PR #236, opened 2025-11-17
doc(memory): add documentation for token based memory and summarization

sahiltyagiii:
# Pull Request

## Type of Change

- [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] 📚 Documentation update
- [ ] 🧹 Code refactoring (no functional changes)
- [ ] ⚡ Performance improvement
- [ ] 🧪 Test coverage improvement
- [ ] 🔧 Build/CI configuration change

## Changes Made

- Documentation added

## AI Provider Impact

- [ ] OpenAI
- [ ] Anthropic
- [ ] Google AI/Vertex
- [ ] AWS Bedrock
- [ ] Azure OpenAI
- [ ] Hugging Face
- [ ] Ollama
- [ ] Mistral
- [ ] All providers
- [x] No provider-specific changes

## Component Impact

- [ ] CLI
- [ ] SDK
- [ ] MCP Integration
- [ ] Streaming
- [ ] Tool Calling
- [ ] Configuration
- [x] Documentation
- [ ] Tests

## Performance Impact

- [x] No performance impact
- [ ] Performance improvement
- [ ] Minor performance impact (acceptable)
- [ ] Significant performance impact (needs discussion)

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive design plan for token-based conversation memory.
  * Documented model-aware context thresholds, configurable session settings, and fallback behavior.
  * Defined approaches for token counting, history preservation, summarization, truncation, caching, and error recovery.
  * Detailed proposed configuration options, APIs, defaults, and backward-compatibility considerations.

### DIFF
--- a/memory-bank/TOKEN-BASED-MEMORY-IMPLEMENTATION-PLAN.md
+++ b/memory-bank/TOKEN-BASED-MEMORY-IMPLEMENTATION-PLAN.md
@@ -1,0 +1,1385 @@
+# Token-Based Conversation Memory Implementation Plan
+
+**Date:** 2025-11-17
+**Status:** Design Phase
+
+---
+
+## Problem Statement
+
+### Critical Issues with Current Implementation
+
+#### 1. **CRITICAL: Single Long Message Breaks System**
+
+The current turn-based implementation **fails catastrophically** when a single message exceeds the token limit:
+
+```typescript
+// User sends a 60k token message
+await neurolink.generate({
+  input: { text: veryLongMessage }, // 60k tokens
+  context: { sessionId: "session-1" }
+});
+
+// Current behavior:
+// 1. Adds message to array (no validation)
+// 2. Checks turn count: 1 turn < 20 turn threshold ✓
+// 3. Sends 60k tokens to LLM → CONTEXT WINDOW ERROR ❌
+```
+
+**Root cause**: Turn-based logic only checks **number of turns**, not **actual token usage**.
+
+**Real-world scenarios**:
+- User pastes entire document as input
+- Multi-paragraph questions
+- Code snippets with extensive context
+- PDF/image analysis with long OCR output
+
+#### 2. Turn-Based Triggering Ignores Token Reality
+
+```typescript
+// Scenario A: 20 short messages (5k tokens total)
+"Hi", "Hello", "How are you?", ...
+// → Triggers summarization unnecessarily ❌
+
+// Scenario B: 5 long messages (80k tokens total)
+Long technical discussions with code examples...
+// → No summarization, exceeds context window ❌
+```
+
+#### 3. Destructive Message Deletion
+
+```typescript
+// After summarization (line 184-187)
+session.messages = [
+  { role: "system", content: summary },
+  ...last10Turns  // Only keeps 20 messages
+];
+// ❌ Deletes 30+ messages permanently
+// ❌ Cannot audit full conversation
+// ❌ Cannot re-summarize with better prompts
+```
+
+#### 4. No Per-Session Flexibility
+
+- Quick Q&A session: Needs low limit (10k tokens)
+- Technical support: Needs high limit (100k tokens)
+- Current: **All sessions use same 20-turn threshold**
+
+---
+
+## Proposed Solution
+
+### High-Level Architecture
+
+```
+┌─────────────────────────────────────────────────────────────┐
+│ SessionMemory                                                │
+├─────────────────────────────────────────────────────────────┤
+│ sessionId: "session-123"                                     │
+│ messages: [                                                  │
+│   { id: "V1StGX", role: "user", content: "..." },           │
+│   { id: "8_Z5jd", role: "assistant", content: "..." },      │
+│   ...                                                        │
+│   { id: "Hi6B-m", role: "user", content: "..." },  ← Pointer│
+│   { id: "summary-yT", role: "system", content: "Summary" }  │
+│   { id: "R8_Z5j", role: "user", content: "..." },           │
+│   ...                                                        │
+│ ]                                                            │
+│ summarizedUpToMessageId: "Hi6B-m"  ← NEW                    │
+│ tokenThreshold: 50000               ← NEW (per-session)     │
+│ lastTokenCount: 48532               ← NEW (cached)          │
+└─────────────────────────────────────────────────────────────┘
+```
+
+### Core Principles
+
+1. **Message-Level Validation**: Check token count BEFORE adding message
+2. **Token-Based Triggering**: Use actual token count, not turn count
+3. **Non-Destructive Storage**: Never delete messages
+4. **Pointer-Based Context**: Only send relevant messages to LLM
+5. **Per-Session Config**: Different limits for different use cases
+6. **Model-Aware Defaults**: Automatically use 80% of each model's context window
+
+### Default Configuration
+
+**Token-based memory is ENABLED BY DEFAULT** with model-aware thresholds:
+
+```typescript
+// No configuration needed - works out of the box!
+const neurolink = new NeuroLink();
+
+await neurolink.generate({
+  provider: 'openai',
+  model: 'gpt-4o',  // 128k context window
+  // → Automatically uses 102,400 tokens (80% of 128k)
+});
+
+await neurolink.generate({
+  provider: 'vertex',
+  model: 'gemini-2.5-flash',  // 1M context window
+  // → Automatically uses 819,200 tokens (80% of 1M)
+});
+
+await neurolink.generate({
+  provider: 'anthropic',
+  model: 'claude-3-5-sonnet-20241022',  // 200k context window
+  // → Automatically uses 160,000 tokens (80% of 200k)
+});
+```
+
+**How it works**:
+1. **Enabled by default**: `NEUROLINK_MEMORY_ENABLED !== "false"`
+2. **Summarization enabled**: `NEUROLINK_SUMMARIZATION_ENABLED !== "false"`
+3. **Dynamic threshold**: Calculates `80%` of model's context window automatically
+4. **Fallback**: Uses `50,000` tokens if model context unknown
+
+**Priority order**:
+- Session override (highest)
+- Environment variable (`NEUROLINK_TOKEN_THRESHOLD`)
+- **Model-based (80% of context) ← DEFAULT**
+- Fallback (50k tokens)
+
+---
+
+## Preventing Context Window Overflow - ALL Cases Covered
+
+### Guarantee: ZERO Context Window Errors
+
+**Requirement**: No matter what happens, the LLM must NEVER receive more tokens than its context window supports.
+
+### Case 1: Single Incoming Message Too Large
+
+**Problem**: User sends 200k token message to GPT-4o (128k context window)
+
+**Solution**: Truncate at ingestion
+
+```typescript
+async validateAndPrepareMessage(content: string, threshold: number): Promise<ChatMessage> {
+  const messageTokens = await countTokens(content);
+  const maxMessageSize = Math.floor(threshold * 0.5); // Max 50% of context window
+
+  if (messageTokens > maxMessageSize) {
+    // TRUNCATE - no choice
+    const truncated = await truncateToTokens(content, maxMessageSize);
+
+    logger.warn('Message truncated', {
+      original: messageTokens,
+      truncated: maxMessageSize,
+      lost: messageTokens - maxMessageSize
+    });
+
+    return {
+      id: nanoid(),
+      content: truncated + "\n\n[Message truncated due to size]",
+      metadata: { truncated: true, originalTokens: messageTokens }
+    };
+  }
+
+  return { id: nanoid(), content };
+}
+```
+
+**Result**: ✅ No single message can exceed 50% of context window
+
+### Case 2: Input to Summarizer Too Large
+
+**Problem**: Need to summarize 500k tokens, but summarizer model (gemini-2.5-flash) only supports 1M input
+
+**Solution**: Chunked summarization
+
+```typescript
+async summarizeSession(session: SessionMemory, threshold: number): Promise<void> {
+  const messagesToSummarize = getMessagesToSummarize(session);
+  const inputTokens = await countTokens(messagesToSummarize);
+
+  // Get summarizer model's context limit
+  const summarizerLimit = TokenUtils.getProviderTokenLimit(
+    config.summarizationProvider,
+    config.summarizationModel
+  );
+
+  let summary: string;
+
+  // If input exceeds summarizer's capacity → CHUNK IT
+  if (inputTokens > summarizerLimit * 0.8) {
+    logger.info('Input too large for summarizer, using chunked approach', {
+      inputTokens,
+      summarizerLimit,
+      chunks: Math.ceil(inputTokens / (summarizerLimit * 0.5))
+    });
+
+    summary = await chunkAndSummarize(messagesToSummarize, summarizerLimit);
+  } else {
+    // Normal single-pass summarization
+    summary = await generateSummary(messagesToSummarize);
+  }
+
+  // Continue with validated summary...
+}
+
+async chunkAndSummarize(messages: ChatMessage[], summarizerLimit: number): Promise<string> {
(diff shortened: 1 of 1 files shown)


memory-bank/TOKEN-BASED-MEMORY-IMPLEMENTATION-PLAN.md
```diff
@@ -0,0 +1,828 @@
...
+export function getConversationMemoryDefaults(): ConversationMemoryConfig {
+  return {
+    enabled: process.env.NEUROLINK_MEMORY_ENABLED === "true",
+    maxSessions: Number(process.env.NEUROLINK_MEMORY_MAX_SESSIONS) || 50,
+    sessionInactivityDays: Number(process.env.NEUROLINK_SESSION_INACTIVITY_DAYS) || 30,
+
+    enableSummarization: process.env.NEUROLINK_SUMMARIZATION_ENABLED === "true",
+    tokenThreshold: Number(process.env.NEUROLINK_TOKEN_THRESHOLD) || DEFAULT_TOKEN_THRESHOLD,
+    summarizationProvider: (process.env.NEUROLINK_SUMMARIZATION_PROVIDER as AIProviderName) || "vertex",
+    summarizationModel: process.env.NEUROLINK_SUMMARIZATION_MODEL || "gemini-2.5-flash",
```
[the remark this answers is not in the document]
sahiltyagiii: yes, 2.5 flash will be used

memory-bank/TOKEN-BASED-MEMORY-IMPLEMENTATION-PLAN.md
```diff
@@ -0,0 +1,828 @@
...
+
+```bash
+# Shared settings
+NEUROLINK_MEMORY_ENABLED=true
+NEUROLINK_MEMORY_MAX_SESSIONS=50
+NEUROLINK_SUMMARIZATION_ENABLED=true
+NEUROLINK_SUMMARIZATION_PROVIDER=vertex
+NEUROLINK_SUMMARIZATION_MODEL=gemini-2.5-flash
+```
+
```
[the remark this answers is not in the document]
sahiltyagiii: yes, 2.5 flash will be used

memory-bank/TOKEN-BASED-MEMORY-IMPLEMENTATION-PLAN.md
```diff
@@ -0,0 +1,828 @@
+# Token-Based Conversation Memory Implementation Plan
+
+**Date:** 2025-11-17
```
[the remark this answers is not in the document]
sahiltyagiii: date is correct

memory-bank/TOKEN-BASED-MEMORY-IMPLEMENTATION-PLAN.md
```diff
@@ -0,0 +1,828 @@
...
+4. **Pointer-Based Context**: Only send relevant messages to LLM
+5. **Per-Session Config**: Different limits for different use cases
+
+---
+
+## Token Counting Infrastructure
+
+### Provider-Specific Implementations
+
+| Provider | Method | Accuracy | Latency | Status |
```
murdore, line 523: @sahiltyagiii add for all provders we support. vertex litellm and others

memory-bank/TOKEN-BASED-MEMORY-IMPLEMENTATION-PLAN.md
```diff
@@ -0,0 +1,828 @@
...
+**Already Available:**
+- ✅ `js-tiktoken@1.0.21` (via langsmith)
+- ✅ `@google/generative-ai@0.24.1`
+- ✅ `@aws-sdk/client-bedrock-runtime`
+
+### Token Counter Interface
+
+```typescript
+// src/lib/services/tokenCounting/types.ts
+export interface TokenCounter {
```
murdore: No interface, only types and that too in the central locations where all other types are defined.

memory-bank/TOKEN-BASED-MEMORY-IMPLEMENTATION-PLAN.md
```diff
@@ -0,0 +1,828 @@
+# Token-Based Conversation Memory Implementation Plan
```
murdore, line 1: Approach needs to mention how it will be enabled by default and what will be the configuration for default. it should be 80% context memory per model being used.

memory-bank/TOKEN-BASED-MEMORY-IMPLEMENTATION-PLAN.md
```diff
@@ -0,0 +1,828 @@
...
+
```
murdore, line 2: We need to add future plans of using multi-modality for summarization of the messages.

memory-bank/TOKEN-BASED-MEMORY-IMPLEMENTATION-PLAN.md
murdore, line 3: Another requirement is how will the implementation work? How do you make sure LLMs don't overflow the token coin text window?

Another requirement would be that for us to be able to identify 80 percent token window fill and keep summarizing till we are able to answer the query, it can be slow but this can be something that we will have to do in the future.

murdore (changes requested):

memory-bank/TOKEN-BASED-MEMORY-IMPLEMENTATION-PLAN.md
sahiltyagiii: @murdore added bhaiya

memory-bank/TOKEN-BASED-MEMORY-IMPLEMENTATION-PLAN.md
sahiltyagiii: okay bhaiya, updated the doc to use types, and add them to centeralized location

memory-bank/TOKEN-BASED-MEMORY-IMPLEMENTATION-PLAN.md
sahiltyagiii: mentioned default config, also will use `getProviderTokenLimit` to get the model's token limit and will use a `0.8x` multiplier for it
updated the doc accordingly

memory-bank/TOKEN-BASED-MEMORY-IMPLEMENTATION-PLAN.md
sahiltyagiii: let's discuss this bhaiya

memory-bank/TOKEN-BASED-MEMORY-IMPLEMENTATION-PLAN.md
sahiltyagiii: for this we will check the token count of message, and if it exceeds we can iteratively summarize it or set summarization prompt to strictly keep response in required tokens. If the message is too big for the summarizer LLM we can break it in chunks to summarize to prevent failing in worst cases.

updated the doc for it

murdore:
Solid plan — the token-vs-turn framing is right, and I like that you've already costed the per-turn `countTokens` latency (the ~100-200ms providers table) and that truncation emits a `logger.warn` + `[Message truncated]` marker rather than failing silently.

Two design points worth resolving before implementation:

1. **Single-oversized-message handling (the headline bug) shouldn't hard-truncate.** The ingestion path truncates a too-large single message to ~50% threshold 'no choice' and continues. Even with the warn + metadata flag, the dropped tokens are unrecoverable and the user gets a silently-shortened answer to a question they didn't actually ask. For the exact scenario this plan opens with (user pastes a 60k/200k-token message), a **reject-with-actionable-error** ('message is X tokens, limit is Y — split it or enable summarization') or **chunk-then-summarize** is safer than truncate, because the caller can act on it. Truncation as a last-resort *send-time* guard (your emergency path) is fine; truncation as the *ingestion* answer to the core use-case is lossy.

2. **'ENABLED BY DEFAULT' (`NEUROLINK_MEMORY_ENABLED !== "false"`) is a behavior change for every existing consumer.** Worth an explicit backward-compat section: does the non-destructive store imply a session-schema change? Do in-flight/existing sessions keep working unchanged? Since this flips default runtime behavior of the SDK, it reads like a major-version bump + a documented opt-out, not a silent default flip.

Both are non-blocking design questions, not code issues (doc-only PR).

murdore:
Hi @sahiltyagiii — thank you for this detailed token-based-memory design doc. Heads up that since this PR was opened, the feature it proposes (token-threshold conversation memory, background summarization, pointer-based context tracking) has actually been implemented and shipped on `release` — see `ConversationMemoryConfig`, `conversationMemoryManager.ts`, `summarizationEngine.ts`, and the `isSummary` / `summarizesFrom` / `summarizedUpToMessageId` fields on `ChatMessage`/`SessionMemory`. So this planning doc is now largely superseded by the real implementation.

If you're up for it, the most useful path would be to repurpose this into user-facing docs that match what actually shipped (and address @murdore's earlier review notes on lossy single-message truncation and the default-enabled behavior). Otherwise we're happy to close it. Could you let us know which you'd prefer? If we don't hear back in a couple of weeks we'll close it as superseded — easy to reopen later.

murdore:
Hi @sahiltyagiii — thanks for the detailed design write-up. A few things to flag:

1. **This is now superseded by shipped code.** Since you opened this, token-threshold conversation memory + background summarization landed on `release` — see `src/lib/types/conversation.ts` (`tokenThreshold`, `summarizedUpToMessageId`, `isSummary`/`summarizesFrom`/`summarizesTo` on `ChatMessageMetadata`/`SessionMemory`), `src/lib/core/conversationMemoryManager.ts`, and `src/lib/context/summarizationEngine.ts`. There's also already published user docs at `docs-site/docs/memory/conversation-memory.md`, `summarization.md`, and `conversation.md`. @murdore flagged this in the thread on 2026-07-08 and asked which way you'd like to take it — we haven't heard back yet.

2. **Please respond to that comment either way:** (a) if you'd like, repurpose this into user-facing docs that describe what actually shipped (the real docs above are a good starting reference, and the shipped design differs from this plan in places — e.g. it's a turn/token hybrid with `maxTurnsPerSession` marked `@deprecated` in favor of `tokenThreshold`, not a pure token-only cutover), or (b) let us know if you're OK with us closing this as superseded — easy to reopen if useful later.

3. **If you do repurpose it**, worth folding in the review feedback already on this PR before it becomes a doc PR: the bot reviewers caught a factual error (Anthropic's SDK has no pre-request token-counting endpoint — it only returns usage post-response, so the 'Count Tokens API' framing needs correcting) and an inconsistent env-var naming scheme (`NEUROLINK_TOKEN_THRESHOLD` vs `NEUROLINK_MEMORY_MAX_TURNS_PER_SESSION`). @murdore's design comment also raised two still-open points worth resolving even for a docs pass: (i) the plan's truncate-on-ingestion behavior for oversized single messages is lossy — document the actual shipped behavior here explicitly since it materially affects users; (ii) whether summarization/memory being on by default is called out as a behavior change in the docs.

No action needed on our end right now beyond a reply from you — but if we don't hear back in the next week or so per @murdore's note, we'll close this as superseded by the shipped implementation.

murdore:
Small doc/code mismatch worth fixing before this lands: the plan repeatedly states memory is enabled by default / opt-out — e.g. "Memory: **ENABLED** (opt-out with `NEUROLINK_MEMORY_ENABLED=false`)" and the sample `enabled: process.env.NEUROLINK_MEMORY_ENABLED !== "false"  // Default: true`.

But the shipped code makes it opt-in: `src/lib/config/conversationMemory.ts` sets `enabled: process.env.NEUROLINK_MEMORY_ENABLED === "true"`, and `docs/features/memory.md` already says "Set `true` to activate memory." Worth aligning the doc to opt-in (`=== "true"`) so it matches what actually shipped. (Low urgency — noting it's a Design-Phase doc under `memory-bank/`, not user-facing.)

murdore:
Re-checked at current head `b926b792664f806afdf25404564e745facc7ee96` — the doc/code mismatch from the 2026-07-16 review is still present verbatim, and the "superseded by shipped code" question from 2026-07-08 / 2026-07-15 is still unanswered, so chasing both rather than re-raising as new:

**1. Opt-in vs. opt-out — still unresolved.** The plan still says:
```
131:1. **Enabled by default**: `NEUROLINK_MEMORY_ENABLED !== "false"`
1053:    enabled: process.env.NEUROLINK_MEMORY_ENABLED !== "false",  // Default: true
1293:- ✅ Memory: **ENABLED** (opt-out with `NEUROLINK_MEMORY_ENABLED=false`)
```
but the shipped `src/lib/config/conversationMemory.ts` uses `enabled: process.env.NEUROLINK_MEMORY_ENABLED === "true"` (opt-in). Same pattern repeats for `NEUROLINK_SUMMARIZATION_ENABLED` at lines 132/1058/1294. None of these lines have changed since the 07-16 comment.

**2. Env-var naming convention — also still unaddressed.** `NEUROLINK_TOKEN_THRESHOLD` (lines 138, 1059, 1096, 1310, 1346) and `NEUROLINK_MEMORY_MAX_TURNS_PER_SESSION` (line 1066/1325) still mix bare-noun and subsystem-prefixed naming, as both coderabbitai and Copilot flagged earlier.

**3. The bigger open question is still the one from 07-08/07-15: repurpose this into user-facing docs matching the shipped turn/token-hybrid implementation, or close as superseded?** That's now ~2 weeks past the "next week or so" window mentioned on 07-15 with no reply. If a decision is made to repurpose rather than close, items 1 and 2 above (plus the still-open Anthropic Count-Tokens-API framing from coderabbitai's line-646 review, and the truncate-on-ingestion / default-enabled design points from the 06-30 review) are the concrete things worth folding into that pass — happy to help however's useful, just flagging it's stalled on a response either way.

murdore:
One design point I don't see raised yet in this thread (current head `b926b792664f806afdf25404564e745facc7ee96`), separate from the still-open opt-in/opt-out, env-var-naming, and "superseded by shipped code" items already flagged above — not repeating those.

`storeConversationTurn` fires summarization without awaiting it, specifically so the next turn isn't blocked (`memory-bank/TOKEN-BASED-MEMORY-IMPLEMENTATION-PLAN.md:814-820`):
```
  // 5. Check if summarization needed (NON-BLOCKING)
  if (this.config.enableSummarization) {
    // Fire and forget - don't block user response
    setImmediate(async () => {
      await this.checkAndSummarize(session, tokenThreshold);
    });
  }
```
`_summarizeSession` reads `startIndex` from the pointer *before* awaiting the summarizer LLM call, then uses that pre-await index to splice the summary in and update the pointer afterward (`:911-913`, `:938`, `:954-959`):
```
  const startIndex = session.summarizedUpToMessageId
    ? this._findMessageIndex(session.messages, session.summarizedUpToMessageId) + 1
    : 0;
  ...
  const summary = await this._generateSummary(messagesToSummarize);
  ...
  const insertIndex = startIndex + splitIndex;
  session.messages.splice(insertIndex, 0, summaryMessage);
  session.summarizedUpToMessageId = lastSummarized.id;
```
Because the fire-and-forget call isn't awaited, a fast follow-up turn on the same session can cross the threshold again and kick off a second `_summarizeSession` before the first one's `splice`/pointer update lands — the second call computes its own `insertIndex` against the pre-splice array. Could that leave `session.messages` with the pointer not immediately followed by a summary message? `buildContextFromPointer`'s "Verify first message is the summary" check (`:885-886`) only warns and falls back to slicing from the pointer rather than preventing this on the write side. Given the "GUARANTEE"/"ABSOLUTE GUARANTEE" language used elsewhere in this doc for token-overflow safety, is single-flight/locking per `sessionId` around `_summarizeSession` intentionally deferred to implementation, or worth calling out explicitly here?

PR #870, opened 2026-03-12
fix(vision): allow unknown models for proxy providers in vision check

pdogra1299:
## Problem

When using the `litellm` provider with a custom/private model name (e.g. `private-large`), NeuroLink threw a pre-flight error even though the underlying model supported vision:

```
Provider litellm with model private-large does not support vision processing.
Supported providers: openai, google-ai, anthropic, azure, vertex, litellm, ...
```

## Root Cause

`ProviderImageAdapter.supportsVision()` and `validateVisionSupport()` use a hardcoded `VISION_CAPABILITIES` allowlist to validate model names. The check uses substring matching against known public model names — custom/private models never match.

**LiteLLM and OpenRouter are proxy providers** that route requests to arbitrary underlying models. NeuroLink cannot know in advance what models are deployed behind them, so a static allowlist is inherently incomplete for these providers.

## Fix

Added a `PROXY_PROVIDERS` constant (`["litellm", "openrouter"]`). In both vision check locations, if the model name doesn't match the allowlist **and** the provider is a proxy, the request is passed through instead of rejected.

The underlying provider (LiteLLM/OpenRouter) will naturally surface an error if the model truly doesn't support vision — no silent failures.

## Behaviour After Fix

| Scenario | Before | After |
|---|---|---|
| `litellm` + known model (e.g. `gpt-4o`) | ✅ passes | ✅ passes |
| `litellm` + custom model (e.g. `private-large`) | ❌ pre-flight error | ✅ passes through |
| `openrouter` + unknown model | ❌ pre-flight error | ✅ passes through |
| Non-proxy provider + unsupported model | ❌ throws (correct) | ❌ throws (unchanged) |

## Files Changed

- `src/lib/adapters/providerImageAdapter.ts` — patched `supportsVision()` and `validateVisionSupport()`
- `src/lib/providers/googleAiStudio.ts`, `googleVertex.ts` — pre-existing prettier formatting fixes (unblocked pre-commit hook)

## Summary by CodeRabbit

* **New Features**
  * Proxy providers now allow additional vision models by deferring validation to the underlying provider instead of relying solely on a static allowlist, reducing false rejections for unknown models.

* **Style**
  * Minor code formatting refinements for clearer conditional expressions.

### DIFF
--- a/src/lib/adapters/providerImageAdapter.ts
+++ b/src/lib/adapters/providerImageAdapter.ts
@@ -45,6 +45,26 @@ const IMAGE_LIMITS = {
 } as const;
 
 /**
+ * Proxy providers that route to arbitrary underlying models.
+ * Vision capability cannot be statically determined for these — pass requests
+ * through and let the underlying provider surface errors if needed.
+ */
+const PROXY_PROVIDERS = new Set(["litellm", "openrouter"]);
+
+/**
+ * Normalize provider name/alias to its canonical form for vision checks.
+ */
+function normalizeVisionProvider(provider: string): string {
+  const lower = provider.toLowerCase();
+  switch (lower) {
+    case "or":
+      return "openrouter";
+    default:
+      return lower;
+  }
+}
+
+/**
  * Vision capability definitions for each provider
  */
 const VISION_CAPABILITIES = {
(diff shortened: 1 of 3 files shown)


murdore:
@coderabbitai review

PR #900, opened 2026-03-25
fix(proxy): crash recovery with launchd daemon and setup overhaul

murdore:
## Summary

- **Setup installs launchd daemon** on macOS instead of starting a foreground process that dies with the terminal. Added `--no-service` flag for foreground mode.
- **Guard process restarts proxy** via `launchctl kickstart` before falling back to Claude settings cleanup.
- **Manual start blocked when launchd is active** — prevents port conflicts that left both processes broken.
- **Signal-aware shutdown exit codes** — SIGINT (Ctrl+C) exits 0 so launchd doesn't restart; SIGTERM exits 1 so launchd does restart. Fixes `KeepAlive.SuccessfulExit=false` actually working.
- Hoisted plist constants to file top to avoid TDZ errors.
- Added `managedBy` field to `ProxyState` type.
- Added Claude Proxy to features doc sidebar.

## Test plan

- [ ] `neurolink proxy setup` installs launchd service and proxy auto-starts as daemon
- [ ] Kill the proxy process (`kill <pid>`) and verify launchd restarts it within 15s
- [ ] `neurolink proxy start` refuses with clear error when launchd service is active
- [ ] `neurolink proxy setup --no-service` starts proxy in foreground (old behavior)
- [ ] Ctrl+C (SIGINT) stops proxy and launchd does NOT restart
- [ ] `neurolink proxy uninstall` then `neurolink proxy start` works for manual mode
- [ ] `neurolink proxy status` shows `managedBy: launchd` or `managedBy: manual`

## Summary by CodeRabbit

* **New Features**
  * Introduced Claude Proxy with OAuth pooling, rate-limit failover, token refresh, and crash recovery
  * Added macOS daemon service integration for persistent proxy management
  * Added `--no-service` flag to proxy setup for manual-only operation
  * Enhanced restart health verification with improved failover handling


murdore:
@coderabbitai full review

PR #972, opened 2026-04-18
fix(tools): tool filter not working

Yaswanth-2874:
Description
Fixed a bug where tool filtering was not being respected in Google AI Studio and Google Vertex providers. Previously, SDK tools were automatically merged with user-provided tools, ignoring the user's tool selection filter.
Related Issues
N/A - Internal bug fix
Type of Change
Motivation and Context
When users explicitly passed a filtered set of tools via options.tools, the providers were incorrectly merging them with all available SDK tools (from getAllTools()), causing the filter to be ignored. This fix ensures only user-specified tools are used.
Changes Made
- Modified src/lib/providers/googleAiStudio.ts:
  - Removed automatic SDK tool merging in stream and generate methods
  - Tools now respect user's explicit filter in options.tools
  - Updated logging to reflect actual tool count
  
- Modified src/lib/providers/googleVertex.ts:
  - Same fix applied for stream and generate methods
  - Consistent tool handling across both providers
Breaking Changes
- [x] No breaking changes
Testing
- [x] Unit tests added/updated
- [x] Tested with multiple providers: google-ai-studio, google-vertex
- [x] Manual testing completed
Code Quality
- [x] Code follows the project's style guidelines (ESLint passes)
- [x] Self-review of code completed
- [x] TypeScript strict mode compliance
- [x] Commit message follows format: type(scope): description
Documentation
N/A - Bug fix, no API changes
Dependencies
- [x] No dependency changes
Performance Impact
- [x] No performance impact
Security Considerations
- [x] No security implications

### DIFF
--- a/src/lib/providers/googleAiStudio.ts
+++ b/src/lib/providers/googleAiStudio.ts
@@ -539,17 +539,16 @@ export class GoogleAIStudioProvider extends BaseProvider {
     // Need to check early if we should route to native SDK
     const gemini3CheckShouldUseTools =
       !options.disableTools && this.supportsTools() && !wantsStructuredOutput;
-    const optionTools = options.tools || {};
-    const sdkTools = gemini3CheckShouldUseTools ? await this.getAllTools() : {};
-    const combinedToolCount =
-      Object.keys(optionTools).length + Object.keys(sdkTools).length;
-    const hasTools = gemini3CheckShouldUseTools && combinedToolCount > 0;
+    const tools = options.tools || {};
+
+    const hasTools =
+      gemini3CheckShouldUseTools && Object.keys(tools).length > 0;
 
     if (isGemini3Model(gemini3CheckModelName) && hasTools) {
       // Merge SDK tools into options for native SDK path
       let mergedOptions = {
         ...options,
-        tools: { ...sdkTools, ...optionTools },
+        tools: tools,
       };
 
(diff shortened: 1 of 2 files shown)


BoraYaswanthReddy:
@coderabbitai

> src/lib/providers/googleVertex.ts (1)
> 1296-1340: Dead field in return type: baseToolCount is now a hard-coded 0.
> 
> Since the SDK merge was removed, baseToolCount is always 0 at line 1339. The field still appears in the return-type contract of resolveAISDKStreamTools (line 1296) but no callers can meaningfully use it. Consider dropping the field (and updating callers) to keep the tool-resolution surface aligned with the "options-only" model.

This has nothing to do with this pr objective, it doesnt even belong to this pr

> src/lib/providers/googleAiStudio.ts (1)
> 537-632: Variable shadowing: tools is declared twice in executeStream.
> 
> const tools = options.tools || {} at line 539 is shadowed by let tools: Record<string, Tool> | undefined inside the try block at line 620. The two bindings refer to different values (the outer one is the raw merged map for Gemini 3 routing; the inner one is the sanitized map for streamText), which makes the control flow harder to follow. Renaming the outer binding (e.g., routingTools) avoids the shadow without behavior change.

This is good suggestion but naming would be better if it is consistent everywhere
This logic is clear enough , we dont need to consider renaming

PR #973, opened 2026-04-18
refactor(processors): dynamic-import exceljs, move to optionalDependencies

murdore:
## Summary
- Convert static exceljs import to lazy dynamic import in ExcelProcessor
- Move exceljs (22 MB) from dependencies to optionalDependencies
- Narrow catch to ERR_MODULE_NOT_FOUND with install hint

## Impact
- Default installs: no change (optional deps install by default)
- Excel processing works when exceljs is installed
- 22 MB savings for --no-optional consumers

## Test plan
- [ ] pnpm run build succeeds
- [ ] pnpm run check passes
- [ ] Excel file processing works


murdore:
@coderabbitai full review

PR #988, opened 2026-04-25
test(observability): regression guard for mcp.tool.name on Pipeline B spans

murdore:
## Summary

Curator P2-5 reported that NeuroLink-emitted tool spans did not carry a tool-name attribute, breaking per-tool error analysis in Langfuse. The P1-3 fix in 9.55.11 already plumbed the attribute on every internal span:

| Span | Attribute key | Source |
|---|---|---|
| `neurolink.tool.execute` | `tool.name` | `src/lib/neurolink.ts:8765` |
| `neurolink.tool.registry.execute` | `gen_ai.tool.name` | `src/lib/mcp/toolRegistry.ts:332` |
| `neurolink.tools.execute_custom` | `tool.name` | `src/lib/core/modules/ToolsManager.ts:542` |

This PR adds a regression-guard suite so it can't silently regress. **No production-code change.**

The suite uses real OpenTelemetry (`InMemorySpanExporter` from `@opentelemetry/sdk-trace-base`) and the public `sdk.executeTool()` API.

## Verification

```bash
pnpm run build
npx tsx test/continuous-test-suite-issue-05-mcp-tool-name-spans.ts
```

Expected: `Results: 4 passed, 0 failed, 0 skipped`.

## Test plan

- [x] Suite runs and passes against current `release`
- [x] No source-code changes — bug is already fixed
- [x] Regression-guard documents the contract
- [x] One small follow-up (out of scope): attribute key differs between spans (`tool.name` vs `gen_ai.tool.name`); a downstream normaliser may help if Langfuse queries filter on a single key

## Tests

* Introduced new OpenTelemetry span capture helper utility for test infrastructure, enabling precise validation of real-time telemetry data and span attributes throughout the system
* Added comprehensive test suite to validate that tool-name attributes are properly captured in execution spans across multiple scenarios, including successful execution, error handling, and non-existent tool requests

### DIFF
--- a/test/continuous-test-suite-issue-05-mcp-tool-name-spans.ts
+++ b/test/continuous-test-suite-issue-05-mcp-tool-name-spans.ts
@@ -1,0 +1,277 @@
+#!/usr/bin/env tsx
+/**
+ * Continuous Test Suite: Issue #5 — mcp.tool.name on Pipeline B spans
+ *
+ * Curator P1-5: NeuroLink-emitted spans (`neurolink.tool.execute`,
+ * `neurolink.tool.registry.execute`, `neurolink.tools.execute_custom`) must
+ * carry `mcp.tool.name` / `tool.name` / `gen_ai.tool.name` so per-tool
+ * error analysis works in Langfuse.
+ *
+ * Strategy: REAL OTel via InMemorySpanExporter. Register a real custom tool
+ *           on the public SDK; invoke it via sdk.executeTool() (deterministic,
+ *           no LLM needed). Inspect every span emitted; assert the tool-name
+ *           attribute is present.
+ *
+ * Run: pnpm run build && npx tsx test/continuous-test-suite-issue-05-mcp-tool-name-spans.ts
+ */
+import "dotenv/config";
+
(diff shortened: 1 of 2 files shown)


murdore:
Force-pushed `5581d2a8` addressing 3 of CodeRabbit's findings:

- **`dumpAllSpans`** now fails when zero spans are captured AND asserts that a `neurolink.tool.execute` span exists with a tool-name attribute, rather than always reporting PASS with `(no spans captured)`.
- **bogus-tool FAIL message** no longer prints `undefined=undefined` when no attribute is found — it now says `attr not found (expected does_not_exist_xyz)`.
- **`installSpanCapture`** snapshots the module-level exporter into a local `captured` const after the null-check, dropping the four `exporter!` non-null assertions. Added a doc comment explaining the one-shot install / shared-reset semantics.

**Skipped** the doc table line-range nit — that doc was already removed from this PR in the prior commit (`a51ae8d6`), so the file isn't tracked here.

Suite re-verified: 4/4 PASS, 0 fail.

@coderabbitai review

murdore:
@coderabbitai full review

PR #989, opened 2026-04-25
fix(observability): emit generation:end exactly once on stream finalize

murdore:
## Summary

Curator P2-4: cost listeners that subscribe to `generation:end` previously received **zero** events from `sdk.stream()` calls. The doc described "fires twice"; the bug on shipped 9.56.x is the opposite direction — `stream()` emitted `stream:complete` but never `generation:end`, leaving any listener (cost-tracker, audit log, alerting) with nothing.

## Reproduction (real providers, before fix)

```
generate / vertex            count=1   PASS
stream  / vertex             count=0   FAIL  ← bug
generate / google-ai-studio  count=1   PASS
stream  / google-ai-studio   count=0   FAIL  ← bug
generate / litellm           count=1   PASS
stream  / litellm            count=0   FAIL  ← bug
```

After fix: all stream rows return `count=1`.

## Fix

In `runStandardStreamRequest`'s `processedStream` generator, emit `generation:end` exactly once in the `finally` block with the final stream state (provider, model, content, usage, finishReason, toolsUsed, prompt, temperature, maxTokens, success, error, `pipelineAHandled: true`). Hoist `resolvedUsage` to the generator scope so it's available in `finally`. The event payload mirrors `generate()`'s shape so listeners receive a consistent contract across both APIs.

## Backward compatibility

Additive only. Listeners that previously received zero events on streams now receive one event with the same shape `generate()` produces. Any listener that already expected the event will start working; any listener that didn't will still ignore the new event.

## Verification

```bash
pnpm run build
npx tsx test/continuous-test-suite-issue-04-generation-end-dedup.ts
```

Expected: 6 passed (3 generate + 3 stream); the OpenAI rows may SKIP/FAIL on env-specific quota / tool-injection issues unrelated to this fix.

## Test plan

- [x] Suite passes against fixed `release` for vertex, google-ai-studio, litellm
- [x] No regression on `generate()` paths (still count=1)
- [x] `pipelineAHandled: true` flag preserved for Langfuse exporter dedup

## Release Notes

* **Bug Fixes**
  * Fixed event handling to ensure stream completion events are emitted exactly once per operation in concurrent scenarios.

* **Tests**
  * Added test suite validating event emission behavior across multiple AI providers.

### DIFF
--- a/src/lib/neurolink.ts
+++ b/src/lib/neurolink.ts
@@ -97,6 +97,7 @@ import type {
   MCPTool,
   RoutingDecision,
   MetricsTraceContext,
+  StreamGenerationEndContext,
 } from "./types/index.js";
 import { emergencyContentTruncation } from "./context/emergencyTruncation.js";
 import {
@@ -475,6 +476,41 @@ function isNonRetryableProviderError(error: unknown): boolean {
  * same NeuroLink instance would clobber each other's trace context.
  */
 const metricsTraceContextStorage = new AsyncLocalStorage<MetricsTraceContext>();
+
+/**
+ * Curator P2-4 dedup (concurrency-safe): native providers emit
+ * `generation:end` on the shared SDK emitter. We attach a fresh
+ * mutable `dedupContext` object directly to the per-call
+ * `StreamOptions` (under `_streamDedupContext`) so each stream gets
+ * its own instance — concurrent streams have different option objects
+ * and therefore different contexts, so they cannot interfere.
+ *
+ * Native provider emit sites read `options._streamDedupContext` and
+ * flip `.providerEmitted = true` before emitting; the orchestration's
+ * finally block reads the same closed-over reference and skips its
+ * own emit when the flag is set.
+ *
+ * This avoids the AsyncLocalStorage approach which doesn't reliably
+ * propagate through async-generator yield boundaries when iteration
+ * happens from outside the original `run()` scope (e.g. when the
+ * consumer drives `for await of result.stream` after `sdk.stream(...)`
+ * returns).
+ */
+export const STREAM_DEDUP_CONTEXT_KEY = "_streamDedupContext" as const;
+
+/**
+ * Native providers call this from their `generation:end` emit sites,
+ * passing the same `options` object they received. Safe no-op when
+ * the field isn't set.
+ */
+export function markStreamProviderEmittedGenerationEnd(
+  options: { _streamDedupContext?: StreamGenerationEndContext } | undefined,
+): void {
+  const ctx = options?._streamDedupContext;
+  if (ctx) {
+    ctx.providerEmitted = true;
+  }
+}
 
 export class NeuroLink {
   private mcpInitialized = false;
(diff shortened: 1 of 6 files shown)


murdore:
Force-pushed `db8d0932` after applying the same recipe used on #988:
- **Rebased** onto latest `origin/release` (now `a66ed14c`).
- **Dropped** the per-issue doc `docs/curator-feedback-fixes/issue-04-stream-generation-end.md` from the PR (kept locally for reference; not tracked).

Final diff:
```
src/lib/neurolink.ts                               |  60 ++++++-
test/continuous-test-suite-issue-04-generation-end-dedup.ts | 179 +++++++++++
test/helpers/envGuard.ts                           |  43 +++++
3 files changed, 281 insertions(+), 1 deletion(-)
```

Suite re-verified post-rebase: 6 PASS (vertex/google-ai-studio/litellm × generate+stream), 1 SKIP (OpenAI quota), 1 FAIL pre-existing OpenAI tool-injection env issue unrelated to this fix.

@coderabbitai full review

murdore:
Force-pushed addressing reviewer Finding #1 (double-emit on native Google streams):

The native Gemini 3 paths in `googleVertex.ts:2339` and `googleAiStudio.ts:1042/1076` already emit `generation:end` themselves — they're the only Pipeline B observation source for those paths (per the comment in `initializeMetricsListeners` at neurolink.ts:3083: 'Native providers (Bedrock, Ollama, Gemini 3) do NOT set [pipelineAHandled] — Pipeline B remains their only observation source').

**Fix**: register a counter listener at the top of `runStandardStreamRequest` and only fire the orchestration emit when the counter is zero (no native emit happened during the stream's lifetime). This preserves:
- AI-SDK paths: orchestration emit fires (count=1, our richer payload with `pipelineAHandled: true`)
- Native paths: provider emit fires (count=1, original payload, Pipeline B observability preserved)
- Exactly-once contract on both

Suite re-verified: 6 PASS (vertex/google-ai-studio/litellm × generate+stream count=1 each).

@coderabbitai full review

murdore:
Force-pushed addressing reviewer Finding #1 follow-up (concurrency-safe dedup):

The previous fix used a counter-listener on the shared SDK emitter and would mis-attribute a concurrent unrelated stream's `generation:end` to this stream, suppressing the orchestration emit incorrectly.

**New approach**: AsyncLocalStorage scoped per-stream:
- New `streamGenerationEndContext` ALS in `neurolink.ts` and matching `StreamGenerationEndContext` type in `src/lib/types/streamDedup.ts` (per CLAUDE.md rule 2).
- Exported helper `markStreamProviderEmittedGenerationEnd()` — native providers call it from their emit sites; safe no-op outside a stream context.
- `runStandardStreamRequest` wraps the `processedStream` IIFE in `streamGenerationEndContext.run(dedupContext, ...)` so any native emit within this stream's async chain flips the per-stream flag.
- Native emit sites in `googleVertex.ts:2339` and `googleAiStudio.ts:1042/1076` now call `markStreamProviderEmittedGenerationEnd()` before emitting.
- Concurrent streams on the same SDK instance get independent ALS scopes — they cannot suppress each other.

Suite re-verified: 6 PASS (vertex/google-ai-studio/litellm × generate+stream count=1).

@coderabbitai full review

murdore:
Force-pushed addressing the reviewer's concern that AsyncLocalStorage doesn't reliably propagate through async-generator yield boundaries.

**New approach**: drop ALS entirely. Attach a fresh mutable `dedupContext` object directly onto `enhancedOptions._streamDedupContext`. Native providers receive these options through their normal call chain and call `markStreamProviderEmittedGenerationEnd(options)`, which mutates the same object the orchestration's finally block reads via closure capture.

**Why it's concurrency-safe**: each `sdk.stream()` call creates its own `enhancedOptions` object → its own dedupContext instance. Concurrent streams literally pass through different option objects so their contexts cannot interfere. No ALS, no shared listeners, no async-chain assumptions.

Wire-up:
- `src/lib/neurolink.ts:6572-6586` attaches `dedupContext` to `enhancedOptions._streamDedupContext` and reads it directly in the finally block (closure capture).
- `markStreamProviderEmittedGenerationEnd(options)` now takes options and mutates `options._streamDedupContext.providerEmitted = true`.
- `googleVertex.ts:2345` and `googleAiStudio.ts:1047/1083` pass their `options` / `params.options` to the helper.

Suite re-verified: 6 PASS (vertex/google-ai-studio/litellm × generate+stream count=1).

@coderabbitai full review

PR #992, opened 2026-04-25
feat(fallback): providerFallback callback + modelChain config

murdore:
## Summary

Curator P2-3: SDK consumers had no way to plug in their own fallback policy when a provider returned model-access-denied. The existing `fallbackProvider` (BZ-1341) is a single string driven from env vars; not suitable for runtime / per-call control.

## New public surface

Two opt-in config knobs (instance-level and per-call):

```ts
new NeuroLink({
  providerFallback: async (err) => ({ model: "kimi-latest" }),
  modelChain:       ["sonnet-4-5", "kimi-latest", "glm-latest"],
});

await sdk.generate({
  provider: "litellm",
  model: "sonnet-4-5",
  input: { text: "hi" },
  providerFallback: ...,   // per-call override
  modelChain: ...,         // per-call override
});

sdk.getEventEmitter().on("model.fallback", (e) => {
  // { requestedProvider, requestedModel, fallbackProvider, fallbackModel, reason, kind, timestamp }
});
```

## Orchestration

`runWithFallbackOrchestration()` wraps `generate()` and `stream()`:

1. Run inner. Return on success.
2. If error AND `looksLikeModelAccessDenied(error)` AND a callback or chain is configured, walk the orchestration:
   - Resolve effective callback (per-call > instance > synthesised from `modelChain`).
   - For each attempt: invoke callback → emit `model.fallback` event → re-call inner with `{provider, model}`.
   - Stop on first success, on `null` callback return, or after `modelChain.length+5` attempts.

`looksLikeModelAccessDenied` matches both Issue #1's typed `ModelAccessDeniedError` (by `name`/`code`) **and** raw message strings ("team not allowed", "team can only access"). Works whether or not Issue #1 has merged.

## Reproduction (real LiteLLM, denied=`sonnet-4-5`, allowed=`open-large`)

```
Before:
[FAIL] 3.0 type surface — neither providerFallback nor modelChain in dist .d.ts
[FAIL] 3.1 callback     — callback never invoked despite real denial
[FAIL] 3.3 modelChain   — chain ignored
[FAIL] 3.5 event        — events=0
Results: 0 passed, 4 failed

After:
[PASS] 3.0 type surface — providerFallback=true, modelChain=true
[PASS] 3.1 callback      — callbackFired=1; resultModel=open-large
[PASS] 3.3 modelChain    — chain progressed; resultModel=open-large
[PASS] 3.5 event         — events=1
Results: 4 passed, 0 failed
```

## Backward compatibility

Additive only. When neither `providerFallback` nor `modelChain` is set, the orchestration is a no-op — existing `generate()`/`stream()` behavior unchanged.

## Verification

```bash
pnpm run build
npx tsx test/continuous-test-suite-issue-03-fallback-hook.ts
```

## Test plan

- [x] 4/4 reproductions pass against real LiteLLM
- [x] Per-call overrides (test 3.6/3.7 in plan) work — covered by 3.1/3.3 with explicit per-call options
- [x] No behavior change when both options are unset
- [x] Compatible with Issue #1's typed `ModelAccessDeniedError` AND with raw message-match (when #1 hasn't merged)

## Release Notes

* **New Features**
  * Added automatic fallback mechanism when model access is denied, enabling configuration of alternative providers and models
  * Support for model chains to sequentially attempt fallback models on access denial
  * Fallback options configurable at SDK initialization level or overridden per request

* **Tests**
  * Added comprehensive test suite for provider fallback functionality

### DIFF
--- a/src/lib/neurolink.ts
+++ b/src/lib/neurolink.ts
@@ -310,6 +310,40 @@ function mcpCategoryToErrorCategory(
  * For example, a NOT_FOUND error for a model causes 6 retries of a 418KB
  * message, wasting ~628,000 tokens and adding 10+ seconds of latency.
  */
+/**
+ * Curator P2-3: detect model-access-denied without requiring the typed
+ * ModelAccessDeniedError class to be present (Issue #1 ships that class
+ * separately). Matches LiteLLM "team not allowed" / "team can only access
+ * models=[...]" plus typed-error markers when present.
+ */
+function looksLikeModelAccessDenied(error: unknown): boolean {
+  if (!error) {
+    return false;
+  }
+  const e = error as { name?: string; code?: string; message?: string };
+  if (e.name === "ModelAccessDeniedError") {
+    return true;
+  }
+  if (e.code === "MODEL_ACCESS_DENIED") {
+    return true;
+  }
+  const msg =
+    typeof e.message === "string"
+      ? e.message
+      : error instanceof Error
+        ? error.message
+        : String(error);
+  if (!msg) {
+    return false;
+  }
+  const lower = msg.toLowerCase();
+  return (
+    (lower.includes("team") && lower.includes("not allowed")) ||
+    lower.includes("team can only access") ||
+    /not\s+allowed\s+to\s+access\s+(this\s+)?model/i.test(msg)
+  );
+}
+
 function isNonRetryableProviderError(error: unknown): boolean {
   // Check for typed error classes from providers
   if (error instanceof InvalidModelError) {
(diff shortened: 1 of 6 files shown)


murdore:
Force-pushed addressing reviewer Finding #2 + applying same recipe as the rest:

- **Rebased** onto latest `origin/release`.
- **Finding #2 fix**: added `streamWithIterationFallback()` wrapper around `sdk.stream()`. The standard `runWithFallbackOrchestration` only catches errors thrown while the StreamResult is being created; many providers (LiteLLM at `litellm.ts:517`) throw inside the async generator during iteration. The wrapper now:
  - Forwards chunks transparently while iteration succeeds.
  - On iteration error AND zero chunks yielded AND `looksLikeModelAccessDenied`, resolves the next callback / chain target, emits `model.fallback` (with `phase: "iteration"` for distinguishability), and recurses into `executeStreamRequest` with the new model.
  - On error after chunks were already yielded, propagates — mid-stream switch isn't safe (consumer has half a response).
  - Only wraps when fallback config is present; otherwise returns the original result unchanged (no behavior change for callers that don't opt in).

Suite re-verified: 4/4 PASS.

@coderabbitai full review

PR #994, opened 2026-04-25
fix(conversation-memory): stop persisting abort sentinel + typed AbortError + read-time filter (SI-069/SI-071)

murdore:
## Summary

Closes the production memory-pollution bug surfaced by Curator SI-069 (Habeeb, 2026-04-16) and SI-071 (Sahil Tyagi, 2026-04-18), where TARA was emitting the literal string `[generation was interrupted]` to end users in Slack.

- **Bug 1**: `handleGenerateTextInternalFailure`'s abort branch was synthesising a fabricated assistant turn into Redis-backed conversation memory on every aborted generate. After ~6+ such entries accumulated, instruction-tuned models (Claude Sonnet 4-6 on Vertex) treated the pattern as a few-shot example and emitted the sentinel as their response.
- **Bug 2**: There was no upstream defence against existing polluted Redis sessions. Every consumer of the SDK had to build their own sanitizer (Curator did — 80+ lines).
- **Architectural gap**: `ErrorCategory` enum had no `ABORT` value, so consumers were forced to message-string match `DOMException`s to detect cancellation.

This PR fixes all three at the source. Net diff: **+183 / −39** in production code, zero public-API breaks.

### 1. No fabricated content on aborts (`src/lib/neurolink.ts`)
`handleGenerateTextInternalFailure` no longer calls `storeConversationTurn` in the abort branch. The previous "needed for title generation" justification was incorrect — `RedisConversationMemoryManager.generateConversationTitle` reads `userMessage` only, never the assistant turn. Aborts now leave memory untouched, identical to every other failure category (auth, quota, network, validation, overflow, timeout). Symmetric error contract.

### 2. Read-time filter at the prompt-builder boundary (`src/lib/utils/conversationMemory.ts`)
`getConversationMessages` (the single function that turns stored memory into the messages array sent to providers) now drops:
- assistant turns with empty / whitespace-only content **and** no tool activity
- assistant turns matching the legacy sentinel `[generation was interrupted]`

This is defence-in-depth and **self-heals every already-polluted Redis session on next read** — no migration script needed. Polluted sessions are flagged with the `neurolink.memory.polluted_turns_dropped` span attribute so Langfuse traces show the cleanup happening.

Tool-bearing assistant turns are explicitly preserved (empty surface text + `tool:start`/`tool:end` events stay), to keep `repairToolPairs` invariants intact. `tool_call`/`tool_result` messages with empty content are also preserved (they legitimately have empty content per `redisConversationMemoryManager.ts:1870`).

### 3. Typed AbortError (`src/lib/constants/enums.ts`, `src/lib/utils/errorHandling.ts`, `src/lib/neurolink.ts`)
- New `ErrorCategory.ABORT = "abort"`
- New `ERROR_CODES.OPERATION_ABORTED = "OPERATION_ABORTED"`
- New `ErrorFactory.aborted(originalError)` factory
- `isAbortError()` extended to recognise the typed shape

Aborts now throw `NeuroLinkError` with `category: "abort"`, `code: "OPERATION_ABORTED"`, `retriable: false`. Consumers can branch on `error.category` instead of message-string matching.

**Backwards compatibility preserved**: `isAbortError()` continues to detect raw `DOMException`/`AbortError` shapes, and inner SDK paths still throw `DOMException` from short-circuit checks. So existing callers that match on `error.name === "AbortError"` continue to work without changes — incremental migration is safe.

### 4. Telemetry parity (`src/lib/neurolink.ts`)
The `executeGenerateTextInternalWithSpan` catch block was setting `status: ERROR` for all failures, which made `ContextEnricher` map aborts to Langfuse `level=ERROR`. Aborts are user-initiated cancellations, not system faults — they should be `WARNING`. Now:

- **Aborts**: stamp `ai.finishReason="aborted"` + `neurolink.aborted=true`, leave span status unset → `ContextEnricher.applyNonErrorLangfuseLevel` maps to `level=WARNING` with `statusMessage="Generation aborted by client"`. Matches Curator's existing telemetry-gaps Issue 5a expectations.
- **Non-abort errors**: `recordException(error)` + `setStatus(ERROR)` (per OTEL conventions; previously we only did `setStatus`).

### Side-effect bug also fixed
`handleGenerateTextInternalFailure` was calling `emitter.emit("error", err)` unconditionally. Node's `EventEmitter` rethrows the original error from inside `emit("error", e)` if no `"error"` listener is registered, which **short-circuited the catch block before the typed-error wrap could run**. Now gated by `listenerCount("error") > 0`. Discovered while debugging the wrap; fixed inline because it's the same code path.

### Belt-and-braces
The upper-layer `storeConversationTurn` guard (`src/lib/utils/conversationMemory.ts:230-240`) now also rejects content equal to the legacy sentinel. Even if a future caller bypasses Fix 1, the storage layer refuses to persist the string. Defence in depth.

## Bug-locations summary

| Layer | File | Line(s) | What changed |
|---|---|---|---|
| Producer (the bug) | `src/lib/neurolink.ts` | 4706–4745 | Removed sentinel write + log message updated |
| Span lifecycle | `src/lib/neurolink.ts` | 4486-ish | Aborts no longer set ERROR status |
| Wrap to typed error | `src/lib/neurolink.ts` | runGenerateTextInternalFlow.catch | Convert raw DOMException → NeuroLinkError(ABORT) |
| EventEmitter rethrow | `src/lib/neurolink.ts` | end of failure handler | `listenerCount("error") > 0` gate |
| Read filter | `src/lib/utils/conversationMemory.ts` | `getConversationMessages` | Drop polluted assistant turns, preserve tool-bearing |
| Storage guard | `src/lib/utils/conversationMemory.ts` | `storeConversationTurn` | Reject sentinel writes too |
| Typed error | `src/lib/utils/errorHandling.ts` | new `ErrorFactory.aborted` + `OPERATION_ABORTED` code + `isAbortError` extension | |
| Enum | `src/lib/constants/enums.ts` | `ErrorCategory` | + `ABORT = "abort"` |

## Test plan

New suite: `test/continuous-test-suite-session-memory-bugs.ts` (run via `pnpm run test:session-memory-bugs`). 5 cases that pin every layer of the fix:

- [x] **Test 1** — Pre-aborted signal + `getSessionMessages` inspection: memory unchanged, no sentinel entry. **Deterministic**, no provider call needed for the assertion.
- [x] **Test 2** — Pre-pollute session with 6 sentinel turns + 1 empty turn via `setSessionMessages`, then ask "What is the capital of France?". Real LLM call (vertex/gemini-2.5-flash). Asserts model response is not the sentinel.
- [x] **Test 3** — Direct call to `getConversationMessages` utility with polluted history. Asserts 2 sentinel + 2 empty assistant turns are filtered out, valid turns preserved. **Deterministic**, no LLM.
- [x] **Test 4** — Pre-aborted signal, catch the thrown error, assert `error.name === "NeuroLinkError"`, `category === "abort"`, `code === "OPERATION_ABORTED"`, `retriable === false`. **Deterministic**, no LLM.
- [x] **Test 5** — Pre-pollute with a tool-bearing assistant turn (empty surface text + `tool:start` event) alongside sentinel + plain-empty turns. Assert filter drops sentinel + plain-empty but **preserves** the tool-bearing turn. Regression guard for `repairToolPairs` invariants. **Deterministic**, no LLM.

Latest run against vertex:

```
Final Results: 5 passed, 0 failed, 0 skipped (5 total) in 27s
```

### Existing-suite regression checks

- [x] `test/continuous-test-suite-telemetry-gaps.ts`: **15/15 PASS** (Issue 5a "aborted generation → level=WARNING + abort statusMessage" still passes — no regression on the existing ContextEnricher-only test).
- [x] Sanity: ad-hoc 2-turn generate with vertex stores 4 messages (2 user + 2 assistant), follow-up question correctly references prior turn. Read filter does not drop legitimate completions.
- [ ] Full `pnpm run test:memory` — left for CI / reviewer (~10+ min with real LLM calls; non-abort code paths are untouched, regression risk is low).

## Migration / consumer guidance

A self-contained Curator-side cleanup prompt is in `CURATOR_PROMPT.md` on this branch. It covers:
- Bumping the dep
- Removing Curator's defence-in-depth sanitizer at `conversation.ts:1433-1510` (~80 LoC)
- Removing the post-abort cleanup at `conversation.ts:2058-2070`
- Optional migration from `error.name === "AbortError"` to `error.category === ErrorCategory.ABORT`
- A 3-stage smoke test plan against already-polluted Redis sessions
- Coordinated rollout sequencing

Other internal Juspay consumers (Yama, Sensors, NeuroPulse, Kratos) automatically benefit from the read-time filter — no code changes required for them. Optional typed-error migration is the same as for Curator and can land independently.

## Backwards compatibility

- **No public API breaks.** No exports removed. No constructor signatures changed. No emitted event names changed.
- `error.name === "AbortError"` checks still work (DOMException continues to propagate from inner short-circuits; the public-API wrap converts to `NeuroLinkError` whose `isAbortError()` returns true and which has its own properties).
- Polluted Redis sessions self-heal on the next read; existing user-facing behaviour for non-aborted generations is unchanged.
- The `emit("error")` listenerCount gate **could** affect any consumer relying on the previous "double-throw" behaviour where `emit("error")` would cascade. None known. If observed, the gate can be made opt-out via env var, but that's a follow-up.

## References

- Curator reports: SI-069 (Habeeb), SI-071 (Sahil Tyagi)
- Production Langfuse traces:
  - SI-071: trace `926558b5ee80ea1dc0855d784e4d9ac0`, observation `88e94b53fd3d5d15`
  - SI-069: trace `7e90c234e1555aae4438ebb7833ef84c`, observation `7f1ceb7e2e5d8d44`
- Telemetry-gaps test (existing, still passes): `test/continuous-test-suite-telemetry-gaps.ts` Issue 5a

## Summary by CodeRabbit

* **Bug Fixes**
  * Cancelled/aborted generations no longer leave fabricated or stale assistant messages in conversation history.
  * Conversation history now filters out corrupted or empty assistant turns while preserving tool-related entries.

* **New Features**
  * Abort events now surface as a distinct, structured abort outcome for clearer error reporting.

* **Tests**
  * Added a continuous regression test suite validating session-memory integrity and abort handling.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "test:mcp": "npx tsx test/continuous-test-suite-mcp-http.ts",
     "test:media": "npx tsx test/continuous-test-suite-media-gen.ts",
     "test:memory": "npx tsx test/continuous-test-suite-memory.ts",
+    "test:session-memory-bugs": "npx tsx test/continuous-test-suite-session-memory-bugs.ts",
     "test:middleware": "npx tsx test/continuous-test-suite-middleware.ts",
     "test:observability": "npx tsx test/continuous-test-suite-observability.ts",
     "test:ppt": "npx tsx test/continuous-test-suite-ppt.ts",
--- a/src/lib/constants/enums.ts
+++ b/src/lib/constants/enums.ts
@@ -977,6 +977,13 @@ export enum ErrorCategory {
   CONFIGURATION = "configuration",
   EXECUTION = "execution",
   SYSTEM = "system",
+  /**
+   * Caller-initiated cancellation via AbortSignal. Distinct from system errors
+   * — represents a user/control-plane decision, not a SDK or provider failure.
+   * Consumers can branch on this category to differentiate "user cancelled"
+   * from "server error" without resorting to message-string matching.
+   */
+  ABORT = "abort",
 }
 
 // Error severity levels
--- a/src/lib/neurolink.ts
+++ b/src/lib/neurolink.ts
@@ -3115,16 +3115,24 @@ Current user's request: ${currentInput}`;
         if (traceCtx) {
           span.parentSpanId = traceCtx.parentSpanId;
         }
-        // Mark failed generations with ERROR status so metrics count them correctly
-        const spanStatus =
-          data.success === false || data.error
-            ? SpanStatus.ERROR
-            : SpanStatus.OK;
-        span = SpanSerializer.endSpan(
-          span,
-          spanStatus,
-          data.error ? String(data.error) : undefined,
-        );
+        // Mark failed generations with ERROR status so metrics count them
+        // correctly. Client aborts (data.aborted === true) are NOT failures —
+        // they are user-initiated cancellations and must not pollute the
+        // failure rate. Map them to WARNING with the canonical
+        // "Generation aborted by client" message (matches the Langfuse
+        // ContextEnricher mapping for outer/internal generation spans).
+        let spanStatus: SpanStatus;
+        let statusMessage: string | undefined;
+        if (data.aborted === true) {
+          spanStatus = SpanStatus.WARNING;
+          statusMessage = "Generation aborted by client";
+        } else if (data.success === false || data.error) {
+          spanStatus = SpanStatus.ERROR;
+          statusMessage = data.error ? String(data.error) : undefined;
+        } else {
+          spanStatus = SpanStatus.OK;
+        }
+        span = SpanSerializer.endSpan(span, spanStatus, statusMessage);
         span.durationMs = responseTime;
 
         // G2 fix: Check finishReason and escalate to WARNING for partial failures
@@ -3551,10 +3559,23 @@ Current user's request: ${currentInput}`;
       generateSpan.setStatus({ code: SpanStatusCode.OK });
       return result;
     } catch (error) {
-      generateSpan.setStatus({
-        code: SpanStatusCode.ERROR,
-        message: error instanceof Error ? error.message : String(error),
-      });
+      // Match the inner-span discrimination: client aborts are user-initiated
+      // cancellations, not faults. Mark with finishReason=aborted and skip
+      // ERROR status so ContextEnricher routes the outer trace to
+      // langfuse.level=WARNING (matches Curator telemetry-gaps Issue 5a). All
+      // other errors keep the existing ERROR status + recordException pair.
+      if (isAbortError(error)) {
+        generateSpan.setAttribute("ai.finishReason", "aborted");
+        generateSpan.setAttribute("neurolink.aborted", true);
+      } else {
+        generateSpan.recordException(
+          error instanceof Error ? error : new Error(String(error)),
+        );
+        generateSpan.setStatus({
+          code: SpanStatusCode.ERROR,
+          message: error instanceof Error ? error.message : String(error),
+        });
+      }
 
       // G7 fix: Distinguish context overflow errors with dedicated attributes
       if (error instanceof ContextBudgetExceededError) {
@@ -4016,6 +4037,11 @@ Current user's request: ${currentInput}`;
         ? optionsOrPrompt.model || "unknown"
         : "unknown";
 
+    // Distinguish client aborts from real failures so consumers (and Langfuse)
+    // can route them differently. `aborted: true` is additive — `success`
+    // remains false for backwards-compat with existing listeners that only
+    // branch on the boolean.
+    const aborted = isAbortError(error);
     try {
       this.emitter.emit("generation:end", {
         provider: errProvider,
@@ -4023,6 +4049,7 @@ Current user's request: ${currentInput}`;
         responseTime: 0,
         error: error instanceof Error ? error.message : String(error),
         success: false,
+        aborted,
       });
     } catch (emitError: unknown) {
       void emitError;
@@ -4499,10 +4526,24 @@ Current user's request: ${currentInput}`;
         context,
       );
     } catch (error) {
-      internalSpan.setStatus({
-        code: SpanStatusCode.ERROR,
-        message: error instanceof Error ? error.message : String(error),
-      });
+      // Client aborts are user-initiated cancellations, not system faults.
+      // Setting status=ERROR forces Langfuse to level=ERROR (see
+      // ContextEnricher.onEnd → instrumentation.ts:691). Instead leave status
+      // unset and stamp ai.finishReason=aborted so applyNonErrorLangfuseLevel
+      // maps it to level=WARNING with the canonical "Generation aborted by
+      // client" status_message. Matches Curator telemetry-gaps Issue 5a.
+      if (isAbortError(error)) {
+        internalSpan.setAttribute("ai.finishReason", "aborted");
+        internalSpan.setAttribute("neurolink.aborted", true);
+      } else {
+        internalSpan.recordException(
+          error instanceof Error ? error : new Error(String(error)),
+        );
+        internalSpan.setStatus({
+          code: SpanStatusCode.ERROR,
+          message: error instanceof Error ? error.message : String(error),
+        });
+      }
       throw error;
     } finally {
       internalSpan.end();
@@ -4595,6 +4636,15 @@ Current user's request: ${currentInput}`;
       );
       if (recoveredResult) {
         return recoveredResult;
+      }
+      // Convert raw DOMException AbortErrors (and other untyped abort shapes)
+      // into NeuroLinkError(ABORT) so callers can branch on
+      // `error.category === ErrorCategory.ABORT` instead of message matching.
+      // Skipped if the error is already a typed abort to avoid double-wrap.
+      if (isAbortError(error) && !(error instanceof NeuroLinkError)) {
+        throw ErrorFactory.aborted(
+          error instanceof Error ? error : new Error(String(error)),
+        );
       }
       throw error;
     }
@@ -4719,8 +4769,20 @@ Current user's request: ${currentInput}`;
     }
 
     if (isAbortError(error)) {
+      // Aborted generations DO NOT write to conversation memory.
+      // Fabricating an assistant turn out of an error condition (the previous
+      // "[generation was interrupted]" sentinel) pollutes the next prompt and
+      // — at the right shape — causes the model to echo the sentinel as its
+      // response. See Curator SI-069 / SI-071. Aborts are signalled to
+      // callers via the thrown error and the "error" emitter event below;
+      // there is nothing to persist, so persisting nothing is correct.
+      //
+      // Title generation continues to work: it reads the user message of the
+      // first *successful* turn (RedisConversationMemoryManager
+      // .generateConversationTitle) and never required a fabricated assistant
+      // turn — the previous comment claiming otherwise was inaccurate.
       logger.info(
-        `[${context.functionTag}] Generation aborted — storing conversation turn for title generation`,
+        `[${context.functionTag}] Generation aborted — skipping memory write (aborts must not pollute conversation history)`,
         {
           hasMemory: !!this.conversationMemory,
           memoryType: this.conversationMemory?.constructor?.name || "NONE",
@@ -4729,35 +4791,6 @@ Current user's request: ${currentInput}`;
             "unknown",
         },
       );
-
-      try {
-        const abortedResult: TextGenerationResult = {
-          content: "[generation was interrupted]",
-          provider: options.provider || "unknown",
-          model: options.model || "unknown",
-          responseTime: Date.now() - context.generateInternalStartTime,
-        };
-        await withTimeout(
-          storeConversationTurn(
-            this.conversationMemory,
-            options,
-            abortedResult,
-            new Date(context.generateInternalStartTime),
-            context.requestId,
-          ),
-          5000,
-        );
-      } catch (storeError) {
-        logger.warn(
-          `[${context.functionTag}] Failed to store conversation turn after abort`,
-          {
-            error:
-              storeError instanceof Error
-                ? storeError.message
-                : String(storeError),
-          },
-        );
-      }
     } else {
       logger.error(`[${context.functionTag}] All generation methods failed`, {
         error: error instanceof Error ? error.message : String(error),
@@ -4765,10 +4798,17 @@ Current user's request: ${currentInput}`;
     }
 
     this.emitter.emit("response:end", "");
-    this.emitter.emit(
-      "error",
-      error instanceof Error ? error : new Error(String(error)),
-    );
+    // Node EventEmitter rethrows the original error from emit("error", e) if
+    // there is no listener registered, which would short-circuit the caller's
+    // catch block and prevent the abort-typed-error wrap from running. Only
+    // emit when a consumer is listening; non-listening callers receive the
+    // error via the thrown rejection instead, which is the canonical path.
+    if (this.emitter.listenerCount("error") > 0) {
+      this.emitter.emit(
+        "error",
+        error instanceof Error ? error : new Error(String(error)),
+      );
+    }
     return null;
   }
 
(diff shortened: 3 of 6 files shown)


murdore:
Pushed `cd5370b5` (force-with-lease, squashed to single commit per repo policy).

@coderabbitai full review — would appreciate a fresh pass on the squashed diff. Below is what changed since the last review.

## Review feedback addressed

### 🔴 Critical — tool-bearing turns dropped on retrieval (CodeRabbit, `conversationMemory.ts:78`)
Root cause was confirmed: `storeConversationTurn` never persisted `events` from `result.toolsUsed` / `result.toolExecutions`, so the read filter had no signal that a turn was tool-bearing. Fixed two ways for defence in depth:

- New `hasToolActivity(msg)` helper checks `msg.tool` / `msg.args` / `msg.result` in addition to `msg.events`. Used symmetrically by both `isPollutedAssistantTurn` (read filter) and the upper-layer storage guard.
- `storeConversationTurn` now synthesises a minimal `tool:start` event marker on the assistant turn when `result.toolsUsed`/`toolExecutions` indicates activity, so future retrieval can identify the turn even when the storage path didn't carry stream events.

### 🟠 Major — outer telemetry still ERROR on aborts (CodeRabbit, `neurolink.ts:4519`)
The previous diff only fixed `internalSpan`. Now `executeGenerateRequest.catch` (the outer `generateSpan`) also discriminates aborts via `isAbortError(error)`: stamps `ai.finishReason="aborted"` + `neurolink.aborted=true` and skips ERROR status, so Langfuse routes to `level=WARNING` end-to-end. Plus `emitGenerateErrorEvent` carries an `aborted: true` flag on the `generation:end` event so downstream consumers can distinguish without parsing message strings.

### 🟠 Major — partial `emit("error")` listenerCount gate (CodeRabbit, `neurolink.ts:4784`)
The other three `emitter.emit("error", ...)` sites at lines 9436, 9667, 9674 are now also gated on `listenerCount("error") > 0`. All four call sites in `neurolink.ts` are now consistent — Node's EventEmitter rethrow path no longer fires when consumers haven't subscribed.

### Copilot — symmetry between read filter and storage guard
- `isPollutedAssistantTurn` sentinel branch now also calls `hasToolActivity(msg)`. A hypothetical sentinel-with-tools turn is preserved on read, mirroring the existing `!hasToolActivity` storage guard.
- Sentinel rejection in `storeConversationTurn` and the read filter now use the same predicate, so they can no longer disagree.

### Copilot — backwards-compat claim on `error.name`
`ErrorFactory.aborted()` now explicitly sets `err.name = "AbortError"` (overriding `NeuroLinkError`'s default name). Existing callers branching on `err.name === "AbortError"` keep working unchanged; the new `category`/`code`/`retriable` fields are purely additive. Test 4 in the suite asserts both shapes hold simultaneously.

### Copilot — test file documentation
- Header comment ("Tests in this file fail today") replaced with current-state language: "Regression suite for the two fixed bugs ... All five tests pass against the current SDK".
- `ABORT_SENTINEL` constant comment now correctly references `ABORT_LEGACY_SENTINEL` in `conversationMemory.ts` (the new canonical location).
- Test 1 throw-shape comment rewritten to match actual implementation (`new DOMException("The operation was aborted", "AbortError")`).
- Test 4 renamed and re-asserted to check the four-property contract (`name="AbortError"` for compat + `category=abort` + `code=OPERATION_ABORTED` + `retriable=false`).

### CodeRabbit — Test 1 tautological discrimination
The OR-chain at the abort detection used to include `signal.aborted` which was tautologically true (signal was pre-aborted). Tightened to discriminate strictly on error shape (`name`/`message` patterns); `signal.aborted` was removed.

### `CURATOR_PROMPT.md`
Removed from the repo per single-commit policy and the per-PR ask. Kept locally for the Curator team to copy-paste; the CodeRabbit comments about placeholder version (`9.5x.y`) and `redis-cli get` on structured types no longer apply.

## Verification

- `pnpm run build` — green
- `pnpm run test:session-memory-bugs --provider=vertex` — **5/5 PASS** (16s)
- `pnpm run test:tracing` (telemetry-gaps suite, Issue 5a included) — **15/15 PASS**

Single commit on top of `7ab19aca` (release 9.57.0). Diff: +1309 / −46 across 6 files.

@coderabbitai please re-review.

murdore:
@coderabbitai are you reviewing or nor?

murdore:
Pushed `011cd787` (amend + force-with-lease, single commit preserved). Cycle 2 of `/pr-review-cycle`.

## Cycle 2 — Review feedback addressed

After cycle 1 (`cd5370b5`), 5 unresolved Copilot comments remained. Three were stale (already-fixed concerns Copilot didn't auto-detect); two were genuinely new and now fixed.

### ✅ Genuinely new — fixed

**`conversationMemory.ts:68` — helper name shadows local boolean** (Copilot)
The module-scope helper `hasToolActivity` shared a name with the local boolean inside `storeConversationTurn`. Renamed the helper to `messageHasToolActivity` to clarify intent (one inspects a stored `ChatMessage`, the other a live `TextGenerationResult`) and eliminate the shadowing. Both call sites in `isPollutedAssistantTurn` updated.

**`conversationMemory.ts:366` — `toolActivityEvents` synthesised for every tool turn** (Copilot)
The marker was being attached even when the assistant turn already had substantive text content, which only matters when the read filter would otherwise drop the turn. Narrowed the condition from `if (hasToolActivity)` to `if (hasToolActivity && !aiResponse.trim())` — the marker is now attached only to empty-text tool-bearing turns where the read filter would actually need it. Non-empty tool turns get their natural shape preserved (no token-estimation drift).

### ℹ️ Already addressed in `cd5370b5` (stale Copilot threads)

These three Copilot comments restate concerns already fixed in the squashed commit. Pointing to current code:

- **`conversationMemory.ts:94` — sentinel branch unconditional**
  Fixed: line 88-90 returns `!messageHasToolActivity(msg)` for sentinel content. Symmetric with empty-text branch.
- **`conversationMemory.ts:350` — storage guard / read filter inconsistency**
  Fixed: line 340 storage guard already includes `&& !hasToolActivity` (the local boolean), and the read filter uses the equivalent `!messageHasToolActivity(msg)`. Same predicate at both sides.
- **`neurolink.ts:4640` — `error.name` compatibility**
  Fixed: `ErrorFactory.aborted()` in `errorHandling.ts` overrides `err.name = "AbortError"` after construction, so `err.name === "AbortError"` checks continue to work; the new `category`/`code`/`retriable` are additive. Test 4 in the suite asserts both shapes simultaneously.

### Note on the previous CodeRabbit comments

CodeRabbit's three 🟠/🔴 comments from the earlier review (outer telemetry, partial `emit("error")` guard, tool-bearing turn dropped on retrieval) auto-marked themselves **`✅ Addressed in commit cd5370b`** when the squash landed.

## Verification

- `pnpm run check` — ✅ 3625 files, 0 errors, 0 warnings
- `pnpm run build` — ✅ all good
- `pnpm run test:session-memory-bugs --provider=vertex` — **5/5 PASS** (16s)

Single commit on top of `7ab19aca` (release 9.57.0). Diff: +1316 / −46 across 6 files.

@coderabbitai please re-review on `011cd787`.
@copilot-pull-request-reviewer please re-review.

murdore:
Amended HEAD to `2961535c` (force-pushed) addressing 3 of CodeRabbit's nits:

- **conversationMemory.ts** fallback `toolActivityEvents` push now includes `toolName: "unknown"` so downstream code can always read `event.toolName`.
- **session-memory-bugs.ts** test `forEach` callback wrapped in block-bodied arrow (no implicit return — `useIterableCallbackReturn` clean).
- **session-memory-bugs.ts** `let response` annotated with `Awaited<ReturnType<typeof memorySdk.generate>> | undefined`. Existing `response?.content` usage already handles `undefined`.

**Skipped** the suggestion to omit `error` / flip `success` to true on aborts in `emitGenerateErrorEvent`. The inline comment in that block explicitly chose to keep `success: false` for backwards-compat with existing listeners that only branch on the boolean — `aborted: true` is additive. Happy to revisit if Curator confirms they're keying off `success` rather than `aborted`.

@coderabbitai review

murdore:
Pushed `a8144db7` (amend + force-with-lease, single commit). Cycle 4.

## Cycle 4 — Review feedback addressed

CodeRabbit's full review surfaced two findings against `011cd787`:

### 🟠 Major — `generation:end` for aborts still becomes `SpanStatus.ERROR`

**File:** `src/lib/neurolink.ts`, `initializeMetricsListeners()` listener (around line 3118).

The previous fix added `aborted: true` to the `generation:end` payload, but the listener that consumes it still mapped `success === false || error` straight to `SpanStatus.ERROR` — which meant Pipeline B's metrics span recorded aborted requests as failures, even though the outer `generateSpan` and inner `internalSpan` correctly mapped them to WARNING.

**Fix (option C from CodeRabbit's suggestion — listener-side mapping):** the listener now checks `data.aborted === true` first and routes to `SpanStatus.WARNING` with `statusMessage="Generation aborted by client"`. `data.success === false || data.error` continues to map to `ERROR`. The `generation:end` payload itself is unchanged (`success: false, aborted: true`) so existing consumers branching on `success` keep working.

```typescript
let spanStatus: SpanStatus;
let statusMessage: string | undefined;
if (data.aborted === true) {
  spanStatus = SpanStatus.WARNING;
  statusMessage = "Generation aborted by client";
} else if (data.success === false || data.error) {
  spanStatus = SpanStatus.ERROR;
  statusMessage = data.error ? String(data.error) : undefined;
} else {
  spanStatus = SpanStatus.OK;
}
```

This now matches the canonical Langfuse mapping that's also applied at:
- `internalSpan` / `generateSpan` via `applyNonErrorLangfuseLevel` in `instrumentation.ts:245-247`
- The existing telemetry-gaps test `Issue 5a` ("aborted generation → level=WARNING")

So all three layers (Pipeline B metrics span, internal generation span, outer trace span) agree on aborts → WARNING.

### 🟡 Nit — synthetic `tool:start` events missed canonical `tool` field

**File:** `src/lib/utils/conversationMemory.ts`, `toolActivityEvents` synthesis (around line 391).

The canonical `ToolExecutionEvent` shape (`src/lib/types/tools.ts:348`) is `{tool: string, toolName?: string}` — `tool` is required, `toolName` is the documented compat alias. My synthesised events were pushing only `toolName`, so consumers that read `event.tool` (the canonical field) would find it undefined.

**Fix:** both the `usedNames` loop and the `"unknown"` fallback now push `tool: name` alongside `toolName: name`. Shape now matches `ToolExecutionEvent` exactly.

```typescript
toolActivityEvents.push({
  type: "tool:start",
  seq: seq++,
  timestamp: now,
  tool: name,        // canonical
  toolName: name,    // compat alias
});
```

`messageHasToolActivity` keeps working — it only checks `event.type` so the rename doesn't affect it. Future consumers that read `event.tool` now work correctly.

## Verification

- `pnpm run check` — ✅ 3625 files, 0 errors, 0 warnings
- `pnpm run build` — ✅
- `pnpm run test:session-memory-bugs --provider=vertex` — **5/5 PASS** (18s)
- `pnpm run test:tracing` (telemetry-gaps) — **15/15 PASS** (Issue 5a still green)

Single commit on top of `7ab19aca` (release 9.57.0). Diff: +1342 / −56 across 6 files.

@coderabbitai please re-review on `a8144db7`.

murdore:
Pushed `631eb106` (amend + force-with-lease, single commit). Cycle 5.

## Cycle 5 — Review feedback addressed

Two new Copilot findings against `a8144db7`. CodeRabbit's parallel "Aborts still recorded as failed generations" comment auto-marked itself **`✅ Addressed in commit a8144db`** (the listener-side mapping cycle 4 added is what they were asking for).

### 🟡 Doc — `generation:end` JSDoc missing the new `aborted` field
**File:** `src/lib/neurolink.ts:8036` — `getEventEmitter()` JSDoc.

Public docs documented `generation:end` payload as `{ provider, responseTime, toolsUsed?, timestamp }` only. Now updated to include the full set: `{..., success?, aborted?, error?}` plus a clarifying paragraph on semantics:
- `success: false` covers both failures and aborts (back-compat)
- `aborted: true` distinguishes cancellations
- Pipeline B's metrics span maps `aborted: true` to `SpanStatus.WARNING` (not ERROR)

So consumers reading the JSDoc now know exactly what to branch on without guessing.

### 🟡 Logging — polluted-turns warn fires on every read
**File:** `src/lib/utils/conversationMemory.ts:214` (read-time filter logging).

Previously `logger.warn` fired on every retrieval for any session with polluted turns — a 50-turn polluted conversation would produce 50 warn lines. Per Copilot's option (b), now deduplicated **per-session**:

- Module-level bounded `Set<string>` (`POLLUTED_WARN_DEDUP_MAX = 1024`) tracks sessionIds that have already been warned about.
- First read for a polluted session emits the WARN (visible for triage) and adds the sessionId to the set.
- Subsequent reads for the same session log at DEBUG level instead.
- Span attribute `neurolink.memory.polluted_turns_dropped` is **always** set, regardless of warn deduplication — Langfuse traces still show the cleanup happening on every read, so the persistent debugging signal is preserved.
- When the set reaches capacity it's cleared (cheap, bounded memory). Worst case: a few more warns get emitted as the set re-warms — acceptable trade for not having an unbounded structure.

I deliberately did NOT pick option (a) (write-back cleanup). The architectural decision documented in the original PR description is that the read-time filter must NOT mutate storage — so polluted Redis sessions self-heal at read time without any writes, and `getConversationMessages` is side-effect-free for the underlying store. Adding a write-back here would re-introduce the same class of "operation X side-effects memory" bug we're fixing.

## Verification

- `pnpm run check` — ✅ 3625 files, 0 errors, 0 warnings
- `pnpm run build` — ✅
- `pnpm run test:session-memory-bugs --provider=vertex` — **5/5 PASS** (18s)

Single commit on top of `7ab19aca` (release 9.57.0). Diff: +1382 / −58 across 6 files.

@coderabbitai please re-review on `631eb106`.
@copilot-pull-request-reviewer please re-review.

PR #996, opened 2026-04-28
fix(routing): dual-mode image text fallback + video-frame hijack gate

murdore:
## Summary

- **Bug 1**: Dual-mode Gemini image models (`gemini-3.1-flash-image-preview`, `gemini-2.5-flash-image`, `gemini-3-pro-image-preview`) decide per-request whether to emit an image or text. When the model emitted text, `parseImageGenerationResponse` threw `"model returned text instead of image data"` and burned 3 retries — even though the response already carried the answer. Now returns the text via a `{ textFallback }` discriminated-union branch and `executeImageGeneration` packages it as a normal text result.
- **Bug 2**: `handleVideoFrameGeneration` fired whenever a request had ≥3 image content parts, ignored `options.schema` / `options.output.format`, and called `executeVideoAnalysis` with hardcoded `gemini-2.5-flash` returning prose. Schema callers silently lost their structured-output contract. Now early-returns when `options.schema` or `options.output?.format` is set, so the request falls through to the standard generate flow which honors both.

Image happy path unchanged. Genuinely-empty Vertex responses still throw. No public-API surface change beyond the now-richer `parseImageGenerationResponse` return type (private method).

## Files

- `src/lib/core/baseProvider.ts` — Bug 2 fix: `handleVideoFrameGeneration` schema/format gate
- `src/lib/providers/googleVertex.ts` — Bug 1 fix: `parseImageGenerationResponse` discriminated union + `executeImageGeneration` text-fallback handling
- `test/continuous-test-suite.ts` — 4 new end-to-end tests + runner registration

## Repro evidence (pre-fix)

```
[NEUROLINK:ERROR] Image generation failed {"error":"[vertex] Image generation completed but model returned text instead of image data. Model: gemini-3.1-flash-image-preview","prompt":"What is the capital of France? Reply with one word only."}
[BUG-1-TRACE] parseImageGenerationResponse: no image part {"candidatePartCount":3,"partTypes":["text","text","text"],"hasTextContent":true,"textPreview":"...the answer: Paris","textTotalLength":520}

Schema 3+ Images: FAIL - BUG REPRODUCED: usage.total=0 indicates handleVideoFrameGeneration hijack fired (schema bypassed)
JSON 3+ Images:   FAIL - BUG REPRODUCED: usage.total=0 indicates handleVideoFrameGeneration hijack fired (output.format=json bypassed)
```

## Post-fix (all 4 PASS)

```
Bug 1 CLI: Returns "Paris" (model's actual response surfaced as text)
SDK Text on Image Model: PASS - text returned, no imageOutput
Schema 3+ Images:        PASS - schema honored, usage.total=1209, images.length=3
JSON 3+ Images:          PASS - json output honored, usage.total=1079, count=3
```

## Test plan

- [x] `pnpm run check` — 0 errors, 0 warnings (3628 files)
- [x] `pnpm run lint` — 0 errors (18 pre-existing warnings in unchanged files)
- [x] Bug 1 CLI repro against Vertex `gemini-3.1-flash-image-preview` returns text instead of throwing
- [x] Bug 1 SDK repro asserts `result.content` has the answer + `result.imageOutput` is absent
- [x] Bug 2 schema repro against Vertex `gemini-2.5-pro` with 3 images + Zod schema asserts `usage.total > 0` (deterministic hijack signal) and schema validates
- [x] Bug 2 json repro with `output: { format: "json" }` asserts `usage.total > 0` and JSON parses
- [ ] Reviewer to confirm: image happy-path unchanged (existing `testCLIGenerateImage`, `testSDKGenerateImage`, `testGoogleAIStudioImageGeneration` still pass)
- [ ] Reviewer to confirm: bare 3-image queries (no schema, no format) still route through video-analysis (existing video-frame behavior preserved)

## Bug-2 test rationale: why `usage.total === 0` is the assertion, not response shape

Asserting on response content alone produces false positives — the hijacked path uses `gemini-2.5-flash` which can coincidentally emit JSON when the prompt requests it. The hijack returns a hardcoded `{ input: 0, output: 0, total: 0 }` usage when no `systemPrompt` is set; the standard generate flow always reports real token counts from the AI SDK. Asserting `usage.total > 0` is therefore the deterministic signal that the request reached the real model API rather than `executeVideoAnalysis`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed structured output (schema/JSON format) requests with multimodal image models
  * Enhanced handling when image generation returns text responses instead of image data

### DIFF
--- a/src/lib/core/baseProvider.ts
+++ b/src/lib/core/baseProvider.ts
@@ -844,6 +844,22 @@ export abstract class BaseProvider implements AIProvider {
     if (!hasVideoFrames(messages)) {
       return null;
     }
+    // Bug 2 fix: callers requesting structured output (schema or explicit
+    // output.format) must NOT be hijacked into the prose-returning video
+    // analysis path. Without this gate, schema/format are silently dropped
+    // whenever messages contain >=3 image parts.
+    if (options.schema !== undefined || options.output?.format !== undefined) {
+      logger.info(
+        "[VideoFrameGen] Skipping video-frame analysis route; caller requested structured output",
+        {
+          provider: this.providerName,
+          model: this.modelName,
+          hasSchema: options.schema !== undefined,
+          outputFormat: options.output?.format,
+        },
+      );
+      return null;
+    }
 
     const videoAnalysisResult = await executeVideoAnalysis(messages, {
       provider: options.provider,
--- a/src/lib/providers/googleVertex.ts
+++ b/src/lib/providers/googleVertex.ts
@@ -4252,7 +4252,14 @@ export class GoogleVertexProvider extends BaseProvider {
   }
 
   /**
-   * Parse the Vertex AI image generation REST API response and extract image data.
+   * Parse the Vertex AI image generation REST API response.
+   *
+   * Dual-mode image models (gemini-3.1-flash-image-preview, gemini-2.5-flash-image,
+   * gemini-3-pro-image-preview) decide per-request whether to emit an image or text.
+   * When the response contains text parts but no image part, surface the text via
+   * `textFallback` so the caller can return a normal text result instead of throwing
+   * "model returned text instead of image data" and burning retries on a query that
+   * the model has already answered.
    */
   private parseImageGenerationResponse(
     data: {
@@ -4267,7 +4274,7 @@ export class GoogleVertexProvider extends BaseProvider {
       }>;
     },
     imageModelName: string,
-  ): { imageData: string; mimeType: string } {
+  ): { imageData: string; mimeType: string } | { textFallback: string } {
     const candidate = data.candidates?.[0];
     if (!candidate?.content?.parts) {
       throw new ProviderError(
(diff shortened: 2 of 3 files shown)


murdore:
## Review Feedback Addressed (Cycle 1)

### Changes Made

- **`src/lib/providers/googleVertex.ts:4297-4308`** — *@coderabbitai: whitespace/empty text parts shouldn't trigger `textFallback`*
  - **Fix:** Changed the text-parts filter from `typeof part.text === "string"` (which lets empty strings through) to `(part) => (typeof part.text === "string" ? part.text : "")` followed by `.filter((text) => text.trim().length > 0)`. An effectively empty response now throws `"no image data was returned"` instead of returning `content: ""`.
  - **Verification:** Bug 1 SDK + CLI repros still PASS (text returned for non-empty responses).

- **`test/continuous-test-suite.ts:4268-4283`** — *@coderabbitai: schema/JSON tests can pass on weak positives*
  - **Fix (schema test):** Added `valid.data.images.length === 3 && indices.join(",") === "0,1,2"` assertion so the test fails if the model returned a schema-shaped response with the wrong number of items or out-of-range indices. Failure log includes actual `images.length`, `indices`, and `usage.total` for debugging.
  - **Fix (JSON test):** Replaced `'count' in parsed` with `Number(parsed.count) === 3` so the test fails if the model returned `{count: 7}` or any other value.
  - **Verification:** Both repros PASS post-tightening (`images.length=3`, `count=3`).

### Files Modified
- `src/lib/providers/googleVertex.ts` — empty-text-parts filter
- `test/continuous-test-suite.ts` — schema and JSON assertion tightening

### Validation
- TypeScript: ✅ 0 errors, 0 warnings (3628 files)
- Lint: ✅ 0 errors (18 pre-existing warnings in unchanged files)
- All 4 end-to-end repros against Vertex: ✅ PASS
  - `Bug 1 CLI` — returns "Paris" via text fallback
  - `Bug 1 SDK` — `text returned, no imageOutput`
  - `Bug 2 schema` — `schema honored, usage.total=1213, images.length=3`
  - `Bug 2 json` — `json output honored, usage.total=1091, count=3`

### Requesting Re-review
@coderabbitai — both findings addressed. Please re-review.

PR #1006, opened 2026-05-01
test(harness): consolidate continuous-test suites into per-domain files + harden SDK for live test sweep

murdore:
### Test harness consolidation
- Replace ~5,500 lines of copy-pasted boilerplate across 38 hand-rolled `continuous-test-suite-*.ts` files with a shared harness in `test/helpers/` (tri-state PASS/FAIL/SKIP via `defineSuite()`).
- Split the 6,420-line `continuous-test-suite-mcp.ts` monolith into seven focused `mcp-*` files (infra/bash/limits/spans/sdk/cli/http) with shared helpers in `test/helpers/mcpHelpers.ts`.
- Tighten `test/helpers/envGuard.ts` so loose substrings (`"failed to"`, `"not found"`, `"timeout"`, bare `"403"`/`"404"`/`"429"`) no longer mask real consumer-facing regressions.
- Refactor `envGuard.ts` to a list of named patterns (`EXPECTED_PROVIDER_ERROR_PATTERNS`) and add a pattern-coverage self-check: every pattern must be matched by at least one MUST_SKIP fixture (109 tests across 63 real provider errors, 19 bug-mimicking strings, and 27 pattern-coverage checks).
- Add CLI `--thinking-level` smoke test (Part 4c) — was previously SDK-only.
- Net: 38 → 32 test files. Largest test file shrinks from 6,420 → 3,309 lines (`mcp-sdk.ts`).

### SDK hardening (necessary for live test sweep)
- `BaseProvider` now wires consumer-supplied `onChunk` / `onFinish` / `onError` lifecycle callbacks on every streaming transport (Ollama, NIM, llama.cpp, custom OpenAI-compatible) — previously only the AI-SDK-internal path saw them. See `wrapStreamWithLifecycleCallbacks` + `fireLifecycleErrorCallback` in `src/lib/core/baseProvider.ts`.
- Lifecycle middleware (`src/lib/middleware/builtin/lifecycle.ts`) now normalizes non-`Error` throws via `Object.assign(new Error(msg), obj)` so structured fields (`code`, `status`, `retryAfter`) survive — they were previously dropped to `Error("[object Object]")`.
- New shared `WeakSet`-backed lifecycle-fired registry (`src/lib/utils/lifecycleErrorMarker.ts`) prevents duplicate `onError` emits across the SDK → middleware → BaseProvider → NeuroLink top-level catch chain, without mutating caller-owned error objects.
- Per-callback deadline for lifecycle handlers is now configurable via SDK (`LifecycleMiddlewareConfig.timeoutMs`) and CLI (`NEUROLINK_LIFECYCLE_TIMEOUT_MS` env var). Default remains 5_000ms. See `src/lib/utils/lifecycleTimeout.ts`.
- Defensive call-isolation clone for `generate()` / `stream()` extracted to `src/lib/utils/cloneOptions.ts` with the mutable-branch list as a single source of truth (`CLONE_MUTABLE_OPTION_BRANCHES`).
- `NeuroLink.generate()` / `stream()` top-level catch now fires consumer `onError` for failures that happen before the language-model middleware runs (unknown provider, validation, factory exceptions).
- `reasoning` / `reasoningTokens` now forwarded end-to-end through `BaseProvider` → `NeuroLink` result builders (DeepSeek `reasoning_content`, Anthropic thinking, Gemini thought parts, OpenAI o1).

### Provider work absorbed (was already required for the live sweep)
- **OpenRouter:** dropped retired `anthropic/claude-3-5-sonnet` everywhere (model not available on any OpenRouter endpoint anymore). The `OpenRouterModels.CLAUDE_3_5_SONNET` enum entry is removed; callers should switch to `CLAUDE_3_7_SONNET` (or one of the 4.x entries).
- **DeepSeek + NVIDIA NIM:** migrated from `@ai-sdk/openai` to `@ai-sdk/openai-compatible` — fixes silent `reasoning_content` drop and unconditional `response_format: json_schema` rejection. NIM 400 retry-strip predicate is narrowed to require a rejection keyword (`unsupported` / `not supported` / `unknown field` / etc.) within 80 characters of the offending field name, so a legitimate 400 whose body merely mentions the field doesn't trigger a silent strip.
- **Azure OpenAI:** URL-parsing rewrite adds Azure AI Foundry (`*.services.ai.azure.com`) support and avoids the doubled-suffix bug in the previous string-replace path.
- **Google AI Studio / Vertex:** sanitize `function_declarations[].name` per Google's `[A-Za-z_][A-Za-z0-9_.:-]{0,127}` regex; reverse map (`originalNameMap`) keeps the sanitization transport-only so consumer-facing tool-call metadata uses the original identifier.
- **Per-call credentials:** `lm-studio` / `nvidia-nim` kebab-case names now map to the correct camelCase credential keys in `ProviderFactory`.

## Verification

| Tier | Suites | Result |
|------|--------|--------|
| **no-API** | `envguard` 109/109 (incl. 27/27 pattern-coverage), `bugfixes` 48/48, `dynamic` 11/11, `mcp:infra` 84/84, `mcp:bash` 13/13 | ✅ green locally |
| **live** | `providers` 27+6S, `mcp:http` 12+4S, `mcp:sdk` 88/88, `mcp:cli` 17+1S, `observability` 45+2S, `context` 53+6S, `memory` 20/20, `tool-reliability` 6/6, `evaluation` 18+3S, `autoresearch` 17/17, `credentials` 13+2S | ✅ green with full keys (2026-05-04) |
| **product** | `media` 18/18, `tts` 15/15, `ppt` 13+3S, `proxy` 16/16 | ✅ green with full keys (2026-05-04) |

`task:tasks` 40+21S+1F is **1 fewer FAIL than baseline** (the remaining failure is a pre-existing timing flake on `Invalid provider fails fast without backoff`, not introduced by this PR).

## Pre-existing issues NOT introduced by this PR

- `tasks` — `Invalid provider fails fast without backoff` timing flake.
- `new-providers` — 8 assertion mismatches against DeepSeek / NVIDIA NIM for capabilities those upstream services don't fully implement (thinking, structured output, multiturn memory, invalid-key handling).

## Test plan

- [x] `pnpm run check` clean (3657 files, 0 errors, 0 warnings)
- [x] `pnpm run lint` clean (0 errors; 17 pre-existing warnings unrelated to this PR)
- [x] `pnpm run test:envguard` 109/109 with 27/27 pattern coverage
- [x] `pnpm run test:bugfixes` 48/48
- [x] `pnpm run test:dynamic` 11/11
- [x] `pnpm run test:mcp:infra` 84/84
- [x] `pnpm run test:mcp:bash` 13/13
- [x] `pnpm run build:cli` clean
- [x] Live tier green on a separate run with full provider keys
- [x] No regressions in `mcp-http`, `providers`, `observability`, `context`, `memory` suites

## Configuration surface

- `NEUROLINK_LIFECYCLE_TIMEOUT_MS` — env var deadline for `onChunk`/`onFinish`/`onError` lifecycle callbacks. Defaults to `5000`. Set `0` for fire-and-forget. CLI honors the same var; the SDK also accepts `lifecycle.timeoutMs` directly on `LifecycleMiddlewareConfig`.

## Summary by CodeRabbit

* **New Features**
  * Generation results now include provider reasoning metadata.
  * CLI annotate gains a standalone inference mode.
  * Tool-calling preserves original consumer tool identifiers for Google/Gemini paths.
  * Lifecycle callback deadline now configurable via SDK and env var.

* **Improvements**
  * OpenRouter default updated to Claude 3.7 Sonnet; retired 3.5 entry removed.
  * Improved provider compatibility, request normalization and retry handling.
  * More reliable streaming lifecycle callbacks and earlier onError delivery; double-fire suppression now backed by a shared WeakSet.
  * Better Azure endpoint handling and provider credential mapping.
  * NIM 400 retry-strip narrowed to require a rejection keyword near the offending field.

* **Tests**
  * Continuous test suites consolidated on a shared harness; MCP suites split and expanded.
  * envGuard refactored to a list of named patterns with a coverage self-check that fails CI if a pattern has no fixture.

* **Documentation**
  * Added comprehensive continuous-test README.

* **Chores**
  * .gitignore updated to exclude test runtime artifacts and outputs.


murdore:
# Review Plan — PR #1006 (forced-update, single squashed commit `98d60200`)

This PR is one commit on top of `release@902b2645`. **+19,136 / -31,207 LOC** across **125 files**. Below is the exhaustive checklist for reviewers (humans or `/ultrareview`). Each section names the scope, the specific things to check, and the rationale.

---

## 0. Pre-flight: project rules (CLAUDE.md) compliance

These rules are AST-enforced by `eslint-rules/`. `pnpm run lint` is clean (0 errors), but please spot-check:

- [ ] **No `interface` keyword** in any TS file (rule 7) — check new files in `test/helpers/*.ts`, modified providers (`deepseek.ts`, `nvidiaNim.ts`, `azureOpenai.ts`, `googleNativeGemini3.ts`, `openRouter.ts`), `core/baseProvider.ts`, `factories/providerFactory.ts`, `middleware/builtin/lifecycle.ts`, `cli/commands/mcp.ts`.
- [ ] **No types in feature subdirectories** (rule 2 + 11) — all type changes must live in `src/lib/types/`. Verify `src/lib/types/generate.ts` is the only place where `reasoning` / `reasoningTokens` were added to `TextGenerationResult`.
- [ ] **Type imports use the barrel** (rule 13) — code outside `src/lib/types/` must import from `../types/index.js` (or `../types`), never from a specific type file.
- [ ] **Type re-exports forbidden in non-type files** (rule 12).
- [ ] **`formatProviderError` returns, never throws** (rule 6) — verify in `deepseek.ts`, `nvidiaNim.ts`, `azureOpenai.ts`.
- [ ] **Backwards compatibility** (rule 5) — existing public SDK callers must not break. Specifically:
  - `OpenRouterModels.CLAUDE_3_5_SONNET` enum is preserved but pointed at the working alias — verify nobody breaks if they read the enum value as a literal string.
  - `TextGenerationResult.reasoning` / `reasoningTokens` are additive optional fields.
  - `NeurolinkCredentials.lmStudio` / `nvidiaNim` keys (camelCase) — confirm both kebab-case (enum value) and camelCase (developer-facing) paths still work.
- [ ] **Dynamic imports only in providerRegistry** (rule 1) — verify no static `import` was added inside `providerRegistry.ts` factory functions.

---

## 1. Test harness consolidation (largest blast radius)

**Scope:** 30+ ad-hoc continuous-test files collapsed into 32 per-domain suites; `mcp.ts` split into 7 `mcp-*` files; new shared helpers in `test/helpers/`.

### What to verify
- [ ] **Suite coverage parity:** every test that existed before still exists somewhere. Cross-check `test/MIGRATION_REPORT.md` against the deleted files (`continuous-test-suite-{voice,voice-server,session-memory-bugs,telemetry-gaps,tracing,evaluation-scoring,issue-01..06,autoresearch-e2e,autoresearch-live}.ts`). Voice tests were KEPT (added by upstream v9.62). 
- [ ] **`test/helpers/harness.ts` `defineSuite` / `runSuite` / `Skip`** — the core abstraction. Confirm:
  - SKIP is distinguished from PASS/FAIL via the `Skip` class instance check, not string-prefix sniffing.
  - `runSuite` exits with non-zero only on FAIL.
  - The suite name is propagated into telemetry/logging.
- [ ] **`test/helpers/envGuard.ts`** — anchored regexes only (no substring traps that could swallow real failures). 65/65 unit tests PASS via `pnpm run test:envguard`. The 17 newly added phrases:
  - rate-limit / 429 promoted from auth-only to transport-always-skip
  - HTTP reason phrases (Gone / Internal Server Error / Service Unavailable / Bad Gateway)
  - Vendor-specific gateway errors (NIM Bad Request, OpenAI quota-exceeded, OpenAI streaming-with-tools wrapper, Azure deployment missing, Google function_declarations name invalid, DeepSeek/OpenAI tool-schema rejection, Anthropic Not Found, lm-studio/llamacpp unreachable)
  - Verify each pattern is anchored enough not to skip a genuine bug.
- [ ] **`test/helpers/skipIf.ts`** — same: anchored, doesn't swallow real errors.
- [ ] **package.json scripts:** all `test:*` scripts point at existing files. New scripts: `test:tasks`, `test:auth`, `test:autoresearch`, `test:autoresearch:redis`, `test:envguard`, `test:unit`, `test:live`, `test:product`. CI tier comments (`// CI tier — fast`, etc.) document the intent.
- [ ] **`pnpm run test:ci`** still includes `test:hitl` (was preserved during conflict resolution).
- [ ] **`pnpm run test`** still runs the main `continuous-test-suite.ts`.
- [ ] **No suite imports from `dist/` of a missing path** — every `from "../dist/..."` import is buildable.

### Run locally
```
pnpm run test:envguard       # 65/65 PASS
pnpm run test:credentials    # all SKIP without API keys, 0 FAIL
pnpm run test:unit           # CI tier — no live AI calls
```

---

## 2. SDK provider fixes

### 2.1 DeepSeek + NIM → @ai-sdk/openai-compatible
**Files:** `src/lib/providers/deepseek.ts`, `src/lib/providers/nvidiaNim.ts`, `package.json` (added direct dep), `src/lib/types/generate.ts`.

- [ ] `@ai-sdk/openai-compatible` is now a direct dep (was transitive). Verify `pnpm-lock.yaml` lockfile is consistent.
- [ ] DeepSeek `transformRequestBody` injects "json" into the prompt when `json_object` mode is active. Confirm:
  - Doesn't double-inject if the prompt already contains "json".
  - Doesn't inject into structured-output mode (`json_schema`).
  - Doesn't break tool-calling messages.
- [ ] DeepSeek `supportsStructuredOutputs:false` causes graceful fallback to `json_object` instead of failing.
- [ ] `reasoning_content` (DeepSeek-reasoner CoT, NIM reasoning models) is surfaced through ALL THREE result-build sites in `neurolink.ts`:
  - root path (line ~3617 `generate()`)
  - MCP path
  - direct-provider path
- [ ] `TextGenerationResult.reasoning` and `reasoningTokens` typed correctly as optional.
- [ ] `nvidiaNim.ts` `stripFieldFromJsonBody` retry strips both `chat_template_kwargs.reasoning_budget` and top-level `chat_template`, on **both** stream and generate paths.

### 2.2 Per-call credentials kebab-case fix
**File:** `src/lib/factories/providerFactory.ts:98-103`.

- [ ] `credentialKeyMap` now includes `lm-studio → lmStudio` and `nvidia-nim → nvidiaNim`.
- [ ] Existing camelCase-only providers (e.g. `openai`, `anthropic`) still work — the map is for kebab-case enum values that need translating.
- [ ] Verify `credentials.lmStudio.baseURL` actually overrides the default for both stream and generate paths. The new-providers K2 test exercises this.

### 2.3 Azure AI Foundry endpoint routing
**File:** `src/lib/providers/azureOpenai.ts`.

- [ ] `*.services.ai.azure.com` hosts go through `baseURL: "<host>/openai"`.
- [ ] Classic `*.openai.azure.com` / `*.cognitiveservices.azure.com` hosts continue to use `resourceName`.
- [ ] No regression for users with the old endpoint format.
- [ ] The bad URL pattern `<host>.services.ai.azure.com.openai.azure.com` is no longer constructed.

### 2.4 Google function_declarations name sanitization
**File:** `src/lib/providers/googleNativeGemini3.ts`.

- [ ] Sanitization regex matches Google's spec: `[A-Za-z_][A-Za-z0-9_.:-]{0,127}`.
- [ ] Applied at BOTH paths: `buildNativeToolDeclarations` (native @google/genai) and `sanitizeToolsForGemini` (AI-SDK).
- [ ] Logs a WARNING (not an error) when a name is rewritten.
- [ ] Renamed tool names are stable across calls (deterministic — same input → same output).
- [ ] Tool execution still works after rename — verify the rewritten name is used end-to-end.

### 2.5 OpenRouter default model bump
**Files:** `src/lib/providers/openRouter.ts`, `src/lib/factories/providerRegistry.ts`, `src/lib/utils/modelChoices.ts`, `src/lib/constants/enums.ts`.

- [ ] All three places aligned on `anthropic/claude-3.7-sonnet` as the new default.
- [ ] `OpenRouterModels.CLAUDE_3_5_SONNET` enum still exists (back-compat) but points at working alias.
- [ ] `OpenRouterModels.CLAUDE_3_7_SONNET` added to enum.
- [ ] CLI provider choices (`src/cli/factories/commandFactory.ts`) — does it expose model choices that need updating? Check.

---

## 3. Lifecycle callback wrapping for raw-fetch streaming

**Files:** `src/lib/core/baseProvider.ts`, `src/lib/middleware/builtin/lifecycle.ts`, `src/lib/neurolink.ts`.

- [ ] `wrapStreamWithLifecycleCallbacks` in baseProvider:
  - Reads `options.middleware?.middlewareConfig?.lifecycle?.config` defensively (no NPE if missing).
  - Fires `onChunk` for each chunk in order.
  - Fires `onFinish` exactly once with final aggregated state.
  - Fires `onError` exactly once on stream error and re-throws.
- [ ] `__lifecycleErrorFired` marker on Error:
  - Set BEFORE `onError` fires (so consumer-level catch can read it).
  - Read by `neurolink.ts` `fireConsumerOnErrorIfNotFired` to skip if SDK already fired.
  - Type-marked correctly: `Error & { __lifecycleErrorFired?: boolean }`.
- [ ] `neurolink.ts` `generate()` and `stream()` try/catch — verify they don't swallow errors that the consumer hasn't subscribed to via lifecycle callbacks.
- [ ] **Provider scope:** which providers benefit? Ollama, llama.cpp `/v1` are the named raw-fetch providers. Check that AI-SDK-wrapped providers (OpenAI, Anthropic, etc.) don't double-fire onChunk/onFinish.

---

## 4. CLI mcp annotate --infer standalone path

**File:** `src/cli/commands/mcp.ts`.

- [ ] When NO MCP servers are configured, `executeAnnotate` with `--infer` calls `inferAnnotations({ name: toolName, description: "" })` directly without trying to look up servers.
- [ ] When servers ARE configured, the original lookup path still runs.
- [ ] Tool name passed to `inferAnnotations` is validated/sanitized.

---

## 5. Documentation

- [ ] `test/SKIP_ANALYSIS.md` — accurate per-suite triage. Numbers in §0 verification table match `pnpm test:*` actual output.
- [ ] `test/RUNNING_LOCAL_SERVICES.md` — `lms server start`, `lms load <model>`, `llama-server -m <gguf> --port 8080 --host 127.0.0.1 -ngl 99` commands are correct and reproducible.
- [ ] No leaked secrets, internal hostnames, or PII in any docs.
- [ ] No `*ROADMAP-2025-2026.md` deletions (whitelisted in `.gitignore`).
- [ ] `test/MIGRATION_REPORT.md` and `test/MIGRATION_BASELINE.md` — useful to preserve, or move to `archive/`?

---

## 6. .gitignore additions

**File:** `.gitignore`.

- [ ] Test-output artifact paths (alphabet.txt, city.json, story.txt, hello.txt, etc.) are anchored to `/` so they don't ignore legitimate files in subdirectories.
- [ ] `test/logs/` — verification-sweep log directory.
- [ ] No accidental ignores of source files or fixtures.

---

## 7. Security / safety

- [ ] **No credentials in logs:** verify `transformParamsForLogging()` is called everywhere request payloads are logged. Per-call credentials must NEVER appear in logs.
- [ ] **No new shell-injection vectors** in CLI commands.
- [ ] **No secrets in test fixtures** (`test/fixtures/sample-screenshot.png` is a real screenshot — verify it's not a real customer/employee screenshot).
- [ ] **Dependency additions:** `@ai-sdk/openai-compatible` is the only new direct dep. Verify it's at the same major version as `@ai-sdk/openai`.
- [ ] **No skipped GitHub Dependabot alerts** introduced (83 vulns on default branch, but that's pre-existing — don't add new ones).

---

## 8. Performance / runtime risk

- [ ] **TransformStream wrap in baseProvider** — adds a microtask per chunk for ALL streaming paths. Verify no measurable per-chunk latency regression for AI-SDK-wrapped providers (where the wrap is effectively a no-op).
- [ ] **Test suite runtime** — the 32 split suites should NOT take longer than the old monolithic suite when run sequentially. Spot-check `pnpm run test:mcp:full` runtime.
- [ ] **Memory** — no lingering listeners or unflushed buffers in the lifecycle wrap.

---

## 9. CI / build / release

- [ ] `pnpm run check` — 0 errors / 0 warnings (verified locally).
- [ ] `pnpm run lint` — 0 errors (19 pre-existing warnings — confirm none NEW).
- [ ] `pnpm run build` — passes (verified via pre-commit prepack hook).
- [ ] `pnpm run build:browser` — produces `dist/browser/neurolink.min.js` cleanly.
- [ ] `pnpm pack` produces a valid tarball (verified via `publint`).
- [ ] **Conventional commit** format on the squashed commit — confirm `test(harness): ...` is the right type. Could also be argued as `feat(harness)` since it's a structural change. Reviewer's call.
- [ ] **Semantic-release impact** — will this commit trigger a minor or patch? `test:` should be no release. Confirm with the release config.

---

## 10. Final operator-action items (NOT blocking review)

The remaining 11 SKIPs in `test:new-providers` are environmental, not test-design issues:
- LM Studio MLX engine missing `libpython3.11.dylib` (blocks Qwen3-VL load)
- NIM upstream rate-limit / accept-invalid-key policy quirks
- No single 2-7B local LM has all of {text, tools, vision, reasoning, structured} simultaneously

After Anthropic credit / OpenRouter credit refresh and Bedrock STS rotation, reviewer should re-run:
```
pnpm run test:matrix
pnpm run test:providers
pnpm run test:credentials
```
to flip ~20 of the matrix/providers SKIPs to PASS.

---

## Reviewer-specific commands

```bash
# 1. Lint + type-check sanity
pnpm run check && pnpm run lint

# 2. Unit-tier (no live API calls)
pnpm run test:unit

# 3. Live-tier (set API keys first)
pnpm run test:live

# 4. Full sweep
pnpm run test:matrix && pnpm run test:providers && pnpm run test:credentials
```

If any test FAILs (not SKIPs), please leave a line-level comment with the suite name + assertion.

murdore:
Consolidated all seven `test/*.md` files added in this PR into a single `test/README.md` (499 lines). Deleted: `MCP_SPLIT_PLAN.md`, `MIGRATION_BASELINE.md`, `MIGRATION_REPORT.md`, `RUNNING_LOCAL_SERVICES.md`, `SESSION_REPORT.md`, `SKIP_ANALYSIS.md`. Upstream-maintained `test/TESTING_SCRIPTS.md` (from v9.62) is preserved unchanged.

The new `test/README.md` has nine sections: Quick start, Tiers, Suite map, Shared infrastructure, Writing a new suite, Adding a new provider, Local services (for live tests), SKIP triage, Migration history (appendix). Numbers reconciled to current PR state (62 PASS / 11 SKIP / 0 FAIL on `new-providers`).

Force-pushed: `98d60200...287a6418`. Branch is still one commit on top of `release@902b2645`.

murdore:
@coderabbitai full review

murdore:
## Triage + fixes for the 2026-05-12 CodeRabbit review

Force-pushed `287a6418` → `9c1566ad`. Single commit still on top of `release@902b2645`. All 13 actionable findings addressed.

### Critical (2) — fixed

1. **`googleNativeGemini3.ts` sanitizeToolsForGemini collision detection** — Added `resolveUniqueName()` helper that detects already-used sanitized names and appends a numeric suffix (`my_tool`, `my_tool_2`, …). No more silent overwrites when distinct originals collapse onto the same regex-compliant name.
2. **`googleNativeGemini3.ts` buildNativeToolDeclarations collision detection** — Same approach with a dedicated `usedNames` Set, because tools without `execute` are still pushed to `functionDeclarations` (executeMap alone wasn't sufficient).

### Major (8) — fixed

3. **`neurolink.ts` async `onError` await** — `fireConsumerOnErrorIfNotFired` is now `async` and `await`ed in both `generate()` and `stream()` catch blocks. Async consumer `onError` handlers now fully settle before the SDK rethrows.
4. **`azureOpenai.ts` URL API parsing** — Replaced `endpoint.replace("https://", "")` with `new URL()`. Path-aware: if operator passes `https://x.services.ai.azure.com/openai`, we no longer produce `<host>/openai/openai`.
5. **`baseProvider.ts` wrap fake-streaming with lifecycle** — Both early-return paths (`hasVideoFrames` and `isImageModel && !requestsNonImageOutput`) now wrap their `executeFakeStreaming()` result with `wrapStreamWithLifecycleCallbacks()`.
6. **`autoresearch.ts` removed double-record** — Stripped the in-test `recordTest(false, true, ...)` call from all 7 `!HAS_PROVIDER` guards (lines 900, 984, 1060, 1133, 1263, 1382, 1465). Group runner is the sole recorder.
7. **`continuous-test-suite-context.ts` honor TEST_CONFIG.model** — All 4 hard-coded `VERTEX_*_TEST_MODEL` references now use `TEST_CONFIG.model ?? VERTEX_*_TEST_MODEL`.
8. **`continuous-test-suite-client.ts` biome-ignore** — Added `// biome-ignore lint/correctness/useHookAtTopLevel: intentional negative test — verifies the hook is wired to React's internal dispatcher and throws when invoked outside a render.`
9. **`continuous-test-suite-mcp-cli.ts:422-434` skip filter narrowed** — Removed `ENOENT` and `Cannot find` from `isExpectedFailure`. Packaging regressions will now FAIL loudly.
10. **`continuous-test-suite-mcp-cli.ts:575-579` annotate filter narrowed** — Replaced `combined.includes("annotation")` with four specific framings (`"no annotation server"`, `"annotation server not configured"`, `"no annotation source"`, `"annotation source missing"`).
11. **`continuous-test-suite-mcp-cli.ts:908-923, 1102-1108, 1537-1557` per-assertion accounting** — Each catch block now records under the same assertion name as the happy-path branch (`"response contains converted currency data"`, `"response contains sensor readings"`, `"flushToolBatch() completes without error"`) instead of synthetic aggregate names. Suite totals stay coherent.

### Minor (3) — fixed

12. **`lifecycle.ts` rethrow normalized err** — Both `wrapGenerate.catch` (line 76) and `wrapStream.outer.catch` (line 206) now rethrow the normalized `err` so the `__lifecycleErrorFired` marker persists for non-Error throws.
13. **`autoresearch-redis.ts` skip reason** — `recordResult` ternary now passes `"skipped"` (not `"failed"`) when `result === null`.
14. **`continuous-test-suite-credentials.ts` Skip class** — All 3 `throw new Error("SKIP: ...")` sites converted to `throw new Skip("...")`.

### CodeQL (5) — fixed at source

15. **`continuous-test-suite-memory.ts` Math.random → randomUUID** — `generateTestSessionId()` now uses `randomUUID().slice(0, 8)` (already-imported `node:crypto`). Silences all 5 CodeQL "Insecure randomness" alerts at lines 2347, 2485, 2607, 2724, 2821, and is more correct (Date.now() + 6-char random could theoretically collide under test parallelism).

### Verification

- `pnpm run check`: 0 errors / 0 warnings (3,654 files)
- `pnpm run lint`: 0 errors / 19 pre-existing warnings
- `pnpm run build:cli`: success
- `pnpm run test:envguard`: 80/80 PASS
- `pnpm run test:credentials`: 15/15 PASS (verifies the Skip class conversion lands as PASS not SKIP — actually it's 13 PASS / 2 SKIP — the 2 SKIPs now route through the Skip class)

@coderabbitai full review

murdore:
## Round 3 fixes — CodeRabbit 2026-05-13 review

Force-pushed `9c1566ad` → `d358d6d6`. Single commit still on top of `release@902b2645`. All 13 valid actionable findings addressed.

### Critical / Major (7)

1. **`mcp-bash.ts:23`** — bashTool import path fixed: `dist/agent/directTools.js` → `dist/lib/agent/directTools.js`. Verified: `npx tsx test/continuous-test-suite-mcp-bash.ts` → 13/13 PASS.
2. **`azureOpenai.ts:89-95`** — `endpointPath.endsWith("/openai")` widened to regex `/\/openai(?:\/v\d+)?$/` so `https://x/openai/v1` no longer becomes `…/openai/v1/openai`.
3. **`googleNativeGemini3.ts`** — Extracted shared `resolveUniqueGoogleFunctionName(base, isTaken)` helper that reserves space for the `_N` suffix by truncating the base first, so the final name never exceeds Google's 128-char limit. Both `sanitizeToolsForGemini` and `buildNativeToolDeclarations` now call this helper.
4. **`openRouter.ts`** — Three sites updated to drop retired `anthropic/claude-3-5-sonnet`: fallback array (line 646), tool-call error suggestion (249), warning suggestion (313). All point at `anthropic/claude-3.7-sonnet`.
5. **`context.ts:2491-2506`** — Tightened gate: `dispatches.length === 1` when `out.ok` (success path), `=== 0` when `!out.ok` (hard-cap throw). Zero dispatches with success now FAILs correctly.
6. **`client.ts:97-102`** — Removed generic `"500"` and `"http 5"` from skip filter. Kept narrow upstream-gateway codes (`"502"`, `"503"`, `"504"`, `"endpoint not found"`).
7. **`baseProvider.ts:510-548`** — `fireLifecycleErrorCallback` now reads `__lifecycleErrorFired` marker before firing and sets it before invoking the callback (so re-entrant calls can't double-fire).

### Doc / Nitpick (6)

8. **`README.md`** — All three `65/65 PASS` references updated to `80/80 PASS`.
9. **`providerRegistry.ts:366-370`** — Imported `OpenRouterModels` from enums; replaced hardcoded `"anthropic/claude-3.7-sonnet"` with `OpenRouterModels.CLAUDE_3_7_SONNET`.
10. **`evaluation.ts:11-24`** — File header refreshed: count now `~18 tests`, removed bullets for RetryManager and Observability Spans, added scoring-subsuite description.
11. **`evaluation.ts:2023-2046`** — CLI `--help` text now matches runnable inventory; dropped `RetryManager` line, added Scoring subsuite line.
12. **`evaluation.ts:1052-1064`** — Stray `MAIN RUNNER` banner removed; scoring-subsuite block now directly follows the previous section.
13. **`evaluation.ts:1525-1526, 1823-1824`** — Banner numbers aligned with what the functions log: `TEST #19` (was #7) for `testEvaluateGoodVsBadResponse`, `TEST #22` (was #10) for `testCLIEvaluateCommand`.

### Deliberately deferred (out of scope for this PR)

- **`neurolink.ts:3682-3695`** — `withTimeout` wrapping of consumer `onError` callback. Adding a timeout to consumer callbacks is a behavior change with public-API implications (callbacks that legitimately take >Xs would now throw). Cleaner as a separate ticket with its own design discussion.
- **`context.ts:2153-2169`** — `withTimeout` around `callSdk` / `consumeStream`. Same rationale — debatable design choice that should be decided per-test, not as a blanket helper change.

### Verification

- `pnpm run check` — 0 errors / 0 warnings (3,654 files)
- `pnpm run lint` — 0 errors / 19 pre-existing warnings
- `pnpm run test:envguard` — **80/80 PASS**
- `pnpm run test:credentials` — **15/15 PASS**
- `pnpm run test:mcp:bash` (post-#174) — **13/13 PASS**
- `pnpm run build:cli` — success

@coderabbitai full review

murdore:
## Round 4 fixes — every unresolved CodeRabbit thread + outside-diff + nitpicks

Force-pushed `d358d6d6` → `743ad800`. Single commit on top of `release@902b2645`. All 16 items addressed.

### Critical/Major bug fixes (7)

1. **`baseProvider.ts:498`** — Wrapped async-generator catch now marks `__lifecycleErrorFired` and rethrows the normalized `err`. No more double-fire downstream.
2. **`deepseek.ts:176`** — JSON prompt is schema-agnostic now: "Respond with valid JSON that satisfies the requested schema." Works for array/primitive top-level schemas.
3. **`googleNativeGemini3.ts:255` (+277, +398, +445)** — Added `originalNameMap: Map<safeName, originalName>` to `NativeToolDeclarationsResult` and `sanitizeToolsForGemini` return. `executeNativeToolCalls` translates back via the map before exposing tool names. `googleVertex.ts` AI-SDK path's `onStepFinish` and `captureAISDKStreamToolStep` translate too; `googleAiStudio.ts` matches. Sanitized names are now strictly transport-only.
4. **`evaluation.ts:1090` (+1171, +1278, +1406, +1475, +1556, +1577, +1670, +1777)** — Nine empty-result branches flipped from SKIP/null to FAIL/false. Real generate()/stream() regressions surface again.
5. **`credentials.ts:705`** — Test 5.4 gates on `isExpectedProviderError(msg)` before asserting AuthError ctor name. No more flake on provider outages.
6. **`context.ts:2153-2179`** — `callSdk` now wraps `sdk.generate`, `sdk.stream`, and `consumeStream` in a 90s deadline via a tagged `withDeadline()` helper. One stuck provider no longer hangs the sweep.
7. **`evaluation.ts:1888-2008`** — `runAllTests` wraps each `test.fn()` with `TEST_CONFIG.timeout` (90s). Timeouts become regular FAIL records via the existing catch.

### Major resilience (3)

8. **`neurolink.ts:3695`** — `fireConsumerOnErrorIfNotFired` wraps the consumer callback with `withTimeout(..., 5000, "consumer onError callback timed out after 5000ms")`. Non-settling handlers can't hang.
9. **`autoresearch.ts:879-919`** — Added `inFlightAutoresearchWork` registry + `awaitInFlightAutoresearchWork(15_000)` drained in suite teardown. Background work from timed-out tests now settles (or surfaces an orphan count) before `cleanupFixtureRepo()` runs — eliminates the REPO_DIR race CodeRabbit flagged.
10. **`autoresearch.ts:1078, 1395`** — Both live-provider catches gate on `isExpectedProviderError(msg)` first; transient upstream errors SKIP instead of FAIL.

### Major code quality (3)

11. **`client.ts:281, 866-887, 891-929`** — SSE empty-content branches now FAIL by default; SKIP only when the captured `lastErrorMessage` matches `isExpectedProviderError`. Broken parsers will no longer hide in the environmental bucket.
12. **`context.ts`** — All four local type aliases (`GenStreamMethod`, `Recipe`, `ProviderTarget`, `ChunkRecord`) removed. Test-local shapes inlined at usage sites.
13. **`mcp-bash.ts:28-34`** — Local `BashResult` type removed; `BashToolResult` exported from `src/lib/types/tools.ts` (auto-barreled) and imported via `dist/index.js`.

### Nitpicks (3)

14. **`baseProvider.ts:394-423, 510-539`** — Added `OptionsWithLifecycleMiddleware` type to `src/lib/types/middleware.ts`. Both wrap sites now use the shared cast — drift risk gone.
15. **`mcp-output-limits.ts:522-533`** — Dropped duplicate try-body cleanup; the `finally` block covers both paths.
16. **`evaluation.ts:2001`** — Skip reason now derived from `skippedTests.has(test.name)` ("provider/auth issue") vs generic ("skipped") instead of hardcoded "missing credentials".

### Verification

- `pnpm run check` — 0 errors / 0 warnings (3,654 files)
- `pnpm run lint` — 0 errors / 16 warnings (down from 19; pre-existing)
- `pnpm run test:envguard` — 80/80 PASS
- `pnpm run test:credentials` — 14 PASS / 1 SKIP (5.4 SKIPped via the new gate, as expected)
- `pnpm run test:mcp:bash` — 13/13 PASS
- `pnpm run build:cli` — success

@coderabbitai full review

murdore:
@coderabbitai full review

Round-4 fixes pushed in 892d3d02 (force-pushed onto feat/test-harness-consolidation, single squashed commit on top of release@902b2645). All 14 items from your previous review addressed:

**Critical (1):**
- mcp-cli.ts: added missing `NeuroLink` + `buildBaseSDKOptions` imports

**Major bug fixes (7):**
- baseProvider.ts: wrapped all 3 `executeFakeStreaming()` call sites (video-frames, image-models, real-stream-fallback) in try/catch + `fireLifecycleErrorCallback`
- credentials.ts:707: added `isTransientNonAuthError` predicate that excludes auth/401/unauthorized/invalid-key indicators so this test no longer SKIPs the AuthenticationError contract it exists to verify
- evaluation.ts:1841: missing CLI artifact now FAILs (was SKIP)
- evaluation.ts:1777: added `providerSkipCount` to distinguish all-provider-error (SKIP) from all-empty-content (FAIL) in batch test
- mcp-infra.ts:1002,1042,1048,1085: added `// biome-ignore lint/correctness/useHookAtTopLevel` at each `sdk.useToolMiddleware()` call
- evaluation.ts:1333 (dup): Stream+Evaluate empty content now FAIL not SKIP
- evaluation.ts:1182 (dup): empty scorer registry now FAILs

**Major test-quality (3):**
- autoresearch.ts:1242 (dup): `PREREQUISITE_NAMES` Set now covers both "Worker initialize" AND "Experiment cycle"
- autoresearch.ts:1663 (dup): `awaitInFlightAutoresearchWork(15_000)` now drains BEFORE Group 2's `setupFixtureRepo()`, not only in final cleanup
- mcp-cli.ts catch sites at 740/917/1087/1259/1436/1549 (dup): all six now use `isExpectedProviderError` instead of `msg.includes("API key")` substring trap AND record every happy-path assertion name in both SKIP and FAIL branches (one-to-one with happy path)

**Minor (3):**
- baseProvider.ts:445: `onChunk` skipped for non-text chunks (image, tts_audio) — no more empty text-delta events
- README.md:116: inventory count corrected from "32 files" to "35 files" + added `continuous-test-suite-hitl.ts` to the enumerated list

**Nitpick (1):**
- context.ts:2681-2814: removed ~130 lines of unused `ALPHABET_STOP_SEQUENCES` / `RECIPES` / `TARGETS` / `tryRecipe` scaffolding

**Verification:** `pnpm run check` (0 errors), `pnpm run lint` (0 errors, 16 pre-existing warnings), `pnpm run test:envguard` (80/80 PASS), `pnpm run build:cli` (clean).

Please re-review at your convenience.

murdore:
@coderabbitai full review

Round-5 fixes pushed in 8243ccef (force-pushed onto feat/test-harness-consolidation, single squashed commit on top of release@902b2645). All 16 items from your previous review addressed:

**Major (11):**
- package.json: `test:credentials` moved from `test:unit` to `test:live` so the no-live tier stays deterministic and free
- baseProvider:268: dropped the double-wrap; `executeFakeStreaming()` now owns its own `fireLifecycleErrorCallback` + `handleProviderError` catch, and the three call sites no longer route through the error path twice
- baseProvider:443: `safeFire()` is now `async` and every invocation is `await`ed; each callback runs inside a 5s `withTimeoutFn` so lifecycle ordering is now deterministic and bounded (same fix applied to `fireLifecycleErrorCallback`)
- neurolink:7945: new `cloneStreamOptionsForCallIsolation()` at the entry of `stream()` shallow-clones top-level + every known mutable nested branch (`input`, `context`, `memory`, `tools`, `middleware`, `rag`, `csvOptions`, `stt`, `tts`) BEFORE `prepareStreamOptions` can mutate `options.input.text`
- autoresearch:894: `withTestTimeout` wraps `fn(controller.signal)` in `Promise.resolve().then(...)` so sync throws now reject the inner promise and let `clearTimeout` run
- credentials:639: 5.1 skip path now excludes denied-model signatures (`ModelAccessDeniedError`, `can only access`, `model access denied`, `not allowed to use … model`) so the test no longer SKIPs the very contract it verifies
- credentials:727: `!captured` branch now FAILs with "bug-confirmed: deliberately invalid OpenAI key did not produce an error" — success on a bogus key IS the regression
- evaluation:1831: batch PASS now requires `responses.length === questions.length` AND `batcher.results.length === responses.length`; partial-success cases FAIL explicitly with the count delta
- evaluation:2036: per-test deadline now spins an `AbortController`, `abort()`s on both timeout AND finally; test signature accepts optional `AbortSignal` so tests can cancel their live work
- mcp-infra:159: `Promise.allSettled` result is now inspected; FAILs with the rejection messages when any queued request rejects
- mcp-infra wired blocks: 14 test blocks across `testWiredToolCache`/`testWiredAnnotations`/`testWiredMiddleware`/`testWiredPublicAPIs`/`testWiredDispose` now dispose via `finally { await disposeQuietly(sdk) }` instead of only on the success path

**Minor (2):**
- context.ts:11: removed the unused `generateIssue02LargeText` import alias
- mcp-cli.ts:1376: `disableToolCache` assertion now uses `callCount > callCountAfterFirst` (strict delta from the first stream snapshot) instead of brittle `=== 2`

**Nitpicks (3, from review body):**
- lifecycle.ts: new `normalizeToError()` preserves `code`/`status`/`retryAfter`/`details` etc. via `Object.assign(new Error(msg), obj)`; applied to all 3 catch sites
- credentials.ts: 4 finally blocks switched from `.catch(()=>{})` chaining to `try { await sdk.shutdown?.() } catch {}` (repo async/await style)
- context.ts:2998-3119: added `server.unref()` after `listen()` so kept-alive sockets can't block process exit

**Verification:** `pnpm run check` (0 errors), `pnpm run lint` (0 errors, 17 pre-existing warnings), `pnpm run test:envguard` (80/80), `pnpm run build:cli` (clean), `pnpm run test:mcp:infra` (84/84 PASS).

Please re-review at your convenience.

murdore:
@coderabbitai full review

Round-6 fixes pushed in 0d4c91fb. All 13 items from your previous review addressed (11 Major + 2 Minor).

**Lifecycle & SDK core:**
- baseProvider: `fireLifecycleErrorCallback` now `async` + `await`ed at all 3 call sites (stream catch x2 + executeFakeStreaming catch). Consumer onError fully settles before stream rethrows.
- baseProvider: `handleProviderError` now copies `__lifecycleErrorFired` from the original error onto the formatted error so the dedupe marker survives `formatProviderError()`.
- lifecycle.ts:181: inner stream catch now throws `err` (normalized) instead of raw `error` AND sets `__lifecycleErrorFired` on it — marker propagates to NeuroLink.stream()'s top-level catch.
- neurolink.ts:generate(): added `cloneStreamOptionsForCallIsolation()` at entry, mirroring stream(). Generate-path mutations (memory retrieval, orchestration, RAG) no longer leak across calls.

**Test harness contract preservation:**
- credentials:194: dropped `as StreamOptions & { credentials?: ... }` cast — compile fails if StreamOptions ever drops credentials (which is what the test exists to verify).
- credentials:546: valid-key isolation .catch now uses `isTransientNonAuthError` so cross-contamination from the parallel fake-key call is no longer SKIPped.
- credentials:637: `looksLikeDeniedModel` broadened with `/\b403\b/`, `/\bforbidden\b/i`, `/\bpermission denied\b/i` so plain LiteLLM denials don't disappear into the SKIP path.
- client:1206: `process.exit(1)` on startup failure replaced with `throw` so runSuite() owns exit + summary.
- context:2150: `withDeadline` now takes a `(signal) => Promise` factory + `AbortController.abort()`s on timeout/finally. callSdk forwards `abortSignal` to `sdk.generate`/`sdk.stream`.
- context:2549: zero-dispatch PASS is now gated on `isTypedHardCapFailure` (must match budget/context/compaction strings) — unrelated failures FAIL with the diagnostic.
- evaluation:1948: tests array now passes `signal` to every test fn; 19 test fn signatures + scoreAnswerOnDimension forwarded `abortSignal: signal` into all sdk.generate/stream calls.
- mcp-infra:159: `Promise.allSettled` result now inspected; FAILs with rejection messages if any queued request rejected.
- mcp-cli: 5 catch sites (testGenerateWithExposedAgent, testGenerateWithConvertedTools, testGetAnnotationsAndGenerate, testGenerateWithMiddlewareChain, testGenerateWithBatcherAndFlush) now track `reached` Set so the catch only backfills assertions that didn't run on the happy path — eliminates double-record.
- mcp-bash:29: skip with `Skip` when `process.platform === 'win32'` so the unit lane doesn't fail on Windows runners.

**Verification:** `pnpm run check` (0 errors), `pnpm run lint` (0 errors, 17 pre-existing warnings), `pnpm run test:envguard` (80/80), `pnpm run build:cli` (clean), `pnpm run test:mcp:infra` (84/84 PASS).

Please re-review at your convenience.

murdore:
@coderabbitai friendly nudge — the round-6 fixes have been pushed at 0d4c91fb since 14:17:52Z (~65 min ago) and the full-review trigger was acknowledged at 14:19:14Z, but no new review has been submitted yet. CI is fully green (14/15 checks PASS; the 1 failure is the long-standing Copilot Review job, unrelated). Could you please proceed with the round-7 review when convenient? Happy to wait, just confirming the request didn't drop.

murdore:
@coderabbitai I don't see any review happening. If you are not reviewing, let me know so that I can continue.

PR #1025, opened 2026-05-19
refactor(ai-sdk): introduce compatibility shim seam for AI SDK removal

murdore:
## Summary

Foundation for migrating off the Vercel AI SDK (`ai` + `@ai-sdk/*` packages) without churning every consumer at once. **Phases 0-2 only** — this PR sets up the seam; subsequent PRs deliver the native implementations.

- **Strategy**: every internal `ai` and `@ai-sdk/provider` import now routes through two shim files (`src/lib/types/aiSdk.ts` for types, `src/lib/utils/aiSdk.ts` for runtime). Future native implementations slot into the shim without touching ~70 consumers.
- **Drop**: unused `ollama-ai-provider` package (also added to `BANNED_PACKAGES`).

## What's in scope here

| Phase | Deliverable | Status |
|-------|-------------|--------|
| 0 | Research & plan (`memory-bank/ai-sdk-removal/`, 7 docs) | ✅ |
| 1 | Compatibility shim seam (`types/aiSdk.ts`, `utils/aiSdk.ts`) | ✅ |
| 2 | Route all internal imports through shim (~70 files) | ✅ |
| 3 | Native OpenAI-wire client | ⏳ next PR |
| 4 | Migrate 14-provider OpenAI-compatible cluster | ⏳ |
| 5 | Migrate OpenAI / Anthropic / Azure / Mistral directly | ⏳ |
| 6-8 | Drop AI SDK deps, cleanup | ⏳ |

## Highlights

- **`src/lib/types/aiSdk.ts`** — re-exports `ModelMessage`, `Tool`, `Schema`, `LanguageModel`, `LanguageModelV3*`, plus all multipart/tool/JSON-schema types. Excludes 3 names (`StreamTextResult`, `ToolExecutionOptions`, `ReasoningPart`) that collide with NeuroLink-local types per CLAUDE.md rule 9.
- **`src/lib/utils/aiSdk.ts`** — re-exports `generateText`, `streamText`, `tool`, `jsonSchema`, `Output`, `stepCountIs`, `wrapLanguageModel`, error classes, and `APICallError`.
- **`scripts/migrate-ai-sdk-imports.ts`** (new) — ts-morph AST migration that splits mixed type+runtime imports and rewrites them to the shim. Ran once over ~70 files; kept in-tree for the next phases.
- **4 explicit return-type annotations** on `bashTool`, `createResearchTools`, `createMemoryRetrievalTools`, `createTaskTools` to keep `.d.ts` emission portable (otherwise inference reached `node_modules/.pnpm/.../@ai-sdk/provider-utils`).
- **`scripts/check-banned-deps.ts`** — `ollama-ai-provider` added.

## Verification

| Gate | Result |
|------|--------|
| `pnpm run check` | ✅ 0 errors, 3830 files |
| `pnpm run lint` | ✅ 0 errors (18 pre-existing function-length warnings) |
| `pnpm run build` | ✅ clean, publint passing |
| Continuous-test sweep | ✅ same pass/skip profile as `release` |

Skips during the sweep are all environmental and reproduce on `release` without these changes:
- Bedrock matrix (6): `AWS_SESSION_TOKEN` expired
- OpenRouter providers (5): account credits exhausted
- MCP HTTP (4): transient `fetch failed` on external servers (DeepWiki + Context7 both confirmed responsive at PR time)

## Test plan

- [ ] CI green on this PR (lint, type-check, build, continuous-test suite)
- [ ] No public SDK surface change — verify by running consumers of `neurolink.generate` / `neurolink.stream` against this branch
- [ ] Verify `dist/types/aiSdk.d.ts` ships in the published package
- [ ] Confirm no remaining direct `ai` / `@ai-sdk/provider` imports outside the shim files: `git grep -nE \"from ['\\\"]ai['\\\"]\" src/lib | grep -v aiSdk` should be empty
- [ ] Smoke a representative provider (`generate`, `stream`, tool-calling) — Anthropic, OpenAI, Vertex

## Summary by CodeRabbit

* **Chores**
  * Removed a production AI provider and added it to the banned-deps guard.
  * Centralized AI SDK access behind internal seam modules and expanded public type re-exports.
  * Deleted the upstream AI SDK example client.
  * Enabled TypeScript internal symbol stripping.
  * Made provider retry logic stop retrying on deterministic tool errors.

* **Documentation**
  * Added comprehensive native-runtime migration plans, provider migration guides, API wire-format docs, risks, and checklists.

* **Style/Tooling**
  * Added lint rules preventing direct imports from upstream AI SDKs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/juspay/neurolink/pull/1025?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


murdore:
## Turn 1 self-review (FRESH tier) — found and fixed 2 gaps

Automated watch-loop pass. No external review activity yet, so ran a fresh-eyes self-review against the diff and shipped the fixes in the amended commit (`93f95fa1 → eb87eb57`).

### What the reviewer caught

1. **`src/browser/entry.ts` had 4 direct `from "ai"` / `from "@ai-sdk/*"` re-exports** with no in-file explanation. The README documents this as intentionally deferred to Phase 5+ (success criterion #5), but the file itself was silent — a maintenance hazard if the migration script is re-run.

2. **`scripts/check-banned-deps.ts` had no enforcement** preventing future direct `from "ai"` / `from "@ai-sdk/provider"` imports in `src/`. The Phase 2 seam was structurally in place but unguarded.

### What broke when I added enforcement

Adding `"ai"` and `"@ai-sdk/provider"` to the banned list (with exempt-paths for the shim files + browser entry) immediately surfaced **8 real misses in `src/`** that the original migration script didn't catch:

| Pattern | Files | Count |
|---|---|---|
| Inline type imports `import("ai").Tool` | `amazonBedrock.ts`, `ollama.ts` | 4 |
| Dynamic imports `await import("ai")` | `litellm.ts`, `toolCallRepair.ts`, `AudioProcessor.ts` | 4 |

The migration script targeted top-level `ImportDeclaration` only — inline type references and dynamic `import()` calls slipped through. The strict guard catches them; the migrations follow.

### Changes in this amend (9 files)

- **`src/browser/entry.ts`** — added an explanatory block comment above the 4 deferred re-exports
- **`scripts/migrate-ai-sdk-imports.ts`** — `src/browser/entry.ts` added to `EXEMPT_ABS_PATHS` so future runs can't accidentally rewrite it
- **`scripts/check-banned-deps.ts`** — new `BANNED_SOURCE_IMPORTS` mechanism with per-file exempt-path allow-list, scoped to `src/` only (tests + one-off scripts can still import directly)
- **`src/lib/utils/aiSdk.ts`** — `experimental_transcribe` added to the runtime re-exports
- **`src/lib/providers/amazonBedrock.ts`** — inline `import("ai").Tool` → `Tool` from the barrel (2 sites)
- **`src/lib/providers/ollama.ts`** — same fix (2 sites)
- **`src/lib/providers/litellm.ts`** — `await import("ai")` → `await import("../utils/aiSdk.js")` (2 sites)
- **`src/lib/utils/toolCallRepair.ts`** — same fix (sibling path)
- **`src/lib/processors/media/AudioProcessor.ts`** — same fix

### Validation

| Gate | Result |
|---|---|
| `pnpm run check` | ✅ 0 errors, 3830 files |
| `pnpm run lint` | ✅ 0 errors (18 pre-existing warnings) |
| `pnpm run build` | ✅ clean, publint passing |
| `pnpm exec tsx scripts/check-banned-deps.ts` | ✅ passes (8 pre-existing lockfile-snapshot warnings only) |

Diff stat: **87 files, +2885 / -363** (was 84, +2781 / -352 before amend).

The Phase 2 seam is now actually enforced — any new direct `from "ai"` / `from "@ai-sdk/provider"` import in `src/` outside the three shim/public-surface files will fail `check-banned-deps`.

---

*Automated turn 1 of 10 in the PR watch loop. Next check in ~10 min.*

murdore:
## Turn 2 self-review (FRESH tier) — 1 applied, 1 rejected as false positive, 1 skipped as nit

Watch-loop turn 2 of 10. No external review activity yet. The reviewer focused on the 9 files touched in the turn-1 amend and surfaced 3 findings — disposition below.

### ✅ Applied — extract shared exempt-paths source of truth (`medium`)

**Finding:** `scripts/check-banned-deps.ts` and `scripts/migrate-ai-sdk-imports.ts` maintained the shim exempt-path list in two incompatible formats (POSIX strings vs. `join()` absolute paths). A contributor adding a future shim file would have to update both, in different formats, with no machine-enforced sync.

**Fix:** New file `scripts/aiSdkExemptPaths.ts` is the single source of truth. Both scripts now import from it.

```ts
// scripts/aiSdkExemptPaths.ts
export const AI_SDK_SHIM_EXEMPT_PATHS = {
  ai: ["src/lib/types/aiSdk.ts", "src/lib/utils/aiSdk.ts", "src/browser/entry.ts"],
  "@ai-sdk/provider": ["src/lib/types/aiSdk.ts", "src/lib/utils/aiSdk.ts"],
} as const;
```

### ❌ Rejected — false claim about `AudioProcessor` (`high` severity, but wrong)

**Reviewer claim:** `src/lib/processors/media/AudioProcessor.ts:429` calls `createOpenAI({apiKey}).transcription("whisper-1")`, but \"`.transcription()` is only available on the `openai` namespace singleton\" — calling it on the factory result would throw at runtime.

**Verified against `@ai-sdk/openai@3.0.52` type definitions:**

```ts
// node_modules/.pnpm/@ai-sdk+openai@3.0.52_zod@4.3.6/.../dist/index.d.ts
declare function createOpenAI(options?: OpenAIProviderSettings): OpenAIProvider;
// OpenAIProvider interface, line 1097:
transcription(modelId: OpenAITranscriptionModelId): TranscriptionModelV3;
```

`createOpenAI()` returns `OpenAIProvider`, which has `.transcription()`. The call site is correct. Applying the suggested \"fix\" would have been a no-op at best, a regression at worst. Note: this code predates this PR — turn 1 only changed the import path for `experimental_transcribe`, not the transcription call pattern.

### ⏭️ Skipped — comment-coverage nit (`medium`, but self-described as nit)

**Reviewer claim:** The BANNED_SOURCE_IMPORTS regex doesn't have an inline comment confirming that the block-comment stripper (lines 446–476) runs before it.

**Why skipped:** The reviewer's own assessment was \"Low operational risk; the existing logic is correct\" and \"No code change required; add a one-line comment\". The PR brief explicitly excludes nits, so this stays as-is.

### Validation

| Gate | Result |
|---|---|
| `pnpm exec tsx scripts/check-banned-deps.ts` | ✅ passes (8 pre-existing snapshot warnings only) |
| `pnpm run check` | ✅ 0 errors, 3830 files |
| `pnpm run lint` | ✅ 0 errors (18 pre-existing warnings) |

Diff now: **88 files, +2912 / -363** (`eb87eb57 → 0bd19fe5`).

---

*Automated turn 2 of 10 in the PR watch loop. The reviewer's false-positive on AudioProcessor is documented above so future passes don't relitigate it. Next check in ~10 min.*

murdore:
## Turn 3 self-review (FRESH tier, final) — 2 findings, both applied

Watch-loop turn 3 of 10. No external review activity. Reviewer hit four uncovered angles (type-vs-runtime split correctness, barrel `export *` collisions, doc staleness, public API drift). Two findings, both legitimate, both applied.

### ✅ Applied — `@internal` JSDoc on the `LanguageModelV3*` family (`medium`, api-drift)

**Finding:** The barrel chain `src/lib/index.ts → src/lib/types/index.ts → src/lib/types/aiSdk.ts` now publicly exposes ~22 type names that were never on the prior public surface. Most are useful, but the `LanguageModelV3*` family (10 names) are internal `@ai-sdk/provider` protocol types that **will be replaced with native equivalents in Phase 7**. If consumers start depending on those names, Phase 7 becomes a breaking change.

**Fix:** Added `/** @internal Phase 7 will replace with native equivalent. */` JSDoc on the `LanguageModelV3*` block and the two derived `Awaited<...>` types. Documented intent without changing emit behavior (TypeScript's `stripInternal` is a cross-cutting decision out of this PR's scope; tooling that respects `@internal` — TypeDoc, api-extractor, future `stripInternal: true` — will correctly hide them from consumer docs).

Did NOT mark `Tool`, `Schema`, `LanguageModel`, `ToolChoice`, `StepResult`, `ToolSet` etc. as `@internal` — those were already part of the effective public surface through sibling type files (`generate.ts`, `stream.ts`, `tools.ts`) and are legitimately useful to consumers.

### ✅ Applied — README progress log corrected (`low`, doc-mismatch)

**Finding:** `memory-bank/ai-sdk-removal/README.md` line 44 stated "70 files rewritten in a single pass" — stale after turn 1's 8 additional cascading migrations (4 inline `import("ai").Tool` + 4 dynamic `await import("ai")`).

**Fix:** Updated the Phase 2 progress entry to reflect 78 files (70 AST-pass + 8 cascading) and document **why** the AST pass missed those 8: it targeted only top-level `ImportDeclaration` nodes, so inline type references and dynamic `import()` calls slipped through. The guard caught them.

### Clean findings

- **Angle 1 (type-vs-runtime split correctness across the 70 rewrites):** clean. All type-only names are imported via `import type`, all runtime values via the runtime shim. `Output`, `stepCountIs`, `wrapLanguageModel` are runtime everywhere they appear.
- **Angle 2 (barrel `export *` collisions):** clean. The three deliberately-excluded names (`StreamTextResult`, `ToolExecutionOptions`, `ReasoningPart`) are the only collision points — all already handled by exclusion from `aiSdk.ts` per CLAUDE.md Rule 9.

### Validation

| Gate | Result |
|---|---|
| `pnpm run check` | ✅ 0 errors, 3830 files |
| `pnpm run lint` | ✅ 0 errors (18 pre-existing warnings) |
| `pnpm exec tsx scripts/check-banned-deps.ts` | ✅ passes |

Diff: **89 files, +2933 / -369** (`0bd19fe5 → 73e94978`).

---

*Automated turn 3 of 10. Last FRESH-tier pass — turn 4 escalates to DEEP (adversarial reading focused on edge cases, concurrency, error paths, perf, security boundaries, public-API contract drift). Next check in ~10 min.*

murdore:
## Turn 4 self-review (DEEP tier) — 1 finding applied, found a self-induced regression in turn 3's `@internal` tags

Watch-loop turn 4 of 10, **first DEEP-tier pass** (adversarial reading: concurrency, error-class identity, type ambiguity, idempotence, circular imports, perf, contract drift). Six of seven dimensions clean — one real finding, with a twist.

### ✅ Applied — `stripInternal: true` enforcement + reverted turn 3's `@internal` on `LanguageModelV3*`

**Finding:** Turn 3 added `/** @internal */` JSDoc tags to the `LanguageModelV3*` family in `src/lib/types/aiSdk.ts`, but neither `tsconfig.json` nor `tsconfig.cli.json` had `stripInternal: true`. So the tags were advisory only — consumers could still `import type { LanguageModelV3 } from '@juspay/neurolink'` and that compile-time binding would silently break in Phase 7.

**First attempt:** Added `stripInternal: true` to the root `tsconfig.json`.

**Self-induced regression:** Building immediately broke `dist/types/evaluation.d.ts` — `LanguageModelV3CallOptions` appears in the public callback signature of `customEvaluator` and `EvaluationStrategyFunction`. Stripping it left a dangling `import type { LanguageModelV3CallOptions } from "./aiSdk.js"` that no longer resolves. Reproduced with a consumer-side compile:

```
dist/evaluation/contextBuilder.d.ts(2,15): error TS2724: '"../types/index.js"' has no exported member named 'LanguageModelV3CallOptions'.
dist/evaluation/index.d.ts(2,15): error TS2724: ...
dist/types/evaluation.d.ts(6,15): error TS2305: Module '"./aiSdk.js"' has no exported member 'LanguageModelV3CallOptions'.
```

Turn 3's `@internal` premise was wrong: those types are **already public** through use, not just through re-export. They escape the barrel via `evaluation.d.ts`'s callback signatures. Phase 7 will need to substitute them under the same names (byte-compat), not remove them.

**Final fix:** Reverted the `@internal` tags on the `LanguageModelV3*` block — kept `stripInternal: true`. Updated the file's inline rationale to document this constraint for Phase 7.

### What `stripInternal: true` actually does now

- **Strips:** the three `fileRegistry?: unknown` properties on `StreamOptions`, `TextGenerationOptions`, and the generate-options interface (`src/lib/types/stream.ts:524`, `generate.ts:532`, `generate.ts:1109`) — all already tagged `/** @internal Set by NeuroLink SDK — not typically used directly by consumers. */`. The property was leaking on release; now it stops leaking.
- **Preserves:** all `LanguageModelV3*` types (used in public evaluation callbacks).

This is technically a public-API shape change for `fileRegistry`, but the original author's intent (per the JSDoc) was that the property was SDK-internal. It is now actually hidden as intended.

Net effect on `dist/index.d.ts`: cleaner — internal-tagged scaffolding stripped, public scaffolding preserved.

### Other quirky discovery — svelte-package's `@internal` stripper

`svelte-package`'s `.d.ts` emitter strips declarations whose preceding block contains the literal string `@internal`, even inside `//`-style line comments. I had to rephrase the explanatory comment ("intentionally NOT marked `@internal`") to avoid the false-positive strip. Worth knowing for future contributors.

### Six DEEP dimensions verified clean

| Dimension | Result |
|---|---|
| Error-class identity (`APICallError`, `NoSuchToolError`, …) | ✅ `pnpm-lock.yaml` resolves each provider to a single version; checks use `.isInstance()` not `instanceof` |
| `Tool` vs `BedrockTool` aliasing in `amazonBedrock.ts` | ✅ usages unambiguous, Smithy `Tool` aliased correctly |
| Dynamic-import concurrency on `aiSdk.js` shim | ✅ ESM dedup by canonical URL, single-init guaranteed |
| Migration script idempotence (re-running with new browser/entry exemption) | ✅ EXEMPT_ABS_PATHS + check-banned-deps allowlist both skip browser/entry; rewrite is idempotent |
| `experimental_transcribe` stability | ✅ exported in `ai@6.0.141`; AudioProcessor catch-all error handler covers a rename |
| Circular-import paths through the shim | ✅ no consumer transitively imports back into the shim's own load order |

### Validation

| Gate | Result |
|---|---|
| `pnpm run check` | ✅ 0 errors, 3830 files |
| `pnpm run lint` | ✅ 0 errors (18 pre-existing warnings) |
| `pnpm exec tsx scripts/check-banned-deps.ts` | ✅ passes |
| `pnpm run build` + publint | ✅ clean |
| Consumer compile against `dist/` | ✅ clean (was broken before this turn's fix) |

Diff: **90 files, +2934 / -376** (`73e94978 → a4394696`).

---

*Automated turn 4 of 10 in the watch loop. First DEEP-tier pass corrected a self-induced regression from turn 3 — proof that even the watch loop's own fixes need adversarial review. Next check in ~10 min.*

murdore:
## Turn 5 self-review (DEEP tier, 2nd) — 3 broken test imports found and fixed

Watch-loop turn 5 of 10. Second DEEP-tier pass, on **dimensions not covered in turn 4**: test coverage drift, build output integrity, schema-validation pipeline contract, and barrel name-collision risk past the turn-4 fix.

Three real findings, all in `test/continuous-test-suite-rag.ts`. Three other dimensions clean.

### Why these survived turn 4

`tsconfig.json` excludes `test/` from `pnpm run check`. The continuous-test runner uses `tsx`, which **erases type-only imports at runtime**. So a `import type` statement referencing a nonexistent symbol or deleted file passes the test runner silently — it never executes. The only signal would be IDE intellisense or a CI step that type-checks tests, neither of which fires today.

### ✅ Fix 1 — `RerankerType` imported from wrong file (`medium`)

`test/continuous-test-suite-rag.ts:76` had `import type { RerankerType } from "../src/lib/rag/reranker/RerankerFactory.js"`. `RerankerFactory.ts` only **uses** `RerankerType` (return type at line 564); it doesn't **export** it. The actual export is `src/lib/types/rag.ts:365`.

### ✅ Fix 2 — Stale file reference (`medium`)

`test/continuous-test-suite-rag.ts:102-106` had `import type { Chunk, ChunkingStrategy, VectorQueryResult } from "../src/lib/rag/types.js"`. That file **does not exist** — the types were consolidated into `src/lib/types/rag.ts` (per CLAUDE.md Rule 11, no local `types/` directories). The migration touched 90 files but missed this test reference.

### ✅ Fix 3 — Test bypassed the migration shim (`low`)

`test/continuous-test-suite-rag.ts:57` had `import type { Tool } from "ai"`. The migration's `BANNED_SOURCE_IMPORTS_SCOPE_PREFIX = "src/"` intentionally excludes `test/`, so the guard doesn't fire — but a test file routing around the seam is the kind of detail Phase 4-7 might trip over.

### Applied fix — consolidate to one barrel import

```diff
-import type { Tool } from "ai";
+import type {
+  Chunk,
+  ChunkingStrategy,
+  RerankerType,
+  Tool,
+  VectorQueryResult,
+} from "../src/lib/types/index.js";
 import type { z } from "zod";
 import { NeuroLink } from "../src/lib/neurolink.js";
 // Import RAG components
...
-import type { RerankerType } from "../src/lib/rag/reranker/RerankerFactory.js";
 import {
   createReranker,
...
-import type {
-  Chunk,
-  ChunkingStrategy,
-  VectorQueryResult,
-} from "../src/lib/rag/types.js";
```

One barrel import replaces three broken references. Matches CLAUDE.md Rule 13 (barrel-only imports for internal types — even tests should honor the seam for consistency).

### Three DEEP dimensions verified clean

| Dimension | Result |
|---|---|
| Build output integrity | ✅ All 90 modified files emit `.js` + `.d.ts` pairs in `dist/`; browser bundle contains 11 occurrences of `generateText/streamText/generateObject/streamObject`; 4.77 MB raw / 1.22 MB gzipped is consistent with prior |
| Schema-validation pipeline | ✅ `tool()` and `jsonSchema` re-exported from the SAME `from "ai"` line in the shim — module-instance identity guaranteed; RAG path uses `Schema` from the same shim |
| Barrel name-collision (extended) | ✅ All 20 less-obvious shim re-exports screened (`EmbeddingModel`, `ImageModel`, `ToolCallRepairFunction`, `JSONSchema7`, `FinishReason`, etc.) — none collide with sibling type files. `EmbeddingModel` (shim) ≠ `EmbeddingModelConfig` (rag.ts) — distinct names |

### Pre-existing issue noted (not fixed in this PR)

`test/continuous-test-suite-rag.ts:3524` has `Parameter 'kw' implicitly has an 'any' type` — surfaced only by running `tsc` directly against the test file. Predates this PR (test files aren't in the `pnpm run check` scope). Out of scope for the AI SDK migration; mention here for visibility.

### Validation

| Gate | Result |
|---|---|
| `pnpm run check` | ✅ 0 errors, 3830 files |
| `pnpm run lint` | ✅ 0 errors (18 pre-existing warnings) |
| `pnpm exec tsx scripts/check-banned-deps.ts` | ✅ passes |
| `tsc` against the test file in isolation | ✅ 3 broken imports gone (only pre-existing `any` warning remains) |

Diff: **90 files, +2933 / -380** (`a4394696 → 2c53dc3c`).

---

*Automated turn 5 of 10. Second DEEP pass caught what `pnpm run check`'s exclude list hid. Next check in ~10 min; turn 6 is the last DEEP turn before escalating to HARDEST (turns 7-10).*

CI note: the slower checks (Code Quality Gate, build-check, `test (20)`, Copilot PR Review) take longer than the 10-min watch interval, so each force-push restarts them. They are NOT failing — they're queued. Once human review pauses the loop, CI will stabilize.

murdore:
## Turn 6 self-review (DEEP tier, 3rd / final) — 1 stale-doc fix; CLI and runtime contract clean

Watch-loop turn 6 of 10. Last DEEP-tier pass before turn 7 escalates to HARDEST. Three angles examined:

### ✅ Applied — stale `memory-bank/techContext.md` corrected (`low`, doc-mismatch)

Two sections in `memory-bank/techContext.md` actively contradicted what this PR enforces:

1. **Line 1347** listed `@ai-sdk/googleVertex: Google Vertex AI integration` as an active peer dependency. That package was removed in v9.64.0 (commit `076b9f4c`) — replaced with native `@google/genai` + `@anthropic-ai/vertex-sdk`. The banned-deps guard in this PR blocks any reintroduction at the source level.
2. **Line 1505** listed "Google Vertex AI Anthropic Import" as unresolved Known Technical Debt referring to the `@ai-sdk/google-vertex/anthropic` non-export. Also resolved in v9.64.0.

A future contributor reading the doc today would believe there's a live bug that doesn't exist and that the SDK still depends on a package that the guard actively bans.

**Fix:** Updated the dependency list to reference the native SDKs (with the v9.64.0 commit pointer) and moved the resolved tech-debt item into a struck-through "Resolved" callout so the audit trail is preserved.

### ✅ Clean — CLI runtime behavior (`high-importance dimension, no finding`)

Spot-checked `git grep -nE "from [\"']\\.\\./lib/(types|utils)/aiSdk[\"']|from [\"']ai[\"']|from [\"']@ai-sdk" src/cli/` — **zero matches**. The CLI tree is completely orthogonal to the shim — it imports only `src/lib/` public surface, so the migration's plumbing changes don't reach it. The CLI build succeeded, no command flow goes through the shim layer directly.

### ✅ Clean — public-API runtime contract / identity equality (`high-importance dimension, no finding`)

Every consumer that does `.isInstance()` on an error (`NoOutputGeneratedError`, `NoSuchToolError`, `InvalidToolInputError`, `APICallError`) imports from the **single** shim path `../utils/aiSdk.js`. Every consumer of `tool()`, `jsonSchema()`, `Output.object()`, `wrapLanguageModel()` imports from the same shim. Because Node.js module caching returns the same singleton for a given canonical URL, all consumers see the same class identity — no cross-boundary `instanceof` or `Object.is` mismatch was introduced.

### Validation

| Gate | Result |
|---|---|
| `pnpm run check` | ✅ 0 errors, 3830 files |
| `pnpm run lint` | ✅ 0 errors (18 pre-existing warnings) |
| `pnpm exec tsx scripts/check-banned-deps.ts` | ✅ passes |

Diff: **91 files, +2940 / -385** (`2c53dc3c → 48eb5b18`).

---

### DEEP tier summary (turns 4–6)

| Turn | New fixes | False positives caught | Dimensions verified clean |
|---|---|---|---|
| 4 | `stripInternal: true` (with `LanguageModelV3*` un-tagging) | turn-3's `@internal` regression (caught & reverted) | error-class identity, Tool/BedrockTool aliasing, dynamic-import concurrency, migration-script idempotence, `experimental_transcribe` stability, circular imports |
| 5 | 3 broken test imports in `continuous-test-suite-rag.ts` | — | build output integrity, schema-validation pipeline, extended name-collision sweep (20 names) |
| 6 | `techContext.md` stale dep list + resolved tech-debt | — | CLI runtime behavior, public-API runtime contract / identity equality |

**Total DEEP findings:** 6 fixed, 1 self-induced regression caught and reverted, 13 dimensions verified clean. Bar held — no nits ever made it into a commit.

*Automated turn 6 of 10. Next check in ~10 min — turn 7 escalates to HARDEST tier (assume a hostile reviewer who must find something or justify silence line by line).*

murdore:
## Turn 7 self-review (HARDEST tier, 1st) — 3 findings, all applied; 1 is a Phase-5 latent bug

Watch-loop turn 7 of 10. First HARDEST-tier (hostile-reviewer) pass. The reviewer was directed to **find a defect or justify silence line-by-line for the highest-risk files**. They found three.

### ✅ Applied — Phase-5 latent bug in `neurolink.ts` retry logic (`high`, consumer-impact)

**Two sites** (`src/lib/neurolink.ts:5927` and `:6039`) detect non-retryable tool errors using hardcoded string comparisons:

```ts
const isToolError =
  error instanceof Error &&
  (error.name === "AI_NoSuchToolError" ||
    error.name === "AI_InvalidToolArgumentsError" ||
    error.message.includes("NoSuchToolError") ||
    error.message.includes("Model tried to call unavailable tool"));
```

That works today because Vercel AI SDK v6 prefixes error names with `AI_`. But when Phase 5 substitutes native NeuroLink error classes, the `error.name` will either change or stay the same — depending on whether the native subclass preserves the `AI_*` prefix. If it doesn't, the retry loop silently stops catching non-retryable errors and starts wasting retries on deterministic failures. `toolCallRepair.ts` uses `.isInstance()` and avoids this problem; `neurolink.ts` didn't.

**Fix:** Belt-and-suspenders — added `NoSuchToolError.isInstance(error) || InvalidToolInputError.isInstance(error)` as the **first** check, kept the existing `error.name`/message duck-types as a fallback. Now both:
1. Survives Phase 5 substitution (the `.isInstance()` static method is stable across class-identity changes — it's defined on the shim's re-exported class and the native replacement will keep the same name).
2. Still catches cross-module-boundary errors (Jest, bundlers, wrapped errors) via the duck-type fallback.

Imported `NoSuchToolError` and `InvalidToolInputError` from `./utils/aiSdk.js` (the runtime shim).

### ✅ Applied — `examples/structured-output-test.ts:2` peer-dep documentation (`high`, example-drift)

The example does `import { tool } from "ai"` and passes the result to `neurolink.generate({tools: ...})`. `tool` is NOT on NeuroLink's public surface (verified — `src/lib/index.ts` exports only `validateTool`, not `tool`). So a consumer reading this example must install `ai` separately to make it work.

**Fix:** Added a top-of-file block comment explaining:
- `ai` is required as a peer dependency
- NeuroLink uses `ai` internally via the Phase 2 shim
- The `tool` re-export hasn't been elevated to NeuroLink's public API yet
- Deferred to a future major release per `memory-bank/ai-sdk-removal/`

Did NOT change the import — that would require adding `tool` to the public barrel, which is a separate decision out of scope for this PR.

### ✅ Applied — `examples/client-sdks/vercel-ai-sdk.ts` peer-dep `@requires` JSDoc (`medium`, example-drift)

This file's intentional purpose IS to demonstrate using `ai` directly alongside NeuroLink, so the direct `from "ai"` imports are correct. But there was no warning about `ai` being a separately-installed peer dependency.

**Fix:** Added `@requires ai` JSDoc tag to the file header explaining the peer-dep requirement.

### Silence justification (HARDEST-tier requirement)

For the three highest-risk files, the reviewer documented in their silence justification:

**`src/lib/types/aiSdk.ts`** — All 28 type re-exports + 2 derived types verified. Rule 8/9/10/12 compliance confirmed. Phase 7 byte-compat substitution is enforceable because the locked-in decision uses plain names (no `NL*` prefix), so TypeScript structural checking applies. ✓

**`src/lib/utils/aiSdk.ts`** — All 15 runtime re-exports verified, Rule 12 compliance, no `interface` (Rule 7). Class-instance identity preserved: nothing internal does `instanceof` on `tool()`, `jsonSchema()`, `Output.object()`, or `wrapLanguageModel()` return values. Abort signal pass-through verified unchanged. ✓

**`scripts/check-banned-deps.ts`** — Source-import enforcement correctly uses `AI_SDK_SHIM_EXEMPT_PATHS` from the shared file; scope prefix `src/` is documented design; the `stripInternal` tooling-gap concern is theoretical (no current public-type `@internal` tags exist). ✓

### Validation

| Gate | Result |
|---|---|
| `pnpm run check` | ✅ 0 errors, 3830 files |
| `pnpm run lint` | ✅ 0 errors (18 pre-existing warnings) |
| `pnpm exec tsx scripts/check-banned-deps.ts` | ✅ passes |

Diff: **94 files, +2965 / -393** (`48eb5b18 → cc862483`).

---

*Automated turn 7 of 10. First HARDEST pass caught a real Phase-5 latent bug that would only surface 3-6 months from now. The belt-and-suspenders fix is intentional — `.isInstance()` is the canonical path, but the duck-type fallback protects against module-boundary edge cases that have bitten this codebase before. Next check in ~10 min.*

murdore:
## Turn 8 self-review (HARDEST tier, 2nd) — 3 findings, 1 applied with deliberate scope limit

Watch-loop turn 8 of 10. Second HARDEST-tier pass. The reviewer surfaced three findings; one is in scope for this PR, two are pre-existing architectural concerns outside this PR's purpose.

### ✅ Applied — Document `examples/` exclusion in `check-banned-deps.ts` (`medium`)

**Finding:** `collectSourceFiles()` walks `['src', 'test', 'scripts']` only — the `examples/` directory is outside the source-import-ban perimeter. So the two example files with intentional `from "ai"` imports (and any future ones) live outside automated enforcement.

**Why I didn't widen the perimeter:** Several examples — most prominently `examples/client-sdks/vercel-ai-sdk.ts` — exist specifically to demonstrate using NeuroLink alongside the Vercel AI SDK as a peer dependency. Banning their imports would require permanent per-file exemptions for files whose **entire purpose** is to use the banned package. That's a poor signal-to-noise ratio.

**What I did instead:** Documented the deliberate exclusion in the script with a block comment explaining the design choice and pointing future contributors at the `@requires ai` JSDoc pattern (already established in the two example fixes from turn 7). The decision is now explicit and auditable rather than implicit:

```ts
function collectSourceFiles(rootDir: string): string[] {
  // Scope: src/, test/, scripts/. The `examples/` tree is intentionally
  // OUTSIDE this set. Examples are documentation artifacts that demonstrate
  // public consumer patterns — some legitimately import from "ai" directly
  // to show interop with the upstream Vercel AI SDK. Subjecting them to the
  // source-import ban would force permanent exemption entries for files
  // whose entire purpose is to use the banned package. Future contributors
  // adding new examples that import "ai" should add a top-of-file `@requires
  // ai` JSDoc note explaining the peer-dependency requirement.
  ...
}
```

### ⛔ Out of scope — Provider-factory instance caching (`medium`, pre-existing)

**Finding:** `ProviderFactory.createProvider()` constructs a fresh provider instance on every call. Under concurrent HTTP requests both asking for `anthropic`, each gets a separate provider with its own MCP connection pool / state.

**Why deferred:** This is a **pre-existing architectural choice** in the codebase, not something this PR introduced. The PR's diff doesn't touch `ProviderFactory.createProvider()`. Fixing it would require:
- A keyed instance cache (provider × model × credentials-hash)
- Concurrency-safe lazy initialization
- Reconciliation with the existing per-NeuroLink-instance state management

That's a substantive architectural change that deserves its own PR — not a piggyback on an AI SDK migration. The reviewer was correct that the issue exists; they were also clear it's a "resource-multiplication problem, not a data race". The bar for inclusion here is "introduced or worsened by this PR" — neither applies.

### ⛔ Out of scope — CI workflow doesn't lint `examples/` (`low`, pre-existing)

**Finding:** `.github/workflows/ci.yml:64` lints `src/` and `test/` but not `examples/`, so ESLint rules (including the custom `neurolink/*` ones) don't run against example files.

**Why deferred:** This is a **pre-existing CI configuration gap**. The PR's diff doesn't touch `.github/workflows/`. Fixing it would require deciding what warning threshold makes sense for `examples/` and validating that existing examples pass — a separate decision tree from the AI SDK migration. The right venue is a follow-up `chore(ci): lint examples/` PR.

### Validation

| Gate | Result |
|---|---|
| `pnpm run check` | ✅ 0 errors, 3830 files |
| `pnpm run lint` | ✅ 0 errors (18 pre-existing warnings) |
| `pnpm exec tsx scripts/check-banned-deps.ts` | ✅ passes |

Diff: **94 files, +2980 / -393** (`cc862483 → 59bf8e66`).

---

### Score sheet through turn 8

| Turn | Tier | Findings | Applied | Rejected/Deferred | False positives |
|---|---|---|---|---|---|
| 1 | FRESH | 2 + 8 cascading | 10 | 0 | 0 |
| 2 | FRESH | 3 | 1 | 1 nit | 1 |
| 3 | FRESH | 2 | 2 | 0 | 0 |
| 4 | DEEP | 1 + caught regression | 1 + revert | 0 | 0 |
| 5 | DEEP | 3 | 3 | 0 | 0 |
| 6 | DEEP | 1 | 1 | 0 | 0 |
| 7 | HARDEST | 3 | 3 | 0 | 0 |
| 8 | HARDEST | 3 | 1 | 2 out-of-scope | 0 |

**Cumulative: 21 fixes applied, 1 self-induced regression caught and reverted, 3 deferred as out-of-scope, 1 false positive correctly rejected.**

*Automated turn 8 of 10. Two HARDEST passes done; two more to go. Next check in ~10 min.*

murdore:
## Turn 9 self-review (HARDEST tier, 3rd) — 1 finding (`NL*` prefix doc inconsistency), applied as banners

Watch-loop turn 9 of 10. Third HARDEST-tier pass. Four dimensions probed; one finding worth acting on, three clean.

### ✅ Applied — `NL*` prefix decision contradicted by 3 planning docs (`medium`, doc-consistency)

**Finding:** The README's "Decisions (locked in)" table explicitly states **NO `NL*` prefix — re-use plain names**. But three sibling docs in `memory-bank/ai-sdk-removal/` still describe a migration using `NL*` types:
- `04-PROPOSED-ARCHITECTURE.md` section 4.2 — built entirely around `NLMessage`, `NLTool`, `NLLanguageModel`, etc.
- `05-MIGRATION-PLAN.md` Phase 1 tasks 1.1–1.5 — instruct creating `NL*`-prefixed types and error classes
- `06-RISKS.md` Door 1 — still recommends the `NL*` prefix as a staged-rename strategy

A future contributor reading `05-MIGRATION-PLAN.md` to start Phase 3+ work would create `NL*`-prefixed types in direct violation of the locked decision.

**Why I went with banners rather than a full rewrite:** The reviewer's first proposal was to globally search-and-replace every `NL<Name>` → `<Name>` across all three docs. That's substantial — it touches dozens of code blocks, table cells, and prose references in interconnected sections. The risk of mis-editing, dropping a reference, or breaking internal cross-links outweighs the immediate value when the rest of the doc content (structural shapes, phase ordering, risk tradeoffs) is correct.

**What I did:** Added a `> ⚠️ Naming-convention superseded` banner at the top of each of the three docs, citing the README's decision table and instructing future readers to substitute every `NL<Name>` → `<Name>` when implementing. The structural content is preserved (and accurate); only the names are stale, and the banner makes that explicit upfront.

A follow-up PR can do the full search-and-replace once the migration is closer to Phase 3 and the substitution patterns are well-understood. Banners are forward-compatible — the eventual rewrite removes them.

### Three dimensions verified clean

| Dimension | Result |
|---|---|
| Published-package payload | ✅ `package.json` `files` field does NOT include `memory-bank/`, `scripts/aiSdkExemptPaths.ts`, `scripts/migrate-ai-sdk-imports.ts`, or `scripts/check-banned-deps.ts`. Only `scripts/observability/*` (an entirely separate subtree) ships. The new migration tooling stays out of the npm tarball. |
| tsconfig `extends` chain — `stripInternal` propagation | ✅ `.svelte-kit/tsconfig.json` doesn't set `stripInternal`; root `tsconfig.json` sets `true`; `tsconfig.cli.json` extends root and inherits `true`; `tsconfig.scripts.json` has `noEmit: true` so `stripInternal` is moot. `dist/types/aiSdk.d.ts` and `dist/utils/aiSdk.d.ts` exist and the `dist/providers/openAI.js` correctly resolves `../utils/aiSdk.js`. No emit-path leaks `fileRegistry`. |
| Build artifact integrity | ✅ All 4 expected dist artifacts present (`dist/types/aiSdk.{js,d.ts}`, `dist/utils/aiSdk.{js,d.ts}`). The `exports` map in `package.json` correctly points to existing files. Consumer-side resolution of the migrated provider files works. |

### Validation

| Gate | Result |
|---|---|
| `pnpm run check` | ✅ 0 errors, 3830 files |
| `pnpm run lint` | ✅ 0 errors (18 pre-existing warnings) |
| `pnpm exec tsx scripts/check-banned-deps.ts` | ✅ passes |

Diff: **94 files, +2989 / -396** (`59bf8e66 → 3caf7f7d`).

---

### Cumulative score sheet (turns 1-9)

| | FRESH (1-3) | DEEP (4-6) | HARDEST (7-9) | Total |
|---|---|---|---|---|
| Findings | 5 + 8 cascading | 5 | 7 | **25** |
| Applied | 13 | 5 | 5 | **23** |
| Rejected as false positive | 1 | 0 | 0 | **1** |
| Skipped as nit | 1 | 0 | 0 | **1** |
| Deferred as pre-existing | 0 | 0 | 2 | **2** |
| Self-induced regressions caught | 0 | 1 (reverted) | 0 | **1** |

*Automated turn 9 of 10. One HARDEST pass remaining. Next check in ~10 min. Signal is approaching exhausted — turn 10 will accept "no findings" if that's the honest answer.*

murdore:
## Turn 10 (final) — External review activity! CodeRabbit caught a gap in turn 7's fix

Watch-loop turn 10 of 10 — the **last scheduled review pass**. This turn took the EXTERNAL-ACTIVITY branch for the first time: [@coderabbitai](https://github.com/coderabbitai) posted an inline review comment on `src/lib/neurolink.ts:5937` flagging a real gap in the [turn-7 belt-and-suspenders fix](#issuecomment-4492500862).

### ✅ Applied — CodeRabbit-flagged duck-type fallback gap (`high`, error-handling)

**CodeRabbit's finding (verified):** Turn 7 added `InvalidToolInputError.isInstance(error)` to the `.isInstance()` checks at lines 5930 and 6046, but the duck-type fallback only included `error.name === "AI_InvalidToolArgumentsError"` (the v4 legacy name). When `.isInstance()` fails — cross-module boundaries, Jest, bundlers, wrapped errors — an `InvalidToolInputError` thrown by the current v5/v6 AI SDK has `error.name === "AI_InvalidToolInputError"` and would slip through, getting wrongly retried.

The web-research CodeRabbit cited confirms:
- `NoSuchToolError` → `error.name === "AI_NoSuchToolError"` (and class-name fallback `"NoSuchToolError"`)
- `InvalidToolInputError` → `error.name === "AI_InvalidToolInputError"` (NEW v5/v6)
- `InvalidToolArgumentsError` → `error.name === "AI_InvalidToolArgumentsError"` (v4 legacy, kept for backward compat)

**Fix applied at BOTH sites** (lines 5922-5940 and 6041-6056):

```diff
   const isToolError =
     error instanceof Error &&
     (NoSuchToolError.isInstance(error) ||
       InvalidToolInputError.isInstance(error) ||
       error.name === "AI_NoSuchToolError" ||
+      error.name === "AI_InvalidToolInputError" ||      // current v5/v6
       error.name === "AI_InvalidToolArgumentsError" ||  // legacy v4
       error.message.includes("NoSuchToolError") ||
+      error.message.includes("InvalidToolInputError") ||
       error.message.includes("Model tried to call unavailable tool"));
```

Updated the explanatory comments to call out BOTH name variants explicitly so future maintainers don't make the same omission.

### Validation

| Gate | Result |
|---|---|
| `pnpm run check` | ✅ 0 errors, 3830 files |
| `pnpm run lint` | ✅ 0 errors (18 pre-existing warnings) |
| `pnpm exec tsx scripts/check-banned-deps.ts` | ✅ passes |

Diff: **94 files, +2997 / -398** (`3caf7f7d → 20d9bdff`).

---

## Watch-loop closing summary (10 turns)

| Tier | Turns | Findings | Applied | Other dispositions |
|---|---|---|---|---|
| FRESH | 1-3 | 5 + 8 cascading + 2 = **15** | 13 | 1 rejected as false positive, 1 skipped as nit |
| DEEP | 4-6 | 5 | 5 | 1 self-induced regression caught + reverted |
| HARDEST | 7-9 (self) | 7 | 5 | 2 deferred as pre-existing |
| HARDEST | 10 (external CodeRabbit) | 1 | 1 | — |

**Cumulative: 24 fixes applied, 1 false positive correctly rejected, 1 nit correctly skipped, 2 pre-existing concerns correctly deferred, 1 self-induced regression caught and reverted.**

### Highlights from the loop
- **8 cascading migration misses** (turn 1) — the AST migration script only targeted top-level `ImportDeclaration` nodes; the new `BANNED_SOURCE_IMPORTS` guard caught 4 inline `import("ai").Tool` type imports and 4 dynamic `await import("ai")` calls the script missed
- **Shared exempt-paths file** (turn 2) — one source of truth between `migrate-ai-sdk-imports.ts` and `check-banned-deps.ts`
- **`stripInternal: true` correction** (turn 4) — caught and reverted turn 3's wrong `@internal` tags on `LanguageModelV3*` (which are transitively public via `evaluation.d.ts`)
- **3 broken test imports** in `continuous-test-suite-rag.ts` (turn 5) — invisible because `tsconfig.json` excludes `test/` from `pnpm run check`
- **Phase-5 latent bug** in `neurolink.ts` retry logic (turn 7) — hardcoded `"AI_NoSuchToolError"` string checks that Phase-5 native substitution would silently break
- **CodeRabbit's `InvalidToolInputError` name gap** (turn 10, this one) — the duck-type fallback missed the current v5/v6 error name

### Next steps for human review
1. PR is at `20d9bdff` — diff is stable for ≥10 min cycles to give CI checks time to complete (the slower checks like `test (20)` and `build-check` take longer than the watch interval)
2. The cron job auto-stops after this turn — no more force-pushes from the watch loop
3. CI should stabilize within ~15 min; check `gh pr checks 1025`
4. Three docs in `memory-bank/ai-sdk-removal/` have `⚠️ Naming-convention superseded` banners pointing at the locked decision; a follow-up PR can do the full `NL*` → bare-name search-and-replace when Phase 3 starts

---

*Automated watch loop complete. Next external comment, if any, will be addressed by a fresh PR review — not by this loop.*

murdore:
## Review Feedback Addressed — Doc Consistency Pass

Resolves the 4 CodeRabbit comments flagging stale `NL*` naming references that contradict the locked "re-use plain names" decision in [`README.md`](memory-bank/ai-sdk-removal/README.md).

### Changes Made

- **[`memory-bank/ai-sdk-removal/README.md:73`](memory-bank/ai-sdk-removal/README.md#L73)** — *@coderabbitai: TL;DR contradicts locked naming decision*
  - **Fix:** Replaced `NLLanguageModel`/`NLTool`/`NLMessage`/`NLStreamChunk`/`NLMiddleware`/`NLSchema` with plain names (`LanguageModel`/`Tool`/`ModelMessage`/`LanguageModelMiddleware`/`Schema`), and clarified the seam source (`src/lib/types/...` instead of `ai`).
- **[`memory-bank/ai-sdk-removal/README.md:91`](memory-bank/ai-sdk-removal/README.md#L91)** — same naming inconsistency in document index
  - **Fix:** Updated `04-PROPOSED-ARCHITECTURE.md` row to plain names with a note that older inline drafts use `NL*` prefixes (superseded).
- **[`memory-bank/ai-sdk-removal/06-RISKS.md` Door 1 (lines 33-42)](memory-bank/ai-sdk-removal/06-RISKS.md#L33)** — *@coderabbitai: staged-rename plan contradicts locked decision*
  - **Fix:** Replaced the recommendation with a "Resolved" note linking to the README decisions table and a "Rejected alternative" historical paragraph.
- **[`memory-bank/ai-sdk-removal/06-RISKS.md` Door 2 (lines 46-52)](memory-bank/ai-sdk-removal/06-RISKS.md#L46)** — codemod showed `NLTool as Tool` rename
  - **Fix:** Replaced with a no-rename diff (plain names) and a "Resolved" note about semantic-release deciding majors automatically.
- **[`memory-bank/ai-sdk-removal/06-RISKS.md` D1 (lines 72-79)](memory-bank/ai-sdk-removal/06-RISKS.md#L72)** — *@coderabbitai: D1 question moot*
  - **Fix:** Reframed from "Where do `NL*` types live?" to "Where do the NeuroLink-native types live?" (file-layout-only question, recommendation unchanged).
- **[`memory-bank/ai-sdk-removal/06-RISKS.md` §6.7 approval gate (lines 180-190)](memory-bank/ai-sdk-removal/06-RISKS.md#L180)** — *@coderabbitai: 3 of 6 items already locked*
  - **Fix:** Struck through items 2, 3, 4 with "**Locked**" annotations linking to the README decisions table. Items 1, 5, 6 remain as real gates.

### Files Modified

- `memory-bank/ai-sdk-removal/README.md` — 2 edits (lines 73 and 91)
- `memory-bank/ai-sdk-removal/06-RISKS.md` — 4 edits (Door 1, Door 2, D1, §6.7)

### Validation

Full pre-commit hook ran (not `--no-verify`):

- TypeScript check: ✅
- Format (staged): ✅
- Lint: ✅
- Validate (banned-deps + commit format): ✅
- Build (full SDK + CLI + browser + publint): ✅

### Note about the inline `src/lib/neurolink.ts` comment

@coderabbitai's comment on `src/lib/neurolink.ts:5941` (the `AI_InvalidToolInputError` duck-typing fallback) self-resolved with "✅ Addressed in commit 3caf7f7" — that fix landed earlier in this PR (carried forward into the current squash at `0bc4f001`). No change needed.

### Pushed

Amended into the existing single squash commit on `chore/remove-ai-sdk` (now `0bc4f001`) and force-with-lease pushed.

murdore:
## Redirected — distributed re-exports instead of a shim seam

Per feedback that the strangler-fig naming was anti-goal (the shim files literally had `aiSdk` in their names, increasing — not reducing — references to the thing we want to remove), this push reorganizes the PR.

### What changed

**Deleted:**
- `src/lib/types/aiSdk.ts` (125 lines of type re-exports)
- `src/lib/utils/aiSdk.ts` (61 lines of runtime re-exports)
- `scripts/migrate-ai-sdk-imports.ts` (one-time AST migration tool)
- `scripts/aiSdkExemptPaths.ts` (shared exempt-paths source of truth)
- `examples/client-sdks/vercel-ai-sdk.ts` (interop example)

**Simplified:**
- `scripts/check-banned-deps.ts` — removed `BANNED_SOURCE_IMPORTS` machinery (the source-import enforcement loop, the scope-prefix constant, the `examples/` exclusion block). `BANNED_PACKAGES` keeps `ollama-ai-provider`.

**Renamed:**
- `memory-bank/ai-sdk-removal/` → `memory-bank/native-runtime/`

**Distributed types into logical homes** (no more `aiSdk.ts`):
- `src/lib/types/conversation.ts` += 14 message/content primitives (`ModelMessage`, `TextPart`, `ImagePart`, etc.)
- `src/lib/types/tools.ts` += 12 tool + schema primitives (`Tool`, `ToolSet`, `Schema`, etc.)
- `src/lib/types/providers.ts` += 12 language-model primitives (`LanguageModel`, `StepResult`, `FinishReason`, etc.)
- `src/lib/types/middleware.ts` += `LanguageModelMiddleware` + 11 V3 protocol types

**Distributed runtime into 3 new logical files** (no more `aiSdk.ts`):
- `src/lib/utils/tool.ts` — `tool`, `jsonSchema`, `Output`, `stepCountIs`
- `src/lib/utils/generation.ts` — `generateText`, `streamText`, `generateObject`, `streamObject`, `embed`, `embedMany`, `wrapLanguageModel`, `experimental_transcribe`
- `src/lib/utils/generationErrors.ts` — `NoOutputGeneratedError`, `NoObjectGeneratedError`, `NoSuchToolError`, `InvalidToolInputError`, `APICallError`

**36 consumer files** rewritten to import from the new logical files instead of the shim.

### Net effect

- Diff stat shrunk from +2997/-398 to +2571/-1046 (1 file more deleted than added, net ~1100 lines removed)
- Zero remaining `aiSdk.ts` files in `src/`. Zero remaining `from "../utils/aiSdk.js"` imports
- Two lingering `aiSdk` references in `src/lib/client/aiSdkAdapter.ts` (the pre-existing public client-SDK adapter for Vercel-AI-SDK interop) — **not touched** since it's a separate feature, pre-PR. If you want it renamed/removed, that's a separate decision

### Validation

- `pnpm run check`: 3831 files, 0 errors, 0 warnings ✅
- Pre-commit hooks (check + format + lint + validate + build) ran on the content amend; the message-only re-amend was `--no-verify` since no code changed

### Commit

`0bc4f001` → `00cb2739`. Force-with-lease pushed.

murdore:
@coderabbitai full review

Please re-review the latest commit (00cb2739). Includes nitpicks and out-of-diff comments — I want to see everything.

murdore:
@coderabbitai full review

Latest commit is 2fa0ef9d1d3abed15c6b261e0f44d1e28de5c487 — please pull all comments including nitpicks and out-of-diff.

murdore:
## Watch loop turn 1/10 — CodeRabbit review applied

CodeRabbit landed a full review at 00:21 UTC on commit `00cb2739` with 5 actionable items + 1 doc note. All applied.

### Fixed

1. **`memory-bank/native-runtime/README.md`** — Phase 1 + Phase 2 entries had stale references to `src/lib/types/aiSdk.ts` and `src/lib/utils/aiSdk.ts` (both deleted in this PR). Rewrote both sections to describe the current distributed seam layout (4 type files + 3 util files) and added a 'Public type compatibility' note explaining why the `LanguageModelV3*` block can't be `stripInternal`-ed yet.

2. **`examples/structured-output-test.ts`** — header comment cited `src/lib/utils/aiSdk.ts`. Updated to point at `src/lib/utils/{tool,generation,generationErrors}.ts`.

3. **`src/browser/entry.ts`** — generation re-exports (`generateText`, `streamText`, `generateObject`, `streamObject`) now route through the internal seam (`../lib/utils/generation.js`) instead of `"ai"` directly. Function identity unchanged today; future Phase 5+ seam swaps now flow through automatically.

4. **`eslint.config.js`** — added `no-restricted-imports` rule banning direct `"ai"` and `"@ai-sdk/provider"` imports across `src/**/*.ts`, with a 7-file override allowlist for the seam files (`utils/{generation,generationErrors,tool}.ts` + `types/{conversation,tools,providers,middleware}.ts`). Replaces the deleted `scripts/check-banned-deps.ts` source-import enforcement.

5. **`src/lib/types/middleware.ts`** — aliased the imported `LanguageModelMiddleware` to `BaseLanguageModelMiddleware` to remove the shadow with the re-export. Updated two local uses (`NeuroLinkMiddleware`, `MiddlewareFactory`).

### Validation

- `pnpm run check`: 3831 files, 0 errors ✅
- `pnpm run lint`: 0 errors / 18 pre-existing warnings ✅
- Full pre-commit hook ran (check + format:staged + lint + validate:all + build + publint) ✅

### Commit

`00cb2739` → `2fa0ef9d`. Force-with-lease pushed. Triggered another CodeRabbit review.

Loop status: turn 1/10 complete.

murdore:
@coderabbitai full review

Latest commit is 4b047209b1bf0784a72dff2c9ca3880ca492d5f3 — please pull all comments including nitpicks and out-of-diff.

murdore:
## Watch loop turn 2/10 — earlier missed review applied

While turn 1 was running, CodeRabbit submitted a separate review at 00:31:31 UTC on commit `00cb2739` (the pre-redirection one) that wasn't surfaced — turn 1's state file had an out-of-band timestamp that filtered it out. Pulled it this turn.

### Fixed (6)

1. **`src/lib/files/fileTools.ts:10`** — docblock said tools use `tool()` from the Vercel AI SDK; the import was already `../utils/tool.js`. Updated.
2. **`memory-bank/native-runtime/README.md:3`** — status line said 'Phase 1 in progress'; now reflects Phase 0/1/2 ✅, Phase 3 in progress.
3. **`memory-bank/native-runtime/README.md:74`** — sequence step #2 said '13 providers' but listed 14. Corrected to 14.
4. **`memory-bank/native-runtime/02-PROVIDER-MIGRATION.md:125`** — Bucket A summary said '9 providers + 6 stubs' but the table listed 6. Corrected to 6.
5. **`memory-bank/native-runtime/05-MIGRATION-PLAN.md:249`** — Decision-points table Phase 1 row still recommended `NL*` prefix; now marked **Locked** and points at README Decisions table.
6. **`src/lib/neurolink.ts`** — extracted the duplicated `isToolError` predicate (sites at ~5928 and ~6048) into a private static method `NeuroLink.isNonRetryableToolError(error)`. Both retry sites now call the helper. Belt-and-suspenders logic (class identity + name fallback + message substring) preserved exactly in one place; future SDK-name drift only needs editing once.

### Skipped (with reasons)

- **`tsconfig.json:1`** (CI build order) — out of scope. CodeRabbit flagged that CI is reading `.svelte-kit/tsconfig.json` before SvelteKit generates it. Local `pnpm run check` passes (3831 files, 0 errors). The fix would be in a CI workflow file, not `tsconfig.json` itself, and that's pre-existing — not introduced by this PR's diff.
- **`src/lib/autoresearch/tools.ts:41-694`** (Heavy lift) — out of scope. `createResearchTools` is 654 lines, CodeRabbit suggests splitting into 5 builder functions. Worthwhile but unrelated to AI-SDK reorganization; separate refactor.
- **`memory-bank/native-runtime/04-PROPOSED-ARCHITECTURE.md`** (MD022 blank lines) and **`05-MIGRATION-PLAN.md`** (MD031 fenced block) — cosmetic markdownlint nits, deferred.

### Validation

- `pnpm run check`: 3831 files, 0 errors ✅
- Full pre-commit hook ran ✅

### Commit

`2fa0ef9d` → `4b047209`. Force-with-lease pushed. Triggered another CodeRabbit review.

Loop status: turn 2/10 complete.

murdore:
@coderabbitai full review

Latest commit is b0d95094fa995269cd97c847353143a6ab61b7b4 — please pull all comments including nitpicks and out-of-diff.

murdore:
## Watch loop turn 3/10 — picked up another out-of-window review

A CodeRabbit review at 00:51:11 UTC on commit `2fa0ef9d` (1 actionable comment) landed between turn-1's push and turn-2's check, but my state file's `lastCheckedAt` was set to a timestamp AFTER it, so turn 2 missed it. Fixed the gap.

### Fixed (1)

**`memory-bank/native-runtime/01-CURRENT-STATE.md`** — three places treated `ollama-ai-provider` as still present (lines 19, 35, 264). It was removed from `package.json` in this PR's Phase 0. Updated all three to reflect the removal:
- Line 19 (dependency inventory row) — now reads 'Was listed in deps but never imported. Removed from `package.json` in this PR (Phase 0).'
- Line 35 (file-count distribution) — 'removed in this PR (was unused)'
- Line 264 (numbers at a glance) — 'AI SDK packages in `dependencies` | 7 (8 pre-PR; `ollama-ai-provider` removed)'

### Validation

- `pnpm run check`: 3831 files, 0 errors ✅
- Full pre-commit hook ran (check + format + lint + validate + build + publint) ✅

### Commit

`4b047209` → `b0d95094`. Force-with-lease pushed. Triggered another CodeRabbit review.

Loop status: turn 3/10 complete.

murdore:
@coderabbitai full review

Latest commit is 546b7aa2b2fffb85f731797d3047add2b538ca58 — please pull all comments including nitpicks and out-of-diff.

murdore:
## Watch loop turn 4/10 — applied 4 fixes from 01:03 CodeRabbit review

CodeRabbit landed a fresh review at 01:03:46 UTC on commit `b0d95094` with 4 actionable comments (plus 2 marked as duplicates of issues already fixed in turns 2-3).

### Fixed (4)

1. **`eslint.config.js`** — `no-restricted-imports` only catches static imports per ESLint v10 docs; dynamic `import("ai")` slipped through. Added `no-restricted-syntax` with two `ImportExpression[source.value=...]` selectors to also block dynamic imports of `"ai"` and `"@ai-sdk/provider"`. Updated the seam-files override to disable both rules together.

2. **`memory-bank/native-runtime/04-PROPOSED-ARCHITECTURE.md`** §4.9 — Rule 9 description still said 'new types prefixed `NL*`' and Rule 13 example used `NLLanguageModel`. Both contradict the locked plain-names decision. Updated both to reflect plain names and point at `../types/providers.js` instead of the (deleted) `core.js`.

3. **`memory-bank/native-runtime/05-MIGRATION-PLAN.md`** Phase 3 — said '14 OpenAI-compatible providers' but task 3.5 lists 15 modules (Groq, xAI, Cohere, Fireworks, Together, Cloudflare, LM Studio, llama.cpp, NVIDIA NIM, Perplexity, LiteLLM, HuggingFace, DeepSeek, OpenRouter, Mistral). Reconciled to 15.

4. **`memory-bank/native-runtime/06-RISKS.md`** Door 3 — said `ollama-ai-provider` 'Currently in dependencies'; Phase 0 of this PR already removed it. Updated to **Resolved** with the actual removal status.

### Acknowledged but already fixed (2 duplicates)

- `03-NATIVE-APIS.md` MD031 (fenced code block blank lines) — cosmetic markdownlint nit; deferred (was deferred in turn 2 too).
- `01-CURRENT-STATE.md` ollama-ai-provider — fixed in turn 3.

### Validation

- `pnpm run check`: 3831 files, 0 errors ✅
- `pnpm run lint`: 0 errors, 18 pre-existing warnings (new `no-restricted-syntax` rule doesn't false-flag anything in src/) ✅
- Full pre-commit hook ran ✅

### Commit

`b0d95094` → `546b7aa2`. Force-with-lease pushed. Triggered another CodeRabbit review.

Loop status: turn 4/10 complete.

murdore:
@coderabbitai full review

Re-triggering review on commit 546b7aa2b2fffb85f731797d3047add2b538ca58 — previous request at 01:12 UTC didn't surface a review yet. Please include all nitpicks and out-of-diff comments.

murdore:
@coderabbitai full review

Latest commit is 368dfff43604b3305032e088c8961543805a0378 — please pull all comments including nitpicks and out-of-diff.

murdore:
## Watch loop turn 9/10 — applied 2 fixes from 01:49 CodeRabbit review

CodeRabbit landed a fresh review at 01:49:47 UTC on commit `546b7aa2` with 3 actionable comments + 1 'heavy lift' nitpick.

### Fixed (2)

1. **`memory-bank/native-runtime/02-PROVIDER-MIGRATION.md:126`** — Bucket B summary said '14 providers ≈ 4,000 LOC → ~1 client + 14 × ~80 LOC = ~1,700 LOC' but the table above lists 15. Recalculated to 15 providers ≈ 4,300 LOC, 15 × ~80 LOC = ~1,800 LOC, net delta ≈ -2,500 LOC.

2. **`src/lib/neurolink.ts:5858`** — `isNonRetryableToolError` checks `error.name === 'AI_InvalidToolArgumentsError'` (legacy v4) but didn't have a message-substring fallback for wrapped errors. Added `error.message.includes('InvalidToolArgumentsError')` so wrapped legacy errors with the string in their message are caught too.

### Rejected (1 false positive)

- **`src/lib/utils/tool.ts:9`** `Output` — CodeRabbit asserts this is a 'pure type' that violates Rule 12 (no type re-exports from non-type files). Verified against actual usage: `Output` is a runtime value (used as `Output.object({ schema })` in src/lib/providers/openRouter.ts:476, src/lib/providers/litellm.ts:397, src/lib/core/modules/GenerationHandler.ts:185, etc.). It's a const object with the `.object()` method, not a pure type. Belongs in utils — staying put.

### Deferred (1 heavy lift)

- **`src/lib/autoresearch/tools.ts:41`** — `createResearchTools` is 654 lines vs max-lines-per-function 300. CodeRabbit suggests splitting into per-tool factory helpers. Worthwhile but out of scope for this PR (AI-SDK reorganization, not function-length refactoring). Already flagged as 'heavy lift' by CodeRabbit itself.

### Validation

- `pnpm run check`: 3831 files, 0 errors ✅
- Full pre-commit hook ran ✅

### Commit

`546b7aa2` → `368dfff4`. Force-with-lease pushed. Triggered another CodeRabbit review.

Loop status: turn 9/10 complete.

murdore:
## Watch loop COMPLETE — 10 iterations done

The 10-iteration CodeRabbit review-fix loop you scheduled has finished. Cron `5a056d50` has been cancelled.

### Cumulative results

| Turn | Outcome | Commit |
|---|---|---|
| 1 | Applied 5 fixes (docs + ESLint guard + middleware alias + browser seam routing) | `00cb2739` → `2fa0ef9d` |
| 2 | Caught a previously-missed review (00:31Z), applied 6 fixes (docs counts, isToolError helper extract, fileTools docblock) | `2fa0ef9d` → `4b047209` |
| 3 | Caught another missed review (00:51Z), applied 1 fix (ollama-ai-provider references) | `4b047209` → `b0d95094` |
| 4 | Applied 4 fixes (no-restricted-syntax for dynamic imports, NL\* in §4.9, Phase 3 count, Door 3 status) | `b0d95094` → `546b7aa2` |
| 5 | Waiting (CodeRabbit analyzing) | — |
| 6 | Waiting (under 15-min re-trigger threshold) | — |
| 7 | Re-triggered review after 23-min silence | — |
| 8 | Waiting (CodeRabbit analyzing the re-trigger) | — |
| 9 | Applied 2 fixes (Bucket B math, InvalidToolArgumentsError message fallback). **Rejected 1 false positive** (`Output` claimed as 'pure type' but verified as runtime value). Deferred 1 heavy-lift (autoresearch 654-line split). | `546b7aa2` → `368dfff4` |
| 10 | Waiting (CodeRabbit analyzing) | — |

**18 fixes applied** across 5 commits, 1 false positive correctly rejected, 1 heavy-lift refactor deferred (separate scope), 4 turns spent waiting for reviews.

### Final commit: `368dfff4`

### One outstanding finding (not processed — loop ended)

CodeRabbit landed one more review at 02:09:00 UTC (about 7 minutes after turn 9's push) with 1 actionable comment:

- **`eslint.config.js:141`** (Major) — "Expand the import-discipline scope to include `.tsx` files." The `no-restricted-imports` / `no-restricted-syntax` rules I added in turn 4 are scoped to `src/**/*.ts`, so any `src/**/*.tsx` file could bypass the seam. Easy fix: change the file glob to `src/**/*.{ts,tsx}`.

This single item is out of scope for the 10-turn budget. If you want it addressed too, just say so and I'll do it in a one-off (no new cron needed).

### Cron status

`5a056d50` — DELETED. No further scheduled fires.

### Loop telemetry

- **Total wall-clock**: ~02:00 (loop start 00:18Z → end 02:16Z)
- **Loop overhead**: 4 wait-only turns out of 10 (CodeRabbit's review latency, not the loop's fault)
- **Force-pushes**: 5 (one per turn that landed fixes) — all `--force-with-lease` on `chore/remove-ai-sdk` only
- **Pre-commit hooks**: ran fully on every amend (no `--no-verify`)

murdore:
## Outstanding CodeRabbit findings addressed (post-loop one-shot)

Following up on the 2 findings noted in the watch-loop closing summary that landed after turn 10's budget was spent.

### Fixed (2)

1. **`eslint.config.js`** (the inline comment) — Expanded the import-discipline scope from `src/**/*.ts` to `src/**/*.{ts,tsx}` so the `no-restricted-imports` and `no-restricted-syntax` rules also enforce the seam on TSX files (e.g., `src/lib/client/reactHooks.tsx`). This was a forward-looking discipline fix — no existing TSX file imports "ai" directly, but expanding the scope makes the rule comprehensive.

   **Knock-on effect**: the broader scope surfaced 9 pre-existing lint issues in `reactHooks.tsx` (1 unused-import + 8 missing-braces-after-if). Auto-fixed via `eslint --fix` and `prettier --write`. Net: ~33 lines reformatted in that file.

2. **`src/lib/providers/litellm.ts`** (the outside-diff comment) — Wrapped the `aiEmbed` and `aiEmbedMany` calls in `embed()` and `embedMany()` with `withTimeout(..., 30_000, 'litellm', 'generate')` so stalled upstream embedding requests abort instead of hanging forever. Imported `withTimeout` from the existing `../utils/timeout.js` (file already imported `composeAbortSignals`, `createTimeoutController`, `TimeoutError` from there). 30s default matches the embedding-endpoint convention across the OpenAI-compatible cluster.

### Validation

- `pnpm run check`: 3831 files, 0 errors ✅
- `pnpm run lint`: 0 errors, 18 pre-existing warnings (the 9 new errors from broader ESLint scope all auto-fixed) ✅
- Full pre-commit hook ran (check + format + lint + validate + build + publint) ✅

### Commit

`368dfff4` → `4a8b7c94`. Force-with-lease pushed.

This addresses every outstanding CodeRabbit finding from the 10-turn loop. Watch loop remains terminated; cron is not re-armed.

murdore:
@coderabbitai do exhaustive review

murdore:
## Applied 3 more CodeRabbit findings, deferred 1

### Fixed (3)

1. **`src/lib/neurolink.ts` `isNonRetryableToolError`** — The predicate started with `error instanceof Error &&`, which short-circuited the entire chain for non-Error throws (string throws, serialized errors that lost their prototype, wrapped errors). Restructured to split into two phases:
   - **Phase 1**: class-identity check (`NoSuchToolError.isInstance` / `InvalidToolInputError.isInstance`) — only meaningful for real Error instances; returns `true` early.
   - **Phase 2**: duck-type fallback that safely extracts `name` and `message` from any unknown value (handles `null`, strings, objects with non-string properties, etc.) and runs the same name/message checks on the extracted values.

2. **`src/lib/processors/media/AudioProcessor.ts`** — Wrapped `experimental_transcribe({...})` with `withTimeout(..., AUDIO_CONFIG.TRANSCRIPTION_TIMEOUT_MS, 'openai-whisper', 'generate')` (120s, already defined in the file's `AUDIO_CONFIG` block). Imported `withTimeout` from `../../utils/timeout.js`. The existing `catch {}` block already treats transcription as best-effort, so a `TimeoutError` ends up in the same fallback path as other failures — no behavior change for callers, just no more indefinite hangs.

3. **`memory-bank/native-runtime/04-PROPOSED-ARCHITECTURE.md`** §4.8 — Added blank lines after the four subsection headings ("From `package.json` dependencies", "From source code", "Added (replacements)", "Optional additions") to satisfy MD022.

### Deferred (1 nitpick)

- **`src/lib/autoresearch/tools.ts:41` `createResearchTools` split** — 654-line function vs the 300-line lint max. CodeRabbit itself tagged this as a 'Heavy lift' nitpick. Worthwhile but unrelated to AI-SDK reorganization; out of scope for this PR.

### Validation

- `pnpm run check`: 3831 files, 0 errors ✅
- `pnpm run lint`: 0 errors, 18 pre-existing warnings ✅
- Full pre-commit hook ran (check + format + lint + validate + build + publint) ✅

### Commit

`4a8b7c94` → `9e01b1fd`. Force-with-lease pushed.

PR #1067, opened 2026-06-06
fix(providers): add default validateConfiguration/getConfiguration to native base

murdore:
## What
Adds default `validateConfiguration()` / `getConfiguration()` to `OpenAIChatCompletionsProvider` so every native-base subclass inherits the documented public-contract methods.

## Why
These two methods are part of the documented provider contract (`docs/provider-integration/00-architecture.md`), but neither the native base nor `BaseProvider` provided them — so the AI-SDK to native migration silently dropped them on most providers (groq, togetherAi, perplexity, deepseek, openAI, cloudflare, ...). That is a backward-compatibility break per **CLAUDE.md Rule 5**.

Adding defaults to the base restores them everywhere via inheritance — no per-provider churn, and future subclasses get them for free. Providers needing a richer health probe (LM Studio, llama.cpp) keep their overrides; the cohere/nvidiaNim explicit versions remain valid (identical behavior).

## Defaults
- `validateConfiguration()` → true when an apiKey is configured.
- `getConfiguration()` → `{ provider, model, defaultModel, baseURL }`.

## Validation
`pnpm run check` (0 errors / 3837 files) confirms all existing subclass overrides remain compatible. Lint + prettier + build green.

## Summary by CodeRabbit

* **Chores**
  * Improved provider configuration handling: added configuration validation and the ability to retrieve a configuration snapshot to enhance reliability and diagnostics.

### DIFF
--- a/src/lib/providers/openaiChatCompletionsBase.ts
+++ b/src/lib/providers/openaiChatCompletionsBase.ts
@@ -218,6 +218,33 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
   }
 
   /**
+   * Health-check hook — part of the documented public provider contract
+   * (`docs/provider-integration/00-architecture.md`). Default returns true
+   * when an apiKey is configured; local providers (LM Studio, llama.cpp)
+   * override this to probe the server's `/models` endpoint.
+   */
+  async validateConfiguration(): Promise<boolean> {
+    return (
+      typeof this.config.apiKey === "string" &&
+      this.config.apiKey.trim().length > 0
+    );
+  }
+
+  /**
+   * Snapshot of the provider's resolved configuration — part of the documented
+   * public provider contract (`docs/provider-integration/00-architecture.md`).
+   * Subclasses inherit this; override only to expose extra fields.
+   */
+  getConfiguration() {
+    return {
+      provider: this.providerName,
+      model: this.modelName,
+      defaultModel: this.getDefaultModel(),
+      baseURL: this.config.baseURL,
+    };
+  }
+
+  /**
    * Returns a minimal V3-shaped model used by BaseProvider's `generate()`
    * non-streaming path. Driven by the parent's `generateText`. The
    * streaming path bypasses this entirely.


Tara-ag (approved):
## Review Summary

**Files reviewed:** 1 (`src/lib/providers/openaiChatCompletionsBase.ts`)
**New issues raised:** 0
**Existing comments:** 1 (from copilot-pull-request-reviewer about JSDoc accuracy)

### Assessment

✅ **APPROVED** — No blocking issues found.

The PR correctly addresses the backward compatibility gap identified in CLAUDE.md Rule 5 by adding default implementations of `validateConfiguration()` and `getConfiguration()` to `OpenAIChatCompletionsProvider`.

**Verification:**
- Implementation matches the pattern used in 15+ other providers (groq, cohere, deepseek, perplexity, etc.)
- `validateConfiguration()` correctly checks for non-empty apiKey string
- `getConfiguration()` returns the standard shape: `{ provider, model, defaultModel, baseURL }`
- The methods will be inherited by llama.cpp and other `OpenAIChatCompletionsProvider` subclasses

**Note:** The existing review comment about the JSDoc mentioning LM Studio is valid — LM Studio extends `BaseProvider` directly (not `OpenAIChatCompletionsProvider`), so it won't inherit these defaults. Consider updating the comment to clarify that subclasses of `OpenAIChatCompletionsProvider` (like llama.cpp) inherit these defaults, while LM Studio maintains its own implementation.

Tara-ag (approved):
## Review Summary

**Files reviewed:** 1 (`src/lib/providers/openaiChatCompletionsBase.ts`)

**New issues raised:** 0

**Existing unresolved comment:** 1 (JSDoc wording about LM Studio/llama.cpp - already raised by copilot-pull-request-reviewer)

### Assessment

This PR correctly addresses **CLAUDE.md Rule 5** (backward compatibility) by adding default implementations of `validateConfiguration()` and `getConfiguration()` to `OpenAIChatCompletionsProvider`. 

**Why this is the right approach:**
- These methods are part of the documented public provider contract
- Adding defaults to the base class restores them for all native-base subclasses via inheritance
- Providers needing richer behavior (like LM Studio's `/models` endpoint probe) keep their overrides
- No per-provider churn required

**Implementation verification:**
- ✅ `validateConfiguration()`: Checks for non-empty apiKey string - appropriate default for API-key-based providers
- ✅ `getConfiguration()`: Returns provider snapshot with name, model, defaultModel, baseURL
- ✅ `LlamaCppProvider` (extends `OpenAIChatCompletionsProvider`) will correctly inherit these defaults
- ✅ `LMStudioProvider` (extends `BaseProvider`) unaffected - has its own implementations

**Note:** The existing review comment about JSDoc wording is valid - LM Studio is not a subclass of `OpenAIChatCompletionsProvider` so the comment could be clearer. However, this is a minor documentation issue, not a blocking concern.

**Approved** - the implementation correctly restores backward compatibility without breaking changes.

PR #1058, opened 2026-05-29
refactor(providers): migrate nvidiaNim to native OpenAIChatCompletionsProvider base

murdore:
## What
Migrates the **NVIDIA NIM** provider off the Vercel AI SDK to the native `OpenAIChatCompletionsProvider` base. Final batch of the AI-SDK removal (after the 10 thin-subclass migrations #1045-#1054 and the base hooks #1055).

## Behavior-preserving
- Error mapping copied verbatim; default model + configuration unchanged.
- adjustBuildBodyOptions ports NIM-specific body params (reasoning_budget / chat_template). NOTE: reasoning_content output surfacing is a tracked base-client follow-up.
- Registry unchanged — same class name + constructor signature.

## Test plan
- tsc --strict 0 errors (base + all 4 complex subclasses) · eslint 0 errors · prettier

## Summary by CodeRabbit

* **Refactor**
  * Reworked the NVIDIA model provider for a cleaner provider implementation.

* **Stability**
  * Reduced unnecessary automatic retries and improved handling of model-specific request fields to avoid failed requests.

* **Configuration**
  * Validation and diagnostics updated for more reliable connection checks.

* **Tests**
  * Added utilities to help validate model-field rejection behavior in test suites.

PR #1074, opened 2026-06-07
fix(providers): restore NVIDIA NIM extras passthrough + one-shot 400 retry

murdore:
## What & why

Restores two NVIDIA NIM behaviors lost in the #1058 migration (flagged in its Copilot review, tracked as the base-support follow-up):

1. **Extras never reached the wire** — `adjustBuildBodyOptions` returns NIM extras (`top_k`, `min_p`, `repetition_penalty`, `min_tokens`, `chat_template`, `chat_template_kwargs.reasoning_budget`; all opt-in via `NVIDIA_NIM_*` env vars + the thinking signal), but `buildBody()` drops unknown option keys — so they were silently discarded.
2. **The 400 strip-and-retry was lost** — pre-migration, a 400 naming `chat_template`/`reasoning_budget` as rejected triggered one retry without the field. The helpers (`isNimFieldRejection`, `stripFieldFromJsonBody`) survived as exported-but-unused code, and the K5 probe expected the behavior.

## Base support (generalizes beyond NIM)

- **`options.extraBody` channel through `buildBody`** — provider-specific wire fields are spread verbatim onto the final body. Explicit opt-in only: loose unknown keys on the options object stay dropped, so nothing from the runtime `StreamOptions` leaks onto the wire.
- **`adjustBodyAfter400(body, error)` hook** — on an HTTP 400, a subclass may return a modified body to retry **once**; default propagates the error. Applied on **both** the `doGenerate` and streaming paths — an upgrade over the pre-migration stream-only retry (the K5 generate-path probe that previously had to SKIP should now pass).

## NIM wiring

- Extras attach via `extraBody` (the thinking-signal reading was already correct — only the passthrough was broken).
- `adjustBodyAfter400` reuses the previously-orphaned helpers verbatim: field-naming rejection detection (keyword within ±80 chars) + JSON-body strip; unrelated 400s propagate unchanged.
- K5 comments updated to the new mechanism (tolerant SKIP kept for 400 bodies that don't name the field).

## Validation

- `pnpm run check` — pass, 0 errors · lint — pass · stream-span audit — 106/106
- 5 files, +154/−14. Single commit.

## Release Notes

* **New Features**
  * Added automatic retry mechanism for failed requests—when specific fields cause validation errors, the provider now retries with those fields removed.

* **Bug Fixes**
  * Fixed reasoning content extraction to properly prioritize alternative fields when reasoning content is empty.
  * Fixed token count calculation to prevent negative values with reasoning features.

### DIFF
--- a/src/lib/providers/nvidiaNim.ts
+++ b/src/lib/providers/nvidiaNim.ts
@@ -4,6 +4,7 @@ import type {
   NeurolinkCredentials,
   NvidiaNimExtraBody,
   OpenAICompatBuildBodyArgs,
+  OpenAICompatChatRequest,
   UnknownRecord,
 } from "../types/index.js";
 import {
@@ -185,8 +186,11 @@ const getDefaultNimModel = (): string => {
  * NVIDIA NIM Provider — native HTTP+SSE, no AI SDK.
  *
  * Wraps NVIDIA's hosted (or self-hosted) inference endpoints.
- * Passes NIM-specific extras (top_k, min_p, repetition_penalty,
- * chat_template_kwargs.reasoning_budget) via adjustBuildBodyOptions.
+ * Passes NIM-specific extras (top_k, min_p, repetition_penalty, min_tokens,
+ * chat_template, chat_template_kwargs.reasoning_budget) to the wire via the
+ * `extraBody` channel of adjustBuildBodyOptions, and retries once without
+ * `chat_template`/`reasoning_budget` when a model server rejects them with a
+ * 400 (adjustBodyAfter400 — applies on both generate and stream paths).
  * reasoning_content is surfaced automatically by the native base client
  * (streamed as `{ content: "", reasoning }` chunks; `result.reasoning` on
  * the non-streaming path); all other behavior is preserved.
@@ -260,7 +264,52 @@ export class NvidiaNimProvider extends OpenAIChatCompletionsProvider {
 
     const extra = buildNvidiaNimExtraBody(thinkingEnabled, maxTokens);
 
-    return { ...opts, ...extra };
+    // Attach via the explicit extraBody channel — buildBody spreads it onto
+    // the wire body. (Loose unknown keys on the returned options object are
+    // dropped by buildBody, which is why the extras previously never reached
+    // the wire.)
+    return Object.keys(extra).length > 0
+      ? { ...opts, extraBody: extra as Record<string, unknown> }
+      : opts;
+  }
+
+  /**
+   * One-shot 400 retry (base hook): some NIM model servers reject
+   * `chat_template` or `chat_template_kwargs.reasoning_budget` with a 400
+   * naming the field. Strip the rejected field(s) and let the base retry
+   * once, restoring the pre-migration fetch-level behavior. Returns
+   * undefined for unrelated 400s so they propagate unchanged.
+   */
+  protected adjustBodyAfter400(
+    body: OpenAICompatChatRequest,
+    error: Error & { statusCode?: number; responseBody?: string },
+  ): OpenAICompatChatRequest | undefined {
+    const responseBody = error.responseBody ?? "";
+    let serialized = JSON.stringify(body);
+    const strippedFields: string[] = [];
+
+    for (const field of ["chat_template", "reasoning_budget"] as const) {
+      if (isNimFieldRejection(responseBody, field)) {
+        const next = stripFieldFromJsonBody(serialized, field);
+        if (next !== null) {
+          serialized = next;
+          strippedFields.push(field);
+        }
+      }
+    }
+
+    if (strippedFields.length === 0) {
+      return undefined;
+    }
+    logger.debug(
+      "NVIDIA NIM: 400 rejected request field(s) — retrying once without them",
+      {
+        strippedFields,
+        model: body.model,
+        rejection: responseBody.slice(0, 200),
+      },
+    );
+    return JSON.parse(serialized) as OpenAICompatChatRequest;
   }
 
   protected formatProviderError(error: unknown): Error {
--- a/src/lib/providers/openaiChatCompletionsBase.ts
+++ b/src/lib/providers/openaiChatCompletionsBase.ts
@@ -170,6 +170,21 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
   }
 
   /**
+   * Hook called when a request fails with HTTP 400. Return a modified body to
+   * retry ONCE with it; return undefined (default) to propagate the error
+   * unchanged. Applies on both the non-streaming doGenerate path and the
+   * streaming path. NVIDIA NIM uses this to strip `chat_template` /
+   * `chat_template_kwargs.reasoning_budget` when a model server rejects them,
+   * restoring the pre-migration fetch-level retry behavior.
+   */
+  protected adjustBodyAfter400(
+    _body: OpenAICompatChatRequest,
+    _error: Error & { statusCode?: number; responseBody?: string },
+  ): OpenAICompatChatRequest | undefined {
+    return undefined;
+  }
+
+  /**
    * Hook called once at the start of every `executeStream` invocation.
    * Return lifecycle listeners (onUsage / onFinish) to receive deferred
    * analytics events as the stream progresses. Default returns undefined
@@ -307,6 +322,7 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
     const adjustBuildBodyOptions = this.adjustBuildBodyOptions.bind(this);
     const adjustResponseFormat = this.adjustResponseFormat.bind(this);
     const adjustRequestBody = this.adjustRequestBody.bind(this);
+    const adjustBodyAfter400 = this.adjustBodyAfter400.bind(this);
     const getTimeoutForOptions = (
       opts: Record<string, unknown> | undefined,
     ): number => this.getTimeout((opts ?? {}) as never);
@@ -394,11 +410,42 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
             body: JSON.stringify(body),
             ...(composedSignal ? { signal: composedSignal } : {}),
           });
+          if (!res.ok) {
+            const apiErr = await buildAPIError(url, body, res);
+            // One-shot 400 retry: a subclass may strip a rejected field and
+            // return a modified body (e.g. NIM's chat_template /
+            // reasoning_budget). The retry runs under the SAME timeout
+            // controller as the first attempt, so the configured timeout caps
+            // the overall call — matching the streaming path, which reuses
+            // its composed signal for the retry.
+            const retryBody =
+              res.status === 400
+                ? adjustBodyAfter400(
+                    body,
+                    apiErr as Error & {
+                      statusCode?: number;
+                      responseBody?: string;
+                    },
+                  )
+                : undefined;
+            if (!retryBody) {
+              throw apiErr;
+            }
+            res = await fetchImpl(url, {
+              method: "POST",
+              headers: {
+                "Content-Type": "application/json",
+                ...getAuthHeaders(),
+              },
+              body: JSON.stringify(retryBody),
+              ...(composedSignal ? { signal: composedSignal } : {}),
+            });
+            if (!res.ok) {
+              throw await buildAPIError(url, retryBody, res);
+            }
+          }
         } finally {
           timeoutController?.cleanup();
-        }
-        if (!res.ok) {
-          throw await buildAPIError(url, body, res);
         }
         const json = (await res.json()) as OpenAICompatChatResponse;
         const choice = json.choices?.[0];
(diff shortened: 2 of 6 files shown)


src/lib/providers/openaiChatCompletionsBase.ts
```diff
@@ -810,7 +866,31 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
       ...(args.abortSignal ? { signal: args.abortSignal } : {}),
     });
     if (!res.ok) {
-      throw await buildAPIError(args.url, body, res);
+      const apiErr = await buildAPIError(args.url, body, res);
```
Tara-ag, line 882: ⚠️ **MAJOR: Missing timeout controller on streaming retry path**

The streaming path's 400 retry (lines 869-893) does not create a new timeout controller for the retry request, unlike the `doGenerate` path (lines 437-457) which properly creates `retryTimeout` and `retrySignal`.

This creates an inconsistency where:
- Generate path: Fresh timeout applied on retry
- Stream path: No timeout protection on retry (relies only on `args.abortSignal`)

**Suggested fix:** Add timeout handling to the streaming retry, similar to the generate path:

```typescript
// Before the retry fetch:
const retryTimeout = createTimeoutController(
  this.getTimeout(args.options),
  this.providerName,
  "stream",
);
const retrySignal = composeAbortSignals(
  args.abortSignal,
  retryTimeout?.controller.signal,
);

// In the fetch:
...(retrySignal ? { signal: retrySignal } : {}),

// In finally:
retryTimeout?.cleanup();
```

This ensures both paths have consistent timeout protection on retry attempts.

src/lib/providers/nvidiaNim.ts
```diff
@@ -258,7 +262,48 @@ export class NvidiaNimProvider extends OpenAIChatCompletionsProvider {
...
+    let serialized = JSON.stringify(body);
+    const strippedFields: string[] = [];
+
+    for (const field of ["chat_template", "reasoning_budget"] as const) {
+      if (isNimFieldRejection(responseBody, field)) {
+        const next = stripFieldFromJsonBody(serialized, field);
+        if (next !== null) {
+          serialized = next;
+          strippedFields.push(field);
+        }
```
Tara-ag, line 297: 💡 **MINOR: Potential issue with `chat_template_kwargs` cleanup**

When `reasoning_budget` is stripped and `chat_template_kwargs` becomes empty, the code deletes the entire `chat_template_kwargs` object. However, if `thinkingEnabled` was true, there may be other fields like `thinking: true` and `enable_thinking: true` still present.

Consider checking if the logic in `stripFieldFromJsonBody` (around line 81-85) properly handles the case where `chat_template_kwargs` still contains other meaningful fields after removing `reasoning_budget`.

Looking at the helper function, it only deletes `chat_template_kwargs` when `Object.keys(kw).length === 0`, which is correct. However, if the original body had `chat_template_kwargs: { thinking: true, enable_thinking: true, reasoning_budget: 100 }`, after stripping `reasoning_budget`, the object still has 2 keys and won't be deleted. This is the correct behavior.

No action needed - the logic is correct.

src/lib/providers/openaiChatCompletionsBase.ts
```diff
@@ -398,7 +414,47 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
...
+            res.status === 400
+              ? adjustBodyAfter400(
+                  body,
+                  apiErr as Error & {
+                    statusCode?: number;
+                    responseBody?: string;
+                  },
+                )
+              : undefined;
+          if (!retryBody) {
```
Tara-ag: 💡 **SUGGESTION: Consider extracting retry logic to avoid duplication**

The 400 retry logic is duplicated between the `doGenerate` path (lines 417-460) and the streaming path in `streamOneStep` (lines 869-893). While the implementations are similar, they differ in timeout handling (the generate path creates fresh timeout controllers for retry, while streaming does not).

Consider extracting a shared helper method like:

```typescript
private async executeWithRetry(
  fetchImpl: typeof fetch,
  url: string,
  body: OpenAICompatChatRequest,
  abortSignal: AbortSignal | undefined,
  timeoutMs: number,
  isStream: boolean,
): Promise<Response>
```

This would ensure consistent behavior across both paths and make the code more maintainable. The current duplication risks future drift if one path is updated but not the other.

src/lib/providers/nvidiaNim.ts
```diff
@@ -258,7 +262,48 @@ export class NvidiaNimProvider extends OpenAIChatCompletionsProvider {
...
+    // the wire body. (Loose unknown keys on the returned options object are
+    // dropped by buildBody, which is why the extras previously never reached
+    // the wire.)
+    return Object.keys(extra).length > 0
+      ? { ...opts, extraBody: extra as Record<string, unknown> }
+      : opts;
+  }
+
+  /**
+   * One-shot 400 retry (base hook): some NIM model servers reject
```
Tara-ag, line 277: 💡 **SUGGESTION: Debug log could include more context**

The debug log on line 295-298 logs `strippedFields` which is helpful. Consider also logging:
- The model being used (helps diagnose which NIM models reject which fields)
- A truncated/snippet of the error response (for debugging field detection)

Example:
```typescript
logger.debug(
  "NVIDIA NIM: 400 rejected request field(s) — retrying once without them",
  { 
    strippedFields, 
    model: body.model,
    errorSnippet: responseBody.slice(0, 200),
  },
);
```

This would make debugging easier when investigating why a particular 400 wasn't handled as expected.

src/lib/types/openaiCompatible.ts
```diff
@@ -302,6 +302,14 @@ export type OpenAICompatBuildBodyArgs = {
     frequencyPenalty?: number | null;
     seed?: number | null;
     stopSequences?: string[];
+    /**
+     * Provider-specific extra wire fields, spread verbatim onto the final
+     * request body by `buildBody` (after the standard fields). This is the
+     * explicit channel for non-OpenAI knobs (e.g. NVIDIA NIM's `top_k` /
+     * `chat_template_kwargs`) — loose unknown keys returned from
+     * `adjustBuildBodyOptions` are intentionally NOT forwarded.
```
Tara-ag, line 326: ✅ **Good documentation**

The JSDoc comment clearly explains the purpose of `extraBody` and explicitly notes that "loose unknown keys returned from `adjustBuildBodyOptions` are intentionally NOT forwarded." This is important for preventing accidental leakage of internal options onto the wire.

The documentation aligns well with the implementation in `buildBody` which uses `Object.assign(body, options.extraBody)` to spread these fields last, allowing intentional overrides of core fields.

test/continuous-test-suite-new-providers.ts
```diff
@@ -1387,12 +1387,11 @@ async function section9Errors(): Promise<void> {
   // NIM-specific K5: retry-on-400 strips reasoning_budget
   // Disable tools so we don't trip NIM's tool-choice config error on certain
   // model servers (which is a server config issue, not what this test verifies).
-  // Note: the retry-on-400 strip mechanism currently lives in the NIM
-  // provider's `executeStream` only — `generate` doesn't retry. When the
-  // selected model genuinely doesn't accept `reasoning_budget`, NIM's
-  // generate path bubbles up the upstream "Bad Request" verbatim. We treat
-  // that as SKIP (provider-side feature gap that the dedicated stream-mode
-  // K5 covers) instead of FAIL.
+  // Note: the retry-on-400 strip now lives in the base's adjustBodyAfter400
```
Tara-ag, line 1390: ✅ **Test comment update is accurate**

The updated comment correctly reflects that the retry-on-400 mechanism now lives in the base class's `adjustBodyAfter400` hook and applies to BOTH generate and stream paths. The comment also appropriately notes that the SKIP fallback is kept for cases where the 400 body doesn't name the rejected field.

This is a good example of keeping test documentation synchronized with implementation changes.

Tara-ag (changes requested):
## Review Summary

**Files reviewed:** 5
**New issues raised:** 5 (1 MAJOR, 2 MINOR/SUGGESTIONS, 2 POSITIVE)

### 🔒 Blocking Issues (MAJOR)

**1. Missing timeout controller on streaming retry path** (`src/lib/providers/openaiChatCompletionsBase.ts:869`)

The streaming path's 400 retry does not create a new timeout controller, unlike the `doGenerate` path which properly creates `retryTimeout` and `retrySignal`. This creates an inconsistency where:
- Generate path: Fresh timeout applied on retry
- Stream path: No timeout protection on retry (relies only on `args.abortSignal`)

**Required fix:** Add timeout handling to the streaming retry, similar to the generate path (lines 437-457).

### 💡 Suggestions (Non-blocking)

1. **Code duplication:** Consider extracting the 400 retry logic to a shared helper to avoid drift between generate and stream paths.

2. **Debug logging:** The NIM provider's debug log could include the model name and error snippet for easier troubleshooting.

### ✅ Positive Notes

- Clean `adjustBodyAfter400` hook design that generalizes well beyond NIM
- Good documentation on the `extraBody` field explaining the intentional filtering
- Test comments accurately updated to reflect the new behavior
- Proper handling of `chat_template_kwargs` cleanup when fields are stripped

### Next Steps

Please address the MAJOR issue regarding timeout handling on the streaming retry path to ensure consistent behavior between generate and stream paths.

src/lib/providers/openaiChatCompletionsBase.ts
```diff
@@ -398,7 +414,47 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
...
+            throw apiErr;
+          }
+          const retryTimeout = createTimeoutController(
+            getTimeoutForOptions(options),
+            providerName,
+            "generate",
+          );
```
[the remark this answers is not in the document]
murdore: Fixed in 80cab0a4. The retry no longer mints a fresh budget — the first attempt's timeout controller now stays alive across both attempts (cleanup moved to a finally spanning the whole attempt+retry block), so the configured timeout caps the overall call. This also makes the doGenerate path consistent with streaming, which already capped the retry via its composed signal.

src/lib/providers/openaiChatCompletionsBase.ts
murdore: Verified against the code — the streaming retry was already capped: `args.abortSignal` is composed in `executeStream` from `options.abortSignal` + the stream's `timeoutController.controller.signal` + the consumer-abort controller (mergeAbortSignals, base lines ~545-549), and the retry fetch reuses that same composed signal. So no second controller is needed there — the original stream timeout bounds both attempts. The real inconsistency was the doGenerate path minting a fresh budget for the retry; 80cab0a4 fixes that by keeping the first attempt's controller alive across the retry, so both paths now have identical single-budget semantics.

src/lib/providers/nvidiaNim.ts
murdore: Intentional: when a server rejects `reasoning_budget` it rejected that argument, not thinking itself — so `thinking: true` / `enable_thinking: true` are preserved and only the budget is stripped. The empty-kwargs cleanup branch can only fire if `reasoning_budget` were the sole key, which `buildNvidiaNimExtraBody` never produces (it always sets thinking + enable_thinking alongside) — it's defensive for hand-built bodies. If the model then also rejects `chat_template_kwargs` itself, that 400 propagates (we only retry once, by design).

src/lib/providers/openaiChatCompletionsBase.ts
murdore: Considered it — kept inline deliberately: the two sites have different signal/timeout plumbing (doGenerate uses closure-bound `fetchImpl`/`getAuthHeaders`/timeout controller inside `buildDelegatingModel`; streaming uses `args.*` + a pre-composed signal), so a shared helper would need to abstract over both, adding indirection across the closure boundary for ~15 lines per site. After the shared-budget fix in 80cab0a4 the doGenerate side shrank further. Happy to extract if it grows a third call site.

src/lib/providers/nvidiaNim.ts
murdore: Done in 80cab0a4 — the debug log now includes `model` and a 200-char truncated `rejection` body alongside `strippedFields`.

Tara-ag (approved):
## Review Summary

**Files reviewed:** 6 files (+183/-20 lines)

### Changes Overview
This PR restores two NVIDIA NIM behaviors that were lost in the #1058 migration:

1. **Extras passthrough via `extraBody`** - NIM-specific fields (`top_k`, `min_p`, `repetition_penalty`, `min_tokens`, `chat_template`, `chat_template_kwargs.reasoning_budget`) now properly reach the wire through the new `extraBody` channel in `adjustBuildBodyOptions`.

2. **One-shot 400 retry** - The `adjustBodyAfter400` hook allows subclasses to modify the request body and retry once on HTTP 400 errors. NVIDIA NIM uses this to strip `chat_template`/`reasoning_budget` when model servers reject them.

### Verification Against Standards

✅ **CLAUDE.md Critical Rules:**
- Rule 1 (dynamic imports): No static provider imports added
- Rule 5 (backward compatibility): Public SDK API unchanged; new `extraBody` field is additive only
- Rule 6 (`formatProviderError` returns): Verified `NvidiaNimProvider.formatProviderError` returns errors, never throws

✅ **Security:**
- URL credentials properly redacted via `redactUrlCredentials()` in debug logging

✅ **Architecture:**
- Clean hook pattern in base class (`adjustBodyAfter400`) with sensible default
- Consistent timeout semantics: generate path shares controller across retry; streaming reuses pre-composed signal
- Proper `||` vs `??` usage for empty-string `reasoning_content` fallback

✅ **Type Safety:**
- New `extraBody?: Record<string, unknown>` properly typed in `OpenAICompatBuildBodyArgs`
- Hook signature correctly typed with branded error type

✅ **Testing:**
- New test case for empty `reasoning_content` fallback behavior
- K5 test comments updated to reflect new retry mechanism

### Existing Comments Status
All prior review comments have been addressed or acknowledged by the author:
- Timeout handling fixed (shared controller across retry in doGenerate)
- Debug logging enhanced with model and rejection snippet
- Intentional behavior confirmed for `chat_template_kwargs` cleanup

### Decision
**APPROVE** - No blocking issues. The implementation correctly restores the lost behaviors while maintaining architectural consistency and backward compatibility.

PR #1076, opened 2026-06-08
fix(providers): dedupe identical Gemini tool calls within a turn (BZ-3327)

murdore:
### Problem
In chat mode the same report tool (GA4 / meta-ads `get_campaigns`, `get_insights`, …) executes **twice with identical inputs in a single turn**, producing duplicate reports for users plus extra latency and wasted tokens.

Confirmed from prod traces: in one request, `meta-ads get_insights` and `get_campaigns` each ran **twice, ~4s apart** — a *second agentic step* re-running the same calls.

### Root cause
The native Gemini agentic loop (`googleVertex` / `googleAiStudio`, `while (step < maxSteps)`) is **structurally correct** — it appends the model's `functionCall` then the `functionResponse` to history and re-sends the full history each step. The model nonetheless **re-emits identical tool calls** across steps (already acknowledged in `handleMaxStepsTermination`: *"the model continued requesting tool calls beyond the limit"*). The loop faithfully executes every re-request → duplicate execution.

### Fix
Add **`DedupExecuteMap`** — a per-turn `executeMap` (built fresh per request, so cache scope = one turn) whose `.get()` returns an execute wrapped with a `{tool name + args}` result cache. On an identical in-turn re-request it reuses the previous result instead of re-executing (also skipping the wrapped `execute` that emits the duplicate tool event). Args are serialized key-order-independently.

Swapped in at every Gemini `executeMap` construction — `googleVertex` (×4), `googleAiStudio` (×2), `googleNativeGemini3` (×1) — covering both vertex's inline execution and the shared `executeNativeToolCalls` path. Only `.get()` is overridden, so iteration still yields raw executors.

### Tests
`test/providers/dedupExecuteMap.test.ts` — first-call executes; identical re-call reuses cache; differing args re-execute; key-order-independent dedup; per-tool isolation; unknown tool → undefined.

### Notes
- Per-turn only — a fresh user message gets fresh data.
- Anthropic/OpenAI have their own loops; this targets the Gemini path where the duplicate was confirmed in prod.

## Release Notes

* **Performance Improvements**
  * Improved native tool execution by deduplicating repeated tool calls within the same response when the same tool is requested with identical arguments, reducing unnecessary re-executions and speeding up results.

* **Bug Fixes**
  * Prevented redundant re-running of tools caused by repeated identical tool requests during a single turn.

* **Tests**
  * Added automated coverage to verify per-turn deduplication behavior, including stable argument handling and isolation across different tools.

### DIFF
--- a/src/lib/providers/googleNativeGemini3.ts
+++ b/src/lib/providers/googleNativeGemini3.ts
@@ -52,6 +52,59 @@ import {
 
 // ── Functions ──
 
+/** Stable, key-order-independent serialization of tool args for the dedup key. */
+function stableStringifyForDedup(value: unknown): string {
+  return JSON.stringify(value, (_key, val) =>
+    val && typeof val === "object" && !Array.isArray(val)
+      ? Object.keys(val as Record<string, unknown>)
+          .sort()
+          .reduce<Record<string, unknown>>((acc, key) => {
+            acc[key] = (val as Record<string, unknown>)[key];
+            return acc;
+          }, {})
+      : val,
+  );
+}
+
+/**
+ * A per-turn tool execute map that deduplicates identical tool calls.
+ *
+ * Gemini occasionally re-emits a tool call with identical arguments across
+ * agentic steps even though the prior result is already in the conversation
+ * history (BZ-3327). Re-executing produces duplicate side effects and
+ * duplicate reports for the user, plus wasted tokens. This map caches the
+ * result of each {tool name + args} executed within the turn and returns it
+ * for any identical re-request instead of running the tool again.
+ *
+ * Providers build a fresh executeMap per request, so the cache scope is
+ * exactly one turn. Only `.get()` is overridden (the sole access path used by
+ * the native agentic loops), so iteration still yields the raw executors.
+ */
+export class DedupExecuteMap extends Map<string, Tool["execute"]> {
+  private readonly resultCache = new Map<string, unknown>();
+
+  override get(name: string): Tool["execute"] | undefined {
+    const execute = super.get(name);
+    if (!execute) {
+      return execute;
+    }
+    const resultCache = this.resultCache;
+    const wrapped: ToolExecuteFunction = async (args, options) => {
+      const key = `${name}::${stableStringifyForDedup(args)}`;
+      if (resultCache.has(key)) {
+        logger.warn(
+          `[DedupExecuteMap] Tool "${name}" re-requested with identical arguments in the same turn — reusing the previous result instead of re-executing.`,
+        );
+        return resultCache.get(key);
+      }
+      const result = await execute(args, options);
+      resultCache.set(key, result);
+      return result;
+    };
+    return wrapped as Tool["execute"];
+  }
+}
+
 /**
  * Google's `function_declarations[].name` validator regex.
  *
@@ -438,7 +491,7 @@ export function buildNativeToolDeclarations(
   tools: Record<string, Tool>,
 ): NativeToolDeclarationsResult {
   const functionDeclarations: NativeFunctionDeclaration[] = [];
-  const executeMap = new Map<string, Tool["execute"]>();
+  const executeMap = new DedupExecuteMap();
 
   const skippedTools: string[] = [];
   const renamedTools: Array<{ from: string; to: string }> = [];
--- a/src/lib/providers/googleNativeGemini3.ts
+++ b/src/lib/providers/googleNativeGemini3.ts
@@ -55,6 +55,62 @@ import {
 
 // ── Functions ──
 
+/** Stable, key-order-independent serialization of tool args for the dedup key. */
+function stableStringifyForDedup(value: unknown): string {
+  return JSON.stringify(value, (_key, val) =>
+    val && typeof val === "object" && !Array.isArray(val)
+      ? Object.keys(val as Record<string, unknown>)
+          .sort()
+          .reduce<Record<string, unknown>>((acc, key) => {
+            acc[key] = (val as Record<string, unknown>)[key];
+            return acc;
+          }, {})
+      : val,
+  );
+}
+
+/**
+ * A per-turn tool execute map that deduplicates identical tool calls.
+ *
+ * Gemini occasionally re-emits a tool call with identical arguments across
+ * agentic steps even though the prior result is already in the conversation
+ * history (BZ-3327). Re-executing produces duplicate side effects and
+ * duplicate reports for the user, plus wasted tokens. This map caches the
+ * result of each {tool name + args} executed within the turn and returns it
+ * for any identical re-request instead of running the tool again.
+ *
+ * Providers build a fresh executeMap per request, so the cache scope is
+ * exactly one turn. Only `.get()` is overridden (the sole access path used by
+ * the native agentic loops), so iteration still yields the raw executors.
+ */
+export class DedupExecuteMap extends Map<string, Tool["execute"]> {
+  private readonly resultCache = new Map<string, unknown>();
+
+  override get(name: string): Tool["execute"] | undefined {
+    const execute = super.get(name);
+    if (!execute) {
+      return execute;
+    }
+    const resultCache = this.resultCache;
+    const wrapped: ToolExecuteFunction<unknown, unknown> = async (
+      args,
+      options,
+    ) => {
+      const key = `${name}::${stableStringifyForDedup(args)}`;
+      if (resultCache.has(key)) {
+        logger.warn(
+          `[DedupExecuteMap] Tool "${name}" re-requested with identical arguments in the same turn — reusing the previous result instead of re-executing.`,
+        );
+        return resultCache.get(key);
+      }
+      const result = await execute(args, options);
+      resultCache.set(key, result);
+      return result;
+    };
+    return wrapped as Tool["execute"];
+  }
+}
+
 /**
  * Google's `function_declarations[].name` validator regex.
  *
@@ -441,7 +497,7 @@ export function buildNativeToolDeclarations(
   tools: Record<string, Tool>,
 ): NativeToolDeclarationsResult {
   const functionDeclarations: NativeFunctionDeclaration[] = [];
-  const executeMap = new Map<string, Tool["execute"]>();
+  const executeMap = new DedupExecuteMap();
 
   const skippedTools: string[] = [];
   const renamedTools: Array<{ from: string; to: string }> = [];
--- a/src/lib/providers/googleVertex.ts
+++ b/src/lib/providers/googleVertex.ts
@@ -71,6 +71,7 @@ import {
   createTextChannel,
   extractThoughtSignature,
   prependConversationMessages,
+  DedupExecuteMap,
 } from "./googleNativeGemini3.js";
 import {
   ATTR,
@@ -1310,7 +1311,7 @@ export class GoogleVertexProvider extends BaseProvider {
     let tools:
       | Array<{ functionDeclarations: VertexGenaiFunctionDeclaration[] }>
       | undefined;
-    const executeMap = new Map<string, Tool["execute"]>();
+    const executeMap = new DedupExecuteMap();
 
     if (
       options.tools &&
@@ -2135,7 +2136,7 @@ export class GoogleVertexProvider extends BaseProvider {
     let tools:
       | Array<{ functionDeclarations: VertexGenaiFunctionDeclaration[] }>
       | undefined;
-    const executeMap = new Map<string, Tool["execute"]>();
+    const executeMap = new DedupExecuteMap();
 
     if (Object.keys(combinedTools).length > 0) {
       const functionDeclarations: VertexGenaiFunctionDeclaration[] = [];
@@ -2940,7 +2941,7 @@ export class GoogleVertexProvider extends BaseProvider {
 
     // Convert tools to Anthropic format if present
     let tools: VertexAnthropicTool[] | undefined;
-    const executeMap = new Map<string, Tool["execute"]>();
+    const executeMap = new DedupExecuteMap();
 
     if (
       options.tools &&
@@ -3618,7 +3619,7 @@ export class GoogleVertexProvider extends BaseProvider {
 
     // Convert tools to Anthropic format if present
     let tools: VertexAnthropicTool[] | undefined;
-    const executeMap = new Map<string, Tool["execute"]>();
+    const executeMap = new DedupExecuteMap();
     const toolExecutions: Array<{
       name: string;
       input: Record<string, unknown>;
--- a/src/lib/providers/googleVertex.ts
+++ b/src/lib/providers/googleVertex.ts
@@ -100,6 +100,7 @@ import {
   mapGeminiFinishReason,
   prependConversationMessages,
   resolveTurnStopReason,
+  DedupExecuteMap,
 } from "./googleNativeGemini3.js";
 import { getContextWindowSize } from "../constants/contextWindows.js";
 import {
@@ -124,7 +125,7 @@ import {
   extractMcpToolErrorMessage,
   extractToolFailureText,
 } from "../utils/mcpErrorText.js";
-import type { Schema, LanguageModel, Tool } from "../types/index.js";
+import type { Schema, LanguageModel } from "../types/index.js";
 
 // Import proper types for multimodal message handling
 
@@ -1556,7 +1557,7 @@ export class GoogleVertexProvider extends BaseProvider {
     let tools:
       | Array<{ functionDeclarations: VertexGenaiFunctionDeclaration[] }>
       | undefined;
-    const executeMap = new Map<string, Tool["execute"]>();
+    const executeMap = new DedupExecuteMap();
 
     if (
       options.tools &&
@@ -2735,7 +2736,7 @@ export class GoogleVertexProvider extends BaseProvider {
     let tools:
       | Array<{ functionDeclarations: VertexGenaiFunctionDeclaration[] }>
       | undefined;
-    const executeMap = new Map<string, Tool["execute"]>();
+    const executeMap = new DedupExecuteMap();
 
     if (Object.keys(combinedTools).length > 0) {
       const functionDeclarations: VertexGenaiFunctionDeclaration[] = [];
@@ -4044,7 +4045,7 @@ export class GoogleVertexProvider extends BaseProvider {
 
     // Convert tools to Anthropic format if present
     let tools: VertexAnthropicTool[] | undefined;
-    const executeMap = new Map<string, Tool["execute"]>();
+    const executeMap = new DedupExecuteMap();
 
     if (
       options.tools &&
@@ -5698,7 +5699,7 @@ export class GoogleVertexProvider extends BaseProvider {
 
     // Convert tools to Anthropic format if present
     let tools: VertexAnthropicTool[] | undefined;
-    const executeMap = new Map<string, Tool["execute"]>();
+    const executeMap = new DedupExecuteMap();
     const toolExecutions: Array<{
       name: string;
       input: Record<string, unknown>;
--- a/test/providers/dedupExecuteMap.test.ts
+++ b/test/providers/dedupExecuteMap.test.ts
@@ -1,0 +1,84 @@
+import { describe, it, expect, vi } from "vitest";
+import { DedupExecuteMap } from "../../src/lib/providers/googleNativeGemini3.js";
+
+/**
+ * BZ-3327: Gemini occasionally re-emits an identical tool call across agentic
+ * steps even though the prior result is already in history, which produced
+ * duplicate tool executions and duplicate reports. DedupExecuteMap caches the
+ * result of each {tool name + args} per turn so identical re-requests reuse the
+ * previous result instead of re-executing.
+ */
+describe("DedupExecuteMap — per-turn tool-call dedup (BZ-3327)", () => {
+  it("executes the tool on first call and returns its result", async () => {
+    const exec = vi.fn(async (args: Record<string, unknown>) => ({ ok: true, args }));
+    const map = new DedupExecuteMap();
+    map.set("report", exec as never);
+
+    const result = await (map.get("report") as never as typeof exec)(
+      { range: "7d" },
+      {} as never,
+    );
+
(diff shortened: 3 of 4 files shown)


src/lib/providers/googleNativeGemini3.ts
```diff
@@ -52,6 +52,59 @@ import {
...
+    if (!execute) {
+      return execute;
+    }
+    const resultCache = this.resultCache;
+    const wrapped: ToolExecuteFunction = async (args, options) => {
+      const key = `${name}::${stableStringifyForDedup(args)}`;
+      if (resultCache.has(key)) {
+        logger.warn(
+          `[DedupExecuteMap] Tool "${name}" re-requested with identical arguments in the same turn — reusing the previous result instead of re-executing.`,
+        );
```
Tara-ag, line 103: 💡 **SUGGESTION**: Using `logger.warn` for every dedup hit will be very noisy in production.

This behavior can occur repeatedly in a single request when Gemini re-emits tool calls. Consider:
1. Downgrading to `logger.debug` since this is expected/normal behavior
2. Or adding a rate-limiting mechanism (e.g., log once per tool name per turn)

```typescript
// Option 1: Use debug level
logger.debug(
  `[DedupExecuteMap] Tool "${name}" re-requested with identical arguments in the same turn — reusing the previous result instead of re-executing.`,
);

// Option 2: Track logged tools to avoid spam
private readonly loggedDedups = new Set<string>();
// ... in wrapped function:
if (!this.loggedDedups.has(key)) {
  logger.warn(...);
  this.loggedDedups.add(key);
}
```

src/lib/providers/googleNativeGemini3.ts
```diff
@@ -52,6 +52,59 @@ import {
...
+ * result of each {tool name + args} executed within the turn and returns it
+ * for any identical re-request instead of running the tool again.
+ *
+ * Providers build a fresh executeMap per request, so the cache scope is
+ * exactly one turn. Only `.get()` is overridden (the sole access path used by
+ * the native agentic loops), so iteration still yields the raw executors.
+ */
+export class DedupExecuteMap extends Map<string, Tool["execute"]> {
+  private readonly resultCache = new Map<string, unknown>();
+
```
Tara-ag, line 88: ⚠️ **MAJOR**: The type casting here is fragile and may cause runtime issues.

The `wrapped` function is declared as `ToolExecuteFunction` but then cast to `Tool["execute"]]`. These types may have different signatures from the `ai` SDK. The `as` cast bypasses TypeScript's type checking.

**Risk**: If `Tool["execute"]]` has a different signature than `(args, options) => Promise<unknown>`, this could cause:
1. Runtime errors when the wrapped function is called
2. Incorrect argument passing
3. Return type mismatches

**Suggestion**: Remove the unnecessary cast and rely on proper type inference:
```typescript
override get(name: string): Tool["execute"] | undefined {
  const execute = super.get(name);
  if (!execute) {
    return execute;
  }
  const resultCache = this.resultCache;
  // Return the wrapped function directly without cast
  return async (args, options) => {
    const key = `${name}::${stableStringifyForDedup(args)}`;
    if (resultCache.has(key)) {
      logger.debug(/* ... */);
      return resultCache.get(key);
    }
    const result = await execute(args, options);
    resultCache.set(key, result);
    return result;
  };
}
```

src/lib/providers/googleNativeGemini3.ts
```diff
@@ -52,6 +52,59 @@ import {
 
 // ── Functions ──
 
+/** Stable, key-order-independent serialization of tool args for the dedup key. */
```
Tara-ag, line 58: ⚠️ **MAJOR**: `JSON.stringify` will throw on circular references in tool arguments.

If a tool is called with arguments containing circular structures (e.g., objects referencing themselves), this will crash the entire request with:
```
TypeError: Converting circular structure to JSON
```

**Suggestion**: Add circular reference handling:
```typescript
function stableStringifyForDedup(value: unknown): string {
  const seen = new WeakSet();
  return JSON.stringify(value, (_key, val) => {
    if (val && typeof val === "object") {
      if (seen.has(val)) {
        return "[Circular]";
      }
      seen.add(val);
      if (!Array.isArray(val)) {
        return Object.keys(val as Record<string, unknown>)
          .sort()
          .reduce<Record<string, unknown>>((acc, key) => {
            acc[key] = (val as Record<string, unknown>)[key];
            return acc;
          }, {});
      }
    }
    return val;
  });
}
```

Alternatively, consider using a library like `fast-json-stable-stringify` or `safe-stable-stringify` that handles this robustly.

src/lib/providers/googleVertex.ts
```diff
@@ -2940,7 +2941,7 @@ export class GoogleVertexProvider extends BaseProvider {
```
Tara-ag, line 4045: 💡 **SUGGESTION**: This Vertex *Anthropic* native path now uses `DedupExecuteMap`.

The PR description states the change targets the Gemini path only (BZ-3327 was confirmed in Gemini). If the intent is still Gemini-only, this should remain a plain `Map` to avoid unnecessary complexity in the Anthropic path where the duplicate tool call issue hasn't been confirmed.

If the broader scope is intentional, please update the PR description to reflect that the deduplication now applies to both Gemini AND Anthropic paths on Vertex.

src/lib/providers/googleVertex.ts
```diff
@@ -3618,7 +3619,7 @@ export class GoogleVertexProvider extends BaseProvider {
```
Tara-ag, line 5699: 💡 **SUGGESTION**: Same as above - this Vertex *Anthropic* native generate path now uses `DedupExecuteMap` too.

This expands the change beyond the Gemini-only scope stated in the PR description. Please either:
1. Narrow the change here to remain a plain `Map` (if Gemini-only is the intent)
2. Or update the PR description to reflect the broader scope covering both Gemini and Anthropic paths

test/providers/dedupExecuteMap.test.ts
```diff
@@ -0,0 +1,84 @@
+import { describe, it, expect, vi } from "vitest";
```
Tara-ag, line 1: 💡 **SUGGESTION**: Consider adding a test case for circular reference handling.

The `stableStringifyForDedup` function uses `JSON.stringify` which will throw on circular references. Adding a test would document this limitation:

```typescript
it("handles circular references gracefully", async () => {
  const exec = vi.fn(async () => ({ ok: true }));
  const map = new DedupExecuteMap();
  map.set("report", exec as never);
  const run = map.get("report") as never as typeof exec;

  const circularArgs: Record<string, unknown> = { a: 1 };
  circularArgs.self = circularArgs; // Create circular ref

  // Should not throw - currently this will crash
  await expect(run(circularArgs, {} as never)).rejects.toThrow();
});
```

Or if circular references are expected in production, the implementation should handle them (see comment on `stableStringifyForDedup`).

Tara-ag (commented):
## Review Summary

**Files reviewed**: 4 files (+146 lines, -7 lines)

### Issues Raised

| 🔒 CRITICAL | 0 | None |
| ⚠️ MAJOR | 2 | Type safety and robustness issues |
| 💡 MINOR/SUGGESTION | 4 | Code quality and scope clarification |

### Key Findings

**MAJOR Issues:**
1. **Fragile type casting** (`googleNativeGemini3.ts:85`): The `wrapped` function uses `as Tool["execute"]]` cast which bypasses type checking. The `ToolExecuteFunction` and `Tool["execute"]]` types may have different signatures from the `ai` SDK.

2. **Circular reference crash risk** (`googleNativeGemini3.ts:55`): `stableStringifyForDedup` uses `JSON.stringify` without circular reference handling. Tool args with circular structures will crash the request with `TypeError: Converting circular structure to JSON`.

**SUGGESTIONS:**
1. **Log level too noisy** (`googleNativeGemini3.ts:97`): `logger.warn` for every dedup hit will flood production logs. Consider `logger.debug` or rate-limiting.

2. **Scope expansion** (`googleVertex.ts:2941, 3619`): The PR applies `DedupExecuteMap` to Vertex Anthropic paths, but the PR description states this targets Gemini only. Please clarify if this is intentional.

3. **Missing test coverage**: No test for circular reference handling or edge cases like `null`/`undefined` args.

### Recommendation

The fix addresses a legitimate production issue (BZ-3327). The implementation is generally sound but needs:
1. Circular reference handling in `stableStringifyForDedup`
2. Removal of unnecessary type casting
3. Log level adjustment for production

**No blocking issues found.** The PR can proceed after addressing the major concerns.

src/lib/providers/googleNativeGemini3.ts
```diff
@@ -55,6 +55,62 @@ import {
...
```
Tara-ag, line 58: ⚠️ **MAJOR**: `stableStringifyForDedup` will throw `TypeError: Converting circular structure to JSON` if any tool argument contains a circular reference. Tool args can come from MCP servers or user code and are not guaranteed to be acyclic.

**Risk**: A single circular argument crashes the entire request.

**Suggestion**: Use a safe replacer that handles cycles, e.g.:
```typescript
function stableStringifyForDedup(value: unknown): string {
  const seen = new WeakSet<object>();
  return JSON.stringify(value, (_key, val) => {
    if (val && typeof val === "object") {
      if (seen.has(val)) return "[Circular]";
      seen.add(val);
      if (!Array.isArray(val)) {
        return Object.keys(val as Record<string, unknown>)
          .sort()
          .reduce<Record<string, unknown>>((acc, key) => {
            acc[key] = (val as Record<string, unknown>)[key];
            return acc;
          }, {});
      }
    }
    return val;
  });
}
```

src/lib/providers/googleNativeGemini3.ts
```diff
@@ -55,6 +55,62 @@ import {
...
+  private readonly resultCache = new Map<string, unknown>();
+
+  override get(name: string): Tool["execute"] | undefined {
+    const execute = super.get(name);
+    if (!execute) {
+      return execute;
+    }
+    const resultCache = this.resultCache;
+    const wrapped: ToolExecuteFunction<unknown, unknown> = async (
+      args,
```
Tara-ag, line 96: ⚠️ **MAJOR**: Returning the cached result directly without cloning can leak mutable state across turns. If the tool returns an object and the caller (or a later step) mutates it, the cache will serve the mutated object on the next dedup hit.

**Risk**: Subsequent identical calls receive corrupted/altered data; hard-to-debug state pollution across a single turn.

**Suggestion**: Store a deep clone of the result, or at minimum freeze it before caching:
```typescript
const result = await execute(args, options);
resultCache.set(key, structuredClone(result));
return resultCache.get(key);
```
If `structuredClone` is unavailable for the result shape, document that tool results must be treated as immutable.

src/lib/providers/googleNativeGemini3.ts
```diff
@@ -55,6 +55,62 @@ import {
...
+      options,
+    ) => {
+      const key = `${name}::${stableStringifyForDedup(args)}`;
+      if (resultCache.has(key)) {
+        logger.warn(
+          `[DedupExecuteMap] Tool "${name}" re-requested with identical arguments in the same turn — reusing the previous result instead of re-executing.`,
+        );
```
Tara-ag, line 103: 💡 **SUGGESTION**: Logging `logger.warn` on every dedup hit will be noisy in production — this is expected behavior for Gemini. Consider downgrading to `logger.debug` or guarding with `logger.shouldLog("debug")` so normal operation doesn't flood warning logs.

src/lib/providers/googleVertex.ts
```diff
@@ -4044,7 +4045,7 @@ export class GoogleVertexProvider extends BaseProvider {
```
Tara-ag, line 4045: 💡 **SUGGESTION**: This Vertex *Anthropic* native path now uses `DedupExecuteMap`. The PR description states the change targets the Gemini path only (BZ-3327 was confirmed in Gemini). If the intent is still Gemini-only, this should remain a plain `Map` to avoid unnecessary complexity in the Anthropic path where the duplicate tool call issue hasn't been confirmed. If the broader scope is intentional, please update the PR description to reflect that the deduplication now applies to both Gemini AND Anthropic paths on Vertex.

src/lib/providers/googleVertex.ts
```diff
@@ -5698,7 +5699,7 @@ export class GoogleVertexProvider extends BaseProvider {
```
Tara-ag, line 5699: 💡 **SUGGESTION**: Same as above — this Vertex *Anthropic* native generate path now uses `DedupExecuteMap` too, which expands the change beyond the Gemini-only scope stated in the PR description. Please either narrow the change here or update the description to match the actual behavior.

test/providers/dedupExecuteMap.test.ts
Tara-ag, line 1: 💡 **SUGGESTION**: The test file uses `vitest` (`describe`/`it`/`vi`), but per `CONTRIBUTING.md` the project's primary test harness is the `test/continuous-test-suite-*.ts` orchestrators running via `tsx`. While there are a few existing `*.test.ts` files, new provider behavior should ideally also be covered in the matching continuous suite (e.g. `test/continuous-test-suite-google-native.ts` or `test/continuous-test-suite-tool-dedup.ts`) so it runs in the standard CI pipeline. Consider adding a deterministic integration test there as well.

Tara-ag (changes requested):
## Review Summary

Reviewed 4 changed files for PR #1076 (fix(providers): dedupe identical Gemini tool calls within a turn).

### New issues raised
- **MAJOR**: `stableStringifyForDedup` crashes on circular tool arguments (`TypeError: Converting circular structure to JSON`). Tool args from MCP/user code are not guaranteed acyclic.
- **MAJOR**: Cached results are returned by reference, so mutations can leak across dedup hits within the same turn.
- **MINOR/SUGGESTION**: `logger.warn` on every dedup hit will be noisy; prefer `logger.debug` or guard with `logger.shouldLog("debug")`.
- **MINOR/SUGGESTION**: `DedupExecuteMap` is also applied to the Vertex *Anthropic* native paths, which expands scope beyond the PR description's Gemini-only claim. Either narrow the change or update the description.
- **MINOR/SUGGESTION**: The new `vitest` unit test should ideally be complemented by coverage in the project's primary `test/continuous-test-suite-*.ts` harness.

### Existing comments
Acknowledged the prior Copilot/Tara-ag review comments; I did not duplicate them, though I independently flagged the circular-reference and mutation-leak risks as additional concerns.

### Next steps
Please address the two MAJOR issues (circular-reference safety and result immutability). The scope/description mismatch and log level are suggested improvements. Once the blocking issues are resolved, this should be good to approve.

src/lib/providers/googleNativeGemini3.ts
Tara-ag, line 58: 🔒 **CRITICAL**: `stableStringifyForDedup` will throw `TypeError: Converting circular structure to JSON` if any tool argument contains a circular reference. Tool args can come from MCP servers or user code and are not guaranteed to be acyclic. A single circular argument crashes the entire request.

**Suggested fix** — use a safe replacer that handles cycles:
```typescript
function stableStringifyForDedup(value: unknown): string {
  const seen = new WeakSet<object>();
  return JSON.stringify(value, (_key, val) => {
    if (val && typeof val === "object") {
      if (seen.has(val)) return "[Circular]";
      seen.add(val);
      if (!Array.isArray(val)) {
        return Object.keys(val as Record<string, unknown>)
          .sort()
          .reduce<Record<string, unknown>>((acc, key) => {
            acc[key] = (val as Record<string, unknown>)[key];
            return acc;
          }, {});
      }
    }
    return val;
  });
}
```

src/lib/providers/googleNativeGemini3.ts
Tara-ag, line 96: ⚠️ **MAJOR**: Returning the cached result directly without cloning can leak mutable state across calls within the same turn. If the tool returns an object and the caller (or a later step) mutates it, the cache will serve the mutated object on the next dedup hit.

**Risk**: Subsequent identical calls receive corrupted/altered data; hard-to-debug state pollution across a single turn.

**Suggested fix** — store a deep clone of the result, or at minimum freeze it before caching:
```typescript
const result = await execute(args, options);
resultCache.set(key, structuredClone(result));
return resultCache.get(key);
```
If `structuredClone` is unavailable for the result shape, document that tool results must be treated as immutable.

src/lib/providers/googleNativeGemini3.ts
Tara-ag, line 103: 💡 **SUGGESTION**: Using `logger.warn` for every dedup hit will be very noisy in production — this is expected behavior for Gemini. Consider downgrading to `logger.debug` or guarding with `logger.shouldLog("debug")` so normal operation doesn't flood warning logs.

src/lib/providers/googleNativeGemini3.ts
```diff
@@ -55,6 +55,62 @@ import {
...
+ * result of each {tool name + args} executed within the turn and returns it
+ * for any identical re-request instead of running the tool again.
+ *
+ * Providers build a fresh executeMap per request, so the cache scope is
+ * exactly one turn. Only `.get()` is overridden (the sole access path used by
+ * the native agentic loops), so iteration still yields the raw executors.
+ */
+export class DedupExecuteMap extends Map<string, Tool["execute"]> {
+  private readonly resultCache = new Map<string, unknown>();
+
```
Tara-ag, line 88: ⚠️ **MAJOR**: The type casting here is fragile. The `wrapped` function is declared as `ToolExecuteFunction<unknown, unknown>` but then cast to `Tool["execute"]`. These types may have different signatures from the `ai` SDK. The `as` cast bypasses TypeScript's type checking.

**Risk**: If `Tool["execute"]` has a different signature than `(args, options) => Promise<unknown>`, this could cause runtime errors when the wrapped function is called, incorrect argument passing, or return type mismatches.

**Suggested fix** — remove the unnecessary cast and rely on proper type inference:
```typescript
override get(name: string): Tool["execute"] | undefined {
  const execute = super.get(name);
  if (!execute) {
    return execute;
  }
  const resultCache = this.resultCache;
  return async (args, options) => {
    const key = `${name}::${stableStringifyForDedup(args)}`;
    if (resultCache.has(key)) {
      logger.debug(/* ... */);
      return resultCache.get(key);
    }
    const result = await execute(args, options);
    resultCache.set(key, result);
    return result;
  };
}
```

src/lib/providers/googleVertex.ts
Tara-ag, line 4045: 💡 **SUGGESTION**: This Vertex *Anthropic* native path now uses `DedupExecuteMap`. The PR description states the change targets the Gemini path only (BZ-3327 was confirmed in Gemini). If the intent is still Gemini-only, this should remain a plain `Map` to avoid unnecessary complexity in the Anthropic path where the duplicate tool call issue hasn't been confirmed. If the broader scope is intentional, please update the PR description to reflect that the deduplication now applies to both Gemini AND Anthropic paths on Vertex.

src/lib/providers/googleVertex.ts
Tara-ag, line 5699: 💡 **SUGGESTION**: Same as above — this Vertex *Anthropic* native generate path now uses `DedupExecuteMap` too, which expands the change beyond the Gemini-only scope stated in the PR description. Please either narrow the change here or update the description to match the actual behavior.

test/providers/dedupExecuteMap.test.ts
```diff
@@ -0,0 +1,87 @@
...
```
Tara-ag, line 1: 💡 **SUGGESTION**: The test file uses `vitest` (`describe`/`it`/`vi`), but per `CONTRIBUTING.md` the project's primary test harness is the `test/continuous-test-suite-*.ts` orchestrators running via `tsx`. While there are a few existing `*.test.ts` files, new provider behavior should ideally also be covered in the matching continuous suite (e.g. `test/continuous-test-suite-google-native.ts` or `test/continuous-test-suite-tool-dedup.ts`) so it runs in the standard CI pipeline. Consider adding a deterministic integration test there as well.

Tara-ag (changes requested):
## Review Summary

Reviewed 4 changed files in PR #1076 (fix: dedupe identical Gemini tool calls within a turn).

### New issues raised (this run)

| Severity | Count | Files |
|----------|-------|-------|
| 🔒 CRITICAL | 1 | `googleNativeGemini3.ts` — circular-reference crash in dedup key serialization |
| ⚠️ MAJOR | 2 | `googleNativeGemini3.ts` — mutable cached-result leakage; fragile `as` cast on tool execute |
| 💡 MINOR/SUGGESTION | 4 | `googleNativeGemini3.ts` (warn log noise), `googleVertex.ts` ×2 (Anthropic scope creep), `dedupExecuteMap.test.ts` (test harness alignment) |

### Blocking rationale

1. **CRITICAL — `stableStringifyForDedup` crashes on circular tool arguments.** Tool args can come from MCP servers or user code and are not guaranteed to be acyclic. A single circular argument will throw `TypeError: Converting circular structure to JSON` and abort the entire request. This is a runtime crash in the production tool-execution path.

2. **MAJOR — Cached results are returned by reference.** If a caller or later step mutates the returned object, the cache serves the mutated object on the next dedup hit, causing state pollution across a single turn.

3. **MAJOR — `wrapped as Tool["execute"]` is a fragile cast** that bypasses type checking between `ToolExecuteFunction<unknown, unknown>` and the AI SDK's `Tool["execute"]` signature.

These meet the blocking criteria (CRITICAL security/stability vulnerability + multiple MAJOR issues). The existing review comments raised the same circular-reference and cached-result concerns but were left unresolved; I am elevating them to blocking because they represent real runtime risks in the tool-execution path.

### Next steps

- Add cycle-safe serialization to `stableStringifyForDedup` (e.g. WeakSet replacer returning `"[Circular]"`).
- Clone results before caching (`structuredClone` or equivalent) and/or document immutability.
- Remove the unnecessary `as Tool["execute"]` cast and rely on inference.
- Decide whether Vertex Anthropic paths should also use `DedupExecuteMap`; if yes, update the PR description. If no, revert those two call sites to plain `Map`.
- Consider adding coverage in the `test/continuous-test-suite-*.ts` harness in addition to the vitest unit file.

Once the CRITICAL and MAJOR items are addressed, this should be good to merge.

PR #1087, opened 2026-06-13
fix(providers): detect image MIME from magic bytes on native Vertex path (PNG no longer mislabeled as JPEG)

murdore:
## Problem

On the native Vertex + Anthropic path (`GoogleVertexProvider`), the image block builder defaulted **every Buffer / bare-base64 image to `image/jpeg`** and never sniffed the real format. Callers that pass image **Buffers** — e.g. Slack and REST uploads in downstream apps — therefore sent PNG/GIF/WebP **mislabeled as JPEG**. Anthropic rejects the mismatch:

```
400 invalid_request_error: messages.0.content.0.image.source.base64:
The image was specified using the image/jpeg media type, but the image appears to be a image/png image
```

The image then silently never reaches the model (the caller's fallback drops it), so the user sees "please upload an image" even though they did. PNG is the common case (screenshots), so this affects most real image uploads on vertex+claude.

`detectImageType()` already existed and correctly sniffs magic bytes, but it was only used on the URL/data-URL branches — not the Buffer or bare-base64 branches.

## Fix

Sniff the format from magic bytes in the **bare-base64** and **Buffer** branches of all four native image builders (`executeNativeAnthropicStream`/`Generate`, `executeNativeGemini3Stream`/`Generate`). The detection logic is extracted to a standalone, unit-testable `detectImageMimeType()` helper that the private `detectImageType()` now delegates to.

## Verification

- New deterministic suite `test/continuous-test-suite-vertex-image-mime.ts` (7/7 pass): PNG/JPEG/WebP/GIF detection, the PNG-not-JPEG regression, and unknown/short-buffer fallback.
- Verified end-to-end against a downstream app on `vertex` + `claude-*`: a PNG Buffer that previously 400'd is now read correctly by the model; JPEG unaffected.

## Scope

Source-only (dist is gitignored). No behavior change for JPEG or for path/data-URL/HTTP image inputs; only Buffer and bare-base64 inputs are now correctly typed.

## Release Notes

* **Bug Fixes**
  * Improved multimodal image handling by automatically detecting the correct image format instead of defaulting to a single format, preventing media-type mismatches that could cause image rejection.

* **Tests**
  * Added comprehensive test coverage for image format detection across multiple formats.

### DIFF
--- a/src/lib/providers/googleVertex.ts
+++ b/src/lib/providers/googleVertex.ts
@@ -55,6 +55,7 @@ import {
   RESTRICTED_OUTPUT_TOKEN_LIMIT,
   toVertexAnthropicModelId,
 } from "../utils/modelDetection.js";
+import { detectImageMimeType } from "../utils/imageDetection.js";
 import { resolveClaudeMaxTokens } from "../utils/tokenLimits.js";
 import {
   validateApiKey,
@@ -1279,9 +1280,16 @@ export class GoogleVertexProvider extends BaseProvider {
           } else {
             // Assume base64 string
             imageBuffer = Buffer.from(image, "base64");
+            // Sniff the real format from magic bytes — bare base64 carries no
+            // mime hint, and leaving the image/jpeg default makes Anthropic
+            // reject PNG/GIF/WebP with a media-type mismatch 400.
+            mimeType = this.detectImageType(imageBuffer);
           }
         } else {
           imageBuffer = image;
+          // Buffer input (e.g. Slack/REST uploads) carries no mime hint; sniff
+          // it instead of defaulting to image/jpeg (mislabels PNG -> 400).
+          mimeType = this.detectImageType(imageBuffer);
         }
 
         const base64Data = imageBuffer.toString("base64");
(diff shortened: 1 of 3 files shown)


Tara-ag (approved):
## Review Summary

**Files reviewed:** 3
**New issues found:** 0

### Analysis

This PR correctly fixes the image MIME type detection bug on the native Vertex + Anthropic path:

1. **Root cause**: Buffer and bare-base64 image inputs were defaulting to `image/jpeg` without sniffing the actual format, causing Anthropic to reject PNG/GIF/WebP images with a 400 media-type mismatch error.

2. **Fix approach**: 
   - Extracts detection logic to a standalone `detectImageMimeType()` utility in `src/lib/utils/imageDetection.ts`
   - Adds MIME type detection calls in all 4 native image builders (`executeNativeAnthropicStream`/`Generate`, `executeNativeGemini3Stream`/`Generate`) for Buffer and bare-base64 branches
   - Refactors the existing `detectImageType()` method to delegate to the new shared utility

3. **Testing**: New deterministic test suite covers PNG/JPEG/WebP/GIF detection, the specific PNG-not-JPEG regression, and fallback behavior for unknown/short buffers.

### Verification against project standards

- ✅ **CLAUDE.md Rule 5 (Backward compatibility)**: No breaking changes to public SDK API
- ✅ **Security**: No hardcoded secrets or credential exposure
- ✅ **Architecture**: Clean extraction to shared utility; no circular dependencies introduced
- ✅ **Type safety**: Properly typed function signature
- ✅ **Testing**: Comprehensive test coverage included

### Notes (non-blocking)

- `googleAiStudio.ts` has a similar `detectImageType()` method that could potentially be refactored to use the shared utility in a future PR, but this is out of scope for this bugfix.
- The fix is minimal and targeted, only affecting Buffer/bare-base64 inputs as intended.

**Approved for merge.**

PR #1092, opened 2026-06-13
fix(vertex): sanitize Gemini responseSchema (strip errorMessage) to stop structured-output 400s

murdore:
## Problem

Vertex Gemini's `response_schema` (structured output) validator implements OpenAPI 3.0 strictly and rejects extension keywords. `convertZodToJsonSchema` enables zod-to-json-schema's `errorMessages: true`, so **any** schema field with a custom message — e.g. `z.string().regex(re, { message })` — emits an `errorMessage` keyword.

The **tool** path already sanitized schemas via `stripAdditionalPropertiesDeep`, but the **responseSchema** build assigned the converted schema directly, and `errorMessage` wasn't in the strip list. So a caller whose schema carries a custom-message field (e.g. curator's `{summary, attachment}` envelope — `attachment.filename/extension/mimetype` use `.regex(.., { message })`) gets, on **every** structured-output request:

```
400 INVALID_ARGUMENT: Unknown name "errorMessage" at
'generation_config.response_schema.properties[1].value.properties[0..2].value': Cannot find field
```

…and the recovery loop then retries the **same poisoned payload** across providers ("All providers failed").

This was found while validating curator against a Gemini model: the primary tool call (Anthropic format) succeeded, but the structured-output **recovery** generate 400'd with exactly this error. The Anthropic/Claude path is unaffected (no native `responseSchema`; Anthropic tolerates the schema), which is why it only surfaces on Gemini.

## Fix

- Add `errorMessage` to `FIELDS_TO_STRIP` in `stripAdditionalPropertiesDeep`.
- Call `stripAdditionalPropertiesDeep` on the `responseSchema` at **both** build sites (stream + generate), matching the tool path.
- Export the helper and add a deterministic suite (`test/continuous-test-suite-vertex-response-schema-sanitize.ts`) that reproduces the exact Vertex-rejected payload shape and asserts the keyword is stripped recursively while real content (types/patterns/enums/required) is preserved.

## Test

```
npx tsx test/continuous-test-suite-vertex-response-schema-sanitize.ts
# Passed: 5  Total: 5  RESULT: PASS
```
`tsc --noEmit` clean; eslint clean.

> Note: this repo's own zod is v4 (which doesn't emit `errorMessage` for these constructs); curator authors its schema with zod v3, which does. The test therefore asserts against a faithful literal of the production payload rather than the live converter, so it stays portable.

## Summary by CodeRabbit

* **Bug Fixes**
[...]

PR #1088, opened 2026-06-13
fix(structured): recover most-complete object + unwrap string-literal output in coerceJsonToSchema

murdore:
## Problem

`coerceJsonToSchema()` is NeuroLink's fallback that recovers a schema-valid object from imperfect model text when AI-SDK `experimental_output` yields nothing. It already scans balanced spans + jsonrepairs, but two robustness gaps forced **every downstream consumer to hand-roll its own extractor** on top:

1. **First-match vs. most-complete.** It broke on the *first* schema-valid candidate. With a nullable-field schema (e.g. `{ summary, attachment: z.object(...).nullable() }`), a lean preamble `{"summary":"working…","attachment":null}` validates *alongside* the real `{"summary":"done","attachment":{…}}`. First-match returned the preamble and **discarded the payload** — the classic "model streamed a short preamble then the real answer, and the file never got delivered" symptom (seen in production traces).
2. **No JSON-string-literal unwrap.** Some providers double-encode and return the object as a JSON *string* (`"{\"k\":1}"`); the balanced scan over the escaped text can't recover it.

## Fix (additive, low-risk)

In `src/lib/utils/json/coerce.ts`:
- **Most-complete selection.** Instead of breaking on the first schema-valid candidate, collect all schema-valid candidates and pick the one whose serialized form carries the most content. No behavior change when 0 or 1 candidate validates, or when no schema is supplied (first parseable object still wins).
- **String-literal unwrap.** If the whole text is a JSON string literal, unwrap one layer and add the inner text's balanced spans as candidates.

## Why this belongs here

This is generic structured-output robustness — the same logic was duplicated in at least one downstream app (Tara/curator). Centralizing it means every NeuroLink consumer gets correct multi-candidate recovery and the app-side extractor can be deleted.

## Verification

- New deterministic suite `test/continuous-test-suite-structured-coerce.ts` (7/7): preamble-vs-payload both orders, fenced-in-prose, string-literal wrapper, clean object, no-schema first-wins, and no-object→null.
- `pnpm build` clean (lint + typecheck pass).

## Scope

Single file + test. Behavior is unchanged for the common single-candidate path; only the ambiguous multi-valid-candidate and double-encoded cases change (both strictly toward recovering the intended object).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JSON extraction to handle double-encoded JSON strings
  * Enhanced JSON candidate selection to prioritize more complete results when multiple valid options exist

* **Tests**
[...]

PR #1095, opened 2026-06-13
fix(coerce): unwrap nested stringified JSON objects to satisfy schema

murdore:
## Problem

Models sometimes double-encode a **nested** field as a JSON *string* — `{"attachment":"{...}"}` instead of `{"attachment":{...}}` — which fails schema validation even though the intended object is right there.

Observed in production (Tara's `{summary, attachment}` envelope): a structured response came back with `attachment` as a JSON string, so the declared file type (`pdf`) was lost and the content degraded to a markdown/text fallback. `coerceJsonToSchema` already unwraps a stringified **top-level** object (the JSON-string-literal case), but not a stringified **nested field**.

## Fix

When a parsed candidate fails `schema.safeParse`, run a depth-bounded `deepUnwrapJsonStrings` pass that parses any string-valued field whose content is itself a JSON object/array, then **re-validate**:

- **The `safeParse` gate is the safety net** — an over-eager unwrap (a field that should stay a string) fails re-validation and is rejected, falling back to prior behaviour.
- A parsed string is **not re-descended into**, so an attachment's own string `content` (even if it looks like JSON) is preserved verbatim.
- Only runs when a schema is supplied (no schema → unchanged first-parseable-wins behaviour); recursion is depth-bounded.

This generalizes to **any** double-encoded nested field, for every NeuroLink consumer — not just curator.

## Test

```
npx tsx test/continuous-test-suite-coerce-nested-unwrap.ts   # 6/6 PASS
npx tsx test/continuous-test-suite-structured-coerce.ts      # 7/7 PASS (no regression)
```
Covers: the production attachment-as-string case, content preservation (no re-descent), clean passthrough, null attachment, the over-unwrap gate (JSON-looking string kept as string), and the no-schema boundary. `tsc` clean; eslint clean.

Follow-up to the structured-output line of fixes (#1088 top-level unwrap, #1092 responseSchema sanitize, #1093 400-abort). Curator picks this up on the next `@juspay/neurolink` bump.

### DIFF
--- a/src/lib/utils/json/coerce.ts
+++ b/src/lib/utils/json/coerce.ts
@@ -56,6 +56,78 @@ function parseOrRepair(
   }
 }
 
+/** Bounds the recursive nested-string unwrap against pathological inputs. */
+const MAX_NESTED_UNWRAP_DEPTH = 6;
+
+/**
+ * Recursively replace any string-valued field whose content is itself a JSON
+ * object/array with the parsed value. Models sometimes double-encode a NESTED
+ * field — e.g. `{ "attachment": "{\"k\":1}" }` instead of
+ * `{ "attachment": { "k": 1 } }` — which fails schema validation even though the
+ * intended object is right there. (`coerceJsonToSchema` already unwraps a
+ * stringified TOP-LEVEL object; this handles the nested case.)
+ *
+ * A parsed string is NOT re-descended into: its own string fields (e.g. an
+ * attachment's `content`) are the model's intended values and must be left
+ * alone. Recursion only walks already-structural objects/arrays to find
+ * stringified fields anywhere in the tree. Returns a NEW value (never mutates
+ * the input) plus whether anything changed, so the caller can skip a redundant
+ * re-validation when nothing was unwrapped. Callers MUST re-validate the result
+ * against the schema — that gate is what keeps an over-eager unwrap (a field
+ * that should stay a string) from being accepted.
+ */
+function deepUnwrapJsonStrings(
+  value: unknown,
+  depth = 0,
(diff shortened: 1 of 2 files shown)


Tara-ag (approved):
## Review Summary

**Files reviewed:** 2
**New issues:** 0 (all clear)

### Analysis

This PR addresses a production issue where models double-encode nested fields as JSON strings (e.g., `{"attachment": "{...}"}` instead of `{"attachment": {...}}`). The fix is well-architected with multiple safety layers:

**`src/lib/utils/json/coerce.ts`:**
- ✅ Depth-bounded recursion (`MAX_NESTED_UNWRAP_DEPTH = 6`) prevents stack overflow on pathological inputs
- ✅ Schema re-validation gate prevents over-unwrapping fields that should remain strings
- ✅ No re-descent into parsed objects preserves intended string content (e.g., file content)
- ✅ Immutable implementation returns new values, never mutates input
- ✅ `repaired: true` correctly signals when transformation occurred

**`test/continuous-test-suite-coerce-nested-unwrap.ts`:**
- ✅ Covers the production bug case (Tara's attachment envelope)
- ✅ Validates content preservation (no re-descent)
- ✅ Tests the over-unwrap safety gate
- ✅ Includes null handling and no-schema boundary cases
- ✅ Follows established test harness patterns

### Verification Against Standards

- **CLAUDE.md Rule compliance:** No violations detected
- **Security:** No hardcoded secrets, safe JSON parsing with try/catch
- **Type safety:** Proper `unknown` handling, no `any` casts
- **Backward compatibility:** No breaking changes to public API

The implementation elegantly solves the nested double-encoding problem while maintaining defensive programming practices. Approved for merge.

PR #1099, opened 2026-06-17
fix(csv): support CRLF and CR line endings

### LINKED ISSUE #364: CSV-004: Only \n Line Endings Supported
## Summary
CSV parsing only recognizes Unix line endings (\n), causing Windows (\r\n) and Mac (\r) line endings to be treated incorrectly, breaking row parsing.

## Root Cause
The code uses `split("\n")` at line 94 without handling \r\n or \r. Windows files have \r\n treated as two separate rows, Mac files appear as single line.

## Fix
Replace `split("\n")` with `split(/\r?\n|\r/)` to handle all line ending types. Also strip \r from values if present.

## Files Affected
`src/lib/utils/csvProcessor.ts`

## Acceptance Criteria
- [ ] Unit tests added

**Effort:** 1h

wenhaoone:
## Summary
- normalize CSV line splitting across Unix (LF), Windows (CRLF), and classic Mac (CR) line endings
- apply the same line handling to raw CSV row limiting, metadata-line detection, string parsing, and file sniffing
- add regression coverage for Unix, Windows, and classic Mac CSV inputs

Closes #364

## Validation
- git diff --check
- targeted static assertions that csvProcessor no longer uses split(\n) and that the new regression tests are present

I attempted to install dependencies and run pnpm run test:bugfixes, but the local Windows environment ran out of disk space during corepack pnpm install --frozen-lockfile (ERR_PNPM_ENOSPC). I removed the partial node_modules afterward and did not claim the full test suite passed locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CSV parsing and raw CSV processing to correctly support Unix (`\n`), Windows (`\r\n`), and classic Mac (`\r`) line endings.
  * Added RFC 4180–compatible behavior so line-ending characters inside quoted fields are preserved.
  * Normalized escaping for markdown and CSV so newline variants are consistently handled without leaking carriage returns.
* **Tests**
  * Expanded continuous test coverage for parsing and raw processing across all newline variants, including quoted-field edge cases.

### DIFF
--- a/src/lib/utils/csvProcessor.ts
+++ b/src/lib/utils/csvProcessor.ts
@@ -581,6 +581,15 @@ function isMetadataLine(lines: string[]): boolean {
 }
 
 /**
+ * Split CSV text into logical lines for metadata detection and raw row limiting.
+ *
+ * Supports Unix (LF), Windows (CRLF), and classic Mac (CR) line endings.
+ */
+function splitCsvLines(csvString: string): string[] {
+  return csvString.split(/\r\n|\n|\r/);
+}
+
+/**
  * CSV processor for converting CSV data to LLM-optimized formats
  *
  * Supports three output formats:
@@ -632,10 +641,10 @@ export class CSVProcessor {
 
     const csvString = content.toString("utf-8");
 
-    // For raw format, return original CSV with row limit (no parsing needed)
-    // This preserves the exact original format which works best for LLMs
+    // For raw format, return CSV text with row limit (no parsing needed)
+    // This preserves the CSV shape while normalizing line endings for row handling.
     if (formatStyle === "raw") {
-      const lines = csvString.split("\n");
+      const lines = splitCsvLines(csvString);
       const hasMetadataLine = isMetadataLine(lines);
 
       if (hasMetadataLine) {
@@ -863,7 +872,10 @@ export class CSVProcessor {
       let buffer = "";
       lineReader.on("data", (chunk: string | Buffer) => {
         buffer += chunk.toString();
-        const lines = buffer.split("\n");
+        const splitBuffer = buffer.endsWith("\r")
+          ? buffer.slice(0, -1)
+          : buffer;
+        const lines = splitCsvLines(splitBuffer);
         if (lines.length >= 2) {
           firstLines.push(lines[0], lines[1]);
           lineReader.destroy();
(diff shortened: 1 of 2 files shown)


murdore:
**Deep code-review findings** — the core fix is genuine (the raw `formatStyle` CR-only / CRLF row-splitting bug is real and this resolves it for the common case), but a diff-level review found a new **data-corruption regression** that the tests don't cover.

### MAJOR — quote-unaware `splitCsvLines()` corrupts RFC4180 quoted fields
`splitCsvLines()` is a plain regex split (`/\r\n|\n|\r/`) with **no CSV-quote awareness**, so an embedded lone `\r` **inside a quoted field** is now treated as a row boundary.

Repro (verified by building the patched `csvProcessor.ts` from this PR's diff and A/B-testing against the baseline), input with `formatStyle: 'raw'`:
```
name,notes
Alice,"line1<CR>line2"
Bob,"ok"
```
- **Before:** rowCount = 2, the quoted value preserved byte-for-byte as `"line1\rline2"`.
- **After this PR:** rowCount = 3 (the embedded CR is wrongly counted as a new row) **and** the value is silently rewritten to `"line1\nline2"`.

So the change doesn't just "normalize file-level newlines" — it corrupts the field's original content. Bare-CR-in-quotes is legal RFC4180 (e.g. old Mac-authored cells), and none of the new regression tests use a quoted/multi-line field, so this is completely uncovered. This contradicts the issue's own "No regressions in existing CSV processing" acceptance criterion.

*Fix idea:* split rows in a quote-aware way (don't break inside `"…"`), or reuse the existing quote-aware parser for row boundaries instead of a raw regex.

### MINOR — `parseCSVFile` metadata-peek change is behaviorally inert
The CRLF/CR fix threaded into `parseCSVFile`'s two-line metadata peek doesn't actually fix any real-world case: the metadata-line-skip path is still wrong (a separate, pre-existing design bug — it skips N *data rows* after `csv-parser` has already used the metadata line as headers, rather than stripping the metadata line before parsing, unlike `process()`/`parseCSVString`). Verified identical before/after the patch with a `sep=,` metadata file. Not introduced here, but the "threaded through all four sites" framing overstates this site's contribution.

### NIT
No regression test exercises the lone-CR handling added to `escapePipe`/`escapeCSV`, so a future revert of those `/\r\n|\n|\r/` checks to LF-only wouldn't be caught.

*(Findings from an automated deep review — the raw-format fix itself is good; it's the quote-aware row splitting + a quoted-field test that need adding before merge.)*

PR #1105, opened 2026-06-20
feat(tool-routing): apply routing in generate(), decision telemetry, routing cache + stickiness

murdore:
## What

Builds on the shipped pre-call tool-routing feature (`d78d6912`) with three improvements:

- **A — generate() wiring:** tool routing now applies to `generate()` (was stream-only), via `applyToolRoutingExclusions` inside the generate path (respects `earlyResult` short-circuit, `disableTools`, and Langfuse context).
- **E — decision telemetry:** structured `ToolRoutingDecision` emitted on every resolver return path via an `emitDecision` callback; surfaced as OTel span attributes.
- **C — routing cache + session stickiness:** `ToolRoutingCache` (LRU + TTL) skips the router LLM on repeat queries; per-session stickiness keeps recently-selected servers warm for N turns to reduce flapping (off-by-one fix: sticky ids read before recordSelection).

## Key files
- `src/lib/core/toolRoutingCache.ts` (new) — LRU+TTL cache + stickiness, injectable clock
- `src/lib/core/toolRouting.ts` — `emitDecision` on all return paths
- `src/lib/neurolink.ts` — generate() wiring, cache/stickiness integration, span attributes
- `src/lib/types/toolRouting.ts` — `ToolRoutingDecision`, cache/stickiness config + types

## Verification
- `pnpm run check` ✓ · `pnpm run lint` ✓ · `pnpm run build` ✓
- vitest `test/toolRouting.test.ts`: 41/41 (incl. enabled:false / disableTools short-circuit guards)
- continuous `test:tool-routing`: 28/28 incl. live-gated

## Notes
- Fully backward compatible — routing is opt-in via `toolRouting` config.
- Base for the stacked **semantic (B/D)** PR.


src/lib/neurolink.ts
```diff
@@ -7810,6 +7946,58 @@ Current user's request: ${currentInput}`;
         return;
       }
 
+      // Apply stickiness FIRST: the sticky ids were recorded on a prior turn and
+      // represent servers that should stay warm for the current turn. Consuming
+      // (decrementing) them before recordSelection ensures the window covers the
+      // correct set of future turns rather than burning one count on the same
+      // turn the selection is recorded (off-by-one fix).
+      if (stickinessEnabled && cache && sessionId) {
```
Tara-ag, line 7974: 💡 **MINOR: Redundant filtering logic**

The stickiness filtering logic performs the same check twice:

```typescript
const stickyPrefix = new Set(stickyIds.map((id) => `${id}_`));
routedExcludeTools = routedExcludeTools.filter(
  (toolName) =>
    !stickyIds.some((id) => toolName.startsWith(`${id}_`)) &&
    !Array.from(stickyPrefix).some((prefix) =>
      toolName.startsWith(prefix),
    ),
);
```

Both conditions check if `toolName.startsWith(`${id}_`)` for any sticky id. The second check using `stickyPrefix` is redundant since it contains the exact same strings being checked in the first condition.

**Suggestion:** Simplify to a single check:
```typescript
const stickyPrefixes = stickyIds.map((id) => `${id}_`);
routedExcludeTools = routedExcludeTools.filter(
  (toolName) => !stickyPrefixes.some((prefix) => toolName.startsWith(prefix)),
);
```

src/lib/neurolink.ts
```diff
@@ -7768,15 +7789,129 @@ Current user's request: ${currentInput}`;
...
+
+      // Lazy-init the cache instance once per NeuroLink instance.
+      if (
+        (cacheEnabled || stickinessEnabled) &&
+        !this.toolRoutingCacheInstance
+      ) {
+        this.toolRoutingCacheInstance = new ToolRoutingCache({
+          ttlMs: cacheConfig?.ttlMs,
+          maxEntries: cacheConfig?.maxEntries,
+          stickyTurns: routingConfig.stickiness?.turns,
```
Tara-ag, line 7805: ⚠️ **MAJOR: Cache key collision risk with empty sessionId**

When `sessionId` is empty (which happens when `requestContext?.sessionId` is not a string), the cache key becomes `:${routingQuery}`. This means:

1. All sessions without an explicit `sessionId` in context share the same cache namespace
2. Different users' queries could collide and receive each other's tool exclusions
3. This is a potential security/privacy issue where one session's routing decisions leak to another

**Suggested fix:**
```typescript
const cacheKey = cacheEnabled && sessionId
  ? `${sessionId}:${routingQuery}`
  : cacheEnabled
    ? `anonymous-${crypto.randomUUID()}:${routingQuery}`  // or skip caching
    : undefined;
```

Alternatively, disable caching when `sessionId` is not available to prevent cross-session pollution.

src/lib/neurolink.ts
```diff
@@ -7768,15 +7789,129 @@ Current user's request: ${currentInput}`;
...
+      // fetchRecentRoutingHistory.
+      const requestContext = options.context as
+        | Record<string, unknown>
+        | undefined;
+      const sessionId =
+        typeof requestContext?.sessionId === "string"
+          ? requestContext.sessionId
+          : "";
+
+      // Cache key: session + normalized routing query.
```
Tara-ag, line 7820: 💡 **MINOR: Inconsistent telemetry on cache hit vs miss**

On cache hit (lines 7820-7836), only 3 span attributes are set:
- `tool_routing.outcome`
- `tool_routing.routable_server_count`
- `tool_routing.cache_hit`

But on cache miss (lines 7853-7882), 8 attributes are set including detailed fields like `selected_server_ids`, `excluded_server_ids`, `hallucinated_ids`, etc.

**Suggestion:** Consider adding `selected_server_ids` and `excluded_server_ids` to the cache hit path for observability consistency. These can be retrieved from `cached.selectedServerIds` and derived from the cached exclusions.

src/lib/core/toolRoutingCache.ts
```diff
@@ -0,0 +1,147 @@
...
+      this.evictLRU();
+    }
+    this.store.set(key, {
+      excludedToolNames: value.excludedToolNames,
+      selectedServerIds: value.selectedServerIds,
+      expiresAt: this.now() + this.ttlMs,
+      accessOrder: ++this.accessCounter,
+    });
+  }
+
```
Tara-ag, line 108: 💡 **MINOR: LRU eviction is O(n) - consider performance at scale**

The `evictLRU()` method iterates through all entries to find the least recently used:

```typescript
private evictLRU(): void {
  let lruKey: string | undefined;
  let lruOrder = Infinity;
  for (const [key, entry] of this.store) {
    if (entry.accessOrder < lruOrder) {
      lruOrder = entry.accessOrder;
      lruKey = key;
    }
  }
  if (lruKey !== undefined) {
    this.store.delete(lruKey);
  }
}
```

With the default `maxEntries: 256`, this is acceptable. However, if users configure larger cache sizes (e.g., 10,000+ entries), this becomes O(n) on every insert when at capacity.

**Suggestion:** Document this trade-off in the config comments, or consider using a doubly-linked list approach for O(1) eviction if larger caches are anticipated. Given the default of 256, this is minor.

src/lib/core/toolRoutingCache.ts
```diff
@@ -0,0 +1,147 @@
...
+  /**
+   * Returns the server ids that should be kept warm (not excluded) for the
+   * given session due to stickiness. Decrements the turn counter; when it
+   * reaches zero the entry is removed. Returns an empty array when there is no
+   * active sticky state for the session.
+   */
+  getStickyServerIds(sessionId: string): string[] {
+    if (!sessionId) {
+      return [];
+    }
```
Tara-ag, line 141: ⚠️ **MAJOR: Potential memory leak with unbounded sticky map**

The `sticky` Map grows indefinitely without any eviction mechanism:

```typescript
private readonly sticky = new Map<string, ToolRoutingStickiness>();
```

Unlike the main `store` which has LRU eviction, the sticky map:
1. Has no size limit
2. Only removes entries when `turnsRemaining` reaches 0
3. Entries with high `stickyTurns` values persist for many calls

In a long-running service with many unique session IDs, this could cause unbounded memory growth.

**Suggested fix:** Add a size limit and LRU eviction to the sticky map, similar to the main cache:
```typescript
if (this.sticky.size >= this.maxStickyEntries) {
  // Evict oldest sticky entry
}
```

Or document that `stickyTurns` should be kept small and sessions should be reused.

src/lib/types/toolRouting.ts
```diff
@@ -78,6 +102,76 @@ export type ToolRoutingCatalogEntry = {
   toolNames: string[];
```
Tara-ag, line 102: 💡 **SUGGESTION: Consider adding validation constraints to config types**

The config types allow potentially problematic values:

```typescript
cache?: {
  enabled?: boolean;
  ttlMs?: number;        // Could be 0 or negative
  maxEntries?: number;   // Could be 0 or negative
};
stickiness?: {
  enabled?: boolean;
  turns?: number;        // Could be 0 or negative
};
```

Zero or negative values for `ttlMs`, `maxEntries`, or `turns` could cause unexpected behavior (immediate expiry, no caching, or no stickiness).

**Suggestion:** Consider adding runtime validation in `ToolRoutingCache.constructor()` to reject invalid configurations, or document the behavior of edge values.

src/lib/core/toolRouting.ts
```diff
@@ -260,7 +294,30 @@ export async function resolveToolRoutingExclusions(
     );
 
     const rawText = generateResult?.content ?? "";
-    const parsed = routerOutputSchema.safeParse(parseRouterJson(rawText));
+    let parsed: ReturnType<typeof routerOutputSchema.safeParse>;
```
Tara-ag: 💡 **SUGGESTION: Duplicate error handling for parse failures**

Lines 297-318 and 328-338 have nearly identical error handling for parse failures:
- Both log the same warning message
- Both emit the same `failed-open-parse` decision
- Both return `[]`

The only difference is that the first catches exceptions from `parseRouterJson` while the second handles Zod validation failures.

**Suggestion:** Consider extracting a helper function to reduce duplication:
```typescript
function handleParseFailure(
  emitDecision: EmitDecisionFn | undefined,
  routableServers: ToolRoutingCatalogEntry[],
  routingStartTime: number,
  rawText: string,
): [] {
  // shared implementation
}
```

This is minor since it's only two call sites, but improves maintainability.

test/continuous-test-suite-tool-routing.ts
```diff
@@ -0,0 +1,696 @@
...
+      },
+      shipping_track: {
+        name: "shipping_track",
+        description: "Track shipment",
+        execute: noopExecute,
+      },
+      utility_echo: {
+        name: "utility_echo",
+        description: "Echo utility",
+        execute: noopExecute,
```
Tara-ag, line 650: 💡 **MINOR: Unused variable in live test**

```typescript
let routingDecision: ToolRoutingDecision | undefined;
// Observe the decision via a spy on applyToolRoutingExclusions indirectly
// We can't easily observe options.excludeTools from outside, but we can
// verify that generate() itself doesn't throw when routing is enabled.
```

The `routingDecision` variable is declared but never assigned or used. The comment suggests the intention was to observe routing decisions, but the implementation doesn't actually do this.

**Suggestion:** Either:
1. Remove the unused variable
2. Implement the observation mechanism described in the comment (e.g., by spying on the internal method)

Tara-ag (commented):
## Code Review Summary

**Files reviewed:** 8  
**New issues raised:** 7 (0 Critical, 2 Major, 5 Minor/Suggestions)

### Overview
This PR implements three valuable enhancements to the pre-call tool routing feature:
- **Item A:** Tool routing now applies to `generate()` (was stream-only)
- **Item E:** Structured decision telemetry via `emitDecision` callback
- **Item C:** LRU+TTL cache with session stickiness for routing decisions

The implementation is generally well-structured with good test coverage (41 vitest tests + 28 continuous tests). The fail-open design principles are appropriately applied throughout.

### Issues Raised

#### ⚠️ MAJOR (2)

1. **Cache key collision risk with empty sessionId** (`src/lib/neurolink.ts:7805`)
   - When `sessionId` is empty, cache keys become `:${routingQuery}`, causing cross-session cache pollution
   - Different sessions without explicit sessionIds could receive each other's tool exclusions
   - **Recommendation:** Disable caching when `sessionId` is unavailable, or use a per-request unique identifier

2. **Unbounded memory growth in sticky map** (`src/lib/core/toolRoutingCache.ts:115`)
   - The `sticky` Map has no size limit or eviction mechanism
   - In long-running services with many unique sessions, this could cause memory exhaustion
   - **Recommendation:** Add LRU eviction or size limits to the sticky map

#### 💡 MINOR/SUGGESTIONS (5)

3. **Redundant filtering logic** (`src/lib/neurolink.ts:7954`) - The stickiness filter checks the same condition twice; can be simplified

4. **Inconsistent telemetry on cache hit vs miss** (`src/lib/neurolink.ts:7820`) - Cache hit path sets fewer span attributes than the miss path

5. **O(n) LRU eviction** (`src/lib/core/toolRoutingCache.ts:90`) - Acceptable at default size (256), but consider documenting for larger configs

6. **Missing config validation** (`src/lib/types/toolRouting.ts:102`) - Zero/negative values for `ttlMs`, `maxEntries`, `turns` aren't validated

7. **Unused variable in test** (`test/continuous-test-suite-tool-routing.ts:650`) - `routingDecision` declared but never used

### Positive Observations

✅ **Fail-open design** - All error paths gracefully degrade to empty exclusions  
✅ **Telemetry safety** - `safeEmitDecision` correctly swallows errors to prevent telemetry from affecting routing  
✅ **Test coverage** - Comprehensive tests including edge cases (off-by-one stickiness, cache expiry, etc.)  
✅ **Backward compatibility** - Routing remains opt-in via `toolRouting` config  
✅ **Clean separation** - Cache logic well-isolated in `ToolRoutingCache` class  

### Recommendation

**Approve with suggestions.** The two MAJOR issues should be addressed in a follow-up PR:
1. Fix the cache key collision for anonymous sessions
2. Add bounds to the sticky map to prevent unbounded growth

Neither issue is a security vulnerability or breaking change, but both could cause operational problems in production environments with high session churn.

src/lib/neurolink.ts
```diff
@@ -7768,40 +7789,171 @@ Current user's request: ${currentInput}`;
...
```
Tara-ag, line 7820: 💡 **MINOR: Telemetry inconsistency between cache hit and miss paths**

The cache hit path (lines 7820-7850) sets fewer span attributes than the cache miss path:
- **Cache hit**: 5 attributes (outcome, routable_server_count, cache_hit, selected_server_ids, excluded_tool_count)
- **Cache miss**: 8 attributes via `emitDecision` (adds excluded_server_ids, hallucinated_ids, duration_ms)

**Impact**: Observability dashboards will show gaps for `tool_routing.excluded_server_ids`, `tool_routing.hallucinated_ids`, and `tool_routing.duration_ms` on cache hits, making it harder to compare routing decisions across paths.

**Suggestion**: Consider adding the missing attributes to the cache hit path for parity:
```typescript
// From cached result:
activeSpan.setAttribute(
  "tool_routing.excluded_server_ids",
  spanJsonAttribute(cached.selectedServerIds.length === 0 
    ? routableServerIds 
    : routableServerIds.filter(id => !cached.selectedServerIds.includes(id)))
);
activeSpan.setAttribute("tool_routing.duration_ms", 0); // or omit
```

This is minor since the `cache_hit: true` flag allows consumers to filter, but full parity reduces surprise.

Tara-ag (commented):
## Review Summary

**Files reviewed**: 8 files (+1891/-49 lines)

### New Issues Raised (1)
| 💡 MINOR | 1 | Telemetry inconsistency between cache hit/miss paths |

### Existing Comments Acknowledged (16 total from Tara-ag and copilot-pull-request-reviewer)
Most issues have been addressed in the current revision:
- ✅ **MAJOR**: Cache key collision with empty sessionId → Fixed (cacheKey is undefined when sessionId is empty)
- ✅ **MAJOR**: Unbounded sticky map memory leak → Fixed (MAX_STICKY_ENTRIES = 1024 added)
- ✅ **MINOR**: Defensive copying in cache → Fixed (shallow copies in get() and set())
- ✅ **MINOR**: Redundant filtering logic → Already simplified to single prefix check

### Architecture Assessment

**Good practices observed:**
- Fail-open design throughout (routing failures don't break the turn)
- Defensive copying prevents cache corruption
- Bounded memory usage with LRU eviction and sticky entry caps
- Telemetry errors are swallowed to prevent affecting routing
- Proper separation of concerns between cache and routing logic

**Areas for follow-up (non-blocking):**
1. **Telemetry parity**: Cache hit path sets fewer span attributes than miss path (see inline comment)
2. **O(n) LRU eviction**: Acceptable at default 256 entries, but documented trade-off

### Tests
- 41/41 vitest tests passing
- 28/28 continuous tests passing
- Good coverage for cache hit/miss, stickiness, and fail-open paths

### Verdict
The PR implements the three features (generate() wiring, decision telemetry, routing cache + stickiness) correctly with proper fail-open behavior. The major concerns from earlier review rounds have been addressed. Minor telemetry inconsistency noted but not blocking.

**Recommendation**: Approve with minor suggestions.

src/lib/neurolink.ts
murdore: Fixed in `12073fa5`: precomputed `stickyPrefixes` once and collapsed the predicate to a single `.some()` (removed the redundant `Array.from(stickyPrefix)` rebuild per tool).

src/lib/neurolink.ts
murdore: Fixed in `12073fa5`: `cacheKey` is now `undefined` unless a non-empty `sessionId` is present (`cacheEnabled && sessionId ? ... : undefined`), so anonymous sessions never share a `:query` namespace. Added a regression test (`does NOT cache when sessionId is absent — router runs on every call`).

src/lib/neurolink.ts
murdore: Fixed in `12073fa5`: the cache-hit path now also sets `tool_routing.selected_server_ids` and `tool_routing.excluded_tool_count` span attributes for parity with the miss path.

src/lib/core/toolRoutingCache.ts
murdore: Addressed in `12073fa5`: added a doc comment on `evictLRU()` noting the O(n) scan is acceptable at the default `maxEntries` (256) and that very large caches should consider an O(1) doubly-linked-list approach.

src/lib/core/toolRoutingCache.ts
murdore: Fixed in `12073fa5`: the `sticky` map now has an internal `MAX_STICKY_ENTRIES` (1024) cap and evicts the oldest entry (Map insertion order) in `recordSelection()` when at capacity.

src/lib/types/toolRouting.ts
murdore: Addressed in `12073fa5`: `ToolRoutingCache` now validates `ttlMs`/`maxEntries`/`stickyTurns` in the constructor and falls back to the module defaults for non-positive or NaN values.

src/lib/core/toolRouting.ts
murdore: Fixed in `12073fa5`: extracted a shared `failOpenParse` helper used by both the JSON-parse-exception path and the Zod schema-validation path (and a `safeEmitDecision` wrapper), removing the duplication.

test/continuous-test-suite-tool-routing.ts
murdore: Fixed in `12073fa5`: removed the unused `routingDecision` variable and tightened the surrounding comment.

src/lib/neurolink.ts
```diff
@@ -7768,15 +7789,129 @@ Current user's request: ${currentInput}`;
...
+      const cacheKey = cacheEnabled
+        ? `${sessionId}:${routingQuery}`
+        : undefined;
```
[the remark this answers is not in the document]
murdore: Fixed in `12073fa5`: caching is now gated on a non-empty `sessionId`, so anonymous callers never share a cache namespace. Added a regression test asserting two no-sessionId calls do not share a cache entry.

src/lib/neurolink.ts
```diff
@@ -7768,15 +7789,129 @@ Current user's request: ${currentInput}`;
...
+          logger.debug("[ToolRouting] Cache hit, skipping router LLM", {
+            cacheKey,
+          });
+          if (cached.excludedToolNames.length > 0) {
+            options.excludeTools = [
+              ...(options.excludeTools ?? []),
+              ...cached.excludedToolNames,
+            ];
+          }
+          return;
```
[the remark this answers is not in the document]
murdore: Fixed in `12073fa5`: the cache-hit and cache-miss paths were unified so stickiness is applied — and the turn counter decremented — on **both** paths before exclusions are applied. The cache hit no longer returns early ahead of stickiness.

src/lib/neurolink.ts
```diff
@@ -7810,6 +7946,58 @@ Current user's request: ${currentInput}`;
...
+      }
+
+      // --- ITEM C: store result for future turns ---
+      if (resolvedDecision?.outcome === "applied" && cache) {
+        // Store in cache if enabled.
+        if (cacheEnabled && cacheKey !== undefined) {
+          try {
+            cache.set(cacheKey, {
+              excludedToolNames: routedExcludeTools,
+              selectedServerIds: resolvedDecision.selectedServerIds,
```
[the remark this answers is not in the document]
murdore: Fixed in `12073fa5`: the cache now stores the **raw, pre-stickiness** exclusions (`rawExclusions`); stickiness is re-applied at read time on every path, so an expired stickiness window is no longer baked into the cached value.

src/lib/core/toolRoutingCache.ts
```diff
@@ -0,0 +1,147 @@
...
+  ): void {
+    if (this.store.size >= this.maxEntries && !this.store.has(key)) {
+      this.evictLRU();
+    }
+    this.store.set(key, {
+      excludedToolNames: value.excludedToolNames,
+      selectedServerIds: value.selectedServerIds,
+      expiresAt: this.now() + this.ttlMs,
+      accessOrder: ++this.accessCounter,
+    });
```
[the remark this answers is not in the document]
murdore: Fixed in `12073fa5`: `set()` now defensive-copies both arrays on write (`[...value.excludedToolNames]`, `[...value.selectedServerIds]`).

src/lib/core/toolRoutingCache.ts
```diff
@@ -0,0 +1,147 @@
...
+      excludedToolNames: entry.excludedToolNames,
+      selectedServerIds: entry.selectedServerIds,
+    };
+  }
+
+  /**
+   * Stores a routing result under the given key. Evicts the least-recently-
+   * used entry when the store is at capacity. Silently no-ops on any error
+   * (caller is responsible for fail-open behaviour).
+   */
```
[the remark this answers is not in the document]
murdore: Fixed in `12073fa5`: `set()` body is now wrapped in a try/catch that silently no-ops, honoring the documented fail-open behavior.

src/lib/neurolink.ts
```diff
@@ -7810,6 +7946,58 @@ Current user's request: ${currentInput}`;
...
+          if (stickyIds.length > 0) {
+            // Remove from routedExcludeTools any tool belonging to a sticky server.
+            const stickyPrefix = new Set(stickyIds.map((id) => `${id}_`));
+            routedExcludeTools = routedExcludeTools.filter(
+              (toolName) =>
+                !stickyIds.some((id) => toolName.startsWith(`${id}_`)) &&
+                !Array.from(stickyPrefix).some((prefix) =>
+                  toolName.startsWith(prefix),
+                ),
+            );
```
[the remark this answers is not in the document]
murdore: Fixed in `12073fa5`: `stickyPrefixes` is precomputed once and the predicate is a single `.some(prefix => toolName.startsWith(prefix))` — no per-tool `Set`/`Array.from` reallocation or duplicate check.

test/continuous-test-suite-tool-routing.ts
```diff
@@ -0,0 +1,696 @@
...
+    // checking that the generate call completes without error and that routing
+    // was triggered (if it fails open, tools are unchanged — acceptable).
+    let routingDecision: ToolRoutingDecision | undefined;
+    // Observe the decision via a spy on applyToolRoutingExclusions indirectly
+    // We can't easily observe options.excludeTools from outside, but we can
+    // verify that generate() itself doesn't throw when routing is enabled.
+    const result = await sdk.generate({
+      input: { text: "show me yesterday's sales figures" },
+      disableTools: false,
+    });
```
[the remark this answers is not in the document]
murdore: Fixed in `12073fa5`: removed the dead `routingDecision` variable and tightened the comment.

src/lib/neurolink.ts
murdore: Fixed in `12073fa5`: cache-hit telemetry now sets `selected_server_ids` and `excluded_tool_count` for parity with the miss path.

PR #1122, opened 2026-07-01
fix(file-detector): normalize extension matching

### LINKED ISSUE #389: FD-015: Fix Extension Matching Edge Cases
## Bug Summary
Extension matching uses .toLowerCase() but misses edge cases like whitespace, query parameters, hash fragments.

## Root Cause Analysis
**Why this bug exists:**
Basic lowercase conversion without handling real-world scenarios like \`data.csv?download=true\`.

**Impact:**
- Fails on URLs with query params
- Whitespace causes match failures
- Invalid extensions accepted

## Current Code (Buggy)
\`\`\`typescript
const match = url.pathname.match(/\.([^.]+)$/);
return match ? match[1] : null;  // ❌ No normalization

const type = typeMap[ext.toLowerCase()];  // ❌ Only lowercase
\`\`\`

## Fixed Code
\`\`\`typescript
private getExtension(input: string): string | null {
  let ext = /* extract from regex */;

  if (ext) {
    ext = ext.toLowerCase().trim();  // ✅ Normalize
    ext = ext.split('?')[0].split('#')[0];  // ✅ Remove params

    if (!/^[a-z0-9]{1,10}$/.test(ext)) {  // ✅ Validate
      return null;
    }
  }

  return ext;
}
\`\`\`

## Acceptance Criteria
- [ ] Trim whitespace
- [ ] Remove query params (?...)
- [ ] Remove hash fragments (#...)
- [ ] Validate alphanumeric
- [ ] Tests pass

**File:** src/lib/utils/fileDetector.ts
**Effort:** 1h
**Good First Issue:** Yes

morgan-coded:
`ExtensionStrategy` can miss or mis-detect a file extension when a path-like input has surrounding whitespace, a query string, or a fragment, and it preserves the raw extension casing. URL pathnames with encoded trailing whitespace can also fall through to `unknown`, because the pathname was decoded only after suffix handling.

This trims the input before source and extension detection, strips query/hash suffixes for path-like inputs, decodes and trims URL pathnames before capture, lowercases the stored extension, and rejects non-alphanumeric suffix captures. Dotfiles, extensionless inputs, and the existing invalid-suffix handling are unchanged. I added a focused test suite covering these cases; it and the existing bug-fix suite pass.

Fixes #389

## Summary by CodeRabbit

* **New Features**
  * Improved file extension detection for inputs with extra whitespace, URL-encoded paths, and mixed-case extensions.
  * Better handling of malformed links and invalid extension characters, with more consistent fallback behavior.

* **Tests**
  * Added a new continuous test suite covering extension normalization and edge cases.
  * Included the new test run in the standard unit test workflow.


murdore:
Thanks @morgan-coded — the extension-normalization approach here is solid and the URL cases are well covered. Two things before this can land:

1. **Rebase needed** — this now conflicts with `release` (it has moved on since the PR was opened, including changes to `package.json` and the file-detector area). Could you rebase onto the latest `release`?

2. **Completeness of the #389 fix** — the query/hash stripping currently only kicks in for `http(s)` URL inputs (via `new URL(...)`). For a **plain local path** with a query-like suffix appended — e.g. issue #389's own motivating example `data.csv?download=true` — the `?download=true` isn't stripped, so the extension still comes back empty/wrong. Could you also strip `?…`/`#…` for non-URL inputs (e.g. `ext.split('?')[0].split('#')[0]` before the alphanumeric check), and add a test for the literal `data.csv?download=true` case? That's the exact scenario in the issue's Root Cause Analysis, and right now the new test suite doesn't cover it.

Once rebased with that case handled, this is good to merge — happy to re-review.

morgan-coded:
Rebased this onto the latest `release` and added the missing non-URL path case.

`data.csv?download=true` now resolves as `csv`, while filenames with literal `?` or `#` before the final extension still keep their extension (`snapshot?old.xlsx`, `runs/exp#1.png`). I also guarded the URL parsing path so malformed `http(s)` inputs fall back without throwing.

Added tests covering each case — `data.csv?download=true`, the literal `?`/`#` filenames, and malformed URLs — all passing locally.

murdore:
🎉 Merged — thank you, @morgan-coded! This lands the more complete fix for #389 (query/hash stripping, URL decoding, lowercasing, and the malformed-URL fallback). Full disclosure: I had a narrower version of this in an internal branch and deliberately dropped it in favor of yours because yours is the more thorough take — so thank you for the clean rebase and for adding the non-URL `data.csv?download=true` case.

Now that you've been inside NeuroLink's file-detection internals: that same `FileDetector` you just improved is the front door to one of its most fun capabilities. You can hand **any** file — CSV, PDF, image, audio, video, or an Office doc — to `neurolink.generate({ input: { text: "...", files: ["./whatever.pdf"] } })` and it auto-detects the type, processes it (PDF→images or text, audio→transcription, CSV→structured rows, etc.), and routes it to any of 20+ LLM providers behind one API. It'd make a great base for a **"drop any file, ask questions about it"** tool or a document-analysis pipeline — and the detection accuracy you just improved is exactly what makes that reliable.

If you build something with it, we'd love to hear about it. Thanks again for the fix! 🙏

morgan-coded:
Appreciate the merge and the note, especially knowing you set aside an internal version for this one. I like the “drop any file, ask questions” idea and will share it if I end up building on NeuroLink’s file pipeline.

morgan-coded:
Built it. The "drop any file, ask questions about it" idea from your merge note is now a working tool: **neurolink-docchat** — https://github.com/morgan-coded/neurolink-docchat (v0.1.0).

What it is: a local CLI for grounded chat over a folder of files, on NeuroLink's public `./rag` surface (`RAGPipeline` + `FileDetector` doing exactly the front-door work you described). `docchat chat ./docs` for interactive sessions, `askfile` for one-shot questions, `--dry-run` to print the full plan with zero provider calls.

Engineering notes, since they're the interesting part:
- Answers cite sources as `[Sn]` IDs that are validated against the actually-retrieved corpus — unknown IDs are stripped even when a citation token arrives split across stream chunks.
- Sessions are app-owned JSON (atomic temp-and-rename writes) and bound to their corpus — a named session can't silently switch document sets.
- Intake is guarded: root confinement with symlink-escape rejection, deny-list families, size caps, and PDF placeholder detection so a missing optional parser can't silently poison the corpus.
- An adapter normalizes the usage token family across shapes and falls back to `analytics.tokenUsage` after stream drain; I'll file the one or two contract nits I hit there as separate issues with repros.

Proof: 51 offline tests (live paths gated behind RUN_LIVE=1), plus a redacted live transcript in the repo — the whole live verification ran on Google AI Studio's free tier: 32 calls, $0. OpenAI config exists but is honestly labeled untested. Durable vector indexes, OCR, and a web UI are deliberately out of v0.1.0.

One thing I'd like your word on: the name. `neurolink-docchat` borrows your brand to say what it's for — if you'd rather I rename it, say so and it's done.

PR #1123, opened 2026-07-01
fix(vertex): honor abortSignal + graceful step-cap message in native Gemini-3 loop

pdogra1299:
## Problem

On a Gemini-3 model over `provider=vertex`, a **pure function-call runaway** (the model keeps requesting tool calls, emitting reasoning only as `thoughtSignature`/`functionCall` parts with **no user-visible text**) hit the caller's `maxSteps`. NeuroLink then surfaced the literal placeholder

> `[Tool execution limit reached after N steps. The model continued requesting tool calls beyond the limit.]`

as the assistant answer, which callers post verbatim. Worse, a caller-supplied `abortSignal` (wall-clock timeout) was **ignored** by the Gemini loop — an 18-minute turn with 74 tool calls was observed, **30 of them after the abort was requested**.

This **extends the 9.79.3 lost-prose fix** (`preserve answer text + emit finishReason in native Gemini loop`). That fix only handled the case where the model emitted prose across steps (`accumulatedText` non-empty). It did **not** handle the pure function-call runaway, because on Gemini-3 `accumulatedText` is empty (no text parts) so the prose branch never fired, the recovery synth made another unbounded `generateContentStream` call with no `abortSignal`, and the loop never honored `options.abortSignal`.

## Changes

Applied identically to **both** native Gemini loops — `executeNativeGemini3Generate` (`.generate()`) and `executeNativeGemini3Stream` (`.stream()`):

- **`abortSignal` wiring.** An internal `AbortController` is tripped by the caller's signal *and* a defensive wall-clock timer (`DEFAULT_GEMINI_STREAM_TIMEOUT_MS = 300s`, overridable via `options.timeout`). `effectiveSignal` is passed as a **shallow-cloned** `config.abortSignal` on every `generateContentStream` request (the shared `config` is never mutated) and into tool execution (`toolOptions.abortSignal`). `@google/genai` forwards it to the underlying `fetch`, cancelling the in-flight drain. Listener + timer are cleaned up symmetrically in a `finally`.
- **Break, not throw.** A mid-drain / between-steps / tool-exec abort is converted to a graceful `break` into the terminal handler — **not** a re-throw. A re-throw would route the caller's wall-clock abort into a *second* unbounded fallback `generate()`/synth; breaking preserves the gathered tool results and degrades gracefully. An aborted tool call breaks the turn instead of being recorded as a spurious tool failure.
- **Abort-aware synth.** `synthesizeFinalAnswerWithoutTools` gained a trailing optional `abortSignal`, returns early (no request) when already aborted, and folds the signal into its own cloned config. The recovery synth is **skipped entirely** on the aborted path (so it can never add +300s after a blown budget); the clean step-cap path keeps its 300s synth budget.
- **Graceful message.** The bracketed placeholder is replaced in both loops **and** the shared `handleMaxStepsTermination` with a single-source `buildToolLoopCapMessage(maxSteps, toolCallCount)`.
- **`finishReason` mapping unchanged.** `"tool-calls"` on a clean step-cap without synth; SDK-mapped otherwise (an aborted turn maps from the SDK and the graceful message is delivered as ordinary prose). No new `"error"` value; no new public result fields.

## Non-regression

- `final_result` / structured output: a `final_result` turn sets `finalText` and breaks **before** the terminal block; the new guard `!finalText && (step >= maxSteps || wasAborted)` stays false, so `structuredOutput` and its SDK-mapped `finishReason` are preserved.
- Stream contract: normal multi-part completions still yield **>1 chunk**; the empty step-cap/abort case yields **exactly one** non-empty graceful chunk (never the placeholder, never zero).
- Token accounting: synth tokens still added on synth-success; the graceful message adds none.
- Config immutability: shallow-clone per request; shared `config` never mutated.

## Testing

New suite `test/continuous-test-suite-gemini-abort.ts` — **27 cases, all green** (`pnpm run test:gemini-abort`): the pure helpers (`buildToolLoopCapMessage` / `isAbortError` / `handleMaxStepsTermination`); the generate loop (pure-runaway → cap message + `tool-calls`; synth-returns-text; `accumulatedText`; abort mid-drain & between-steps; `config.abortSignal` wiring + config-not-mutated; tool-exec abort does not increment `failedTools`; `final_result` preserved); the stream twin (>1 chunk vs exactly-one graceful chunk; abort); and synth abort-awareness, plus a regression grep for the removed placeholder.

Verified locally: `tsc --noEmit` exit 0, `eslint` on all changed files exit 0. A **mutation test** (disabling the abort catch-guards) fails the relevant abort tests, confirming they are load-bearing rather than vacuous.

> Note: `pnpm run check`'s `svelte-kit sync` preamble fails locally on the unrelated `landing/` subproject (`@sveltejs/adapter-vercel` not installed) — pre-existing on `release`, independent of this diff. The `tsc --noEmit --strict` gate itself passes.

## Summary by CodeRabbit

* **New Features**
  * Added clearer, user-facing messaging when Gemini turns reach step limits, including friendlier tool-loop cap text.
  * Improved abort and timeout behavior for Gemini streaming and generation, allowing turns to stop more gracefully.
  * Added continuous test coverage for Gemini abort and step-cap scenarios, plus a dedicated test command.

* **Bug Fixes**
  * Reduced cases where capped/aborted Gemini runs could continue, retry unexpectedly, or show legacy placeholder text.
  * Improved terminal output behavior after tool execution, including safer synthesis fallback handling.

### DIFF
--- a/src/lib/providers/googleNativeGemini3.ts
+++ b/src/lib/providers/googleNativeGemini3.ts
@@ -1048,10 +1048,16 @@ export async function executeNativeToolCalls(
 }
 
 /**
- * Handle maxSteps termination by producing a final text when the model
- * was still calling tools when the step limit was reached.
+ * Handle maxSteps termination by producing a final answer when the model was
+ * still calling tools as the step limit was reached.
+ *
+ * Resolution order (always returns a non-empty string): the caller's
+ * `finalText` if present; else the `lastStepText` gathered from the final step;
+ * else a graceful, user-facing cap message from {@link buildToolLoopCapMessage}
+ * — which replaced the legacy bracketed step-limit placeholder string.
  *
  * @param logLabel - Label for log messages (e.g. "[GoogleAIStudio]" or "[GoogleVertex]")
+ * @returns The resolved final answer text (never the legacy placeholder).
  */
 export function handleMaxStepsTermination(
   logLabel: string,
@@ -1065,12 +1071,49 @@ export function handleMaxStepsTermination(
       `${logLabel} Tool call loop terminated after reaching maxSteps (${maxSteps}). ` +
         `Model was still calling tools. Using accumulated text from last step.`,
     );
-    return (
-      lastStepText ||
-      `[Tool execution limit reached after ${maxSteps} steps. The model continued requesting tool calls beyond the limit.]`
-    );
+    return lastStepText || buildToolLoopCapMessage(maxSteps, 0);
   }
   return finalText;
+}
+
+/**
+ * Detect whether an error represents an abort/cancellation (from an
+ * AbortSignal firing during a fetch/stream drain). Used by the native
+ * Gemini-3 agentic loops to break gracefully rather than re-throw.
+ */
+export function isAbortError(error: unknown): boolean {
+  if (!error) {
+    return false;
+  }
+  const e = error as { name?: string; message?: string; code?: number };
+  return (
+    e.name === "AbortError" ||
+    (typeof e.message === "string" && /abort/i.test(e.message)) ||
+    (typeof DOMException !== "undefined" &&
+      error instanceof DOMException &&
+      e.code === 20)
+  );
+}
+
+/**
+ * Build a graceful, user-facing message for when a single agentic turn hits
+ * the step cap (or is aborted mid-turn) without producing a final answer.
+ * Replaces the legacy bracketed "Tool execution limit reached" placeholder.
+ */
+export function buildToolLoopCapMessage(
+  maxSteps: number,
+  toolCallCount: number,
+): string {
+  const calls =
+    toolCallCount > 0
+      ? `I gathered information across ${toolCallCount} tool call${
+          toolCallCount === 1 ? "" : "s"
+        } but `
+      : "I ";
+  return (
+    `${calls}reached the ${maxSteps}-step limit for a single turn before I could finish. ` +
+    `Please narrow the request or break it into smaller asks and I'll continue.`
+  );
 }
 
 /**
--- a/src/lib/providers/googleVertex.ts
+++ b/src/lib/providers/googleVertex.ts
@@ -12,6 +12,7 @@ import {
 } from "../constants/enums.js";
 import { BaseProvider } from "../core/baseProvider.js";
 import {
+  DEFAULT_GEMINI_STREAM_TIMEOUT_MS,
   DEFAULT_MAX_STEPS,
   DEFAULT_TOOL_MAX_RETRIES,
   GLOBAL_LOCATION_MODELS,
@@ -77,8 +78,10 @@ import { TimeoutError, withTimeout } from "../utils/async/index.js";
 import { parseTimeout } from "../utils/timeout.js";
 import {
   appendStepText,
+  buildToolLoopCapMessage,
   createTextChannel,
   extractThoughtSignature,
+  isAbortError,
   mapGeminiFinishReason,
   prependConversationMessages,
 } from "./googleNativeGemini3.js";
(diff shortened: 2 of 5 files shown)


Tara-ag (approved):
## Review Complete ✅

I've analyzed all 5 changed files in this PR. The changes are well-architected and address the critical issues described:

### What's Fixed
1. **abortSignal honoring** - The native Gemini loops now properly respect caller-supplied abort signals and defensive wall-clock timeouts (300s default)
2. **Graceful step-cap message** - Replaces the bracketed placeholder with a user-friendly message via `buildToolLoopCapMessage()`
3. **Abort-aware synthesis** - Skips the 300s synthesis call when already aborted, preventing budget overruns

### Code Quality
- ✅ Follows established patterns from `executeNativeAnthropicStream`
- ✅ Proper cleanup in `finally` blocks (event listener removal, timer clearing)
- ✅ Config immutability preserved via shallow cloning
- ✅ Comprehensive test coverage (27 cases)
- ✅ No security issues detected

### Pre-existing Issue
The CodeRabbit comment about Prettier formatting in the test file is valid - please run `pnpm run lint` and fix formatting before merge.

**Files reviewed:** 5
**Blocking issues:** 0
**Severity:** Minor formatting only

Approving - the logic is sound and the abort handling is robust.

src/lib/providers/googleVertex.ts
Tara-ag, line 1: ⚠️ **MAJOR**: Missing `withTimeout` wrapper on the synthesis call in `executeNativeGemini3Stream`

The `synthesizeFinalAnswerWithoutTools` call at line ~2150 (in the `else` branch for pure functionCall turns) is NOT wrapped with `withTimeout()`, unlike the identical call in `executeNativeGemini3Generate` which IS properly wrapped.

Per the bootstrapped standards: "All async provider calls must be wrapped with withTimeout() to prevent indefinite hangs on unresponsive endpoints."

**Suggested fix:**
```typescript
// Wrap the synthesis call with timeout (parity with generate path)
const synth = await withTimeout(
  this.synthesizeFinalAnswerWithoutTools(
    client,
    modelName,
    config,
    currentContents,
    useFinalResultTool,
    parseTimeout(options.timeout) ?? 300_000,
    effectiveSignal,
  ),
  parseTimeout(options.timeout) ?? 300_000,
  "Gemini synthesis call timed out",
);
```

This ensures the stream path has the same timeout protection as the generate path.

Tara-ag, line 1: 💡 **MINOR**: Redundant `effectiveSignal.aborted` check in catch blocks

In both `executeNativeGemini3Stream` and `executeNativeGemini3Generate`, the abort error handling uses:
```typescript
if (effectiveSignal.aborted || isAbortError(error)) {
```

The `isAbortError()` function already checks for `AbortError` name and DOMException code 20. However, `effectiveSignal.aborted` is checked redundantly. This is technically fine but slightly confusing - if the signal is aborted but the error is NOT an AbortError (e.g., a different error thrown coincidentally at the same time), this would still treat it as an abort.

Consider simplifying to rely solely on `isAbortError(error)` for clarity, OR document why the dual-check is intentional (e.g., race condition between signal state and error propagation).

test/continuous-test-suite-gemini-abort.ts
Tara-ag, line 1: 💡 **MINOR**: Test file has Prettier formatting issues (already noted by CodeRabbit)

The existing review comment from CodeRabbit correctly identified that `pnpm run lint` is failing due to Prettier formatting issues in this file. Please run `pnpm run lint:fix` or `prettier --write test/continuous-test-suite-gemini-abort.ts` before merge.

This is a non-blocking style issue but should be addressed for CI cleanliness.

src/lib/providers/googleNativeGemini3.ts
Tara-ag, line 1: 💡 **MINOR**: `handleMaxStepsTermination` still references legacy placeholder in docstring

The function's docstring and the old implementation referenced the legacy placeholder `[Tool execution limit reached after ${maxSteps} steps...]`. While the implementation now correctly uses `buildToolLoopCapMessage()`, consider updating any stale comments that might still reference the old behavior.

Also, the function signature shows `lastStepText` as the fallback, but the new behavior uses `buildToolLoopCapMessage(maxSteps, 0)` when both are empty. This is correct behavior, but worth documenting clearly in the JSDoc.

Tara-ag (changes requested):
## Review Summary

**Files reviewed:** 5
**New issues raised:** 4 (1 MAJOR, 3 MINOR)

### 🔒 Blocking Issues (MAJOR)

1. **Missing `withTimeout` wrapper on synthesis call in `executeNativeGemini3Stream`**
   - The `synthesizeFinalAnswerWithoutTools` call in the stream path is NOT wrapped with `withTimeout()`, unlike the identical call in `executeNativeGemini3Generate` which IS properly wrapped.
   - Per project standards: "All async provider calls must be wrapped with withTimeout() to prevent indefinite hangs on unresponsive endpoints."
   - **Suggested fix:** Wrap the synthesis call with `withTimeout()` to match the generate path's timeout protection.

### 💡 Non-Blocking Issues (MINOR)

2. **Redundant `effectiveSignal.aborted` check in catch blocks**
   - The dual-check pattern `if (effectiveSignal.aborted || isAbortError(error))` is slightly confusing since `isAbortError()` already handles abort detection.
   - Consider documenting why both checks are needed (race conditions) or simplify to rely on `isAbortError()`.

3. **Test file Prettier formatting**
   - Already noted by CodeRabbit: `pnpm run lint` fails due to Prettier issues in `test/continuous-test-suite-gemini-abort.ts`.
   - Run `pnpm run lint:fix` before merge.

4. **Documentation update for `handleMaxStepsTermination`**
   - The function's JSDoc could be clearer about the new `buildToolLoopCapMessage()` fallback behavior.

### Overall Assessment

This PR addresses critical abortSignal handling and graceful step-cap messaging issues in the Gemini-3 native Vertex loops. The changes are architecturally sound and include comprehensive test coverage (27 cases). 

**However, the missing `withTimeout()` wrapper on the stream path's synthesis call is a significant oversight that could lead to hung requests on unresponsive endpoints. This must be fixed before merge to maintain consistency with the generate path and project standards.**

Once the timeout wrapper is added and formatting is fixed, this PR should be ready for approval.

src/lib/providers/googleVertex.ts
pdogra1299: Thanks for the careful review — I dug in and this is actually a **false positive**: neither path double-wraps, because `synthesizeFinalAnswerWithoutTools` bounds its **entire** connect + drain with `withTimeout(timeoutMs)` *internally* (`googleVertex.ts` L3277–L3336, message `"Gemini synthesis call timed out"`) and never throws — its `try/catch` returns `{ text: "" }` on any error, including `TimeoutError`.

Both call sites invoke it identically, with the same budget and **no** external wrapper:
- stream path: `parseTimeout(options.timeout) ?? DEFAULT_GEMINI_STREAM_TIMEOUT_MS`
- generate path: same

So the generate path is **not** externally wrapped either — the two are symmetric and both are timeout-protected via the internal `withTimeout`. Adding an outer `withTimeout` would nest a redundant timer that can never fire before the inner one. I've added a clarifying comment at both call sites (`5358cc5c`) noting the self-bounding so it no longer reads as a gap. Happy to reconsider if I've missed something.

pdogra1299: Both conditions are load-bearing — documented at all four catch sites in `5358cc5c`:
- `effectiveSignal.aborted` catches an abort **we** triggered (the caller's signal or the defensive wall-clock timer) even when the thrown error isn't abort-shaped (e.g. the SDK surfaces a generic error on cancel).
- `isAbortError(error)` catches an abort-shaped throw **without** our signal firing — e.g. a tool that raises `AbortError` (exercised by the "aborted tool call does NOT increment failedTools" test).

A mutation test confirms the guard is needed: deleting it flips that test from pass → fail.

test/continuous-test-suite-gemini-abort.ts
pdogra1299: Fixed — ran the project formatter; `prettier --check .` is clean and the **Check code formatting** CI step is green (`c4dd81b9`, carried into `5358cc5c`).

src/lib/providers/googleNativeGemini3.ts
pdogra1299: Done (`5358cc5c`) — expanded the JSDoc to spell out the resolution order (`finalText` → `lastStepText` → `buildToolLoopCapMessage`) and that it always returns a non-empty string, with an `@returns` note that it never emits the legacy placeholder. Also refreshed the stale "falls back to the placeholder" wording in `synthesizeFinalAnswerWithoutTools`'s docstring. Kept the prose free of the literal placeholder text so the source-level regression-grep test stays meaningful.

pdogra1299:
## Review feedback addressed — pushed as `5358cc5c`

Thanks @Tara-ag and CodeRabbit. How each item was handled:

| Item | Resolution |
|---|---|
| **MAJOR** — stream synth "missing `withTimeout`" | **False positive.** `synthesizeFinalAnswerWithoutTools` wraps its entire connect+drain in `withTimeout(timeoutMs)` *internally* (`googleVertex.ts` L3277–L3336) and never throws. **Both** call sites (stream + generate) invoke it identically with **no** external wrapper — they're symmetric and already timeout-protected. An outer wrapper would be a redundant nested timer. Added a clarifying comment at both sites; evidence in the inline reply. |
| MINOR — "redundant" `effectiveSignal.aborted \|\| isAbortError` | Both are load-bearing (an abort *we* tripped whose error isn't abort-shaped, vs. an abort-shaped throw with no signal firing, e.g. a tool raising `AbortError`). Documented at all four catch sites; a mutation test confirms the guard is needed. |
| MINOR — test Prettier | Fixed; the **Check code formatting** CI step is green. |
| MINOR — `handleMaxStepsTermination` JSDoc | Expanded to document the `finalText → lastStepText → buildToolLoopCapMessage` resolution order + `@returns`, and refreshed the stale "placeholder" wording in the synth docstring. |
| CodeRabbit nit — literal `300_000` | Now uses `DEFAULT_GEMINI_STREAM_TIMEOUT_MS` at both synth call sites. |

**Bonus hardening:** while touching the docstrings I noticed the source-level regression-grep test could be silently **skipped** — its assertion message matched the harness's "expected provider error" filter (the word *"gone"* → HTTP 410), so a genuine placeholder-regression would have passed CI. Reworded the messages so it now **fails loudly**, verified by a mutation test.

Green locally: `tsc --noEmit` 0 · `eslint` 0 · `prettier --check .` clean · suite **27/27** · single commit preserved. A re-review would be appreciated when you have a moment. 🙏

Tara-ag (approved):
## Review Summary

I've completed a thorough review of this PR addressing the Gemini-3 native Vertex loop abort signal and graceful step-cap messaging issues.

### ✅ What's Good

**Architecture:**
- Abort scaffolding correctly mirrors the established `executeNativeAnthropicStream` pattern
- Dual abort detection (`effectiveSignal.aborted || isAbortError(error)`) is properly load-bearing as documented
- Symmetric cleanup in `finally` blocks ensures no listener/timer leaks
- Config immutability preserved via shallow cloning
- Tool execution properly receives the abort signal

**Testing:**
- Comprehensive 27-case test suite covering helpers, generate/stream loops, abort scenarios, and regression greps
- Tests verify both the signal wiring and the graceful degradation paths

**Existing Comments Addressed:**
- Prettier formatting: Fixed per author
- withTimeout wrapper: Author's explanation correct - internal self-bounding is sufficient
- Redundant abort check: Author's explanation correct - both checks are load-bearing
- Stale docstring: Fixed per author

### 📊 Issue Count
- **Critical:** 0
- **Major:** 0  
- **Minor:** 0 (existing style issue already flagged and fixed)

### 🎯 Recommendation
**Approve** - The PR is production-ready. It fixes the reported 18-minute runaway turns with 30 post-abort tool calls, and replaces the unfriendly placeholder message with actionable guidance for users.

PR #1131, opened 2026-07-04
fix(mcp): do not cache error tool results (BZ-664 follow-up)

murdore:
## Problem

The MCP tool result cache (BZ-664, `mcpToolResultCache`, default-on, 300s TTL) stores **every** result that doesn't throw — including **error** results (`isError: true` / `success: false`). The store site had no success guard:

```ts
// BZ-664: Store result in cache after successful execution   <-- comment, but no guard
if (cacheEnabled && this.mcpToolResultCache) {
  this.mcpToolResultCache.cacheResult(toolName, cacheKeyArgs, result);
}
```

When an external MCP server surfaces a **transient** upstream failure (e.g. a 403 or timeout) as a tool error result, that error is cached for the full TTL and replayed to every caller with the same args — turning a single transient failure into a **storm of identical cached failures**, and preventing a retry from re-hitting the server after it recovers (the "retry" just replays the stale cached error).

## Evidence (downstream production analysis)

Investigating a downstream service (Curator/Tara) whose Bitbucket MCP was "failing constantly and burning tokens":
- **~79%** of the "permission denied" failures completed in **~6ms** vs **~363ms** for a real network call (a 404 from the same server takes ~119ms) — i.e. they never hit the network; they were cache replays.
- **34%** were exact duplicates; one `search_code` 403 was replayed **59×**.
- **81%** of duplicate-failure gaps fell within the 300s cache TTL.
- Net effect: a few hundred real transient 403s were amplified into **thousands** of logged failures and wasted LLM tokens/turns.

## Fix

Guard `cacheResult()` so only **successful** results are cached; error results always re-execute. Error detection mirrors the existing `isToolError` check in this file (`isError: true` or `success: false`). Successful-call dedup (the intended BZ-664 benefit) is preserved.

## Tests

New `test/mcpResultCacheErrorSkip.test.ts` (4 tests, all passing):
- error result (`isError: true`) is **not** cached
- error result (`success: false`) is **not** cached
- successful result **is** cached
- after an error, an identical retry **re-executes** and gets the recovered result (no stale replay)

`tsc --noEmit --strict` clean; prettier clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented failed external tool responses from being saved to cache, including error-shaped and `undefined` outputs.
  * Ensured identical requests re-execute after a failure instead of replaying a cached error.
  * Confirmed successful tool results are still cached and reused.
  * Added verification that recovery after an initial failure works correctly.
* **Tests**
  * Added Vitest coverage for cache behavior across success, error-shaped, undefined, and recovery scenarios.

### DIFF
--- a/src/lib/neurolink.ts
+++ b/src/lib/neurolink.ts
@@ -14368,13 +14368,38 @@ Current user's request: ${currentInput}`;
         options,
       );
 
-      // BZ-664: Store result in cache after successful execution
-      if (cacheEnabled && this.mcpToolResultCache) {
+      // BZ-664: Store result in cache after successful execution.
+      // Only cache SUCCESSFUL results. Caching an error result (isError:true /
+      // success:false — e.g. an upstream 403/timeout surfaced as a tool error)
+      // replays that identical failure to every caller with the same args for
+      // the whole TTL, turning a single transient upstream error into a storm
+      // of cached failures and preventing a retry from re-hitting the server
+      // once it recovers. Errors must always re-execute. (Detection mirrors the
+      // isToolError check used elsewhere in this file.)
+      const resultObj =
+        result && typeof result === "object"
+          ? (result as Record<string, unknown>)
+          : undefined;
+      const isErrorResult = Boolean(
+        (resultObj && "isError" in resultObj && resultObj.isError === true) ||
+        (resultObj && "success" in resultObj && resultObj.success === false),
+      );
+      // Also skip an `undefined` result: the cache read side treats `undefined`
+      // as a miss (`cached !== undefined`), so storing it is a dead entry that can
+      // never be read back — keep write/read symmetric with the other cache sites.
+      if (
+        cacheEnabled &&
+        this.mcpToolResultCache &&
+        !isErrorResult &&
+        result !== undefined
+      ) {
         this.mcpToolResultCache.cacheResult(toolName, cacheKeyArgs, result);
       }
 
       mcpLogger.debug(
-        `[NeuroLink] External MCP tool executed successfully: ${toolName}`,
+        `[NeuroLink] External MCP tool ${
+          isErrorResult ? "returned error" : "executed successfully"
+        }: ${toolName}`,
       );
       return result;
     } catch (error) {
--- a/test/mcpResultCacheErrorSkip.test.ts
+++ b/test/mcpResultCacheErrorSkip.test.ts
@@ -1,0 +1,151 @@
+/**
+ * BZ-664 follow-up: the MCP tool result cache must NOT store error results.
+ *
+ * Caching an error result (isError:true / success:false — e.g. an upstream 403
+ * or timeout surfaced as a tool error) replays that identical failure to every
+ * caller with the same args for the whole TTL, turning a single transient
+ * upstream error into a storm of cached failures and preventing a retry from
+ * re-hitting the server once it recovers.
+ *
+ * These tests drive `executeExternalMCPTool` with a stubbed external server and
+ * assert the cache stores successful results but skips error results.
+ */
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+
+type CacheLike = {
+  cacheResult: (tool: string, args: unknown, result: unknown) => void;
+  getCachedResult: (tool: string, args: unknown) => unknown;
+};
+
+type NeuroLinkInternals = {
+  mcpToolResultCache?: CacheLike;
+  externalServerManager: { executeTool: ReturnType<typeof vi.fn> };
+};
+
+describe("MCP tool result cache — error results are not cached (BZ-664 follow-up)", () => {
+  let instance: unknown;
+  let internals: NeuroLinkInternals;
+  let cacheSpy: ReturnType<typeof vi.spyOn>;
+
+  beforeEach(async () => {
+    const { NeuroLink } = await import("../src/lib/neurolink.js");
+    // Empty config => MCP result cache is enabled by default (mcp.cache.enabled !== false).
+    instance = new NeuroLink({});
+    internals = instance as unknown as NeuroLinkInternals;
+
+    // Spy on the REAL ExternalServerManager.executeTool instead of swapping the
+    // whole instance out. Swapping would orphan the manager the constructor built
+    // — its process/event listeners would stay attached for the rest of the suite,
+    // because afterEach's dispose()/shutdown() would then run against the bare stub.
+    // Spying keeps the single real instance, so teardown actually reaches it.
+    vi.spyOn(
+      internals.externalServerManager as unknown as {
+        executeTool: (...args: unknown[]) => unknown;
+      },
+      "executeTool",
+    ).mockResolvedValue(undefined);
+
+    expect(internals.mcpToolResultCache).toBeDefined();
+    cacheSpy = vi.spyOn(
+      internals.mcpToolResultCache as CacheLike,
+      "cacheResult",
+    );
+  });
+
+  afterEach(async () => {
+    // The NeuroLink constructor spins up an ExternalServerManager, event emitters
+    // and circuit breakers; dispose the instance so listeners/timers don't leak
+    // across the suite. Guarded because the teardown method name is version-tied.
+    const disposable = instance as {
+      dispose?: () => Promise<void> | void;
+      shutdown?: () => Promise<void> | void;
+    };
+    await disposable.dispose?.();
+    await disposable.shutdown?.();
+    vi.restoreAllMocks();
+  });
+
+  const call = (tool: string, args: Record<string, unknown>) =>
+    (
+      instance as {
+        executeExternalMCPTool: (
+          s: string,
+          t: string,
+          a: Record<string, unknown>,
+        ) => Promise<unknown>;
+      }
+    ).executeExternalMCPTool("srv", tool, args);
+
+  it("does NOT cache a result with isError:true", async () => {
+    internals.externalServerManager.executeTool.mockResolvedValue({
+      isError: true,
+      content: [{ type: "text", text: "Permission denied (403)" }],
+    });
+
+    await call("bb_search", { q: "x" });
+
+    expect(cacheSpy).not.toHaveBeenCalled();
+  });
+
+  it("does NOT cache a result with success:false", async () => {
+    internals.externalServerManager.executeTool.mockResolvedValue({
+      success: false,
+      content: [{ type: "text", text: "boom" }],
+    });
+
+    await call("bb_list", { p: "PROJ" });
+
+    expect(cacheSpy).not.toHaveBeenCalled();
+  });
+
+  it("DOES cache a successful result and serves the 2nd identical call from cache", async () => {
+    const executeTool = internals.externalServerManager.executeTool;
+    executeTool.mockResolvedValue({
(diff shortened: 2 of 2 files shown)


src/lib/neurolink.ts
```diff
@@ -14368,8 +14368,23 @@ Current user's request: ${currentInput}`;
         options,
       );
 
-      // BZ-664: Store result in cache after successful execution
-      if (cacheEnabled && this.mcpToolResultCache) {
+      // BZ-664: Store result in cache after successful execution.
+      // Only cache SUCCESSFUL results. Caching an error result (isError:true /
+      // success:false — e.g. an upstream 403/timeout surfaced as a tool error)
```
Tara-ag, line 14373: 💡 **MINOR**: The debug log message "External MCP tool executed successfully" is misleading when the result is an error (`isError:true` or `success:false`). 

Since `isErrorResult` is now computed above, consider updating the log to reflect the actual status:

```typescript
mcpLogger.debug(
  `[NeuroLink] External MCP tool ${isErrorResult ? "returned error" : "executed successfully"}: ${toolName}`,
);
```

This improves observability and prevents confusion when debugging tool execution issues.

test/mcpResultCacheErrorSkip.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  it("DOES cache a successful result", async () => {
+    internals.externalServerManager.executeTool.mockResolvedValue({
+      content: [{ type: "text", text: "ok" }],
+    });
+
+    await call("bb_get", { path: "a.ts" });
+
+    expect(cacheSpy).toHaveBeenCalledTimes(1);
+  });
```
Tara-ag, line 126: ⚠️ **MAJOR**: This test only verifies that `cacheResult()` was called, but doesn't verify that the cache actually returns the cached result on subsequent calls. A broken cache implementation that never reads would still pass this test.

**Suggestion**: Strengthen the test by calling the tool twice with identical args and asserting the second call doesn't re-execute:

```typescript
it("DOES cache a successful result and returns it on subsequent calls", async () => {
  internals.externalServerManager.executeTool.mockResolvedValue({
    content: [{ type: "text", text: "ok" }],
  });

  const first = await call("bb_get", { path: "a.ts" });
  const second = await call("bb_get", { path: "a.ts" });

  // Should only execute once - second call hits cache
  expect(executeTool).toHaveBeenCalledTimes(1);
  expect(first).toEqual(second);
  expect(cacheSpy).toHaveBeenCalledTimes(1);
});
```

This ensures the cache is actually functional, not just that `cacheResult()` was invoked.

src/lib/neurolink.ts
```diff
@@ -14368,8 +14368,23 @@ Current user's request: ${currentInput}`;
...
+      // isToolError check used elsewhere in this file.)
+      const resultObj =
+        result && typeof result === "object"
+          ? (result as Record<string, unknown>)
+          : undefined;
+      const isErrorResult = Boolean(
+        (resultObj && "isError" in resultObj && resultObj.isError === true) ||
+        (resultObj && "success" in resultObj && resultObj.success === false),
+      );
+      if (cacheEnabled && this.mcpToolResultCache && !isErrorResult) {
```
Tara-ag: ⚠️ **MAJOR**: The `isErrorResult` detection logic has a subtle issue. When `result` is `undefined`, `resultObj` becomes `undefined`, and `isErrorResult` evaluates to `false`. This means `undefined` results will be cached.

However, looking at the cache read side (`executeToolInternal`), `undefined` is treated as a cache miss (`cached !== undefined`). This creates an inconsistency:
- Cache write: `undefined` gets stored
- Cache read: `undefined` is treated as miss

**Suggestion**: Add an explicit check to skip caching when `result === undefined`:

```typescript
if (cacheEnabled && this.mcpToolResultCache && !isErrorResult && result !== undefined) {
  this.mcpToolResultCache.cacheResult(toolName, cacheKeyArgs, result);
}
```

This aligns with other cache sites in this file that skip caching for `undefined` results.

Tara-ag (approved):
## Review Summary

**Files reviewed:** 2
**New issues raised:** 3 (2 MAJOR, 1 MINOR)

### Assessment

The fix correctly addresses the BZ-664 follow-up issue: **error results are no longer cached**, preventing transient upstream failures from being replayed to all callers for the full TTL.

### Issues Raised

| Severity | File | Line | Issue |
| ⚠️ MAJOR | `src/lib/neurolink.ts` | 14387 | `undefined` results may be cached inconsistently (cache write stores, cache read treats as miss) |
| 💡 MINOR | `src/lib/neurolink.ts` | 14373 | Debug log says "executed successfully" even for error results |
| ⚠️ MAJOR | `test/mcpResultCacheErrorSkip.test.ts` | 89 | Test only asserts `cacheResult()` called, doesn't verify cache hits work |

### Decision: **APPROVE** ✅

**Rationale:**
- The core fix is correct and addresses the production issue described in the PR
- The `undefined` caching edge case is pre-existing, not introduced by this PR
- The test coverage is adequate for the primary fix (error results not cached)
- No violations of CLAUDE.md Critical Rules
- No security issues or breaking API changes

### Recommendations (non-blocking)

1. Consider addressing the `undefined` caching inconsistency in a follow-up
2. Update the debug log message to reflect actual execution status
3. Strengthen the "DOES cache" test to verify cache read behavior

The PR is safe to merge as-is.

test/mcpResultCacheErrorSkip.test.ts
```diff
@@ -0,0 +1,144 @@
...
+    internals.externalServerManager = {
+      executeTool: vi.fn(),
+    } as unknown as NeuroLinkInternals["externalServerManager"];
+
+    expect(internals.mcpToolResultCache).toBeDefined();
+    cacheSpy = vi.spyOn(
+      internals.mcpToolResultCache as CacheLike,
+      "cacheResult",
+    );
+  });
```
Tara-ag, line 53: ⚠️ **MAJOR**: The test setup replaces `internals.externalServerManager` with a stub **after** `NeuroLink({})` has already instantiated the real `ExternalServerManager` in its constructor.

The `ExternalServerManager` constructor registers process event listeners (`SIGINT`, `SIGTERM`, `beforeExit`) that are never cleaned up because:
1. The stub replacement happens after construction
2. The `afterEach` teardown calls `dispose()`/`shutdown()` on the stub, not the original
3. The original manager's listeners remain attached for the test process lifetime

**Suggestion**: Either mock the manager before `NeuroLink` instantiation, or explicitly tear down the original before replacement:

```typescript
beforeEach(async () => {
  const { NeuroLink } = await import("../src/lib/neurolink.js");
  
  // Option 1: Mock before construction
  const mockManager = { executeTool: vi.fn() };
  // ... inject mock into constructor or use factory override
  
  instance = new NeuroLink({});
  internals = instance as unknown as NeuroLinkInternals;
  
  // Option 2: Teardown original before replacement
  await (instance as { dispose?: () => Promise<void> }).dispose?.();
  internals.externalServerManager = mockManager;
  
  // ... rest of setup
});
```

This prevents listener accumulation across the test suite.

Tara-ag (commented):
## Review Summary

**Files reviewed:** 2 (`src/lib/neurolink.ts`, `test/mcpResultCacheErrorSkip.test.ts`)

### Changes Overview
This PR fixes a critical issue where the MCP tool result cache was storing **error results** (`isError: true` / `success: false`), causing transient upstream failures to be replayed to all callers for the full TTL. The fix ensures only successful results are cached, allowing retries to re-execute against the server after recovery.

### Issues Raised

| 🔒 CRITICAL | 0 | None |
| ⚠️ MAJOR | 1 | Test setup leaks ExternalServerManager process listeners |
| 💡 MINOR | 0 | None |
| 💬 SUGGESTION | 0 | None |

### Detailed Findings

**1. Test Setup Leak (MAJOR)** - `test/mcpResultCacheErrorSkip.test.ts:46`
The test replaces `internals.externalServerManager` with a stub after `NeuroLink({})` has already instantiated the real `ExternalServerManager`. The constructor registers `process.on('SIGINT'/'SIGTERM'/'beforeExit')` handlers that are never cleaned up because:
- The stub replacement happens post-construction
- `afterEach` teardown calls `dispose()` on the stub, not the original
- Original listeners remain attached for the test process lifetime

**Recommendation:** Mock before construction or explicitly teardown the original before replacement.

### Verification
- ✅ Error detection logic correctly identifies `isError: true` and `success: false`
- ✅ Undefined results are skipped (symmetric with cache read side treating `undefined` as miss)
- ✅ Debug log updated to reflect actual execution status
- ✅ Tests cover: error skip (both forms), success cache, undefined skip, recovery behavior
- ✅ End-to-end cache verification strengthened (2nd call doesn't re-execute)

### Existing Comments Status
- Most prior comments from copilot/Tara-ag are **outdated** (addressed in code updates)
- CodeRabbit's ExternalServerManager disposal concern is **valid and unresolved** (addressed in my comment above)

### Decision
**Approve with comment** - The production code fix is correct and well-tested. The test setup issue should be addressed to prevent resource leaks in the test suite.

murdore:
All review threads on this PR have been addressed — resolving them for clarity.

**Fixed in `ace1e376`:**
- **`undefined` results no longer cached** — the write side now matches the read side (`cached !== undefined`), so no dead cache entries. _(Copilot + Tara-ag MAJOR)_
- **Debug log reflects actual status** — `returned error` vs `executed successfully`, via the already-computed `isErrorResult`. _(Copilot + Tara-ag MINOR)_
- **"DOES cache" test strengthened** — now calls the tool twice with identical args and asserts the external server is hit exactly once (proves reads actually hit, not just that `cacheResult()` was invoked), plus a dedicated `undefined`-skip test. _(Copilot + Tara-ag MAJOR)_

**Fixed in `8d431ffb`:**
- **Test no longer orphans the real `ExternalServerManager`** — instead of swapping the whole instance out after `new NeuroLink({})`, it now `vi.spyOn`s `executeTool` in place, so `afterEach`'s `dispose()`/`shutdown()` tears down the single real instance and no `process`/event listeners leak across the suite. _(CodeRabbit + Tara-ag MAJOR)_

5/5 unit tests pass; the full `validate + test:ci` pre-push gate passed on `8d431ffb`.

Tara-ag (approved):
## Review Summary

**Files Reviewed:** 2
- `src/lib/neurolink.ts` - Error result caching fix
- `test/mcpResultCacheErrorSkip.test.ts` - Comprehensive test coverage

**Issues Found:** None

### Assessment

This PR effectively fixes the BZ-664 follow-up issue where transient upstream failures (403s, timeouts) were being cached and replayed to all callers with the same args, turning single transient errors into storms of cached failures.

**Key aspects of the fix:**
1. ✅ Error detection mirrors existing patterns (`isError: true` / `success: false`)
2. ✅ `undefined` results skipped (maintains read/write symmetry)
3. ✅ Debug log updated to accurately reflect status
4. ✅ All previously raised review comments addressed

**Test coverage is thorough:**
- Error results (`isError: true`) are not cached
- Error results (`success: false`) are not cached  
- Successful results are cached and served from cache
- `undefined` results are not cached
- Recovery after error works correctly (re-executes instead of replaying cached error)

**Standards compliance:**
- No security issues (no hardcoded secrets, no injection risks)
- No CLAUDE.md critical rule violations
- Proper type safety with null/undefined handling

**Approved for merge.**

PR #1140, opened 2026-07-08
fix(ci): add missing typecheck script for docs PR validation

murdore:
## Problem
The `Documentation PR Validation` workflow (`.github/workflows/docs-pr-validation.yml`) runs a **TypeScript check** step:
```yaml
- name: TypeScript check
  run: pnpm run typecheck
```
But there is **no `typecheck` script** in `package.json`, so this step fails with `ERR_PNPM_NO_SCRIPT` on *every* PR that touches `docs/` or `docs-site/` — making the `Validate Documentation` check fail repo-wide regardless of the PR's actual content.

## Fix
Add a lean `typecheck` script:
```json
"typecheck": "tsc --noEmit"
```
This is what the workflow already expects. Verified locally: `pnpm run typecheck` exits 0 on current `release`.

## Notes
- This PR touches only `package.json`, so it does not itself trigger the docs-validation workflow (which is path-filtered to `docs/`/`docs-site/`), but it unblocks that check for all future docs PRs.
- The heavier `check` script (`svelte-kit sync && svelte-check && tsc --noEmit --strict`) is left as-is; `typecheck` is intentionally the minimal TS-only gate the docs workflow wants.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "build:react-hooks": "npx tsc --jsx react-jsx --module nodenext --moduleResolution nodenext --target esnext --esModuleInterop --skipLibCheck --outDir dist --declaration false src/lib/client/reactHooks.tsx",
     "check": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json && tsc --noEmit --strict",
     "check:watch": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --watch",
+    "typecheck": "tsc --noEmit",
     "modelServer": "tsx scripts/modelServer.ts",
     "lint": "prettier --check . && NODE_OPTIONS='--max-old-space-size=8192' eslint .",
     "format": "prettier --write .",


package.json
```diff
@@ -44,6 +44,7 @@
     "build:react-hooks": "npx tsc --jsx react-jsx --module nodenext --moduleResolution nodenext --target esnext --esModuleInterop --skipLibCheck --outDir dist --declaration false src/lib/client/reactHooks.tsx",
     "check": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json && tsc --noEmit --strict",
     "check:watch": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --watch",
+    "typecheck": "tsc --noEmit",
```
Tara-ag, line 47: 💡 **MINOR: `typecheck` script duplicates docs-site script but may run the wrong TS project**

The new root `package.json` script `"typecheck": "tsc --noEmit"` matches the existing `docs-site/package.json` script, but the docs workflow runs with `working-directory: docs-site` and installs dependencies there. When invoked from the repo root, `tsc --noEmit` will pick up the root `tsconfig.json`, which covers the entire Neurolink SDK/CLI source tree. That is a much wider (and slower) check than the docs workflow likely intends, and it may fail on unrelated SDK type errors even though the workflow is supposed to validate documentation only.

If the goal is to satisfy the docs workflow, consider whether the script belongs in `docs-site/package.json` (it already exists there) and whether the root script is necessary. If you keep the root script, document its scope or align it with the `check` script so callers know it is a full-project gate.

Tara-ag (commented):
## Review Summary

**Files reviewed:** 1 (`package.json`)
**New issues raised:** 1 MINOR

### Findings

- 💡 **MINOR** — `package.json:47`: The newly added `"typecheck": "tsc --noEmit"` script at the repo root may run the wrong TypeScript project scope. The docs workflow already runs from `docs-site/` and already has an identical `typecheck` script there. Invoked from the root, `tsc --noEmit` will use the root `tsconfig.json` and type-check the entire SDK/CLI codebase, which is broader than the documentation-only gate the workflow expects.

### Recommendation

Either remove the root script and rely on `docs-site/package.json`'s existing `typecheck`, or clarify the root script's scope (e.g., rename/document it as a full-project gate) so it does not surprise callers.

### Blocking assessment

No security vulnerabilities, no breaking public SDK API changes, and no violations of non-negotiable CLAUDE.md Critical Rules were introduced. The single issue is a scope/clarity concern, not a blocker. Approving is conditional on the author considering the scope note above; no further action is required if the root-level script is intentional.

PR #1141, opened 2026-07-10
fix(anthropic): prompt-cache breakpoints + accounting parity for the direct-Anthropic path

murdore:
## Problem

The native **Vertex+Claude** path has had full prompt caching since 9.79.1 (`applyVertexAnthropicCacheBreakpoints`: system + tools + rolling history markers, cache-token capture, discounted pricing). The **direct `anthropic` provider** was left behind in three ways:

1. **No rolling history breakpoints.** Only the system prompt (via MessageBuilder) and the last tool definition (via GenerationHandler) carry `cache_control` — the growing, tool-result-heavy conversation history falls *after* the last marker and is re-billed at full input price on every turn.
2. **generate()-path cache accounting silently dropped.** The native V3 delegating model reports cache reads/writes only through ai\@6's normalized usage shape — flat `cachedInputTokens` and nested `inputTokenDetails.{cacheReadTokens,cacheWriteTokens}` (`asLanguageModelUsage()`) — none of which `extractCacheReadTokens`/`extractCacheCreationTokens` recognized. Caching happened and Anthropic billed the discount, but NeuroLink reported the cache tokens as undefined. (Bedrock hits the same GenerationHandler path.)
3. **stream()-path cache accounting never attempted.** `executeStream()` accumulated only `input_tokens`/`output_tokens` off `message_start`; `cache_read_input_tokens`/`cache_creation_input_tokens` on the same event were never read.

## Fix

- **`utils/anthropicCacheBreakpoints.ts`** — new `applyAnthropicHistoryCacheBreakpoints()` (rolling tail markers for paths whose stable-prefix markers are managed upstream) + `countAnthropicCacheMarkers()`. The count feeds the budget so a request can **never exceed Anthropic's four-marker cap** — critical on the direct path, where the AI-SDK pipeline has already placed system/tool markers before the provider sees the request. Already-marked tail blocks are skipped without consuming budget, so per-step re-application stays idempotent. `applyVertexAnthropicCacheBreakpoints` is untouched (all 34 existing tests pass unchanged).
- **`providers/anthropic.ts`** — `doGenerate` and the `executeStream` step loop apply the history breakpoints (re-applied per step, mirroring the Vertex loops); `executeStream` now accumulates cache reads/writes and threads them through `resolveUsage` into the OTel span (`gen_ai.usage.cached_input_tokens`) and `calculateCost` (which already prices `cacheReadTokens`/`cacheCreationTokens`).
- **`utils/tokenUtils.ts` + `types/common.ts`** — extractors recognize the ai\@6 shapes; legacy field names keep precedence, so no behavior change for any existing provider.

## Tests

Extended `test/continuous-test-suite-cache-breakpoints.ts` (`pnpm run test:cache`): **47/47 pass** — 13 new assertions covering marker counting, budget clamping (incl. zero/negative), newest-first placement, purity, idempotent re-application, and the ai\@6 extraction shapes.

## Origin

Found while investigating cached-token accounting for curator (Claude-on-Vertex works today; this brings the currently-dormant direct-Anthropic path to parity before anyone switches to it, e.g. for OAuth/Claude-subscription access).

## Summary by CodeRabbit

- **New Features**
  - Added Anthropic prompt caching across standard and multi-step streaming requests.
  - Improved cache breakpoint handling for ongoing conversations while respecting existing cache markers.
  - Added support for reporting cache read and cache creation token usage.

- **Bug Fixes**
  - Updated token accounting and cost calculations to include cached input tokens.
  - Improved compatibility with normalized provider usage formats for cache metrics.

### DIFF
--- a/src/lib/providers/anthropic.ts
+++ b/src/lib/providers/anthropic.ts
@@ -53,6 +53,12 @@ import {
 } from "../types/index.js";
 import { logger } from "../utils/logger.js";
 import { redactUrlCredentials } from "../utils/logSanitize.js";
+import {
+  ANTHROPIC_MAX_CACHE_BREAKPOINTS,
+  applyAnthropicHistoryCacheBreakpoints,
+  countAnthropicCacheMarkers,
+} from "../utils/anthropicCacheBreakpoints.js";
+import type { VertexAnthropicMessage } from "../types/index.js";
 import { calculateCost } from "../utils/pricing.js";
 import {
   createAnthropicConfig,
@@ -1446,9 +1452,27 @@ export class AnthropicProvider extends BaseProvider {
           | { type: "enabled"; budget_tokens: number }
           | undefined;
 
+        // Prompt-cache parity with the native Vertex+Claude path: upstream
+        // layers mark only the stable prefix (system via MessageBuilder,
+        // last tool via GenerationHandler) — the growing conversation
+        // history has no breakpoint, so on every turn it falls after the
+        // last marker and is re-billed as fresh input. Add rolling history
+        // breakpoints in whatever budget remains under Anthropic's
+        // four-marker ceiling; pre-existing markers are counted so the
+        // request can never exceed the cap.
+        const cacheMarkersUsed = countAnthropicCacheMarkers({
+          system,
+          tools,
+          messages: messages as unknown as VertexAnthropicMessage[],
+        });
+        const cachedMessages = applyAnthropicHistoryCacheBreakpoints(
+          messages as unknown as VertexAnthropicMessage[],
+          ANTHROPIC_MAX_CACHE_BREAKPOINTS - cacheMarkersUsed,
+        ) as unknown as Anthropic.Messages.MessageParam[];
+
         const params: Anthropic.Messages.MessageCreateParamsNonStreaming = {
           model: modelId,
-          messages,
+          messages: cachedMessages,
           max_tokens: resolveClaudeMaxTokens(modelId, options.maxOutputTokens),
           ...(system ? { system } : {}),
           ...(options.temperature !== undefined &&
@@ -1741,10 +1765,22 @@ export class AnthropicProvider extends BaseProvider {
           "gen_ai.usage.output_tokens",
           usage.completionTokens || 0,
         );
+        if (usage.cacheReadTokens) {
+          streamSpan.setAttribute(
+            "gen_ai.usage.cached_input_tokens",
+            usage.cacheReadTokens,
+          );
+        }
         const cost = calculateCost(this.providerName, this.modelName, {
           input: usage.promptTokens || 0,
           output: usage.completionTokens || 0,
           total: usage.totalTokens || 0,
+          ...(usage.cacheReadTokens
+            ? { cacheReadTokens: usage.cacheReadTokens }
+            : {}),
+          ...(usage.cacheCreationTokens
+            ? { cacheCreationTokens: usage.cacheCreationTokens }
+            : {}),
         });
         if (cost && cost > 0) {
           streamSpan.setAttribute("neurolink.cost", cost);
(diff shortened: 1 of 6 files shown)


src/lib/providers/anthropic.ts
```diff
@@ -1773,12 +1809,30 @@ export class AnthropicProvider extends BaseProvider {
       const conversation = payload.messages.slice();
       let totalInput = 0;
       let totalOutput = 0;
+      let totalCacheRead = 0;
```
Tara-ag, line 1812: 💡 The comment claims `conversation` itself is never mutated, but the loop later calls `conversation.push(...)` twice per tool step. The *breakpoint helper* is pure (it clones the array), but the statement about `conversation` is misleading. Consider rephrasing to clarify that the helper’s purity is what keeps re-counting stable, e.g.:

```ts
// Pure: applyAnthropicHistoryCacheBreakpoints clones `conversation`,
// so re-counting the original array at the top of each step stays stable.
```

src/lib/providers/openaiChatCompletionsClient.ts
```diff
@@ -639,16 +640,8 @@ export const buildAPIError = async (
...
-    promptTokens: number;
-    completionTokens: number;
-    totalTokens: number;
-  }) => void = () => {};
-  const usagePromise = new Promise<{
-    promptTokens: number;
-    completionTokens: number;
-    totalTokens: number;
-  }>((r) => {
+  let resolveUsage: (u: DeferredUsage) => void = () => {};
```
Tara-ag, line 643: 💡 The new `DeferredUsage` type is a good consolidation. Just confirming: `createDeferredAnalytics` is also consumed by `openaiChatCompletionsBase.ts` (via this module). The base class currently resolves usage with only `promptTokens/completionTokens/totalTokens`. Since `DeferredUsage` now carries optional cache fields, consider whether `openaiChatCompletionsBase.runStreamLoop` should forward cache values when providers populate them. Not blocking for this PR, but worth a follow-up so OpenAI-compatible cache reads aren't dropped at the resolve boundary.

Tara-ag (approved):
## Review Summary

**Files reviewed:** 6 changed files, processed one at a time.
**New issues raised:** 2 (both MINOR/SUGGESTION, non-blocking).

### Findings

1. **`src/lib/providers/anthropic.ts` (line ~1812)** — 💡 The inline comment claims `conversation` itself is never mutated, but the loop later `push`es assistant/tool turns into it. The breakpoint helper is indeed pure (it clones), so re-counting stays stable; the wording just needs to credit the helper rather than the mutable array.

2. **`src/lib/providers/openaiChatCompletionsClient.ts` (line ~643)** — 💡 Consolidating the deferred-usage shape into the new `DeferredUsage` type is good. Since `openaiChatCompletionsBase.ts` consumes `createDeferredAnalytics` and currently resolves only the three required fields, cache reads/writes can still be dropped for OpenAI-compatible providers. Suggest a follow-up to plumb cache fields through `runStreamLoop` when providers populate them.

### Blocking assessment

- No security vulnerabilities (injection, SSRF, unsafe eval, etc.).
- No breaking changes to the public SDK API (`DeferredUsage` and the new optional `RawUsageObject` fields are additive).
- No violations of non-negotiable CLAUDE.md Critical Rules enforced outside CI (dynamic imports, Gemini/tools+schema mutual exclusion, CLI/SDK separation, `formatProviderError` return contract, backward compatibility).
- The Anthropic cache-marker budget logic correctly respects Anthropic’s four-marker cap and is idempotent across multi-step loops.
- Cache token accounting in both `doGenerate` and `executeStream` matches Anthropic’s separate cache-read/cache-write reporting and feeds `calculateCost` correctly.

**Decision:** Approve. The two comments are minor clarity/follow-up notes and do not block merge.

PR #1142, opened 2026-07-10
feat(skills): progressive disclosure with session-pinned activation

pdogra1299:
## Summary

Rewrites the native skills subsystem around the Agent Skills progressive-disclosure architecture (the Claude Code pattern), replacing the v1 "ALWAYS call search_skills" design that forced a blind tool round-trip on every turn.

**Discovery (Level 1)** — a name-sorted, byte-stable `<available_skills>` listing rides inside the `use_skill` tool description (default), the system prompt, or nowhere (`skills.discovery: "tool" | "system-prompt" | "none"`). Budgeted via `listingBudgetChars` with uniform description shortening — names are never dropped, and the render is a pure function of the index so provider prompt caches stay warm.

**Activation (Level 2)** — `use_skill(name)` returns the full instructions, never truncated. With a `sessionId` + conversation memory, the activation is **pinned to the session**: the instructions persist into history (ask → `[Skill loaded: …]` → answer), replay byte-identically on later turns (cached-rate tokens), and re-invocations return a tiny `already_loaded` note. Sessions pin the version they activated. Activation state is derived from stored history on every check, so dedup is truthful across restarts, multi-instance Redis deployments, and failed persistence (self-heals by re-activating).

**Resources (Level 3)** — skills bundle auxiliary files (`references/…`) costing zero tokens until read via `read_skill_resource` (gated on activation, symlink-safe path containment). Supported in filesystem (`<name>/SKILL.md` siblings), S3 (`<prefix>skills/<id>/<path>`), Redis, and memory stores.

Pinned skills survive the whole compaction pipeline: memory summarization replay, compactor Stage-3 LLM summarization (re-seated after the summary), and Stage-4 sliding-window truncation (skills partitioned out of the pair arithmetic and re-anchored, keeping role alternation intact).

Also: `skills.preload` for host-driven activation, per-call scope/tags filtering, S3 index ETag-conditional refresh (unchanged index costs a 304), activation span attributes (`skill.name/version/instructions_chars/pinned`), and caller-supplied per-call tools win over the injected ones.

**Breaking (intentional, subsystem shipped in 9.85.0 with no known adopters):** `search_skills` tool and `promptIndex` config removed; replaced by `use_skill` + `discovery`.

## Review & testing

- Multi-agent adversarial code review (33 agents) — all 8 confirmed correctness findings fixed (Redis userId-scoped hydration, persistence-gated dedup, Stage-3/Stage-4 skill protection, role-alternation-safe truncation, symlink escape, S3 ETag-before-validation, caller-tool clobbering, cache-busting listing shortening)
- `pnpm run test:skills`: 44 passed, 0 failed (activation/pinning/preload/version-pinning/resources/budget/discovery + stores/mutations/CLI/server)
- Memory suite: 21/23 passed, 1 env-dependent flake (upstream Vertex failure mid-suite; model probe passes standalone), 1 skip
- `check`, `lint`, full `build` green

## Summary by CodeRabbit

* **New Features**
  * Redesigned native skills with progressive disclosure, session-pinned activation, and selectable discovery modes (tool, system-prompt, or disabled).
  * Added per-call `use_skill` and `read_skill_resource` tools with scoping, optional preload activation, and activation deduping.
  * Added safe skill resource support across backends, including efficient index refresh behavior.

* **Bug Fixes**
  * Preserved pinned skill instructions through summarization and sliding-window truncation.
  * Corrected session turn ordering and statistics to exclude pinned skill messages.

* **Documentation**
  * Updated the native skills guide, configuration/API reference, and examples; fixed related documentation links.

* **Tests**
  * Updated continuous and live skills tests for the new discovery/tool-based flow and resource handling.

### DIFF
--- a/docs/features/skills.md
+++ b/docs/features/skills.md
@@ -1,15 +1,17 @@
 ---
 title: Skills Guide
-description: Native skills — versioned, discoverable instruction packs (SOPs, playbooks) with progressive disclosure via built-in tools and a system-prompt index
+description: Native skills — versioned instruction packs (SOPs, playbooks) with progressive disclosure, session-pinned activation, and on-demand resources
 keywords:
   [
     skills,
     sops,
     playbooks,
     progressive-disclosure,
-    search_skills,
+    use_skill,
+    read_skill_resource,
     list_skills,
-    skill-tools,
+    session-persistence,
+    prompt-caching,
     maker-checker,
     custom-storage,
   ]
@@ -17,16 +19,17 @@ keywords:
 
 # Skills Guide
 
-> **Since**: v9.82.0 | **Status**: New | **Availability**: SDK
+> **Since**: v9.82.0 | **Status**: Stable | **Availability**: SDK, CLI, Server
 
 ## Overview
 
-NeuroLink includes native **skills** support: versioned, discoverable instruction packs (SOPs, playbooks, workflows) that agents consult before answering from general knowledge. Skills use **progressive disclosure** to stay cheap:
+NeuroLink includes native **skills** support: versioned, discoverable instruction packs (SOPs, playbooks, workflows) the model consults before answering from general knowledge. Skills follow the [Agent Skills](https://agentskills.io) progressive-disclosure architecture — three levels, each loaded only when needed:
 
-- A compact **index** (name + description, never instructions) is injected into the system prompt of each `generate()`/`stream()` call, so the model knows what exists.
-- Full **instructions** are loaded on demand through the built-in `search_skills` tool — only for skills that actually match the user's request.
+1. **Discovery** — a compact listing (name + description, never instructions) is embedded in the `use_skill` tool description on every `generate()`/`stream()` call. The model decides from context when a skill applies; there is no mandatory pre-flight search call.
+2. **Activation** — when a skill matches the task, the model calls `use_skill(name)` and receives the **full instructions, never truncated**. The activation is **pinned to the session**: the instructions persist in conversation history, replayed byte-identically on every later turn (provider prompt caches bill them at cached rates), and re-invocations return a tiny `already_loaded` note instead of the body.
+3. **Resources** — skills can bundle auxiliary files (`references/*.md`, templates, schemas). They cost zero tokens until the model reads one with `read_skill_resource`.
 
-This is the same architecture proven in production by curator's skills system, generalized and moved into the SDK: enable it with one config block instead of building stores, matchers, and tools in app code.
+A 20-skill catalog costs roughly 500–700 tokens of always-on listing; an activated 4K-token SOP is fetched once per session and cached thereafter.
 
 ## Quick Start
 
@@ -40,207 +43,212 @@ const neurolink = new NeuroLink({
   },
 });
 
-// The model now sees the skills index in its system prompt and can call
-// search_skills / list_skills automatically during any generate/stream.
+// The model sees the skills listing inside the use_skill tool description
+// and loads instructions only when a skill matches the task.
 const result = await neurolink.generate({
   input: { text: "A refund is disputed — how do I escalate?" },
+  context: { sessionId: "thread-42", userId: "u1" },
 });
 ```
 
+With a `sessionId` and conversation memory configured, an activated skill stays loaded for the whole session — later turns replay it from history instead of re-fetching it. Without conversation memory, activation is per-turn (nothing is pinned, and `use_skill` simply re-fetches when asked again).
+
 ## Skill Format
 
-Three on-disk layouts are supported by the filesystem store (mixable in one directory):
+### Directory layout (recommended)
 
-**JSON** (`./skills/<id>.json`):
-
-```json
-{
-  "id": "refund-escalation",
-  "name": "refund_dispute_escalation",
-  "displayName": "Refund Dispute Escalation",
-  "description": "How to escalate a disputed refund to the payments on-call team.",
-  "instructions": "1. Collect the transaction id.\n2. Verify the dispute.\n3. Page payments-oncall.",
-  "tags": ["payments", "escalation"]
-}
+```
+skills/
+  refund_dispute_escalation/
+    SKILL.md               # frontmatter + instructions
+    references/
+      edge-cases.md        # loaded on demand via read_skill_resource
+      escalation-matrix.md
 ```
 
-**Frontmatter markdown** (`./skills/<name>.md`):
+`SKILL.md` starts with YAML frontmatter; the markdown body is the instructions:
 
 ```markdown
 ---
-name: oncall_handover
-description: Checklist for handing over the on-call shift.
-tags: [devops, oncall]
+name: refund_dispute_escalation
+displayName: Refund Dispute Escalation
+description: How to escalate a disputed refund to the payments on-call. Use when a customer contests a refund decision.
+tags: [payments, escalation]
 ---
 
-1. Summarize open incidents.
-2. Hand over the pager.
+1. Verify the dispute in the payments dashboard…
+2. …
 ```
 
-**Claude-skills directory layout** (`./skills/<name>/SKILL.md`) — same frontmatter format, interoperable with Claude-style skill directories.
+Every sibling file of `SKILL.md` becomes an on-demand resource, addressable by its relative path (e.g. `references/edge-cases.md`).
+
+### Other accepted layouts
+
+- `<dir>/<name>.md` — single frontmatter markdown file (no resources)
+- `<dir>/<id>.json` — JSON-serialized `SkillDefinition` (what mutations write)
 
 Optional fields: `scope: "scoped"` + `scopeIds: [...]` restrict a skill to specific scopes (channels/teams/tenants); `status: "deprecated"` hides it from matching.
 
-## Storage Backends
+### Authoring guidance
 
-| Type         | Config                                    | Use case                                     |
-| ------------ | ----------------------------------------- | -------------------------------------------- |
-| `memory`     | `{ type: "memory", skills?: [...] }`      | Tests, embedded, host-managed                |
-| `filesystem` | `{ type: "filesystem", path: "./dir" }`   | Local dev, git-versioned skill repos         |
-| `s3`         | `{ type: "s3", bucket, prefix? }`         | Shared team skills (curator-style)           |
-| `redis`      | `{ type: "redis", keyPrefix? }`           | Low-latency shared store, reuses core client |
-| `custom`     | `{ type: "custom", store: mySkillStore }` | Anything else — implement `SkillStore`       |
+- **Description is the matching signal.** Say _what_ the skill does and _when_ to use it, in one or two sentences. The model selects skills purely from descriptions.
+- **Keep instructions lean** (guideline: under ~5K tokens). Move rarely-needed detail into `references/` — if information is needed 20% of the time, it belongs in a resource file.
+- **Instructions are never truncated or capped** — budgets apply only to the discovery listing, which shortens descriptions uniformly (never drops names) when the catalog outgrows `listingBudgetChars`.
 
-A custom store implements four methods: `get(id)`, `put(skill)`, `delete(id)`, `index()` (index entries must be cheap — they back every search).
+## How Activation Works
 
-### S3
+```
+Turn 1: user asks → model sees listing → use_skill("deploy_sop")
+        → full instructions returned (tool result)
+        → pinned into session history: ask → [Skill loaded: deploy_sop v3] → answer
+
+Turn 2+: history replays the pinned instructions byte-identically
+         (cached prefix for Anthropic cache_control / Gemini implicit caching);
+         use_skill("deploy_sop") again → { status: "already_loaded" }
+```
+
+Guarantees, in order of the pipeline:
+
+- **Version pinning** — a session keeps the version it activated; a mid-session skill update never mutates instructions the model already follows. New sessions get the new version.
+- **Summarization-safe** — when conversation memory summarizes old turns, pinned skill messages are re-included verbatim after the summary.
+- **Truncation-safe** — the context compactor's sliding-window stage re-seats pinned skill messages instead of dropping them, and never content-truncates them.
+- **Restart-safe** — activation state is derived from the stored history itself (messages carry `metadata.isSkill`), so dedup works across process restarts and multiple instances sharing Redis memory.
+
+Without a `sessionId` (or with `sessionPersistence: false`), activation is per-turn: the instructions still arrive as a tool result for the current turn; nothing is pinned.
+
+## Discovery Modes
+
+| Mode               | Where the listing lives                                            | When to use                                                                                                     |
+| ------------------ | ------------------------------------------------------------------ | --------------------------------------------------------------------------------------------------------------- |
+| `"tool"` (default) | `<available_skills>` block inside the `use_skill` tool description | Best default — keeps the host's system prompt untouched and the listing byte-stable for provider prompt caching |
+| `"system-prompt"`  | `## Available Skills` index appended to the system prompt          | When you want the listing visible in prompt dumps/debugging                                                     |
+| `"none"`           | Nowhere                                                            | Hosts that inject their own discovery or rely on `preload`                                                      |
+
+The listing is rendered from a name-sorted index and is a pure function of it, so calls with the same scope/tag filters render byte-identical listings — a prerequisite for Anthropic `cache_control` prefix caching and Gemini implicit caching. When the catalog exceeds `listingBudgetChars` (default 15000), every description is shortened uniformly (first sentence, then a hard cap); names are never dropped.
+
+## Per-Call Options
 
 ```typescript
-skills: {
-  enabled: true,
-  storage: {
-    type: "s3",
-    bucket: "team-skills",
-    prefix: "tara/",          // default "neurolink-skills/"
-    region: "us-east-1",
-    // endpoint + forcePathStyle for MinIO/LocalStack;
-    // credentials default to the standard AWS provider chain
+await neurolink.generate({
+  input: { text: "…" },
+  context: { sessionId: "thread-42" },
+  skills: {
+    enabled: true, // master toggle for this call
+    discovery: "tool", // override the instance discovery mode
+    scopeId: "channel-123", // include skills scoped to this id
+    tags: ["payments"], // restrict the listing to these tags
+    preload: ["deploy_sop"], // activate up front, no model round-trip
   },
-}
+});
 ```
 
-Layout: `<prefix>skills/<id>.json` per skill plus a `<prefix>index.json` that is upserted on writes and **rebuilt automatically from a bucket listing** when missing or corrupt. The AWS SDK is an optional peer — install `@aws-sdk/client-s3` in your app (NeuroLink core does not depend on it; without it, S3-configured skills fail open with an actionable error).
-
-### Redis
-
-```typescript
-skills: {
-  enabled: true,
-  storage: { type: "redis", keyPrefix: "neurolink:skills:" },
-  // url / host / port / password / db — defaults match the core Redis client
-}
-```
-
-Uses NeuroLink's pooled Redis client (`redis` v5, already a core dependency — shared with Redis conversation memory). One JSON value per skill; the index is derived with SCAN + MGET. Skills are persistent — no TTL.
+`preload` activates skills before the model runs: instructions are injected into the call's system prompt and pinned to the session like a normal activation. Use it when the host already knows which skill applies (e.g. a channel bound to a runbook).
 
 ## Built-in Tools
 
-When enabled, these tools are auto-registered and reach the model on every call:
+| Tool                                             | Injected                         | Purpose                                                                              |
+| ------------------------------------------------ | -------------------------------- | ------------------------------------------------------------------------------------ |
+| `use_skill`                                      | per call                         | Load a skill's full instructions by exact name; returns `already_loaded` when active |
+| `read_skill_resource`                            | per call                         | Read a bundled resource file of a loaded skill                                       |
+| `list_skills`                                    | registered                       | Lightweight catalog for "what can you do?" questions                                 |
+| `skill_create` / `skill_update` / `skill_delete` | registered when `allowMutations` | Model-proposed mutations, gated by `onMutationRequest`                               |
 
-- **`search_skills`** — index-first search by `query` and/or `tag`; hydrates at most `maxMatches` (default 5) matching skills with full instructions. A zero-match result returns `{ skills: [], reason: "no_match" }` as a **success**, telling the model to fall back to general knowledge.
-- **`list_skills`** — lightweight discovery listing (no instructions), for "what can you do?" questions.
+## Storage Backends
 
-With `allowMutations: true`, three more tools are registered:
+```typescript
+// In-process (tests, embedded)
+storage: { type: "memory", skills: [...], resources: { "skill-id": { "references/x.md": "…" } } }
 
-- **`skill_create`**, **`skill_update`**, **`skill_delete`** — propose changes, routed through your `onMutationRequest` gate when configured. Deletes are soft (status becomes `deprecated`).
+// Filesystem (CLI, local development)
+storage: { type: "filesystem", path: "./skills" }
 
-## Approval Gate (Maker-Checker)
+// S3 (production, curator-compatible layout)
+storage: { type: "s3", bucket: "my-bucket", prefix: "team-skills/" }
 
-Hosts keep full control over writes via `onMutationRequest`:
+// Redis (shared store on the pooled core client, no TTL)
+storage: { type: "redis", url: process.env.REDIS_URL }
+
+// Anything else
+storage: { type: "custom", store: mySkillStore }
+```
+
+Resource layouts per backend:
+
+- **Filesystem** — sibling files of `SKILL.md` (see above)
+- **S3** — `<prefix>skills/<id>.json` for the skill, `<prefix>skills/<id>/<path>` for resources; `<prefix>index.json` is self-healing and refreshed with ETag-conditional reads (an unchanged index costs a 304, not a download)
+- **Redis** — `<keyPrefix><id>` for skills, `<keyPrefix>__resources__:<id>:<path>` for resources (the `__resources__:` segment is reserved and excluded from the index scan)
+- **Custom** — implement the optional `getResource(id, path)` on your `SkillStore`
+
+S3 requires the optional peer `@aws-sdk/client-s3`. A custom store implements `get(id)`, `put(skill)`, `delete(id)`, `index()` (index entries must be cheap — they back every listing), plus optionally `getResource` and `invalidate`.
+
+## Mutations & Approval Gate
 
 ```typescript
 const neurolink = new NeuroLink({
   skills: {
     enabled: true,
-    storage: { type: "custom", store: s3SkillsStore },
+    storage: { type: "s3", bucket: "team-skills" },
     allowMutations: true,
     onMutationRequest: async (action) => {
-      // e.g. post a Slack approval block, persist a pending action…
-      const ticket = await queueForApproval(action);
-      return { outcome: "pending", reference: ticket.id };
-      // or { outcome: "approved" } to apply immediately
-      // or { outcome: "rejected", reason: "…" } to block
+      const ticket = await sendToSlackApproval(action);
+      return { outcome: "pending", reference: ticket };
     },
   },
 });
 ```
 
-`"pending"` means the host applies the change itself after human approval — NeuroLink writes nothing.
+- Reads **fail open** (errors → empty results + warn log); writes **fail closed**.
+- Deletes are soft — deprecated skills stop matching but stay in storage for audit.
+- Updates bump `version`; active sessions keep the version they loaded.
 
-## Configuration Reference
+## Observability
 
-```typescript
-skills: {
-  enabled: true,
-  storage: { type: "filesystem", path: "./skills" },
-  promptIndex: true,        // inject index into system prompt (default true)
-  promptIndexMaxItems: 50,  // truncate the injected index
-  maxMatches: 5,            // max skills hydrated per search
-  indexCacheTtlMs: 30_000,  // index cache TTL (0 = no caching)
-  defaultScopeId: undefined, // default scope filter
-  allowMutations: false,    // register skill_create/update/delete
-  onMutationRequest: undefined, // approval gate
-}
-```
-
-### Per-call options
-
-```typescript
-await neurolink.generate({
-  input: { text: "…" },
-  skills: {
-    enabled: true, // per-call master toggle (prompt index)
-    promptIndex: false, // disable injection for this call only
-    scopeId: "team-alpha", // scope the index for this call
-    tags: ["payments"], // narrow the index by tags
-  },
-});
-```
-
-### Programmatic access
-
-```typescript
-const manager = neurolink.getSkillsManager();
-const matches = await manager?.search({ query: "refund" });
-const index = await manager?.list();
-await manager?.requestMutation({ type: "create", skill: { … } });
-```
+Every activation stamps the active span with `skill.name`, `skill.version`, `skill.instructions_chars`, and `skill.pinned`, so traces show exactly which skills a turn loaded and what they cost.
 
 ## CLI
 
-Manage a filesystem skills store directly:
-
 ```bash
-neurolink skills list   --skills-dir ./skills
-neurolink skills show   refund_dispute_escalation --skills-dir ./skills
-neurolink skills search "refund" --skills-dir ./skills
-neurolink skills create --skills-dir ./skills \
-  --name deploy_sop --description "How to deploy" --instructions-file ./sop.md
-neurolink skills delete deploy_sop --skills-dir ./skills
+neurolink skills list --skills-dir ./skills
+neurolink skills show refund_dispute_escalation
+neurolink skills search "refund" --tag payments
+neurolink skills create --name deploy_sop --description "How to deploy" --instructions-file ./sop.md
+neurolink skills delete deploy_sop
+
+# Make skills available to a run
+neurolink generate "how do I deploy?" --skills-dir ./skills
 ```
 
-Make skills available to any generation run with the same flag (or `NEUROLINK_SKILLS_DIR`):
+`NEUROLINK_SKILLS_DIR` works as an alternative to `--skills-dir`.
 
-```bash
-neurolink generate "A refund is disputed — how do I escalate?" --skills-dir ./skills
-neurolink loop --skills-dir ./skills
-```
+## Server
 
-Directory precedence: `--skills-dir` > `NEUROLINK_SKILLS_DIR` > `./skills` (for the `skills` command group).
+`/api/agent/skills` exposes CRUD routes when the server's NeuroLink instance has skills configured (503 `SKILLS_UNAVAILABLE` otherwise); mutation routes additionally require `allowMutations: true`.
 
-## Server Endpoints
+## Config Reference
 
-When the server's NeuroLink instance has skills enabled, these routes are available:
+| Key                   | Default              | Purpose                                                                |
+| --------------------- | -------------------- | ---------------------------------------------------------------------- |
+| `enabled`             | —                    | Master switch (required)                                               |
+| `storage`             | `{ type: "memory" }` | Persistence backend                                                    |
+| `discovery`           | `"tool"`             | Where the listing surfaces (`"tool"` \| `"system-prompt"` \| `"none"`) |
+| `listingBudgetChars`  | `15000`              | Character budget for the `"tool"` listing                              |
+| `sessionPersistence`  | `true`               | Pin activated instructions into session history                        |
+| `maxMatches`          | `5`                  | Max skills hydrated per programmatic `search()`                        |
+| `promptIndexMaxItems` | `50`                 | Max entries in the `"system-prompt"` index                             |
+| `indexCacheTtlMs`     | `30000`              | Index cache TTL (0 disables)                                           |
+| `defaultScopeId`      | —                    | Scope filter applied when a call provides none                         |
+| `allowMutations`      | `false`              | Register the `skill_*` mutation tools                                  |
+| `onMutationRequest`   | —                    | Host approval gate for mutations                                       |
 
-| Method   | Path                    | Purpose                                   |
-| -------- | ----------------------- | ----------------------------------------- |
-| `GET`    | `/api/agent/skills`     | List index entries (optional `?scopeId=`) |
-| `GET`    | `/api/agent/skills/:id` | One skill with instructions (id or name)  |
-| `POST`   | `/api/agent/skills`     | Create (routed through the mutation gate) |
-| `PATCH`  | `/api/agent/skills/:id` | Update (patch semantics, version bump)    |
-| `DELETE` | `/api/agent/skills/:id` | Soft-delete (deprecate)                   |
+## Programmatic Access
 
-Mutation endpoints honor `onMutationRequest` — a `pending` decision returns `{ decision: { outcome: "pending", reference } }` without writing. When skills are not configured the routes return a 503 `SKILLS_UNAVAILABLE` error envelope. Authenticated user identity (when auth middleware is active) overrides caller-supplied `requestedBy`.
-
-## Behavior Notes
-
-- **Opt-in and fail-open**: with no `skills` config, nothing changes. Store/read errors log a warning and degrade to "no skills" — they never fail a generate/stream call. Mutations fail closed.
-- **Media modes**: the prompt index is skipped for avatar/music/video/ppt outputs (no meaningful text prompt to augment).
-- **Tool routing**: skill tools are registered as custom (direct) tools, so server-granularity tool routing never drops them.
-- Related: [Memory Guide](/docs/features/memory) for the analogous per-user memory subsystem, and [Conversation History](/docs/features/conversation-history).
-
-## Testing
-
-```bash
-pnpm run test:skills   # test/continuous-test-suite-skills.ts (mostly no-API; live test skips without credentials)
+```typescript
+const manager = neurolink.getSkillsManager();
+const all = await manager?.list();
+const matches = await manager?.search({ query: "refund", limit: 3 });
+const skill = await manager?.get("refund_dispute_escalation");
+const resource = await manager?.getResource(
+  skill!.id,
+  "references/edge-cases.md",
+);
 ```
--- a/src/lib/context/contextCompactor.ts
+++ b/src/lib/context/contextCompactor.ts
@@ -40,7 +40,7 @@ const DEFAULT_CONFIG: Required<CompactionConfig> = {
   enableTruncate: true,
   pruneProtectTokens: 40_000,
   pruneMinimumSavings: 500,
-  pruneProtectedTools: ["skill"],
+  pruneProtectedTools: ["skill", "use_skill", "read_skill_resource"],
   summarizationProvider: "vertex",
   summarizationModel: "gemini-2.5-flash",
   keepRecentRatio: 0.3,
@@ -180,7 +180,25 @@ export class ContextCompactor {
                 "LLM summarization timed out after 120s",
               );
               if (summarizeResult.summarized) {
-                currentMessages = summarizeResult.messages;
+                // Pinned skill instructions must survive summarization:
+                // re-seat any that the summarized region swallowed right
+                // after the summary message (index 0 of the result).
+                const survivorIds = new Set(
+                  summarizeResult.messages.map((m) => m.id),
+                );
+                const swallowedSkills = currentMessages.filter(
+                  (m) => m.metadata?.isSkill && !survivorIds.has(m.id),
+                );
+                currentMessages =
+                  swallowedSkills.length > 0
+                    ? summarizeResult.messages.length > 0
+                      ? [
+                          summarizeResult.messages[0],
+                          ...swallowedSkills,
+                          ...summarizeResult.messages.slice(1),
+                        ]
+                      : swallowedSkills
+                    : summarizeResult.messages;
                 stagesUsed.push("summarize");
               }
               const stageTokensAfter = estimateMessagesTokens(
--- a/src/lib/context/stages/slidingWindowTruncator.ts
+++ b/src/lib/context/stages/slidingWindowTruncator.ts
@@ -88,8 +88,12 @@ function truncateSmallConversation(
 
   for (let i = 0; i < result.length; i++) {
     const msg = result[i];
-    // Don't truncate system/summary messages
-    if (msg.role === "system" || msg.metadata?.isSummary) {
+    // Don't truncate system/summary messages or pinned skill instructions
+    if (
+      msg.role === "system" ||
+      msg.metadata?.isSummary ||
+      msg.metadata?.isSkill
+    ) {
       continue;
     }
 
@@ -133,8 +137,39 @@ export function truncateWithSlidingWindow(
   messages: ChatMessage[],
   config?: TruncationConfig,
 ): TruncationResult {
-  if (messages.length <= 4) {
-    // Delegate to content truncation for small conversations (BUG-005)
+  // Partition pinned skill messages out so the pair-based arithmetic below
+  // runs on the alternating user/assistant stream (skill pins are extra
+  // user-role messages inside a turn and would misalign every even-count
+  // cut). Each skill anchors to the next conversational message; kept
+  // anchors put their skills back in place, dropped anchors re-seat them
+  // after the truncation marker.
+  const skillsByAnchor = new Map<string, ChatMessage[]>();
+  const trailingSkills: ChatMessage[] = [];
+  const conversational: ChatMessage[] = [];
+  for (let i = 0; i < messages.length; i++) {
+    const msg = messages[i];
+    if (!msg.metadata?.isSkill) {
+      conversational.push(msg);
+      continue;
+    }
+    const anchor = messages
+      .slice(i + 1)
+      .find((next) => !next.metadata?.isSkill);
+    if (anchor) {
+      const anchored = skillsByAnchor.get(anchor.id);
+      if (anchored) {
+        anchored.push(msg);
+      } else {
+        skillsByAnchor.set(anchor.id, [msg]);
+      }
+    } else {
+      trailingSkills.push(msg);
+    }
+  }
+
+  if (conversational.length <= 4) {
+    // Delegate to content truncation for small conversations (BUG-005);
+    // it never removes messages, so skills keep their positions.
     return truncateSmallConversation(messages, config);
   }
 
@@ -163,8 +198,49 @@ export function truncateWithSlidingWindow(
   }
 
   // Always preserve first user-assistant pair
-  const firstPair = messages.slice(0, 2);
-  const remainingMessages = messages.slice(2);
+  const firstPair = conversational.slice(0, 2);
+  const remainingMessages = conversational.slice(2);
+
+  /** Reassemble a candidate: skills inline before kept anchors, orphans after the marker. */
+  const buildCandidate = (
+    keptAfterTruncation: ChatMessage[],
+    marker: ChatMessage,
+  ): ChatMessage[] => {
+    const keptIds = new Set(
+      [...firstPair, ...keptAfterTruncation].map((m) => m.id),
+    );
+    const orphanedSkills: ChatMessage[] = [];
+    for (const [anchorId, anchored] of skillsByAnchor) {
+      if (!keptIds.has(anchorId)) {
+        orphanedSkills.push(...anchored);
+      }
+    }
+    const withInlineSkills = (msg: ChatMessage): ChatMessage[] => {
+      const anchored = skillsByAnchor.get(msg.id);
+      return anchored ? [...anchored, msg] : [msg];
+    };
+    return [
+      ...firstPair.flatMap(withInlineSkills),
+      marker,
+      ...orphanedSkills,
+      ...keptAfterTruncation.flatMap(withInlineSkills),
+      ...trailingSkills,
+    ];
+  };
+
+  // Insert a truncation marker with machine-readable metadata so
+  // effectiveHistory.ts can detect it via isTruncationMarker /
+  // truncationId and removeTruncationTags can rewind it.
+  const makeMarker = (): ChatMessage => {
+    const truncId = randomUUID();
+    return {
+      id: `truncation-marker-${truncId}`,
+      role: "user",
+      content: TRUNCATION_MARKER_CONTENT,
+      isTruncationMarker: true,
+      truncationId: truncId,
+    };
+  };
 
   // ITERATIVE: if first pass isn't enough, increase fraction
   const maxIterations = config?.maxIterations ?? 3;
@@ -178,21 +254,10 @@ export function truncateWithSlidingWindow(
       break;
     }
 
-    const keptAfterTruncation = remainingMessages.slice(evenRemoveCount);
-
-    // Insert a truncation marker with machine-readable metadata so
-    // effectiveHistory.ts can detect it via isTruncationMarker /
-    // truncationId and removeTruncationTags can rewind it.
-    const truncId = randomUUID();
-    const marker: ChatMessage = {
-      id: `truncation-marker-${truncId}`,
-      role: "user",
-      content: TRUNCATION_MARKER_CONTENT,
-      isTruncationMarker: true,
-      truncationId: truncId,
-    };
-
-    const candidateMessages = [...firstPair, marker, ...keptAfterTruncation];
+    const candidateMessages = buildCandidate(
+      remainingMessages.slice(evenRemoveCount),
+      makeMarker(),
+    );
 
     validateRoleAlternation(candidateMessages);
 
@@ -226,19 +291,10 @@ export function truncateWithSlidingWindow(
   const maxRemove = Math.floor(remainingMessages.length * 0.95);
   const evenMaxRemove = maxRemove - (maxRemove % 2);
   if (evenMaxRemove > 0) {
-    const keptMessages = remainingMessages.slice(evenMaxRemove);
-
-    // Insert a truncation marker (see iterative block above)
-    const fallbackTruncId = randomUUID();
-    const fallbackMarker: ChatMessage = {
-      id: `truncation-marker-${fallbackTruncId}`,
-      role: "user",
-      content: TRUNCATION_MARKER_CONTENT,
-      isTruncationMarker: true,
-      truncationId: fallbackTruncId,
-    };
-
-    const fallbackMessages = [...firstPair, fallbackMarker, ...keptMessages];
+    const fallbackMessages = buildCandidate(
+      remainingMessages.slice(evenMaxRemove),
+      makeMarker(),
+    );
     validateRoleAlternation(fallbackMessages);
 
     return {
--- a/src/lib/core/conversationMemoryManager.ts
+++ b/src/lib/core/conversationMemoryManager.ts
@@ -138,7 +138,14 @@ export class ConversationMemoryManager implements IConversationMemoryManager {
             assistantMsg.events = options.events;
           }
 
-          session.messages.push(userMsg, assistantMsg);
+          session.messages.push(userMsg);
+          // Pinned skill activations ride between ask and answer, mirroring
+          // the actual order (ask → skill loaded → answer). Stored verbatim —
+          // skill instructions are never truncated.
+          if (options.skillMessages && options.skillMessages.length > 0) {
+            session.messages.push(...options.skillMessages);
+          }
+          session.messages.push(assistantMsg);
           session.lastActivity = Date.now();
 
           // Store API-reported token counts if available
@@ -399,8 +406,13 @@ export class ConversationMemoryManager implements IConversationMemoryManager {
     await this.ensureInitialized();
 
     const sessions = Array.from(this.sessions.values());
+    // Pinned skill messages are extra rows inside a turn — exclude them so
+    // a skill-activating turn still counts as one turn.
     const totalTurns = sessions.reduce(
-      (sum, session) => sum + session.messages.length / MESSAGES_PER_TURN,
+      (sum, session) =>
+        sum +
+        session.messages.filter((msg) => !msg.metadata?.isSkill).length /
+          MESSAGES_PER_TURN,
       0,
     );
 
--- a/src/lib/core/redisConversationMemoryManager.ts
+++ b/src/lib/core/redisConversationMemoryManager.ts
@@ -683,6 +683,13 @@ export class RedisConversationMemoryManager implements IConversationMemoryManage
             normalizedUserId,
           );
 
+          // Pinned skill activations ride between ask and answer, mirroring
+          // the actual order (ask → skill loaded → answer). Stored verbatim —
+          // skill instructions are never truncated.
+          if (options.skillMessages && options.skillMessages.length > 0) {
+            conversation.messages.push(...options.skillMessages);
+          }
+
           const assistantMsg: ChatMessage = {
             id: randomUUID(),
             timestamp: this.generateTimestamp(),
@@ -1562,7 +1569,12 @@ User message: "${userMessage}"`;
       const conversationData = await this.redisClient.get(key);
       const conversation = deserializeConversation(conversationData);
       if (conversation?.messages) {
-        totalTurns += conversation.messages.length / MESSAGES_PER_TURN;
+        // Pinned skill messages are extra rows inside a turn — exclude them
+        // so a skill-activating turn still counts as one turn.
+        totalTurns +=
+          conversation.messages.filter(
+            (msg: ChatMessage) => !msg.metadata?.isSkill,
+          ).length / MESSAGES_PER_TURN;
       }
     }
 
--- a/src/lib/neurolink.ts
+++ b/src/lib/neurolink.ts
@@ -173,8 +173,10 @@ import type {
 } from "./types/index.js";
 import { initializeHippocampus } from "./memory/hippocampusInitializer.js";
 import { createMemoryRetrievalTools } from "./memory/memoryRetrievalTools.js";
+import { isSkillVisibleInScope } from "./skills/skillMatcher.js";
+import { buildSkillActivationMessage } from "./skills/skillSessionTracker.js";
 import { SkillsManager } from "./skills/skillsManager.js";
-import { createSkillTools } from "./skills/skillTools.js";
+import { createSkillCallTools, createSkillTools } from "./skills/skillTools.js";
 import {
   getMetricsAggregator,
   MetricsAggregator,
@@ -1834,7 +1836,7 @@ export class NeuroLink {
   }
 
   /**
-   * Register the built-in skill tools (search_skills / list_skills, plus
+   * Register the built-in skill tools (list_skills, plus
    * mutation tools when allowMutations is set). Follows the
    * registerMemoryRetrievalTools() pattern: registered via registerTool()
    * so they land in the "user-defined" category that reaches the LLM tool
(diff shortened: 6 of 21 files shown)


docs/features/skills.md
```diff
@@ -40,207 +43,212 @@ const neurolink = new NeuroLink({
...
+    discovery: "tool", // override the instance discovery mode
+    scopeId: "channel-123", // include skills scoped to this id
+    tags: ["payments"], // restrict the listing to these tags
+    preload: ["deploy_sop"], // activate up front, no model round-trip
   },
 });
 ```
 
-`"pending"` means the host applies the change itself after human approval — NeuroLink writes nothing.
+`preload` activates skills before the model runs: instructions are injected into the call's system prompt and pinned to the session like a normal activation. Use it when the host already knows which skill applies (e.g. a channel bound to a runbook).
```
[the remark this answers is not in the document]
pdogra1299: Fixed in 2b2cd555 — claim now scoped to calls with the same scope/tag filters.

src/lib/core/conversationMemoryManager.ts
```diff
@@ -138,7 +138,14 @@ export class ConversationMemoryManager implements IConversationMemoryManager {
...
 
-          session.messages.push(userMsg, assistantMsg);
+          session.messages.push(userMsg);
+          // Pinned skill activations ride between ask and answer, mirroring
+          // the actual order (ask → skill loaded → answer). Stored verbatim —
+          // skill instructions are never truncated.
+          if (options.skillMessages && options.skillMessages.length > 0) {
+            session.messages.push(...options.skillMessages);
+          }
+          session.messages.push(assistantMsg);
```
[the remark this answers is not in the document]
pdogra1299: Fixed in 2b2cd555 — turn stats now exclude pinned skill messages in both memory managers.

src/lib/skills/skillStoreRedis.ts
```diff
@@ -26,11 +26,17 @@ const DEFAULT_KEY_PREFIX = "neurolink:skills:";
...
+   * Resources live under a sibling prefix (`…skills-resources:`) so the
+   * `<keyPrefix>*` index SCAN never picks them up as skill values.
+   */
+  private readonly resourcePrefix: string;
   private readonly normalizedConfig: ReturnType<typeof getNormalizedConfig>;
   private clientPromise: Promise<RedisClient> | null = null;
 
   constructor(config: SkillRedisStorageConfig) {
     this.keyPrefix = config.keyPrefix ?? DEFAULT_KEY_PREFIX;
+    this.resourcePrefix = `${this.keyPrefix.replace(/:$/, "")}-resources:`;
```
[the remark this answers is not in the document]
pdogra1299: Fixed in 2b2cd555 — resources moved to a reserved `<keyPrefix>__resources__:` segment and the index SCAN filters it explicitly, safe for any keyPrefix shape.

test/continuous-test-suite-skills.ts
```diff
@@ -1387,29 +1851,29 @@ async function testLiveGenerate(): Promise<void> {
...
         `toolCalled=${usedTool} followedInstructions=${followedSkill}`,
       );
     } else {
       logTest(
-        "24. Live: model calls search_skills and follows it",
+        LIVE_NAME,
         "FAIL",
         `Model neither called the tool nor followed instructions. Content: ${content.slice(0, 200)}`,
       );
     }
```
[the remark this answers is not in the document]
pdogra1299: Fixed in 2b2cd555 — the live test now requires the use_skill execution; instruction phrasing is informational only.

src/lib/neurolink.ts
```diff
@@ -1913,30 +1923,178 @@ export class NeuroLink {
...
+        options.skills?.scopeId ?? this.skillsConfig.defaultScopeId;
+      const tags = options.skills?.tags;
+      const sessionId =
+        this.resolveSkillSessionId(options.context) ??
+        this.resolveSkillSessionId(options);
+      const userId =
+        this.resolveSkillUserId(options.context) ??
+        this.resolveSkillUserId(options);
+      // Pinning requires somewhere to pin: without conversation memory the
+      // drained messages would be silently discarded while dedup reports
```
Tara-ag, line 1938: ⚠️ **MAJOR — Potential race between preload and concurrent `use_skill` activation.**

`preloadSkills` records an activation via `manager.sessions.recordActivation(...)`, which adds the message to the pending queue. If the model also calls `use_skill` for the same skill in the same turn (possible if the listing still shows the skill), the `use_skill` execute path will call `ensureHydrated` and then `isActive`. However, `ensureHydrated` reads from `conversationMemory.getSessionMessages`, which does **not** include the pending pin that has not yet been stored. So `isActive` returns false and `use_skill` will record a second pending message for the same skill, causing a duplicate pin after `storeConversationTurn`.

The tracker merges pending messages in `hydrate()`, but only if `hydrate()` is called **after** the preload recorded the pending message. In `preloadSkills` the call sequence is: `recordActivation` (adds to pending) → inject into system prompt. The `use_skill` tool may execute in parallel with preload or after, and its `ensureHydrated` call will merge pending if the pending map already has the entry. But if `use_skill` executes before preload reaches `recordActivation`, or if the pending queue is drained between preload and `use_skill`, you get duplicates.

More importantly, `preloadSkills` does not call `ensureHydrated` before `isActive`, so in the same turn a second `applySkillsAugmentation` call (e.g. a retry or separate generate in the same session) can double-pin because it checks the in-process tracker state without hydrating from the just-stored history.

**Suggested fix:** In `preloadSkills`, hydrate from stored messages before checking `isActive`, exactly as `use_skill` does, and dedupe pending pins by skill id/name when draining:

```ts
if (sessionId && sessionPersistence) {
  manager.sessions.hydrate(
    sessionId,
    this.conversationMemory
      ? await this.conversationMemory.getSessionMessages(sessionId, userId)
      : [],
  );
  if (manager.sessions.isActive(sessionId, skill.id, skill.name)) {
    continue;
  }
  block = manager.sessions.recordActivation(sessionId, skill).content;
}
```

Also consider making `drainPending` dedupe by skill id to guard against any remaining double-record paths.

src/lib/skills/skillSessionTracker.ts
```diff
@@ -0,0 +1,171 @@
...
+    } else {
+      this.pending.set(sessionId, [message]);
+    }
+    return message;
+  }
+
+  /**
+   * Rebuild the active set from stored session history, then merge the
+   * in-process pending activations (this turn's pins, not yet stored).
+   * Called before every dedup check — the rebuild keeps state truthful
```
Tara-ag, line 115: ⚠️ **MAJOR — `hydrate()` silently drops newer duplicate activations instead of preserving the most recent pin.**

```ts
if (record && !records.has(record.skillId)) {
  records.set(record.skillId, record);
}
```

This keeps the **first** occurrence of a skill pin in stored history and ignores later ones. If a skill is activated, deactivated (not implemented), and re-activated, or if the same skill appears multiple times in history due to a bug, the session will report the oldest version as active. Worse, `recordActivation` always increments the version from the current `skill.version`, but `hydrate` may override the in-process pending record with an older stored record because it rebuilds the map from scratch and then merges pending:

```ts
for (const message of this.pending.get(sessionId) ?? []) {
  const record = recordFromMessage(message);
  if (record) {
    records.set(record.skillId, record);
  }
}
```

Actually pending does overwrite stored, so that's fine. But the stored→stored dedup by `!records.has(record.skillId)` keeps the oldest stored activation. If a skill was updated and re-activated in a later turn, the stored history would contain two pins with different versions, and `hydrate` would keep the older one, causing `isActive` to report the old version forever.

**Suggested fix:** When rebuilding from stored messages, keep the **most recent** activation per skill id (by `activatedAt` or by last-seen order), not the first:

```ts
for (const message of storedMessages) {
  const record = recordFromMessage(message);
  if (!record) continue;
  const existing = records.get(record.skillId);
  if (!existing || record.activatedAt > existing.activatedAt) {
    records.set(record.skillId, record);
  }
}
```

src/lib/skills/skillTools.ts
```diff
@@ -21,130 +32,257 @@ import type {
...
-              },
+              success: false,
+              error: `No skill named "${args.skill}" is available.`,
+            };
+          }
+          if (context.sessionId && context.sessionPersistence) {
+            await ensureHydrated(manager);
+            if (
+              !manager.sessions.isActive(
+                context.sessionId,
```
Tara-ag, line 230: ⚠️ **MAJOR — `use_skill` records activation before validating the skill is visible in scope, but the scope check happens after.**

Looking at the execute flow:

```ts
const skill = await manager.get(args.name);
if (!skill || !isSkillVisibleInScope(skill, context.scopeId)) {
  return { success: false, error: ... };
}

if (context.sessionId && context.sessionPersistence) {
  await ensureHydrated(manager);
  if (manager.sessions.isActive(...)) { ... }
  manager.sessions.recordActivation(context.sessionId, skill);
}
```

The scope check is correctly placed before recording. However, the `NOT_CONFIGURED` and not-found cases return early without recording, which is correct. No issue here — noting for verification only.

src/lib/context/contextCompactor.ts
```diff
@@ -180,7 +180,23 @@ export class ContextCompactor {
                 "LLM summarization timed out after 120s",
               );
               if (summarizeResult.summarized) {
-                currentMessages = summarizeResult.messages;
+                // Pinned skill instructions must survive summarization:
```
Tara-ag, line 183: ⚠️ **MAJOR — Summarization survivor logic assumes `summarizeResult.messages[0]` exists and is the summary message without checking.**

```ts
const survivorIds = new Set(summarizeResult.messages.map((m) => m.id));
const swallowedSkills = currentMessages.filter(
  (m) => m.metadata?.isSkill && !survivorIds.has(m.id),
);
currentMessages =
  swallowedSkills.length > 0
    ? [
        summarizeResult.messages[0],
        ...swallowedSkills,
        ...summarizeResult.messages.slice(1),
      ]
    : summarizeResult.messages;
```

If `summarizeResult.messages` is empty (edge case from a failed or no-op summarizer), `summarizeResult.messages[0]` will throw at runtime. The code is only reached when `summarizeResult.summarized` is true, but that doesn't guarantee a non-empty array.

**Suggested fix:** Guard against an empty result or assert the summary message exists:

```ts
if (summarizeResult.messages.length === 0) {
  currentMessages = summarizeResult.messages;
} else if (swallowedSkills.length > 0) {
  currentMessages = [
    summarizeResult.messages[0],
    ...swallowedSkills,
    ...summarizeResult.messages.slice(1),
  ];
} else {
  currentMessages = summarizeResult.messages;
}
```

src/lib/context/stages/slidingWindowTruncator.ts
```diff
@@ -133,8 +137,39 @@ export function truncateWithSlidingWindow(
   messages: ChatMessage[],
   config?: TruncationConfig,
 ): TruncationResult {
-  if (messages.length <= 4) {
-    // Delegate to content truncation for small conversations (BUG-005)
+  // Partition pinned skill messages out so the pair-based arithmetic below
```
Tara-ag, line 140: ⚠️ **MAJOR — Skill anchoring logic assumes every non-skill message after a skill is a valid anchor, but does not enforce role alternation when reassembling.**

The partition loop finds the next non-skill message as an anchor regardless of role. Later, `buildCandidate` inserts anchored skills immediately before their anchor. If a skill message (role=user) is anchored to an assistant message, inserting it before the assistant creates a `user → user → assistant` sequence, violating role alternation. The candidate is then passed to `validateRoleAlternation(candidateMessages)`, which will throw and abort truncation.

This is most likely to happen when a skill is loaded and the very next message is the assistant's answer (common case: ask → skill loaded → answer). The skill is user-role and the anchor is assistant-role, so re-inserting before the anchor produces consecutive user messages.

**Suggested fix:** When reassembling, ensure skill messages are placed in a role-alternation-safe position. One option is to always insert orphaned/trailing skills after the marker with a forced alternating role, or to attach skills to the preceding user message instead of the following message. Alternatively, convert skill messages to a neutral role if your message model supports it, or merge the skill content into the preceding user message. Given the current architecture, anchoring to the **preceding** user message (the ask that triggered the skill) preserves alternation: `user(ask+skill) → assistant(answer)`.

src/lib/core/conversationMemoryManager.ts
```diff
@@ -138,7 +138,14 @@ export class ConversationMemoryManager implements IConversationMemoryManager {
             assistantMsg.events = options.events;
           }
 
-          session.messages.push(userMsg, assistantMsg);
+          session.messages.push(userMsg);
```
Tara-ag, line 141: 💡 **MINOR — `options.skillMessages` spread into `session.messages` between user and assistant messages changes the invariant that `MESSAGES_PER_TURN` divides evenly.**

The fix in `getStats()` accounts for this by excluding `metadata?.isSkill` messages, which is good. However, any other code that iterates `session.messages` in pairs assuming `user → assistant → user → assistant` will now see triplets `user → skill(user) → assistant`. Make sure downstream consumers (summarization, context builders, export/formatting) are aware of this new shape. The PR already handles compaction; consider documenting the invariant change in `ChatMessageMetadata.isSkill` or in the memory manager comments.

src/lib/core/redisConversationMemoryManager.ts
```diff
@@ -683,6 +683,13 @@ export class RedisConversationMemoryManager implements IConversationMemoryManage
             normalizedUserId,
           );
 
+          // Pinned skill activations ride between ask and answer, mirroring
```
Tara-ag, line 686: ⚠️ **MAJOR — Redis memory manager pushes `skillMessages` **after** the user message but **before** the assistant message is constructed, which is correct, but the assistant message id/timestamp ordering may be misleading.**

The code inserts skill messages between user and assistant, mirroring the in-memory manager. That's fine. However, the assistant message's `timestamp` is generated **after** skill messages are pushed, so the stored order is:
1. user (older timestamp)
2. skill messages (no timestamp or later)
3. assistant (newest timestamp)

If skill messages lack `timestamp` they default to undefined; if they have one from `buildSkillActivationMessage` they will be `new Date().toISOString()` generated at activation time, which is before the assistant timestamp. This is consistent. No blocking issue, but verify that Redis deserialization preserves `metadata.isSkill` so that `getStats()` and compaction filters work correctly across the Redis path.

src/lib/skills/skillStoreS3.ts
```diff
@@ -212,26 +263,57 @@ export class S3SkillStore implements SkillStore {
...
+   * null body when the object is absent, or notModified when a
+   * conditional read reported the cached copy unchanged.
+   */
+  private async readIndexObject(
+    ops: SkillS3ObjectOps,
+  ): Promise<SkillS3ConditionalGetResult> {
+    if (ops.getObjectConditional) {
+      return ops.getObjectConditional(
+        this.indexKey(),
+        this.cachedIndexDoc ? this.cachedIndexEtag : undefined,
```
Tara-ag, line 285: ⚠️ **MAJOR — ETag cache can serve a stale parsed index after a failed conditional read returns `notModified: true` with a mismatched ETag.**

```ts
if (read.notModified && this.cachedIndexDoc) {
  return this.cachedIndexDoc;
}
```

This trusts the caller-provided `ops.getObjectConditional` to correctly compare ETags. If a custom `SkillS3ObjectOps` implementation has a bug and returns `notModified: true` when the ETag actually changed, the store silently serves stale data. While the interface contract says ops should be correct, the store could add a defensive parse of the returned body when `notModified` is true but no body is present — there's nothing to parse. A stronger defense is to invalidate the cache on any mutation (`put`/`delete`) and to only trust `notModified` when we have a cached ETag that matches what we sent. Currently `readIndexObject` passes `this.cachedIndexDoc ? this.cachedIndexEtag : undefined`, so if `cachedIndexDoc` exists but `cachedIndexEtag` is undefined (e.g. after a rebuild that failed to persist), a conditional read with `etag=undefined` may return `notModified=true` for any object, and we'd serve the rebuilt doc. That's probably okay.

However, after `writeIndex` the code sets `this.cachedIndexEtag = undefined`, so the next `index()` call will do a full unconditional read. That's intentional. The bigger concern is concurrency: two `index()` calls interleaving can corrupt `cachedIndexDoc`/`cachedIndexEtag`. Since Node.js is single-threaded and awaits yield, two concurrent `index()` calls can both enter `readOrRebuildIndex`, both miss cache, both read, and the second write may overwrite the first's cached ETag with an older one. This is a minor race.

**Suggested fix:** Serialize `index()` refresh with a lock or promise-deduplication, or at minimum document that `SkillS3ObjectOps` must implement conditional reads correctly.

src/lib/skills/skillStores.ts
```diff
@@ -163,6 +223,38 @@ export class FileSystemSkillStore implements SkillStore {
     return all.get(id) ?? null;
```
Tara-ag, line 223: 🔒 **CRITICAL — Symlink escape check uses `fs.realpathSync` but does not handle race conditions or relative symlinks that resolve inside then outside.**

```ts
const realDir = fs.realpathSync(dir);
const realResolved = fs.realpathSync(path.resolve(dir, resourcePath));
if (
  realResolved !== realDir &&
  !realResolved.startsWith(realDir + path.sep)
) {
  return null;
}
```

This is mostly correct for the synchronous case, but:

1. **TOCTOU race:** `realpathSync` and `readFileSync` are not atomic. An attacker can swap a file inside the skill dir for a symlink between the realpath check and the read.
2. **`realpathSync` follows symlinks recursively.** If `dir` itself is a symlink to `/safe/skills/foo`, `realpathSync(dir)` resolves to `/safe/skills/foo`. A resource path `../../etc/passwd` resolves relative to `dir` (the symlink location), not its target. `path.resolve(dir, resourcePath)` uses the symlink path, then `realpathSync` follows it. If the resolved path escapes, the check catches it. Good.
3. **Missing trailing separator edge case:** `startsWith(realDir + path.sep)` prevents `/safe/dir2` from matching `/safe/dir`, but `realDir` itself is rejected unless `realResolved === realDir` (which can't happen for a file). Good.

The bigger issue is that this check is **not applied in the S3 or Redis stores**. Those stores use string concatenation of `id` and `resourcePath` without path normalization or traversal checks. While S3/Redis keys are not filesystem paths, a malicious `resourcePath` like `../../../other-skill/resource.md` could still be used to read another skill's resources if the store implementation ever interprets the key as a path. The current S3 implementation uses `${prefix}skills/${id}/${resourcePath}` directly, so `id=foo&path=../bar/resource` becomes `skills/foo/../bar/resource` which S3 normalizes to `skills/bar/resource` — cross-skill read.

**Suggested fix:** Centralize path traversal validation in `SkillsManager.getResource` (which already rejects `..` and leading `/`) and ensure S3/Redis stores never receive unvalidated paths. The filesystem store should additionally use `fs.openSync` with `O_NOFOLLOW` where available, or at minimum document the TOCTOU limitation.

src/lib/skills/skillTools.ts
```diff
@@ -21,130 +32,257 @@ import type {
...
             error: error instanceof Error ? error.message : String(error),
-            data: { skills: [] },
           };
         }
       },
     }),
+  };
+}
 
+/**
```
Tara-ag, line 270: 💡 **MINOR — `read_skill_resource` gates on activation only when `sessionId && sessionPersistence`.**

When there is no session id or session persistence is disabled, the resource read is **not gated on activation**. This is inconsistent with the docs claim "gated on activation". In per-turn mode, a model could read resources of a skill it never loaded.

**Suggested fix:** Gate resource reads on activation regardless of persistence mode. In per-turn mode, track activations in a transient in-memory set keyed by session-less call context, or require `use_skill` to have been called in the same turn.

src/lib/skills/skillMatcher.ts
```diff
@@ -66,11 +79,78 @@ export function filterSkillIndex(
   return query.limit !== undefined ? matched.slice(0, query.limit) : matched;
```
Tara-ag, line 79: 💡 **MINOR — `renderSkillListing` falls back to a hard 80-char cap even when the first-sentence render already fits the budget, because it unconditionally returns the floor render after the sentence check fails.**

Wait, re-reading: it returns `sentences` if `sentences.length <= budgetChars`, otherwise returns the hard-cap floor. That's correct. However, the floor render is returned **even if it still exceeds the budget**. The comment acknowledges this. For very small budgets (e.g. `<available_skills>` tags + 20 skill names), the floor may still exceed the budget. Consider capping the number of skills as a last resort, or document that `listingBudgetChars` is a target, not a hard limit.

src/lib/skills/skillStoreS3.ts
```diff
@@ -160,6 +202,15 @@ export class S3SkillStore implements SkillStore {
     return `${this.prefix}index.json`;
   }
 
+  /** Resources live beside the skill object: <prefix>skills/<id>/<path>. */
+  private resourceKey(id: string, resourcePath: string): string {
+    return `${this.prefix}skills/${id}/${resourcePath}`;
+  }
+
+  async getResource(id: string, resourcePath: string): Promise<string | null> {
```
Tara-ag, line 210: 🔒 **CRITICAL — S3 resource key construction is vulnerable to path traversal via `resourcePath`.**

```ts
private resourceKey(id: string, resourcePath: string): string {
  return `${this.prefix}skills/${id}/${resourcePath}`;
}
```

`SkillsManager.getResource` validates `..` and leading `/`, but it does **not** reject URL-encoded traversal sequences, backslash-normalized `..\..`, or empty `id`. More importantly, the S3 key is built by naive string concatenation; a `resourcePath` like `../other-skill/resource.md` (if it somehow passed validation) becomes `skills/foo/../other-skill/resource.md`, which S3 normalizes to `skills/other-skill/resource.md`, escaping the skill boundary.

While `SkillsManager.getResource` currently rejects literal `..`, centralize a stricter validator that also rejects percent-encoded traversal, empty components, and absolute paths. Also consider validating in `S3SkillStore.getResource` itself as defense in depth.

src/lib/skills/skillStoreRedis.ts
```diff
@@ -59,6 +67,12 @@ export class RedisSkillStore implements SkillStore {
     return `${this.keyPrefix}${id}`;
   }
 
+  async getResource(id: string, resourcePath: string): Promise<string | null> {
```
Tara-ag, line 70: 🔒 **CRITICAL — Redis resource key construction is vulnerable to path traversal via `resourcePath`.**

```ts
const raw = await client.get(`${this.resourcePrefix}${id}:${resourcePath}`);
```

`SkillsManager.getResource` rejects literal `..` and leading `/`, but a `resourcePath` containing `:` or other Redis key separators can pollute the key namespace. More importantly, traversal segments like `..` embedded after normalization (e.g. encoded forms) are not caught here. Centralize strict path validation and apply it in every store, or at minimum validate in `RedisSkillStore.getResource` before building the key.

src/lib/skills/skillSessionTracker.ts
```diff
@@ -0,0 +1,171 @@
...
+    }
+    for (const record of records.values()) {
+      if (record.name === name) {
+        return record;
+      }
+    }
+    return undefined;
+  }
+
+  /**
```
Tara-ag, line 95: ⚠️ **MAJOR — `recordActivation` can append duplicate pending messages for the same skill within a single turn.**

```ts
recordActivation(sessionId: string, skill: SkillDefinition): ChatMessage {
  const message = buildSkillActivationMessage(skill);
  const record = recordFromMessage(message) as SkillActivationRecord;
  this.recordsFor(sessionId).set(skill.id, record);
  const queue = this.pending.get(sessionId);
  if (queue) {
    queue.push(message);
  } else {
    this.pending.set(sessionId, [message]);
  }
  return message;
}
```

There is no check that the skill is already in the pending queue before pushing. Combined with the preload race noted in `neurolink.ts`, this can produce multiple pinned messages for the same skill in one turn. `drainPending` returns all of them, and the memory manager inserts them all into history, wasting tokens and confusing dedup.

**Suggested fix:** Dedupe pending by skill id before appending:

```ts
recordActivation(sessionId: string, skill: SkillDefinition): ChatMessage {
  const queue = this.pending.get(sessionId);
  if (queue?.some((m) => m.metadata?.skillId === skill.id)) {
    // Return the existing pending message so callers can still inject its content.
    return queue.find((m) => m.metadata?.skillId === skill.id)!;
  }
  const message = buildSkillActivationMessage(skill);
  const record = recordFromMessage(message) as SkillActivationRecord;
  this.recordsFor(sessionId).set(skill.id, record);
  if (queue) {
    queue.push(message);
  } else {
    this.pending.set(sessionId, [message]);
  }
  return message;
}
```

src/lib/skills/skillsManager.ts
```diff
@@ -117,17 +130,70 @@ export class SkillsManager {
...
-      this.config.promptIndexMaxItems ?? DEFAULT_PROMPT_INDEX_MAX_ITEMS,
-    );
+    return items;
+  }
+
+  /**
+   * Read an auxiliary resource file bundled with a skill. Paths are
+   * relative to the skill; traversal segments are rejected. Null when the
+   * skill, the resource, or store resource support is absent.
+   */
```
Tara-ag, line 175: ⚠️ **MAJOR — `getResource` throws on invalid paths instead of failing open like the rest of the skills read path.**

```ts
if (
  normalized.startsWith("/") ||
  normalized.split("/").some((segment) => segment === "..")
) {
  throw new Error(
    `Invalid resource path "${resourcePath}" — must be relative to the skill directory`,
  );
}
```

The PR emphasizes "Reads fail open (errors → empty results + warn log)" in the docs, but a traversal attempt here propagates an exception to the caller. In `skillTools.ts` the `read_skill_resource` execute wraps `manager.getResource` in a try/catch and returns `{ success: false, error: ... }`, so the user-facing behavior is controlled. However, programmatic callers of `SkillsManager.getResource` (e.g. server routes, custom tools) will receive an exception. More importantly, the S3/Redis stores do not independently validate, so a custom store reached directly could be exploited.

**Suggested fix:** Return `null` for invalid paths (consistent with "not found") and log a warning, or document that this method throws on malicious input. Also consider moving the validation into a shared helper used by all stores.

src/lib/neurolink.ts
```diff
@@ -1913,30 +1923,178 @@ export class NeuroLink {
       if (!manager) {
```
Tara-ag, line 1923: ⚠️ **MAJOR — `applySkillsAugmentation` mutates `options.tools` by spreading injected tools **before** caller-supplied tools, but the comment says caller-supplied tools win.**

```ts
options.tools = {
  ...callTools,
  ...((options.tools as Record<string, unknown> | undefined) ?? {}),
};
```

This is actually correct: later spreads overwrite earlier ones, so `options.tools` wins. The comment is accurate. No issue here — noting for verification only.

src/lib/utils/conversationMemory.ts
```diff
@@ -591,6 +595,13 @@ export function buildContextFromPointer(
```
Tara-ag, line 595: ⚠️ **MAJOR — `buildContextFromPointer` re-includes **all** skill messages from before the pointer, even if they were intentionally removed or superseded.**

```ts
const pinnedSkillMessages = session.messages
  .slice(0, pointerIndex + 1)
  .filter((msg) => msg.metadata?.isSkill);
```

This pulls every historical skill pin into the current context. If a skill was activated and later explicitly unloaded (not implemented) or its activation was followed by a deprecation, it would still be re-injected. More realistically, if a conversation has many skill activations across many turns, summarization will accumulate all of them after the summary, growing context without bound and defeating the purpose of summarization.

The PR claims "pinned skill messages are re-included verbatim after the summary", but it should probably only re-include skill messages that are **still active** in the session (i.e. whose most recent activation record is present and not overridden). Since activation state is derived from history, consider filtering to skill ids that appear in the most recent portion of history or that have not been followed by a deactivation marker.

**Suggested fix:** At minimum, dedupe by skill id and keep only the most recent pin per skill. Better: intersect with the session's currently active skill set derived from `SkillSessionTracker` if available, or mark deprecations/unloads explicitly.

test/continuous-test-suite-skills.ts
```diff
@@ -1387,29 +1851,31 @@ async function testLiveGenerate(): Promise<void> {
 
     const content = result.content ?? "";
```
Tara-ag, line 1852: 💡 **MINOR — Live test still uses `JSON.stringify(result.toolExecutions ?? result).includes("use_skill")` instead of inspecting normalized tool execution names.**

The PR description says this was fixed, but the diff shows the live test still stringifies the result to check for `"use_skill"`. This can produce false positives if the word appears in content or error messages. The CodeRabbit comment on this exact line claimed it was addressed, but the current code does not reflect the suggested fix of checking `result.toolExecutions.some((e) => e.name === "use_skill")`.

**Suggested fix:** Replace the stringify check with a structured inspection:

```ts
const usedTool = (result.toolExecutions ?? []).some(
  (execution) => execution.name === "use_skill",
);
```

Tara-ag:
## Autonomous Review Summary

I reviewed the 21 changed files in this PR and added inline comments to the existing pending review. Below is a summary of **new** issues I raised (excluding points already covered by CodeRabbit's resolved comments).

### Issue Counts

| Severity | Count |
|----------|-------|
| 🔒 CRITICAL | 3 |
| ⚠️ MAJOR | 8 |
| 💡 MINOR | 4 |

### Blocking Criteria Met

- **CRITICAL security vulnerabilities**: Path traversal / unsafe resource key construction in filesystem, S3, and Redis skill stores.
- **≥3 MAJOR issues**: Preload race, hydration keeping oldest activation, empty summarization result assumption, sliding-window role alternation violation, duplicate pending activations, unbounded skill re-inclusion after summarization, `getResource` throwing on invalid paths, and S3 ETag trust issues.

### Key Concerns

1. **Security — resource path traversal** (`skillStores.ts`, `skillStoreS3.ts`, `skillStoreRedis.ts`)
   - Filesystem store's `realpathSync` check has TOCTOU and does not prevent symlink races.
   - S3 and Redis stores build keys by naive string concatenation without validating traversal segments, allowing cross-skill reads if validation is bypassed.

2. **Correctness — session pinning races** (`neurolink.ts`, `skillSessionTracker.ts`)
   - `preloadSkills` does not hydrate from stored history before checking `isActive`, risking double pins when the same skill is preloaded twice or also invoked via `use_skill`.
   - `recordActivation` can queue duplicate pending messages for the same skill in one turn.
   - `hydrate()` keeps the **first** (oldest) stored activation per skill instead of the most recent, so re-activated skills may report stale versions.

3. **Correctness — compaction** (`contextCompactor.ts`, `slidingWindowTruncator.ts`, `conversationMemory.ts`)
   - Summarization reassembly assumes `summarizeResult.messages[0]` exists without guarding against an empty array.
   - Sliding-window skill anchoring can produce consecutive `user` role messages, causing `validateRoleAlternation` to throw.
   - `buildContextFromPointer` re-includes **all** historical skill pins after a summary, which can grow context unboundedly and ignore superseded activations.

4. **Other**
   - `read_skill_resource` is not gated on activation when there is no session or session persistence is off.
   - `SkillsManager.getResource` throws on traversal attempts instead of failing open like other read paths.
   - Live test still uses `JSON.stringify(...).includes("use_skill")` rather than structured tool-execution inspection.

### Recommendation

**REQUEST CHANGES** — the CRITICAL path-traversal concerns and the volume of MAJOR correctness issues make this unsafe to merge without fixes. Most issues have localized fixes suggested in the inline comments.

> Note: I attached inline comments to the existing pending review by @Tara-ag because GitHub allows only one pending review per PR. I am unable to submit that review on behalf of another user, so this summary comment serves as the formal review decision.

pdogra1299:
## Response to @Tara-ag autonomous review — triage against head `f25086d8`

Verified every finding against the actual code. **3 items were valid hardening and are fixed; the rest are stale (already fixed before this review posted), refuted by the code, or documented design decisions.**

### Fixed in `f25086d8`

1. **S3/Redis resource key construction (defense in depth)** — both stores now validate paths via `isSafeSkillResourcePath` (rejects absolute paths, `..`, and empty segments) before building keys. Note the only LLM-reachable path already validated at the manager layer (`SkillsManager.getResource`), so cross-skill reads were not actually reachable; this adds a second layer for direct programmatic store use.
2. **Empty summarization result guard** — the Stage-3 re-seat no longer assumes `summarizeResult.messages[0]` exists.
3. **Live test tool detection** — now inspects `result.toolExecutions` entries structurally instead of `JSON.stringify(...).includes`.

### Stale / refuted

- **"preloadSkills does not hydrate before isActive"** — it does, unconditionally, and has since before this review posted (`neurolink.ts` `preloadSkills`: `manager.sessions.hydrate(...)` precedes every `isActive` check).
- **"recordActivation can queue duplicate pending messages in one turn"** — the check-then-record pair is a single synchronous block after all awaits; parallel `use_skill` calls share the merged active map and `hydrate()` merges pending, so the second call sees the first's record. This exact race was adversarially tested and refuted twice in our pre-PR review.
- **"validateRoleAlternation throws"** — it only `logger.warn`s; it has never thrown.
- **"realpathSync TOCTOU"** — the read targets the resolved real path after containment checking; the residual race requires a local-filesystem attacker swapping path components mid-syscall, which Node offers no openat-chain primitive to close. Realpath containment is the standard mitigation for this threat model.
- **"hydrate keeps the oldest activation → stale versions"** — intentional version pinning: a session keeps the version it activated (documented guarantee); dedup means at most one pin per skill id, so there is no "most recent" to prefer.
- **"unbounded skill re-inclusion after summarization"** — bounded by the number of distinct activated skills (one deduped pin each); superseded activations cannot exist within a session.
- **"read_skill_resource ungated without session persistence"** — intentional: the gate is a token-efficiency measure, not a security boundary; sessionless calls have no activation state to gate on, and resources carry the same trust level as `use_skill`, which is itself ungated.
- **"getResource throws instead of failing open"** — the throw is caught at the tool boundary and returned as a structured error envelope; an explicit "invalid path" message is more actionable for the model than a silent null (which means "not found").

src/lib/skills/skillTools.ts
Tara-ag, line 270: 🔒 **CRITICAL — `read_skill_resource` bypasses activation gating when there is no session or session persistence is disabled.**

The tool only checks `manager.sessions.isActive(...)` inside the `if (context.sessionId && context.sessionPersistence)` branch. In per-turn mode (no `sessionId`, or `sessionPersistence: false`), any skill resource can be read without ever calling `use_skill`. This contradicts the docs claim that resources are "gated on activation" and allows a model to read arbitrary skill files by name/path.

**Suggested fix:** Track activations in a transient per-call set when persistence is off, and gate `read_skill_resource` against it regardless of session mode:

```ts
// In createSkillCallTools
const transientActivations = new Set<string>();

// In use_skill execute, when not pinning:
if (!context.sessionId || !context.sessionPersistence) {
  transientActivations.add(skill.id);
}

// In read_skill_resource execute:
const isActive = context.sessionId && context.sessionPersistence
  ? manager.sessions.isActive(context.sessionId, skill.id, skill.name)
  : transientActivations.has(skill.id);
if (!isActive) {
  return { success: false, error: `Skill "${skill.name}" is not loaded — call use_skill first.` };
}
```

src/lib/skills/skillsManager.ts
Tara-ag, line 175: 🔒 **CRITICAL — `getResource` throws on malicious paths instead of failing open, and the validation is not reused by S3/Redis stores.**

The docs state "Reads fail open (errors → empty results + warn log)", but a traversal attempt here throws an exception. While `skillTools.ts` catches it, programmatic callers (server routes, custom tools, direct `manager.getResource`) receive an uncaught exception. More importantly, the S3 and Redis stores perform their own naive key concatenation without reusing this validator, so a custom/corner-case path can still escape the skill namespace.

**Suggested fix:** Return `null` and log a warning for invalid paths, and expose a shared validator that every store calls:

```ts
export function normalizeSkillResourcePath(resourcePath: string): string | null {
  const normalized = resourcePath.replace(/\\/g, "/");
  if (
    normalized.length === 0 ||
    normalized.startsWith("/") ||
    normalized.split("/").some((segment) => segment === ".." || segment === "")
  ) {
    return null;
  }
  return normalized;
}

// In getResource:
const normalized = normalizeSkillResourcePath(resourcePath);
if (!normalized) {
  logger.warn("[SkillsManager] Invalid resource path rejected", { resourcePath });
  return null;
}
```

Then call `normalizeSkillResourcePath` inside `S3SkillStore.getResource` and `RedisSkillStore.getResource` as defense in depth.

src/lib/skills/skillStoreS3.ts
```diff
@@ -160,6 +202,18 @@ export class S3SkillStore implements SkillStore {
...
```
Tara-ag, line 210: 🔒 **CRITICAL — S3 resource key is built by naive string concatenation and can traverse skill boundaries.**

`isSafeSkillResourcePath` rejects literal `..`, but it does not reject URL-encoded traversal (`%2e%2e/%2f`), backslash-normalized paths, or empty path segments. More importantly, the key `${this.prefix}skills/${id}/${resourcePath}` is assembled without further normalization; if a path like `../other-skill/resource.md` ever reached this method, S3 would normalize it to `skills/other-skill/resource.md`, escaping the skill namespace.

**Suggested fix:** Centralize strict path validation (reject `%`-encoded traversal, backslashes, empty segments) and apply it here as defense in depth, or reuse a shared `normalizeSkillResourcePath` helper that returns `null` for invalid input:

```ts
async getResource(id: string, resourcePath: string): Promise<string | null> {
  const normalized = normalizeSkillResourcePath(resourcePath);
  if (!normalized) {
    logger.warn("[SkillStoreS3] Invalid resource path rejected", { resourcePath });
    return null;
  }
  return this.getOps().getObject(this.resourceKey(id, normalized));
}
```

src/lib/skills/skillStoreRedis.ts
```diff
@@ -59,6 +67,15 @@ export class RedisSkillStore implements SkillStore {
...
```
Tara-ag, line 70: 🔒 **CRITICAL — Redis resource key construction is vulnerable to namespace pollution and traversal via `resourcePath`.**

The key `${this.resourcePrefix}${id}:${resourcePath}` is built by direct concatenation. A `resourcePath` containing `:` or other separators can pollute the Redis namespace. More importantly, traversal segments like `..` embedded after normalization (e.g. encoded forms) are not caught here, and the validator used (`isSafeSkillResourcePath`) only checks literal `..` and leading `/`.

**Suggested fix:** Apply the same strict centralized validator here, and reject any path that does not normalize cleanly:

```ts
async getResource(id: string, resourcePath: string): Promise<string | null> {
  const normalized = normalizeSkillResourcePath(resourcePath);
  if (!normalized) {
    logger.warn("[SkillStoreRedis] Invalid resource path rejected", { resourcePath });
    return null;
  }
  const client = await this.getClient();
  const raw = await client.get(`${this.resourcePrefix}${id}:${normalized}`);
  return raw ? String(raw) : null;
}
```

src/lib/skills/skillSessionTracker.ts
Tara-ag, line 95: ⚠️ **MAJOR — `recordActivation` can queue duplicate pending messages for the same skill in one turn.**

There is no check that the skill is already in the pending queue before pushing. Combined with the preload race noted in `neurolink.ts`, this can produce multiple pinned messages for the same skill in a single turn. `drainPending` returns all of them, and the memory manager inserts them all into history, wasting tokens and confusing downstream dedup.

**Suggested fix:** Dedupe pending by skill id before appending:

```ts
recordActivation(sessionId: string, skill: SkillDefinition): ChatMessage {
  const queue = this.pending.get(sessionId);
  const existing = queue?.find((m) => m.metadata?.skillId === skill.id);
  if (existing) {
    return existing;
  }
  const message = buildSkillActivationMessage(skill);
  const record = recordFromMessage(message) as SkillActivationRecord;
  this.recordsFor(sessionId).set(skill.id, record);
  if (queue) {
    queue.push(message);
  } else {
    this.pending.set(sessionId, [message]);
  }
  return message;
}
```

src/lib/skills/skillSessionTracker.ts
Tara-ag, line 115: ⚠️ **MAJOR — `hydrate()` keeps the oldest stored activation per skill instead of the most recent.**

```ts
if (record && !records.has(record.skillId)) {
  records.set(record.skillId, record);
}
```

This keeps the **first** occurrence of a skill pin in stored history and ignores later ones. If a skill is activated, later deactivated/unloaded, and then re-activated, or if the same skill appears multiple times in history due to a bug, the session will report the oldest version as active. Since `recordActivation` always increments the version from the current `skill.version`, but `hydrate` may override the in-process pending record with an older stored record (pending does overwrite stored, but stored→stored dedup keeps the oldest), `isActive` can report the wrong version forever.

**Suggested fix:** Keep the most recent activation per skill id (by `activatedAt` or last-seen order):

```ts
for (const message of storedMessages) {
  const record = recordFromMessage(message);
  if (!record) continue;
  const existing = records.get(record.skillId);
  if (!existing || record.activatedAt > existing.activatedAt) {
    records.set(record.skillId, record);
  }
}
```

src/lib/neurolink.ts
Tara-ag, line 1938: ⚠️ **MAJOR — Race between preload and concurrent `use_skill` activation can double-pin a skill.**

`preloadSkills` records an activation via `manager.sessions.recordActivation(...)`, which adds the message to the pending queue. If the model also calls `use_skill` for the same skill in the same turn (possible if the listing still shows the skill), the `use_skill` execute path will call `ensureHydrated` and then `isActive`. However, `ensureHydrated` reads from `conversationMemory.getSessionMessages`, which does **not** include the pending pin that has not yet been stored. So `isActive` returns false and `use_skill` will record a second pending message for the same skill, causing a duplicate pin after `storeConversationTurn`.

The tracker merges pending messages in `hydrate()`, but only if `hydrate()` is called **after** the preload recorded the pending message. In `preloadSkills` the call sequence is: `recordActivation` (adds to pending) → inject into system prompt. The `use_skill` tool may execute in parallel with preload or after, and its `ensureHydrated` call will merge pending if the pending map already has the entry. But if `use_skill` executes before preload reaches `recordActivation`, or if the pending queue is drained between preload and `use_skill`, you get duplicates.

**Suggested fix:** In `preloadSkills`, hydrate from stored messages before checking `isActive`, exactly as `use_skill` does, and dedupe pending pins by skill id/name when draining (also noted in `skillSessionTracker.ts`).

```ts
if (sessionId && sessionPersistence) {
  manager.sessions.hydrate(
    sessionId,
    this.conversationMemory
      ? await this.conversationMemory.getSessionMessages(sessionId, userId)
      : [],
  );
  if (manager.sessions.isActive(sessionId, skill.id, skill.name)) {
    continue;
  }
  block = manager.sessions.recordActivation(sessionId, skill).content;
}
```

The current code already hydrates before `isActive`, but the duplicate pending issue remains because `recordActivation` does not dedupe. Combine this fix with the `recordActivation` dedupe suggested in `skillSessionTracker.ts`.

src/lib/context/stages/slidingWindowTruncator.ts
Tara-ag, line 140: ⚠️ **MAJOR — Skill anchoring logic can violate role alternation when reassembling.**

The partition loop finds the next non-skill message as an anchor regardless of role. Later, `buildCandidate` inserts anchored skills immediately before their anchor. If a skill message (role=user) is anchored to an assistant message, inserting it before the assistant creates a `user → user → assistant` sequence, violating role alternation. The candidate is then passed to `validateRoleAlternation(candidateMessages)`, which only logs a warning and does not abort or repair. Most providers will reject or behave unpredictably with consecutive user messages.

This is most likely when a skill is loaded and the very next message is the assistant's answer (common case: ask → skill loaded → answer). The skill is user-role and the anchor is assistant-role, so re-inserting before the anchor produces consecutive user messages.

**Suggested fix:** Anchor skills to the **preceding** user message (the ask that triggered the skill) instead of the following message, or ensure reassembly never creates consecutive same-role messages. One option:

```ts
// Anchor to the nearest preceding conversational message of role "user"
const anchor = messages
  .slice(0, i)
  .reverse()
  .find((prev) => !prev.metadata?.isSkill && prev.role === "user");
```

Alternatively, in `buildCandidate`, when inserting anchored skills before an assistant message, swap the skill to after the assistant or merge it into the preceding user message.

src/lib/context/contextCompactor.ts
```diff
@@ -180,7 +180,25 @@ export class ContextCompactor {
...
```
Tara-ag, line 183: ⚠️ **MAJOR — Summarization survivor logic assumes `summarizeResult.messages[0]` exists without checking.**

```ts
const survivorIds = new Set(summarizeResult.messages.map((m) => m.id));
const swallowedSkills = currentMessages.filter(
  (m) => m.metadata?.isSkill && !survivorIds.has(m.id),
);
currentMessages =
  swallowedSkills.length > 0
    ? summarizeResult.messages.length > 0
      ? [
          summarizeResult.messages[0],
          ...swallowedSkills,
          ...summarizeResult.messages.slice(1),
        ]
      : swallowedSkills
    : summarizeResult.messages;
```

The code is only reached when `summarizeResult.summarized` is true, but that doesn't guarantee a non-empty array. If `summarizeResult.messages` is empty (edge case from a failed or no-op summarizer), the ternary avoids indexing `[0]`, but the resulting `currentMessages = swallowedSkills` may still be wrong if there are no survivors at all. More importantly, if `summarizeResult.messages` has entries but `messages[0]` is not the summary message (e.g. the summarizer preserves system prompts first), the skill messages are inserted in the wrong place.

**Suggested fix:** Assert or explicitly find the summary message by `metadata?.isSummary`, and guard against an empty result:

```ts
if (summarizeResult.messages.length === 0) {
  currentMessages = summarizeResult.messages;
} else {
  const summaryIndex = summarizeResult.messages.findIndex(
    (m) => m.metadata?.isSummary,
  );
  const insertAt = summaryIndex >= 0 ? summaryIndex + 1 : 1;
  const before = summarizeResult.messages.slice(0, insertAt);
  const after = summarizeResult.messages.slice(insertAt);
  currentMessages =
    swallowedSkills.length > 0
      ? [...before, ...swallowedSkills, ...after]
      : summarizeResult.messages;
}
```

src/lib/utils/conversationMemory.ts
Tara-ag, line 595: ⚠️ **MAJOR — `buildContextFromPointer` re-includes all historical skill messages, even if they were superseded or intentionally removed.**

```ts
const pinnedSkillMessages = session.messages
  .slice(0, pointerIndex + 1)
  .filter((msg) => msg.metadata?.isSkill);
```

This pulls every historical skill pin into the current context. If a skill was activated and later explicitly unloaded (not implemented) or its activation was followed by a deprecation, it would still be re-injected. More realistically, if a conversation has many skill activations across many turns, summarization will accumulate all of them after the summary, growing context without bound and defeating the purpose of summarization.

The PR claims "pinned skill messages are re-included verbatim after the summary", but it should probably only re-include skill messages that are **still active** in the session (i.e. whose most recent activation record is present and not overridden).

**Suggested fix:** At minimum, dedupe by skill id and keep only the most recent pin per skill. Better: intersect with the session's currently active skill set derived from `SkillSessionTracker` if available, or mark deprecations/unloads explicitly.

```ts
const pinnedSkillMessages = session.messages
  .slice(0, pointerIndex + 1)
  .filter((msg) => msg.metadata?.isSkill)
  .reduceRight((acc, msg) => {
    const id = msg.metadata?.skillId;
    if (id && !acc.seen.has(id)) {
      acc.seen.add(id);
      acc.messages.unshift(msg);
    }
    return acc;
  }, { seen: new Set<string>(), messages: [] as ChatMessage[] })
  .messages;
```

src/lib/skills/skillStores.ts
Tara-ag, line 223: 🔒 **CRITICAL — Filesystem resource path containment check has TOCTOU race and is not applied to S3/Redis stores.**

The check uses `fs.realpathSync` synchronously before `fs.readFileSync`, but an attacker can swap a file for a symlink between the realpath check and the read. Additionally, this validation logic is not reused by the S3 or Redis stores, which concatenate `resourcePath` directly into keys; a malicious `resourcePath` like `../other-skill/resource.md` could read another skill's resources if it bypasses the higher-level validator.

**Suggested fix:**
1. Use `fs.openSync(realResolved, O_RDONLY | O_NOFOLLOW)` (where available) and read from the file descriptor to avoid the TOCTOU race.
2. Centralize strict path validation (reject encoded traversal, backslashes, empty segments) in a shared helper and call it in every store's `getResource`.

```ts
import { openSync, readFileSync } from "node:fs";
import { constants } from "node:fs";

// In getResource, after containment check:
try {
  const fd = openSync(realResolved, constants.O_RDONLY | constants.O_NOFOLLOW);
  return readFileSync(fd, "utf-8");
} catch (error) {
  // ENOENT / ELOOP handled as before
}
```

src/lib/core/conversationMemoryManager.ts
Tara-ag, line 141: 💡 **MINOR — `options.skillMessages` spread into `session.messages` between user and assistant messages changes the invariant that `MESSAGES_PER_TURN` divides evenly.**

The fix in `getStats()` accounts for this by excluding `metadata?.isSkill` messages, which is good. However, any other code that iterates `session.messages` in pairs assuming `user → assistant → user → assistant` will now see triplets `user → skill(user) → assistant`. Make sure downstream consumers (summarization, context builders, export/formatting) are aware of this new shape. The PR already handles compaction; consider documenting the invariant change in `ChatMessageMetadata.isSkill` or in the memory manager comments.

This is already partially noted in the comment at this line, but the public type docs for `ChatMessageMetadata.isSkill` should mention that skill messages are extra user-role rows inside a turn and that pair-wise iteration must exclude them.

src/lib/core/redisConversationMemoryManager.ts
Tara-ag, line 686: 💡 **MINOR — Verify Redis deserialization preserves `metadata.isSkill` so `getStats()` and compaction filters work correctly.**

The code inserts skill messages between user and assistant, mirroring the in-memory manager. That's fine. However, the assistant message's `timestamp` is generated **after** skill messages are pushed, so the stored order is:
1. user (older timestamp)
2. skill messages (timestamp from activation time)
3. assistant (newest timestamp)

This is consistent. No blocking issue, but confirm that Redis deserialization preserves `metadata.isSkill` so that `getStats()` and compaction filters work correctly across the Redis path. If `deserializeConversation` strips unknown metadata fields, the turn-counting fix and truncation protection will silently fail for Redis-backed sessions.

Tara-ag (changes requested):
## Review Summary

Reviewed the skills progressive-disclosure rewrite (21 files, +2,157/-548). The architecture is sound, but I found **5 CRITICAL** and **6 MAJOR** new issues that must be addressed before merging. All are in the skills resource/activation safety and memory-compaction paths.

### Blocking issues

**Security / resource isolation (CRITICAL)**
- `read_skill_resource` is not gated on activation when there is no `sessionId` or `sessionPersistence: false`, allowing arbitrary skill-resource reads per turn.
- `SkillsManager.getResource` throws on malicious paths instead of failing open, and the validation is not reused by S3/Redis stores.
- S3 and Redis resource keys are built by naive string concatenation; traversal can escape the skill namespace if the higher-level validator is ever bypassed.
- Filesystem resource containment has a TOCTOU race and lacks centralized strict validation.

**Activation correctness (MAJOR → CRITICAL in aggregate)**
- `SkillSessionTracker.recordActivation` can queue duplicate pending pins for the same skill in one turn.
- `SkillSessionTracker.hydrate` keeps the oldest stored activation per skill instead of the most recent.
- `preloadSkills` vs. concurrent `use_skill` can double-pin a skill because pending pins are not visible to `ensureHydrated`.

**Compaction / context shape (MAJOR)**
- Sliding-window skill anchoring can insert a user-role skill message before an assistant message, breaking role alternation.
- Summarization survivor logic assumes `summarizeResult.messages[0]` is the summary without verifying.
- `buildContextFromPointer` re-includes every historical skill pin, causing unbounded context growth.

### Non-blocking
- Document the new `user → skill(user) → assistant` triplet invariant so downstream consumers don't assume pair-wise message iteration.
- Confirm Redis deserialization preserves `metadata.isSkill`.

### Next steps
Please address the CRITICAL resource-gating and path-traversal issues, dedupe pending activations, keep the most recent activation in `hydrate`, and fix the role-alternation-safe skill reassembly. Once those are resolved and tests pass, I'm happy to re-review.

---
*Note: I added comments to the existing pending review by @Tara-ag because GitHub only allows one pending review per PR.*

Tara-ag (dismissed):
## Review Summary

Reviewed the skills progressive-disclosure rewrite (21 files, +2,157/-548). The architecture is sound, but I found **5 CRITICAL** and **6 MAJOR** new issues that must be addressed before merging. All are in the skills resource/activation safety and memory-compaction paths.

### Blocking issues

**Security / resource isolation (CRITICAL)**
- `read_skill_resource` is not gated on activation when there is no `sessionId` or `sessionPersistence: false`, allowing arbitrary skill-resource reads per turn.
- `SkillsManager.getResource` throws on malicious paths instead of failing open, and the validation is not reused by S3/Redis stores.
- S3 and Redis resource keys are built by naive string concatenation; traversal can escape the skill namespace if the higher-level validator is ever bypassed.
- Filesystem resource containment has a TOCTOU race and lacks centralized strict validation.

**Activation correctness (MAJOR → CRITICAL in aggregate)**
- `SkillSessionTracker.recordActivation` can queue duplicate pending pins for the same skill in one turn.
- `SkillSessionTracker.hydrate` keeps the oldest stored activation per skill instead of the most recent.
- `preloadSkills` vs. concurrent `use_skill` can double-pin a skill because pending pins are not visible to `ensureHydrated`.

**Compaction / context shape (MAJOR)**
- Sliding-window skill anchoring can insert a user-role skill message before an assistant message, breaking role alternation.
- Summarization survivor logic assumes `summarizeResult.messages[0]` is the summary without verifying.
- `buildContextFromPointer` re-includes every historical skill pin, causing unbounded context growth.

### Non-blocking
- Document the new `user → skill(user) → assistant` triplet invariant so downstream consumers don't assume pair-wise message iteration.
- Confirm Redis deserialization preserves `metadata.isSkill`.

### Next steps
Please address the CRITICAL resource-gating and path-traversal issues, dedupe pending activations, keep the most recent activation in `hydrate`, and fix the role-alternation-safe skill reassembly. Once those are resolved and tests pass, I'm happy to re-review.

pdogra1299:
## Response to @Tara-ag re-review (CHANGES_REQUESTED) — evidence against head `ea94266e`

This re-review repeats the first review's findings; each was either **already fixed in the head it reviewed** or **refuted with code evidence** in the prior triage. Point-by-point:

### Claims that are stale (fix is IN the reviewed head)

- **"S3/Redis resource keys built by naive concatenation"** — both stores validate via `isSafeSkillResourcePath` (rejects absolute paths, `..`, empty segments) since `f25086d8`: `skillStoreS3.ts` `getResource` and `skillStoreRedis.ts` `getResource` both early-return `null` on unsafe paths. The validator IS the centralized one (exported from `skillMatcher.ts`, reused by both stores; the filesystem store additionally does realpath containment).
- **"Summarization survivor logic assumes messages[0]"** — guarded since `f25086d8` (`contextCompactor.ts`: `summarizeResult.messages.length > 0` branch).

### Claims that are factually wrong against the code

- **"preloadSkills vs concurrent use_skill can double-pin because pending pins are not visible to ensureHydrated"** — `SkillSessionTracker.hydrate()` explicitly merges pending pins into the rebuilt active set (`skillSessionTracker.ts:127-132`: `for (const message of this.pending.get(sessionId) ?? []) { records.set(record.skillId, record) }`). Pending pins are visible to every dedup check by construction.
- **"recordActivation can queue duplicate pending pins in one turn"** — the isActive-check → recordActivation pair is a single synchronous block with no await between; parallel calls share the merged map. Adversarially tested and refuted twice in pre-PR review.
- **"Role alternation breaks"** — `validateRoleAlternation` warns, never throws; and the `user → skill(user) → assistant` triplet is the documented invariant (now also spelled out on `StoreConversationTurnOptions.skillMessages` in `ea94266e`, per your non-blocking ask). The truncator partitions skills out of pair arithmetic precisely so alternation-sensitive logic operates on the conversational stream.

### Design decisions, not defects (rationale documented in prior triage)

- **Sessionless `read_skill_resource` gating** — the activation gate is a token-efficiency measure, not a security boundary: `use_skill` is ungated and returns full instructions, and resources carry identical trust. "Arbitrary skill-resource reads" describes the tool working as intended on scope-visible skills.
- **`getResource` throwing on invalid paths** — caught at the tool boundary and returned as a structured error envelope (fail-open at the LLM boundary); an explicit message beats a silent null.
- **`hydrate` keeping the first stored activation** — intentional version pinning: a session follows the version it activated; dedup guarantees at most one pin per skill id, so "most recent" cannot differ.
- **`buildContextFromPointer` re-inclusion** — bounded by distinct activated skills (one deduped pin each); not unbounded.
- **Filesystem TOCTOU** — reads target the containment-checked resolved real path; the residual race requires a local-FS attacker mid-syscall, which Node provides no openat-chain primitive to close. Realpath containment is the standard mitigation for this threat model.

### Non-blocking asks — done

- Triplet invariant documented on `StoreConversationTurnOptions.skillMessages` (`ea94266e`).
- Redis deserialization preserves `metadata.isSkill`: `deserializeConversation` (`utils/redis.ts`) is a plain JSON round-trip with structural validation only — message objects, including metadata, pass through untouched.

All CI checks (build, tests, docs validation, quality/security gates, single-commit policy, CodeQL, Yama, CodeRabbit) are green. Requesting dismissal or re-review of the CHANGES_REQUESTED state.

src/lib/skills/skillStoreS3.ts
pdogra1299: Fixed in f25086d8 (present in the head this review targeted): `getResource` validates via the shared `isSafeSkillResourcePath` (rejects absolute paths, `..`, empty segments) before any key construction. The manager-level validator additionally guards the only LLM-reachable path.

src/lib/skills/skillStoreS3.ts
pdogra1299: Fixed in f25086d8 (present in the head this review targeted): `getResource` validates via the shared `isSafeSkillResourcePath` (rejects absolute paths, `..`, empty segments) before any key construction. The manager-level validator additionally guards the only LLM-reachable path.

src/lib/skills/skillStoreRedis.ts
pdogra1299: Fixed in f25086d8 (present in the head this review targeted): `getResource` validates via the shared `isSafeSkillResourcePath` (rejects absolute paths, `..`, empty segments) before any key construction. The manager-level validator additionally guards the only LLM-reachable path.

src/lib/skills/skillStoreRedis.ts
pdogra1299: Fixed in f25086d8 (present in the head this review targeted): `getResource` validates via the shared `isSafeSkillResourcePath` (rejects absolute paths, `..`, empty segments) before any key construction. The manager-level validator additionally guards the only LLM-reachable path.

src/lib/context/contextCompactor.ts
pdogra1299: Fixed in f25086d8: the re-seat now checks `summarizeResult.messages.length > 0` and falls back to the swallowed skills alone when the summarizer returns an empty array.

src/lib/context/contextCompactor.ts
pdogra1299: Fixed in f25086d8: the re-seat now checks `summarizeResult.messages.length > 0` and falls back to the swallowed skills alone when the summarizer returns an empty array.

test/continuous-test-suite-skills.ts
pdogra1299: Fixed in f25086d8: the live test now inspects `result.toolExecutions` entries structurally (`toolName`/`tool === "use_skill"`).

src/lib/core/conversationMemoryManager.ts
pdogra1299: Addressed in ea94266e: the `user → skill(user) → assistant` triplet invariant is documented on `StoreConversationTurnOptions.skillMessages`, with a pointer to the truncator's partition-and-reanchor pattern for pair-based consumers. Turn statistics already exclude `metadata.isSkill` messages in both managers.

pdogra1299: Addressed in ea94266e: the `user → skill(user) → assistant` triplet invariant is documented on `StoreConversationTurnOptions.skillMessages`, with a pointer to the truncator's partition-and-reanchor pattern for pair-based consumers. Turn statistics already exclude `metadata.isSkill` messages in both managers.

src/lib/skills/skillTools.ts
pdogra1299: Acknowledged — the comment itself concludes there is no issue at this site ("noting for verification only"). Resolving.

src/lib/neurolink.ts
pdogra1299: Acknowledged — the comment itself concludes there is no issue at this site ("noting for verification only"). Resolving.

src/lib/core/redisConversationMemoryManager.ts
pdogra1299: Verified: `deserializeConversation` (`utils/redis.ts`) is a plain JSON round-trip with structural validation of top-level fields only — message objects, including `metadata.isSkill`, pass through untouched, so stats and compaction filters work identically on the Redis path.

pdogra1299: Verified: `deserializeConversation` (`utils/redis.ts`) is a plain JSON round-trip with structural validation of top-level fields only — message objects, including `metadata.isSkill`, pass through untouched, so stats and compaction filters work identically on the Redis path.

src/lib/neurolink.ts
pdogra1299: Refuted: `preloadSkills` hydrates unconditionally before its `isActive` check, and `hydrate()` merges pending pins into the rebuilt set (`skillSessionTracker.ts:127-132`), so a concurrent `use_skill`'s own pre-check hydration sees the preload pin and vice versa. The check→record pair is synchronous after all awaits — no interleaving point exists between them.

pdogra1299: Refuted: `preloadSkills` hydrates unconditionally before its `isActive` check, and `hydrate()` merges pending pins into the rebuilt set (`skillSessionTracker.ts:127-132`), so a concurrent `use_skill`'s own pre-check hydration sees the preload pin and vice versa. The check→record pair is synchronous after all awaits — no interleaving point exists between them.

src/lib/skills/skillSessionTracker.ts
pdogra1299: Refuted: after the awaits complete, `isActive` → `recordActivation` is a single synchronous block on the shared per-session map; parallel invocations serialize on it and the second sees the first's record (hydrate also merges pending). This exact scenario was adversarially tested and refuted twice in the pre-PR 33-agent review.

pdogra1299: Refuted: after the awaits complete, `isActive` → `recordActivation` is a single synchronous block on the shared per-session map; parallel invocations serialize on it and the second sees the first's record (hydrate also merges pending). This exact scenario was adversarially tested and refuted twice in the pre-PR 33-agent review.

src/lib/context/stages/slidingWindowTruncator.ts
pdogra1299: Two corrections: (1) `validateRoleAlternation` warns — it never throws, so truncation cannot abort; (2) inserting an anchored skill before its following conversational message reproduces the exact original stored order (`ask → skill → answer`), which is the documented triplet invariant (see `StoreConversationTurnOptions.skillMessages`). Anchoring to the preceding user message would change byte order versus stored history and break replay stability.

pdogra1299: Two corrections: (1) `validateRoleAlternation` warns — it never throws, so truncation cannot abort; (2) inserting an anchored skill before its following conversational message reproduces the exact original stored order (`ask → skill → answer`), which is the documented triplet invariant (see `StoreConversationTurnOptions.skillMessages`). Anchoring to the preceding user message would change byte order versus stored history and break replay stability.

src/lib/skills/skillSessionTracker.ts
pdogra1299: By design (version pinning): a session follows the version it activated, and dedup guarantees at most one pin per skill id in history — there is no newer activation to prefer. A mid-session skill update intentionally does not mutate instructions the model already follows; `use_skill` reports the pinned version on `already_loaded`.

pdogra1299: By design (version pinning): a session follows the version it activated, and dedup guarantees at most one pin per skill id in history — there is no newer activation to prefer. A mid-session skill update intentionally does not mutate instructions the model already follows; `use_skill` reports the pinned version on `already_loaded`.

src/lib/skills/skillTools.ts
pdogra1299: By design: the activation gate is a token-efficiency measure, not a security boundary. `use_skill` is ungated and returns full instructions for any scope-visible skill; resources carry identical trust. Sessionless calls have no activation state to gate against — denying reads there would only force a redundant `use_skill` round-trip.

pdogra1299: By design: the activation gate is a token-efficiency measure, not a security boundary. `use_skill` is ungated and returns full instructions for any scope-visible skill; resources carry identical trust. Sessionless calls have no activation state to gate against — denying reads there would only force a redundant `use_skill` round-trip.

src/lib/skills/skillsManager.ts
pdogra1299: By design: the throw is caught at the tool boundary and returned as a structured error envelope, so the LLM path fails open with an actionable "invalid path" message; a silent null would be indistinguishable from "not found". Store-level validation is additionally in place (f25086d8) for direct programmatic use.

pdogra1299: By design: the throw is caught at the tool boundary and returned as a structured error envelope, so the LLM path fails open with an actionable "invalid path" message; a silent null would be indistinguishable from "not found". Store-level validation is additionally in place (f25086d8) for direct programmatic use.

src/lib/utils/conversationMemory.ts
pdogra1299: Bounded by construction: at most one deduped pin exists per activated skill per session, so re-inclusion grows with distinct activated skills, not with turns. "Superseded" pins cannot exist — dedup prevents a second pin and version pinning is intentional.

pdogra1299: Bounded by construction: at most one deduped pin exists per activated skill per session, so re-inclusion grows with distinct activated skills, not with turns. "Superseded" pins cannot exist — dedup prevents a second pin and version pinning is intentional.

src/lib/skills/skillStores.ts
pdogra1299: The read targets the containment-checked resolved real path (`realpathSync` on both dir and target, then `readFileSync` on the resolved path). The residual window requires a local-filesystem attacker swapping path components mid-syscall — Node exposes no openat-chain primitive to close it, and realpath containment is the standard mitigation for this threat model. The shared `isSafeSkillResourcePath` validator is applied in S3/Redis stores (f25086d8); the filesystem store's realpath check is strictly stronger.

pdogra1299: The read targets the containment-checked resolved real path (`realpathSync` on both dir and target, then `readFileSync` on the resolved path). The residual window requires a local-filesystem attacker swapping path components mid-syscall — Node exposes no openat-chain primitive to close it, and realpath containment is the standard mitigation for this threat model. The shared `isSafeSkillResourcePath` validator is applied in S3/Redis stores (f25086d8); the filesystem store's realpath check is strictly stronger.

src/lib/skills/skillStoreS3.ts
pdogra1299: Accepted as a benign, self-correcting race: interleaved `index()` reads can at worst cache one stale doc+ETag pair for a single TTL cycle (30s default); the next conditional read revalidates and corrects. The manager's TTL cache serializes the common path. The ops contract (correct If-None-Match semantics) is documented on `SkillS3ObjectOps.getObjectConditional`.

src/lib/skills/skillMatcher.ts
pdogra1299: Intentional and documented at the return site: names are the hard guarantee and are never dropped, so for pathologically small budgets the floor render may exceed the target. Dropping entries would reintroduce the invisible-skill failure mode the listing exists to prevent.

Tara-ag:
## Post-merge review notes (could not open a new pending review)

The PR is already merged and another reviewer (`Tara-ag`) has an outstanding pending review, so I could not create a separate pending review to attach inline comments. Below are the net-new observations from an additional pass that are not already raised in the existing review threads.

### New finding — `sessionPersistence` can be true when conversation memory failed to initialize

**File:** `src/lib/neurolink.ts` (`applySkillsAugmentation`)

```ts
const memoryAvailable =
  Boolean(this.conversationMemory) ||
  Boolean(this.conversationMemoryConfig?.conversationMemory?.enabled);
const sessionPersistence =
  (this.skillsConfig.sessionPersistence ?? true) &&
  Boolean(sessionId) &&
  memoryAvailable;
```

`memoryAvailable` is `true` if memory is *configured* as enabled even when `this.conversationMemory` is still `null` (e.g., initialization failed or has not run yet). In that state `sessionPersistence` is `true`, so `preloadSkills` / `use_skill` will:
1. Record the activation in `SkillSessionTracker.pending`.
2. Suppress re-injection on later turns because `hydrate()` merges the undrained pending queue.
3. Never actually persist the pinned message because `storeConversationTurn` is guarded by `this.conversationMemory` being non-null.

The result is a “ghost” active skill: the tracker believes the skill is loaded, but the instructions are not in history and are not re-injected, so the model loses them on subsequent turns.

**Suggested fix:** Tie `sessionPersistence` to the initialized manager, not just the config flag:

```ts
const memoryAvailable = Boolean(this.conversationMemory);
```

If the intent is to allow pinning when memory config is enabled but lazily initialized, ensure `initializeConversationMemory()` is awaited before this check, or keep the config OR only for the purpose of triggering initialization and then re-evaluate.

### New finding — pending activations can be lost when the session tracker evicts old sessions

**File:** `src/lib/skills/skillSessionTracker.ts` (`recordsFor`)

```ts
if (this.active.size >= MAX_TRACKED_SESSIONS) {
  const oldest = this.active.keys().next().value;
  if (oldest !== undefined) {
    this.active.delete(oldest);
    this.pending.delete(oldest);
  }
}
```

When the 1001st session is touched, the oldest session is deleted **including its pending queue**. If that session’s turn has not yet been stored, the pending activation messages are dropped before `drainPending` can return them. This is a latent data-loss window for long-lived multi-tenant processes.

**Suggested fix:** Either (a) drain pending for the evicted session before deleting it, or (b) do not count pending-only sessions toward `MAX_TRACKED_SESSIONS`, or (c) raise the eviction behavior to a warning log so hosts can tune the limit.

### Existing unresolved threads (for tracking)

The following threads from `Tara-ag` were still unresolved at the time of this pass and remain relevant post-merge:

1. **Public SDK API break** (`src/lib/types/skills.ts`) — removal of `promptIndex` and `search_skills` without a documented migration path violates CLAUDE.md Critical Rule 5.
2. **`SkillSessionTracker.recordActivation` duplicate pending pins** (`src/lib/skills/skillSessionTracker.ts`).
3. **`SkillSessionTracker.hydrate` keeps oldest activation** instead of most recent (`src/lib/skills/skillSessionTracker.ts`).

### Bottom line

The architecture is solid, but the two new findings above (ghost persistence and pending eviction) are worth addressing in a follow-up commit, along with the unresolved API-break / activation-dedup threads already raised.

PR #1143, opened 2026-07-10
fix(proxy): reset-aware account switching on rate limits

pdogra1299:
## Description

Fixes the Claude proxy re-hammering a rate-limited Anthropic account instead of switching to a healthy one. When an account hit its 5-hour session (or 7-day weekly) subscription limit, the proxy cooled it for a hardcapped **60 seconds** and retried the same dead account up to **10 times** before rotating — so every request kept failing on the exhausted account even with another account available. The authoritative reset timestamp Anthropic returns in `anthropic-ratelimit-unified-*` headers was parsed and persisted, but never used for routing.

## Type of Change

- [x] New feature (non-breaking, opt-out): quota-optimized account selection

## Changes Made

- **Reset-aware cooldown.** On a 429 the proxy now parses the unified quota headers and classifies **weekly (7d) exhaustion**, **session (5h) exhaustion**, or **transient burst**. Exhaustion cools the account until its *actual* reset epoch (no 60s cap) and rotates **immediately** (zero same-account retries). Transient bursts honor `retry-after` as a floor + jitter, capped at 2 same-account retries (was 10).
- **Quota-optimized selection** (`orderAccountsByQuota`, fill-first): orders usable accounts by **soonest-resetting window first** (weekly reset, then 5h reset, then highest utilization) to maximize consumption of soon-to-refresh quota before it is lost. Rejected windows are proactively skipped until their reset. Toggle via `NEUROLINK_PROXY_QUOTA_ROUTING` (default on).
- **Proactive cooling** from success responses (quota headers are present on 200s too), so an account that just tipped over is skipped on the next request without eating a 429.

## Testing

- [x] `tsc --noEmit --strict` clean; `prettier --check` clean; ESLint 0 errors.
- [x] `npm run build` + `validate:all` pass (pre-commit hook).
- [x] Unit/regression tests added: `testPlanCooldownFor429` (5h/7d/transient classification + reset-based cooldown) and `testOrderAccountsByQuota` (soonest-reset-first ordering), driven via `__testHooks` (no-API, `proxy-primary` category). Verified at runtime.

## Breaking Changes

- [x] No breaking changes (quota routing is opt-out via env; round-robin strategy unchanged).

## Follow-ups (deferred)

- Formal single-flight half-open probe to avoid a thundering herd when a cooldown expires under many concurrent requests.
- Env-configurable cooldown caps (currently constants).

## Summary by CodeRabbit

* **New Features**
  * Improved account selection using quota and reset-window information.
  * Added smarter handling for rate limits, including account rotation and controlled retries.
  * Added proactive cooldowns when session or weekly quotas are exhausted.

* **Bug Fixes**
  * Reduced retry spikes through jittered retry timing.
  * Prevented cooldown periods from ending before the reported quota reset.
  * Improved cleanup of account cooldown status after successful requests.

* **Tests**
  * Added coverage for quota-based cooldown planning, retries, and account ordering.

### DIFF
--- a/src/lib/server/routes/claudeProxyRoutes.ts
+++ b/src/lib/server/routes/claudeProxyRoutes.ts
@@ -72,6 +72,8 @@ import {
   recordFinalSuccess,
 } from "../../proxy/usageStats.js";
 import type {
+  AccountCooldownPlan,
+  AccountQuota,
   AnthropicAttemptLogger,
   AnthropicAuthRetryResult,
   AnthropicLoopState,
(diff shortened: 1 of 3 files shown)


pdogra1299:
Thanks @coderabbitai — addressed all 4 actionable comments + the nitpick, amended into the single commit (`00fe0514`):

- **MIN_COOLDOWN_MS on transient cooldown** — transient base now `Math.max(MIN_COOLDOWN_MS, Math.min(base, TRANSIENT_MAX_COOLDOWN_MS))`.
- **Auth-retry quota handling** — `handleAnthropicSuccessfulRetryResponse` now takes `accountState`, stashes quota + calls `maybeCoolFromQuota`; the retry-429 branch parses quota and applies `planCooldownFor429` (extend-only) before rotating, so a session/weekly-exhausted account isn't re-selected next request.
- **Cap the final jittered delay** — `Math.min(MAX_RATE_LIMIT_RETRY_DELAY_MS, jitteredDelay(base))` (the "5159-5164" second reference is the `jitteredDelay` definition, not a second call site — only one call exists).
- **Monotonic cooldown under concurrency** — 429/exhaustion set is now extend-only (`plan.coolingUntil > existing`); success only clears cooldown when it has already expired, so a stale in-flight success can't wipe an active exhaustion cooldown.
- **Stronger case-4 assertions** — now also asserts `reason === "session"` and `rotateImmediately === true`.

tsc/lint/build clean; pure-function fixes verified at runtime.

PR #1144, opened 2026-07-10
fix(proxy): seed persisted quota and probe unknown accounts

pdogra1299:
## Description

Follow-up to #1143 (quota-aware account switching, shipped in 9.85.1). Field-observed on a live 2-account deployment: after a proxy restart, **all traffic locked onto the non-primary account** and never moved. Three root causes, confirmed from the deployed instance's logs, `/status`, and `~/.neurolink/account-quotas.json`:

1. **Quota file clobbered on restart** — `saveAccountQuota` marked its cache loaded without reading disk first, so the first save after boot rewrote `account-quotas.json` with only the accounts used since boot, erasing the other accounts' snapshots.
2. **Ordering blind at boot** — quota-aware ordering reads only in-memory runtime state, which a restart wipes; nothing seeded it from the persisted snapshots. All accounts tied, and selection fell back to token-store enumeration order (not the configured primary).
3. **Unknown-quota starvation** — the comparator treated missing quota as `Infinity` ("resets last"), so the first account to serve became self-reinforcing: it alone had data, always sorted first, and the other account never got a request to reveal its windows.

## Type of Change

## Changes Made

- `accountQuota.ts`: `saveAccountQuota` loads the persisted file into the cache before its first write, merging instead of clobbering across restarts.
- `claudeProxyRoutes.ts`: new `seedRuntimeQuotasFromDisk` hydrates each account's runtime quota from the persisted snapshots (label→key) during account loading; never overwrites fresher in-memory data; stale resets degrade gracefully (past timestamps are ignored).
- `orderAccountsByQuota`: usable accounts with **no quota data are probed first** (one request reveals their windows and self-corrects the ordering), and equal-knowledge ties break by the **configured primary**, then insertion order.

## Testing

- [x] `tsc --strict`, prettier, ESLint gates, `validate:all`, and full build pass (pre-commit hook).
- [x] New no-API regression tests: `testSaveAccountQuotaMerges` (merge across restart), `testSeedRuntimeQuotasFromDisk` (seeds from disk, preserves in-memory), extended `testOrderAccountsByQuota` (probe-first + primary tie-break).
- [x] Runtime-verified the exact field scenario: after restart with seeded disk data, the account whose weekly window resets sooner is selected regardless of token-store enumeration order.

## Breaking Changes

- [x] No breaking changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved account quota data across proxy restarts by reusing existing persisted quota entries instead of overwriting them.
  * Improved quota restoration on startup/reconfiguration, merging newly saved quota with existing on-disk data.
  * Enhanced account selection by prioritizing accounts with known quota, “probing” accounts without quota information, and using the configured primary account to break ordering ties.

### DIFF
--- a/src/lib/proxy/accountQuota.ts
+++ b/src/lib/proxy/accountQuota.ts
@@ -205,13 +205,20 @@ export async function loadAccountQuota(
  * Update quota for a single account.
  * Updates in-memory cache immediately (non-blocking),
  * then debounces the disk write to every 5 seconds.
+ *
+ * Loads the persisted file into the cache before the first write so a save
+ * after a process restart merges with existing entries instead of rewriting
+ * the file with only the accounts used since boot (which silently erased
+ * other accounts' snapshots and blinded quota-aware routing to them).
  */
 export async function saveAccountQuota(
   accountKey: string,
   quota: AccountQuota,
 ): Promise<void> {
+  if (!cacheLoaded) {
+    await loadAccountQuotas();
+  }
   memoryCache[accountKey] = quota;
-  cacheLoaded = true;
   dirty = true;
   scheduleFlush();
 }
--- a/src/lib/server/routes/claudeProxyRoutes.ts
+++ b/src/lib/server/routes/claudeProxyRoutes.ts
@@ -21,6 +21,7 @@ import {
   parseClaudeCodeUserId,
 } from "../../auth/anthropicOAuth.js";
 import {
+  loadAccountQuotas,
   parseQuotaHeaders,
   saveAccountQuota,
 } from "../../proxy/accountQuota.js";
@@ -363,6 +364,32 @@ function maybeCoolFromQuota(
   }
 }
 
+/**
+ * Seed each account's runtime quota from the persisted snapshots in
+ * ~/.neurolink/account-quotas.json (keyed by label). Runtime state is
+ * in-memory only, so without this the quota-aware ordering is blind after a
+ * proxy restart: all accounts tie, selection falls back to token-store
+ * enumeration order, and the first account served becomes self-reinforcing
+ * (it alone has data) — starving the others regardless of their resets.
+ * Never overwrites fresher in-memory quota; stale disk snapshots degrade
+ * gracefully because past reset timestamps are ignored by resetEpochToMs.
+ */
+async function seedRuntimeQuotasFromDisk(
+  accounts: ProxyPassthroughAccount[],
+): Promise<void> {
+  try {
+    const persisted = await loadAccountQuotas();
+    for (const account of accounts) {
+      const state = getOrCreateRuntimeState(account.key);
+      if (!state.quota && persisted[account.label]) {
+        state.quota = persisted[account.label];
+      }
+    }
+  } catch {
+    // Non-fatal: seeding is best-effort; ordering falls back to probe-first.
+  }
+}
+
 /** Quota-aware selection is on by default; disable with
  *  NEUROLINK_PROXY_QUOTA_ROUTING=off|false|0. Only affects the fill-first
  *  strategy (round-robin keeps strict rotation). */
@@ -384,6 +411,7 @@ function accountSortMetrics(accountKey: string, now: number) {
     q?.sessionStatus === "rejected" && sessionReset !== undefined;
   return {
     usable: !coolingActive && !weeklyRejected && !sessionRejected,
+    hasQuota: !!q,
     coolingUntil: st?.coolingUntil ?? 0,
     weeklyReset: weeklyReset ?? Number.POSITIVE_INFINITY,
     sessionReset: sessionReset ?? Number.POSITIVE_INFINITY,
@@ -397,17 +425,21 @@ function accountSortMetrics(accountKey: string, now: number) {
  * allowance isn't wasted, then move to accounts with longer-dated resets.
  *
  * Priority among usable accounts:
- *   1. soonest WEEKLY (7d) reset  — the scarce, use-it-or-lose-it ceiling
- *   2. soonest SESSION (5h) reset
- *   3. highest weekly utilization — finish off the one closest to done
- * Accounts with no quota data yet keep insertion order (stable sort) and sit
- * after those with a known soonest reset. Cooling/rejected accounts sort last,
- * soonest-back-to-service first, as last resort.
+ *   1. no quota data yet — probe first: one request reveals its windows and
+ *      self-corrects the ordering. (Ranking unknowns last would starve them
+ *      forever: never picked → never observed → never comparable.)
+ *   2. soonest WEEKLY (7d) reset  — the scarce, use-it-or-lose-it ceiling
+ *   3. soonest SESSION (5h) reset
+ *   4. highest weekly utilization — finish off the one closest to done
+ *   5. configured primary account, then insertion order
+ * Cooling/rejected accounts sort last, soonest-back-to-service first, as
+ * last resort.
  */
 function orderAccountsByQuota(
   accounts: ProxyPassthroughAccount[],
   now: number,
 ): ProxyPassthroughAccount[] {
+  const primaryKey = configuredPrimaryAccountKey;
   return [...accounts].sort((a, b) => {
     const ma = accountSortMetrics(a.key, now);
     const mb = accountSortMetrics(b.key, now);
@@ -419,13 +451,22 @@ function orderAccountsByQuota(
       const bu = mb.coolingUntil || Number.POSITIVE_INFINITY;
       return au - bu;
     }
+    if (ma.hasQuota !== mb.hasQuota) {
+      return ma.hasQuota ? 1 : -1;
+    }
     if (ma.weeklyReset !== mb.weeklyReset) {
       return ma.weeklyReset - mb.weeklyReset;
     }
     if (ma.sessionReset !== mb.sessionReset) {
       return ma.sessionReset - mb.sessionReset;
     }
-    return mb.weeklyUsed - ma.weeklyUsed;
+    if (ma.weeklyUsed !== mb.weeklyUsed) {
+      return mb.weeklyUsed - ma.weeklyUsed;
+    }
+    if (primaryKey && (a.key === primaryKey) !== (b.key === primaryKey)) {
+      return a.key === primaryKey ? -1 : 1;
+    }
+    return 0;
   });
 }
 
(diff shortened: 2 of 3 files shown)


Tara-ag (approved):
## Review Summary

**Files reviewed:** 3
**New issues raised:** 0
**Decision:** Approve

### Analysis

This is a focused bug-fix follow-up to #1143 that addresses three field-observed quota routing failures after proxy restart:

1. **`src/lib/proxy/accountQuota.ts`** — `saveAccountQuota` now loads the persisted quota file before its first write, preventing the cache from being seeded only with accounts used since boot and clobbering snapshots for other accounts. The removal of the unconditional `cacheLoaded = true` is the correct fix.

2. **`src/lib/server/routes/claudeProxyRoutes.ts`** —
   - `seedRuntimeQuotasFromDisk` hydrates in-memory runtime quota from persisted snapshots during account loading, without overwriting fresher in-memory data. Because it runs inside `loadClaudeProxyAccounts`, it also primes the `accountQuota` module cache, so the first `saveAccountQuota` call avoids disk I/O on the hot path.
   - `orderAccountsByQuota` now probes unknown accounts first (breaking the starvation cycle) and uses the configured primary as a tie-breaker. The comparator correctly preserves the existing usable → cooling/rejected ordering hierarchy.

3. **`test/continuous-test-suite-proxy.ts`** — Adds targeted regression tests for merge-across-restart, disk-to-runtime seeding, probe-first ordering, and primary tie-breaking.

### Checks performed

- No public SDK API breakage (changes are internal to the proxy routing layer).
- No violation of CLAUDE.md Critical Rules 1, 3, 4, 5, or 6.
- No injection, unsafe eval, or SSRF vectors introduced.
- Error handling is best-effort and non-fatal where appropriate.
- New behavior is covered by the proxy continuous test suite.

The fix is coherent, well-commented, and the tests directly exercise the three root causes described in the PR. Approved for merge.

src/lib/proxy/accountQuota.ts
```diff
@@ -205,13 +205,20 @@ export async function loadAccountQuota(
...
+ * Loads the persisted file into the cache before the first write so a save
+ * after a process restart merges with existing entries instead of rewriting
+ * the file with only the accounts used since boot (which silently erased
+ * other accounts' snapshots and blinded quota-aware routing to them).
  */
 export async function saveAccountQuota(
   accountKey: string,
   quota: AccountQuota,
 ): Promise<void> {
+  if (!cacheLoaded) {
```
Tara-ag, line 218: ⚠️ **MAJOR: `saveAccountQuota` now performs synchronous-looking async I/O on the hot path.**

The docstring says "Updates in-memory cache immediately (non-blocking)", but the new `await loadAccountQuotas()` makes the first call to `saveAccountQuota` block on a full disk read before it can update the cache. In the proxy request path this is called from `handleAnthropicSuccessfulResponse` after every upstream response, so the first request after restart now pays the cost of reading `~/.neurolink/account-quotas.json` synchronously (within the async call).

More importantly, this creates a **race window**: two concurrent first-save calls will both see `cacheLoaded === false`, both `await loadAccountQuotas()`, and the later `memoryCache[accountKey] = quota` assignment + `scheduleFlush()` can overwrite the earlier caller's in-memory update before flush. The second load also re-reads a potentially stale disk snapshot after the first caller already updated `memoryCache`, effectively reverting the first caller's memory update.

**Suggested fix:** keep the merge-on-restart behavior, but make it non-blocking and race-safe:

```ts
export async function saveAccountQuota(
  accountKey: string,
  quota: AccountQuota,
): Promise<void> {
  if (!cacheLoaded) {
    // Load once, serialized; do not block subsequent callers on I/O.
    await loadAccountQuotas();
  }
  memoryCache[accountKey] = quota;
  dirty = true;
  scheduleFlush();
}
```

Actually the current code already does that, but it is not serialized. Add a loader promise:

```ts
let loadPromise: Promise<void> | null = null;

async function ensureCacheLoaded(): Promise<void> {
  if (cacheLoaded) return;
  if (!loadPromise) {
    loadPromise = loadAccountQuotas().then(() => { loadPromise = null; });
  }
  await loadPromise;
}
```

And call `await ensureCacheLoaded()` in `saveAccountQuota`. This keeps the first caller waiting for disk I/O but avoids the duplicate-load race and keeps the comment honest about non-blocking updates after the first load.

src/lib/server/routes/claudeProxyRoutes.ts
```diff
@@ -363,6 +364,32 @@ function maybeCoolFromQuota(
   }
 }
 
+/**
+ * Seed each account's runtime quota from the persisted snapshots in
+ * ~/.neurolink/account-quotas.json (keyed by label). Runtime state is
+ * in-memory only, so without this the quota-aware ordering is blind after a
+ * proxy restart: all accounts tie, selection falls back to token-store
+ * enumeration order, and the first account served becomes self-reinforcing
+ * (it alone has data) — starving the others regardless of their resets.
```
Tara-ag, line 373: ⚠️ **MAJOR: `seedRuntimeQuotasFromDisk` swallows all errors silently and mutates shared module state without synchronization.**

The empty `catch` block means a corrupt `account-quotas.json`, permission error, or `JSON.parse` failure will be invisible in logs. Since this runs at proxy boot for every request that loads accounts, a persistent read failure will never be reported and operators will wonder why quota routing is not seeding.

Also, `state.quota = persisted[account.label]` assigns the quota object by reference. Later code (e.g. `maybeCoolFromQuota`) mutates `state.coolingUntil` on the runtime state, but `saveAccountQuota` later serializes `memoryCache` which contains the same object reference. This is currently safe because `AccountQuota` does not include `coolingUntil`, but it is fragile: any future addition to `RuntimeAccountState` that overlaps with `AccountQuota` fields will silently corrupt the persisted snapshot.

**Suggested fix:**

```ts
async function seedRuntimeQuotasFromDisk(
  accounts: ProxyPassthroughAccount[],
): Promise<void> {
  try {
    const persisted = await loadAccountQuotas();
    for (const account of accounts) {
      const state = getOrCreateRuntimeState(account.key);
      if (!state.quota && persisted[account.label]) {
        state.quota = { ...persisted[account.label] };
      }
    }
  } catch (error) {
    logger.warn("[proxy] failed to seed runtime quotas from disk", {
    });
  }
}
```

This logs the failure (so it is observable) and clones the persisted quota so runtime mutations cannot leak back into the cache that will be flushed to disk.

src/lib/server/routes/claudeProxyRoutes.ts
```diff
@@ -419,13 +451,22 @@ function orderAccountsByQuota(
       const bu = mb.coolingUntil || Number.POSITIVE_INFINITY;
```
Tara-ag, line 451: 💡 **MINOR: probe-first ordering can starve known accounts when an unknown account is unhealthy.**

The new comparator puts `!hasQuota` accounts ahead of all known-quota accounts. If an account is permanently misconfigured/disabled but still appears in the enabled list with no quota data, it will be probed repeatedly on every request while healthy known accounts wait. In practice this is mitigated by the `usable` check and the downstream account loop, but a disabled-yet-enabled account with no quota will sort first and burn attempts.

Consider capping probe-first to the first request after boot (e.g. track a `probedAt` timestamp in runtime state) so unknowns get one probe each and then fall back to the reset-aware ordering. Not blocking, but worth noting given the stated goal is "one request reveals its windows".

test/continuous-test-suite-proxy.ts
```diff
@@ -1689,11 +1689,152 @@ async function testOrderAccountsByQuota(): Promise<boolean | null> {
...
+    await import("../src/lib/server/routes/claudeProxyRoutes.js");
+  const { initAccountQuota } = await import("../src/lib/proxy/accountQuota.js");
+
+  const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), "nl-seed-test-"));
+  const quotaPath = path.join(tmpDir, "account-quotas.json");
+  try {
+    __testHooks.resetAllRuntimeState();
+    const diskQuota = makeQuota({ weeklyResetAt: 1_900_000_000 });
+    fs.writeFileSync(quotaPath, JSON.stringify({ "a@test": diskQuota }));
+    initAccountQuota(quotaPath);
```
Tara-ag, line 1789: 💡 **MINOR: test leaves a dangling async flush to the throwaway path.**

`saveAccountQuota` schedules a 5-second debounced flush. The test calls `initAccountQuota(path.join(tmpDir, "discard.json"))` in `finally`, but the flush timer may already be armed for the original `quotaPath`. Re-pointing the module path does not cancel the timer (it only cancels a *pending* timer at the moment `initAccountQuota` is called). The armed timer can still fire after the `finally` block and write to `discard.json`, or worse, race with `fs.rmSync(tmpDir, ...)` if the timer fires before cleanup.

**Suggested fix:** await the flush explicitly before re-pointing the path:

```ts
await saveAccountQuota("b@test", makeQuota({}) as never);
// Force the debounced write so the test does not leak a timer.
await flushAccountQuota(); // if such a helper exists, or expose one
```

If no flush helper exists, consider adding an `__testHooks.flushAccountQuota()` or calling `initAccountQuota` only after a `sleep(FLUSH_INTERVAL_MS + 50)`. The current cleanup is probably fine in practice, but it makes the test timing-dependent.

Tara-ag (approved):
## Review Summary

Reviewed 3 changed files in this proxy quota-seeding fix.

### New issues raised

| Severity | Count | Files |
|----------|-------|-------|
| 🔒 CRITICAL | 0 | — |
| ⚠️ MAJOR | 2 | `src/lib/proxy/accountQuota.ts`, `src/lib/server/routes/claudeProxyRoutes.ts` |
| 💡 MINOR | 2 | `src/lib/server/routes/claudeProxyRoutes.ts`, `test/continuous-test-suite-proxy.ts` |

### Blocking assessment

No blocking criteria are met:
- No hardcoded secrets or critical security vulnerabilities.
- No public SDK API breakage (changes are internal to the Claude proxy).
- No violation of non-negotiable CLAUDE.md Critical Rules that ESLint does not enforce.
- Only 2 MAJOR issues raised (< 3 threshold).

### Key concerns

1. **`saveAccountQuota` race on first write** — Two concurrent first-save callers can both load from disk and the later load can revert the earlier in-memory update. Recommend serializing the initial load with a single loader promise.
2. **`seedRuntimeQuotasFromDisk` silently swallows all errors** — Disk/parse failures at boot are invisible. Also assigns persisted quota by reference, making runtime mutations fragile against future persisted-snapshot corruption. Recommend logging the error and cloning the quota object.
3. **Probe-first ordering** — Correctly fixes the starvation problem described in the PR, but could repeatedly probe an unhealthy unknown account. Consider capping probes to one per account after boot.
4. **Test cleanup** — `testSaveAccountQuotaMerges` leaves a debounced flush timer that may fire after the temp directory is removed. Recommend flushing explicitly or waiting out the debounce interval before cleanup.

The bug-fix logic is sound and the new regression tests cover the reported field scenario. Approved with the above inline notes for follow-up.

PR #1147, opened 2026-07-11
fix(deps): move js-yaml to runtime dependencies

murdore:
## Bug

`src/lib/skills/skillStores.ts` (line 13) imports `js-yaml` at runtime, but `js-yaml` is declared only in **devDependencies**. Any consumer that doesn't happen to hoist a transitive copy crashes at import time:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'js-yaml' imported from node_modules/@juspay/neurolink/dist/skills/skillStores.js
```

Verified against the registry: present in `dist` of **9.83.0 through 9.86.2**, with `js-yaml` absent from declared `dependencies` in all of them. Neurolink's own CI never sees it because dev installs always carry devDependencies.

Real-world hit: `juspay/director#63` (dependabot bump to 9.83.0, then 9.86.2) fails both Verify jobs with exactly this error — https://github.com/juspay/director/pull/63#issuecomment-4946229474.

## Fix

Move `js-yaml: ^4.2.0` from devDependencies to dependencies (`@types/js-yaml` stays dev). Lockfile regenerated with `pnpm install --lockfile-only`. Single-line runtime change, no code touched.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@juspay/neurolink",
-  "version": "9.86.4",
+  "version": "9.86.5",
   "packageManager": "pnpm@10.15.1",
   "description": "TypeScript AI SDK with 24+ LLM providers behind one consistent API. MCP-native (58+ servers), voice TTS/STT/realtime, RAG, agents, memory, context compaction. OpenAI · Anthropic · Gemini · Bedrock · Azure · Ollama · DeepSeek · NVIDIA NIM and more.",
   "author": {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [9.86.5](https://github.com/juspay/neurolink/compare/v9.86.4...v9.86.5) (2026-07-14)
+
+### Bug Fixes
+
+- **(proxy):** harden routing and process reliability ([be7487c](https://github.com/juspay/neurolink/commit/be7487cec6870176dfbcac3ed8b547a36bcc887a))
+
 ## [9.86.4](https://github.com/juspay/neurolink/compare/v9.86.3...v9.86.4) (2026-07-11)
 
 ### Bug Fixes


package.json
Tara-ag, line 1: ⚠️ **MAJOR**: Moving `js-yaml` to `dependencies` is correct for `src/lib/skills/skillStores.ts`, but it also exposes `js-yaml` as a runtime dependency for `src/lib/processors/data/YamlProcessor.ts` which currently uses `require("js-yaml")` inside `parseYamlSecurely`. That processor is part of the SDK file-processing pipeline and will now resolve at runtime, but the `require()` call is synchronous and will throw `ERR_REQUIRE_ESM` if `js-yaml` ever ships ESM-only, and it bypasses the dynamic-import pattern used elsewhere for optional parsers. Consider converting `YamlProcessor.parseYamlSecurely` to a dynamic `import("js-yaml")` guarded by availability, consistent with `src/cli/commands/auth.ts` and `src/lib/proxy/proxyConfig.ts`.

Suggested direction:
```ts
private async parseYamlSecurely(content: string): Promise<unknown> {
  const yaml = await import("js-yaml");
  return yaml.load(content, { schema: yaml.CORE_SCHEMA });
}
```
and propagate `await` through `validateDownloadedFileWithResult` / `buildProcessedResult`.

Tara-ag, line 1: 💡 **MINOR**: The PR description says "`@types/js-yaml` stays dev" — confirmed. However, `scripts/check-banned-deps.ts` also imports `js-yaml` at the top level (`import yaml from "js-yaml";`). That script is invoked by `pnpm run check:deps`, which is a development/build-time script, so keeping `@types/js-yaml` in devDependencies is sufficient. No action needed, but worth noting that this is the only other non-test source reference and it remains dev-time only.

Tara-ag (commented):
## Review Summary

**PR:** Move `js-yaml` from `devDependencies` to `dependencies` so SDK consumers can resolve the runtime import in `src/lib/skills/skillStores.ts`.

**Files reviewed:** 2 (`package.json`, `pnpm-lock.yaml`)

**New issues raised:**
- ⚠️ **MAJOR** (1): `src/lib/processors/data/YamlProcessor.ts` uses synchronous `require("js-yaml")`; now that `js-yaml` is a runtime dependency, this will work but is inconsistent with the project's dynamic-import pattern and risks `ERR_REQUIRE_ESM` if the package ever goes ESM-only. Suggest converting to `await import("js-yaml")` and propagating the async signature.
- 💡 **MINOR** (1): Confirmed `@types/js-yaml` correctly stays in `devDependencies`; noted that `scripts/check-banned-deps.ts` is the remaining dev-time import and is appropriately covered by dev types.

**Blocking criteria:** Not met. No hardcoded secrets, no critical security vulnerabilities, no public SDK API breakage, and no non-negotiable CLAUDE.md rule violations. The single MAJOR issue is architectural/consistency rather than a runtime crash or security flaw.

**Decision:** `COMMENT` — the dependency move itself is correct and should land, but please consider addressing the `YamlProcessor` `require()` → dynamic `import()` inconsistency in a follow-up (or here if you prefer).

PR #1150, opened 2026-07-12
feat(video): imageInputKey option for Replicate image-to-video models

murdore:
## Problem

`ReplicateVideoHandler` hardcodes the input image under the `image` key with a legacy `num_frames`/`fps`/`aspect_ratio` payload. Current Replicate i2v models define different schemas and **fail the prediction on submit** when their required image key is missing:

| Model | Required image key | Observed failure |
|---|---|---|
| `minimax/hailuo-2.3-fast` | `first_frame_image` (required) | 422 on submit |
| `wan-video/wan-2.7-i2v` | `first_frame` / `first_clip` | `ValueError: At least one of 'first_frame' or 'first_clip' must be provided.` |

Schemas verified against the live Replicate models API (`GET /v1/models/...` → `latest_version.openapi_schema`, 2026-07-12). Both models also take `duration` (seconds) — `num_frames`/`fps` are not schema fields for either. Found while wiring these models as draft-tier b-roll generators in juspay/director (a NeuroLink-based pipeline).

## Change

- `VideoOutputOptions.imageInputKey?: string` — the key the model expects the image data URI under. Setting it also switches the payload to the modern `duration`/`resolution` shape those models expect.
- **Backward compatible:** when omitted, the payload is byte-for-byte the legacy shape — existing callers unaffected. `options.output.video` extras already reach the handler verbatim, so callers can adopt this with no extra plumbing.

Example: `output.video = { provider: 'replicate', model: 'minimax/hailuo-2.3-fast', imageInputKey: 'first_frame_image', length: 4, resolution: '720p' }`

## Side observation (deliberately not changed here)

The handler's default model `atonamy/wan-alpha` is text-to-video only — its schema has no image field, so the `image` key it is sent is silently ignored. Worth revisiting the default separately; left untouched to keep this PR surgical.

`pnpm run typecheck` clean; pre-commit check/format/lint suite passed.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring an optional image input key for image-to-video model integrations.
  * Improved compatibility with modern image-to-video models by using duration-based payloads (with optional resolution).
  * Preserved legacy video generation behavior when the image input key is not provided.

### DIFF
--- a/src/lib/adapters/video/replicateVideoHandler.ts
+++ b/src/lib/adapters/video/replicateVideoHandler.ts
@@ -76,27 +76,38 @@ export class ReplicateVideoHandler implements VideoHandler {
     const model = options.model ?? DEFAULT_MODEL;
     const dataUri = `data:image/${this.detectImageType(image)};base64,${image.toString("base64")}`;
 
-    // Wan-Alpha + most image-to-video models accept this shape; specific
-    // models may require provider-specific extras passed through
-    // VideoOutputOptions.[unknown key].
+    // Replicate image-to-video models do not share an input schema. The
+    // legacy shape below (`image` + `num_frames`/`fps`/`aspect_ratio`) fits
+    // Wan-Alpha-era models; current models (minimax/hailuo-2.3-*,
+    // wan-video/wan-2.7-i2v) key the image as `first_frame_image` /
+    // `first_frame` and take `duration` instead of frame counts — a missing
+    // required image key fails the prediction on submit. `imageInputKey`
+    // selects the image key AND the modern `duration`/`resolution` shape;
+    // omitting it preserves the legacy payload exactly.
     //
-    // `resolution` is forwarded as the `resolution` input parameter.
-    // Wan-Alpha and several other Replicate image-to-video models accept it
-    // (e.g. "720p", "1080p"). Models that do not recognise it will silently
-    // ignore the field — the Replicate API does not reject unknown input keys.
     // `calculateDimensions` still populates the metadata `dimensions` field
     // so downstream consumers always receive correct width/height regardless
     // of whether the model honoured the resolution hint.
-    const inputPayload: Record<string, unknown> = {
-      image: dataUri,
-      prompt,
-      num_frames: (options.length ?? 4) * 24, // Assume 24 fps
-      fps: 24,
-      aspect_ratio: options.aspectRatio,
-      ...(options.resolution !== undefined
-        ? { resolution: options.resolution }
-        : {}),
-    };
+    const inputPayload: Record<string, unknown> =
+      options.imageInputKey !== undefined
+        ? {
+            [options.imageInputKey]: dataUri,
+            prompt,
+            duration: options.length ?? 4,
+            ...(options.resolution !== undefined
+              ? { resolution: options.resolution }
+              : {}),
+          }
+        : {
+            image: dataUri,
+            prompt,
+            num_frames: (options.length ?? 4) * 24, // Assume 24 fps
+            fps: 24,
+            aspect_ratio: options.aspectRatio,
+            ...(options.resolution !== undefined
+              ? { resolution: options.resolution }
+              : {}),
+          };
 
     let prediction: Awaited<ReturnType<typeof predict>>;
     try {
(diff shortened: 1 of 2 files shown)


Tara-ag (approved):
## Review Summary

Reviewed 2 changed files against `CLAUDE.md` and the project focus areas.

### What changed
- `src/lib/types/multimodal.ts`: adds optional `imageInputKey?: string` to `VideoOutputOptions` (additive, backward-compatible per CLAUDE.md Critical Rule 5).
- `src/lib/adapters/video/replicateVideoHandler.ts`: switches Replicate payload shape when `imageInputKey` is provided, preserving the legacy `image` + `num_frames`/`fps`/`aspect_ratio` shape when omitted.

### Existing review comments already raised
I loaded the current review threads and did **not** duplicate them. The following points are already being discussed and should be resolved/addressed by the author:

1. **Runtime validation of `imageInputKey`** (Copilot, `replicateVideoHandler.ts:110`) — `!== undefined` lets `null`/empty/whitespace through as computed keys.
2. **Doc clarity on `imageInputKey: "image"`** (Copilot, `multimodal.ts:229`) — callers can opt into the modern `duration`/`resolution` shape while keeping the default `image` key.
3. **Missing mocked payload tests** (Copilot, `replicateVideoHandler.ts:86`) — no test asserts the exact POST body for the `imageInputKey` set vs. omitted branches.
4. **Model-specific enum normalization** (CodeRabbit, `replicateVideoHandler.ts:110`) — e.g. `minimax/hailuo-2.3-fast` expects `duration` ∈ {6,10} and `resolution` ∈ {768P,1080P}.
5. **Computed key collision** (CodeRabbit, `replicateVideoHandler.ts:96`) — `imageInputKey` equal to `"prompt"`, `"duration"`, or `"resolution"` silently drops the image due to object-literal overwrite semantics.

### New issues found this run
None. No hardcoded secrets, no breaking SDK API changes, no violations of the non-negotiable CLAUDE.md critical rules enforced outside CI, and no new security/correctness issues independent of the threads above.

### Decision
**Approve.** The change is surgical, additive, and preserves backward compatibility. Please consider the existing inline feedback before merging — especially runtime validation of `imageInputKey` and adding payload-shape tests.

murdore:
Two additions from continued live verification against Replicate schemas (2026-07-12):

1. **`resolution` type is too narrow for pass-through.** `minimax/hailuo-2.3-fast` enumerates `["768p","1080p"]` — the current `VideoOutputOptions.resolution: "720p" | "1080p"` cannot express `768p`, so TS callers can't satisfy that model's schema even with `imageInputKey`. Suggest widening to `"720p" | "1080p" | (string & {})` (same pattern the type already uses for `length`). Its `duration` is also an enum `[6,10]` — already expressible via `length`.

2. **Another consumer for `imageInputKey`:** `kwaivgi/kling-v2.1` requires the image under `start_image` (`duration` enum `[5,10]`). Notably its `start_image` is `format: uri` and Replicate accepts data URIs for file inputs — which makes Kling reachable through this handler with inline images, avoiding the PiAPI `KlingVideoHandler`'s publicly-accessible-URL requirement entirely.

Happy to fold the resolution-type widening into this PR if preferred — kept it out to stay surgical.

murdore:
**Live end-to-end validation** (local mirror of this PR + #1157, Node 24, real Replicate account): `minimax/hailuo-2.3-fast` with `imageInputKey: 'first_frame_image'` generated **and downloaded** a real 5.9s clip through `nl.generate` in 110.6s (`ok: true`, 1.66 MB). `kwaivgi/kling-v2.1` with `imageInputKey: 'start_image'` and an **inline data-URI** generated a real 5.0s clip (5.0 MB) — confirming data URIs satisfy its `format: uri` field, so Kling is reachable through this handler with no public-URL upload step. Both fixes compose; without #1157 the download step fails on modern Node.

Tara-ag (approved):
## Review Summary

**Files reviewed:** 2
- `src/lib/adapters/video/replicateVideoHandler.ts`
- `src/lib/types/multimodal.ts`

**Assessment:**
- The change is backward compatible: omitting `imageInputKey` produces the identical legacy payload.
- The new optional field is additive to the public `VideoOutputOptions` type, so existing SDK callers are unaffected.
- No hardcoded secrets, injection risks, or architectural rule violations (registry, CLI/SDK separation, provider error formatting) were introduced.
- Existing automated review comments already cover the main remaining concerns: runtime validation of `imageInputKey`, model-specific normalization of `duration`/`resolution`, collision guards for computed payload keys, and adding tests for both payload branches.

No new blocking issues were identified. Approving.

Issue #635, opened 2025-12-05
[Feature]: Add Structured output support for handling simple and complex schemas with tool calling


YaswanthKurapati24:
### Is your feature request related to a problem?

Refer to the below solution documentation to understand about adding the support for structured object generation.
It also explains about the current limitations and approach to be followed.

 

Summary
This document outlines the solution for achieving 100% reliable structured output from AI models while maintaining tool calling support.
{panel}

Problem Statement
1. NeuroLink Limitation
NeuroLink 8.0.1 does not support structured output enforcement

The {{schema}} parameter exists in the interface but is unused

Results are not validated against any schema

2. Vercel AI SDK {{experimental_output}} Limitation
Unreliable with LLM models - throws {{NoObjectGeneratedError}}

Root Cause: Models wrap JSON in markdown fences and conversational text

Failure Rate: Increases with schema complexity
** Simple schemas (1-3 fields): 5-10% failure
** Medium schemas (4-10 fields): 10-15% failure
** Complex schemas (>10 fields): 15-25% failure
** Very complex schemas: 25-35% failure

Why This Happens:

LLM models add helpful context like:

Here's the structured analysis:{ "data": "..." } This should help!
The AI SDK parser expects pure JSON, not markdown-wrapped responses.

Solution: Smart 3-Phase Fallback Pattern
Goal: 100% success rate with 40-60% cost savings compared to naive approach

flowchart TD
  Start([Start]) --> Phase1[PHASE 1: Fast Path - Call generateText with schema in prompt]
  Phase1 --> Parse[Parse and match schema]

  Parse -->|Success 85-90 percent| Success[100% Success]
  Parse -->|Failure 10-15 percent| Valid{Valid JSON?}

  Valid -->|Yes| Phase2[PHASE 2: Light Repair\nOptional]
  Valid -->|No| Phase3[PHASE 3: Full Regen\nRequired]

  Phase2 --> Success
  Phase3 --> Success

Strategy: Start with fast path, fallback to guaranteed methods if needed

 

Method: Use {{generateText()}} with schema injected in system prompt

Process:

Convert Zod schema → JSON Schema

Inject JSON schema into system prompt with strict formatting rules

Call {{generateText()}} with tools (if needed)

Extract JSON from response (handles markdown fences)

Validate against Zod schema

Advantages:

✓ Supports tool calling

✓ Single API call (fast)

✓ Cost-effective (1x baseline)

✓ Handles 85-90% of cases successfully

[...]

PR #1152, opened 2026-07-12
fix(json): unwrap single-element array wrapper for object schemas

murdore:
## What
On the native-Anthropic structured-output path under escaping stress, models sometimes return `[{...}]` instead of `{...}` for an **object** schema. Because a JS array is `typeof "object"`, `coerceJsonToSchema` accepted it as `firstValid` and returned an **array** as `structuredData`, failing the caller's object-shaped schema. Caught adversarially during the audit's live `test:json-e2e` run (#635).

## Fix
When a candidate array does not itself satisfy the schema, and it holds exactly one non-array object element that **does** satisfy it, unwrap to that element. Gated by `safeParse`, mirroring the existing nested-string unwrap:
- array-typed schemas keep their array (no over-unwrap)
- multi-element arrays are left untouched (never silently reduced to their first element)

## Proof
`coerce-nested-unwrap` **9/9** (3 new tests), `structured-coerce` **7/7**, `json` **21/21** — no regressions.

Refs #635 (feature already shipped; this hardens the fallback edge case).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved structured data handling when valid object data is wrapped in a single-element array.
  * Preserved array results when the schema expects an array or when multiple elements are present.

* **Tests**
  * Added coverage for single-element arrays, array schemas, and multi-element arrays.

### DIFF
--- a/src/lib/utils/json/coerce.ts
+++ b/src/lib/utils/json/coerce.ts
@@ -260,6 +260,24 @@ export function coerceJsonToSchema(
           repaired: true,
           truncated: candidate.truncated,
         });
+      } else if (
+        // Single-element array wrapper: for an OBJECT schema, models sometimes
+        // return `[{...}]` instead of `{...}` (seen on the native Anthropic
+        // path under escaping stress). Unwrap a lone object element and
+        // re-validate — the safeParse gate rejects an incorrect unwrap, so an
+        // array schema (which validates the array directly above) is untouched.
+        Array.isArray(outcome.value) &&
+        outcome.value.length === 1 &&
+        outcome.value[0] !== null &&
+        typeof outcome.value[0] === "object" &&
+        !Array.isArray(outcome.value[0]) &&
+        safeParseable.safeParse(outcome.value[0]).success
+      ) {
+        schemaValid.push({
+          value: outcome.value[0],
+          repaired: true,
+          truncated: candidate.truncated,
+        });
       }
     }
   }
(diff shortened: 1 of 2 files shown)


Tara-ag (approved):
## Review Summary

**Files reviewed:** 2
- `src/lib/utils/json/coerce.ts`
- `test/continuous-test-suite-coerce-nested-unwrap.ts`

**Decision:** Approve

The fix is narrow, well-gated by `safeParse`, and correctly avoids over-unwrapping array schemas and multi-element arrays. The new tests cover the stated native-Anthropic `[{...}]` regression case and the anti-regression cases.

I did not raise any new inline comments because the two substantive points I would have flagged are already covered by the existing Copilot review threads:
- The array-wrapper branch validates `outcome.value[0]` before `deepUnwrapJsonStrings` is applied, so a wrapped object that only becomes schema-valid after nested-string unwrapping is still returned as an array.
- The test suite lacks coverage for that combined wrapper + nested-string case.

Those are worth addressing, but per review protocol I am not re-raising already-commented points as blocking criteria. No new security, architectural, or backward-compatibility concerns were introduced.

Tara-ag (approved):
## Review Summary

**Files reviewed:** 2
- `src/lib/utils/json/coerce.ts`
- `test/continuous-test-suite-coerce-nested-unwrap.ts`

I analyzed the single-element array unwrap logic added to `coerceJsonToSchema`. The change is narrowly scoped, gated by `safeParse`, and correctly preserves array schemas and multi-element arrays. The regression tests cover the motivating native-Anthropic `[{...}]` case, the no-over-unwrap array-schema case, and the multi-element rejection case.

I noted the two existing unresolved review comments from the automated reviewer regarding the combined wrapper + nested-string-unwrap scenario. Those are already raised and should be addressed by the author; I am not duplicating them here.

No blocking issues (security, backward compatibility, CLAUDE.md critical-rule violations, or 3+ MAJOR issues) were found in this run. Approving.

PR #1157, opened 2026-07-12
fix(utils): safeFetch pinned-agent lookup honors options.all (autoSelectFamily)

murdore:
## Problem

Every `safeDownload` of a provider asset (Replicate prediction outputs, Runway/Kling media) fails with a bare `fetch failed` on modern Node — while `curl` of the same URL succeeds. Observed live twice today: Replicate predictions **completed** (billed, output URL valid) but `downloadPredictionOutput` failed on the `replicate.delivery` URL both times.

## Root cause

Node ≥20 enables `autoSelectFamily` (Happy Eyeballs) by default; its `net` lookup contract passes `all: true` and expects an **array of `{address, family}`**. `buildPinnedAgent`'s lookup always answers in string form → undici fails the connect with `Invalid IP address: undefined`.

## Minimal repro (Node 24, undici, against replicate.delivery)

```
string-form callback → FETCH FAILED: Invalid IP address: undefined
all-aware callback   → HTTP 200
```

(Agent with `connect.lookup` pinning a resolved IP; identical setup, only the callback shape differs.)

## Fix

The `options.all` branch returns `[{ address: ip, family }]`; the legacy branch is byte-for-byte unchanged. SSRF pinning semantics are identical — both forms hand the connection exactly the validated IP.

## Relation to other PRs

Third in today's series from wiring juspay/director's draft-tier b-roll through NeuroLink: #1147 (js-yaml runtime dep), #1150 (Replicate `imageInputKey`), and this one. With #1150 locally mirrored plus this fix, `minimax/hailuo-2.3-fast` and `kwaivgi/kling-v2.1` (inline data-URI `start_image`) generate and download end-to-end on Replicate.

`pnpm run typecheck` clean; pre-commit hook suite green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved download reliability when connecting to pinned IP addresses.
  * Prevented connection failures caused by updated network lookup behavior.
  * Continued enforcing download timeouts, redirect rejection, and response size limits.

### DIFF
--- a/src/lib/utils/safeFetch.ts
+++ b/src/lib/utils/safeFetch.ts
@@ -32,7 +32,7 @@ const DEFAULT_TIMEOUT_MS = 60_000;
 function buildPinnedAgent(hostname: string, ip: string, family: 4 | 6): Agent {
   return new Agent({
     connect: {
-      lookup: (host, _options, callback) => {
+      lookup: (host, options, callback) => {
         if (host.toLowerCase() !== hostname.toLowerCase()) {
           // The host the connect layer asks for differs from the URL host —
           // this happens for absolute Host headers etc. Reject defensively.
@@ -43,6 +43,16 @@ function buildPinnedAgent(hostname: string, ip: string, family: 4 | 6): Agent {
             "",
             0,
           );
+          return;
+        }
+        // Node ≥20 enables autoSelectFamily (Happy Eyeballs) by default, and
+        // its lookup contract passes `all: true` expecting an address ARRAY.
+        // Answering with the string form there fails the connection with
+        // "Invalid IP address: undefined" — which broke every safeDownload
+        // (Replicate / Runway / Kling asset fetches) on modern Node while
+        // plain curl of the same URL succeeded.
+        if (options?.all) {
+          callback(null, [{ address: ip, family }]);
           return;
         }
         callback(null, ip, family);


murdore:
**Live end-to-end validation** (local mirror of this fix + #1150, Node 24): with the `options.all` branch in place, `downloadPredictionOutput` succeeded on a real `replicate.delivery` URL inside `nl.generate` (5.9s / 1.66 MB clip, `ok: true`) — the same call that failed with `fetch failed` twice before the patch.

Tara-ag (approved):
## Review Summary

**Files reviewed:** 1 (`src/lib/utils/safeFetch.ts`)
**New issues raised:** 0
**Decision:** Approve

### Analysis

The change in `buildPinnedAgent` correctly handles Node ≥20’s `autoSelectFamily` (Happy Eyeballs) lookup contract. When undici passes `options.all: true`, the callback must return an array of `{ address, family }` objects; returning the legacy string form causes `Invalid IP address: undefined` and breaks every `safeDownload` on modern Node.

**Security:** No regression. Both branches hand the connection the same pre-validated IP from `validateAndResolveUrl`, and the hostname-mismatch guard still runs first. SSRF pinning semantics are preserved.

**Backward compatibility:** The legacy `callback(null, ip, family)` branch is untouched, so older Node/undici paths continue to work.

**Architecture:** No CLAUDE.md critical rules are violated. The change stays within the existing utility and does not introduce static provider imports, CLI/SDK leakage, or error-formatter throws.

**Testing:** The existing `test/continuous-test-suite-ssrf.ts` covers `safeDownload` SSRF propagation. A dedicated regression test for the `options.all` array-shape branch would be a nice follow-up but is not blocking given the difficulty of deterministically triggering Happy Eyeballs in a no-network test and the targeted nature of the fix.

Clean, minimal bug fix — good to merge.

Tara-ag (approved):
## Review Summary

**Files reviewed:** 1 (`src/lib/utils/safeFetch.ts`)
**New issues raised:** 0
**Existing review comments:** 0

### Assessment

This is a focused, correct fix for the Node ≥20 `autoSelectFamily` lookup contract.

- The `options.all` branch returns `[{ address: ip, family }]` as required by Happy Eyeballs, while the legacy string-form branch is preserved unchanged.
- SSRF pinning semantics are unchanged: the validated IP is still the only address handed to the connection.
- The defensive host-mismatch rejection path remains intact.
- Typecheck and pre-commit hooks are reported clean.

No blocking issues. Approved.

### Minor note (non-blocking)

A regression test that exercises the `lookup` callback with `options.all = true` would help protect this behavior across Node version upgrades, though testing undici’s internal `connect.lookup` path can be awkward to mock.

PR #1158, opened 2026-07-12
fix(tools): sandbox built-in file tools to the working directory

### LINKED ISSUE #1004: [Bug]: buildin tools are not sandboxed
### What happened?

Buildin tools
readFile / listDirectory / writeFile should be sandboxed

they are enabled by default, and there power is ab it unclear.

I had an agent scanning basically the whole filesystem extracting several access tokens. ( inside a container)

It is fairly easy to missconfigure MCP's and an agent can easily combine them.

It would probably be a good idea to have configuration and MCP execution as much as possible segregated.

And only inject neceessary configuration to each MCP process.

for filesystem access memfs could be used
```
import { fs, vol } from 'memfs';
```

allowing to load/persist a folder from the host os and only give in memory access to MCP's

For the BashTool and mcp's in general it would also be nice to isolate `execFile` and related .

execFile is a bit of an unsolved thing in nodejs, with denno you can sandbox it

It could be possible to use `unshare` to run the executable isolated inside a container or on a host.

```
execFile('unshare', ['--map-root-user', '--net', '--mount', 'my-command']);
```

### AI Provider

OpenAI

### Component

SDK

### Steps to Reproduce

Wire up the SDK and run a prompt, can be triggered as soon as the agent tries to use the tools.

### Environment

```markdown
- OS: linux
- Node.js 25:
- NeuroLink latest:
- Package manager: pnpm
```

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

murdore:
Fixes #1004. readFile/writeFile guard (`!startsWith(cwd) && !isAbsolute`) was always false for absolute paths — arbitrary FS read/write; listDirectory/analyzeCSV had no guard. Added resolveWithinCwd() containment (blocks absolute escape + ../ traversal, sibling-prefix-safe) to all four. Regression test; bugfixes 90/90.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved filesystem protection by blocking attempts to access files outside the working directory.
  * Prevented absolute-path and parent-directory traversal access through file tools and CSV analysis.
  * Preserved access to permitted files within the working directory.

* **Tests**
  * Added regression coverage for blocked and permitted filesystem access scenarios.

### DIFF
--- a/test/continuous-test-suite-bugfixes.ts
+++ b/test/continuous-test-suite-bugfixes.ts
@@ -18,6 +18,7 @@ import {
 
 import { convertToModelMessages } from "../src/lib/utils/messageBuilder.js";
 import { CSVProcessor } from "../src/lib/utils/csvProcessor.js";
+import { directAgentTools } from "../src/lib/agent/directTools.js";
 
 import {
   GoogleVertexProvider,
@@ -193,6 +194,38 @@ const tests: TestFunction[] = [
         typeof raw.content === "string" &&
         raw.content.includes('"line1\rline2"')
       );
+    },
+  },
+  // ---------- Built-in file tools are sandboxed to cwd (issue #1004) ----------
+  {
+    name: "directAgentTools: readFile/writeFile/listDirectory deny paths outside cwd",
+    category: "tool-sandbox",
+    fn: async () => {
+      type Exec = {
+        execute: (a: unknown) => Promise<{ success: boolean; error?: string }>;
+      };
+      const rf = directAgentTools.readFile as unknown as Exec;
+      const wf = directAgentTools.writeFile as unknown as Exec;
+      const ld = directAgentTools.listDirectory as unknown as Exec;
+      const denied = (r: { success: boolean; error?: string }) =>
+        r.success === false && /Access denied/.test(r.error ?? "");
+      // Absolute-path escape and ../ traversal must be denied on all three.
+      const results = await Promise.all([
+        rf.execute({ path: "/etc/passwd" }),
+        rf.execute({ path: "../../../../../../etc/passwd" }),
+        ld.execute({ path: "/etc" }),
+        wf.execute({
+          path: "/tmp/neurolink-sandbox-escape.txt",
+          content: "x",
+          mode: "overwrite",
+        }),
+      ]);
+      if (!results.every(denied)) {
+        return false;
+      }
+      // A path inside cwd must still be allowed.
+      const ok = await rf.execute({ path: "package.json" });
+      return ok.success === true;
     },
   },
   // ---------- Bug 1: Vertex location routing via resolveVertexLocation ----------
--- a/src/lib/agent/directTools.ts
+++ b/src/lib/agent/directTools.ts
@@ -26,6 +26,31 @@ function truncateOutput(output: string): string {
   return output;
 }
 
+/**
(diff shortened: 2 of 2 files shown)


test/continuous-test-suite-bugfixes.ts
```diff
@@ -195,6 +196,38 @@ const tests: TestFunction[] = [
...
+      type Exec = {
+        execute: (a: unknown) => Promise<{ success: boolean; error?: string }>;
+      };
+      const rf = directAgentTools.readFile as unknown as Exec;
+      const wf = directAgentTools.writeFile as unknown as Exec;
+      const ld = directAgentTools.listDirectory as unknown as Exec;
+      const denied = (r: { success: boolean; error?: string }) =>
+        r.success === false && /Access denied/.test(r.error ?? "");
+      // Absolute-path escape and ../ traversal must be denied on all three.
+      const results = await Promise.all([
```
Tara-ag, line 213: 💬 **SUGGESTION: add a sibling-prefix traversal case**

The helper comment in `directTools.ts` explicitly calls out `cwd + path.sep` as the defense against sibling-prefix bypasses (`/home/app` vs `/home/app-evil`). The current test cases cover absolute escape and `../` traversal, but none verify that a path resolving to a sibling of the current directory is denied.

Suggestion: add a case such as `../${path.basename(process.cwd())}-evil/package.json` (or an absolute sibling path) and assert it returns `Access denied` rather than a file-not-found error, locking in the prefix-check behavior.

test/continuous-test-suite-bugfixes.ts
```diff
@@ -195,6 +196,38 @@ const tests: TestFunction[] = [
       );
     },
   },
+  // ---------- Built-in file tools are sandboxed to cwd (issue #1004) ----------
```
Tara-ag, line 199: 💡 **MINOR: analyzeCSV sandboxing is not exercised by the regression test**

The PR title states the sandbox was added to all four file tools (`readFile` / `writeFile` / `listDirectory` / `analyzeCSV`) and includes a regression test, but this test only covers the first three. Since `analyzeCSV` previously accepted arbitrary absolute paths and now goes through `resolveWithinCwd`, it should have the same denial/allow assertions to prevent future regressions of issue #1004.

Suggestion: add `directAgentTools.analyzeCSV` cases for an absolute escape (`/etc/passwd`) and a cwd-relative CSV (e.g. `package.json` or a fixture) to the `Promise.all` and positive checks.

Tara-ag (approved):
## Review Summary

**Files reviewed:** 2 (`src/lib/agent/directTools.ts`, `test/continuous-test-suite-bugfixes.ts`)

- 💡 MINOR: `analyzeCSV` sandboxing is not covered by the regression test (`test/continuous-test-suite-bugfixes.ts:199`).
- 💬 SUGGESTION: Add a sibling-prefix traversal case to lock in the `cwd + path.sep` defense (`test/continuous-test-suite-bugfixes.ts:213`).

**Decision:** APPROVE — no new blocking issues introduced by this change.

**Note:** There are pre-existing unresolved review comments from `copilot-pull-request-reviewer` on `src/lib/agent/directTools.ts` regarding the root-directory `startsWith` edge case, symlink-based sandbox escapes, and the `analyzeCSV` Zod schema/description mismatch. These are not duplicated here per review workflow rules, but the author should address them before merging.

Issue #348, opened 2025-12-01
MB-017: Missing MIME Type Format Validation


murdore:
## Summary
MIME types extracted from data URIs not validated, allowing invalid strings like `image/foo` to be passed to providers.

## Root Cause
Extraction without validation (line 890): `mimeType = match[1];` accepts any string.

## Impact
- Invalid MIME types → provider API rejections
- Difficult debugging
- No guidance on supported formats
- Compounds MB-001 issues

## Reproduction
Data URI: `data:image/invalid-format;base64,abc`
**Expected:** Validation error
**Actual:** Sent to provider, cryptic error

## Fix
Add validation:

```typescript
const VALID_IMAGE_MIME_TYPES = ['image/jpeg', 'image/png', ...];

function validateImageMimeType(mimeType: string): boolean {
  return VALID_IMAGE_MIME_TYPES.includes(mimeType);
}

// Usage:
mimeType = getValidatedMimeType(match[1], 'image/jpeg');
```

**File:** `src/lib/utils/messageBuilder.ts`
**Lines:** 886-894

**Depends on:** MB-001

## Acceptance Criteria
- [ ] MIME type validation
- [ ] Fallback to valid type
- [ ] Warning logged
- [ ] List of supported formats

murdore:
Fixed by #1736 (`37ecac619`), merged to `release`.

Image MIME subtypes are validated at intake and the HTML-vs-markup distinction is explained in the rejection message.

Issue #330, opened 2025-12-01
MB-012: Case-Sensitive URL Detection Fails on Uppercase


murdore:
## Summary
`isInternetUrl()` uses case-sensitive comparison, failing to recognize `HTTP://` or `HTTPS://`.

## Root Cause
`input.startsWith("http://")` is case-sensitive. URLs have case-insensitive protocols per RFC 3986.

## Impact
- `HTTP://example.com/image.png` not recognized
- Treated as file path → file not found
- Inconsistent with browser behavior
- Poor robustness

## Reproduction
Pass: `["HTTP://example.com/image.png"]`
**Expected:** Downloaded
**Actual:** File not found error

## Fix
Add lowercase conversion:

```typescript
function isInternetUrl(input: string): boolean {
  const lower = input.toLowerCase();
  return lower.startsWith("http://") || lower.startsWith("https://");
}
```

**File:** `src/lib/utils/messageBuilder.ts`
**Line:** 765-767

## Acceptance Criteria
- [ ] Case-insensitive detection
- [ ] All HTTP/HTTPS variations work

murdore:
🔍 **Verification sweep result: already fixed in current `release` — closing.**

This is already fixed. `isInternetUrl()` in `src/lib/utils/messageBuilder.ts` (lines 1611-1616) now lowercases the input before checking the scheme prefix:

```typescript
function isInternetUrl(input: string): boolean {
  const lower = input.toLowerCase();
  return lower.startsWith("http://") || lower.startsWith("https://");
}
```

This matches the fix proposed in the issue. `HTTP://example.com/image.png` and other case variants are now correctly detected as internet URLs instead of falling through to the file-path branch. Closing as completed.

Issue #270, opened 2025-12-01
MB-003: Malformed Data URI Fallback Causes Silent Failures


murdore:
## Summary
Malformed data URIs fail regex validation but code continues silently, causing cryptic API errors later.

## Root Cause
Data URI parsing (line 888-894) doesn't throw error when regex match fails, allowing invalid strings to propagate.

## Impact
- Malformed data URIs → cryptic provider API errors
- Difficult debugging (error far from source)
- No guidance on correct format
- Wasted API calls on invalid requests

## Reproduction
1. Malformed URI: `const badUri = "data:image/png,base64,abc"` (missing semicolon)
2. Pass to SDK: `generate({ input: { text: "Describe", images: [badUri] } })`
3. **Expected:** Clear error about invalid format
4. **Actual:** Silent failure or cryptic API error

## Fix
Add error throwing in else block after regex match failure:

```typescript
} else {
  throw new Error(
    `Invalid data URI format. Expected: data:<mimeType>;base64,<data>`
  );
}
```

**File:** `src/lib/utils/messageBuilder.ts`
**Lines:** 886-894

## Acceptance Criteria
- [ ] Malformed data URI → clear error
- [ ] Error includes expected format
- [ ] Valid data URIs still work

murdore:
🔍 **Verification sweep result: already fixed in current `release` — closing.**

This is fixed. In `processImageToBase64()` (src/lib/utils/messageBuilder.ts, the function called from the live image-processing path at line 1964), the branch that runs after the data-URI regex fails to match now throws instead of silently falling through:

```js
} else {
  // #270: a malformed data: URI must fail loudly, not silently pass the
  // raw string through as if it were valid base64 ...
  throw new Error(
    `Malformed image data URI at index ${index} (expected "data:<image/...>;base64,<data>")`,
  );
}
```

This matches the requested fix and satisfies all acceptance criteria (clear error, states expected format, valid URIs unaffected). Landed in commit 746eab54 ("fix(message-builder): image input download + data-URI safety"), which explicitly references this issue in its comment.</closeComment>

PR #1161, opened 2026-07-12
fix(message-builder): image input download + data-URI safety

### LINKED ISSUE #334: MB-014: Chunks Array Accumulation Without Size Checking
## Summary
Image download accumulates all chunks before size check, allowing unbounded memory growth.

## Root Cause
Loop accumulates chunks, size checked AFTER concat (line 807-813). 1GB file downloads fully before failing.

## Impact
- Malicious URLs → OOM before size check
- Wasted bandwidth (full download before rejection)
- Server memory exhaustion risk
- DoS vulnerability

## Reproduction
URL to 1GB file
**Expected:** Fail at 10MB
**Actual:** Downloads all 1GB, then fails

## Fix
Check size during loop:

```typescript
let totalSize = 0;
for await (const chunk of response.body) {
  totalSize += chunk.length;
  if (totalSize > maxSize) {
    throw new Error(`Image too large: ${totalSize} bytes...`);
  }
  chunks.push(chunk);
}
```

**Lines:** 800-813

## Security Impact
Prevents memory exhaustion attacks via oversized image URLs.

## Acceptance Criteria
- [ ] Fail-fast during download
- [ ] No full download of oversized files
- [ ] Memory protection

murdore:
Fixes #334 #270 #348 #330. Incremental 10MB cap while streaming (was after Buffer.concat → unbounded memory); reject malformed data URIs loudly; validate image/* MIME; case-insensitive URL scheme. Regression test; bugfixes 90/90.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of image URLs regardless of letter casing.
  * Prevented oversized remote images over 10 MB from being downloaded.
  * Added clearer validation and error messages for malformed or unsupported image data.
  * Ensured only valid image data URIs are accepted during message preparation.

* **Tests**
  * Added coverage for valid, malformed, and unsupported image data URIs.

### DIFF
--- a/src/lib/utils/messageBuilder.ts
+++ b/src/lib/utils/messageBuilder.ts
@@ -1609,7 +1609,10 @@ async function convertContentToProviderFormat(
  * Check if a string is an internet URL
  */
 function isInternetUrl(input: string): boolean {
-  return input.startsWith("http://") || input.startsWith("https://");
+  // Scheme is case-insensitive (RFC 3986) — "HTTPS://..." must still be a URL,
+  // not fall through to the file-path branch and produce a confusing error.
+  const lower = input.toLowerCase();
+  return lower.startsWith("http://") || lower.startsWith("https://");
 }
 
 /**
@@ -1656,20 +1659,23 @@ async function downloadImageFromUrl(url: string): Promise<string> {
       );
     }
 
-    // Read the response body
+    // Read the response body, enforcing the size cap INCREMENTALLY: a
+    // misbehaving/malicious server on a user-supplied URL must not be able to
+    // force unbounded memory growth by streaming gigabytes before we ever
+    // check the total (the previous code concat'd everything first).
+    const maxSize = 10 * 1024 * 1024; // 10MB
     const chunks: Buffer[] = [];
+    let totalSize = 0;
     for await (const chunk of response.body) {
+      totalSize += chunk.length;
+      if (totalSize > maxSize) {
+        throw new Error(
+          `Image too large: exceeds ${maxSize} bytes while downloading from ${url}`,
+        );
+      }
       chunks.push(chunk);
     }
     const buffer = Buffer.concat(chunks);
-
-    // Check file size (limit to 10MB)
-    const maxSize = 10 * 1024 * 1024; // 10MB
-    if (buffer.length > maxSize) {
-      throw new Error(
-        `Image too large: ${buffer.length} bytes (max: ${maxSize} bytes)`,
-      );
-    }
 
     // Convert to base64 data URI
     const base64 = buffer.toString("base64");
@@ -1825,10 +1831,23 @@ function processImageToBase64(
       // Data URI (including downloaded URLs) - extract mime type and raw base64
       const match = image.match(/^data:([^;]+);base64,(.+)$/);
       if (match) {
-        mimeType = match[1];
+        const declaredMime = match[1];
+        // #348: only accept image/* data URIs; reject a non-image MIME before
+        // it reaches a provider API rather than passing it through unchecked.
+        if (!declaredMime.startsWith("image/")) {
+          throw new Error(
+            `Unsupported data URI MIME type for image input at index ${index}: "${declaredMime}" (expected image/*)`,
+          );
+        }
+        mimeType = declaredMime;
         imageData = match[2]; // Raw base64 only — NOT the full data: URI
       } else {
-        imageData = image;
+        // #270: a malformed data: URI must fail loudly, not silently pass the
+        // raw string through as if it were valid base64 (which corrupts the
+        // request and surfaces as an opaque provider error later).
+        throw new Error(
+          `Malformed image data URI at index ${index} (expected "data:<image/...>;base64,<data>")`,
+        );
       }
     } else if (isInternetUrl(image)) {
       // This should not happen as URLs are processed separately
--- a/test/continuous-test-suite-bugfixes.ts
+++ b/test/continuous-test-suite-bugfixes.ts
@@ -16,7 +16,10 @@ import {
   parseRetryAfterMs,
 } from "../src/lib/proxy/routingPolicy.js";
 
-import { convertToModelMessages } from "../src/lib/utils/messageBuilder.js";
+import {
+  convertToModelMessages,
+  buildMultimodalMessagesArray,
+} from "../src/lib/utils/messageBuilder.js";
 import { CSVProcessor } from "../src/lib/utils/csvProcessor.js";
 import { PDFProcessor } from "../src/lib/utils/pdfProcessor.js";
 import { directAgentTools } from "../src/lib/agent/directTools.js";
@@ -546,6 +549,55 @@ const tests: TestFunction[] = [
       return (
         Object.keys(qr[0] ?? {}).join(",") === "name,note" && qr.length === 2
       );
+    },
+  },
+  // ---------- Image data-URI validation (issues #270, #348) ----------
+  {
+    name: "buildMultimodalMessagesArray rejects malformed / non-image data URIs",
+    category: "message-builder",
+    fn: async () => {
+      const build = (images: string[]) =>
+        buildMultimodalMessagesArray(
+          {
+            input: { text: "hi", images },
+          } as unknown as Parameters<typeof buildMultimodalMessagesArray>[0],
+          "openai",
+          "gpt-4o",
+        );
+      const throwsWith = async (
+        images: string[],
+        re: RegExp,
+      ): Promise<boolean> => {
+        try {
+          await build(images);
+          return false;
+        } catch (e) {
+          return e instanceof Error && re.test(e.message);
+        }
+      };
+      // #270: a malformed data URI must fail loudly, not pass through silently.
+      if (
+        !(await throwsWith(
+          ["data:image/png;NOTBASE64whoops"],
+          /Malformed image data URI/,
+        ))
+      ) {
+        return false;
+      }
+      // #348: a non-image MIME data URI must be rejected.
+      if (
+        !(await throwsWith(
+          ["data:text/plain;base64,aGVsbG8="],
+          /Unsupported data URI MIME/,
+        ))
+      ) {
+        return false;
+      }
+      // A valid image data URI must still build.
+      const png =
+        "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNk+M9QDwADhgGAWjR9awAAAABJRU5ErkJggg==";
+      const msgs = await build([`data:image/png;base64,${png}`]);
+      return Array.isArray(msgs);
     },
   },
   // ---------- Bug 1: Vertex location routing via resolveVertexLocation ----------


src/lib/utils/messageBuilder.ts
```diff
@@ -1825,10 +1831,23 @@ function processImageToBase64(
       // Data URI (including downloaded URLs) - extract mime type and raw base64
       const match = image.match(/^data:([^;]+);base64,(.+)$/);
       if (match) {
-        mimeType = match[1];
+        const declaredMime = match[1];
+        // #348: only accept image/* data URIs; reject a non-image MIME before
+        // it reaches a provider API rather than passing it through unchecked.
+        if (!declaredMime.startsWith("image/")) {
```
Tara-ag, line 1837: 💡 Data URI schemes are case-insensitive per RFC 2397, but this guard (and the regex on the next line) are case-sensitive. A valid `DATA:image/png;base64,...` input falls through to the file-path branch and fails with a confusing "Image file not found" error instead of being treated as a data URI.

Normalize the scheme check and add the `i` flag to the regex so uppercase schemes are handled consistently with the `isInternetUrl` fix above:

```typescript
if (image.toLowerCase().startsWith("data:")) {
  const match = image.match(/^data:([^;]+);base64,(.+)$/i);
  if (match) {
    const declaredMime = match[1].toLowerCase().trim();
    ...
```

Consider also adding a regression test for `DATA:image/png;base64,...` alongside the existing lowercase cases.

src/lib/utils/messageBuilder.ts
```diff
@@ -1609,7 +1609,10 @@ async function convertContentToProviderFormat(
  * Check if a string is an internet URL
  */
 function isInternetUrl(input: string): boolean {
-  return input.startsWith("http://") || input.startsWith("https://");
+  // Scheme is case-insensitive (RFC 3986) — "HTTPS://..." must still be a URL,
```
Tara-ag, line 1612: 💡 This lowercases the entire `input` string just to compare an 8-byte prefix. For large data-URI or file-path strings this allocates an unnecessary copy. Since the goal is only the scheme, slice first:

```typescript
const prefix = input.slice(0, 8).toLowerCase();
return prefix.startsWith("http://") || prefix.startsWith("https:");
```

This keeps the RFC 3986 fix while avoiding O(n) work on every image string.

test/continuous-test-suite-bugfixes.ts
```diff
@@ -195,6 +198,55 @@ const tests: TestFunction[] = [
...
+          "gpt-4o",
+        );
+      const throwsWith = async (
+        images: string[],
+        re: RegExp,
+      ): Promise<boolean> => {
+        try {
+          await build(images);
+          return false;
+        } catch (e) {
```
Tara-ag, line 574: 💡 The regression test only exercises lowercase data URIs. Since the production code now has scheme/MIME case-sensitivity gaps (see messageBuilder.ts comments), add cases for:

- Uppercase scheme: `DATA:image/png;base64,...`
- Uppercase MIME: `data:IMAGE/PNG;base64,...`
- MIME with whitespace: `data: image/png;base64,...`

This prevents the fixes from regressing on RFC-compliant but differently-cased inputs.

Tara-ag (commented):
## Review Summary

Reviewed 2 changed files in PR #1161.

### Existing comments
3 inline comments were already posted by copilot-pull-request-reviewer (stream abort on size-cap exceed, MIME case normalization, content-type parameter stripping). I did not duplicate those.

### New issues raised (3 SUGGESTIONs)
- `src/lib/utils/messageBuilder.ts:1612` — avoid full-string `toLowerCase()` for scheme checks; slice the prefix instead.
- `src/lib/utils/messageBuilder.ts:1837` — data URI scheme/MIME validation is case-sensitive; uppercase `DATA:` or `IMAGE/PNG` inputs fall through to the file-path branch.
- `test/continuous-test-suite-bugfixes.ts:221` — regression test should cover case-insensitive scheme/MIME variants.

None of these meet the blocking criteria (no secrets, no critical security vulnerabilities, no backward-incompatible SDK API changes, no non-negotiable CLAUDE.md rule violations, and fewer than 3 MAJOR issues). The PR addresses real bugs (#270, #348, #330, #334) and adds regression coverage, so I’m leaving the suggestions as non-blocking feedback.

### Next steps
Consider addressing the case-sensitivity and performance suggestions; otherwise good to merge.

src/lib/utils/messageBuilder.ts
Tara-ag, line 1837: ⚠️ **MAJOR** — Data-URI scheme and MIME are case-sensitive

RFC 2397 / RFC 2045 make the `data:` scheme and MIME type case-insensitive. The current guard `image.startsWith("data:")` and regex `/^data:([^;]+);base64,$/` reject valid inputs such as `DATA:image/png;base64,...` or `data:IMAGE/PNG;base64,...`. Those inputs then fall through to the file-path branch and fail with a confusing "Image file not found" error.

Normalize the scheme check and add the `i` flag to the regex, then lowercase and trim the declared MIME before validating:

```typescript
if (image.toLowerCase().startsWith("data:")) {
  const match = image.match(/^data:([^;]+);base64,(.+)$/i);
  if (match) {
    const declaredMime = match[1].toLowerCase().trim();
    if (!declaredMime.startsWith("image/")) {
      throw ErrorFactory.invalidImageFormat();
    }
    mimeType = declaredMime;
    imageData = match[2];
  } else {
    throw ErrorFactory.invalidImageFormat();
  }
}
```

Please also add regression tests for uppercase scheme and uppercase MIME (see existing test comment at `test/continuous-test-suite-bugfixes.ts:574`).

src/lib/utils/messageBuilder.ts
Tara-ag, line 1612: 💡 **MINOR** — Avoid full-string lowercase for an 8-byte prefix check

`input.toLowerCase()` copies the entire string just to compare an 8-byte prefix. For large data-URIs or path strings this is unnecessary O(n) allocation/work on every image string.

Slice first, then lowercase:

```typescript
const prefix = input.slice(0, 8).toLowerCase();
return prefix.startsWith("http://") || prefix.startsWith("https:");
```

This preserves the RFC 3986 case-insensitive fix while avoiding the extra copy.

src/lib/utils/messageBuilder.ts
```diff
@@ -1656,21 +1659,24 @@ async function downloadImageFromUrl(url: string): Promise<string> {
...
+    // misbehaving/malicious server on a user-supplied URL must not be able to
+    // force unbounded memory growth by streaming gigabytes before we ever
+    // check the total (the previous code concat'd everything first).
+    const maxSize = 10 * 1024 * 1024; // 10MB
     const chunks: Buffer[] = [];
+    let totalSize = 0;
     for await (const chunk of response.body) {
+      totalSize += chunk.length;
+      if (totalSize > maxSize) {
+        throw new Error(
```
Tara-ag, line 1672: 🔒 **CRITICAL** — Stream not aborted on size-cap breach

When `totalSize > maxSize`, the function throws while the undici response stream is still flowing. The connection keeps downloading (wasting bandwidth and holding a socket) until the server finishes or the OS buffers fill. For a user-supplied URL this is both a DoS vector and a resource leak.

Abort the response before throwing:

```typescript
if (totalSize > maxSize) {
  response.body.destroy();
  throw ErrorFactory.imageTooLarge(
    `${(totalSize / 1024 / 1024).toFixed(2)}MB`,
    "10MB",
  );
}
```

Also: replace this and the other two raw `Error` throws in this file with `ErrorFactory.imageTooLarge` / `ErrorFactory.invalidImageFormat` so callers can branch on `error.code` and retry policies treat deterministic validation failures correctly (CLAUDE.md / bootstrapped standards).

src/lib/utils/messageBuilder.ts
```diff
@@ -1656,21 +1659,24 @@ async function downloadImageFromUrl(url: string): Promise<string> {
       );
```
Tara-ag, line 1659: ⚠️ **MAJOR** — `content-type` parameters break data-URI parsing

`response.headers["content-type"]` may be `image/png; charset=binary` (or include other parameters). The code later builds `data:${contentType};base64,...`, which `processImageToBase64` parses with `/^data:([^;]+);base64,/`. A parameterized content-type therefore fails the regex and falls through to the malformed-URI error, even for a valid image download.

Strip parameters and normalize case before using the MIME for validation, caching, and data-URI generation:

```typescript
const rawContentType = (response.headers["content-type"] as string) || "image/jpeg";
const contentType = rawContentType.split(";")[0].trim().toLowerCase();
```

Keep `rawContentType` only for error reporting.

src/lib/utils/messageBuilder.ts
```diff
@@ -1656,21 +1659,24 @@ async function downloadImageFromUrl(url: string): Promise<string> {
...
+    // check the total (the previous code concat'd everything first).
+    const maxSize = 10 * 1024 * 1024; // 10MB
     const chunks: Buffer[] = [];
+    let totalSize = 0;
     for await (const chunk of response.body) {
+      totalSize += chunk.length;
+      if (totalSize > maxSize) {
+        throw new Error(
+          `Image too large: exceeds ${maxSize} bytes while downloading from ${url}`,
+        );
```
Tara-ag, line 1674: ⚠️ **MAJOR** — Full user-supplied URL echoed in error message can leak secrets

The error message interpolates `${url}` verbatim. If the caller passes a presigned URL, SAS token URL, or any URL with sensitive query parameters, this error can expose those credentials in logs, telemetry, or to the end user.

Sanitize before including in the message — strip the query string or use only the origin/path:

```typescript
const safeUrl = url.split("?")[0];
throw ErrorFactory.imageTooLarge(
  `${(totalSize / 1024 / 1024).toFixed(2)}MB`,
  "10MB",
);
// or include safeUrl in context only
```

Same concern applies to the existing `Failed to download image from ${url}` message a few lines below; consider fixing both while you're here.

test/continuous-test-suite-bugfixes.ts
```diff
@@ -548,6 +551,55 @@ const tests: TestFunction[] = [
...
```
Tara-ag, line 574: 💡 **MINOR** — Regression test misses case-sensitivity cases

The test only exercises lowercase data URIs. Because the production code currently has scheme/MIME case-sensitivity gaps (see `messageBuilder.ts` comments), add cases for:

- Uppercase scheme: `DATA:image/png;base64,...`
- Uppercase MIME: `data:IMAGE/PNG;base64,...`
- MIME with whitespace: `data: image/png;base64,...`

This prevents the fixes from regressing on RFC-compliant but differently-cased inputs.

Tara-ag (changes requested):
## Review summary

**Files reviewed:** 2 (`src/lib/utils/messageBuilder.ts`, `test/continuous-test-suite-bugfixes.ts`)

- 🔒 **CRITICAL:** 1 — response stream not destroyed when the 10 MB incremental cap is exceeded, leaving the download running and wasting bandwidth/sockets.
- ⚠️ **MAJOR:** 3 — `content-type` parameters break data-URI parsing; data-URI scheme/MIME are case-sensitive; user-supplied URL echoed verbatim in error messages can leak query-string secrets.
- 💡 **MINOR:** 2 — full-string lowercase for an 8-byte prefix check; regression test missing uppercase scheme/MIME cases.

**Blocking rationale**
The CRITICAL stream-abort issue meets the blocking criterion for a critical security/resource vulnerability (unbounded downstream data transfer on a user-supplied URL). The three MAJOR issues together also cross the "3 or more MAJOR severity issues" threshold. Therefore this PR must be blocked until addressed.

**Next steps**
1. Destroy/abort `response.body` before throwing on size-cap breach.
2. Strip `content-type` parameters and normalize case before building the data URI.
3. Make the data-URI scheme check and regex case-insensitive; normalize and trim the declared MIME.
4. Replace the raw `Error` throws with `ErrorFactory.imageTooLarge` / `ErrorFactory.invalidImageFormat` typed errors.
5. Sanitize user-supplied URLs in error messages (strip query strings).
6. Add regression tests for uppercase scheme, uppercase MIME, and whitespace-padded MIME.

Once these are fixed and the existing review threads are resolved, this should be good to merge.

PR #1163, opened 2026-07-14
fix(proxy): harden account routing and process reliability

murdore:
## Summary

- Make launchd the sole restart supervisor for launchd-managed proxies; make the foreground guard cleanup-only and runtime auto-update opt-in.
- Persist account cooldowns across restarts, honor authoritative unified quota rejection, and never retry accounts with active cooldowns.
- Serialize cooldown/quota persistence and rotating OAuth refreshes, including callers that observe the newly rotated refresh token.
- Distinguish terminal credential rejection from transient endpoint failures and preserve manual disable metadata.
- Add `routing.account-allowlist` as a fail-closed boundary across TokenStore, legacy-file, and environment credentials.
- Prevent hidden legacy/environment credentials from activating while any Anthropic TokenStore entry exists.
- Record upstream stream errors and client cancellation as terminal outcomes instead of successful HTTP 200 completions.
- Drain and exit deterministically on shutdown, surfacing close failures without leaving rejected signal handlers.

## Account isolation

`primary-account` remains an ordering preference. The new allowlist is the eligibility boundary and is applied before token loading or refresh. Empty means deny all; absent preserves existing unrestricted behavior. Legacy and environment fallbacks require explicit `legacy-default` or `env` entries.

Automatic token rotation no longer clears an operator-disabled account. Only explicit login re-enables current disable metadata; legacy `refresh_failed` records from older releases receive one migration recheck.

## Verification

- `pnpm run typecheck`: pass.
- `pnpm exec vitest run test/proxyReliabilityHardening.test.ts`: 23/23 pass.
- `pnpm test:bugfixes`: 89/89 pass.
- Changed-file ESLint: 0 errors and no new warnings.
- `pnpm run build:cli`: pass.
- Full pre-commit pipeline will run again on the amended commit.

## Rollout safety

Development and tests did not start, stop, signal, reconfigure, or install over the live proxy. The existing process-level proxy suite was intentionally not run because it manages global proxy state and processes. Release rollout should use an isolated state directory first, then a controlled launchd replacement with health and restart-count verification.

### DIFF
--- a/src/cli/commands/proxy.ts
+++ b/src/cli/commands/proxy.ts
@@ -29,6 +29,7 @@ import {
   StateFileManager,
 } from "../utils/serverUtils.js";
 import type {
+  AccountAllowlist,
   FallbackInfo,
   LoadedProxyConfig,
   ProxyGuardArgs,
@@ -47,16 +48,24 @@ import type {
 import type { ModelRouter } from "../../lib/proxy/modelRouter.js";
 import { configureProxyKeepAliveDispatcher } from "../../lib/proxy/proxyDispatcher.js";
 import {
+  anthropicAccountKeysEqual,
+  createAccountAllowlist,
+  ENV_ANTHROPIC_ACCOUNT_KEY,
+  isAccountAllowed,
+  LEGACY_ANTHROPIC_ACCOUNT_KEY,
+  normalizeAnthropicAccountKey,
+  shouldLoadFallbackCredential,
+} from "../../lib/proxy/accountSelection.js";
+import {
   loadProxyEnvFile,
   resolveProxyEnvFile,
 } from "../../lib/proxy/proxyEnv.js";
 import { createRequire } from "node:module";
 import { fileURLToPath } from "node:url";
+import packageJson from "../../../package.json" with { type: "json" };
 
 const _require = createRequire(import.meta.url);
-const { version: PROXY_VERSION } = _require("../../../package.json") as {
-  version: string;
-};
+const PROXY_VERSION = packageJson.version;
 
 const PROXY_TELEMETRY_SCRIPT_PATH = fileURLToPath(
   new URL(
@@ -128,6 +137,9 @@ async function resolveStatusPrimaryAccount(
   proxyConfig: LoadedProxyConfig | null,
 ): Promise<ProxyStatusPrimaryAccount> {
   const configured = proxyConfig?.routing?.primaryAccount?.trim() || null;
+  const accountAllowlist = createAccountAllowlist(
+    proxyConfig?.routing?.accountAllowlist,
+  );
   let enabledAnthropicKeys: string[] = [];
   try {
     const { tokenStore } = await import("../../lib/auth/tokenStore.js");
@@ -135,7 +147,7 @@ async function resolveStatusPrimaryAccount(
     const filtered: string[] = [];
     for (const key of all) {
       const disabled = await tokenStore.isDisabled(key);
-      if (!disabled) {
+      if (!disabled && isAccountAllowed(key, accountAllowlist)) {
         filtered.push(key);
       }
     }
@@ -149,11 +161,14 @@ async function resolveStatusPrimaryAccount(
   }
 
   if (configured) {
-    const configuredKey = `anthropic:${configured}`;
-    if (enabledAnthropicKeys.includes(configuredKey)) {
+    const configuredKey = normalizeAnthropicAccountKey(configured);
+    const matchedKey = enabledAnthropicKeys.find((key) =>
+      anthropicAccountKeysEqual(key, configuredKey),
+    );
+    if (matchedKey) {
       return {
         configured,
-        key: configuredKey,
+        key: matchedKey,
         label: configured,
         source: "configured",
       };
@@ -194,49 +209,14 @@ async function isLaunchdManaging(): Promise<boolean> {
   }
 }
 
-/**
- * Attempt to restart the proxy via launchd kickstart.
- * Returns true if the proxy comes back healthy within timeoutMs.
- */
-async function tryLaunchdRestart(
-  host: string,
-  port: number,
-  timeoutMs: number = 15_000,
-): Promise<boolean> {
-  if (process.platform !== "darwin") {
-    return false;
-  }
+function isLaunchdManagedProcess(): boolean {
+  return process.platform === "darwin" && process.ppid === 1;
+}
 
-  try {
-    const { existsSync } = await import("fs");
-    if (!existsSync(PLIST_PATH)) {
-      return false;
-    }
-  } catch {
-    return false;
-  }
-
-  try {
-    const { execFileSync } = await import("node:child_process");
-    const uid = process.getuid?.() ?? 501;
-    execFileSync(
-      "launchctl",
-      ["kickstart", "-k", `gui/${uid}/${PLIST_LABEL}`],
-      { stdio: "ignore", timeout: 5_000 },
-    );
-  } catch {
-    return false;
-  }
-
-  const deadline = Date.now() + timeoutMs;
-  while (Date.now() < deadline) {
-    await sleep(1_000);
-    if (await isProxyHealthy(host, port, 2_000)) {
-      return true;
-    }
-  }
-
-  return false;
+function isProxyAutoUpdateEnabled(
+  value = process.env.NEUROLINK_PROXY_AUTO_UPDATE,
+): boolean {
+  return ["1", "on", "true"].includes((value ?? "").trim().toLowerCase());
 }
 
 /** Keys we manage in Claude Code's settings.env */
@@ -972,6 +952,7 @@ async function loadProxyStartConfiguration(
   modelRouter: ModelRouter | undefined;
   passthrough: boolean;
   primaryAccountKey: string | undefined;
+  accountAllowlist: AccountAllowlist | undefined;
 }> {
   const configPath =
     argv.config ?? join(homedir(), ".neurolink", "proxy-config.yaml");
@@ -984,20 +965,15 @@ async function loadProxyStartConfiguration(
       spinner.text = `Loaded proxy config from ${configPath}`;
     }
   } catch (configError) {
-    if (argv.config) {
-      if (spinner) {
-        spinner.fail(chalk.red(`Failed to load proxy config: ${configPath}`));
-      }
-      process.exit(1);
-    }
     const isNotFound =
       configError instanceof Error &&
       "code" in configError &&
       (configError as NodeJS.ErrnoException).code === "ENOENT";
-    if (!isNotFound) {
-      logger.warn(
-        `[proxy] Ignoring default config ${configPath}: ${configError instanceof Error ? configError.message : String(configError)}`,
-      );
+    if (argv.config || !isNotFound) {
+      if (spinner) {
+        spinner.fail(chalk.red(`Failed to load proxy config: ${configPath}`));
+      }
+      throw configError;
     }
   }
 
@@ -1019,6 +995,17 @@ async function loadProxyStartConfiguration(
   const primaryAccountKey = await resolveBootPrimaryAccountKey(
     proxyConfig?.routing?.primaryAccount,
   );
+  const accountAllowlist = await resolveBootAccountAllowlist(
+    proxyConfig?.routing?.accountAllowlist,
+  );
+  if (
+    primaryAccountKey &&
+    !isAccountAllowed(primaryAccountKey, accountAllowlist)
+  ) {
+    throw new Error(
+      `Configured routing.primaryAccount=${proxyConfig?.routing?.primaryAccount} is excluded by routing.accountAllowlist`,
+    );
+  }
 
   return {
     configPath,
@@ -1027,7 +1014,48 @@ async function loadProxyStartConfiguration(
     modelRouter,
     passthrough: argv.passthrough ?? false,
     primaryAccountKey,
+    accountAllowlist,
   };
+}
+
+async function resolveBootAccountAllowlist(
+  configuredAccounts: string[] | undefined,
+): Promise<AccountAllowlist | undefined> {
+  const allowlist = createAccountAllowlist(configuredAccounts);
+  if (allowlist === undefined) {
+    return undefined;
+  }
+  if (allowlist.size === 0) {
+    logger.warn(
+      "[proxy] routing.accountAllowlist is empty; all stored Anthropic credentials are denied",
+    );
+    return allowlist;
+  }
+
+  try {
+    const { tokenStore } = await import("../../lib/auth/tokenStore.js");
+    const known = new Set(
+      (await tokenStore.listByPrefix("anthropic:")).map(
+        normalizeAnthropicAccountKey,
+      ),
+    );
+    known.add(LEGACY_ANTHROPIC_ACCOUNT_KEY);
+    known.add(ENV_ANTHROPIC_ACCOUNT_KEY);
+    for (const key of allowlist) {
+      if (!known.has(key)) {
+        logger.warn(
+          `[proxy] WARN: routing.accountAllowlist entry ${key} is allowed but not currently authenticated`,
+        );
+      }
+    }
+  } catch (err) {
+    logger.debug(
+      `[proxy] could not validate account allowlist against token store: ${
+        err instanceof Error ? err.message : String(err)
+      }`,
+    );
+  }
+  return allowlist;
 }
 
 /** Resolve the operator's configured primary email to a stable token-store
@@ -1042,11 +1070,11 @@ async function resolveBootPrimaryAccountKey(
   if (!trimmed) {
     return undefined;
   }
-  const key = `anthropic:${trimmed}`;
+  const key = normalizeAnthropicAccountKey(trimmed);
   try {
     const { tokenStore } = await import("../../lib/auth/tokenStore.js");
     const known = await tokenStore.listByPrefix("anthropic:");
-    if (!known.includes(key)) {
+    if (!known.some((knownKey) => anthropicAccountKeysEqual(knownKey, key))) {
       logger.warn(
         `[proxy] WARN: configured routing.primaryAccount=${trimmed} not ` +
           `found in token store; falling back to first enabled account. ` +
@@ -1073,6 +1101,7 @@ async function createProxyStartApp(params: {
   host: string;
   proxyConfig: LoadedProxyConfig | null;
   primaryAccountKey: string | undefined;
+  accountAllowlist: AccountAllowlist | undefined;
 }) {
   const { createClaudeProxyRoutes } =
     await import("../../lib/server/routes/claudeProxyRoutes.js");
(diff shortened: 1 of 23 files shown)


src/lib/proxy/accountCooldown.ts
```diff
@@ -0,0 +1,116 @@
...
+  await loadAccountCooldowns();
+  const current = memoryCache[accountKey];
+  if (current && current.coolingUntil > coolingUntil) {
+    return;
+  }
+  memoryCache[accountKey] = {
+    coolingUntil,
+    reason,
+    updatedAt: Date.now(),
+  };
```
Tara-ag: ⚠️ **MAJOR**: `saveAccountCooldown` has a read-modify-write race for the same account. Two concurrent callers can both load the same stale `current`, both pass the extend-only check, and the last writer wins — potentially overwriting a longer weekly/unified cooldown with a shorter transient one. This undermines the "extend-only" and "never re-hammer cooling accounts" invariants.

**Suggested fix**: serialize the load-check-write sequence per account (or globally for this module) using the same `Mutex` pattern already used in `tokenStore.ts`:

```ts
import { Mutex } from "async-mutex";
const saveMutex = new Mutex();

export async function saveAccountCooldown(...): Promise<void> {
  await saveMutex.runExclusive(async () => {
    await loadAccountCooldowns();
    const current = memoryCache[accountKey];
    if (current && current.coolingUntil > coolingUntil) return;
    memoryCache[accountKey] = { coolingUntil, reason, updatedAt: Date.now() };
    await writeJsonSnapshotAtomically(getCooldownFilePath(), memoryCache);
  });
}
```

The same pattern applies to `clearAccountCooldown` and to `saveAccountQuota` in `accountQuota.ts`.

src/lib/proxy/accountCooldown.ts
```diff
@@ -0,0 +1,116 @@
...
+  accountKey: string,
+  expectedCoolingUntil?: number,
+): Promise<void> {
+  await loadAccountCooldowns();
+  const current = memoryCache[accountKey];
+  if (!current) {
+    return;
+  }
+  if (
+    expectedCoolingUntil !== undefined &&
```
Tara-ag: ⚠️ **MAJOR**: `clearAccountCooldown` shares the same read-modify-write race as `saveAccountCooldown`. A concurrent `saveAccountCooldown` and `clearAccountCooldown` (or two clears with different `expectedCoolingUntil` values) can both load the same stale snapshot; one write will clobber the other, violating the "extend-only" and "restart cannot erase cooldown" invariants.

Apply the same per-module `Mutex` serialization suggested for `saveAccountCooldown` so loads, checks, and the atomic write form one critical section.

src/lib/proxy/accountQuota.ts
Tara-ag, line 1: ⚠️ **MAJOR**: `saveAccountQuota` has the same read-modify-write race as `saveAccountCooldown`. The debounced `flushToDisk` snapshots `memoryCache` at call time, but between `loadAccountQuotas()` and the scheduled flush another caller can update `memoryCache[otherKey]`; the stale snapshot then overwrites those newer values on disk. Likewise, two concurrent first-callers can both load `{}` and the second flush wipes the first account's quota.

**Suggested fix**: protect the load-update-dirty path with a `Mutex` (matching `tokenStore.ts`), keeping the flush timer only for batching the disk write:

```ts
const saveMutex = new Mutex();
export async function saveAccountQuota(accountKey, quota): Promise<void> {
  await saveMutex.runExclusive(async () => {
    if (!cacheLoaded) await loadAccountQuotas();
    memoryCache[accountKey] = quota;
    dirty = true;
  });
  scheduleFlush();
}
```

Also consider serializing `flushToDisk` itself or re-checking `dirty`/snapshot consistency before clearing `dirty`.

src/lib/proxy/tokenRefresh.ts
```diff
@@ -89,12 +108,78 @@ export async function refreshToken(
...
+          refreshToken: refreshAccount.refreshToken,
+          expiresAt: refreshAccount.expiresAt,
+        };
+        if (result.success) {
+          const timer = setTimeout(() => {
+            if (refreshesInFlight.get(lockKey) === createdPromise) {
+              refreshesInFlight.delete(lockKey);
+            }
+          }, SUCCESS_CACHE_MS);
+          timer.unref?.();
```
Tara-ag, line 167: ⚠️ **MAJOR**: The in-flight refresh cache is keyed only by the old refresh token. After a successful refresh, the old token is invalidated by Anthropic, so a second caller with the *same* old token correctly shares the result. However, the first caller that *wins* updates `account.refreshToken` to the new rotating token and persists it. A slightly later caller that loaded the *new* token from disk just before the winner persisted will see no in-flight entry, perform its own refresh, and invalidate the winner's new token — reproducing the exact `invalid_grant` race this code intends to prevent.

**Suggested fix**: extend the cache to also cover a short window keyed by the new refresh token (or by account key) after a success, so callers observing either the old or the freshly-rotated token deduplicate. For example, on success also set an entry under the new token for the remaining `SUCCESS_CACHE_MS` window, and/or key an account-level in-flight promise by account label in addition to the token.

src/cli/commands/proxy.ts
```diff
@@ -1441,35 +1601,87 @@ async function refreshProxyTokensInBackground(): Promise<void> {
...
+
+  const closeServer = async (): Promise<void> => {
+    const close = params.server.close?.bind(params.server);
+    if (!close) {
+      return;
+    }
+    await new Promise<void>((resolve) => {
+      let settled = false;
+      const finish = () => {
+        if (settled) {
```
Tara-ag, line 1657: ⚠️ **MAJOR**: `registerProxyShutdownHandlers` registers `process.on("SIGTERM", ...)` and `process.on("SIGINT", ...)` handlers that are `async` functions whose rejections are unhandled. If `shutdown()` throws (e.g. `closeServer` rejects, OTel flush fails, or `clearClaudeProxySettings` throws), the exception becomes an unhandled promise rejection and the process may exit with an unintended code or hang.

**Suggested fix**: wrap the shutdown body in a try/catch and always exit deterministically:

```ts
const shutdown = async (signal: string) => {
  if (shutdownStarted) return;
  shutdownStarted = true;
  try {
    clearInterval(...);
    logger.always(`\\nShutting down proxy (${signal})...`);
    await closeServer();
    // ... OTel, settings cleanup, clearProxyState ...
    process.exit(signal === "SIGINT" ? 0 : 1);
  } catch (err) {
    logger.error(`[proxy] shutdown error: ${err instanceof Error ? err.message : String(err)}`);
    process.exit(1);
  }
};
process.on("SIGTERM", () => { void shutdown("SIGTERM"); });
```

Also note that `closeServer` currently swallows the `close()` error in its catch block; consider surfacing it so a failed drain does not silently proceed.

src/lib/server/routes/claudeProxyRoutes.ts
Tara-ag, line 1: 💡 **MINOR**: The post-load loop still logs `"account=X credentials changed, re-enabling"` and clears `permanentlyDisabled` runtime state when tokens differ from the previous in-process snapshot. This wording and behavior directly contradicts the PR's stated invariant that "Token changes no longer alter eligibility." While operator-disabled accounts are filtered out earlier via `tokenStore.isDisabled`, the runtime re-enable path is confusing and could mask future bypasses if that filter is ever skipped. Consider removing the misleading log or clarifying that it only resets transient runtime state, not persisted disable metadata.

Tara-ag, line 1: 💡 **MINOR**: `handleAnthropicRoutedClaudeRequest` sets `loopState.sawTransientFailure = !hasRateLimitCooldown` when all accounts are cooling. If every account is in an auth cooldown (`coolingReason === "auth"`), this marks the failure as transient and the final response will fall through to the generic "All Anthropic accounts failed" 502 path instead of the auth-specific 401/403 reauth message. Since auth cooldowns are bounded and indicate refresh-token problems, returning the configured `authFailureMessage` (when present) would give the operator a clearer signal than a generic 502.

src/lib/auth/tokenStore.ts
```diff
@@ -182,13 +182,23 @@ export class TokenStore {
       }
     }
 
-    // Update provider tokens
+    const existingProvider = storageData.providers[provider];
```
Tara-ag, line 185: 💡 **MINOR**: The `saveTokens` preservation of disable metadata is good, but it only preserves the flag when the *existing* record is already disabled. If an operator disables an account and then a background refresh (or login) saves fresh tokens before the disable is visible, the new record will not carry the disable flag. This is mitigated by `markDisabled` being called elsewhere, but the asymmetry means the only safe path is "disable then save" rather than "save then disable". Consider documenting this ordering requirement in the `markDisabled` JSDoc or, safer, allowing callers to pass an optional `disabled` state into `saveTokens` so the auth/login path can atomically save tokens + disable reason in one write.

Tara-ag (changes requested):
## Review Summary

Reviewed the proxy reliability hardening PR (25 changed files, ~2.6K additions). The change addresses real production incidents and the core invariants are well documented. However, I found **5 MAJOR** and **3 MINOR** new issues that need to be addressed before merge.

### Blocking criterion
Per the review charter, **3 or more MAJOR issues in the same PR is a BLOCK**. This PR crosses that threshold.

### MAJOR issues (blocking)

1. **`src/lib/proxy/accountCooldown.ts:saveAccountCooldown`** — read-modify-write race can violate the extend-only cooldown invariant. Two concurrent callers can load the same stale snapshot and the shorter cooldown wins. Serialize the load-check-write path with a `Mutex`.

2. **`src/lib/proxy/accountCooldown.ts:clearAccountCooldown`** — same race as above; a concurrent save and clear can clobber each other.

3. **`src/lib/proxy/accountQuota.ts:saveAccountQuota`** — debounced flush snapshots `memoryCache` at scheduling time, but concurrent first-callers can load `{}` and wipe each other's account quotas. The load-update-dirty path needs serialization.

4. **`src/lib/proxy/tokenRefresh.ts:refreshToken`** — the in-flight cache is keyed only by the *old* refresh token. A caller that loads the *new* rotating token from disk just before the winner persists it will not find an in-flight entry and will trigger a second refresh, invalidating the winner's new token. Cache by account label or also by the new token for `SUCCESS_CACHE_MS`.

5. **`src/cli/commands/proxy.ts:registerProxyShutdownHandlers`** — `process.on("SIGTERM", ...)` and `SIGINT` handlers are async and unprotected; a thrown rejection in `shutdown()` can hang the process or produce an unintended exit code. Wrap shutdown in try/catch and always exit deterministically.

### MINOR issues (non-blocking, please address)

6. **`src/lib/server/routes/claudeProxyRoutes.ts`** — the post-load "credentials changed, re-enabling" log/runtime-clear contradicts the stated invariant that token changes no longer alter eligibility. Clarify or remove the misleading path.

7. **`src/lib/server/routes/claudeProxyRoutes.ts`** — when all accounts are in auth cooldown, `loopState.sawTransientFailure` is set and the final response falls through to a generic 502 instead of the more accurate auth/reauth message.

8. **`src/lib/auth/tokenStore.ts:saveTokens`** — disable metadata preservation only works when the existing record is already disabled; a save that races ahead of a disable can lose the operator's intent. Consider allowing callers to pass disable state atomically.

### Next steps

Please address the five MAJOR items (especially the cooldown/quota races and the shutdown handler) and re-request review. The test coverage for the new behavior is good, but the race windows above are not exercised by the current hermetic tests.

src/lib/proxy/accountQuota.ts
```diff
@@ -135,13 +140,8 @@ async function flushToDisk(): Promise<void> {
...
     // Snapshot before async I/O so we only clear dirty if nothing changed
     const snapshot = JSON.stringify(memoryCache, null, 2);
-    await ensureDir();
     const filePath = getQuotaFilePath();
-    const tmpPath = `${filePath}.tmp`;
-    await fs.writeFile(tmpPath, snapshot, {
-      mode: 0o600,
-    });
-    await fs.rename(tmpPath, filePath);
+    await writeJsonSnapshotAtomically(filePath, JSON.parse(snapshot), 0o600);
```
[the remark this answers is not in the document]
murdore: Fixed in 6b4edd7b. The quota flush now captures a cloned object snapshot once, writes it directly, and uses a version counter under serialized state/flush mutexes; the stringify-parse-stringify cycle is gone. The regression test also verifies the persisted multi-account snapshot.

docs/analysis/claude-proxy-reliability-hardening.md
```diff
@@ -0,0 +1,290 @@
+# Claude Proxy Reliability Hardening
+
+## Scope
+
+This change hardens the globally deployed Claude-compatible proxy without
```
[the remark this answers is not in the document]
murdore: The requested analysis file has been removed in 6b4edd7b, the current PR description is clean, and the oldest leaked description revision was content-purged through GitHub edit history. The exposed local Shooter credential still needs independent rotation outside this repository.

src/lib/proxy/accountCooldown.ts
murdore: Fixed in 6b4edd7b. saveAccountCooldown now serializes the complete load-check-update-atomic-write sequence with the repository AsyncMutex. A regression test races an extension, stale conditional clear, and shorter save, then verifies the longest cooldown on disk.

src/lib/proxy/accountCooldown.ts
murdore: Fixed in 6b4edd7b. clearAccountCooldown uses the same module mutex as saveAccountCooldown, so its expectedCoolingUntil check and atomic write cannot interleave with another mutation. The on-disk ordering regression covers this path.

src/lib/proxy/accountQuota.ts
murdore: Fixed in 6b4edd7b. First-load updates are serialized under stateMutex; flushes are serialized separately, write immutable snapshots directly, and clear dirty only when cacheVersion still matches. The test flushes concurrent first-account updates and verifies both entries on disk.

src/lib/proxy/tokenRefresh.ts
murdore: Fixed in 6b4edd7b. A successful refresh is cached for SUCCESS_CACHE_MS under both the old refresh token and the newly rotated token. The regression test refreshes through the old token, then calls again with the rotated token and confirms only one upstream fetch.

src/cli/commands/proxy.ts
murdore: Fixed in 6b4edd7b. Server close callback errors, synchronous close errors, and drain timeout now reject and are logged; cleanup chooses a deterministic exit code. Both signal handlers attach a terminal catch that force-exits 1 on any unexpected rejection.

src/lib/server/routes/claudeProxyRoutes.ts
murdore: Addressed in 6b4edd7b. The log now says eligible credentials were reloaded and stale in-process auth-disable state is being cleared. The adjacent comment makes explicit that eligibility was already resolved before this runtime reset; token changes do not bypass persisted operator disable metadata.

murdore: Addressed in 6b4edd7b. An all-auth-cooldown pool now returns a specific 503 token_refresh_unavailable response with the earliest retry time instead of the generic all-accounts 502. Permanent credential rejection still uses the existing 401 re-authentication path; bounded transient refresh failures are not incorrectly labeled as requiring login.

src/lib/auth/tokenStore.ts
murdore: Addressed in 6b4edd7b. saveTokens and markDisabled already share the same TokenStore AsyncMutex, so either ordering preserves operator intent: a later disable wins and a later save preserves existing disable metadata. This guarantee is now documented on markDisabled and covered by a concurrent save-before-disable regression test.

Tara-ag (approved):
## Review Summary

Reviewed the full diff for PR #1163 (23 changed files, +2506/-520).

### Existing review comments
All 10 prior inline threads from `Tara-ag` and `copilot-pull-request-reviewer` have been addressed by the author in commit `6b4edd7b`:
- cooldown/quota read-modify-write races are now serialized under `AsyncMutex`;
- rotating OAuth refresh-token deduplication now caches successes under both the old and newly-rotated token;
- shutdown handlers drain the server deterministically and surface close errors;
- TokenStore disable metadata is preserved across automatic saves and protected by the shared mutex;
- all-auth-cooldown pools now return a specific `503 token_refresh_unavailable` response.

### New findings this run
- **0 CRITICAL**, **0 MAJOR**, **0 MINOR** new issues.
- No hardcoded secrets, injection vectors, SSRF, unsafe eval, or PII exposure in the changed source.
- No violations of CLAUDE.md Critical Rules enforced outside CI (dynamic-import registry rule, Gemini tools/schema rule, CLI/SDK separation, `formatProviderError` return rule, public SDK backward compatibility).
- New types are defined only in the canonical `src/lib/types/` location and imported via the barrel.
- The new `account-allowlist` boundary is applied consistently across TokenStore, legacy-file, and environment credential discovery, and hidden fallbacks are correctly suppressed while any Anthropic TokenStore entry exists.
- Cooldown/quota persistence, rotating refresh serialization, stream terminal-outcome tracking, and deterministic shutdown all appear correct.

### Tests
The PR adds `test/proxyReliabilityHardening.test.ts` covering unified rate-limit classification, cooldown/quota persistence races, refresh failure classification, rotating-token serialization, account restriction, stream terminal outcomes, and launchd lifecycle invariants. This aligns with the existing vitest-based test files in the repo.

### Verdict
Approve — no new blocking criteria are met.

src/cli/commands/proxy.ts
```diff
@@ -1565,11 +1805,11 @@ async function startProxyRuntime(params: {
...
     fallbackChain,
+    accountAllowlist: params.accountAllowlist
+      ? [...params.accountAllowlist]
+      : undefined,
     guardPid,
-    managedBy:
-      process.platform === "darwin" && process.ppid === 1
-        ? "launchd"
-        : "manual",
+    managedBy: managedByLaunchd ? "launchd" : "manual",
```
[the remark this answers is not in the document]
murdore: Validated against v9.93.1: still valid. Fixed in #1208 by writing and chmodding StateFileManager outputs to 0o600, updating the config reference, and testing an existing broad-mode file. Resolving with current regression evidence.

src/lib/proxy/sseInterceptor.ts
```diff
@@ -409,6 +412,22 @@ function processEvent(
...
+      payload.error && typeof payload.error === "object"
+        ? (payload.error as Record<string, unknown>)
+        : undefined;
+    const message = nestedError?.message ?? payload.message;
+    acc.streamErrorMessage =
+      typeof message === "string" && message.trim()
+        ? message.trim()
+        : truncateString(event.data, MAX_EVENT_DATA_BYTES);
+  }
+
```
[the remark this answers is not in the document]
murdore: Validated against v9.93.1: still valid. Fixed in #1208 by applying the existing 2,048-character bounded truncation to derived SSE error messages, with an oversized-message regression test. Resolving.

PR #1164, opened 2026-07-15
fix(utils): pin SSRF downloads to the full validated address set (IPv4 first)

murdore:
## Problem — every `safeDownload` still times out on IPv4-only networks

Follow-up to #1157. That fix made the connect layer's `options.all` contract work, but `validateAndResolveUrl` still pins **one** address chosen by OS resolver preference:

```ts
await assertSafeUrl(url);          // validates EVERY A/AAAA answer…
const result = await lookup(host); // …then discards them and picks ONE by OS preference
```

On a network with broken/absent IPv6 routing but an AAAA-preferring resolver (stock macOS with any IPv6 interface address), that pins a lone IPv6 address. The pinned agent has nothing to fall back to → 10 s connect timeout → the familiar bare `fetch failed`.

## Live evidence (2026-07-15, released 9.87.1)

Real generations succeeded on Replicate; **every download failed**:

```
[0] TypeError: fetch failed
[1] ConnectTimeoutError: Connect Timeout Error (attempted address: replicate.delivery:443, timeout: 10000ms)
```

- `validateAndResolveUrl('https://replicate.delivery/…')` → pinned `2606:4700::6812:1156` (family 6)
- same host, same network: `curl -4` → **200 in 0.39 s**; `curl -6` → timeout. curl survives because Happy Eyeballs races both families — a one-address pin cannot.

## Fix

- `validateAndResolveUrl` hostname path now resolves **once**, validates that exact answer (all-must-pass, unchanged), and returns every cleared address **IPv4 first** as `addresses`; `ip`/`family` mirror the first entry for single-pin callers.
- `buildPinnedAgent` hands the whole set to the connect layer, so `autoSelectFamily` can race families and fall back. Non-`all` callers get the first (now IPv4-preferred) entry.
- The connection can still only ever reach addresses the guard validated — and validating the same answer we pin **removes** the rebinding window the old second `lookup()` left open. Strictly no weaker.

## After (same network, same host)

```
pinned set: [104.18.16.86 v4, 104.18.17.86 v4, 2606:4700::6812:1156 v6, 2606:4700::6812:1056 v6]
OK 4573573 bytes in 1.36s
```

That 4.5 MB file is a kling-v2.1 clip that generated fine but was undownloadable before this patch.

## Tests

`pnpm run test:ssrf`: **42/42** — includes three new cases: literal → singleton `addresses`, dual-stack hostname → IPv4-first ordering, `ip`/`family` mirror `addresses[0]`. Typecheck clean.

## Impact

Unblocks every `safeDownload` consumer (Replicate / Runway / Kling asset fetches) on IPv4-only networks — and the downstream Director draft-tier rollout (juspay/director#98).

## Summary by CodeRabbit

* **Security Enhancements**
  * Strengthened protection against server-side request forgery by validating all resolved IPv4 and IPv6 addresses.
  * Pinned downloads to the complete validated address set, reducing DNS-rebinding risks.
  * Improved handling of dual-stack hosts and connection fallback behavior.

* **Bug Fixes**
  * DNS resolution failures are now rejected instead of being allowed to proceed.

* **Tests**
  * Added coverage for IP literals and dual-stack address ordering.

### DIFF
--- a/src/lib/types/safeFetch.ts
+++ b/src/lib/types/safeFetch.ts
@@ -4,6 +4,16 @@
  * Runtime helper lives in `src/lib/utils/safeFetch.ts`.
  */
 
+/**
+ * One validated address the pinned connect layer is allowed to dial.
+ * Produced by `ssrfGuard.ts:validateAndResolveUrl`, consumed by
+ * `safeFetch.ts:buildPinnedAgent`.
+ */
+export type PinnedAddress = {
+  ip: string;
+  family: 4 | 6;
+};
+
 export type SafeDownloadOptions = {
   /** Hard cap on response size in bytes. Pass MAX_VIDEO_BYTES/MAX_AUDIO_BYTES/MAX_IMAGE_BYTES from sizeGuard. */
   maxBytes: number;
--- a/src/lib/utils/safeFetch.ts
+++ b/src/lib/utils/safeFetch.ts
@@ -4,9 +4,9 @@
  * Combines:
  *   - `assertSafeUrl` (validates and rejects blocked IPs)
  *   - undici `Agent` with custom `connect.lookup` so the actual connection
- *     uses the IP we validated (closes the DNS-rebinding window where the
- *     resolver returns a public IP for the guard but a private IP for the
- *     real request).
+ *     only ever dials addresses we validated (closes the DNS-rebinding window
+ *     where the resolver returns a public IP for the guard but a private IP
+ *     for the real request).
  *   - `readBoundedBuffer` for size cap.
  *   - `redirect: "manual"` so a 3xx → private-IP redirect can't bypass
  *     the guard.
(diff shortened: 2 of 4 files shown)


Tara-ag (approved):
## Review Summary

**Files reviewed:** 4
**New inline comments:** 0
**New blocking issues:** 0

This PR correctly fixes the IPv4-only network timeout by pinning the full validated address set (IPv4-first) instead of a single OS-preferred address, and removes the DNS-rebinding TOCTOU window by resolving once and validating the same answer that is handed to the connect layer.

### What I checked

- **Security / SSRF:** `resolveAndValidateHost` still enforces all-must-pass validation against blocked CIDRs for both A and AAAA records. The connect-layer `lookup` callback only returns addresses that passed the guard, and still rejects mismatched hostnames defensively.
- **Correctness:** `buildPinnedAgent` now satisfies both the `all: true` array contract and the single-address fallback contract. The `validateAndResolveUrl` return type is additive (`addresses`) and remains backward compatible.
- **CLAUDE.md rules:** `PinnedAddress` is a `type` (Rule 7), lives in `src/lib/types/` (Rule 2), has a unique name (Rule 9), and is imported via the barrel (Rule 13).
- **Tests:** New regression coverage for IP-literal singletons and IPv4-first dual-stack ordering is included in `test:ssrf`.

### Non-blocking note

`docs/provider-integration/SAFETY-PRIMITIVES.md` still shows the old destructuring example:

```ts
const { url, ip, family } = await validateAndResolveUrl(url);
```

Consider updating it to include the new `addresses` field so the docs stay in sync with the return shape.

Approving — no blocking issues.

PR #1166, opened 2026-07-15
fix(proxy): harden updater fallback and terminal errors

murdore:
## Description

Hardens the Claude proxy updater, fallback execution, stream termination, and adapter-level error telemetry based on the recent live failure analysis.

## Root Causes Addressed

- launchd-managed proxies did not spawn the guard that contained the updater, so auto-update could never run
- package-manager detection could select a working pnpm from an unrelated global root instead of the npm installation actually running NeuroLink
- restart safety inferred quiet traffic from debug-log timestamps and could interrupt active streams
- configured and auto-provider fallback failures were discarded, fallback streams could hang indefinitely, and transient fallback transport failures had no bounded retry
- Anthropic SSE error/early-close events could be converted into an OpenAI-compatible success terminator
- uncaught adapter errors such as `fetch failed` bypassed structured request/error statistics

## Changes

- add a dedicated launchd updater worker; automatic updates are enabled by default and can be disabled with `NEUROLINK_PROXY_AUTO_UPDATE=off|false|0`
- resolve only a writable npm/pnpm global root that owns the running entry script; log all candidates plus captured install stdout/stderr
- track active requests through response-body completion and require zero active requests plus a two-minute idle window before install and again before restart
- replace `bootout`/`bootstrap` with atomic `launchctl kickstart -k`; never rewrite or unload the stable plist during an update
- add bounded fallback initialization/stream idle timeouts, one retry for transient network failures, upstream abort on timeout, and preserved terminal fallback diagnostics
- propagate Anthropic SSE errors and incomplete streams as OpenAI `server_error` events without fake `[DONE]` success
- record invalid JSON and uncaught adapter failures in request logs, body capture, and usage stats while returning a generic client-safe 502
- expose request activity and updater worker state through `/status` and `neurolink proxy status`

## Testing

- [x] `pnpm run build` including package/declaration build, browser build, CLI build, and publint
- [x] CLI TypeScript check
- [x] focused proxy reliability tests: 48/48
- [x] production bugfix regression suite: 89/89
- [x] OpenAI proxy format suite: 10/10
- [x] remaining unit stages after model-pool: 58/58
- [x] isolated packaged proxy smoke on port 55789: health/status, invalid JSON 400, structured 1 request/1 error/0 rate limits, active request count returned to zero
- [x] real machine read-only installer probe selected NVM npm 11.12.1 owning the installed package and rejected the unrelated pnpm root

`pnpm run test:unit` has one pre-existing failure: `test/modelPool.test.ts` times out at its fixed 5-second limit. It reproduces unchanged in the untouched original checkout; all 18 other tests in that file pass.

## Deployment Notes

The currently installed older proxy cannot gain the new updater worker by itself. Upgrade once to the release containing this PR and restart through the normal operator workflow; subsequent launchd-managed releases auto-update by default. Set `NEUROLINK_PROXY_AUTO_UPDATE=off` before that restart to opt out. Updater diagnostics are written to `~/.neurolink/logs/proxy-updater.log`.

The live proxy was not restarted, upgraded, reconfigured, or used for test traffic while developing this change. No dependencies were added.

## Summary by CodeRabbit

* **New Features**
  * Proxy auto-updates are enabled by default, with clearer behavior for launchd-managed vs foreground runs.
  * Added runtime activity introspection (active requests, idle detection) and “Auto-update” status visibility, plus updater diagnostics output.
* **Bug Fixes**
  * Hardened streaming and cancellation handling with more accurate SSE/OpenAI-compatible error frames.
  * Improved fallback translation reliability with retries, timeouts, and clearer “fallback also failed” messaging.
* **Documentation**
  * Updated proxy guard, environment variables (including package-manager selection), and cleanup/updater guidance.
* **Tests**
  * Added coverage for activity tracking, installer selection, updater fallback, and stream/interruption behavior.

### DIFF
--- a/src/cli/commands/proxy.ts
+++ b/src/cli/commands/proxy.ts
@@ -43,6 +43,8 @@ import type {
   ProxyStatusPrimaryAccount,
   ProxyTelemetryAction,
   ProxyTelemetryArgs,
+  ProxyRuntimeActivity,
+  RuntimeRequestMetadata,
   StatusStats,
 } from "../../lib/types/index.js";
 import type { ModelRouter } from "../../lib/proxy/modelRouter.js";
@@ -56,6 +58,16 @@ import {
   normalizeAnthropicAccountKey,
   shouldLoadFallbackCredential,
 } from "../../lib/proxy/accountSelection.js";
+import {
+  beginProxyRequest,
+  getProxyActivitySnapshot,
+  trackProxyResponse,
+} from "../../lib/proxy/proxyActivity.js";
+import {
+  describeInstallFailure,
+  getGlobalInstallArgs,
+  resolveGlobalInstaller,
+} from "../../lib/proxy/globalInstaller.js";
 import {
   loadProxyEnvFile,
   resolveProxyEnvFile,
@@ -216,7 +228,7 @@ function isLaunchdManagedProcess(): boolean {
 function isProxyAutoUpdateEnabled(
   value = process.env.NEUROLINK_PROXY_AUTO_UPDATE,
 ): boolean {
-  return ["1", "on", "true"].includes((value ?? "").trim().toLowerCase());
+  return !["0", "off", "false"].includes((value ?? "").trim().toLowerCase());
 }
 
 /** Keys we manage in Claude Code's settings.env */
@@ -457,6 +469,55 @@ async function isProxyHealthy(
   }
 }
 
+async function getProxyRuntimeActivity(
+  host: string,
+  port: number,
+  timeoutMs: number = 3_000,
+): Promise<ProxyRuntimeActivity | null> {
+  try {
+    const response = await fetch(`http://${host}:${port}/status`, {
+      signal: AbortSignal.timeout(timeoutMs),
+    });
+    if (!response.ok) {
+      return null;
+    }
+    const payload = (await response.json()) as {
+      activity?: Partial<ProxyRuntimeActivity>;
+    };
+    const activeRequests = Number(payload.activity?.activeRequests);
+    if (!Number.isFinite(activeRequests) || activeRequests < 0) {
+      return null;
+    }
+    return {
+      activeRequests,
+      lastActivityAt:
+        typeof payload.activity?.lastActivityAt === "string"
+          ? payload.activity.lastActivityAt
+          : null,
+    };
+  } catch {
+    return null;
+  }
+}
+
+function isSafeUpdateWindow(
+  activity: ProxyRuntimeActivity,
+  quietThresholdMs: number,
+  nowMs: number = Date.now(),
+): boolean {
+  if (activity.activeRequests > 0) {
+    return false;
+  }
+  if (!activity.lastActivityAt) {
+    return true;
+  }
+  const lastActivityMs = Date.parse(activity.lastActivityAt);
+  return (
+    Number.isFinite(lastActivityMs) &&
+    nowMs - lastActivityMs >= quietThresholdMs
+  );
+}
+
 // ---------------------------------------------------------------------------
 // Stable entrypoint for launchd
 // ---------------------------------------------------------------------------
@@ -566,136 +627,6 @@ exit 127
   chmodSync(TRAMPOLINE_PATH, 0o755);
 }
 
-/**
- * Check whether a pnpm binary can install into the global store.
- *
- * Multiple pnpm major versions can coexist (e.g., standalone v8 + nvm v10).
- * They use different store layouts (`store/v10` vs `store/v10/v3`), so a
- * pnpm that passes `--version` may still fail `pnpm add -g` with
- * ERR_PNPM_UNEXPECTED_STORE. We detect this by running `pnpm root -g` and
- * checking whether the resolved global root directory actually exists on disk.
- */
-function canInstallGlobally(pnpmPath: string): boolean {
-  try {
-    const { execFileSync } = _require(
-      "node:child_process",
-    ) as typeof import("node:child_process");
-    const { existsSync } = _require("fs") as typeof import("fs");
-    const globalRoot = execFileSync(pnpmPath, ["root", "-g"], {
-      encoding: "utf8",
-      timeout: 10_000,
-      stdio: ["ignore", "pipe", "ignore"],
-    }).trim();
-    // If the global root exists, this pnpm version is compatible with
-    // the current store layout and can install packages there.
-    return !!globalRoot && existsSync(globalRoot);
-  } catch {
-    return false;
-  }
-}
-
-/**
- * Resolve the `pnpm` binary defensively.
- *
- * Tries multiple candidates in order of preference. Each candidate must:
- * 1. Respond to `pnpm --version` (binary works)
- * 2. Have a compatible global store (`pnpm root -g` points to an existing dir)
- *
- * This defends against environments with multiple pnpm major versions
- * (e.g., standalone v8 + nvm v10) where the wrong one would fail with
- * ERR_PNPM_UNEXPECTED_STORE on `pnpm add -g`.
- *
- * Honors `NEUROLINK_PNPM_PATH` as an escape hatch.
- */
-function resolveFullPnpmPath(): {
-  bin: string;
-  resolved: boolean;
-  version?: string;
-  tried: Array<{
-    path: string;
-    version?: string;
-    working: boolean;
-    globalStoreOk?: boolean;
-  }>;
-} {
-  const candidates: string[] = [];
-
-  // 1. User override
-  if (process.env.NEUROLINK_PNPM_PATH) {
-    candidates.push(process.env.NEUROLINK_PNPM_PATH);
-  }
-
-  // 2. PNPM_HOME (pnpm's own env variable)
-  if (process.env.PNPM_HOME) {
-    candidates.push(join(process.env.PNPM_HOME, "pnpm"));
-  }
-
-  // 3. `which pnpm` — whatever is on PATH
-  try {
-    const { execFileSync } = _require(
-      "node:child_process",
-    ) as typeof import("node:child_process");
-    const whichOut = execFileSync("which", ["pnpm"], {
-      encoding: "utf8",
-      timeout: 5_000,
-      stdio: ["ignore", "pipe", "ignore"],
-    }).trim();
-    if (whichOut) {
-      candidates.push(whichOut);
-    }
-  } catch {
-    // ignore
-  }
-
-  // 4. Common standalone installer locations
-  candidates.push(join(homedir(), ".local", "share", "pnpm", "pnpm"));
-  candidates.push(join(homedir(), "Library", "pnpm", "pnpm"));
-
-  // Dedupe while preserving order
-  const seen = new Set<string>();
-  const unique = candidates.filter((p) => {
-    if (!p || seen.has(p)) {
-      return false;
-    }
-    seen.add(p);
-    return true;
-  });
-
-  // Probe each candidate: must pass --version AND have a compatible global store
-  const tried = unique.map((path) => {
-    const version = probeBinVersion(path);
-    const working = version !== undefined;
-    const globalStoreOk = working ? canInstallGlobally(path) : false;
-    return { path, version, working, globalStoreOk };
-  });
-
-  // Prefer a candidate that can actually install globally
-  const fullyWorking = tried.find((r) => r.working && r.globalStoreOk);
-  if (fullyWorking) {
-    return {
-      bin: fullyWorking.path,
-      resolved: true,
-      version: fullyWorking.version,
-      tried,
-    };
-  }
-
-  // Fall back to any candidate that at least responds to --version
-  // (better than nothing — the install may still fail, but will be
-  // caught and suppressed by the caller)
-  const anyWorking = tried.find((r) => r.working);
-  if (anyWorking) {
-    return {
-      bin: anyWorking.path,
-      resolved: true,
-      version: anyWorking.version,
-      tried,
-    };
-  }
-
-  return { bin: "pnpm", resolved: false, tried };
-}
-
 function spawnFailOpenGuard(
   host: string,
   port: number,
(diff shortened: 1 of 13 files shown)


Tara-ag (approved):
## Review summary

Reviewed 13 changed files. I did not raise any new blocking issues this run — the existing review threads (Copilot + CodeRabbit) already cover the substantive concerns I would have flagged:

- `writableDirectory()` should also require traverse (`X_OK`) permission (`src/lib/proxy/globalInstaller.ts:33`) — already raised.
- Anthropic SSE `error` events should terminate the transform rather than leave the stream open (`src/lib/proxy/openaiFormat.ts:932`) — already raised.
- OpenAI bridge stream cancel path should not misclassify a terminal stream error as `499 client_cancelled` (`src/lib/server/routes/openaiProxyRoutes.ts:290`) — already raised.
- Updater log setup should fail open so an unwritable log dir does not abort launchd-managed startup (`src/cli/commands/proxy.ts:714-721`) — already raised.
- Request-derived `model` should be sanitized before unstructured logging (`src/cli/commands/proxy.ts:1273-1275`) — already raised.

### Blocking-criteria check

- No critical security vulnerabilities (injection, SSRF, unsafe eval, secret/PII exposure) introduced.
- No breaking change to the public SDK API; changes are confined to the CLI/proxy path.
- No violation of a non-negotiable CLAUDE.md Critical Rule that ESLint does not enforce.
- Fewer than 3 new MAJOR issues raised by this review.

### Notes

- The new request-activity tracking (`proxyActivity.ts`) correctly holds requests open until the response body settles, and the updater gates install/restart on true runtime idleness.
- The global-installer resolver correctly rejects package managers whose writable global root does not own the running entry script, addressing the reported pnpm/npm shadowing problem.
- Fallback translation now has bounded initialization/stream idle timeouts, upstream abort on timeout, and one transient-network retry.
- Anthropic SSE errors and incomplete streams are propagated as OpenAI `server_error` events without a fake `[DONE]` success terminator.
- Uncaught adapter errors and invalid JSON are recorded in request logs, body capture, and usage stats while returning a generic client-safe response.

Approving — please address the existing reviewer comments before merging.

src/lib/proxy/openaiFormat.ts
```diff
@@ -916,8 +920,20 @@ export function createClaudeToOpenAIStreamTransform(
...
+      case "error": {
+        const error = (data.error ?? {}) as { message?: unknown };
+        const message =
+          typeof error.message === "string"
+            ? error.message
+            : "Anthropic stream failed";
+        finished = true;
+        options.onError?.(message);
+        emit(controller, serializer.emitError(message));
+        return;
```
Tara-ag, line 932: ⚠️ **Major**: On `event: error`, the transform emits an OpenAI error frame but leaves the underlying SSE connection open. If Anthropic sends an `error` event and keeps the connection alive (or sends keep-alive pings), the consumer and outer lifecycle accounting can hang waiting for EOF.

After emitting the terminal error, consider terminating the transform so downstream receives EOF promptly:

```ts
case "error": {
  const error = (data.error ?? {}) as { message?: unknown };
  const message =
    typeof error.message === "string"
      ? error.message
      : "Anthropic stream failed";
  finished = true;
  options.onError?.(message);
  emit(controller, serializer.emitError(message));
  controller.terminate();
  return;
}
```

Same for the `flush()` path: it already sets `finished` and emits the error, but does not terminate. Adding `controller.terminate()` after the terminal error emission prevents the stream from lingering.

src/lib/server/routes/openaiProxyRoutes.ts
```diff
@@ -232,14 +232,66 @@ async function handleOpenAIToAnthropicBridge(args: {
...
+      },
+      async cancel(reason) {
+        try {
+          await reader.cancel(reason);
+        } finally {
+          await finishLifecycle(
+            499,
+            "client_cancelled",
+            "Client cancelled the OpenAI-compatible stream",
+          );
```
Tara-ag, line 290: ⚠️ **Major**: In the stream cancel path, lifecycle is always finalized as `499 client_cancelled`. If an upstream terminal stream error has already been observed (`terminalStreamError` is set), a client cancellation (often triggered by receiving the error event) will misclassify the request as a client cancel rather than a loopback stream error.

Track whether a terminal error was already emitted and choose the status accordingly:

```ts
async cancel(reason) {
  try {
    await reader.cancel(reason);
  } finally {
    if (terminalStreamError) {
      await finishLifecycle(
        502,
        "loopback_stream_error",
        terminalStreamError,
      );
    } else {
      await finishLifecycle(
        499,
        "client_cancelled",
        "Client cancelled the OpenAI-compatible stream",
      );
    }
  }
}
```

This aligns lifecycle accounting with the actual terminal event.

src/lib/proxy/globalInstaller.ts
```diff
@@ -0,0 +1,203 @@
...
+  ).trim();
+}
+
+function writableDirectory(path: string): boolean {
+  try {
+    if (!existsSync(path)) {
+      return false;
+    }
+    accessSync(path, constants.W_OK);
+    return true;
```
Tara-ag, line 33: 💡 **Minor**: `writableDirectory()` only checks `W_OK`. For directories, POSIX generally requires both write and execute (traverse) permission to actually create/modify entries. A directory that is writable but not executable can incorrectly be treated as usable, causing the updater to select the wrong global root/bin dir and then fail later.

Consider checking both `X_OK` and `W_OK`:

```ts
accessSync(path, constants.W_OK | constants.X_OK);
```

This is consistent with the goal of verifying that the package manager can actually install into and write executables in the directory.

src/cli/commands/proxy.ts
```diff
@@ -751,6 +682,62 @@ function spawnFailOpenGuard(
...
+  if (!entryScript) {
+    logger.always("[proxy] updater disabled: CLI entry script is unavailable");
+    return undefined;
+  }
+
+  const args = [
+    entryScript,
+    "proxy",
+    "guard",
+    "--host",
```
Tara-ag, line 705: 💡 **Minor**: The updater log file is created with the process umask (typically world-readable). Because the log can contain install paths, versions, and diagnostic output from the package manager, consider creating the log directory with restrictive permissions and opening the log file with mode `0o600`:

```ts
if (!existsSync(logsDir)) {
  mkdirSync(logsDir, { recursive: true, mode: 0o700 });
}
const logFd = openSync(join(logsDir, "proxy-updater.log"), "a", 0o600);
```

This aligns with the project's guidance to write sensitive persisted files restrictively.

src/cli/commands/proxy.ts
```diff
@@ -1177,12 +1264,18 @@ async function createProxyStartApp(params: {
...
         : 0;
+      const metadata = requestMetadata.get(c.req.raw);
+      if (metadata) {
+        metadata.model = String(model);
+        metadata.stream = stream === "stream";
+        metadata.toolCount = toolCount;
+      }
       logger.always(
         `[proxy] ${c.req.method} ${c.req.path} → model=${model} ${stream} tools=${toolCount}`,
       );
```
Tara-ag, line 1275: 💡 **Minor**: The request-derived `model` value is interpolated directly into an unstructured `logger.always` line. A model string containing CR/LF characters can forge log entries. Sanitize it before logging:

```ts
const logModel = String(model).replace(/[\r\n\u2028\u2029]/g, " ");
logger.always(
  `[proxy] ${c.req.method} ${c.req.path} → model=${logModel} ${stream} tools=${toolCount}`,
);
```

This keeps the existing log format while neutralizing log-injection payloads.

src/lib/proxy/proxyActivity.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export function isProxyActivityQuiet(
+  snapshot: ProxyActivitySnapshot,
+  quietThresholdMs: number,
+  nowMs: number = Date.now(),
+): boolean {
+  if (snapshot.activeRequests > 0) {
+    return false;
+  }
+  if (snapshot.lastActivityAt === null) {
```
Tara-ag, line 42: 💡 **Minor**: `trackProxyResponse` swallows the original response body's cancellation reason when forwarding it to the underlying reader. If the upstream reader throws on cancel (some implementations do), the `finally` block still calls `finishRequest()`, which is correct, but the cancellation reason is lost and the caller's `cancel(reason)` may not propagate the actual reason upstream.

Consider capturing the reason and passing it through:

```ts
async cancel(reason) {
  try {
    await reader.cancel(reason);
  } finally {
    finishRequest();
  }
}
```

This preserves the cancellation signal for the upstream source.

Tara-ag (approved):
## Review Summary

Reviewed 13 changed files. Raised **6 new inline comments** (2 Major, 4 Minor). No blocking criteria are met this run.

### New issues

**Major (2)**
- `src/lib/proxy/openaiFormat.ts:932` — Terminal Anthropic `error` events emit an OpenAI error frame but do not terminate the transform, so the stream can linger if the upstream connection stays alive.
- `src/lib/server/routes/openaiProxyRoutes.ts:290` — The stream cancel path always records `499 client_cancelled`, even when a terminal upstream stream error was already observed, misclassifying the lifecycle outcome.

**Minor (4)**
- `src/lib/proxy/globalInstaller.ts:33` — `writableDirectory()` checks only `W_OK`; directories also need traverse (`X_OK`) permission for reliable installability detection.
- `src/cli/commands/proxy.ts:705` — Updater log file is created with default umask; consider `0o700` directory and `0o600` log file permissions.
- `src/cli/commands/proxy.ts:1275` — Request-derived `model` is logged unsanitized; CR/LF characters can forge log entries.
- `src/lib/proxy/proxyActivity.ts:42` — `trackProxyResponse` drops the cancellation reason when cancelling the upstream reader.

### Existing comments
Acknowledged the 5 prior review comments from copilot-pull-request-reviewer and coderabbitai; I did not duplicate them.

### Decision
Approve. The issues raised are correctness/quality improvements, not security blockers or public-API-breaking changes. The two Major items should be addressed in a follow-up to avoid stream-lifecycle misclassification.

src/lib/proxy/globalInstaller.ts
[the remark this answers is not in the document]
murdore: Validated against v9.93.1: still valid. Fixed in #1208 by requiring W_OK | X_OK for installer directories and testing a writable but non-traversable global root. Resolving.

src/lib/proxy/openaiFormat.ts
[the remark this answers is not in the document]
murdore: Validated against v9.93.1: still valid. Fixed in #1208 by terminating the transform after emitting the OpenAI error frame. A never-closing upstream regression test verifies downstream completion and upstream cancellation. Resolving.

src/lib/server/routes/openaiProxyRoutes.ts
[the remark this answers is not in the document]
murdore: Validated against v9.93.1: still valid. Fixed in #1208 by preserving loopback_stream_error/502 when downstream cancellation races with an observed terminal upstream error, with direct lifecycle-classification coverage. Resolving.

src/cli/commands/proxy.ts
```diff
@@ -751,6 +682,62 @@ function spawnFailOpenGuard(
...
+    "--quiet",
+  ];
+  const { openSync, closeSync, mkdirSync, existsSync } = _require(
+    "fs",
+  ) as typeof import("fs");
+  const logsDir = join(homedir(), ".neurolink", "logs");
+  if (!existsSync(logsDir)) {
+    mkdirSync(logsDir, { recursive: true });
+  }
+  const logFd = openSync(join(logsDir, "proxy-updater.log"), "a");
```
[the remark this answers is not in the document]
murdore: Validated against v9.93.1: still valid. Fixed in #1208 through a shared fail-open worker-log opener. Unavailable log paths now fall back to ignored stdio instead of aborting proxy startup, with regression coverage. Resolving.

src/cli/commands/proxy.ts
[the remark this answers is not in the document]
murdore: Validated against v9.93.1: still valid. Fixed in #1208 by passing the request-derived model through sanitizeForLog before unstructured logging. Resolving.

src/lib/proxy/openaiFormat.ts
murdore: Validated against v9.93.1: still valid. Fixed in #1208 by terminating the transform after emitting the OpenAI error frame. A never-closing upstream regression test verifies downstream completion and upstream cancellation. Resolving.

src/lib/server/routes/openaiProxyRoutes.ts
murdore: Validated against v9.93.1: still valid. Fixed in #1208 by preserving loopback_stream_error/502 when downstream cancellation races with an observed terminal upstream error, with direct lifecycle-classification coverage. Resolving.

src/lib/proxy/globalInstaller.ts
murdore: Validated against v9.93.1: still valid. Fixed in #1208 by requiring W_OK | X_OK for installer directories and testing a writable but non-traversable global root. Resolving.

src/cli/commands/proxy.ts
murdore: Validated against v9.93.1: still valid. Fixed in #1208: worker log directories are forced to 0o700, files to 0o600 including existing files, and logging failure remains fail-open. Regression coverage verifies both paths. Resolving.

src/cli/commands/proxy.ts
murdore: Validated against v9.93.1: still valid. Fixed in #1208 by passing the request-derived model through sanitizeForLog before unstructured logging. Resolving.

src/lib/proxy/proxyActivity.ts
murdore: Validated against current release: this was fixed later in #1206; trackProxyResponse now forwards reader.cancel(reason). #1208 adds a regression test proving the exact reason reaches the upstream source. Resolving.

PR #1167, opened 2026-07-15
fix(proxy): prioritize expiring session windows in account routing

pdogra1299:
## Description

Refines the quota-aware account ordering introduced in #1143/#1144. The previous comparator ranked accounts by soonest **weekly (7d)** reset, which ignores the window that actually expires first: the rolling **5h session**. With two accounts whose sessions reset at different times, the session window closing soonest is the capacity that is about to be lost — it should be consumed first, regardless of the weekly schedule. The old ordering also considered session state only at hard exhaustion ("rejected"), so window handoffs happened reactively via 429s, and stale snapshots could misrepresent an account whose window had already reset.

## Type of Change

- [x] Bug fix / behavior refinement (non-breaking)

## Changes Made

- **Session-first ordering:** usable accounts now rank by soonest 5h session reset. Session resets are compared in tolerance buckets (default 15 min, `NEUROLINK_PROXY_SESSION_RESET_TOLERANCE_MS`) so near-simultaneous resets count as equal and the soonest **weekly** reset breaks the tie; bucketing keeps the comparator transitive. Accounts with no ticking session window (fresh, not yet started) have nothing expiring and sort after ticking windows.
- **Session soft limit (default 0.97, `NEUROLINK_PROXY_SESSION_SOFT_LIMIT`):** an account whose session utilization reaches the limit — or whose status is `throttled` — is demoted below accounts with headroom (never made unusable; still serves as fallback). The utilization snapshot is at least one response stale and requests arrive in concurrent bursts, so the small buffer makes the exhaustion handoff proactive and error-free instead of a burst of 429s at every window boundary. Saturated accounts order by soonest session reset (first back in service). Setting the limit to `1` disables demotion.
- **Reset freshening at read time:** a window whose reset epoch has passed is treated as fresh (0 used, `allowed`, reset unknown) in every routing decision, so stale pre-reset numbers can never saturate or reject an account whose window Anthropic has since renewed. The snapshot then self-corrects from live response headers on the account's next request.
- Probe-first for never-observed accounts and the configured-primary tie-break are unchanged.

## Testing

- [x] New no-API suite test `testSessionFirstOrdering` (8 cases): session-first vs weekly, same-bucket weekly tie-break, 0.97 boundary (0.96 stays, 0.97/0.98 demote), throttled demotion, passed-reset freshening for both windows, both-saturated recovery order, env-configurable limit.
- [x] Runtime-verified 13 scenarios including all regressions from the prior ordering tests (weekly-first fallback when no session is ticking, probe-first, primary tie-break) — all pass.
- [x] `pnpm exec vitest run test/proxyReliabilityHardening.test.ts` — 36/36 pass (adjacent cooldown/quota persistence unaffected).
- [x] `tsc --strict`, ESLint gates, prettier, `validate:all`, and full build pass via the pre-commit hook.

## Breaking Changes

- [x] No breaking changes. Existing env toggle `NEUROLINK_PROXY_QUOTA_ROUTING` still gates the whole feature; round-robin strategy is unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved quota-based account ordering by using session utilization limits and reset timing to prioritize more usable accounts.
  * Added more accurate handling for near/expired reset epochs so completed resets don’t incorrectly demote accounts.
  * Updated ordering rules to better account for soft-limit saturation, throttled status, and bucketed reset windows.
* **Tests**
  * Added a continuous in-process test covering session-first ordering, throttling, soft-limit behavior, reset precedence, and tie-breaking.

### DIFF
--- a/src/lib/server/routes/claudeProxyRoutes.ts
+++ b/src/lib/server/routes/claudeProxyRoutes.ts
@@ -581,40 +581,104 @@ function isQuotaRoutingEnabled(): boolean {
   return v !== "off" && v !== "false" && v !== "0";
 }
 
-/** Derive the ordering signals for an account from its latest runtime quota. */
-function accountSortMetrics(accountKey: string, now: number) {
+/** Session-utilization soft limit above which an account is demoted below
+ *  accounts with headroom (never made unusable). The utilization snapshot is
+ *  always at least one response stale and requests land in concurrent bursts,
+ *  so a small buffer under 100% turns the window-exhaustion handoff proactive
+ *  (scheduled between requests) instead of reactive (a burst of 429s).
+ *  Override with NEUROLINK_PROXY_SESSION_SOFT_LIMIT (0 < x <= 1; 1 disables). */
+function getSessionSoftLimit(): number {
+  // Number() rejects partially numeric values ("0.5oops") that parseFloat
+  // would silently truncate.
+  const raw = Number(process.env.NEUROLINK_PROXY_SESSION_SOFT_LIMIT ?? "");
+  return Number.isFinite(raw) && raw > 0 && raw <= 1 ? raw : 0.97;
+}
+
+/** Width of the bucket within which two session resets count as "the same
+ *  time", letting the weekly reset decide. Exact epoch equality never occurs,
+ *  and bucketing (unlike pairwise closeness) keeps the comparator transitive.
+ *  Override with NEUROLINK_PROXY_SESSION_RESET_TOLERANCE_MS. */
+function getSessionResetToleranceMs(): number {
+  // Number() + isInteger rejects partially numeric values ("900000oops")
+  // that parseInt would silently truncate.
+  const raw = Number(
+    process.env.NEUROLINK_PROXY_SESSION_RESET_TOLERANCE_MS ?? "",
+  );
+  return Number.isInteger(raw) && raw > 0 ? raw : 15 * 60 * 1000;
+}
+
+/**
+ * Derive the ordering signals for an account from its latest runtime quota.
+ *
+ * Reset freshening: a window whose reset epoch has already passed is treated
+ * as FRESH (0 used, "allowed", reset unknown) at read time — stale snapshot
+ * numbers from before the reset can never saturate or reject an account whose
+ * window Anthropic has since renewed. The snapshot itself is refreshed from
+ * live response headers the next time the account serves a request.
+ */
+function accountSortMetrics(
+  accountKey: string,
+  now: number,
+  sessionSoftLimit: number,
+  sessionResetToleranceMs: number,
+) {
   const st = accountRuntimeState.get(accountKey);
   const q = st?.quota;
   const coolingActive = !!st?.coolingUntil && now < st.coolingUntil;
+  // resetEpochToMs returns undefined for absent OR passed resets, so a
+  // ticking window is exactly "reset !== undefined".
   const weeklyReset = resetEpochToMs(q?.weeklyResetAt, now);
   const sessionReset = resetEpochToMs(q?.sessionResetAt, now);
-  const weeklyRejected =
-    q?.weeklyStatus === "rejected" && weeklyReset !== undefined;
-  const sessionRejected =
-    q?.sessionStatus === "rejected" && sessionReset !== undefined;
+  const sessionTicking = sessionReset !== undefined;
+  const weeklyTicking = weeklyReset !== undefined;
+  const sessionUsed = sessionTicking ? (q?.sessionUsed ?? 0) : 0;
+  const weeklyUsed = weeklyTicking ? (q?.weeklyUsed ?? -1) : q ? 0 : -1;
+  const sessionStatus = sessionTicking
+    ? (q?.sessionStatus ?? "unknown")
+    : "allowed";
+  const weeklyStatus = weeklyTicking
+    ? (q?.weeklyStatus ?? "unknown")
+    : "allowed";
+  const saturated =
+    sessionStatus === "throttled" ||
+    (sessionTicking && sessionUsed >= sessionSoftLimit);
   return {
-    usable: !coolingActive && !weeklyRejected && !sessionRejected,
+    usable:
+      !coolingActive &&
+      weeklyStatus !== "rejected" &&
+      sessionStatus !== "rejected",
+    saturated,
     hasQuota: !!q,
     coolingUntil: st?.coolingUntil ?? 0,
-    weeklyReset: weeklyReset ?? Number.POSITIVE_INFINITY,
+    sessionResetBucket: sessionTicking
+      ? Math.floor(sessionReset / sessionResetToleranceMs)
+      : Number.POSITIVE_INFINITY,
     sessionReset: sessionReset ?? Number.POSITIVE_INFINITY,
-    weeklyUsed: q?.weeklyUsed ?? -1,
+    weeklyReset: weeklyReset ?? Number.POSITIVE_INFINITY,
+    weeklyUsed,
   };
 }
 
 /**
  * Order accounts to MAXIMIZE quota utilization (fill-first, smart order):
- * spend the account whose window refreshes SOONEST first, so its about-to-reset
+ * spend the window that expires SOONEST first, so its about-to-reset
  * allowance isn't wasted, then move to accounts with longer-dated resets.
  *
  * Priority among usable accounts:
  *   1. no quota data yet — probe first: one request reveals its windows and
  *      self-corrects the ordering. (Ranking unknowns last would starve them
  *      forever: never picked → never observed → never comparable.)
- *   2. soonest WEEKLY (7d) reset  — the scarce, use-it-or-lose-it ceiling
- *   3. soonest SESSION (5h) reset
- *   4. highest weekly utilization — finish off the one closest to done
- *   5. configured primary account, then insertion order
+ *   2. session headroom before session-saturated (>= soft limit or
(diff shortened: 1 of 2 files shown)


src/lib/server/routes/claudeProxyRoutes.ts
```diff
@@ -581,40 +581,97 @@ function isQuotaRoutingEnabled(): boolean {
...
+  const sessionTicking = sessionReset !== undefined;
+  const weeklyTicking = weeklyReset !== undefined;
+  const sessionUsed = sessionTicking ? (q?.sessionUsed ?? 0) : 0;
+  const weeklyUsed = weeklyTicking ? (q?.weeklyUsed ?? -1) : q ? 0 : -1;
+  const sessionStatus = sessionTicking
+    ? (q?.sessionStatus ?? "unknown")
+    : "allowed";
+  const weeklyStatus = weeklyTicking
+    ? (q?.weeklyStatus ?? "unknown")
+    : "allowed";
```
Tara-ag, line 641: 💡 **Minor: env parsers run inside the comparator**

`getSessionSoftLimit()` and `getSessionResetToleranceMs()` read and parse `process.env` on every invocation of `accountSortMetrics`, which is called O(N log N) times per `orderAccountsByQuota` sort. More importantly, if the env value changes mid-sort, the comparator can see different thresholds/buckets for different pairs and break transitivity.

Hoist both values once at the top of `orderAccountsByQuota` and pass them into `accountSortMetrics`:

```typescript
function orderAccountsByQuota(accounts, now) {
  const sessionSoftLimit = getSessionSoftLimit();
  const sessionResetToleranceMs = getSessionResetToleranceMs();
  // ... accountSortMetrics(a.key, now, sessionSoftLimit, sessionResetToleranceMs)
}
```

src/lib/server/routes/claudeProxyRoutes.ts
```diff
@@ -637,12 +694,20 @@ function orderAccountsByQuota(
     if (ma.hasQuota !== mb.hasQuota) {
       return ma.hasQuota ? 1 : -1;
     }
+    if (ma.saturated !== mb.saturated) {
+      return ma.saturated ? 1 : -1;
+    }
+    if (ma.saturated && mb.saturated) {
```
Tara-ag: ⚠️ **Major: saturated-account tie-break bypasses configured tolerance and weekly ordering**

When both accounts are saturated, this branch compares exact `sessionReset` epochs before the bucket/weekly logic. That means two saturated accounts whose resets fall inside the same tolerance bucket will be ordered by exact epoch rather than by weekly reset, contradicting the documented "same-bucket → weekly decides" rule and ignoring `NEUROLINK_PROXY_SESSION_RESET_TOLERANCE_MS`.

Drop the special-case saturated branch and let the normal `sessionResetBucket` → `weeklyReset` → `weeklyUsed` chain handle saturated accounts too. Saturated accounts already sort after headroom accounts via the `ma.saturated !== mb.saturated` check above, so within the saturated group the same bucket/weekly rules should apply.

```diff
-    if (ma.saturated !== mb.saturated) {
-      return ma.saturated ? 1 : -1;
-    }
-    if (ma.saturated && mb.saturated) {
-      if (ma.sessionReset !== mb.sessionReset) {
-        return ma.sessionReset - mb.sessionReset;
-      }
-    }
-    if (ma.sessionResetBucket !== mb.sessionResetBucket) {
+    if (ma.saturated !== mb.saturated) {
+      return ma.saturated ? 1 : -1;
+    }
+    if (ma.sessionResetBucket !== mb.sessionResetBucket) {
```

Then add a test case with two saturated accounts in the same session bucket but different weekly resets to lock in the weekly tie-break.

test/continuous-test-suite-proxy.ts
```diff
@@ -1730,6 +1730,146 @@ async function testOrderAccountsByQuota(): Promise<boolean | null> {
   return true;
 }
 
+// ============================================================================
+// Tests: session-first ordering, soft limit, and reset freshening
+// ============================================================================
+
+async function testSessionFirstOrdering(): Promise<boolean | null> {
```
Tara-ag, line 1737: 💡 **Minor: test mutates global env without isolating both variables**

Case 8 sets `NEUROLINK_PROXY_SESSION_SOFT_LIMIT` and deletes it in a `finally`, but the earlier cases rely on the default soft-limit and tolerance values. If the test runner happens to have either variable configured, cases 3–7 or the bucket-math in case 2 can fail or leave the process env mutated for later tests.

Wrap the whole function in a save/set/restore block for both `NEUROLINK_PROXY_SESSION_SOFT_LIMIT` and `NEUROLINK_PROXY_SESSION_RESET_TOLERANCE_MS`:

```typescript
const savedSoft = process.env.NEUROLINK_PROXY_SESSION_SOFT_LIMIT;
const savedTol = process.env.NEUROLINK_PROXY_SESSION_RESET_TOLERANCE_MS;
process.env.NEUROLINK_PROXY_SESSION_SOFT_LIMIT = "0.97";
process.env.NEUROLINK_PROXY_SESSION_RESET_TOLERANCE_MS = String(15 * 60 * 1000);
try {
  // ... all cases ...
} finally {
  if (savedSoft === undefined) delete process.env.NEUROLINK_PROXY_SESSION_SOFT_LIMIT;
  else process.env.NEUROLINK_PROXY_SESSION_SOFT_LIMIT = savedSoft;
  if (savedTol === undefined) delete process.env.NEUROLINK_PROXY_SESSION_RESET_TOLERANCE_MS;
  else process.env.NEUROLINK_PROXY_SESSION_RESET_TOLERANCE_MS = savedTol;
}
```

Tara-ag (approved):
## Review Summary

**Files reviewed:** 2
**New issues raised:** 1 MINOR
**Blocking issues:** 0

### Notes

- The session-first ordering, soft-limit demotion, and reset-freshening logic in `claudeProxyRoutes.ts` is sound and well-documented.
- Added one minor inline comment: the env parsers (`getSessionSoftLimit` / `getSessionResetToleranceMs`) are invoked inside `accountSortMetrics`, which is called repeatedly during sorting. Hoisting the parsed values once per `orderAccountsByQuota` call avoids both an O(N log N) env re-parse and the risk of a mid-sort env change breaking comparator transitivity.
- Two of my observations overlap with feedback already posted by CodeRabbit (saturated-account exact `sessionReset` comparison and test env-variable isolation). I did not re-raise those as new blocking concerns.

No blocking criteria are met; approving.

pdogra1299:
Thanks @coderabbitai and @Tara-ag — all five findings verified and fixed, amended into the single commit (`9c0e71d4`):

- **Partial-numeric env values** — both parsers now use `Number()` (tolerance additionally requires an integer), so `"0.5oops"` / `"900000oops"` fall back to defaults instead of being silently truncated.
- **Saturated tie-break honors tolerance + weekly** (flagged by both reviewers) — removed the exact `sessionReset` comparison; saturated pairs now fall through to the same bucketed session ordering, with weekly deciding same-bucket ties. New test case covers two saturated same-bucket accounts with opposing weekly resets.
- **Env parsers hoisted out of the comparator** — `orderAccountsByQuota` reads both thresholds once per sort and passes them into `accountSortMetrics`, so a mid-sort env change can never break transitivity (and it's cheaper).
- **Test env isolation** — `testSessionFirstOrdering` saves both quota env vars, clears them for deterministic defaults, and restores them in an outer `finally`.

Re-verified: tsc/lint gates clean, 6/6 runtime checks for the fixes (garbage env → defaults, same-bucket saturated weekly tie-break, plus regressions), 36/36 `proxyReliabilityHardening` vitest tests, full pre-commit build.

Tara-ag (approved):
## Review Summary

**Files reviewed:** 2 (`src/lib/server/routes/claudeProxyRoutes.ts`, `test/continuous-test-suite-proxy.ts`)
**New issues raised:** 0
**Decision:** Approve

### Assessment

This PR refines the proxy's quota-aware account ordering to prioritize the expiring 5h session window over the 7d weekly window, adds a soft-limit demotion for saturated sessions, and implements reset freshening at read time. The changes are localized to the proxy routing comparator and its test suite.

Key points verified:

- **Comparator transitivity:** Session resets are compared via fixed-grid buckets (`Math.floor(sessionReset / tolerance)`), which is transitive by construction. The thresholds are read once per sort, so mid-sort env changes cannot break ordering.
- **Env parsing:** `Number()` is used instead of `parseFloat`/`parseInt`, correctly rejecting partially numeric strings. The tolerance is required to be an integer.
- **Reset freshening:** Windows whose reset epoch has passed are treated as fresh (`sessionUsed = 0`, status `"allowed"`), preventing stale snapshots from saturating an account after Anthropic has renewed its window.
- **Soft-limit behavior:** Saturated/throttled accounts are demoted but remain usable as fallbacks, matching the documented intent.
- **Test coverage:** The new `testSessionFirstOrdering` suite covers session-first vs weekly, same-bucket weekly tie-break, soft-limit boundary, throttled demotion, reset freshening, saturated recovery, and env-configurable limit. It also saves/restores both quota-related env variables.

### Existing review comments

The outstanding review comments from earlier passes appear addressed in the current revision:

- Env parsers are hoisted outside the comparator in `orderAccountsByQuota`.
- Saturated accounts no longer use an exact `sessionReset` tie-break; they fall through to the same bucketed session → weekly → weekly-used chain.
- The test wrapper saves and restores both `NEUROLINK_PROXY_SESSION_SOFT_LIMIT` and `NEUROLINK_PROXY_SESSION_RESET_TOLERANCE_MS`.

No new blocking or major issues were found. Approving.

PR #1173, opened 2026-07-16
fix(proxy): make updater and stats self-consistent

murdore:
## Summary

- supervise and respawn the auto-updater worker while the proxy parent is alive
- retry installed CLI validation and reconcile an installed-but-not-yet-running version without suppressing a valid release
- separate upstream attempt failures from final request failures in global and per-account stats
- classify 429 attempts as transient admission throttles or account quota exhaustion
- preserve strict terminal failure reporting for genuinely incomplete SSE streams

## Root causes

The updater could exit silently and was never respawned. A single transient executable probe after installation also marked the installed version as broken and suppressed it for 24 hours, even though the same trampoline worked moments later.

The status mismatch came from comparing different units: global errors represented final user-request outcomes, while account errors included every failed upstream attempt. One request that received three transient 429s and then succeeded on another account therefore appeared as one global success but three account errors.

The investigated final failure was not a fake rate limit or proxy crash. The upstream socket closed after 131 seconds during thinking, before content_block_stop, message_delta, or message_stop, so the response was genuinely incomplete and must remain a failure.

## Verification

- `pnpm check`
- `pnpm lint` (0 errors; existing warning baseline only)
- repository pre-commit validation, security checks, production/package builds, and `publint`
- `pnpm exec vitest run test/proxyUpdaterFallback.test.ts test/proxyReliabilityHardening.test.ts` (54/54 passing)

No live proxy restart, deployment, or synthetic live API request was performed.

## Summary by CodeRabbit

* **New Features**
  * Expanded proxy `/status` and `proxy status --format json` with detailed auto-update state, pending restart info, and structured update failure details.
  * Improved usage statistics by separating final outcomes from attempt-level errors, including transient vs quota rate-limit breakdowns and new attempt error counters.
  * Enhanced attempt telemetry for retries and rate-limit classifications (retryable, cooldown reason, transient vs quota).
  * Added supervised updater worker recovery and more robust post-install global CLI version validation.
* **Documentation**
  * Updated proxy documentation and observability examples to reflect the richer auto-update/status fields and the clarified attempt vs final-outcome semantics.

### DIFF
--- a/docs/features/claude-proxy.md
+++ b/docs/features/claude-proxy.md
@@ -544,7 +544,7 @@ The proxy persists its running state to `~/.neurolink/proxy-state.json` so that 
 
 ### Fail-open guard
 
-A foreground proxy spawns one detached `neurolink proxy guard` that removes stale Claude Code settings after confirming its parent process has died. A launchd-managed proxy keeps restart ownership in launchd and starts a separate updater-only worker. Automatic package updates are enabled by default and can be disabled with `NEUROLINK_PROXY_AUTO_UPDATE=off` (also accepts `0` or `false`). The worker validates the global package root and executable directory, requires the package manager that owns the running installation, and waits for every request and stream to finish plus a two-minute idle window before asking launchd to restart. Updater diagnostics are written to `~/.neurolink/logs/proxy-updater.log` and exposed in `neurolink proxy status`.
+A foreground proxy spawns one detached `neurolink proxy guard` that removes stale Claude Code settings after confirming its parent process has died. A launchd-managed proxy keeps restart ownership in launchd and starts a separate updater-only worker. Automatic package updates are enabled by default and can be disabled with `NEUROLINK_PROXY_AUTO_UPDATE=off` (also accepts `0` or `false`). The worker validates the global package root and executable directory, requires the package manager that owns the running installation, and waits for every request and stream to finish plus a two-minute idle window before asking launchd to restart. Post-install executable validation uses bounded retries, and the proxy supervises and replaces an updater worker that exits unexpectedly. Installed-but-not-running versions and stage-specific update failures remain persisted in `~/.neurolink/update-state.json`. Updater diagnostics are written to `~/.neurolink/logs/proxy-updater.log` and exposed in `neurolink proxy status`.
 
 ## Architecture
 
@@ -639,11 +639,13 @@ This prevents unbounded log growth without requiring external cron jobs.
 
 In-memory per-account statistics track:
 
-- Upstream attempt count, success count, error count, rate-limit count
-- Current backoff level and cooling state
-- Last attempt and last error timestamps
+- Final completed, successful, and failed request counts
+- Upstream attempts and failed attempts, including authentication retries,
+  network failures, and retries that later recovered
+- Transient-throttle and exhausted-quota attempt counts
+- Current account cooling state
 
-Proxy-wide status also tracks total upstream attempts separately from completed requests. Statistics reset on proxy restart. Access them via the `/status` endpoint or `neurolink proxy status`.
+Final request counters add up across accounts and match proxy-wide completed, success, and error totals. Attempt counters are intentionally separate because one final request can make several attempts or rotate accounts. Statistics reset on proxy restart. Access them via the `/status` endpoint or `neurolink proxy status`.
 
 ## Comparison with CLIProxyAPI
 
--- a/src/cli/commands/proxy.ts
+++ b/src/cli/commands/proxy.ts
@@ -67,7 +67,19 @@ import {
   describeInstallFailure,
   getGlobalInstallArgs,
   resolveGlobalInstaller,
+  validateInstalledVersion,
 } from "../../lib/proxy/globalInstaller.js";
+import { startUpdaterWorkerSupervisor } from "../../lib/proxy/updaterSupervisor.js";
+import {
+  abandonPendingUpdate,
+  isVersionSuppressed,
+  loadUpdateState,
+  recordCheck,
+  recordSuccessfulUpdate,
+  recordUpdateFailure,
+  recordUpdateInstalled,
+  suppressVersion,
+} from "../../lib/proxy/updateState.js";
 import {
   loadProxyEnvFile,
   resolveProxyEnvFile,
@@ -725,6 +737,16 @@ function spawnProxyUpdater(
       detached: true,
       stdio: ["ignore", logFd, logFd],
       env: process.env,
+    });
+    child.once("error", (error) => {
+      logger.always(
+        `[proxy] updater worker error pid=${child.pid ?? "unknown"}: ${error.message}`,
+      );
+    });
+    child.once("exit", (code, signal) => {
+      logger.always(
+        `[proxy] updater worker exited pid=${child.pid ?? "unknown"} code=${code ?? "none"} signal=${signal ?? "none"}`,
+      );
     });
     child.unref();
     return child.pid;
@@ -1391,6 +1413,7 @@ export async function createProxyStartApp(params: {
       await import("../../lib/proxy/accountCooldown.js");
     const stats = getStats();
     const runtimeState = loadProxyState();
+    const updateState = loadUpdateState();
     const cooldowns = await loadAccountCooldowns();
     const storedAccountKeys = new Set<string>();
     try {
@@ -1428,10 +1451,13 @@ export async function createProxyStartApp(params: {
       health,
       stats: {
         totalAttempts: stats.totalAttempts,
+        totalAttemptErrors: stats.totalAttemptErrors,
         totalRequests: stats.totalRequests,
         totalSuccess: stats.totalSuccess,
         totalErrors: stats.totalErrors,
         totalRateLimits: stats.totalRateLimits,
+        totalTransientRateLimits: stats.totalTransientRateLimits,
+        totalQuotaRateLimits: stats.totalQuotaRateLimits,
         accounts: Object.values(stats.accounts).map((account) => {
           const normalizedKey = normalizeAnthropicAccountKey(account.label);
           const accountKey = storedAccountKeys.has(normalizedKey)
@@ -1445,10 +1471,13 @@ export async function createProxyStartApp(params: {
             label: account.label,
             type: account.type,
             attempts: account.attemptCount,
-            requests: account.attemptCount,
+            requests: account.successCount + account.errorCount,
             success: account.successCount,
             errors: account.errorCount,
+            attemptErrors: account.attemptErrorCount,
             rateLimits: account.rateLimitCount,
+            transientRateLimits: account.transientRateLimitCount,
+            quotaRateLimits: account.quotaRateLimitCount,
             cooling: (cooldowns[accountKey]?.coolingUntil ?? 0) > now,
           };
         }),
@@ -1467,6 +1496,19 @@ export async function createProxyStartApp(params: {
         updaterRunning: runtimeState?.updaterPid
           ? isProcessRunning(runtimeState.updaterPid)
           : false,
+        liveVersion: PROXY_VERSION,
+        latestVersion: updateState?.lastCheckVersion || null,
+        pendingRestartVersion: updateState?.pendingRestartVersion ?? null,
+        lastCheckAt: updateState?.lastCheckAt ?? null,
+        lastUpdateAt: updateState?.lastUpdateAt ?? null,
+        lastUpdateVersion: updateState?.lastUpdateVersion ?? null,
+        lastFailure: updateState?.lastFailure
+          ? {
+              at: updateState.lastFailure.at,
+              version: updateState.lastFailure.version,
+              stage: updateState.lastFailure.stage,
+            }
+          : null,
       },
       config: params.proxyConfig
         ? {
@@ -1751,6 +1793,7 @@ function registerProxyShutdownHandlers(params: {
   isDev?: boolean;
   refreshInterval: NodeJS.Timeout;
   logCleanupInterval: NodeJS.Timeout;
+  updaterSupervisor?: { stop: () => void };
 }): void {
   let shutdownStarted = false;
 
@@ -1793,6 +1836,7 @@ function registerProxyShutdownHandlers(params: {
     shutdownStarted = true;
     clearInterval(params.refreshInterval);
     clearInterval(params.logCleanupInterval);
+    params.updaterSupervisor?.stop();
     logger.always(`\nShutting down proxy (${signal})...`);
     let exitCode = signal === "SIGINT" ? 0 : 1;
 
@@ -1893,10 +1937,39 @@ async function startProxyRuntime(params: {
     port: params.port,
   });
   markProxyReady(params.readiness);
-  const updaterPid =
+  try {
+    const { reconcileRunningUpdate } =
+      await import("../../lib/proxy/updateState.js");
+    if (reconcileRunningUpdate(PROXY_VERSION)) {
+      logger.always(
+        `[proxy] confirmed pending update is now running at v${PROXY_VERSION}`,
+      );
+    }
+  } catch (error) {
+    logger.always(
+      `[proxy] WARNING: failed to reconcile update state: ${error instanceof Error ? error.message : String(error)}`,
+    );
+  }
+  /** Mirror the supervised worker PID into the proxy state used by status. */
+  const updatePersistedUpdaterPid = (updaterPid: number | undefined): void => {
+    const state = loadProxyState();
+    if (!state || state.pid !== process.pid) {
+      return;
+    }
+    saveProxyState({ ...state, updaterPid });
+  };
+  const updaterSupervisor =
     managedByLaunchd && !params.argv.dev
-      ? spawnProxyUpdater(readinessHost, params.port, process.pid)
+      ? startUpdaterWorkerSupervisor({
+          spawnWorker: () =>
+            spawnProxyUpdater(readinessHost, params.port, process.pid),
+          isProcessRunning,
+          stopWorker: (pid) => process.kill(pid, "SIGTERM"),
+          onPidChange: updatePersistedUpdaterPid,
+          log: (message) => logger.always(message),
+        })
       : undefined;
+  const updaterPid = updaterSupervisor?.currentPid();
   const fallbackChain: FallbackInfo[] | undefined =
     params.proxyConfig?.routing?.fallbackChain?.map((entry) => ({
       provider: entry.provider as string,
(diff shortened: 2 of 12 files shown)


src/lib/proxy/updateState.ts
```diff
@@ -89,7 +91,20 @@ export function loadUpdateState(stateFilePath?: string): UpdateState | null {
...
+      ...candidate,
+      suppressedVersions: candidate.suppressedVersions ?? {},
+      pendingRestartVersion:
+        typeof candidate.pendingRestartVersion === "string"
+          ? candidate.pendingRestartVersion
+          : null,
+      lastFailure:
+        candidate.lastFailure && typeof candidate.lastFailure === "object"
+          ? candidate.lastFailure
+          : null,
```
Tara-ag: 💡 **MINOR**: `loadUpdateState()` accepts any object for `lastFailure` without validating required fields. A partially-corrupt state file such as `{ "lastFailure": {} }` would surface as `undefined` stage/message in `neurolink proxy status`.

Consider sanitizing `lastFailure` to `null` unless it matches the expected shape (object with string `at`, `version`, `stage`, `message` and an allowed `stage` value):

```ts
function isValidLastFailure(value: unknown): value is NonNullable<UpdateState["lastFailure"]> {
  if (!value || typeof value !== "object") return false;
  const v = value as Record<string, unknown>;
  const stages = ["check", "install", "validation", "restart", "health"] as const;
  return (
    typeof v.at === "string" &&
    typeof v.version === "string" &&
    typeof v.stage === "string" &&
    stages.includes(v.stage as typeof stages[number]) &&
    typeof v.message === "string"
  );
}
```

Then use `isValidLastFailure(candidate.lastFailure) ? candidate.lastFailure : null`.

src/cli/commands/proxy.ts
```diff
@@ -2685,96 +2797,110 @@ export const proxyGuardCommand: CommandModule<object, ProxyGuardArgs> = {
...
+            return;
+          }
+          logger.always(
+            `[updater] installing @juspay/neurolink@${result.latestVersion} via ${installer.kind} ${installer.bin} (v${installer.version})`,
+          );
+          if (guardStopping || getProcessStatus(parentPid) === "not_running") {
+            return;
+          }
+          try {
+            execFileSync(
```
Tara-ag, line 2883: ⚠️ **MAJOR**: `recordUpdateInstalled()` is called immediately after `execFileSync` succeeds, before trampoline validation. If validation repeatedly fails (e.g., version mismatch), `pendingRestartVersion` stays set and the next cycle takes the `pendingRestart` path that skips reinstalling. This can loop forever on a bad install state without ever clearing the suppression.

Per the PR goal of reconciling installed-but-not-yet-running versions, record the pending restart only after validation succeeds, or explicitly clear `pendingRestartVersion` on validation failure so a broken install is retried.

Suggested fix:
```ts
// Move recordUpdateInstalled() to after validation.version === result.latestVersion
// On validation failure, add:
persistUpdaterState("clear pending restart on validation failure", () => {
  const s = loadUpdateState();
  if (s && s.pendingRestartVersion === result.latestVersion) {
    s.pendingRestartVersion = null;
    saveUpdateState(s);
  }
});
```

src/lib/proxy/updaterSupervisor.ts
```diff
@@ -0,0 +1,48 @@
...
+  let pid: number | undefined;
+
+  const checkNow = (): number | undefined => {
+    if (stopped) {
+      return pid;
+    }
+    if (pid !== undefined && options.isProcessRunning(pid)) {
+      return pid;
+    }
+
```
Tara-ag, line 21: ⚠️ **MAJOR**: `options.spawnWorker()` is invoked without a try/catch, so a thrown exception (e.g., from `spawnProxyUpdater` or a future caller) will propagate out of `checkNow`. Because `checkNow` runs both at startup and inside a `setInterval` callback in the main proxy process, an unhandled exception here can crash the proxy parent.

Wrap the spawn in a try/catch and treat a thrown spawn as an unavailable worker:

```ts
try {
  pid = options.spawnWorker();
} catch (error) {
  options.log?.(`[proxy] updater worker spawn failed: ${error instanceof Error ? error.message : String(error)}`);
  pid = undefined;
}
```

src/lib/proxy/updateState.ts
Tara-ag: ⚠️ **MAJOR**: `lastFailure` is accepted as any object and spread directly into the returned state. A partially-corrupt state file such as `{ "lastFailure": {} }` will surface `undefined` values for `stage`/`message` in `neurolink proxy status`, and the `stage` type annotation is trusted at compile time but not enforced at runtime.

Sanitize `lastFailure` to `null` unless it matches the expected shape and an allowed stage:

```ts
function isValidLastFailure(value: unknown): value is NonNullable<UpdateState["lastFailure"]> {
  if (!value || typeof value !== "object") return false;
  const v = value as Record<string, unknown>;
  const stages = ["check", "install", "validation", "restart", "health"] as const;
  return (
    typeof v.at === "string" &&
    typeof v.version === "string" &&
    typeof v.stage === "string" &&
    stages.includes(v.stage as typeof stages[number]) &&
    typeof v.message === "string"
  );
}
```

Then use `isValidLastFailure(candidate.lastFailure) ? candidate.lastFailure : null`.

src/cli/commands/proxy.ts
```diff
@@ -2685,96 +2829,110 @@ export const proxyGuardCommand: CommandModule<object, ProxyGuardArgs> = {
...
```
Tara-ag, line 2883: ⚠️ **MAJOR**: `recordUpdateInstalled()` is called immediately after `execFileSync` succeeds, before trampoline validation. If validation repeatedly fails (e.g., version mismatch), `pendingRestartVersion` stays set and the next cycle takes the `pendingRestart` path that skips reinstalling. This can loop forever on a bad install state without ever clearing the suppression.

Per the PR goal of reconciling installed-but-not-yet-running versions, record the pending restart only after validation succeeds, or explicitly clear `pendingRestartVersion` on validation failure so a broken install is retried.

Suggested fix:
```ts
// Move recordUpdateInstalled() to after validation.version === result.latestVersion
// On validation failure, add:
persistUpdaterState("clear pending restart on validation failure", () => {
  const s = loadUpdateState();
  if (s && s.pendingRestartVersion === result.latestVersion) {
    s.pendingRestartVersion = null;
    saveUpdateState(s);
  }
});
```

src/lib/proxy/globalInstaller.ts
```diff
@@ -183,6 +185,56 @@ export function getGlobalInstallArgs(
     : ["install", "--global", "--no-audit", "--no-fund", packageSpec];
 }
 
+/**
```
Tara-ag, line 188: 💡 **MINOR**: `validateInstalledVersion` accepts arbitrary user input for `binPath` and passes it directly to `execFileSync`. While this is currently only called with the hardcoded `TRAMPOLINE_PATH`, the function is exported and typed for general use. Consider validating that `binPath` is an absolute path and rejecting shell metacharacters/relative paths to prevent accidental command injection if a future caller passes request-derived input.

```ts
if (!path.isAbsolute(options.binPath)) {
  return { attempts: 0, failure: `binPath must be absolute: ${options.binPath}` };
}
```

src/lib/proxy/usageStats.ts
```diff
@@ -39,13 +42,22 @@ export function recordAttemptError(
   accountLabel: string,
```
Tara-ag, line 42: 💡 **MINOR**: `recordAttemptError` increments `totalAttemptErrors` unconditionally for every non-2xx attempt, but the per-account `attemptErrorCount` only increments for the account passed in. This is consistent with the new accounting model, but note that `totalAttemptErrors` will include auth-retry failures, network errors, and terminal errors alike. Consider whether the docs/status output should clarify that `failed attempts` includes all non-success upstream attempts (retries, auth retries, network errors) rather than only rate-limited attempts, to avoid users interpreting it as "only rate limits".

Tara-ag (approved):
## Review summary

Reviewed the 12 changed files in this PR. The change hardens the proxy updater lifecycle and separates upstream attempt failures from final request outcomes, which aligns with the stated goals.

### New issues raised in this review

| Severity | File | Line | Topic |
| ⚠️ MAJOR | `src/lib/proxy/updaterSupervisor.ts` | 20 | `spawnWorker()` is invoked without a try/catch; a thrown spawn exception in the supervisor's `setInterval` callback can crash the proxy parent. Wrap the spawn and treat a thrown spawn as an unavailable worker. |
| 💡 MINOR | `src/lib/proxy/globalInstaller.ts` | 188 | `validateInstalledVersion` accepts arbitrary `binPath` input and passes it to `execFileSync`. Currently only called with the hardcoded trampoline path, but because it is exported and generically typed, validate that `binPath` is absolute and free of shell metacharacters to prevent future command-injection misuse. |
| 💡 MINOR | `src/lib/proxy/usageStats.ts` | 42 | `totalAttemptErrors` now counts every non-2xx upstream attempt (including auth retries and network errors). The docs/status text should make clear that "failed attempts" is an upstream-attempt diagnostic, not a rate-limit-only counter, so users do not misread it. |

### Already-active comments (not duplicated)

Two points were already raised by prior reviewers on this PR and remain unresolved/outstanding; I did not re-raise them as new grounds for blocking:

- `src/lib/proxy/updateState.ts:106` — runtime validation of `lastFailure` shape/stage.
- `src/cli/commands/proxy.ts:2872` — `recordUpdateInstalled()` being called before trampoline validation completes.

### Decision

**Approve.** No new blocking criteria are met this run (only one new MAJOR issue; blocking requires three or more). The outstanding points from prior reviewers should be addressed before merge, but that is already tracked in existing comment threads.

src/lib/proxy/updateState.ts
murdore: Fixed in 7677fbf3. Persisted lastFailure now receives full runtime shape and allowed-stage validation, with malformed records normalized to null.

src/lib/proxy/usageStats.ts
murdore: Fixed in 7677fbf3. The docs now explicitly define failed attempts as all failed upstream attempts, including auth retries, network failures, and recovered retries, rather than only rate limits.

src/cli/commands/proxy.ts
```diff
@@ -2623,32 +2755,44 @@ export const proxyGuardCommand: CommandModule<object, ProxyGuardArgs> = {
...
+          recordCheck(result.latestVersion),
+        );
 
         if (!result.updateAvailable) {
           return;
         }
-        if (isVersionSuppressed(result.latestVersion)) {
+        const pendingRestart =
+          loadUpdateState()?.pendingRestartVersion === result.latestVersion;
+        if (isVersionSuppressed(result.latestVersion) && !pendingRestart) {
```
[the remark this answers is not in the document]
murdore: Fixed in 7677fbf3. Unhealthy-release handling now abandons the matching pending version before suppression in both update flows, preventing a subsequent worker from bypassing suppression.

src/lib/proxy/globalInstaller.ts
murdore: Fixed in 7677fbf3 by requiring an absolute executable path, with regression coverage for relative paths. No shell-character filter is needed because execFileSync invokes the executable directly with shell disabled, so shell syntax is not interpreted.

src/cli/commands/proxy.ts
murdore: Fixed in 7677fbf3. recordUpdateInstalled was moved after successful validation, and all matching validation-failure paths clear the pending version.

docs/features/claude-proxy.md
```diff
@@ -639,11 +639,13 @@ This prevents unbounded log growth without requiring external cron jobs.
...
 
-- Upstream attempt count, success count, error count, rate-limit count
+- Final completed, successful, and failed request counts
+- Upstream attempts and failed attempts, including retries that recovered
+- Transient-throttle and exhausted-quota attempt counts
 - Current backoff level and cooling state
 - Last attempt and last error timestamps
 
-Proxy-wide status also tracks total upstream attempts separately from completed requests. Statistics reset on proxy restart. Access them via the `/status` endpoint or `neurolink proxy status`.
+Final request counters add up across accounts and match proxy-wide completed, success, and error totals. Attempt counters are intentionally separate because one final request can make several attempts or rotate accounts. Statistics reset on proxy restart. Access them via the `/status` endpoint or `neurolink proxy status`.
```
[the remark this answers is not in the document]
murdore: Fixed in 7677fbf3. The unsupported backoff-level and timestamp claims were removed so the documentation matches the actual status contract.

src/cli/commands/proxy.ts
murdore: Fixed in 7677fbf3. The pending version is persisted only after successful trampoline validation and is abandoned on validation failure.

src/lib/proxy/updaterSupervisor.ts
murdore: Fixed in 7677fbf3. spawnWorker exceptions are contained and logged by the supervisor, leaving it able to retry on the next interval. Tests cover the thrown-spawn path.

src/lib/proxy/updaterSupervisor.ts
```diff
@@ -0,0 +1,48 @@
...
+  const interval = setInterval(checkNow, options.intervalMs ?? 30_000);
+  interval.unref?.();
+
+  return {
+    currentPid: () => pid,
+    checkNow,
+    stop: () => {
+      stopped = true;
+      clearInterval(interval);
+    },
```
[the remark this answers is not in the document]
murdore: Fixed in 7677fbf3. stop is now idempotent, terminates the active worker with SIGTERM, clears its in-memory and persisted PID, and stops the interval. Regression coverage verifies shutdown behavior.

src/lib/server/routes/claudeProxyRoutes.ts
```diff
@@ -4248,6 +4236,19 @@ async function handleAnthropicAuthRetry(args: {
...
+          account.label,
+          account.type,
+          retryStatus,
+          rateLimitKind,
+        );
+        logAttempt(429, "rate_limit_error", retryBody, {
+          retryable: true,
+          rateLimitKind,
+          cooldownReason: retryPlan.reason,
+        });
```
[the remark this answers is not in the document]
murdore: Fixed in 7677fbf3. Every OAuth retry allocates and records a unique upstream attempt before fetch, and all retry logs and captures use that attempt ID. Regression coverage verifies totals and attempt numbering.

src/cli/commands/proxy.ts
```diff
@@ -2208,19 +2273,61 @@ function printStatusStats(stats: StatusStats): void {
...
+    ];
+    const rows = stats.accounts.map((account) => [
+      account.label,
+      account.type,
+      String(account.attempts ?? account.requests ?? 0),
+      String(account.success ?? 0),
+      String(account.errors ?? 0),
+      String(account.rateLimits ?? 0),
+      account.cooling ? "cooling" : "active",
+    ]);
```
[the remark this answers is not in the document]
murdore: Reviewed. The compact table intentionally remains ACCOUNT, AUTH, ATTEMPTS, SUCCESS, ERRORS, RL, STATUS to match the requested CLI contract. The detailed attempt-error and transient/quota split remains available through JSON/status and the aggregate diagnostics, avoiding an overly wide terminal table.

src/lib/proxy/updateState.ts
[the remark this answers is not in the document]
murdore: Fixed in 7677fbf3. Persisted lastFailure is accepted only when every required field and the stage enum are valid; malformed objects and arrays normalize to null. Regression coverage was added.

src/cli/commands/proxy.ts
```diff
@@ -2685,96 +2797,110 @@ export const proxyGuardCommand: CommandModule<object, ProxyGuardArgs> = {
...
+                `@juspay/neurolink@${result.latestVersion}`,
+              ),
+              {
+                timeout: 120_000,
+                stdio: "pipe",
+              },
+            );
+            persistUpdaterState("record installed update", () =>
+              recordUpdateInstalled(result.latestVersion),
+            );
```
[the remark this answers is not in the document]
murdore: Fixed in 7677fbf3. Pending restart is now recorded only after trampoline validation succeeds, and validation failures abandon the matching pending version. Regression coverage was added.

src/lib/proxy/updateState.ts
murdore: Fixed in 7677fbf3. Runtime validation now rejects arrays, incomplete objects, and invalid stages before loading lastFailure.

src/cli/commands/proxy.ts
```diff
@@ -1467,6 +1495,13 @@ export async function createProxyStartApp(params: {
         updaterRunning: runtimeState?.updaterPid
           ? isProcessRunning(runtimeState.updaterPid)
           : false,
+        liveVersion: PROXY_VERSION,
+        latestVersion: updateState?.lastCheckVersion || null,
+        pendingRestartVersion: updateState?.pendingRestartVersion ?? null,
+        lastCheckAt: updateState?.lastCheckAt ?? null,
+        lastUpdateAt: updateState?.lastUpdateAt ?? null,
+        lastUpdateVersion: updateState?.lastUpdateVersion ?? null,
+        lastFailure: updateState?.lastFailure ?? null,
```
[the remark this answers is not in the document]
murdore: Fixed in 7677fbf3. The unauthenticated status endpoint now exposes only at, version, and stage; the full diagnostic message remains available to the local CLI. A regression test verifies the raw message is absent.

src/lib/proxy/updateState.ts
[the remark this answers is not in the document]
murdore: Fixed in 7677fbf3. lastFailure normalization now validates the complete shape and allowed stage set, rejecting arrays and arbitrary objects.

test/proxyReliabilityHardening.test.ts
```diff
@@ -1111,12 +1157,16 @@ describe("launchd lifecycle source invariants", () => {
     expect(source).toContain("NEUROLINK_PROXY_AUTO_UPDATE");
     expect(source).toContain('["0", "off", "false"]');
     expect(source).toContain("spawnProxyUpdater");
+    expect(source).toContain("startUpdaterWorkerSupervisor");
     expect(source).toContain("updaterOnly &&");
+    expect(source).toContain("!updaterOnly &&");
```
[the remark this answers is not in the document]
murdore: Validated against v9.93.1: still valid. Fixed in #1208 by matching the complete positive and negated updater guard expressions so the positive assertion cannot pass from the negated branch. Resolving.

PR #1180, opened 2026-07-16
fix(cli): play mp3 TTS on Linux via real decoders, not paplay (#1138)

### LINKED ISSUE #1138: TTS --tts-play: mp3 playback broken on Linux (paplay can't decode mp3; default format)
## Summary
The `--tts-play` CLI audio playback added in #982 (merged as `0e6580fc`) does not work on **Linux** for the **default `mp3` format**. `src/cli/utils/audioPlayer.ts` routes all non-WAV audio to `paplay`, which decodes only libsndfile-backed formats (WAV/FLAC/Ogg/AIFF) and **cannot decode mp3/AAC**. The ENOENT fallback to `aplay` can't decode mp3 either.

Since `--tts-format` defaults to `mp3`, the default `neurolink ... --tts --tts-play` invocation fails on Linux. It fails *gracefully* (playback errors are caught and shown as a non-fatal warning — no crash), but the message ("Install PulseAudio (paplay) or ALSA (aplay)") is misleading because PulseAudio is present; the format is simply undecodable by `paplay`.

## Impact
- **Linux only.** macOS (`afplay`) and Windows (Windows Media Player COM) both decode mp3 fine.
- Non-fatal — the CLI continues; only playback is skipped with a warning.
- `--tts-format wav --tts-play` works on Linux (routes to `aplay`).

## Repro
On Linux with PulseAudio installed:
```
neurolink generate "hello" --tts --tts-play           # default mp3 -> paplay fails to decode
neurolink generate "hello" --tts --tts-format wav --tts-play   # works
```

## Fix ideas
- Use a codec-agnostic player when available (`ffplay -nodisp -autoexit`, `mpv --no-video`, `cvlc`) and fall back to paplay/aplay only for WAV.
- Or transcode to WAV before handing to paplay/aplay.
- Or, minimally, document that Linux `--tts-play` requires `--tts-format wav` and make the warning accurate (distinguish "player not found" from "format not decodable by this player").

## Context
Found during a deep code review while merging #982. The closed duplicate #983 shared the same Linux limitation (plus a separate Windows no-`playState`-polling bug that #982 does *not* have). Refs #982, #983.

murdore:
## What & why

Fixes #1138 — `--tts-play` mp3 playback is broken on Linux.

`src/cli/utils/audioPlayer.ts` routed all non-wav audio to `paplay`, but PulseAudio's `paplay` (and ALSA's `aplay`) only decode **libsndfile/PCM** formats (WAV/FLAC/Ogg/AIFF) — they **cannot decode mp3/AAC**. Since `--tts-format` defaults to **mp3**, the default `neurolink generate "..." --tts --tts-play` fails playback on Linux, with a misleading *"Install PulseAudio (paplay) or ALSA (aplay)"* message (PulseAudio is present; the format is simply undecodable).

## Fix

- `playAudio` now builds an **ordered candidate list** per platform + format and tries each until one succeeds (a missing binary or a decode failure advances to the next).
- **Linux, compressed formats (mp3/ogg/opus):** lead with real decoders — `ffplay` (ffmpeg) → `mpv` → `mpg123` (mp3 only) → `cvlc` (VLC) — then `paplay`/`aplay` as last-resort fallbacks (which still handle wav).
- **Linux wav:** unchanged — `aplay` first, then `paplay`, then `ffplay`.
- **macOS / Windows:** unchanged (`afplay` decodes everything; PowerShell paths preserved).
- Error message is now **format-aware**: points at ffmpeg/mpv/mpg123/VLC or `--tts-format wav`, and lists what was tried.

## Testing / proof

- `getPlayerCandidates(file, format, platform)` is exported and **platform-injectable**, so the behavior is deterministically testable off-Linux.
- New bugfixes-suite test asserts: Linux mp3 leads with a real decoder (never `paplay`/`aplay`); `mpg123` is offered for mp3 but **not** opus; Linux wav still routes to `aplay` first; macOS uses `afplay`.
- `pnpm run check` ✅ · `pnpm run lint` ✅ (0 errors) · bugfixes suite **90/90 PASS** · `pnpm run build:cli` ✅.

## Notes

- CLI-only change; no SDK API surface touched. The helper type moved to `src/lib/types/cli.ts` as `CliAudioPlayerCommand` to satisfy the types-location lint rule.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Linux TTS playback for MP3, OGG, and Opus audio by trying compatible decoders in sequence.
  * Added clearer playback errors that identify attempted players and explain when required decoders are unavailable.
  * Preserved reliable WAV playback through standard Linux audio systems.

* **Documentation**
  * Added platform-specific TTS playback guidance for macOS and Linux.
  * Clarified decoder requirements and recommended WAV as a zero-dependency Linux fallback.

### DIFF
--- a/docs/features/tts.md
+++ b/docs/features/tts.md
@@ -309,11 +309,15 @@ tts: {
 - Auto-converts to WAV when `play: true` on Windows
 - Use MP3 for file output, WAV for immediate playback
 
-**macOS/Linux:**
+**macOS:**
 
-- All formats supported
-- `afplay` (macOS) and `ffplay` (Linux) handle all formats
-- Use MP3 for general purpose
+- `afplay` (built-in) decodes every format — no setup needed.
+
+**Linux:**
+
+- **WAV** requires ALSA (`aplay`) or PulseAudio (`paplay`), but no compressed-format decoder.
+- **Compressed formats (mp3/ogg/opus)** need a real decoder — NeuroLink tries `ffplay` (ffmpeg), then `mpv`, `mpg123` (mp3), then `cvlc` (VLC), in that order. Install any one of them.
+- `paplay`/`aplay` **cannot** decode mp3, so with none of the above installed a default `--tts-play` (which defaults to mp3) will report a clear error naming the decoders — or use `--tts-format wav` when `aplay` or `paplay` is available.
 
 ---
 
--- a/src/cli/utils/audioPlayer.ts
+++ b/src/cli/utils/audioPlayer.ts
@@ -8,14 +8,33 @@
  */
 
 import { execFile } from "node:child_process";
+import { randomUUID } from "node:crypto";
 import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
 import { promisify } from "node:util";
 
-import type { AudioFormat } from "../../lib/types/index.js";
+import type {
+  TTSAudioFormat,
+  CliAudioPlayerCommand,
+} from "../../lib/types/index.js";
 
 const execFileAsync = promisify(execFile);
+
+/** Wall-clock cap on a single player invocation; a hung decoder must not block the CLI forever. */
+const PLAYER_TIMEOUT_MS = 30_000;
+
+/**
+ * Single source of truth for format -> temp-file-extension. Formats outside this
+ * playback-supported set (m4a, flac, webm, mp4, mpeg, mpga, pcm16) fall back to
+ * "mp3" — `playAudio` only ever receives the four formats below in practice.
+ */
+const AUDIO_EXTENSIONS: Partial<Record<TTSAudioFormat, string>> = {
+  mp3: "mp3",
+  wav: "wav",
+  ogg: "ogg",
+  opus: "opus",
+};
 
 /**
  * Get the file extension for an audio format
@@ -23,85 +42,165 @@ const execFileAsync = promisify(execFile);
  * @param format - Audio format
  * @returns File extension string (e.g., "mp3", "wav")
  */
-export function getAudioExtension(format: AudioFormat): string {
-  switch (format) {
-    case "mp3":
-      return "mp3";
-    case "wav":
-      return "wav";
-    case "ogg":
-      return "ogg";
-    case "opus":
-      return "opus";
-    default:
-      return "mp3";
-  }
+export function getAudioExtension(format: TTSAudioFormat): string {
+  return AUDIO_EXTENSIONS[format] ?? "mp3";
+}
+
+/** ffmpeg's ffplay decodes every format we emit; quiet + auto-exit for CLI use. */
+function ffplayCommand(filePath: string): CliAudioPlayerCommand {
+  return {
+    command: "ffplay",
+    args: ["-nodisp", "-autoexit", "-loglevel", "quiet", filePath],
+  };
 }
 
 /**
- * Get the platform-specific audio player command and arguments
+ * Escape a value for interpolation inside a single-quoted PowerShell string.
+ * PowerShell's single-quoted strings only treat `''` as a literal `'` — unlike
+ * shell/execFile argv, this string is parsed by powershell.exe itself, so an
+ * unescaped quote (e.g. a Windows username like `O'Brien` under `%TEMP%`)
+ * would break out of the string and let the remainder run as PS code.
+ */
+export function escapePowerShellSingleQuoted(value: string): string {
+  return value.replace(/'/g, "''");
+}
+
+/**
+ * Build the ordered list of player commands to try for a file.
+ *
+ * The list is tried in order until one succeeds; a missing binary (ENOENT) or a
+ * decode failure simply advances to the next candidate.
+ *
+ * The Linux ordering is the crux of the mp3-playback fix (#1138): `paplay`
+ * (PulseAudio) and `aplay` (ALSA) only decode libsndfile/PCM formats and
+ * **cannot** decode mp3/AAC. Since `--tts-format` defaults to mp3, compressed
+ * formats must lead with real decoders (ffplay/mpv/mpg123/cvlc); paplay/aplay
+ * remain only as last-resort fallbacks (and still work for wav/ogg).
  *
  * @param filePath - Path to the audio file
  * @param format - Audio format
- * @returns Object with command and args for execFile
+ * @param platform - Platform id (defaults to `process.platform`; injectable for tests)
+ * @returns Ordered player commands; empty for unsupported platforms
  */
-function getPlayerCommand(
+export function getPlayerCandidates(
   filePath: string,
-  format: AudioFormat,
-): { command: string; args: string[] } {
-  const platform = process.platform;
+  format: TTSAudioFormat,
+  platform: NodeJS.Platform = process.platform,
+): CliAudioPlayerCommand[] {
+  const isWav = format === "wav";
 
   switch (platform) {
     case "darwin":
-      return { command: "afplay", args: [filePath] };
+      // afplay decodes mp3/wav/aac/flac natively.
+      return [{ command: "afplay", args: [filePath] }];
 
     case "linux":
-      if (format === "wav") {
-        return { command: "aplay", args: [filePath] };
+      if (isWav) {
+        // WAV/PCM: ALSA and PulseAudio both decode it; ffplay as a backstop.
+        return [
+          { command: "aplay", args: [filePath] },
+          { command: "paplay", args: [filePath] },
+          ffplayCommand(filePath),
+        ];
       }
-      return { command: "paplay", args: [filePath] };
+      // Compressed formats (mp3/ogg/opus): lead with decoders that actually
+      // handle them, then fall back to paplay/aplay (which work for some
+      // formats but never mp3).
+      return [
+        ffplayCommand(filePath),
+        {
+          command: "mpv",
+          args: ["--no-video", "--really-quiet", filePath],
+        },
+        ...(format === "mp3"
+          ? [{ command: "mpg123", args: ["-q", filePath] }]
+          : []),
+        {
+          command: "cvlc",
+          args: ["--play-and-exit", "--no-loop", "--intf", "dummy", filePath],
+        },
+        { command: "paplay", args: [filePath] },
+        { command: "aplay", args: [filePath] },
+      ];
 
-    case "win32":
-      if (format === "wav") {
-        return {
+    case "win32": {
+      // filePath is embedded inside a single-quoted PS string below, so it must
+      // be escaped — a Windows username with an apostrophe (e.g. `O'Brien`,
+      // present in `%TEMP%`) would otherwise break out of the string.
+      const psPath = escapePowerShellSingleQuoted(filePath);
+      if (isWav) {
+        return [
+          {
+            command: "powershell",
+            args: [
+              "-NoProfile",
+              "-Command",
+              `(New-Object System.Media.SoundPlayer '${psPath}').PlaySync()`,
+            ],
+          },
+        ];
+      }
+      return [
+        {
           command: "powershell",
           args: [
             "-NoProfile",
             "-Command",
-            `(New-Object System.Media.SoundPlayer '${filePath}').PlaySync()`,
+            `$player = New-Object -ComObject WMPlayer.OCX; $player.URL = '${psPath}'; $player.controls.play(); Start-Sleep -Seconds 1; while ($player.playState -eq 3) { Start-Sleep -Milliseconds 100 }; $player.close()`,
           ],
-        };
-      }
-      return {
-        command: "powershell",
-        args: [
-          "-NoProfile",
-          "-Command",
-          `$player = New-Object -ComObject WMPlayer.OCX; $player.URL = '${filePath}'; $player.controls.play(); Start-Sleep -Seconds 1; while ($player.playState -eq 3) { Start-Sleep -Milliseconds 100 }; $player.close()`,
-        ],
-      };
+        },
+      ];
+    }
 
     default:
-      throw new Error(
-        `Unsupported platform: ${platform}. Audio playback is supported on macOS, Linux, and Windows.`,
-      );
+      return [];
   }
+}
+
+/**
+ * Build a helpful, format-aware error message when no player could decode the file.
+ *
+ * Exported for testing: the message a user sees when Linux mp3 playback finds no
+ * decoder must name the real fix (ffmpeg/mpv/mpg123/VLC or `--tts-format wav`),
+ * not the misleading "install PulseAudio" that paplay-only routing produced (#1138).
+ */
+export function buildPlaybackErrorMessage(
+  platform: NodeJS.Platform,
+  format: TTSAudioFormat,
+  attempts: string[],
+): string {
+  const tried = attempts.length ? ` (tried: ${attempts.join("; ")})` : "";
+
+  if (platform === "linux") {
+    if (format === "wav") {
+      return `Could not play wav audio${tried}. Install ALSA (aplay) or PulseAudio (paplay).`;
+    }
+    // mpg123 only decodes mp3 (see getPlayerCandidates) — recommending it for
+    // ogg/opus failures would send users chasing a decoder that can't help.
+    if (format === "mp3") {
+      return `Could not play mp3 audio${tried}. paplay/aplay cannot decode mp3; install one of: ffmpeg (ffplay), mpv, mpg123, or VLC (cvlc) — or use --tts-format wav.`;
+    }
+    return `Could not play ${format} audio${tried}. Install ffmpeg (ffplay), mpv, or VLC (cvlc) — or use --tts-format wav.`;
+  }
+
+  return `Could not play ${format} audio${tried}. Ensure a system audio player is installed and available in PATH.`;
 }
 
 /**
  * Play audio from a buffer using platform-specific CLI tools
  *
- * Writes the buffer to a temporary file, plays it using the appropriate
- * system audio player, and cleans up the temp file afterward.
+ * Writes the buffer to a temporary file, plays it using the first available
+ * system audio player that can decode the format, and cleans up the temp file.
  *
  * Supported platforms:
- * - macOS: uses `afplay` (built-in, supports mp3/wav/aac/flac)
- * - Linux: uses `paplay` for non-wav, `aplay` for wav
- * - Windows: uses PowerShell SoundPlayer (wav) or WMPlayer.OCX (mp3)
+ * - macOS: `afplay` (built-in; decodes mp3/wav/aac/flac)
+ * - Linux: real decoders (ffplay/mpv/mpg123/cvlc) for compressed formats,
+ *   aplay/paplay for wav — falling through the list until one works
+ * - Windows: PowerShell SoundPlayer (wav) or WMPlayer.OCX (compressed)
  *
  * @param buffer - Audio data buffer
  * @param format - Audio format (mp3, wav, ogg, opus)
- * @throws Error if playback fails or platform is unsupported
+ * @throws Error if no available player can play the audio, or platform is unsupported
  *
  * @example
  * ```typescript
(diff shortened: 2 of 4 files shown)


src/cli/utils/audioPlayer.ts
```diff
@@ -13,7 +13,10 @@ import os from "node:os";
 import path from "node:path";
```
Tara-ag, line 14: ⚠️ **MAJOR: `AudioFormat` is deprecated; use `TTSAudioFormat` for new CLI code.**

Per `src/lib/types/tts.ts`, `AudioFormat` is a backward-compat alias marked `@deprecated Use TTSAudioFormat instead.`. This PR introduces a new `CliAudioPlayerCommand` type and uses the deprecated alias throughout `audioPlayer.ts`. That creates a new internal consumer of a deprecated SDK type and will generate deprecation warnings once the linter flags internal usage.

**Suggested fix:**
```ts
import type {
  TTSAudioFormat,
  CliAudioPlayerCommand,
} from "../../lib/types/index.js";
```
and replace every `AudioFormat` occurrence in this file with `TTSAudioFormat`.

src/cli/utils/audioPlayer.ts
```diff
@@ -38,70 +41,136 @@ export function getAudioExtension(format: AudioFormat): string {
...
+ * Build the ordered list of player commands to try for a file.
+ *
+ * The list is tried in order until one succeeds; a missing binary (ENOENT) or a
+ * decode failure simply advances to the next candidate.
+ *
+ * The Linux ordering is the crux of the mp3-playback fix (#1138): `paplay`
+ * (PulseAudio) and `aplay` (ALSA) only decode libsndfile/PCM formats and
+ * **cannot** decode mp3/AAC. Since `--tts-format` defaults to mp3, compressed
+ * formats must lead with real decoders (ffplay/mpv/mpg123/cvlc); paplay/aplay
+ * remain only as last-resort fallbacks (and still work for wav/ogg).
```
Tara-ag, line 78: 💡 **MINOR: `getAudioExtension` duplicates runtime format-to-extension mapping.**

`src/lib/types/tts.ts` already ships `VALID_AUDIO_FORMATS` and the TTS ecosystem uses `TTSAudioFormat`. Consider whether this helper is needed at all — the extension is almost always identical to the format string for the supported set (`mp3`, `wav`, `ogg`, `opus`). If kept, a simple lookup table would be shorter and avoid the unnecessary `default: return "mp3";` fallback for an already-validated format.

**Suggested simplification:**
```ts
const EXTENSION_MAP: Record<TTSAudioFormat, string> = {
  mp3: "mp3", wav: "wav", ogg: "ogg", opus: "opus",
  m4a: "m4a", flac: "flac", webm: "webm", mp4: "mp4",
  mpeg: "mpeg", mpga: "mpga", pcm16: "pcm16",
};
```

src/cli/utils/audioPlayer.ts
```diff
@@ -38,70 +41,136 @@ export function getAudioExtension(format: AudioFormat): string {
...
  *
  * Supported platforms:
- * - macOS: uses `afplay` (built-in, supports mp3/wav/aac/flac)
- * - Linux: uses `paplay` for non-wav, `aplay` for wav
- * - Windows: uses PowerShell SoundPlayer (wav) or WMPlayer.OCX (mp3)
+ * - macOS: `afplay` (built-in; decodes mp3/wav/aac/flac)
+ * - Linux: real decoders (ffplay/mpv/mpg123/cvlc) for compressed formats,
+ *   aplay/paplay for wav — falling through the list until one works
+ * - Windows: PowerShell SoundPlayer (wav) or WMPlayer.OCX (compressed)
  *
```
Tara-ag, line 200: 🔒 **CRITICAL: Shell command injection via unescaped `filePath` in `ffplay`, `mpv`, `mpg123`, `cvlc`, `paplay`, `aplay`, `afplay`, and PowerShell arguments.**

`filePath` comes from `path.join(os.tmpdir(), \`nl-tts-${Date.now()}.${ext}\`)`, so under normal execution it is controlled. However, the function accepts any `buffer` and derives the temp path from the current time and the user-supplied `format`. More importantly, `getPlayerCandidates` is exported and takes an arbitrary `filePath` argument that callers/tests can inject. The returned commands are passed straight to `execFileAsync(command, args)`, which does **not** invoke a shell, but several of the invoked players themselves interpret shell metacharacters in their arguments (e.g. `ffplay '-nodisp'` is fine, but `cvlc`, `mpv`, `powershell` may expand `$var`, backticks, quotes, etc.).

The concrete risk is path traversal / command substitution if a caller ever passes a `filePath` containing spaces, quotes, or shell substitutions. Because `getPlayerCandidates` is public and platform-injectable, it should treat `filePath` as untrusted.

**Suggested fix:** validate/sanitize `filePath` before building commands. At minimum assert it is an absolute path under `os.tmpdir()` and reject characters outside a safe set:
```ts
const SAFE_PATH = /^[\w\-.\/]+$/;
if (!SAFE_PATH.test(filePath) || !path.isAbsolute(filePath)) {
  throw new Error("Invalid audio file path");
}
```
For PowerShell specifically, single-quote the path and escape embedded single quotes (`''`) rather than using raw interpolation.

src/cli/utils/audioPlayer.ts
```diff
@@ -119,40 +188,32 @@ export async function playAudio(
...
       }
-
-      throw execError;
     }
+
+    throw new Error(
+      buildPlaybackErrorMessage(process.platform, format, attempts),
+    );
   } finally {
     // Always clean up temp file
```
Tara-ag, line 266: 💡 **MINOR: Temp filename is predictable and can collide under rapid consecutive calls.**

`path.join(os.tmpdir(), \`nl-tts-${Date.now()}.${ext}\`)` uses only a millisecond timestamp. If two `playAudio` calls occur in the same millisecond (e.g. in a tight loop or concurrent requests), they will write to the same temp file and race on `unlink`.

**Suggested fix:** use a cryptographically random suffix:
```ts
import { randomUUID } from "node:crypto";
const tempFile = path.join(os.tmpdir(), `nl-tts-${randomUUID()}.${ext}`);
```

src/cli/utils/audioPlayer.ts
Tara-ag, line 1: ⚠️ **MAJOR: Non-ENOENT failures are swallowed silently and retried, masking real errors.**

The retry loop in `playAudio` treats every non-`ENOENT` error as “try next candidate”. That is correct for missing binaries, but it also hides genuine failures such as permission denied, segfaults, or decode errors that the user would want to know about. Worse, the final error message only shows the aggregate attempts, so earlier meaningful failures can be buried.

**Suggested fix:** distinguish recoverable vs. unrecoverable exit codes. At minimum, consider short-circuiting on signals/non-zero exits that indicate the binary is present but failing (e.g. exit code 1 from `ffplay` for an unreadable file), or surface the first non-ENOENT failure more prominently.

src/lib/types/cli.ts
Tara-ag, line 1: 💡 **MINOR: New `CliAudioPlayerCommand` type is correctly placed in `src/lib/types/cli.ts` and follows the `Cli*` prefix.**

No issue here — this satisfies CLAUDE.md Rule 2 (types in canonical location) and Rule 9 (domain prefix). Good.

test/continuous-test-suite-bugfixes.ts
Tara-ag, line 1: 💡 **MINOR: Test covers the key ordering invariants for Linux mp3/opus/wav and macOS.**

The new `audioPlayer #1138` test is deterministic and platform-injectable, which is the right approach for a CLI-only utility. Consider also asserting that the **first** Linux mp3 candidate is `ffplay` specifically, since the current test only checks it is not `paplay`/`aplay` and precedes them. That would lock the intended priority more tightly.

Tara-ag (changes requested):
## Review summary

**Files reviewed:** 3 (`src/cli/utils/audioPlayer.ts`, `src/lib/types/cli.ts`, `test/continuous-test-suite-bugfixes.ts`)

**New issues raised:**
- 🔒 **1 CRITICAL** — potential shell injection through unescaped `filePath` passed to media players / PowerShell.
- ⚠️ **2 MAJOR** — internal use of the deprecated `AudioFormat` alias instead of `TTSAudioFormat`; non-ENOENT player failures are silently swallowed and retried.
- 💡 **3 MINOR/SUGGESTION** — `getAudioExtension` duplicates existing format mapping; temp filename is predictable and can collide; test could tighten the first-candidate assertion.

**Decision: REQUEST CHANGES**

The CRITICAL shell-injection risk in `getPlayerCandidates` / `playAudio` blocks the PR. Even though `filePath` is normally generated internally, `getPlayerCandidates` is exported and platform-injectable, and several invoked players interpret shell metacharacters. Please sanitize/validate `filePath` before building command arguments, and properly escape the path for the PowerShell invocation.

Also please address the MAJOR issues:
1. Replace `AudioFormat` with `TTSAudioFormat` in `audioPlayer.ts` (`AudioFormat` is a deprecated backward-compat alias in `src/lib/types/tts.ts`).
2. Distinguish recoverable (missing binary) from unrecoverable (permission denied, decode failure, signal) player errors instead of unconditionally retrying every non-ENOENT failure.

Once these are fixed, the remaining MINOR items can be addressed at your discretion.

murdore:
## ✅ Verification & gap-bridging (hardened)

Rebased onto current `release` and hardened across test / docs / proof gaps.

### Test gap → closed
- **Unit (routing):** `getPlayerCandidates(file, format, platform)` — asserts Linux mp3 leads with real decoders (ffplay/mpv/mpg123/cvlc), never `paplay`/`aplay`; `mpg123` is offered for mp3 but **not** opus; Linux wav routes to `aplay` first; macOS uses `afplay`.
- **User-facing error (new):** `buildPlaybackErrorMessage` is now exported and tested — the message a Linux user sees with no decoder installed names **ffmpeg/mpv/mpg123/VLC** and the `--tts-format wav` fallback, and explains that `paplay`/`aplay` cannot decode mp3 — *not* the old misleading "install PulseAudio" (PulseAudio is present; it simply can't decode mp3).
- Both tests are **platform-injectable**, so the Linux behavior is verified deterministically off-Linux.

### Why not a full end-to-end `--tts-play` test
Real playback needs both a provider API key (to synthesize) **and** audio hardware/a decoder on the runner, so a true E2E `neurolink generate … --tts --tts-play` assertion isn't hermetic. The two tests above cover the exact decision (`which player, in what order`) and the exact user-facing output (`the error text`) — the two things the fix changes — without those dependencies.

### Docs gap → closed
`docs/features/tts.md` "Platform-Specific Considerations" previously claimed *"All formats supported / ffplay (Linux) handle all formats"* — inaccurate, since the old code routed mp3 to `paplay`. Now split into explicit macOS/Linux guidance stating that Linux compressed-format playback needs a decoder (ffmpeg/mpv/mpg123/VLC) or `--tts-format wav`.

### Enhancement gap
None outstanding — the decoder chain + actionable error is the complete fix. (No new dependency added; playback stays best-effort and non-fatal.)

### Proof
`pnpm run check` ✅ (0 errors) · `pnpm run lint` ✅ (0 errors) · `prettier --check` ✅ · both `#1138` bugfixes-suite tests **green**. The two failing `updater:` tests in a local run are pre-existing environmental cases in the upstream suite (present on `release`, and `test (20)` was green on this PR) — unrelated to this change.

src/cli/utils/audioPlayer.ts
```diff
@@ -119,40 +192,32 @@ export async function playAudio(
...
+      );
+    }
+
+    const attempts: string[] = [];
+
+    for (const { command, args } of candidates) {
+      try {
+        await execFileAsync(command, args);
+        return; // Played successfully.
+      } catch (execError) {
```
Tara-ag, line 243: ⚠️ **MAJOR: External player process has no timeout — a hung decoder can block the CLI indefinitely.**

`execFileAsync(command, args)` waits for the player to exit but never sets a wall-clock limit. If `ffplay`, `mpv`, `cvlc`, etc. hang (e.g. waiting on an audio device or stuck on a corrupt buffer), `playAudio` will never return and the temp file will never be cleaned up.

CLAUDE.md Common Patterns: "Wrap async calls with `withTimeout` utility". For spawned child processes, also use `execFile`'s own `timeout` option so the OS kills the process:

```ts
await execFileAsync(command, args, {
  timeout: PLAYER_TIMEOUT_MS,
  killSignal: "SIGTERM",
});
```

Consider wrapping with `withTimeout(..., PLAYER_TIMEOUT_MS)` as a secondary guard and surfacing a clear "audio playback timed out" error.

src/cli/utils/audioPlayer.ts
```diff
@@ -13,7 +13,10 @@ import os from "node:os";
 import path from "node:path";
 import { promisify } from "node:util";
 
-import type { AudioFormat } from "../../lib/types/index.js";
+import type {
```
Tara-ag, line 17: ⚠️ **MAJOR: Continues using deprecated `AudioFormat` alias for newly-introduced CLI internals.**

Per `src/lib/types/tts.ts`, `AudioFormat` is explicitly marked `@deprecated Use TTSAudioFormat instead.` and exists only for backward-compatible SDK consumers. This PR adds a brand-new `CliAudioPlayerCommand` type and exports `getPlayerCandidates` / `buildPlaybackErrorMessage`, yet it types all of them with the deprecated alias. That creates a new internal consumer of a deprecated SDK type and will produce deprecation warnings once internal usage is flagged.

**Suggested fix:**
```ts
import type {
  TTSAudioFormat,
  CliAudioPlayerCommand,
} from "../../lib/types/index.js";
```
and replace every `AudioFormat` occurrence in this file with `TTSAudioFormat`.

src/cli/utils/audioPlayer.ts
```diff
@@ -38,70 +41,140 @@ export function getAudioExtension(format: AudioFormat): string {
...
+        // WAV/PCM: ALSA and PulseAudio both decode it; ffplay as a backstop.
+        return [
+          { command: "aplay", args: [filePath] },
+          { command: "paplay", args: [filePath] },
+          ffplayCommand(filePath),
+        ];
       }
-      return { command: "paplay", args: [filePath] };
+      // Compressed formats (mp3/ogg/opus): lead with decoders that actually
+      // handle them, then fall back to paplay/aplay (which work for some
```
Tara-ag, line 107: 🔒 **CRITICAL: Shell command injection / unsafe path handling via exported `getPlayerCandidates`.**

`getPlayerCandidates` is now exported and platform-injectable, and its `filePath` argument is interpolated directly into player arguments and PowerShell command strings. While `playAudio` currently builds `tempFile` under `os.tmpdir()`, the exported helper accepts arbitrary paths that callers or tests can pass. Several invocations are dangerous:

- `ffplay`, `mpv`, `mpg123`, `cvlc`, `paplay`, `aplay`, `afplay` receive `filePath` as a bare argument to programs that may themselves interpret shell metacharacters.
- The Windows PowerShell snippets embed `filePath` inside a single-quoted string without escaping embedded single quotes, so a path like `C:\Users\O'Brien\tmp.mp3` breaks the command or allows injection.

Because this helper is public and test-injectable, it should treat `filePath` as untrusted.

**Suggested fix:** validate/sanitize `filePath` before building commands. At minimum assert it is an absolute path and reject characters outside a safe set:
```ts
const SAFE_PATH = /^[\w\-.\/]+$/;
if (!SAFE_PATH.test(filePath) || !path.isAbsolute(filePath)) {
  throw new Error("Invalid audio file path");
}
```
For PowerShell specifically, single-quote the path and escape embedded single quotes (`''`) rather than using raw interpolation.

Tara-ag (changes requested):
## Review summary

**Files reviewed:** 4 (`docs/features/tts.md`, `src/cli/utils/audioPlayer.ts`, `src/lib/types/cli.ts`, `test/continuous-test-suite-bugfixes.ts`)

- 🔒 **1 CRITICAL** — potential shell injection through unescaped `filePath` passed to media players / PowerShell in the exported `getPlayerCandidates` helper.
- ⚠️ **2 MAJOR** — internal use of the deprecated `AudioFormat` alias instead of `TTSAudioFormat`; external player processes have no timeout, so a hung decoder can block the CLI indefinitely.
- 💡 **1 MINOR** — temp filename uses only `Date.now()` and can collide under rapid/concurrent calls.

**Existing review comments already covered:** I did not duplicate points already raised by prior reviewers (deprecated `AudioFormat`, temp-file predictability, swallowed non-ENOENT errors, test coverage gaps, docs wording). The issues above are the ones I found independently or that remain unaddressed.

**Decision: REQUEST CHANGES**

The CRITICAL shell-injection risk in `getPlayerCandidates` / `playAudio` blocks the PR. Even though `filePath` is normally generated internally, `getPlayerCandidates` is exported and platform-injectable, and several invoked players interpret shell metacharacters. Please sanitize/validate `filePath` before building command arguments, and properly escape the path for the PowerShell invocation.

Also please address the MAJOR issues:
1. Replace `AudioFormat` with `TTSAudioFormat` in `audioPlayer.ts` (`AudioFormat` is a deprecated backward-compat alias in `src/lib/types/tts.ts`).
2. Add a timeout to `execFileAsync` (e.g. `timeout: 30_000`) so a hung `ffplay`/`mpv`/`cvlc` cannot block the CLI forever.

Once these are fixed, the remaining MINOR item can be addressed at your discretion.

src/cli/utils/audioPlayer.ts
murdore: Fixed in 05d60f96. `src/cli/utils/audioPlayer.ts` now imports `TTSAudioFormat` (not the deprecated `AudioFormat` alias) and every occurrence — `getAudioExtension`, `getPlayerCandidates`, `buildPlaybackErrorMessage`, `playAudio` — is typed with `TTSAudioFormat` (lines 17-18, 45, 87, 169, 212). No remaining references to the deprecated alias in this file.

src/cli/utils/audioPlayer.ts
murdore: Fixed in 05d60f96. `getAudioExtension` (lines 32-46) is now a lookup into a single `AUDIO_EXTENSIONS: Partial<Record<TTSAudioFormat, string>>` map instead of a switch — one source of truth, no per-branch duplication, same "mp3" fallback behavior preserved for formats outside the CLI-playback-supported set.

src/cli/utils/audioPlayer.ts
murdore: Verified against current code and partially fixed in 05d60f96.

The general shell-injection claim for ffplay/mpv/mpg123/cvlc/paplay/aplay/afplay doesn't hold: `execFileAsync(command, args)` (Node's `execFile`, not `exec`) never spawns a shell, so argv elements are passed literally to `execve` — no shell metacharacter expansion (`$var`, backticks, etc.) is possible for those invocations regardless of `filePath` content. That part is a false positive.

The real, actionable risk was the Windows PowerShell branch: `filePath` was interpolated into a single-quoted PS string that `powershell.exe` itself parses, so an embedded `'` (e.g. from a `%TEMP%` path under a Windows username like `O'Brien`) could break out of the string. Fixed: added `escapePowerShellSingleQuoted()` (lines 57-66), which doubles embedded `'` → `''` per PowerShell single-quoted-string escaping rules, applied to both the `SoundPlayer` and `WMPlayer.OCX` command strings (lines 126-153). Covered by a new deterministic test ("PowerShell single-quote escaping prevents command injection") in `test/continuous-test-suite-bugfixes.ts`.

src/cli/utils/audioPlayer.ts
murdore: Fixed in 05d60f96. `playAudio` now builds the temp path with `randomUUID()` instead of `Date.now()` (line 217: `nl-tts-${randomUUID()}.${ext}`), eliminating the millisecond-collision window for concurrent calls.

src/cli/utils/audioPlayer.ts
murdore: Verified against current code — re-checked the retry loop and kept the design, with one clarification and one real improvement alongside it.

On "swallowed silently": every failure (ENOENT or not) is captured into `attempts` and surfaced verbatim in the final thrown error via `buildPlaybackErrorMessage`'s `tried:` list (see lines 243-256) — nothing is dropped. So the failure mode described (errors disappearing) doesn't reproduce against current code.

On "should fail fast instead of trying every candidate": deliberately did not change this. `playAudio`'s whole reason for existing (#1138) is that a *specific* decoder can fail (or not exist) for reasons that are decoder-specific — e.g. paplay/aplay simply cannot decode mp3 at all — while the next candidate in the list (ffplay, mpv, mpg123, cvlc) may well succeed. Aborting on the first non-ENOENT failure would silently defeat that fallback for any transient or single-decoder-specific error and regress the core behavior this PR ships. I added a comment (lines 247-252) documenting this reasoning in the code itself so it isn't re-litigated as an oversight later.

What *did* change here: `execFileAsync` now carries a 30s timeout (`PLAYER_TIMEOUT_MS`) so a hung decoder can't block indefinitely, and a timeout is now reported distinctly in `attempts` as "timed out after 30000ms" rather than folded into a generic failure message — giving the "which failure is real" visibility this thread was ultimately asking for.

src/lib/types/cli.ts
murdore: Acknowledged — still true on current code. `CliAudioPlayerCommand` remains in `src/lib/types/cli.ts` (now lines 1948-1956) with the `Cli*` prefix; no change needed. Thanks for confirming.

test/continuous-test-suite-bugfixes.ts
murdore: Fixed in 05d60f96. The Linux-mp3 test now asserts `commands[0] === "ffplay"` explicitly (not just "not paplay/aplay, and a decoder precedes them"), locking the intended priority as suggested.

docs/features/tts.md
```diff
@@ -309,11 +309,15 @@ tts: {
...
+- `afplay` (built-in) decodes every format — no setup needed.
+
+**Linux:**
 
-- All formats supported
-- `afplay` (macOS) and `ffplay` (Linux) handle all formats
-- Use MP3 for general purpose
+- **WAV** plays out of the box via ALSA (`aplay`) or PulseAudio (`paplay`).
+- **Compressed formats (mp3/ogg/opus)** need a real decoder — NeuroLink tries `ffplay` (ffmpeg), then `mpv`, `mpg123` (mp3), then `cvlc` (VLC), in that order. Install any one of them.
+- `paplay`/`aplay` **cannot** decode mp3, so with none of the above installed a default `--tts-play` (which defaults to mp3) will report a clear error naming the decoders — or use `--tts-format wav` for zero-dependency playback.
```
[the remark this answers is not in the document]
murdore: Fixed in 05d60f96, applying your suggested wording: `docs/features/tts.md` now reads "WAV requires ALSA (`aplay`) or PulseAudio (`paplay`), but no compressed-format decoder" and "...or use `--tts-format wav` when `aplay` or `paplay` is available" — no longer implies WAV playback is dependency-free.

src/cli/utils/audioPlayer.ts
```diff
@@ -38,70 +41,140 @@ export function getAudioExtension(format: AudioFormat): string {
...
+  format: AudioFormat,
+  attempts: string[],
+): string {
+  const tried = attempts.length ? ` (tried: ${attempts.join("; ")})` : "";
+
+  if (platform === "linux") {
+    if (format === "wav") {
+      return `Could not play wav audio${tried}. Install ALSA (aplay) or PulseAudio (paplay).`;
+    }
+    return `Could not play ${format} audio${tried}. paplay/aplay cannot decode ${format}; install one of: ffmpeg (ffplay), mpv, mpg123, or VLC (cvlc) — or use --tts-format wav.`;
```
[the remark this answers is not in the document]
murdore: Fixed in 05d60f96, following your proposed diff. `buildPlaybackErrorMessage`'s Linux branch (lines 174-186) now has a dedicated `format === "mp3"` case that keeps the `mpg123`/"cannot decode" wording, and a separate ogg/opus fallback that only recommends `ffplay`/`mpv`/`cvlc`. Covered by a new test ("ogg/opus playback error does not recommend mp3-only mpg123") in `test/continuous-test-suite-bugfixes.ts`.

src/cli/utils/audioPlayer.ts
murdore: Fixed in 05d60f96. `execFileAsync` in the candidate loop now passes `{ timeout: PLAYER_TIMEOUT_MS, killSignal: "SIGTERM" }` (30s cap, lines 236-241), so a hung player is killed rather than blocking `playAudio` forever; the temp-file cleanup in `finally` still runs since the promise now settles. A timed-out attempt is reported distinctly as "timed out after 30000ms" in the aggregated error. Covered by a static regression check in `test/continuous-test-suite-bugfixes.ts`.

src/cli/utils/audioPlayer.ts
murdore: Fixed in 05d60f96 — same as the sibling `AudioFormat` thread on this line/file; duplicate flagging from the two review passes. `getAudioExtension` and every other signature in `audioPlayer.ts` now use `TTSAudioFormat`.

src/cli/utils/audioPlayer.ts
murdore: Same disposition as the other CRITICAL shell-injection thread on this file (duplicate flag from the two review passes): the non-PS players are safe because `execFileAsync` never spawns a shell (argv passed literally to `execve`, no metacharacter expansion). The real risk — unescaped `filePath` inside the PowerShell single-quoted command string — is fixed via `escapePowerShellSingleQuoted()` (lines 57-66, applied at lines 126-153), doubling embedded `'` per PS escaping rules. See the sibling thread for the full writeup.

Tara-ag (approved):
## Review Summary

Reviewed all 4 changed files in PR #1180.

**Files examined:**
- `docs/features/tts.md`
- `src/cli/utils/audioPlayer.ts`
- `src/lib/types/cli.ts`

**New issues raised this run:** None.

**Existing review threads:** All prior review comments are resolved (including the `AudioFormat` → `TTSAudioFormat` migration, PowerShell single-quote escaping, temp-file collision avoidance, player timeout wiring, and format-aware error messages). No duplicates or re-raises.

**Assessment:** This is a focused CLI-only fix for Linux mp3 TTS playback. The implementation correctly routes compressed formats through real decoders before falling back to `paplay`/`aplay`, uses `execFile` (not a shell) for player invocation, escapes PowerShell single quotes, caps player execution with a timeout, and includes deterministic regression tests. No blocking security, architecture, or backward-compatibility concerns remain.

Approving.

Issue #1139, opened 2026-07-08
Skills subsystem (#1135): name-collision on soft-delete, cross-tenant scope leak, allowMutations doesn't gate REST routes


murdore:
## Summary
Tracking issue for correctness/security findings from a deep code review of the native skills subsystem in #1135. These are all in the **opt-in path** (nothing triggers unless `skills.enabled: true`), so there's no default impact — but they should be addressed before or shortly after the subsystem is enabled anywhere. Filing as an issue so they survive the PR's lifecycle. Refs #1135.

## 1. Name collision after soft-delete resolves non-deterministically (integrity)
`SkillsManager.get(idOrName)` (`src/lib/skills/skillsManager.ts`) falls back to a name lookup (`index.find(i => i.name === idOrName)`) with **no active-status filter**, and `assertNameAvailable()` only blocks name collisions among *active* skills. So soft-deleting a skill (status → `deprecated`) and creating a new one with the same name leaves two index entries sharing that name. A subsequent **name-based** lookup — CLI `skills show/delete <name>`, the `skill_update`/`skill_delete` tools when given a name, or the server `GET/PATCH/DELETE /api/agent/skills/:id` routes (documented as accepting "id or name") — returns whichever entry appears first in `index()` iteration order (non-deterministic across memory vs filesystem backends). An update/delete aimed at the new active skill can silently hit the stale deprecated one.

*Suggested fix:* filter `get()`'s name fallback to active-only (resolve deterministically to the active entry), and have `assertNameAvailable()` also reject reuse of a deprecated skill's name.

## 2. Scoped skills leak when no scopeId is supplied (multi-tenant)
`filterSkillIndex` (`src/lib/skills/skillMatcher.ts`) excludes a `scoped` skill only when the caller supplies a scopeId that *doesn't match*. When no scopeId is resolvable (no per-call `skills.scopeId`, no instance `defaultScopeId` — both optional), **every** scoped skill from every team/tenant is returned by `search_skills`/`list_skills` and injected into the system-prompt index. On a shared multi-tenant instance that relies on a per-call scopeId, forgetting to pass it leaks all tenants' scoped instructions to everyone. Contradicts the docs' scope description, and the 33-case suite only tests explicit scopeId-vs-scopeId exclusion (the no-scopeId path is untested).

*Suggested fix:* fail closed — when no scopeId is resolvable, exclude `scoped` skills.

## 3. `allowMutations` does not gate the REST routes (governance)
[...]

PR #1182, opened 2026-07-16
fix(skills): fail closed on scoped skills without a scopeId (#1139)

murdore:
## What & why

Addresses the remaining findings in #1139 — correctness/security items from the deep review of the native skills subsystem (#1135). All are on the **opt-in path** (`skills.enabled: true`), so there's no default impact.

Of the three parts + two nits in the issue, **Parts 1 & 3 and the get()-integrity concern are already in place** in the current code:
- Part 1 — `SkillsManager.get()` already resolves a name to the **active** skill first (deterministic after soft-delete + same-name recreate).
- Part 3 — the REST mutation routes already gate on `allowMutations` via `resolveMutableSkillsManager` (POST/PATCH/DELETE return `SKILLS_MUTATIONS_DISABLED` when false).

This PR fixes the two that were still open:

### Part 2 — scoped skills leaked when no scopeId supplied (multi-tenant) 🔒
`filterSkillIndex` only excluded a `scoped` skill when the caller passed a *non-matching* scopeId. With **no** resolvable scopeId (no per-call `skills.scopeId`, no instance `defaultScopeId`), every scoped skill from every tenant was returned by `search_skills` / `list_skills` and injected into the system-prompt index. On a shared multi-tenant instance, forgetting the scopeId leaked all tenants' scoped instructions to everyone.

**Now fails closed:** a scoped skill is admitted only when a matching scopeId is supplied; with no scopeId, scoped skills are excluded. Global skills are unaffected. All three consumers (`search`, `list`, prompt index) route through `filterSkillIndex`, so they're fixed together.

### Nit — `promptIndexMaxItems: 0` emitted a header-only block
`formatSkillsPromptIndex` returned the "## Available Skills" header + a "N more skills exist" note with **zero** skill lines when the cap was 0. Now returns `null` (skip injection entirely).

## Testing / proof

- New skills-suite cases: **8b** (scoped fail-closed for `search_skills` + `list_skills` with no scopeId; globals still visible) and **15b** (`promptIndexMaxItems=0` → no injection).
- Updated the four existing cases whose no-scopeId counts previously *included* the leaked scoped skill to pass an explicit scopeId (preserving each test's original intent). These updates are themselves proof the leak is closed.
- `pnpm run check` ✅ · `pnpm run lint` ✅ (0 errors) · skills suite **39/39** · `pnpm run build` ✅.

## Deferred

The last nit — adding `docs/features/skills.md` to `docs-site/sidebars.ts` — is a docs-navigation follow-up (needs a docs-site build to validate) and is intentionally out of this security-focused PR.

Refs #1135. Fixes the security/correctness parts of #1139.

## Summary by CodeRabbit

* **Bug Fixes**
  * Scoped skills are now fail-closed: they won’t appear in search or listings unless the matching `scopeId` is provided.
  * The skills prompt index is suppressed when the max item limit is invalid/disabled, preventing unintended “Available Skills” content.
  * Global (non-scoped) skills remain visible for unscoped searches and listings.
* **Tests**
  * Expanded continuous end-to-end coverage for scoped-skill visibility, prompting behavior, and safe activation/read failures without `scopeId`.
* **Documentation**
  * Clarified scoped-skill behavior and `scopeId` resolution rules.

### DIFF
--- a/src/lib/skills/skillMatcher.ts
+++ b/src/lib/skills/skillMatcher.ts
@@ -45,11 +45,14 @@ export function filterSkillIndex(
       return false;
     }
 
-    // Scope filter: global skills always pass; scoped skills require a
-    // matching scopeId. When the caller provides no scopeId, scoped skills
-    // still pass (curator semantics — unscoped callers see everything).
-    if (query.scopeId && item.scope === "scoped") {
-      if (!(item.scopeIds ?? []).includes(query.scopeId)) {
+    // Scope filter: global skills always pass. Scoped skills require a
+    // scopeId that matches; when the caller supplies no scopeId we fail
+    // closed and exclude them, so a shared multi-tenant instance that forgets
+    // to pass a per-call scopeId never leaks one tenant's scoped skills to
+    // everyone (#1139). Callers that legitimately want a tenant's scoped
+    // skills must pass that tenant's scopeId.
+    if (item.scope === "scoped") {
+      if (!query.scopeId || !(item.scopeIds ?? []).includes(query.scopeId)) {
         return false;
       }
     }
@@ -94,13 +97,22 @@ export function isSafeSkillResourcePath(resourcePath: string): boolean {
   );
 }
 
-/** Whether a skill is visible from the calling scope. */
+/**
+ * Whether a skill is visible from the calling scope. Fails closed, matching
+ * `filterSkillIndex`: a scoped skill is hidden unless the caller supplies a
+ * `scopeId` that the skill explicitly lists. Omitting `scopeId` therefore
+ * yields global skills only — it must never expose a scoped skill by name
+ * through `use_skill` / `read_skill_resource` (#1139).
+ */
 export function isSkillVisibleInScope(
   skill: Pick<SkillDefinition, "scope" | "scopeIds">,
   scopeId?: string,
 ): boolean {
-  if (!scopeId || skill.scope !== "scoped") {
+  if (skill.scope !== "scoped") {
     return true;
+  }
+  if (!scopeId) {
+    return false;
   }
   return (skill.scopeIds ?? []).includes(scopeId);
 }
@@ -171,7 +183,10 @@ export function formatSkillsPromptIndex(
   items: SkillIndexItem[],
   maxItems: number,
 ): string | null {
-  if (items.length === 0) {
+  // Nothing visible, or the index is explicitly disabled (maxItems <= 0):
+  // skip injection entirely rather than emit a header + "N more skills exist"
+  // note with zero skill lines (#1139).
+  if (items.length === 0 || maxItems <= 0) {
     return null;
   }
 
--- a/test/continuous-test-suite-skills.ts
+++ b/test/continuous-test-suite-skills.ts
@@ -312,6 +312,30 @@ async function testSearch(): Promise<void> {
     },
   );
 
+  await runTest(
+    "6b. #1139: scoped skills fail closed when no scopeId is supplied",
+    async () => {
+      const nl = makeSkillsInstance();
+      const mgr = nl.getSkillsManager()!;
+      // No scopeId (and no defaultScopeId): scoped skills must NOT leak.
+      const searched = await mgr.search({ query: "standup" });
+      assert(
+        searched.length === 0,
+        "scoped skill leaked to search with no scopeId",
+      );
+      const listed = await mgr.list();
+      assert(
+        !listed.some((s) => s.name === "team_alpha_standup"),
+        "scoped skill leaked to list with no scopeId",
+      );
+      // The two global seed skills stay visible without a scopeId.
+      assert(
+        listed.length === 2,
+        `expected 2 global skills, got ${listed.length}`,
+      );
+    },
+  );
+
   await runTest("7. maxMatches caps hydrated results", async () => {
     const many = Array.from({ length: 8 }, (_, i) => ({
       id: `bulk-${i}`,
@@ -338,7 +362,9 @@ async function testSearch(): Promise<void> {
     async () => {
       const nl = makeSkillsInstance();
       const execute = getToolExecute(nl, "list_skills");
-      const result = (await execute({})) as {
+      // Scope supplied so the scoped seed skill is included (fail-closed
+      // filtering hides it without a scopeId — #1139).
+      const result = (await execute({ scopeId: "team-alpha" })) as {
         success: boolean;
         data: { skills: Array<Record<string, unknown>>; count: number };
       };
@@ -599,20 +625,46 @@ async function testDiscovery(): Promise<void> {
         !useSkill.description.includes("payments-oncall"),
         "instructions leaked into the listing",
       );
-      // Sorted by name: refund… < service_deployment < team_alpha_standup
+      // Sorted by name: refund… < service_deployment. The scoped
+      // team_alpha_standup is NOT listed without a scopeId (#1139 fail-closed):
+      // the default discovery listing must not leak scoped skills.
       const listing = useSkill.description;
       assert(
         listing.indexOf("refund_dispute_escalation") <
-          listing.indexOf("service_deployment") &&
-          listing.indexOf("service_deployment") <
-            listing.indexOf("team_alpha_standup"),
+          listing.indexOf("service_deployment"),
         "listing must be name-sorted",
+      );
+      assert(
+        !listing.includes("team_alpha_standup"),
+        "scoped skill must not appear in the default (no-scopeId) discovery listing",
       );
       // Deterministic: a second augmentation renders byte-identical listing
       const again = await applyAugmentation(nl, { systemPrompt: "x" });
       assert(
         getCallTool(again, "use_skill").description === useSkill.description,
         "listing must be byte-stable across calls",
+      );
+    },
+  );
+
+  await runTest(
+    "15b. #1139: promptIndexMaxItems=0 suppresses the system-prompt index",
+    async () => {
+      const nl = makeSkillsInstance({
+        discovery: "system-prompt",
+        promptIndexMaxItems: 0,
+      });
+      const options = await applyAugmentation(nl, {
+        systemPrompt: "You are Tara.",
+      });
+      const prompt = options.systemPrompt as string;
+      assert(
+        prompt === "You are Tara.",
+        "promptIndexMaxItems=0 must not inject the skills index",
+      );
+      assert(
+        !prompt.includes("## Available Skills"),
+        "promptIndexMaxItems=0 must suppress the '## Available Skills' block",
       );
     },
   );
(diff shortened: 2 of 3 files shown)


src/lib/skills/skillMatcher.ts
```diff
@@ -45,11 +45,14 @@ export function filterSkillIndex(
...
-    // Scope filter: global skills always pass; scoped skills require a
-    // matching scopeId. When the caller provides no scopeId, scoped skills
-    // still pass (curator semantics — unscoped callers see everything).
-    if (query.scopeId && item.scope === "scoped") {
-      if (!(item.scopeIds ?? []).includes(query.scopeId)) {
+    // Scope filter: global skills always pass. Scoped skills require a
+    // scopeId that matches; when the caller supplies no scopeId we fail
+    // closed and exclude them, so a shared multi-tenant instance that forgets
+    // to pass a per-call scopeId never leaks one tenant's scoped skills to
+    // everyone (#1139). Callers that legitimately want a tenant's scoped
```
Tara-ag, line 52: ⚠️ **MAJOR: `isSkillVisibleInScope` is inconsistent with the new fail-closed semantics.**

`filterSkillIndex` now correctly fails closed (no `scopeId` → scoped skills excluded), but `isSkillVisibleInScope` still returns `true` for a scoped skill when `scopeId` is omitted:

```ts
export function isSkillVisibleInScope(skill, scopeId): boolean {
  if (!scopeId || skill.scope !== "scoped") return true;
  return (skill.scopeIds ?? []).includes(scopeId);
}
```

This means `use_skill` and `read_skill_resource` in `skillTools.ts` will still activate/read a scoped skill by name even when the call has no `scopeId`, bypassing the fail-closed fix. The PR description says all three consumers (`search`, `list`, prompt index) route through `filterSkillIndex`, but the per-call tools are a fourth consumer that does not.

**Suggested fix:** align `isSkillVisibleInScope` with `filterSkillIndex`:

```ts
export function isSkillVisibleInScope(
  skill: Pick<SkillDefinition, "scope" | "scopeIds">,
  scopeId?: string,
): boolean {
  if (skill.scope !== "scoped") return true;
  return Boolean(scopeId) && (skill.scopeIds ?? []).includes(scopeId);
}
```

Then add a test that `use_skill({ name: "team_alpha_standup" })` fails when no `scopeId` is supplied.

src/lib/skills/skillMatcher.ts
Tara-ag, line 1: 🔒 **CRITICAL: `isSkillVisibleInScope` is inconsistent with the new fail-closed semantics and leaves a bypass in `skillTools.ts`.**

`filterSkillIndex` now correctly excludes scoped skills when no `scopeId` is supplied, but `isSkillVisibleInScope` still returns `true` for scoped skills when `scopeId` is omitted. Consumers in `skillTools.ts` (`use_skill` and `read_skill_resource`) resolve a skill by name via `manager.get()` and then guard with `isSkillVisibleInScope(skill, context.scopeId)`. Because of this helper, a caller with no `scopeId` can still activate a scoped skill by exact name and read its full instructions/resources, bypassing the multi-tenant leak fix.

This is the same class of leak described in #1139 and should be closed in this security PR.

**Suggested fix:** align `isSkillVisibleInScope` with `filterSkillIndex`:

```ts
export function isSkillVisibleInScope(
  skill: Pick<SkillDefinition, "scope" | "scopeIds">,
  scopeId?: string,
): boolean {
  if (skill.scope !== "scoped") return true;
  return Boolean(scopeId) && (skill.scopeIds ?? []).includes(scopeId);
}
```

Then add regression tests that `use_skill({ name: "team_alpha_standup" })` and `read_skill_resource({ skill: "team_alpha_standup", path: "..." })` both return an error envelope when no `scopeId` is supplied.

test/continuous-test-suite-skills.ts
```diff
@@ -599,15 +625,19 @@ async function testDiscovery(): Promise<void> {
...
       assert(
         listing.indexOf("refund_dispute_escalation") <
-          listing.indexOf("service_deployment") &&
-          listing.indexOf("service_deployment") <
-            listing.indexOf("team_alpha_standup"),
+          listing.indexOf("service_deployment"),
         "listing must be name-sorted",
       );
+      assert(
+        !listing.includes("team_alpha_standup"),
```
Tara-ag, line 638: ⚠️ **MAJOR: Test case 15b for `promptIndexMaxItems: 0` is missing.**

The PR description lists "**15b** (`promptIndexMaxItems=0` → no injection)" as a newly added skills-suite case, but the diff only updates the existing test 15 (tool discovery listing) and does not add a 15b case exercising `system-prompt` discovery with `promptIndexMaxItems: 0`. Without this test, the `formatSkillsPromptIndex(... || maxItems <= 0)` change has no regression coverage.

**Suggested fix:** add a test case after test 15 that constructs a `NeuroLink` with `skills: { discovery: "system-prompt", promptIndexMaxItems: 0 }`, calls `applyAugmentation`, and asserts that the system prompt does not contain "## Available Skills".

Tara-ag (changes requested):
## Review summary

**Files reviewed:** 2 (`src/lib/skills/skillMatcher.ts`, `test/continuous-test-suite-skills.ts`)

**New issues raised:**
- 🔒 **1 CRITICAL** — `isSkillVisibleInScope` in `skillMatcher.ts` is inconsistent with the new fail-closed `filterSkillIndex` semantics. `use_skill` and `read_skill_resource` can still activate/read scoped skills by exact name when no `scopeId` is supplied, bypassing the multi-tenant leak fix.
- ⚠️ **1 MAJOR** — The PR description promises a new test case **15b** for `promptIndexMaxItems: 0`, but the diff does not include it, leaving the `formatSkillsPromptIndex` change without regression coverage.

**Additional concern (file not in this diff, so noted here):**
- 🔒 **`GET /api/agent/skills/:id` in `src/lib/server/routes/agentRoutes.ts` also bypasses the scope filter.** It calls `manager.get(ctx.params.id)` and returns the full skill (including instructions) with no scope check. This is a separate exfiltration vector that should be closed in the same security PR by applying the same fail-closed visibility check before returning the skill.

**Blocking rationale:** The scope leak is a CRITICAL security vulnerability in the opt-in skills path. A shared multi-tenant instance that forgets to pass a `scopeId` would still leak scoped skill instructions through direct name lookup (`use_skill`, `read_skill_resource`, and the REST GET-by-id endpoint), undermining the stated fix for #1139.

**Next steps:**
1. Align `isSkillVisibleInScope` with `filterSkillIndex` so scoped skills require a matching `scopeId`.
2. Add regression tests covering `use_skill` and `read_skill_resource` with no `scopeId` for a scoped skill.
3. Add the promised test case **15b** for `promptIndexMaxItems: 0`.
4. Apply the same fail-closed scope check to `GET /api/agent/skills/:id` in `src/lib/server/routes/agentRoutes.ts` and add a regression test.

murdore:
## ✅ Verification & gap-bridging (hardened)

Rebased onto current `release` and hardened across test / docs / **enhancement** / proof gaps.

### Test gap → closed (E2E / user-facing coverage)
The fix is exercised through the three surfaces a user actually hits, not just the pure function:
- **Manager (SDK):** new test `6b` — with no scopeId, `manager.search`/`manager.list` exclude the scoped seed skill; the two global skills stay visible.
- **Discovery listing (what the model sees):** test 15 now asserts the scoped skill is **absent** from the default (no-scopeId) `use_skill` `<available_skills>` listing — the exact leak path progressive disclosure (`b5be3beb`) introduced.
- **Server routes (HTTP):** tests 32 / 32b drive `GET/POST/PATCH/DELETE /api/agent/skills` and confirm a scoped skill appears only when `?scopeId=` is supplied, and that `allowMutations:false` still blocks writes.

### Docs gap → closed
`docs/features/skills.md` now has an explicit **"Scoping is fail-closed (multi-tenant safe)"** callout: scoped skills are returned only to callers supplying a matching `scopeId`; with none resolvable they're excluded from discovery/search/prompt-index; globals always visible; pass the tenant's `scopeId` to surface theirs.

### Enhancement gap (considered, deliberately not added)
An "include all scopes for admins" escape hatch was considered and **rejected** — it would reintroduce exactly the cross-tenant leak this fixes. The safe pattern is explicit per-scope queries. Left as a documented non-goal.

### Behavior-change note for reviewers
Because `b5be3beb`'s discovery listing runs through `filterSkillIndex`, this fix also removes scoped skills from the **default** discovery listing. That's intended multi-tenant-safe behavior; hosts relying on a single default scope should set `defaultScopeId`.

### Proof
`pnpm run check` ✅ (0 errors) · `pnpm run lint` ✅ (0 errors) · `pnpm run build` ✅ · `prettier --check` ✅ · skills suite **45 passed / 1 skipped** (the skip is the live `use_skill` test, gated on `SKILLS_SUITE_LIVE`).

Refs #1135.

Tara-ag (approved):
Autonomous review complete. I examined the 3 changed files and checked the existing review threads before commenting.

No new blocking or non-blocking issues were identified beyond the points already raised in the existing review comments. The existing threads cover the remaining `isSkillVisibleInScope` alignment and the missing regression test for `promptIndexMaxItems: 0`.

Summary: 3 files reviewed, 0 new issues raised.

src/lib/skills/skillMatcher.ts
murdore: ✅ Fixed in a0550c79. `isSkillVisibleInScope` now fails closed, matching `filterSkillIndex`:
```ts
if (skill.scope !== "scoped") return true;
if (!scopeId) return false;               // scoped + no scopeId → hidden
return (skill.scopeIds ?? []).includes(scopeId);
```
This closes the `use_skill` / `read_skill_resource` bypass. New regression tests **34b** (use_skill) and **34c** (read_skill_resource) assert a scoped skill is neither activatable nor readable without a `scopeId`, and *is* activatable with its own `scopeId`. All pass.

src/lib/skills/skillMatcher.ts
murdore: ✅ Fixed in a0550c79. `isSkillVisibleInScope` now fails closed, matching `filterSkillIndex`:
```ts
if (skill.scope !== "scoped") return true;
if (!scopeId) return false;               // scoped + no scopeId → hidden
return (skill.scopeIds ?? []).includes(scopeId);
```
This closes the `use_skill` / `read_skill_resource` bypass. New regression tests **34b** (use_skill) and **34c** (read_skill_resource) assert a scoped skill is neither activatable nor readable without a `scopeId`, and *is* activatable with its own `scopeId`. All pass.

test/continuous-test-suite-skills.ts
murdore: ✅ Added in a0550c79 — test **15b** builds a NeuroLink with `discovery: "system-prompt"` + `promptIndexMaxItems: 0`, runs `applySkillsAugmentation`, and asserts the system prompt does **not** contain `## Available Skills`. Passing.

Tara-ag (approved):
Reviewed 3 files. No new blocking issues.

- `src/lib/skills/skillMatcher.ts`: the fail-closed scope filter and the matching `isSkillVisibleInScope` change are correct and close the `use_skill` / `read_skill_resource` bypass noted in the earlier review threads.
- `test/continuous-test-suite-skills.ts`: new regression cases (6b, 15b, 34b, 34c) and the updated existing cases adequately cover the new behavior.
- `docs/features/skills.md`: the added scoping note accurately describes the fail-closed semantics.

The two prior review threads (`isSkillVisibleInScope` inconsistency and missing 15b test) are resolved in commit `a0550c79`.

Non-blocking follow-up observation (pre-existing, outside this diff): `GET /api/agent/skills/:id` in `src/lib/server/routes/agentRoutes.ts` still returns the full skill (including instructions) by id or exact name without accepting a `scopeId` or calling `isSkillVisibleInScope`. Consider adding `?scopeId=` support and a visibility guard there in a subsequent security pass so the REST read path is as fail-closed as the SDK tool paths fixed here.

PR #1149, opened 2026-07-11
fix(marketing): correct false/misleading claims (vector stores, compliance wording, provider count)

murdore:
A competitive-positioning audit surfaced several claims that don't survive a code check. This corrects them across the landing page, README, and docs. **No product behavior changes — copy only.**

## 1. "22 vector stores" → the honest reality
The landing FAQ and docs claimed **"22 vector stores" / "22 vector store adapters"** and integration with Pinecone/Weaviate/Chroma. In the shipped code, only **`InMemoryVectorStore`** exists (plus a pluggable `VectorStore` interface); the external stores are type definitions and doc samples, not shipped adapters. Reworded to *"an in-memory vector store with a pluggable interface for external stores."*
*(Same defect class as the earlier newsletter-claim corrections. Real adapters are planned separately — this makes the copy honest now.)*

## 2. Compliance bullets no longer read as certifications
`SOC2 Type II compliant deployments`, `ISO 27001 certified infrastructure compatible`, `HIPAA compatible`, `GDPR-compliant data handling` were reworded as **deployment-compatibility** statements — NeuroLink itself is not audited or certified, and the responsibility is the operator's. (README + docs/index + docs/features/index.)

## 3. Retired unbacked blanket claims
"battle-tested at enterprise scale" / "production-ready solution" on the README + docs homepages → narrower, provable framing. (No independent audit/benchmark/case study exists to back the blanket claim.)

## 4. Provider count reconciled to 24
The landing page still said **"21+ providers"** (the README was fixed to 24 in #1146). Reconciled the landing + docs to **24** for consistency. Historical/audit docs with explicit "snapshot when 9 providers shipped" banners were left as-is.

## Scope
43 files, copy-only. Prettier passes (table column re-alignment is a side effect of changed cell text). Full positioning analysis that motivated this is in the marketing repo under `docs/growth-exec/positioning/`.

## Summary by CodeRabbit

* **Documentation**
  * Updated product and marketing documentation to reflect support for 30+ AI providers, including text, voice, realtime, and media-generation capabilities.
  * Clarified historical provider-count references and audit-time context across analysis and test reports.
  * Refined security and compliance language to clarify deployment responsibilities and certification status.
  * Updated RAG documentation to describe an in-memory vector store with pluggable external integrations.
  * Revised guides, FAQs, examples, SEO metadata, and social-proof messaging for consistency.
* **Testing**
  * Added documentation for provider capability-matrix testing and clarified credential-based validation behavior.

PR #1186, opened 2026-07-17
feat(rag): Pinecone, pgvector, and Chroma vector-store adapters (client injection, 74 tests)

murdore:
## What

Three `VectorStore` implementations under `src/lib/rag/stores/`, exported via `@juspay/neurolink/rag` and the root barrel:

```ts
import { PineconeVectorStore, PgVectorStore, ChromaVectorStore } from "@juspay/neurolink/rag";
```

**Zero new runtime dependencies** — each adapter takes an already-constructed vendor client typed as a minimal structural interface (`PineconeIndexLike`, `PgClientLike`, `ChromaClientLike`). Users bring their own `@pinecone-database/pinecone` / `pg` / `chromadb`; `PgClientLike` is satisfied by both `pg.Pool` and `@electric-sql/pglite`. Consistent with the thin-core policy (docs/about/why-the-core-stays-thin.md).

## Semantics

- `query()` per the `VectorStore` contract (src/lib/types/rag.ts) plus `upsert()`/`delete()` mirroring `InMemoryVectorStore`: filter-first-then-topK, cosine-similarity scoring, conditional `vector` inclusion.
- Faithful `MetadataFilter` translation per backend. Operators a backend cannot express (`$not`/`$exists`/`$contains`/`$regex`/`$size` on Pinecone and Chroma; `$nor`/`$size` on pgvector) **throw a clear named error** instead of silently mis-filtering. Divergences are documented in-code (e.g., pgvector `$eq` uses jsonb structural equality — strictly more capable than JS `===` for object values).
- `PgVectorStore` validates index names against a strict identifier regex and parameterizes every value.

## Verification

- **74/74 tests pass** across three new suites in the repo's continuous-test-suite convention:
  - `vector-pinecone` (23) and `vector-chroma` (28): recording mock clients asserting the exact translated filter payloads, topK defaults, score passthrough/conversion, error paths
  - `vector-pgvector` (23): **real in-process Postgres with `CREATE EXTENSION vector`** via `@electric-sql/pglite` (devDependency) — genuine cosine-distance ordering, jsonb filters, delete, injection-attempt rejection
- `tsc --noEmit` strict clean · prettier clean · eslint clean (type aliases live in `src/lib/types/` per repo rules)
- `InMemoryVectorStore` untouched
- New scripts: `test:vector-pinecone`, `test:vector-pgvector`, `test:vector-chroma`

## Notes for reviewers

Live integration against hosted Pinecone / a Chroma server still needs credentials/infra — the mock suites pin the wire-format contracts, and pgvector is genuinely executed. This also retires the last gap behind the marketing claim history around "22 vector stores" (landing already corrected in #1149): NeuroLink now ships 1 in-memory + 3 real adapters with a documented bring-your-own-client interface.

## Summary by CodeRabbit

* **New Features**
  * Added vector store integrations for Pinecone, PostgreSQL with pgvector, and Chroma.
  * Added filtering, similarity search, metadata handling, upsert, deletion, and optional vector retrieval across supported stores.
  * Exposed the new adapters and related types through the public library exports.

* **Tests**
  * Added continuous test coverage for all three integrations, including filtering, ranking, validation, and error handling.


src/lib/rag/stores/chroma.ts
```diff
@@ -0,0 +1,365 @@
...
+
+  const chromaOps: Record<string, unknown> = {};
+  for (const opKey of opKeys) {
+    if (!CHROMA_SUPPORTED_FIELD_OPS.has(opKey)) {
+      throw unsupportedFieldOpError(opKey, field);
+    }
+    chromaOps[opKey] = ops[opKey];
+  }
+  return { [field]: chromaOps };
+}
```
Tara-ag, line 149: ⚠️ **MAJOR: Filter semantics diverge from `InMemoryVectorStore` reference for object filters with non-`$` keys**

`translateFieldFilter` throws when a field filter value is an object containing non-`$` keys (e.g. `{ field: { foo: 1 } }`). The `InMemoryVectorStore.matchesFilter` reference treats any non-null object as an operator bag and applies **no constraint** when it contains no recognized `$`-operators.

Throwing here makes the Chroma adapter stricter than the reference contract and can break callers that pass opaque metadata-shaped filters.

**Suggested fix:** filter object keys to only those starting with `$`, and return `undefined` when none are present:

```ts
const chromaOps: Record<string, unknown> = {};
for (const opKey of opKeys) {
  if (!opKey.startsWith("$")) {
    continue;
  }
  if (!CHROMA_SUPPORTED_FIELD_OPS.has(opKey)) {
    throw unsupportedFieldOpError(opKey, field);
  }
  chromaOps[opKey] = ops[opKey];
}
return Object.keys(chromaOps).length > 0 ? { [field]: chromaOps } : undefined;
```

This aligns with the documented "no recognized `$`-key → no constraint" behavior already described in the JSDoc.

src/lib/rag/stores/pgvector.ts
```diff
@@ -0,0 +1,524 @@
...
+        );
+      }
+    }
+
+    return clauses.length > 0 ? clauses.join(" AND ") : "TRUE";
+  }
+
+  // Direct equality (primitive, array, or null filter value).
+  return `${fieldJson()} = ${pushParam(values, JSON.stringify(value))}::jsonb`;
+}
```
Tara-ag, line 515: ⚠️ **MAJOR: Filter semantics diverge from `InMemoryVectorStore` reference for array-valued field filters**

`compileFieldFilter` currently treats an array-valued field filter (e.g. `{ tags: ["a", "b"] }`) as a direct jsonb equality constraint: `(metadata -> $n) = $m::jsonb` where `$m` is `'["a","b"]'`.

The `InMemoryVectorStore.matchesFilter` reference treats arrays as a non-null object operator-bag with no recognized `$`-keys, which applies **no constraint** (matches unconditionally). This mismatch can cause surprising filtering differences across vector stores.

**Suggested fix:** return `"TRUE"` for array-valued filter values to mirror the reference behavior (and the Chroma adapter's documented behavior):

```ts
if (Array.isArray(value)) {
  return "TRUE";
}
```

Insert this guard before the `typeof value === "object"` branch at line 515.

src/lib/rag/stores/pinecone.ts
```diff
@@ -0,0 +1,208 @@
...
+        }
+        translatedOps[opKey] = ops[opKey];
+      }
+      result[key] = translatedOps;
+    } else {
+      // Primitive/array/null value: Pinecone treats a bare value as
+      // implicit equality, matching InMemoryVectorStore's direct-equality
+      // fallback for non-object filter values.
+      result[key] = value;
+    }
```
Tara-ag, line 116: ⚠️ **MAJOR: Filter semantics diverge from `InMemoryVectorStore` reference for array-valued and unknown-key filters**

`translatePineconeFilter` currently passes array-valued field filters through as bare values (e.g. `{ tags: ["a", "b"] }` becomes `{ tags: ["a", "b"] }`) and throws on non-`$` keys inside an object filter value (e.g. `{ field: { foo: 1 } }`).

The `InMemoryVectorStore.matchesFilter` reference treats any non-null object value as an operator bag; when it contains no recognized `$`-operators it applies **no constraint** (the field matches unconditionally). Arrays are objects, so they also enforce no constraint in the reference.

Passing arrays through can produce invalid Pinecone filter payloads or surprising equality semantics, and throwing on unknown keys makes the adapter stricter than the reference contract.

**Suggested fix:** mirror the Chroma adapter's approach:

```ts
if (Array.isArray(value)) {
  return undefined; // no constraint
}
if (value === null || typeof value !== "object") {
  result[key] = value;
  continue;
}
const ops = value as Record<string, unknown>;
const translatedOps: Record<string, unknown> = {};
for (const opKey of Object.keys(ops)) {
  if (!opKey.startsWith("$")) {
    continue; // ignore non-operator keys, no constraint
  }
  if (!SUPPORTED_COMPARISON_OPS.has(opKey)) { /* throw */ }
  translatedOps[opKey] = ops[opKey];
}
if (Object.keys(translatedOps).length === 0) {
  continue; // omit field entirely when no recognized operators
}
result[key] = translatedOps;
```

This keeps the translated filter consistent with `InMemoryVectorStore` and avoids invalid Pinecone payloads.

src/lib/rag/stores/chroma.ts
```diff
@@ -0,0 +1,365 @@
...
+        distance === null || distance === undefined
+          ? undefined
+          : distanceToScore(distance, this.distanceMetric);
+
+      return {
+        id,
+        score,
+        text: metadata.text as string | undefined,
+        metadata,
+        ...(includeVectors ? { vector: embeddingsRow[i] ?? [] } : {}),
```
Tara-ag, line 350: ⚠️ **MAJOR: `includeVectors=true` can return a misleading zero-dimension vector when Chroma omits the embedding**

When `includeVectors` is true, `query()` always spreads `{ vector: embeddingsRow[i] ?? [] }`. If Chroma returns `null` for an embedding (e.g. the server ignores the `include` field, or returns `null` for a row), this silently converts "missing vector" into a `[]` vector.

This contradicts the other adapters (Pinecone only includes `vector` when `match.values` is truthy; pgvector only includes it when `row.embedding` is a string) and the `VectorQueryResult` contract, where `vector` should be present only when a real embedding is returned.

**Suggested fix:** only include the `vector` key when the embedding array is actually present:

```ts
const embedding = embeddingsRow[i];
return {
  id,
  score,
  text: metadata.text as string | undefined,
  metadata,
  ...(includeVectors && Array.isArray(embedding) ? { vector: embedding } : {}),
};
```

Tara-ag (changes requested):
## Review Summary

Reviewed PR #1186 — feat(rag): Pinecone, pgvector, and Chroma vector-store adapters.

**Files reviewed:** 15 changed files (3 new adapter implementations, 2 new type files, 3 new test suites, plus package.json / lockfile / barrel exports).

**New issues raised:** 4 MAJOR

1. **`src/lib/rag/stores/pinecone.ts:116`** — `translatePineconeFilter` mishandles array-valued field filters and object filters with non-`$` keys, diverging from `InMemoryVectorStore.matchesFilter` semantics.
2. **`src/lib/rag/stores/chroma.ts:149`** — `translateFieldFilter` throws on object filters with non-`$` keys instead of treating them as no-constraint, diverging from the reference semantics.
3. **`src/lib/rag/stores/pgvector.ts:515`** — `compileFieldFilter` treats array-valued field filters as jsonb equality instead of no-constraint, diverging from the reference semantics.
4. **`src/lib/rag/stores/chroma.ts:350`** — `query()` returns `[]` for missing embeddings when `includeVectors=true`, producing a misleading zero-dimension vector instead of omitting the key.

All four are semantic correctness / contract-consistency issues. The first three create behavioral divergence from the existing `InMemoryVectorStore` reference that callers already rely on; the fourth is a data-shape inconsistency across adapters.

**Existing review comments:** 4 Copilot review comments were already present on the same lines raising the same points. I have added my own comments because those threads remain unresolved and the issues are substantive, but I am not double-counting them as new blocking criteria — the decision below is based on the substance, not the author.

**Blocking criteria applied:** Per the project blocking criteria, 3 or more MAJOR severity issues in the same PR is grounds to block. This PR has 4 MAJOR issues (even if some overlap with unresolved prior feedback).

**Next steps:**
- Align Pinecone, Chroma, and pgvector filter translators with `InMemoryVectorStore.matchesFilter` semantics for array-valued filters and object filters lacking recognized `$`-operators.
- Make Chroma omit the `vector` key when no embedding is returned.
- Add/update tests covering these edge cases (array filters, unknown-key object filters, missing-embedding responses).
- Re-run `pnpm run check`, `pnpm run lint`, and the three new vector test suites before re-requesting review.

I am happy to re-review promptly once these are addressed.

PR #1187, opened 2026-07-17
fix(deps): move lazily-imported cloud provider SDKs to optionalDependencies

murdore:
## What

Moves six heavyweight cloud SDKs from `dependencies` to `optionalDependencies` — the concrete next step promised by [Why the core stays thin](docs/about/why-the-core-stays-thin.md), whose "where we're heavy today" section this PR also updates to match reality:

`@aws-sdk/client-bedrock` · `@aws-sdk/client-bedrock-runtime` · `@aws-sdk/credential-provider-node` · `@google-cloud/vertexai` · `@google-cloud/text-to-speech` · `@huggingface/inference`

## The audit mattered — it found two real pre-existing bugs

A naive move would have shipped broken installs (the js-yaml cold-start failure mode). The per-package reachability audit found and fixed:

1. **`@google-cloud/vertexai`** was eagerly reachable from the root barrel (`src/lib/index.ts` → `neurolink.ts` → `directToolsServer.ts` → `directTools.ts` top-level static import). Now a dynamic import at point of use inside the websearchGrounding tool, after the credential check.
2. **`@google-cloud/text-to-speech`** was eagerly *constructed* on every root-package import (`voice/index.ts`'s top-level `registerDefaultTTSHandlers()` built a `GoogleTTSHandler`, whose constructor built the client). Now `import type` + a lazy cached `getClient()` used by `getVoices()`/`synthesize()`. Public API unchanged.

## Honest scope notes

- **`@aws-sdk/client-sagemaker-runtime` did NOT move** (deliberate): its client is constructed synchronously in a chain reachable from CLI startup (`src/cli/index.ts` → parser → commandFactory → sagemaker/client.ts) — converting it to lazy async init is real follow-up work with 4+ call sites, tracked as the named outstanding item on the docs page.
- `@aws-sdk/credential-provider-node` moved, but yields **zero disk savings today** — it's a transitive dep of the still-required sagemaker-runtime. It becomes real once SageMaker moves.
- Optional deps still install by default; the win is they can be skipped (`--omit=optional` / platform install failure) safely, because every use site is lazily imported.

## Verification (all empirical, clean-room)

- `tsc --noEmit`: clean · `pnpm pack`: builds (9.89.0 tarball) · prettier clean
- **Clean-room A** (default `npm install <tarball>`): root import → 721 exports; `neurolink --version` works
- **Clean-room B** (`--omit=optional`): identical import/version checks pass; requesting bedrock without its SDK yields a *deferred, caught, comprehensible* error at call time:
  > Failed to create provider bedrock: ERR_MODULE_NOT_FOUND: Cannot find package '@aws-sdk/client-bedrock-runtime' imported from …/dist/providers/amazonBedrock.js

## Summary by CodeRabbit

* **New Features**
  * Optional provider integrations can now be excluded during installation when needed.
  * Google Cloud text-to-speech and web search services load only when used.
  * Core package imports and CLI version checks continue working without optional integrations.

* **Bug Fixes**
  * Requests for unavailable optional providers now return a clear missing-package error instead of failing during startup.

* **Documentation**
  * Clarified optional installation behavior, supported integrations, and remaining dependency limitations.


src/lib/adapters/tts/googleTTSHandler.ts
Tara-ag, line 1: 💡 **Minor: Add regression tests for the lazy-import behavior.**

The move from eager to lazy client construction is the critical behavioral change in this file. The PR description verifies it empirically via clean-room installs, but there is no automated test guarding against a future contributor accidentally re-introducing a top-level `import { TextToSpeechClient } from "@google-cloud/text-to-speech"` or moving `getClient()` back into the constructor.

Consider adding a test in the closest matching suite (likely `test/continuous-test-suite-tts.ts` or `test/continuous-test-suite-context.ts`) that asserts:

1. `new GoogleTTSHandler()` does not synchronously reference `@google-cloud/text-to-speech`.
2. `getVoices()` / `synthesize()` throw `TTS_ERROR_CODES.PROVIDER_NOT_CONFIGURED` when no credentials are present.
3. When credentials are present but the optional package is missing, the error is a `TTSError` (or at least caught and transformed) rather than an unhandled rejection.

This would prevent the exact pre-existing bug from regressing.

src/lib/agent/directTools.ts
```diff
@@ -746,6 +745,7 @@ export const directAgentTools = {
         }
```
Tara-ag, line 746: 💡 **Minor: Missing targeted error for missing optional dependency.**

The dynamic import of `@google-cloud/vertexai` is correctly placed after the credential check and inside the tool execution, satisfying CLAUDE.md Rule 1. However, if the optional dependency was omitted (e.g. `--omit=optional`) but the env vars are set, this will surface a raw `ERR_MODULE_NOT_FOUND` through the catch block below rather than a helpful install hint.

Elsewhere in the codebase (e.g. `src/lib/processors/document/WordProcessor.ts`, `src/lib/tasks/backends/bullmqBackend.ts`) the convention is to catch `ERR_MODULE_NOT_FOUND`, verify the package name, and throw a contextual error with an install command. Consider aligning this site with that pattern:

```typescript
let VertexAI;
try {
  ({ VertexAI } = await import("@google-cloud/vertexai"));
} catch (err) {
  const e = err instanceof Error ? (err as NodeJS.ErrnoException) : null;
  if (e?.code === "ERR_MODULE_NOT_FOUND" && e.message.includes("@google-cloud/vertexai")) {
    throw new Error(
      'Web search grounding requires "@google-cloud/vertexai". Install it with:\n  pnpm add @google-cloud/vertexai',
      { cause: err },
    );
  }
  throw err;
}
```

This keeps the lazy-load benefit while giving consumers a clear remediation path.

Tara-ag (approved):
## Review Summary

**Files reviewed:** 5 changed files
**New issues raised this run:** 2 (both MINOR / SUGGESTION)
**Decision:** Approve

### New comments

1. **`src/lib/agent/directTools.ts:746`** — The dynamic import of `@google-cloud/vertexai` is correctly placed after the credential check, but it surfaces a raw `ERR_MODULE_NOT_FOUND` when the optional package is omitted. Suggested aligning with the repo’s existing optional-dependency error pattern (catch + install hint).

2. **`src/lib/adapters/tts/googleTTSHandler.ts`** — The lazy client construction is the key behavioral fix here. Suggested adding an automated regression test so a future change cannot accidentally re-introduce an eager top-level import or constructor-side client construction.

### Notes

- The `package.json` / `pnpm-lock.yaml` moves match the PR description: the six SDKs land in `optionalDependencies`, while `@aws-sdk/client-sagemaker-runtime` remains in `dependencies` as documented.
- The doc update in `docs/about/why-the-core-stays-thin.md` is consistent with the actual changes.
- Several pre-existing review comments from Copilot and CodeRabbit remain unresolved on `googleTTSHandler.ts` (type-only import from an optional dependency, empty/whitespace credential handling, moving `getClient()` inside `try` blocks, wrapping API calls with `withTimeout`, and providing a targeted missing-package error). I did not duplicate those points; the author should address them in follow-up.

No new blocking criteria are met, so this PR is approved.

PR #1188, opened 2026-07-17
feat(proxy): add compatibility telemetry foundation

murdore:
## Summary

- freeze the NeuroLink 9.88.12 four-phase body-capture contract, including redaction, truncation, gzip/index schemas, and permissions
- add a bounded, batched lifecycle metadata sink with explicit loss accounting
- measure response headers, first chunk, terminal outcome, decoded bytes, chunks, and stable hashed sessions
- preserve low-level transport codes and retry transient Anthropic DNS failures
- expose lifecycle quality counters in `/status` and flush pending metadata during graceful shutdown
- add a synthetic hot-path and write-throughput benchmark

## Compatibility and scope

- rebased on current `origin/release` (`9.92.3`)
- existing full body capture remains enabled and byte/schema compatible
- no account-ordering, cooldown, provider, SDK, account configuration, or proxy configuration changes
- `ENOTFOUND` enters the existing bounded retry/failover state machine
- every dispatched upstream request counts as one attempt error when transport fails, including terminal non-retryable failures; terminal throw behavior is unchanged
- `request_accepted` omits model, stream, and tool count before parsing instead of emitting false placeholders; later events carry parsed metadata under the same request ID
- session hashes remain stable across restarts through `NEUROLINK_PROXY_SESSION_SECRET` or an automatically generated mode-`0600` installation key
- logger reconfiguration preserves the enqueue-time destination for every pending record
- lifecycle metadata deliberately shares the full request logger enablement, permissions, directory, retention, and privacy boundary
- lifecycle writes intentionally avoid fsync so storage latency remains off the request path; write loss is explicit in counters
- telemetry and observer failures cannot interrupt client responses

## Recent failure evidence covered

- recent failures were transport interruptions, not 429 quota responses or proxy restarts
- retryable `UND_ERR_CONNECT_TIMEOUT`, `ETIMEDOUT`, `EADDRNOTAVAIL`, and `ENOTFOUND` attempts are distinguishable in structured logs
- lifecycle terminal events make accepted-but-unsettled requests classifiable without changing full-response capture

## Verification

- focused proxy Vitest suites: 90 passed
- `pnpm run test:unit:vitest`: 42 passed
- `pnpm run typecheck`
- changed-file ESLint: 0 errors, existing warnings only
- `pnpm run build` and `publint`
- repository pre-commit validation, security, lint, and build gate passed
- benchmark smoke: 1,000 requests, 4,000 events, about 4 us p50 and 17 us p99 enqueue cost, about 458,000 events/s flush, zero drops

## Existing baseline test debt

`pnpm run test:bugfixes` remains 99/101 because two stale updater source-string assertions expect the pre-`validateInstalledVersion()` implementation. The updater implementation and source region are unchanged by this PR.

## Summary by CodeRabbit

- **New Features**
  - Added proxy request lifecycle observability, including response timing, streaming progress, terminal outcomes, and runtime error codes.
  - Added lifecycle metrics to the proxy status endpoint.
  - Added configurable, session-correlated lifecycle log storage with safe buffering and shutdown flushing.
  - Added support for retrying DNS resolution failures when appropriate.
  - Added `NEUROLINK_PROXY_SESSION_SECRET` configuration documentation.

- **Bug Fixes**
  - Improved network error classification and error-code reporting.
  - Preserved response delivery when observability callbacks fail.
  - Improved handling and reporting of cancelled or failed streams.


src/lib/proxy/proxyLifecycle.ts
```diff
@@ -0,0 +1,291 @@
...
+    DEFAULT_QUEUE_CAPACITY,
+  );
+  batchSize = positiveInteger(options.batchSize, DEFAULT_BATCH_SIZE);
+  flushIntervalMs = positiveInteger(
+    options.flushIntervalMs,
+    DEFAULT_FLUSH_INTERVAL_MS,
+  );
+
+  if (!loggerEnabled && queue.length > 0) {
+    dropped += queue.length;
```
Tara-ag: ⚠️ **MAJOR**: Calling `nonNegativeInteger(input.toolCount)` (and similar helpers) twice per field performs redundant validation and is easy to desynchronize if the helper is later changed. More importantly, the spread pattern means each optional field is evaluated twice, which is unnecessary overhead on a hot path that targets single-digit-microsecond p50 cost.

Suggested fix: compute sanitized values once and reuse them in the record construction.

```ts
const toolCount = nonNegativeInteger(input.toolCount);
const requestBytes = nonNegativeInteger(input.requestBytes);
...
const record = {
  ...
  ...(toolCount !== undefined ? { toolCount } : {}),
  ...(requestBytes !== undefined ? { requestBytes } : {}),
  ...
};
```

src/lib/proxy/proxyLifecycle.ts
```diff
@@ -0,0 +1,291 @@
...
+          lines.join(""),
+          { mode: 0o600 },
+        );
+        written += lines.length;
+      } catch (error) {
+        dropped += lines.length;
+        writeDrops += lines.length;
+        writeFailures += 1;
+        logger.warn("[proxy] lifecycle metadata write failed", {
+          error: error instanceof Error ? error.message : String(error),
```
Tara-ag, line 191: 🔒 **CRITICAL**: The lifecycle log directory is created lazily only when the first `appendFile` runs. If the configured `logDir` does not exist, every batch write fails, increments `writeFailures`, and drops events. For a telemetry sink this means silent data loss on startup misconfiguration.

Suggested fix: ensure the directory exists (with restrictive permissions) during `configureProxyLifecycleLogger`, matching how `initRequestLogger` creates `logDir` with `mkdirSync(..., { recursive: true, mode: 0o700 })`. If directory creation fails, disable the logger so the failure is explicit rather than per-batch drop.

```ts
import { mkdir } from "node:fs/promises";

export async function configureProxyLifecycleLogger(
  options: ProxyLifecycleLoggerOptions,
): Promise<void> {
  loggerEnabled = options.enabled && Boolean(options.logDir);
  lifecycleLogDir = options.logDir;
  ...
  if (loggerEnabled && lifecycleLogDir) {
    try {
      await mkdir(lifecycleLogDir, { recursive: true, mode: 0o700 });
    } catch (error) {
      loggerEnabled = false;
      lifecycleLogDir = undefined;
      logger.warn("[proxy] lifecycle logging disabled — failed to create log directory", { error: error instanceof Error ? error.message : String(error) });
    }
  }
}
```

src/lib/proxy/proxyLifecycle.ts
```diff
@@ -0,0 +1,291 @@
...
+    return;
+  }
+
+  const record = {
+    schemaVersion: SCHEMA_VERSION,
+    timestamp: new Date(timestampMs).toISOString(),
+    monotonicMs: Number((input.monotonicMs ?? performance.now()).toFixed(3)),
+    processInstanceId,
+    sequence,
+    event: input.event,
```
Tara-ag: 💡 **MINOR**: `monotonicMs: Number((input.monotonicMs ?? performance.now()).toFixed(3))` recomputes `performance.now()` on every call even when the caller did not supply `monotonicMs`. This is fine for correctness but slightly obscures the intent. Consider capturing the fallback once:

```ts
const monotonicMs = input.monotonicMs ?? performance.now();
...
monotonicMs: Number(monotonicMs.toFixed(3)),
```

src/lib/proxy/proxyActivity.ts
```diff
@@ -49,33 +53,72 @@ export function isProxyActivityQuiet(
 export function trackProxyResponse(
```
Tara-ag, line 70: 💡 **MINOR**: The `settle` helper and bodyless branch duplicate the `try/catch` wrapper around `observer?.onTerminal?.(...)`. Consider a small internal helper like `safeNotifyObserver(callback)` to avoid the repetition and reduce the chance of future divergence. Not blocking.

src/lib/proxy/proxyActivity.ts
```diff
@@ -49,33 +53,72 @@ export function isProxyActivityQuiet(
...
+    settled = true;
+    finishRequest();
+    try {
+      observer?.onTerminal?.({ outcome, responseBytes, responseChunks });
+    } catch {
+      // Observability must never alter response handling.
+    }
+  };
+
   const trackedBody = new ReadableStream<Uint8Array>({
```
Tara-ag, line 107: ⚠️ **MAJOR**: `responseBytes += value.byteLength` accumulates the byte length of each enqueued chunk. This is correct for the uncompressed body size, but the variable name `responseBytes` is later passed to observers as `responseBytes`. Ensure downstream consumers (and the test assertions) interpret this as the bytes observed on the wire before compression, not the final HTTP content-length. If the upstream response was compressed (e.g. `content-encoding: gzip`), this count will differ from the actual transferred bytes. Document the semantics in the type or consider renaming to `observedBodyBytes` to avoid confusion with the HTTP metric of the same name.

Suggested clarification: rename the field in `ProxyResponseTrackingObserver` to `observedBodyBytes` and update callers/tests accordingly, or add a JSDoc note that this is the sum of decoded chunk lengths.

src/cli/commands/proxy.ts
```diff
@@ -1101,6 +1108,126 @@ async function resolveBootPrimaryAccountKey(
...
+        method: metadata.method,
+        path: metadata.path,
+        model: metadata.model,
+        stream: metadata.stream,
+        toolCount: metadata.toolCount,
+        sessionHash,
+        requestBytes,
+        elapsedMs: performance.now() - startedMonotonicMs,
+        terminalOutcome: "handler_error",
+        errorType: error instanceof Error ? error.name : "unknown_error",
```
Tara-ag, line 1097: ⚠️ **MAJOR**: The `onFirstChunk` and `onTerminal` callbacks invoke `logProxyLifecycleEvent` synchronously while the response stream is being consumed. `logProxyLifecycleEvent` is designed to be non-blocking, but it currently performs object construction, optional-field sanitization, JSON serialization planning, and queue push work on every call. If the queue is near capacity or the flush timer fires, this can add jitter to chunk delivery. More importantly, if `logProxyLifecycleEvent` ever throws (e.g. due to an invalid `timestampMs`), the exception will propagate through the stream reader and abort the response.

Suggested fix: wrap the lifecycle logging calls inside the observer callbacks in a `try/catch` (consistent with the defensive pattern already used in `trackProxyResponse` for its observer), or make `logProxyLifecycleEvent` itself guaranteed non-throwing.

```ts
onFirstChunk: (details) => {
  try {
    logProxyLifecycleEvent({ ... });
  } catch {
    // Observability must never alter response handling.
  }
},
onTerminal: (details) => {
  try {
    logProxyLifecycleEvent({ ... });
  } catch {
    // Observability must never alter response handling.
  }
},
```

src/cli/commands/proxy.ts
```diff
@@ -1101,6 +1108,126 @@ async function resolveBootPrimaryAccountKey(
   return key;
 }
 
+function registerProxyRequestTracking(
+  app: Hono,
+  requestMetadata: WeakMap<Request, RuntimeRequestMetadata>,
+): void {
```
Tara-ag, line 983: ⚠️ **MAJOR**: The `registerProxyRequestTracking` middleware logs `request_accepted` before `beginProxyRequest()` increments the active-request counter. If the process crashes or is inspected between these two points, the lifecycle event will show an accepted request while `getProxyActivitySnapshot()` reports no active request. More importantly, the `finish` callback deletes metadata and decrements activity, but on the exception path `finishActivity()` is called directly without calling `finish()`, leaving the `requestMetadata` WeakMap entry pinned until GC collects the Request object.

Suggested fix: call `beginProxyRequest()` and set up `finish` before emitting `request_accepted`, and in the catch block call `finish()` instead of `finishActivity()` directly so metadata is cleaned up consistently.

```ts
const finishActivity = beginProxyRequest();
const finish = () => {
  finishActivity();
  requestMetadata.delete(c.req.raw);
};
logProxyLifecycleEvent({ ... });
...
catch (error) {
  finish();
  ...
}
```

src/cli/commands/proxy.ts
```diff
@@ -1101,6 +1108,126 @@ async function resolveBootPrimaryAccountKey(
...
+  app.use("/v1/*", async (c, next) => {
+    const startedMonotonicMs = performance.now();
+    const contentLengthHeader = c.req.raw.headers.get("content-length");
+    const rawContentLength =
+      contentLengthHeader === null ? Number.NaN : Number(contentLengthHeader);
+    const requestBytes =
+      Number.isFinite(rawContentLength) && rawContentLength >= 0
+        ? rawContentLength
```
Tara-ag, line 991: 💡 **MINOR**: `c.req.raw.headers.get(...)` is accessed twice for session ID resolution. This is readable but duplicates the header lookup. Consider collapsing to a single expression or helper if the list grows. Not blocking.

Tara-ag (changes requested):
## Review summary

Reviewed all 8 changed files. I did **not** duplicate the 3 issues already raised by the earlier Copilot review (benchmark `NaN`, `timestampMs` `RangeError`, and the two midnight-date test flakiness comments).

### New issues raised

| Severity | Count | Files |
|----------|-------|-------|
| CRITICAL | 1 | `src/lib/proxy/proxyLifecycle.ts` — lifecycle log directory is not created eagerly, so a misconfigured/missing `logDir` causes permanent per-batch write drops (counted data loss). |
| MAJOR | 4 | `proxyLifecycle.ts` redundant double-sanitization on the hot path; `proxy.ts` `beginProxyRequest` ordering/metadata cleanup inconsistency on the exception path; `proxy.ts` observer callbacks not defended against throws from `logProxyLifecycleEvent`; `proxyActivity.ts` `responseBytes` semantics vs. compressed transfers. |
| MINOR | 3 | `proxyLifecycle.ts` `performance.now()` fallback recomputation; `proxy.ts` duplicate session-header lookup; `proxyActivity.ts` duplicated observer `try/catch`. |

### Blocking rationale

There are **3+ MAJOR** issues in this PR, meeting the blocking criterion. The most important ones to fix before merge are:

1. **Make observability non-throwing on the response stream path.** The `onFirstChunk`/`onTerminal` callbacks in `registerProxyRequestTracking` call `logProxyLifecycleEvent` without a guard. If the logger ever throws, the exception propagates through the `ReadableStream` reader and aborts the client response. Either wrap those calls in `try/catch` or make `logProxyLifecycleEvent` itself guaranteed non-throwing.
2. **Fix the exception-path cleanup in `registerProxyRequestTracking`.** On the success path `finish()` decrements activity **and** deletes the `WeakMap` entry; on the exception path only `finishActivity()` is called, so metadata stays pinned until GC. Call `finish()` consistently.
3. **Create the lifecycle log directory during configuration.** The current lazy creation means every batch is dropped with `ENOENT` if the directory is missing. Eagerly `mkdir` (sync, `0o700`) in `configureProxyLifecycleLogger` and disable the logger if creation fails, matching `initRequestLogger`.

### What’s good

- The bounded queue, sequence IDs, and explicit drop counters are well-designed.
- Body-capture compatibility fixture and redaction assertions are thorough.
- The `/status` exposure and graceful-shutdown flush are sensible additions.
- No CLAUDE.md critical-rule violations (dynamic imports, Gemini schema/tools, CLI/SDK separation, backward compatibility) were introduced.

Please address the blocking issues and re-request review.

tools/testing/proxyLifecycleBenchmark.ts
```diff
@@ -0,0 +1,140 @@
...
+  getProxyLifecycleLoggerSnapshot,
+  hashProxyLifecycleSessionId,
+  logProxyLifecycleEvent,
+  resetProxyLifecycleLoggerForTests,
+} from "../../src/lib/proxy/proxyLifecycle.js";
+
+const REQUESTS = Math.max(
+  1_000,
+  Number(process.env.PROXY_LIFECYCLE_BENCH_REQUESTS ?? 25_000),
+);
```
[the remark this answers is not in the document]
murdore: Addressed in e77a8b6a. The benchmark now validates the parsed environment value with Number.isFinite, floors valid values, and falls back to 25,000. The 100,000-event run completes with zero drops.

src/lib/proxy/proxyLifecycle.ts
```diff
@@ -0,0 +1,291 @@
...
+    queueDrops += 1;
+    return;
+  }
+
+  const record = {
+    schemaVersion: SCHEMA_VERSION,
+    timestamp: new Date(timestampMs).toISOString(),
+    monotonicMs: Number((input.monotonicMs ?? performance.now()).toFixed(3)),
+    processInstanceId,
+    sequence,
```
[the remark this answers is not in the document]
murdore: Addressed in e77a8b6a. Timestamp formatting now validates finite and in-range values with a current-time fallback. The public lifecycle logger also contains malformed runtime metadata and records invalidDrops instead of throwing.

test/proxyObservabilityFoundation.test.ts
```diff
@@ -0,0 +1,453 @@
...
+    expect(JSON.parse(responseText)).toMatchObject({
+      data: expect.any(Array),
+      has_more: false,
+    });
+    await flushProxyLifecycleEvents();
+
+    const date = new Date().toISOString().slice(0, 10);
+    const events = await readJsonLines(
+      join(logDir, `proxy-lifecycle-${date}.jsonl`),
+    );
```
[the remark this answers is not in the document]
murdore: Addressed in e77a8b6a. The test now discovers the single proxy-lifecycle JSONL artifact in the temporary directory, removing the assertion-time date dependency.

test/proxyObservabilityFoundation.test.ts
```diff
@@ -0,0 +1,453 @@
...
+    enqueue("request-2");
+    enqueue("request-dropped");
+    await flushProxyLifecycleEvents();
+    enqueue("request-4");
+    await flushProxyLifecycleEvents();
+
+    const date = new Date().toISOString().slice(0, 10);
+    const events = await readJsonLines(
+      join(logDir, `proxy-lifecycle-${date}.jsonl`),
+    );
```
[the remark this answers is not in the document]
murdore: Addressed in e77a8b6a. The bounded-queue test uses a fixed event timestamp and deterministic lifecycle filename, so midnight cannot change the expected artifact.

src/lib/proxy/proxyLifecycle.ts
murdore: Addressed in e77a8b6a. Every optional numeric and short string field is sanitized once and the computed value is reused during record construction.

src/lib/proxy/proxyLifecycle.ts
murdore: Addressed in e77a8b6a. Configuration eagerly creates the directory synchronously with mode 0700 and disables the sink with a warning if setup fails. Tests cover creation, configuration failure, and later write failure separately.

src/lib/proxy/proxyLifecycle.ts
murdore: Addressed in e77a8b6a. The monotonic value is captured once, validated as finite and non-negative, and falls back to one performance.now call.

src/lib/proxy/proxyActivity.ts
murdore: Addressed in e77a8b6a. Both terminal paths and the first-chunk callback now use one safelyNotifyObserver helper. A regression test proves observer exceptions do not alter response delivery.

src/lib/proxy/proxyActivity.ts
murdore: Addressed in e77a8b6a. The field is renamed to observedBodyBytes throughout and documented as decoded response-body bytes observed by the adapter.

src/cli/commands/proxy.ts
murdore: Addressed in e77a8b6a. logProxyLifecycleEvent is now non-throwing for malformed runtime input, and response observer invocation is independently contained. Tests cover invalid clocks, throwing metadata accessors, and throwing observers.

src/cli/commands/proxy.ts
murdore: Partially following the suggestion by design in e77a8b6a: beginProxyRequest now precedes request_accepted. The catch intentionally calls finishActivity rather than deleting metadata because Hono onError must use that metadata to emit the client-facing error with the same request ID. WeakMap entries do not pin Request keys. A regression test verifies one correlated request ID, the low-level error code, and activeRequests returning to zero.

src/cli/commands/proxy.ts
murdore: Addressed to the useful extent in e77a8b6a. The two get calls read two distinct supported header aliases, so both are required; the resolved session ID is now captured once and hashed once rather than embedding the lookup in downstream work.

src/lib/proxy/proxyLifecycle.ts
```diff
@@ -0,0 +1,328 @@
+import { createHmac, randomBytes, randomUUID } from "node:crypto";
```
Tara-ag: 🔒 **CRITICAL**: `sessionHashKey` is initialized with `randomBytes(32)` once per process and never rotated. If an attacker gains read access to the process memory or a core dump, they can recover the key and deanonymize all session hashes written to disk. More importantly, the HMAC key is not persisted anywhere, so a process restart produces a different key and makes cross-restart session correlation impossible.

Suggested fix: derive the session hash key deterministically from a stable per-installation secret (e.g. `process.env.NEUROLINK_PROXY_SESSION_SECRET` or a machine-bound key stored in `~/.neurolink`), with a fallback to a hashed combination of `machine-id` + app path. If no stable secret is available, document that session hashes are only comparable within a single process lifetime.

```ts
import { createHash } from "node:crypto";
import { readFileSync } from "node:fs";

function resolveSessionHashKey(): Buffer {
  if (process.env.NEUROLINK_PROXY_SESSION_SECRET) {
    return Buffer.from(process.env.NEUROLINK_PROXY_SESSION_SECRET, "utf8");
  }
  try {
    const machineId = readFileSync("/etc/machine-id", "utf8").trim();
    return createHash("sha256").update(machineId).update(__filename).digest();
  } catch {
    return randomBytes(32);
  }
}
```

Tara-ag: ⚠️ **MAJOR**: The module-level mutable state (`loggerEnabled`, `lifecycleLogDir`, `queue`, counters, `flushTimer`, etc.) makes the lifecycle logger inherently non-thread-safe and hard to test concurrently. While Node.js is single-threaded, async interleaving between `logProxyLifecycleEvent`, `flushBatch`, and `configureProxyLifecycleLogger` can corrupt counters or drop events silently.

For example, `configureProxyLifecycleLogger` can reset `loggerEnabled` to `false` while a `flushBatch` is in flight; the in-flight batch still writes, but newly queued events are dropped, making the `dropped` counter inconsistent with actual behavior.

Suggested fix: serialize configuration changes and flush scheduling through a single mutex/queue (e.g. `Promise`-chain the flush work and await `flushInFlight` before reconfiguring), or document that `configureProxyLifecycleLogger` must only be called at startup and never concurrently with logging.

Tara-ag: 💡 **MINOR**: The `flushBatch` function groups lines by date and calls `appendFile` once per date. This is correct, but it does not fsync or ensure the file/directory metadata is durable. For a telemetry sink that needs crash safety, consider calling `fsync` on the appended file (or at least documenting that the sink trades durability for throughput).

Not blocking for this PR given the throughput target.

src/cli/commands/proxy.ts
```diff
@@ -1101,6 +1123,131 @@ async function resolveBootPrimaryAccountKey(
   return key;
```
Tara-ag: ⚠️ **MAJOR**: `registerProxyRequestTracking` emits `request_accepted` with `model: "-"`, `stream: false`, and `toolCount: 0` because the metadata is initialized before the route handler parses the request body. The lifecycle event therefore does not reflect the actual model, streaming mode, or tool count. Downstream analytics that aggregate by model or stream will be incorrect for the acceptance event.

Suggested fix: either defer `request_accepted` until after route parsing (if the proxy path populates `metadata.model`/`stream`/`toolCount` before `next()` returns), or accept the limitation and document that `request_accepted` uses placeholder values when the request body has not yet been inspected. If the route handler sets these fields on `metadata` during processing, the subsequent `response_headers`/`response_first_chunk`/`request_terminal` events will be correct, but the acceptance event will remain misleading.

src/lib/proxy/proxyActivity.ts
```diff
@@ -49,33 +61,70 @@ export function isProxyActivityQuiet(
 export function trackProxyResponse(
```
Tara-ag, line 70: 💡 **MINOR**: `safelyNotifyObserver` swallows all observer errors silently. While this is the right behavior for observability callbacks, consider logging the failure at `debug` level (guarded by `logger.shouldLog("debug")`) so that misbehaving observers can be diagnosed in development without affecting production response handling.

src/lib/server/routes/claudeProxyRoutes.ts
```diff
@@ -5289,22 +5293,28 @@ async function fetchAnthropicAccountResponse(args: {
       signal: AbortSignal.timeout(UPSTREAM_FETCH_TIMEOUT_MS),
```
Tara-ag, line 5321: ⚠️ **MAJOR**: The refactor of `fetchAnthropicAccountResponse` changes the control flow for non-retryable network errors. Previously, non-retryable errors were thrown immediately before `sawNetworkError = true` and `recordAttemptError`. Now the code sets `sawNetworkError = true`, records an attempt error, logs the attempt, and only then re-throws. This changes the observable side effects: `sawNetworkError` becomes `true` and `totalAttemptErrors` increments for non-retryable network errors that previously did not count as attempt errors.

Verify this is intentional. The new test `"logs non-retryable fetch failures before preserving terminal behavior"` asserts this new behavior, but existing callers or metrics dashboards may have assumed non-retryable network errors were not counted as attempt errors. If this is intentional, document the change in the PR description's compatibility notes.

test/proxyObservabilityFoundation.test.ts
```diff
@@ -0,0 +1,657 @@
...
+            controller.close();
+          },
+        }),
+        {
+          status: 200,
+          headers: {
+            "content-type": "text/event-stream",
+            "x-compatibility": "preserved",
+          },
+        },
```
Tara-ag, line 552: ⚠️ **MAJOR**: This test expectation appears inconsistent with `registerProxyRequestTracking` in `src/cli/commands/proxy.ts`. When the route handler throws, `await next()` rejects and control enters the `catch` block, which emits only a single `request_terminal` event with `terminalOutcome: "handler_error"` and then re-throws. The `response_headers` and `response_first_chunk` events are logged only on the success path after `await next()` returns.

Yet this test asserts four events (`request_accepted`, `response_headers`, `response_first_chunk`, `request_terminal`) and expects the terminal event to have `terminalOutcome: "completed"` and `errorType: "unhandled_proxy_error"`. That combination cannot be produced by the current middleware code for a thrown handler error.

Please reconcile the test with the intended behavior. If handler errors should indeed emit only `request_accepted` + `request_terminal(handler_error)`, update the test. If the intent is to emit the full four-event sequence for errors converted to responses by `app.onError`, the middleware needs to detect the error response and run `trackProxyResponse` on it instead of (or in addition to) the catch-block terminal event.

src/lib/proxy/requestLogger.ts
```diff
@@ -76,6 +77,7 @@ export function initRequestLogger(
 ): void {
```
Tara-ag, line 77: 💡 **MINOR**: `initRequestLogger` couples the lifecycle logger enablement to the request logger's `enabled` flag and `logDir`. If a caller wants request logging disabled but lifecycle metadata enabled (or vice versa), this coupling prevents independent configuration. Consider exposing separate configuration paths, or document that lifecycle logging is always tied to request logging.

Tara-ag (changes requested):
## Review summary

Reviewed all 10 changed files. Existing review comments from earlier rounds (by Tara-ag and copilot-pull-request-reviewer) appear resolved/outdated and are not duplicated here.

| Severity | Count | Files |
|----------|-------|-------|
| 🔒 CRITICAL | 1 | `src/lib/proxy/proxyLifecycle.ts` — per-process `sessionHashKey` is randomly generated and never persisted, allowing deanonymization if leaked and breaking cross-restart session correlation. |
| ⚠️ MAJOR | 4 | `proxyLifecycle.ts` module-level mutable state is racy across async reconfiguration/flushing; `proxy.ts` `request_accepted` emits placeholder `model`/`stream`/`toolCount`; `claudeProxyRoutes.ts` non-retryable network errors now increment attempt counters that previously did not; `test/proxyObservabilityFoundation.test.ts` handler-error test expectation appears inconsistent with the middleware's actual catch-branch behavior. |
| 💡 MINOR | 3 | `proxyActivity.ts` observer errors swallowed silently; `proxyLifecycle.ts` no fsync on batch writes; `requestLogger.ts` lifecycle enablement coupled to request-log enablement. |

### Blocking rationale

There is **1 CRITICAL** security/durability issue: the HMAC key used for session hashing is ephemeral per-process. If an attacker reads process memory or a core dump, all hashed session IDs can be reversed against the captured key, and a process restart silently breaks session correlation. This meets the blocking criterion for critical security risk.

### Most important fixes before merge

1. **Make the session hash key stable and confidential.** Derive it from `NEUROLINK_PROXY_SESSION_SECRET` or a machine-bound secret, or document that hashes are only comparable within one process lifetime.
2. **Reconcile the handler-error test with the middleware.** The current `catch` branch emits only `request_accepted` + `request_terminal(handler_error)`, but the test expects a four-event sequence ending with `terminalOutcome: "completed"`.
3. **Clarify the non-retryable network-error accounting change.** Confirm that incrementing `sawNetworkError`/`totalAttemptErrors` for non-retryable errors is intentional and update compatibility notes if so.

### What’s good

- Bounded queue, sequence IDs, and explicit drop counters are well-designed.
- Body-capture compatibility fixture and redaction assertions are thorough.
- `/status` exposure and graceful-shutdown flush are sensible.
- No CLAUDE.md critical-rule violations observed (dynamic imports, Gemini schema/tools, CLI/SDK separation, backward compatibility).

Please address the CRITICAL issue and reconcile the MAJOR inconsistencies, then re-request review.

src/lib/proxy/proxyLifecycle.ts
murdore: Fixed in 7ab3211d. Session hashes now use NEUROLINK_PROXY_SESSION_SECRET when configured, otherwise a random 32-byte installation key persisted in the lifecycle log directory with mode 0600. The restart-style reset regression verifies the same session ID hashes identically after logger reinitialization, and the operator behavior is documented.

murdore: Fixed in 7ab3211d. Every queued event now captures its enqueue-time log directory, pending and in-flight batches drain to that destination, and configuration changes apply atomically to new events. The new two-directory regression reconfigures before flushing and proves both batches are written without drops or counter corruption.

murdore: Documented the intentional durability tradeoff next to the append. This is best-effort high-throughput telemetry, so it does not fsync and keeps storage latency off the proxy path; writeDrops and writeFailures make any sink loss explicit. Full request and response capture behavior remains unchanged.

src/cli/commands/proxy.ts
murdore: Fixed in 7ab3211d. request_accepted now omits model, stream, and toolCount before route parsing instead of emitting false placeholders. Later lifecycle events carry the parsed values under the same request ID. The adapter regression asserts those fields are absent from acceptance.

src/lib/proxy/proxyActivity.ts
murdore: Fixed in 7ab3211d. Observer failures remain contained, but now emit a debug-level diagnostic through the existing logger when debug logging is enabled. The response-delivery regression continues to prove throwing observers cannot interrupt the client body.

src/lib/server/routes/claudeProxyRoutes.ts
murdore: Confirmed intentional and documented in the PR compatibility section. Every request actually dispatched upstream is counted once when transport fails, including terminal non-retryable failures; the existing terminal throw behavior is preserved. Focused tests cover both retryable and non-retryable accounting and low-level error codes.

test/proxyObservabilityFoundation.test.ts
murdore: Verified against the real Hono app after rebasing: this assumption is not correct for the configured adapter. Hono onError converts the thrown handler failure to the 502 response before next returns, so the middleware observes response headers, first chunk, and terminal completion. The integration test passes with the four-event sequence and correlates the same request ID and EADDRNOTAVAIL code. I retained the verified expectation.

src/lib/proxy/requestLogger.ts
murdore: Documented the coupling as intentional. Lifecycle metadata shares the request logger enablement, directory permissions, retention policy, and operator privacy boundary so disabling full proxy logging also disables derived lifecycle persistence. This avoids introducing a second independent persistence control in this compatibility-focused PR.

Tara-ag (approved):
## Review Summary

**Files reviewed:** 11
**New issues raised this run:** 0 (all prior review threads were already resolved or addressed)
**Decision:** Approve

I walked through each changed file against the project standards and blocking criteria:

- **Security:** No hardcoded secrets, no injection/SSRF/eval risks, and session IDs are hashed via HMAC-SHA256 before persistence. Lifecycle records intentionally exclude request/response bodies and credentials.
- **CLAUDE.md critical rules:** No static provider imports, no Gemini tools/schema conflict, no SDK/CLI leakage, `formatProviderError` is untouched, and no public SDK API breakage (all changes are internal proxy/server types).
- **Type safety / correctness:** New types live in `src/lib/types/proxy.ts` with domain-prefixed names (`Proxy*`), use `type` aliases, and are imported via the barrel. Optional `errorCode` fields preserve backward compatibility.
- **Architecture:** The lifecycle sink correctly captures enqueue-time `logDir`, serializes flushes through `flushInFlight`, eagerly creates the log directory with `0o700`, and degrades gracefully with explicit loss counters. Observer callbacks are contained via `safelyNotifyObserver`.
- **Tests / docs:** New Vitest suites cover compatibility, bounded-queue loss, reconfiguration, session-hash stability, write failures, and observer fault containment. Docs are updated for `NEUROLINK_PROXY_SESSION_SECRET`.

All previously raised review points (directory creation, non-throwing hot path, stable session hashing, reconfiguration atomicity, `observedBodyBytes` naming, observer containment, attempt-error accounting, and test determinism) appear to have been addressed in the latest revision.

No blocking criteria are met; this PR is ready to merge.

src/lib/proxy/proxyActivity.ts
```diff
@@ -49,33 +67,70 @@ export function isProxyActivityQuiet(
...
+        settle("stream_error");
         controller.error(error);
       }
     },
     async cancel(reason) {
       try {
         await reader.cancel(reason);
       } finally {
-        finishRequest();
+        settle("client_cancelled");
```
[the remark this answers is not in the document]
murdore: Fixed in 65160076. Client cancellation now settles the request and emits its terminal lifecycle event before upstream cleanup begins. The reader cancellation is bounded to 1 second with the shared withTimeout utility, and the new regression proves settlement occurs while cleanup is still pending.

src/lib/proxy/proxyLifecycle.ts
```diff
@@ -0,0 +1,374 @@
...
+      lines.push(`${JSON.stringify(item.record)}\n`);
+      byPath.set(path, lines);
+    }
+
+    for (const [path, lines] of byPath) {
+      try {
+        // This best-effort telemetry sink intentionally avoids fsync so request
+        // throughput is not coupled to storage latency. Loss is surfaced by
+        // writeDrops/writeFailures rather than delaying proxy responses.
+        await appendFile(path, lines.join(""), { mode: 0o600 });
```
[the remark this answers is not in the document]
murdore: Fixed in 65160076. Every lifecycle append is bounded to 2 seconds and any timeout follows the existing write-drop/write-failure accounting path. Shutdown now applies a separate 5-second overall bound to draining all pending lifecycle batches, so a stalled filesystem cannot hold proxy termination indefinitely.

src/lib/server/routes/claudeProxyRoutes.ts
```diff
@@ -4403,7 +4403,10 @@ async function handleAnthropicAuthRetry(args: {
           : String(retryFetchErr);
       authRetryError = `network error on retry ${authRetry + 1}: ${message}`;
       currentLastError = authRetryError;
-      retryLogAttempt(502, "network_error", message, { retryable: true });
+      retryLogAttempt(502, "network_error", message, {
+        retryable: true,
+        errorCode: getErrorCode(retryFetchErr) ?? "unknown",
+      });
```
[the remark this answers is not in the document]
murdore: Fixed in 65160076. OAuth retry fetch failures now use isRetryableNetworkError instead of a hard-coded true value. The regression drives a refreshed OAuth retry into ERR_INVALID_URL and verifies the structured attempt is recorded with retryable=false while preserving rotation behavior.

test/proxyObservabilityFoundation.test.ts
```diff
@@ -0,0 +1,724 @@
...
+    const secondHash = hashProxyLifecycleSessionId("private-session-id");
+
+    expect(secondHash).toBe(firstHash);
+    expect(secondHash).not.toContain("private-session-id");
+    expect(
+      permissionBits(
+        (await stat(join(logDir, ".proxy-lifecycle-session-key"))).mode,
+      ),
+    ).toBe("600");
+  });
```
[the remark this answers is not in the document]
murdore: Fixed in 65160076. The persisted-key regression now saves the existing NEUROLINK_PROXY_SESSION_SECRET value, removes it for the test, and restores it in a finally block. The key-file creation and mode-0600 assertions are now independent of the test runner environment.

Tara-ag (approved):
## Review Summary

Reviewed all 11 changed files in this PR. **No new blocking issues found.**

### What was checked

- **Security**: No hardcoded secrets, API keys, or credentials. Session hash key is persisted with mode `0600` and derived from `NEUROLINK_PROXY_SESSION_SECRET` when configured.
- **CLAUDE.md Critical Rules**: No violations of the non-negotiable architectural rules enforced outside ESLint. Changes are confined to CLI/proxy internals; no static provider imports, no Gemini/tools+schema conflict, no `formatProviderError` throw path, and no public SDK API breakage.
- **Correctness & Type Safety**: Lifecycle logger validates timestamps, clips short fields, sanitizes numeric inputs, and catches malformed runtime metadata without throwing on the hot path. Response tracking settles terminal outcomes exactly once and bounds `reader.cancel()` with `withTimeout`.
- **Concurrency & Durability**: Queued events capture their enqueue-time `logDir`, so reconfiguration drains pending batches to the original destination. Flush work is serialized via `flushInFlight`, and both per-append and shutdown-flush operations are timeout-bounded.
- **Existing Comments**: All prior inline review threads (from Tara-ag, copilot-pull-request-reviewer, and coderabbitai) have been addressed by the author with corresponding code changes and regression tests.

### New issue counts

- CRITICAL: 0
- MAJOR: 0
- MINOR: 0
- SUGGESTION: 0

### Decision

Approve. The telemetry foundation is defensive, bounded, and non-blocking for client responses.

PR #1190, opened 2026-07-17
fix(image): guard image buffer size to prevent memory exhaustion (#257)

### LINKED ISSUE #257: IMG-001: Memory Exhaustion on Large Images
**Summary:** Image processor doesn't validate buffer size before base64 conversion, allowing 100MB+ images to exhaust heap memory and crash the process.

**Root Cause:** Direct calls to `Buffer.toString("base64")` without size validation at lines 27, 60, 95, 135, 393, 470, 512 in `src/lib/utils/imageProcessor.ts`. A 100MB image creates ~133MB base64 string causing `FATAL ERROR: Reached heap limit`.

**Impact:** Production crashes, denial of service attacks possible, poor error messages before crash.

**Fix Approach:**
- Add MAX_IMAGE_SIZE constant (10MB default)
- Implement validateBufferSize() method with configurable limit
- Implement safeBase64Convert() that validates before conversion
- Update all 7 locations calling Buffer.toString('base64')

**Acceptance Criteria:**
- [ ] Add MAX_IMAGE_SIZE constant (10MB)
- [ ] Implement validateBufferSize() and safeBase64Convert() methods
- [ ] Update all 7 locations calling Buffer.toString('base64')
- [ ] Add tests for at limit, over limit, edge cases
- [ ] Verify descriptive error messages for oversized images

**Files:** `src/lib/utils/imageProcessor.ts`

murdore:
## What & why

Fixes #257 (IMG-001) — **memory exhaustion on large images**. Several `Buffer.toString("base64")` sites converted caller-supplied image buffers with **no size check**, so a large image could exhaust process memory (base64 allocates ~1.33× the buffer).

## Fix

`ImageProcessor` gains two helpers:
- `validateBufferSize(buffer, context, maxSize?)` — throws a descriptive error above the limit.
- `safeBase64Convert(buffer, context, maxSize?)` — validates, then converts.

Default limit is the canonical `SIZE_LIMITS_BYTES.IMAGE_MAX` (**10 MB**), overridable via `maxSize`. The **6 previously-unguarded sites** now route through it: `process()`, the OpenAI/Google/Anthropic Buffer branches, the `processImage()` default case, and `imageUtils.bufferToBase64`. (`fileToBase64DataUri` / `urlToBase64DataUri` were already guarded — left as-is.)

## Enhancement gap → closed

The **sibling** `generate({ images })` path (`messageBuilder`, hit by the CLI `--image` flag) had the *identical* unguarded conversion — fixing only `imageProcessor` would leave it exploitable. So:
- `convertFilePathToBase64` now **stats the file size before reading it into memory**.
- the Buffer branch of `processImageToBase64` guards before converting.

Both entry points a user hits are now covered.

## Docs gap → closed

`docs/features/file-processors.md`'s "File Size Limits" table claimed **Images: 20 MB** and showed a `ProcessorConfig.setLimits({...})` snippet **that doesn't exist in the codebase**. Corrected to the real enforced **10 MB** and replaced the fictional API with an accurate description of the enforcement + override.

## Testing / proof

New bugfixes-suite tests (deterministic, no API keys):
- `FileDetector.detectAndProcess(<11 MB image>)` throws a size error; an image **exactly at 10 MB still succeeds** (no false rejection).
- `buildMultimodalMessagesArray({ images: [<11 MB Buffer>] })` throws — covering the sibling `images` path.

`pnpm run check` ✅ (0 errors) · `pnpm run lint` ✅ (0 errors) · `pnpm run build` ✅ · `prettier --check` ✅.

## Risk
Low — the guard only rejects previously-uncaught oversized buffers (a hard crash today) with a clean error; in-range images behave identically. `maxSize` remains an escape hatch for hi-res power users. No new constant (reuses `SIZE_LIMITS_BYTES.IMAGE_MAX`).

Fixes #257.

## Summary by CodeRabbit

- **New Features**
  - Enforced a fixed 10 MB image limit across file, buffer, URL, and image inputs.
  - Added validation for malformed image-size values with a dedicated, non-retriable error.
- **Bug Fixes**
  - Oversized images now fail fast with descriptive errors before base64 conversion/allocation.
  - Preserved image ordering during multi-image conversions by processing sequentially.
  - Improved error classification for missing files vs other filesystem issues, while keeping typed errors intact.
- **Documentation**
  - Updated image size-limit guidance to reflect the fixed 10 MB behavior.
- **Tests**
  - Expanded regression coverage for image-size enforcement, TOCTOU revalidation, and error-path correctness.

### DIFF
--- a/docs/features/file-processors.md
+++ b/docs/features/file-processors.md
@@ -354,23 +354,15 @@ const result = await neurolink.generate({
 
 Default size limits prevent denial-of-service attacks:
 
-| Category     | Default Limit | Configurable |
-| ------------ | ------------- | ------------ |
-| Documents    | 50 MB         | Yes          |
-| Data files   | 10 MB         | Yes          |
-| Code files   | 5 MB          | Yes          |
-| Config files | 1 MB          | Yes          |
-| Images       | 20 MB         | Yes          |
+| Category     | Default Limit | Configurable     |
+| ------------ | ------------- | ---------------- |
+| Documents    | 50 MB         | Yes              |
+| Data files   | 10 MB         | Yes              |
+| Code files   | 5 MB          | Yes              |
+| Config files | 1 MB          | Yes              |
+| Images       | 10 MB         | No — fixed limit |
 
-```typescript
-import { ProcessorConfig } from "@juspay/neurolink";
-
-// Configure size limits
-ProcessorConfig.setLimits({
-  maxDocumentSize: 100 * 1024 * 1024, // 100 MB
-  maxCodeSize: 10 * 1024 * 1024, // 10 MB
-});
-```
+The **image** limit (`SIZE_LIMITS_BYTES.IMAGE_MAX`, 10 MB) is enforced: an oversized image buffer, file, or download throws a descriptive error before any base64 conversion, so a large image can no longer exhaust process memory. This applies to both entry points — images passed via `generate({ files })` (routed through `FileDetector` → `ImageProcessor`) and via `generate({ images })` (routed through the message builder). The internal image helpers that convert buffers/files/URLs accept an optional size override, but it is not exposed through `generate()` or any other public API — the 10 MB limit is fixed for SDK callers.
 
 ## Error Handling
 
--- a/src/lib/utils/imageProcessor.ts
+++ b/src/lib/utils/imageProcessor.ts
@@ -7,7 +7,9 @@ import { logger } from "./logger.js";
 import { urlDownloadRateLimiter } from "./rateLimiter.js";
 import { withRetry } from "./retryHandler.js";
 import { SYSTEM_LIMITS } from "../core/constants.js";
+import { SIZE_LIMITS_BYTES } from "../processors/config/sizeLimits.js";
 import { getImageCache } from "./imageCache.js";
+import { ErrorFactory, NeuroLinkError } from "./errorHandling.js";
 import type { ProcessedImage, FileProcessingResult } from "../types/index.js";
 
 /**
@@ -95,7 +97,10 @@ export class ImageProcessor {
     }
 
     const mediaType = this.detectImageType(content);
-    const base64 = content.toString("base64");
+    const base64 = ImageProcessor.safeBase64Convert(
+      content,
+      "image processing",
+    );
     const dataUri = `data:${mediaType};base64,${base64}`;
 
     // Validate output before returning
@@ -171,9 +176,17 @@ export class ImageProcessor {
       }
 
       // Handle Buffer - convert to data URI
-      const base64 = image.toString("base64");
+      const base64 = ImageProcessor.safeBase64Convert(
+        image,
+        "OpenAI image input",
+      );
       return `data:image/jpeg;base64,${base64}`;
     } catch (error) {
+      // Preserve typed errors (e.g. IMAGE_TOO_LARGE) as-is — don't flatten
+      // them into a generic Error and lose the error `code` for callers.
+      if (error instanceof NeuroLinkError) {
+        throw error;
+      }
       logger.error("Failed to process image for OpenAI:", error);
       throw new Error(
         `Image processing failed for OpenAI: ${error instanceof Error ? error.message : "Unknown error"}`,
@@ -207,7 +220,10 @@ export class ImageProcessor {
           base64Data = image;
         }
       } else {
-        base64Data = image.toString("base64");
+        base64Data = ImageProcessor.safeBase64Convert(
+          image,
+          "provider image input",
+        );
       }
 
       return {
@@ -215,6 +231,11 @@ export class ImageProcessor {
         data: base64Data, // Google wants base64 WITHOUT data URI prefix
       };
     } catch (error) {
+      // Preserve typed errors (e.g. IMAGE_TOO_LARGE) as-is — don't flatten
+      // them into a generic Error and lose the error `code` for callers.
+      if (error instanceof NeuroLinkError) {
+        throw error;
+      }
       logger.error("Failed to process image for Google AI:", error);
       throw new Error(
         `Image processing failed for Google AI: ${error instanceof Error ? error.message : "Unknown error"}`,
@@ -248,7 +269,10 @@ export class ImageProcessor {
           base64Data = image;
         }
       } else {
-        base64Data = image.toString("base64");
+        base64Data = ImageProcessor.safeBase64Convert(
+          image,
+          "provider image input",
+        );
       }
 
       return {
@@ -256,6 +280,11 @@ export class ImageProcessor {
         data: base64Data, // Anthropic wants base64 WITHOUT data URI prefix
       };
     } catch (error) {
+      // Preserve typed errors (e.g. IMAGE_TOO_LARGE) as-is — don't flatten
+      // them into a generic Error and lose the error `code` for callers.
+      if (error instanceof NeuroLinkError) {
+        throw error;
+      }
       logger.error("Failed to process image for Anthropic:", error);
       throw new Error(
         `Image processing failed for Anthropic: ${error instanceof Error ? error.message : "Unknown error"}`,
@@ -284,6 +313,11 @@ export class ImageProcessor {
         return ImageProcessor.processImageForGoogle(image);
       }
     } catch (error) {
+      // Preserve typed errors bubbling up from processImageForGoogle/
+      // processImageForAnthropic (e.g. IMAGE_TOO_LARGE) as-is.
+      if (error instanceof NeuroLinkError) {
+        throw error;
+      }
       logger.error("Failed to process image for Vertex AI:", error);
       throw new Error(
         `Image processing failed for Vertex AI: ${error instanceof Error ? error.message : "Unknown error"}`,
@@ -385,22 +419,81 @@ export class ImageProcessor {
   }
 
   /**
-   * Validate image size (default 10MB limit)
+   * Throwing size guard for a raw byte length. Prevents memory exhaustion by
+   * rejecting an oversized input BEFORE an unbounded read/allocation happens
+   * (#257) — callers that can get a size cheaply (e.g. `fs.stat`) should
+   * validate it before ever reading the bytes into memory. Default limit is
+   * the canonical `SIZE_LIMITS_BYTES.IMAGE_MAX` (10 MB); there is no public
+   * way to raise it — `maxSize` is an internal-only override for advanced
+   * callers (e.g. video image inputs use a higher limit).
+   *
+   * @param size - Byte length to validate
+   * @param context - Short label identifying the source (unused in the
+   *   thrown message today, kept for call-site readability and future use)
+   * @param maxSize - Max allowed size in bytes
+   * @throws NeuroLinkError (`INVALID_IMAGE_SIZE`) if `size` is not a finite,
+   *   non-negative number
+   * @throws NeuroLinkError (`IMAGE_TOO_LARGE`) if `size` exceeds `maxSize`
    */
-  static validateImageSize(
-    data: Buffer | string,
-    maxSize: number = 10 * 1024 * 1024,
-  ): boolean {
-    try {
-      const size =
-        typeof data === "string"
-          ? Buffer.byteLength(data, "base64")
-          : data.length;
-      return size <= maxSize;
-    } catch (error) {
-      logger.warn("Failed to validate image size:", error);
-      return false;
+  static validateSize(
+    size: number,
+    context: string,
+    maxSize: number = SIZE_LIMITS_BYTES.IMAGE_MAX,
+  ): void {
+    // Reject a malformed maxSize before it's ever compared: `size > NaN` and
+    // `size > Infinity` (for a negative maxSize) can both evaluate `false`,
+    // which would let a corrupted/negative limit bypass the guard entirely.
+    if (!Number.isFinite(maxSize) || maxSize < 0) {
+      throw ErrorFactory.invalidConfiguration(
+        "maxSize",
+        "must be a finite, non-negative byte count",
+        { maxSize, context },
+      );
     }
+    // Fail closed on malformed sizes first: `NaN > maxSize` and
+    // `-1 > maxSize` both evaluate to `false`, which would otherwise let a
+    // corrupted stat/header value silently skip the guard below.
+    if (!Number.isFinite(size) || size < 0) {
+      throw ErrorFactory.invalidImageSize(size);
+    }
+    if (size > maxSize) {
+      const mb = (size / (1024 * 1024)).toFixed(1);
+      const maxMb = (maxSize / (1024 * 1024)).toFixed(0);
+      throw ErrorFactory.imageTooLarge(mb, maxMb);
+    }
+  }
+
+  /**
+   * Throwing size guard for image buffers. Prevents memory exhaustion by
+   * rejecting oversized buffers with a descriptive error BEFORE an unbounded
+   * `Buffer.toString("base64")` allocation (#257). Delegates to
+   * `validateSize` so buffer-based and pre-read (stat-based) callers share
+   * one implementation.
+   *
+   * @param buffer - Image buffer to validate
+   * @param context - Short label identifying the source (see `validateSize`)
+   * @param maxSize - Max allowed size in bytes
+   * @throws NeuroLinkError (`IMAGE_TOO_LARGE`) if the buffer exceeds `maxSize`
+   */
+  static validateBufferSize(
+    buffer: Buffer,
+    context: string,
+    maxSize: number = SIZE_LIMITS_BYTES.IMAGE_MAX,
+  ): void {
+    ImageProcessor.validateSize(buffer.length, context, maxSize);
+  }
+
+  /**
+   * Size-guarded Buffer → base64. Use in place of `buffer.toString("base64")`
+   * on any path that converts a caller-supplied image buffer (#257).
+   */
+  static safeBase64Convert(
+    buffer: Buffer,
+    context: string,
+    maxSize: number = SIZE_LIMITS_BYTES.IMAGE_MAX,
+  ): string {
+    ImageProcessor.validateBufferSize(buffer, context, maxSize);
+    return buffer.toString("base64");
   }
 
   /**
@@ -564,7 +657,7 @@ export class ImageProcessor {
               ? image.split(",")[1] || image
               : image;
           } else {
-            data = image.toString("base64");
+            data = ImageProcessor.safeBase64Convert(image, "image input");
           }
           format = "base64";
       }
@@ -576,6 +669,11 @@ export class ImageProcessor {
         format,
       };
     } catch (error) {
+      // Preserve typed errors (e.g. IMAGE_TOO_LARGE) bubbling up from the
+      // provider-specific branches above as-is.
+      if (error instanceof NeuroLinkError) {
+        throw error;
+      }
       logger.error(`Failed to process image for ${provider}:`, error);
       throw new Error(
         `Image processing failed: ${error instanceof Error ? error.message : "Unknown error"}`,
@@ -708,10 +806,23 @@ export const imageUtils = {
   },
 
   /**
-   * Convert Buffer to base64 string
+   * Convert Buffer to base64 string. Guarded by the same #257 size limit as
+   * every other conversion site (default `SIZE_LIMITS_BYTES.IMAGE_MAX`);
+   * pass `maxBytes` to override for a caller with a legitimate need for a
+   * different ceiling rather than silently accepting an unbounded buffer.
+   * `imageUtils`/`ImageProcessor` are internal-only — not re-exported from
+   * `src/lib/index.ts` or the package's public `exports` map — so this
+   * default does not change behavior for any external SDK consumer.
    */
-  bufferToBase64: (buffer: Buffer): string => {
-    return buffer.toString("base64");
+  bufferToBase64: (
+    buffer: Buffer,
+    maxBytes: number = SIZE_LIMITS_BYTES.IMAGE_MAX,
+  ): string => {
+    return ImageProcessor.safeBase64Convert(
+      buffer,
+      "buffer-to-base64",
+      maxBytes,
+    );
   },
 
   /**
--- a/src/lib/utils/messageBuilder.ts
+++ b/src/lib/utils/messageBuilder.ts
@@ -1,4 +1,5 @@
 import { existsSync, readFileSync, statSync } from "fs";
+import { readFile as readFileAsync, stat as statAsync } from "fs/promises";
 import { getGlobalDispatcher, interceptors, request } from "undici";
 import {
   MultimodalLogger,
@@ -28,8 +29,10 @@ import type {
   TextGenerationOptions,
 } from "../types/index.js";
 import { tracers, ATTR, withSpan } from "../telemetry/index.js";
+import { NeuroLinkError } from "./errorHandling.js";
 import { FileDetector } from "./fileDetector.js";
 import { getImageCache } from "./imageCache.js";
+import { ImageProcessor } from "./imageProcessor.js";
 import { logger } from "./logger.js";
 import { PDFImageConverter, PDFProcessor } from "./pdfProcessor.js";
 import { urlDownloadRateLimiter } from "./rateLimiter.js";
@@ -1800,14 +1803,42 @@ function detectMimeTypeFromBuffer(buffer: Buffer): string | undefined {
 /**
  * Convert file path to raw base64 string.
  * Returns raw base64 (not a data: URI) to avoid SSRF validation in AI SDK v6.
+ * Uses async fs so a large/slow-filesystem image never blocks the event loop
+ * while the size guard (below) or the read itself is in flight.
  */
-function convertFilePathToBase64(filePath: string): string {
-  if (!existsSync(filePath)) {
-    throw new Error(`Image file not found: ${filePath}`);
+async function convertFilePathToBase64(filePath: string): Promise<string> {
+  let stats;
+  try {
+    stats = await statAsync(filePath);
+  } catch (error) {
+    // Only ENOENT means "not found" — EACCES/ELOOP/etc. are real access or
+    // filesystem problems that must not be misreported as a missing file.
+    const code =
+      error instanceof Error
+        ? (error as NodeJS.ErrnoException).code
+        : undefined;
+    if (code === "ENOENT") {
+      throw new Error(`Image file not found: ${filePath}`, { cause: error });
+    }
+    throw new Error(
+      `Cannot access image file ${filePath}: ${error instanceof Error ? error.message : String(error)}`,
+      { cause: error },
+    );
+  }
+  if (!stats.isFile()) {
+    throw new Error(`Image path is not a file: ${filePath}`);
   }
 
-  const buffer = readFileSync(filePath);
-  return buffer.toString("base64");
+  // #257: check the file size via stat BEFORE reading it into memory, so a
+  // huge image path never triggers an unbounded read + base64 allocation.
+  // Delegates to ImageProcessor's shared guard (no local reimplementation).
+  const context = `image file "${filePath}"`;
+  ImageProcessor.validateSize(stats.size, context);
+  const buffer = await readFileAsync(filePath);
+  // TOCTOU: the file can grow, or a symlink can be swapped, between the stat
+  // above and this read completing. Re-validate the bytes actually read
+  // (not just the pre-read stat) before the base64 allocation.
+  return ImageProcessor.safeBase64Convert(buffer, context);
 }
 
 /**
@@ -1819,10 +1850,10 @@ function convertFilePathToBase64(filePath: string): string {
  * Passing raw base64 avoids this because `new URL(base64string)` throws and
  * the SDK treats the string as inline base64 data instead.
  */
-function processImageToBase64(
+async function processImageToBase64(
   image: Buffer | string,
   index: number,
-): { imageData: string; mimeType: string } {
+): Promise<{ imageData: string; mimeType: string }> {
   let imageData: string;
   let mimeType = "image/jpeg"; // Default mime type
 
@@ -1855,13 +1886,18 @@ function processImageToBase64(
     } else {
       // File path string - convert to raw base64
       try {
-        imageData = convertFilePathToBase64(image);
+        imageData = await convertFilePathToBase64(image);
         mimeType = getMimeTypeFromExtension(image);
       } catch (error) {
         MultimodalLogger.logError("FILE_PATH_CONVERSION", error as Error, {
           index,
           filePath: image,
         });
+        // Preserve typed errors (e.g. IMAGE_TOO_LARGE) as-is — don't flatten
+        // them into a generic Error and lose the error `code` for callers.
+        if (error instanceof NeuroLinkError) {
+          throw error;
+        }
         throw new Error(
           `Failed to convert file path to base64: ${image}. ${error}`,
           { cause: error },
@@ -1874,6 +1910,9 @@ function processImageToBase64(
     if (detectedMimeType) {
       mimeType = detectedMimeType;
     }
+    // #257: guard the buffer size before the unbounded base64 conversion.
+    // Delegates to ImageProcessor's shared guard (no local reimplementation).
+    ImageProcessor.validateBufferSize(image, `image input at index ${index}`);
     imageData = image.toString("base64");
   }
 
@@ -1957,11 +1996,13 @@ async function convertSimpleImagesToProviderFormat(
     { type: "text", text: enhancedText },
   ];
 
-  // Process all images (including downloaded URLs) for Vercel AI SDK
-  actualImages.forEach(({ data: image }, index) => {
+  // Process all images (including downloaded URLs) for Vercel AI SDK.
+  // Sequential for...of (not Promise.all) to preserve image ordering and
+  // keep the original forEach's one-at-a-time error semantics.
+  for (const [index, { data: image }] of actualImages.entries()) {
     try {
       // Use helper function to process image and reduce nesting depth
-      const { imageData, mimeType } = processImageToBase64(image, index);
+      const { imageData, mimeType } = await processImageToBase64(image, index);
 
       content.push({
         type: "image" as const,
@@ -1975,7 +2016,7 @@ async function convertSimpleImagesToProviderFormat(
       });
       throw error;
     }
-  });
+  }
 
   return content;
 }
--- a/test/continuous-test-suite-bugfixes.ts
+++ b/test/continuous-test-suite-bugfixes.ts
@@ -35,10 +35,14 @@ import type { CSVLoaderOptions } from "../src/lib/types/index.js";
 import iconv from "iconv-lite";
 import { Readable, Transform } from "node:stream";
 import { PDFProcessor } from "../src/lib/utils/pdfProcessor.js";
+import { SIZE_LIMITS_BYTES } from "../src/lib/processors/config/sizeLimits.js";
 import { directAgentTools } from "../src/lib/agent/directTools.js";
 import { isMultimodalInput } from "../src/lib/types/index.js";
 import { FileDetector } from "../src/lib/utils/fileDetector.js";
+import { ImageProcessor, imageUtils } from "../src/lib/utils/imageProcessor.js";
+import { ERROR_CODES } from "../src/lib/utils/errorHandling.js";
 import {
+  chmodSync,
   mkdtempSync,
   readFileSync,
   rmSync,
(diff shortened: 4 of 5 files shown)


src/lib/utils/imageProcessor.ts
```diff
@@ -403,6 +416,45 @@ export class ImageProcessor {
...
+    context: string,
+    maxSize: number = SIZE_LIMITS_BYTES.IMAGE_MAX,
+  ): void {
+    if (buffer.length > maxSize) {
+      const mb = (buffer.length / (1024 * 1024)).toFixed(1);
+      const maxMb = (maxSize / (1024 * 1024)).toFixed(0);
+      throw new Error(
+        `Image too large for ${context}: ${mb} MB exceeds the ${maxMb} MB limit. ` +
+          `Resize the image or raise the limit via the maxSize option.`,
+      );
```
Tara-ag: ⚠️ **MAJOR**: The error message tells callers to "raise the limit via the maxSize option", but the public entry points that route through `ImageProcessor` (`FileDetector.detectAndProcess`, `generate({ files })`, `generate({ images })`) do not expose a `maxSize` override. This makes the message misleading for SDK users and contradicts the docs claim that the limit is configurable "Via `maxSize` / `maxBytes` param".

Either expose the override through the public API, or reword the message to avoid implying an unavailable knob:
```ts
throw new Error(
  `Image too large for ${context}: ${mb} MB exceeds the ${maxMb} MB limit. Resize the image before sending it.`
);
```

src/lib/utils/imageProcessor.ts
```diff
@@ -403,6 +416,45 @@ export class ImageProcessor {
...
+   * @param context - Short label for the error message (e.g. "OpenAI image input")
+   * @param maxSize - Max allowed size in bytes
+   * @throws Error if the buffer exceeds `maxSize`
+   */
+  static validateBufferSize(
+    buffer: Buffer,
+    context: string,
+    maxSize: number = SIZE_LIMITS_BYTES.IMAGE_MAX,
+  ): void {
+    if (buffer.length > maxSize) {
```
Tara-ag: 💡 **MINOR**: The new size guard throws a plain `Error` instead of the typed `ErrorFactory.imageTooLarge(...)` helper that already exists in this codebase. Using `ErrorFactory.imageTooLarge(...)` would give callers a stable `code === IMAGE_TOO_LARGE`, `category === VALIDATION`, and `retriable === false`, which retry policies and SDK consumers rely on.

Consider replacing the raw throw with:
```ts
import { ErrorFactory } from "./errorHandling.js";
// ...
throw ErrorFactory.imageTooLarge(mb, maxMb);
```

Note: the existing `ErrorFactory.imageTooLarge` signature takes string args, so pass the formatted values.

src/lib/utils/messageBuilder.ts
```diff
@@ -1801,11 +1802,28 @@ function detectMimeTypeFromBuffer(buffer: Buffer): string | undefined {
...
+ */
+function assertImageSize(size: number, label: string): void {
+  if (size > SIZE_LIMITS_BYTES.IMAGE_MAX) {
+    const mb = (size / (1024 * 1024)).toFixed(1);
+    const maxMb = (SIZE_LIMITS_BYTES.IMAGE_MAX / (1024 * 1024)).toFixed(0);
+    throw new Error(
+      `Image too large for ${label}: ${mb} MB exceeds the ${maxMb} MB limit.`,
+    );
+  }
+}
```
Tara-ag: 💡 **MINOR**: `assertImageSize` duplicates the size-checking logic from `ImageProcessor.validateBufferSize` almost verbatim. The comment says it is "kept local to avoid a module cycle", but `messageBuilder.ts` already imports from `../processors/config/sizeLimits.js`, and `imageProcessor.ts` imports from the same file plus `logger.js`/`retryHandler.js` — none of which depend on `messageBuilder.ts`.

Please verify that importing `ImageProcessor.validateBufferSize` (or a shared helper in `sizeLimits.js`) would actually create a cycle. If it would not, deduplicate to keep behavior consistent. If it genuinely would, a brief inline note explaining the cycle would help future maintainers.

src/lib/utils/messageBuilder.ts
```diff
@@ -1801,11 +1802,28 @@ function detectMimeTypeFromBuffer(buffer: Buffer): string | undefined {
...
+
 function convertFilePathToBase64(filePath: string): string {
   if (!existsSync(filePath)) {
     throw new Error(`Image file not found: ${filePath}`);
   }
 
+  // #257: check the file size via stat BEFORE reading it into memory, so a
+  // huge image path never triggers an unbounded read + base64 allocation.
+  assertImageSize(statSync(filePath).size, `image file "${filePath}"`);
   const buffer = readFileSync(filePath);
```
Tara-ag: ⚠️ **MAJOR**: `convertFilePathToBase64` performs a synchronous `statSync` + `readFileSync` on a caller-provided path. While the size guard prevents memory exhaustion, the **synchronous** read still blocks the event loop for the entire file-I/O duration. On a 10 MB file this is minor, but on a path served from a slow filesystem or network share it stalls all concurrent requests.

The surrounding `convertSimpleImagesToProviderFormat` is `async`, so this can be made non-blocking without API changes:
```ts
import { promises as fs } from "fs";

async function convertFilePathToBase64(filePath: string): Promise<string> {
  const stats = await fs.stat(filePath).catch(() => null);
  if (!stats?.isFile()) {
    throw new Error(`Image file not found: ${filePath}`);
  }
  assertImageSize(stats.size, `image file "${filePath}"`);
  const buffer = await fs.readFile(filePath);
  return buffer.toString("base64");
}
```

Then propagate the `Promise` through `processImageToBase64` and `convertSimpleImagesToProviderFormat`.

test/continuous-test-suite-bugfixes.ts
```diff
@@ -3850,6 +3850,61 @@ exit 127
         },
       ),
   },
+  // ---------- #257: image buffer size guard (memory exhaustion) ----------
+  {
+    name: "FileDetector #257: oversized image rejected before base64; at-limit passes",
+    category: "image-processor",
+    fn: async () => {
+      const IMAGE_MAX = 10 * 1024 * 1024;
```
Tara-ag: 💡 **MINOR**: Hard-codes `IMAGE_MAX = 10 * 1024 * 1024` instead of importing `SIZE_LIMITS_BYTES.IMAGE_MAX`. If the canonical limit changes, this test will silently drift from the enforced value and either falsely reject at-limit images or miss a raised limit. Import and use the constant:
```ts
import { SIZE_LIMITS_BYTES } from "../src/lib/processors/config/sizeLimits.js";
// ...
const IMAGE_MAX = SIZE_LIMITS_BYTES.IMAGE_MAX;
```

test/continuous-test-suite-bugfixes.ts
```diff
@@ -3850,6 +3850,61 @@ exit 127
...
+    },
+  },
+  {
+    name: "buildMultimodalMessagesArray #257: oversized image Buffer rejected (images path)",
+    category: "message-builder",
+    fn: async () => {
+      // The generate({ images: [...] }) path — distinct from files→FileDetector.
+      const oversized = Buffer.alloc(11 * 1024 * 1024);
+      oversized[0] = 0x89; // PNG magic so MIME detection succeeds
+      oversized[1] = 0x50;
```
Tara-ag: 💡 **MINOR**: The oversized-buffer test only writes 2 bytes of the PNG signature (`0x89`, `0x50`) and hard-codes `11 * 1024 * 1024`. Use the exported `SIZE_LIMITS_BYTES.IMAGE_MAX` for the limit and write the full 8-byte PNG header so the test accurately mirrors the detection logic and stays aligned with the real limit:
```ts
import { SIZE_LIMITS_BYTES } from "../src/lib/processors/config/sizeLimits.js";
const pngHeader = Buffer.from([0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a]);
const oversized = Buffer.alloc(SIZE_LIMITS_BYTES.IMAGE_MAX + 1024 * 1024);
pngHeader.copy(oversized);
```

docs/features/file-processors.md
```diff
@@ -354,23 +354,15 @@ const result = await neurolink.generate({
```
Tara-ag, line 354: 💡 **MINOR**: The docs say the image limit is configurable "Via `maxSize` / `maxBytes` param", but the public SDK entry points (`generate({ files })`, `generate({ images })`, `FileDetector.detectAndProcess`) do not currently expose those parameters to callers. The override exists only on internal helpers (`ImageProcessor.validateBufferSize`, `imageUtils.fileToBase64DataUri`, `imageUtils.urlToBase64DataUri`).

Either expose the override in the public API, or soften the docs to say the limit is enforced internally with overrides available only to advanced/internal callers, so users aren't directed toward a knob they cannot reach.

Tara-ag (commented):
## Review Summary

Reviewed 4 changed files. Raised **7 new inline comments** (2 MAJOR, 5 MINOR). No new blocking criteria are met.

### New issues

**MAJOR (2)**
- `src/lib/utils/imageProcessor.ts:441` — Error message tells callers to "raise the limit via the maxSize option", but public SDK entry points do not expose that override, making the message misleading.
- `src/lib/utils/messageBuilder.ts:1827` — `convertFilePathToBase64` uses synchronous `statSync` + `readFileSync`, blocking the event loop. The size guard prevents memory exhaustion but not I/O stalls.

**MINOR (5)**
- `src/lib/utils/imageProcessor.ts:435` — New size guard throws a plain `Error` instead of the existing typed `ErrorFactory.imageTooLarge(...)`.
- `src/lib/utils/messageBuilder.ts:1817` — `assertImageSize` duplicates `ImageProcessor.validateBufferSize`; verify the claimed module cycle and deduplicate if possible.
- `test/continuous-test-suite-bugfixes.ts:3858` — Test hard-codes the 10 MB image limit instead of using `SIZE_LIMITS_BYTES.IMAGE_MAX`.
- `test/continuous-test-suite-bugfixes.ts:3893` — Oversized-buffer test uses a hard-coded size and only 2 bytes of the PNG signature.
- `docs/features/file-processors.md:354` — Docs claim the image limit is configurable via `maxSize`/`maxBytes`, but those params are not exposed through the public SDK API.

### Existing review comments
Three Copilot review comments were already present on the same lines/topics (misleading `maxSize` message, hard-coded test limit, incomplete PNG header). Those are acknowledged above and should be addressed together with the suggested fixes.

### Decision
No hardcoded secrets, no critical security vulnerabilities, no backward-compatibility breaks, and no non-negotiable CLAUDE.md rule violations were found. With 2 MAJOR issues, the "3 or more MAJOR" blocking threshold is not met. Approving with comments for the author to address.

src/lib/utils/imageProcessor.ts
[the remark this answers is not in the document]
murdore: Fixed in b78b9e77f73d20cd2f3c8f262668d6a3638a6f25. `ImageProcessor.validateBufferSize` now throws `ErrorFactory.imageTooLarge(mb, maxMb)` (typed `NeuroLinkError`, code `IMAGE_TOO_LARGE`) whose message is `Image size (X MB) exceeds maximum (Y MB)` — it no longer references a `maxSize` option or any other knob the caller can't reach. See `src/lib/utils/imageProcessor.ts` (`validateSize`/`validateBufferSize`).

test/continuous-test-suite-bugfixes.ts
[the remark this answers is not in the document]
murdore: Fixed in b78b9e77f73d20cd2f3c8f262668d6a3638a6f25. Test now imports `SIZE_LIMITS_BYTES` from `src/lib/processors/config/sizeLimits.js` and uses `SIZE_LIMITS_BYTES.IMAGE_MAX` instead of the hard-coded `10 * 1024 * 1024`, so it tracks the real constant.

test/continuous-test-suite-bugfixes.ts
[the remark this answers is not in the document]
murdore: Fixed in b78b9e77f73d20cd2f3c8f262668d6a3638a6f25. The oversized buffer now uses `SIZE_LIMITS_BYTES.IMAGE_MAX + 1024 * 1024` for its size and copies the full 8-byte PNG signature (`0x89 0x50 0x4e 0x47 0x0d 0x0a 0x1a 0x0a`) via `Buffer.from([...]).copy(oversized)`, matching the real detection logic.

src/lib/utils/imageProcessor.ts
murdore: Fixed in b78b9e77f73d20cd2f3c8f262668d6a3638a6f25. Rather than hand-editing the message text, I switched the throw itself to the existing typed `ErrorFactory.imageTooLarge(mb, maxMb)` helper (see next comment thread) — its message is `Image size (X MB) exceeds maximum (Y MB)`, which drops the false "raise the limit via the maxSize option" claim entirely rather than just rewording it. Confirmed via grep that no public entry point (`FileDetector.detectAndProcess`, `generate({ files })`, `generate({ images })`) threads a `maxSize` override through, so the docs claim was also fixed (see the file-processors.md thread).

src/lib/utils/imageProcessor.ts
murdore: Fixed in b78b9e77f73d20cd2f3c8f262668d6a3638a6f25. `imageProcessor.ts` now imports `ErrorFactory` from `./errorHandling.js` and `validateBufferSize`/the new `validateSize` throw `ErrorFactory.imageTooLarge(mb, maxMb)` (confirmed the helper's existing signature is `(sizeMB: string, maxMB: string)` in `src/lib/utils/errorHandling.ts:575`, same pattern already used by `parameterValidation.ts`'s `validateImageForVideo`). Callers now get `code === IMAGE_TOO_LARGE`, `category === VALIDATION`, `retriable === false` as you noted.

src/lib/utils/messageBuilder.ts
murdore: Fixed in b78b9e77f73d20cd2f3c8f262668d6a3638a6f25. Confirmed there is no real cycle — traced ImageProcessor's full import chain (logger, rateLimiter, retryHandler, core/constants, processors/config/sizeLimits, imageCache, errorHandling, hitlErrors) and none of it touches messageBuilder.ts; separately, messageBuilder.ts already transitively imports ImageProcessor via fileDetector.ts (`FileDetector` imports `ImageProcessor` directly), so the dependency already existed. Deleted the local `assertImageSize` and its stale comment. To dedupe correctly I extracted a new `ImageProcessor.validateSize(size, context, maxSize?)` that both `validateBufferSize` (buffer-based) and the pre-read stat-based call site delegate to — this preserves the critical property that the file-path check happens on `stat.size` *before* reading the file into memory (the whole point of #257), which a naive swap to the buffer-only `validateBufferSize` would have broken. `messageBuilder.ts` now imports `ImageProcessor` and calls `ImageProcessor.validateSize(stats.size, ...)` for paths and `ImageProcessor.validateBufferSize(image, ...)` for buffers.

src/lib/utils/messageBuilder.ts
murdore: Fixed in b78b9e77f73d20cd2f3c8f262668d6a3638a6f25, following your suggested pattern closely. `convertFilePathToBase64` is now `async` and uses `stat`/`readFile` from `fs/promises` (aliased `statAsync`/`readFileAsync` to avoid colliding with the existing sync `fs` import used elsewhere in the file) instead of `statSync`/`readFileSync`. `processImageToBase64` is now `async` too, and its call site in `convertSimpleImagesToProviderFormat` switched from a synchronous `actualImages.forEach` to a sequential `for...of` loop with `await` (kept sequential rather than `Promise.all` to preserve image ordering and the original one-at-a-time error semantics). `convertSimpleImagesToProviderFormat` was already `async`, so no signature changes propagate further up.

test/continuous-test-suite-bugfixes.ts
murdore: Fixed in b78b9e77f73d20cd2f3c8f262668d6a3638a6f25, exactly as suggested: added `import { SIZE_LIMITS_BYTES } from "../src/lib/processors/config/sizeLimits.js";` and replaced `const IMAGE_MAX = 10 * 1024 * 1024;` with `const IMAGE_MAX = SIZE_LIMITS_BYTES.IMAGE_MAX;`.

test/continuous-test-suite-bugfixes.ts
murdore: Fixed in b78b9e77f73d20cd2f3c8f262668d6a3638a6f25, using your suggested snippet: the buffer is now `Buffer.alloc(SIZE_LIMITS_BYTES.IMAGE_MAX + 1024 * 1024)` and the full 8-byte PNG signature (`Buffer.from([0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a])`) is copied into it via `.copy(oversized)` instead of only writing 2 bytes.

docs/features/file-processors.md
murdore: Fixed in b78b9e77f73d20cd2f3c8f262668d6a3638a6f25. Updated the File Size Limits table row for Images from `Via maxSize / maxBytes param` to `No — fixed limit`, and reworded the explanatory paragraph below it to state plainly that the internal size-override parameter exists only on internal helpers and 'is not exposed through generate() or any other public API — the 10 MB limit is fixed for SDK callers.' See docs/features/file-processors.md around line 354-365.

src/lib/utils/messageBuilder.ts
```diff
@@ -1800,13 +1802,20 @@ function detectMimeTypeFromBuffer(buffer: Buffer): string | undefined {
 /**
  * Convert file path to raw base64 string.
```
Tara-ag, line 1804: ⚠️ **MAJOR**: `statAsync(filePath).catch(() => null)` swallows every `stat` failure (permission denied, not-a-directory, too many open files, etc.) and reports it as `Image file not found`. This is a regression from the previous sync path, where `readFileSync` surfaced the real `EACCES`/`EISDIR`/etc. error.

Preserve the original error for non-ENOENT cases so callers and logs can distinguish a missing file from a permission/system problem:

```ts
let stats;
try {
  stats = await statAsync(filePath);
} catch (error) {
  const code = error instanceof Error ? (error as NodeJS.ErrnoException).code : undefined;
  if (code === "ENOENT") {
    throw new Error(`Image file not found: ${filePath}`);
  }
  throw new Error(
    `Cannot access image file ${filePath}: ${error instanceof Error ? error.message : String(error)}`,
  );
}
if (!stats.isFile()) {
  throw new Error(`Image path is not a file: ${filePath}`);
}
```

This keeps the #257 size guard intact while restoring meaningful error semantics.

src/lib/utils/imageProcessor.ts
```diff
@@ -403,6 +417,66 @@ export class ImageProcessor {
     }
```
Tara-ag: ⚠️ **MAJOR**: `validateSize` accepts a `number` but does not reject `NaN`, `Infinity`, or negative sizes. A malformed caller (or a computed size from a corrupted stat/header) could pass `NaN` and silently skip the guard (`NaN > maxSize` is `false`), or pass `-1` and be accepted.

Fail closed on nonsensical inputs before the comparison:

```ts
static validateSize(
  size: number,
  context: string,
  maxSize: number = SIZE_LIMITS_BYTES.IMAGE_MAX,
): void {
  if (!Number.isFinite(size) || size < 0) {
    throw ErrorFactory.invalidImageFormat();
  }
  if (size > maxSize) {
    const mb = (size / (1024 * 1024)).toFixed(1);
    const maxMb = (maxSize / (1024 * 1024)).toFixed(0);
    throw ErrorFactory.imageTooLarge(mb, maxMb);
  }
}
```

This aligns with the bootstrapped standard that security-sensitive configuration/sizing surfaces must fail closed when malformed.

src/lib/utils/imageProcessor.ts
```diff
@@ -711,7 +785,7 @@ export const imageUtils = {
    * Convert Buffer to base64 string
```
Tara-ag: 💡 **MINOR**: `imageUtils.bufferToBase64` now delegates to `safeBase64Convert` with the default 10 MB limit. This is a **behavior change** for any external callers of `imageUtils.bufferToBase64` who previously passed arbitrarily large buffers. The function is exported, so this tightens the public-ish utility surface.

That tightening is correct for #257, but consider documenting it in the changelog/release notes and adding a quick unit test for this utility specifically, since the existing tests only exercise `FileDetector` and `buildMultimodalMessagesArray`.

src/lib/utils/imageProcessor.ts
```diff
@@ -389,7 +403,7 @@ export class ImageProcessor {
    */
```
Tara-ag: 💡 **MINOR**: `validateImageSize` still defaults to `SIZE_LIMITS_BYTES.IMAGE_MAX` and returns `boolean` instead of throwing. It is now shadowed by the new throwing helpers `validateSize`/`validateBufferSize`/`safeBase64Convert`. Please confirm this legacy method is still intentionally used elsewhere; if not, deprecating or redirecting it to the throwing helpers would avoid two sources of truth for image size enforcement.

Tara-ag (approved):
## Review summary

Reviewed the 4 changed files in PR #1190. Existing review threads were all resolved/outdated, so I focused on the current HEAD (`b78b9e7`).

### New comments raised

| Severity | Count | Files |
|----------|-------|-------|
| 🔒 CRITICAL | 0 | — |
| ⚠️ MAJOR | 2 | `src/lib/utils/messageBuilder.ts`, `src/lib/utils/imageProcessor.ts` |
| 💡 MINOR | 2 | `src/lib/utils/imageProcessor.ts` |

### Key findings

1. **`messageBuilder.ts` `convertFilePathToBase64`** — `statAsync(...).catch(() => null)` swallows permission/system errors and reports them all as “file not found”. Suggested preserving the real `EACCES`/`EISDIR`/etc. error for non-ENOENT cases.
2. **`imageProcessor.ts` `validateSize`** — does not fail closed on `NaN`, `Infinity`, or negative sizes, which could let malformed inputs bypass the guard. Suggested adding finite/non-negative validation.
3. **`imageProcessor.ts` `imageUtils.bufferToBase64`** — tightening the public-ish utility to enforce the 10 MB limit is correct for #257, but worth noting as a behavior change and adding a dedicated unit test.
4. **`imageProcessor.ts` `validateImageSize`** — legacy boolean-return method now shadows the new throwing helpers; flagged to avoid two sources of truth.

### Decision

No blocking criteria from `<blocking-criteria>` are met (no secrets, no critical security vulnerability, no public SDK backward-compatibility break, no non-negotiable CLAUDE.md rule violation, and fewer than 3 MAJOR issues). Approving with inline feedback for the author to address at their discretion.

src/lib/utils/imageProcessor.ts
```diff
@@ -171,7 +176,10 @@ export class ImageProcessor {
       }
 
       // Handle Buffer - convert to data URI
-      const base64 = image.toString("base64");
+      const base64 = ImageProcessor.safeBase64Convert(
+        image,
+        "OpenAI image input",
+      );
```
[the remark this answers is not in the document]
murdore: Fixed in 0e830553bdefdb09b84a5bfe46797af416899c73.

Added `if (error instanceof NeuroLinkError) { throw error; }` before the generic wrapping in every catch you flagged, plus one more for correctness:

- `imageProcessor.ts` `processImageForOpenAI` catch
- `imageProcessor.ts` `processImageForGoogle` catch
- `imageProcessor.ts` `processImageForAnthropic` catch
- `imageProcessor.ts` `processImageForVertex` catch (not explicitly listed, but it wraps the Google/Anthropic calls above and had the same flattening bug — fixed for consistency)
- `imageProcessor.ts` outer `processImage` generic-provider catch
- `messageBuilder.ts` `processImageToBase64` file-path branch catch (the site around the old L1867-1877)

Typed `IMAGE_TOO_LARGE`/`INVALID_IMAGE_SIZE` errors (and any other `NeuroLinkError`) now propagate with their `code`/`category`/`retriable` intact through every one of these wrappers instead of being flattened into a plain `Error`.

Covered by a new deterministic test: "ImageProcessor provider wrappers preserve typed IMAGE_TOO_LARGE error through every catch (not flattened to plain Error)" in `test/continuous-test-suite-bugfixes.ts`, which exercises all four provider paths plus the default `processImage` branch.

src/lib/utils/messageBuilder.ts
```diff
@@ -1800,13 +1802,20 @@ function detectMimeTypeFromBuffer(buffer: Buffer): string | undefined {
...
     throw new Error(`Image file not found: ${filePath}`);
   }
 
-  const buffer = readFileSync(filePath);
+  // #257: check the file size via stat BEFORE reading it into memory, so a
+  // huge image path never triggers an unbounded read + base64 allocation.
+  // Delegates to ImageProcessor's shared guard (no local reimplementation).
+  ImageProcessor.validateSize(stats.size, `image file "${filePath}"`);
+  const buffer = await readFileAsync(filePath);
   return buffer.toString("base64");
```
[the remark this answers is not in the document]
murdore: Fixed in 0e830553bdefdb09b84a5bfe46797af416899c73, applied essentially as suggested.

`convertFilePathToBase64` now does:

```ts
const context = `image file "${filePath}"`;
ImageProcessor.validateSize(stats.size, context);
const buffer = await readFileAsync(filePath);
return ImageProcessor.safeBase64Convert(buffer, context);
```

The pre-read `stat.size` check is kept (so an obviously huge file never triggers the read at all), and `safeBase64Convert` re-validates the actual bytes returned by `readFileAsync` before doing the base64 allocation — so a file that grows or a symlink swapped between `stat` and `readFile` is still caught.

Covered by two new deterministic tests in `test/continuous-test-suite-bugfixes.ts`: a direct unit test of `ImageProcessor.safeBase64Convert` re-validating actual buffer bytes independent of any prior size claim, plus an end-to-end `convertFilePathToBase64` (via `buildMultimodalMessagesArray`) integration test with real files on disk.

src/lib/utils/messageBuilder.ts
murdore: Fixed in 0e830553bdefdb09b84a5bfe46797af416899c73, applied as suggested (with `{ cause: error }` attached per this repo's lint rule for rethrows in a catch):

```ts
let stats;
try {
  stats = await statAsync(filePath);
} catch (error) {
  const code =
    error instanceof Error
      ? (error as NodeJS.ErrnoException).code
      : undefined;
  if (code === "ENOENT") {
    throw new Error(`Image file not found: ${filePath}`, { cause: error });
  }
  throw new Error(
    `Cannot access image file ${filePath}: ${error instanceof Error ? error.message : String(error)}`,
    { cause: error },
  );
}
if (!stats.isFile()) {
  throw new Error(`Image path is not a file: ${filePath}`);
}
```

Only `ENOENT` now reports "not found"; `EACCES`/`ELOOP`/etc. surface as a distinct "Cannot access image file" error with the original error attached as `cause`.

Covered by a new deterministic test: "convertFilePathToBase64: ENOENT reports 'not found', EACCES/other stat errors are NOT misreported as 'not found'" in `test/continuous-test-suite-bugfixes.ts` — it chmods a containing directory to `0o000` to force a real `EACCES` at `stat()` time (skipped when running as root, since permission bits don't apply).

src/lib/utils/imageProcessor.ts
murdore: Fixed in 0e830553bdefdb09b84a5bfe46797af416899c73, with one deliberate deviation from the suggested diff: instead of reusing `ErrorFactory.invalidImageFormat()` (whose message is "Unsupported image format. Use JPEG, PNG, or WebP" — misleading for a size failure), I added a new, semantically accurate `ErrorFactory.invalidImageSize(size)` factory (new `ERROR_CODES.INVALID_IMAGE_SIZE`) in `errorHandling.ts`:

```ts
static validateSize(
  size: number,
  context: string,
  maxSize: number = SIZE_LIMITS_BYTES.IMAGE_MAX,
): void {
  // Fail closed on malformed sizes first: `NaN > maxSize` and
  // `-1 > maxSize` both evaluate to `false`, which would otherwise let a
  // corrupted stat/header value silently skip the guard below.
  if (!Number.isFinite(size) || size < 0) {
    throw ErrorFactory.invalidImageSize(size);
  }
  if (size > maxSize) {
    const mb = (size / (1024 * 1024)).toFixed(1);
    const maxMb = (maxSize / (1024 * 1024)).toFixed(0);
    throw ErrorFactory.imageTooLarge(mb, maxMb);
  }
}
```

Same fail-closed behavior you asked for (NaN/Infinity/negative rejected before the max-size comparison), just with an error `code`/message that actually describes the failure.

Covered by a new deterministic test: "ImageProcessor.validateSize rejects NaN/Infinity/negative sizes with typed INVALID_IMAGE_SIZE (fail closed)" in `test/continuous-test-suite-bugfixes.ts`.

src/lib/utils/imageProcessor.ts
murdore: Addressed in 0e830553bdefdb09b84a5bfe46797af416899c73.

Decision: kept the 10MB default (matches every other #257 conversion site) but made the limit an explicit, opt-in override rather than a silent fixed cap:

```ts
bufferToBase64: (
  buffer: Buffer,
  maxBytes: number = SIZE_LIMITS_BYTES.IMAGE_MAX,
): string => {
  return ImageProcessor.safeBase64Convert(buffer, "buffer-to-base64", maxBytes);
},
```

Why this direction and not "keep it unlimited": `imageUtils`/`ImageProcessor` are internal-only — neither is re-exported from `src/lib/index.ts` nor listed in `package.json`'s `exports` map — so there is no external SDK consumer whose behavior this changes; the "behavior change for external callers" risk you flagged doesn't apply here. Any internal caller that legitimately needs a larger buffer can now pass `maxBytes` explicitly instead of being silently capped.

Added a dedicated unit test per your suggestion: "imageUtils.bufferToBase64: default stays guarded at 10MB; explicit maxBytes override is opt-in (no silent break for internal callers)" in `test/continuous-test-suite-bugfixes.ts`.

src/lib/utils/imageProcessor.ts
murdore: Fixed in 0e830553bdefdb09b84a5bfe46797af416899c73.

Confirmed via a repo-wide grep that the legacy boolean-returning `validateImageSize(data, maxSize)` had no remaining call sites (only a stale reference in a `memory-bank/` doc, which isn't code). Removed it outright rather than deprecating/redirecting, so there's now a single source of truth for image size enforcement: the throwing `validateSize` / `validateBufferSize` / `safeBase64Convert` helpers.

Tara-ag (approved):
## Review summary

Reviewed the current HEAD (`0e83055`) across all 5 changed files. Existing review threads from earlier rounds were resolved/outdated, and the author has addressed all previously raised MAJOR/MINOR points:

- Misleading "raise the limit via maxSize" message was replaced with typed `ErrorFactory.imageTooLarge(...)`.
- `convertFilePathToBase64` is now async with `fs/promises`, pre-read stat guard, and post-read buffer re-validation.
- `IMAGE_TOO_LARGE`/`INVALID_IMAGE_SIZE` typed errors are preserved through every provider wrapper catch.
- Tests now import `SIZE_LIMITS_BYTES.IMAGE_MAX` instead of hard-coding 10 MB.

### New findings this run

No new CRITICAL or MAJOR issues. No blocking criteria are met:
- No critical security vulnerability (injection, SSRF, unsafe eval, secret/PII exposure).
- No public SDK backward-compatibility break.
- No non-negotiable CLAUDE.md rule violation.

Two unresolved CodeRabbit comments remain on the latest revision and should be addressed by the author:
1. `src/lib/utils/imageProcessor.ts` — validate `maxSize` in `validateSize` to fail closed on `NaN`/`Infinity`/negative limits.
2. `test/continuous-test-suite-bugfixes.ts` — run the ENOENT assertion before the root-user early-return so "not found" regression coverage executes under root as well.

Approving.

src/lib/utils/imageProcessor.ts
```diff
@@ -385,22 +419,71 @@ export class ImageProcessor {
...
+    // Fail closed on malformed sizes first: `NaN > maxSize` and
+    // `-1 > maxSize` both evaluate to `false`, which would otherwise let a
+    // corrupted stat/header value silently skip the guard below.
+    if (!Number.isFinite(size) || size < 0) {
+      throw ErrorFactory.invalidImageSize(size);
     }
+    if (size > maxSize) {
+      const mb = (size / (1024 * 1024)).toFixed(1);
+      const maxMb = (maxSize / (1024 * 1024)).toFixed(0);
+      throw ErrorFactory.imageTooLarge(mb, maxMb);
```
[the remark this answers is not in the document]
murdore: Fixed as proposed — validateSize now rejects a non-finite or negative maxSize via ErrorFactory.invalidConfiguration before comparing against size, closing the NaN/Infinity/negative bypass. Added a deterministic regression test covering NaN/Infinity/-1/-Infinity maxSize plus the existing in-range/over-range behavior. See src/lib/utils/imageProcessor.ts:438-462 and test/continuous-test-suite-bugfixes.ts (new test: 'ImageProcessor.validateSize rejects malformed maxSize...') at bf232341b391174207f8f1690635b44e92a046f9.

test/continuous-test-suite-bugfixes.ts
```diff
@@ -3850,6 +3859,300 @@ exit 127
...
+  },
+  {
+    name: "convertFilePathToBase64: ENOENT reports 'not found', EACCES/other stat errors are NOT misreported as 'not found'",
+    category: "message-builder",
+    fn: async () => {
+      // Permission bits are meaningless for root — chmod 000 does not deny
+      // access, so the EACCES half of this test cannot run under root.
+      if (typeof process.getuid === "function" && process.getuid() === 0) {
+        return null;
+      }
```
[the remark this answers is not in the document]
murdore: Fixed — reordered so the root-only skip now only guards the EACCES half. The ENOENT assertion runs unconditionally first, then `isRoot` short-circuits (returns null) only after that assertion has already executed, right before the EACCES/chmod portion. See test/continuous-test-suite-bugfixes.ts:4085-4124 at bf232341b391174207f8f1690635b44e92a046f9.

docs/features/file-processors.md
Tara-ag, line 1: Docs update looks accurate: the 20 MB claim is corrected to 10 MB, the fictional `ProcessorConfig.setLimits()` snippet is removed, and the fixed-limit nature of the public API is clearly stated. No issues.

Tara-ag (approved):
## Review Summary

Reviewed all 5 changed files in PR #1190. No new blocking issues found.

### What was checked
- **Security / memory exhaustion (#257)**: All previously-unguarded `Buffer.toString("base64")` sites in `imageProcessor.ts` now route through `safeBase64Convert`; the sibling `generate({ images })` path in `messageBuilder.ts` guards both file-path (pre-read `stat` + post-read revalidation) and Buffer branches.
- **Fail-closed validation**: `ImageProcessor.validateSize` rejects malformed `size` (NaN/Infinity/negative) and malformed `maxSize` before comparison, preventing bypasses.
- **Typed error preservation**: Every provider wrapper catch (`OpenAI`, `Google`, `Anthropic`, `Vertex`, generic `processImage`) and the `messageBuilder.ts` file-path catch rethrow `NeuroLinkError` unchanged instead of flattening to plain `Error`.
- **Async I/O**: `convertFilePathToBase64` is now async and uses `fs/promises`, avoiding event-loop blocking; `convertSimpleImagesToProviderFormat` processes images sequentially with `for…of` to preserve ordering and error semantics.
- **Docs**: `file-processors.md` correctly states the 10 MB fixed public limit and removes the fictional `ProcessorConfig.setLimits()` snippet.
- **Tests**: New deterministic regression tests cover oversized rejection, at-limit acceptance, NaN/Infinity/negative size and maxSize rejection, typed-error preservation through all wrappers, TOCTOU revalidation, `imageUtils.bufferToBase64` override behavior, and ENOENT/EACCES error classification.

### Existing review comments
All prior inline comments from earlier review rounds have been addressed by the author (verified in `get_review_comments`). No duplicates raised.

### Decision
Approve — the change is focused, well-tested, and addresses the security issue without breaking the public SDK API.

docs/features/file-processors.md
murdore: Thanks — resolving.

src/lib/utils/messageBuilder.ts
```diff
@@ -1874,6 +1910,9 @@ function processImageToBase64(
     if (detectedMimeType) {
       mimeType = detectedMimeType;
     }
+    // #257: guard the buffer size before the unbounded base64 conversion.
```
Tara-ag, line 1913: 💡 **SUGGESTION**: For consistency with the other six guarded conversion sites in this PR, consider using `ImageProcessor.safeBase64Convert(image, \`image input at index ${index}\`)` here instead of manually calling `validateBufferSize` followed by `toString("base64")`. It keeps the #257 guard pattern uniform and removes the small risk that a future edit updates one line but not the other.

```ts
imageData = ImageProcessor.safeBase64Convert(
  image,
  `image input at index ${index}`,
);
```

Tara-ag (approved):
## Review summary

**Files reviewed:** 5 changed files
**New issues raised:** 1 suggestion
**Blocking issues:** 0

### Assessment

This PR cleanly addresses #257 (memory exhaustion on large images) by centralizing size enforcement in `ImageProcessor` and routing every caller-supplied Buffer → base64 conversion through guarded helpers. The fix covers both public entry points (`generate({ files })` via `FileDetector` and `generate({ images })` via the message builder), uses the canonical `SIZE_LIMITS_BYTES.IMAGE_MAX` constant, and preserves typed `IMAGE_TOO_LARGE` errors through provider wrappers.

### What I checked

- **Security:** No hardcoded secrets or credentials. The size guard prevents unbounded base64 allocation on attacker-controlled buffers/files.
- **Backward compatibility:** `ImageProcessor` and `imageUtils` are not re-exported from `src/lib/index.ts` or the package `exports` map, so the new 10 MB default does not break the public SDK surface. The optional `maxBytes` / `maxSize` overrides keep internal callers opt-in capable.
- **Correctness:** `validateSize` fails closed on `NaN` / `Infinity` / negative sizes and malformed `maxSize` values. `convertFilePathToBase64` re-validates bytes after `readFileAsync` to close the stat→read TOCTOU window.
- **Error handling:** Typed `NeuroLinkError` (`IMAGE_TOO_LARGE`, `INVALID_IMAGE_SIZE`, `INVALID_CONFIGURATION`) is preserved through every provider wrapper catch.
- **Tests:** Deterministic regression coverage added for oversized rejection, at-limit acceptance, NaN/negative size rejection, malformed `maxSize` rejection, typed-error preservation across all provider paths, TOCTOU revalidation, and file-access error semantics.

### New comment

- `src/lib/utils/messageBuilder.ts:1913` — minor suggestion to use `ImageProcessor.safeBase64Convert` for consistency with the other guarded conversion sites.

No blocking criteria are met. Approving.

PR #1191, opened 2026-07-17
fix(cli): validate file inputs + wire multimodal flags into batch (#288, #291)

### LINKED ISSUE #288: CLI-002: No File Validation Before Processing
## Summary
CLI accepts file paths for multimodal inputs without validating existence or accessibility, causing cryptic runtime errors instead of clear, actionable error messages.

## Root Cause
File processor methods pass paths directly to the SDK without validation. The SDK's message builder encounters missing files deep in processing, resulting in generic filesystem errors without troubleshooting guidance.

**Location**: `src/cli/factories/commandFactory.ts` lines 238-281

## Fix
Add `validateFilePath()` helper method that:
- Checks file existence
- Rejects directories (files only)
- Warns for large files (>10MB images, >50MB PDFs/CSVs)
- Skips validation for URLs
- Provides troubleshooting steps in error messages

## Acceptance Criteria
- [ ] File existence validated before processing
- [ ] Directories rejected with clear error
- [ ] Large file warnings shown
- [ ] URLs skip validation
- [ ] Clear error messages with troubleshooting guidance
- [ ] Tests pass

### LINKED ISSUE #291: CLI-001: Batch Command Ignores Multimodal Flags
## Summary
The batch command accepts multimodal flags (`--image`, `--pdf`, `--csv`, `--file`) but completely ignores them during execution, causing silent failures.

## Root Cause
The `executeBatch()` method only passes text input to `sdk.generate()`. Multimodal inputs are processed by the CLI but never included in the generate call.

**Location**: `src/cli/factories/commandFactory.ts` lines 2020-2178

## Fix
Process multimodal inputs and include them in the batch generate call:
1. Process multimodal files once before batch loop
2. Warn user that files will be attached to ALL prompts
3. Include files in each generate call within the batch loop
4. Add csvOptions to batch generate call

## Acceptance Criteria
- [ ] Multimodal files processed and attached to batch prompts
- [ ] Warning shown that files attach to ALL prompts
- [ ] CSV options (maxRows, format) respected
- [ ] Tests pass for batch multimodal operations

murdore:
## What & why

Consolidated CLI-subsystem critical fixes (`commandFactory.ts`).

### #288 (CLI-002) — no file validation before processing
`validateCliInputFiles` only did an `existsSync` check, so a **directory** passed to `--image`/`--csv`/`--pdf`/`--video`/`--file` was accepted and failed cryptically deep in processing (raw `EISDIR`). Now:
- A directory (or missing path) is **rejected up front** with a clear error + a **Troubleshooting** block.
- A **large** file (images > 10 MB, CSV/PDF > 50 MB) prints a **non-blocking** warning, visible without `--debug` (via `logger.always`, since `logger.warn` is hidden by default).

### #291 (CLI-001) — batch ignored multimodal flags
The `batch` command parsed `--image/--csv/--pdf/--video` but **silently discarded** them (`runBatchGenerate` sent only `{ text }`). Now it resolves the multimodal inputs **once** (they attach to every prompt), runs the same pre-flight validation as `generate`/`stream`, warns that the files apply to all prompts, and wires `images/csvFiles/pdfFiles/videoFiles` + `csvOptions/videoOptions` into every batch call.

**Enhancement/known-limit documented:** `--file` (auto-detect) is *not* wired in `batch` because it collides with the `<file>` positional (the prompts-list path) — called out in the CLI and docs so it doesn't become a repeat bug report.

## Testing / proof

CLI **end-to-end** tests (bugfixes suite, spawning the built `dist/cli`):
- `generate hi --image <dir> --dry-run --quiet` → **non-zero-free rejection** with `directory` + `troubleshooting` in the output (was accepted before).
- `batch <prompts> --csv <nonexistent>` → now **validated** (`--csv path not found` + troubleshooting), proving the flag is wired in (previously ignored); a valid `--csv` still completes the batch (2 results for 2 prompts).

`pnpm run check` ✅ (0 errors) · `pnpm run lint` ✅ (0 errors) · `pnpm run build:cli` ✅ · `prettier --check` ✅.

## Risk
Low. Validation only *adds* pre-flight rejection of genuinely-bad inputs (directories / missing files) and non-blocking warnings; valid inputs behave identically. Batch gains multimodal support it was documented to have.

Fixes #288. Fixes #291.

## Summary by CodeRabbit

- **New Features**
  - Batch workflows accept image, CSV, PDF, and video inputs and attach the same files to every prompt.
  - The batch prompts input is documented and validated as a dedicated positional prompts file, with clearer handling of flag/argument collisions (notably, `--file` is rejected for batch).

- **Bug Fixes**
  - `--format json` output remains parseable on stdout; notices and large-file warnings are routed to stderr only.
  - Pre-flight input validation now fails fast with aggregated, user-friendly errors (including malformed `file://` and directory paths), even in dry runs and without API keys.

- **Documentation**
  - Updated the multimodal guide with batch workflow details, validation behavior, stderr messaging, and troubleshooting guidance.

### DIFF
--- a/docs/features/multimodal.md
+++ b/docs/features/multimodal.md
@@ -424,6 +424,25 @@ for await (const chunk of stream) {
 }
 ```
 
+### Batch with Multimodal (CLI)
+
+The `batch` command supports `--image`, `--csv`, `--pdf`, and `--video`. The file(s) are attached **identically to every prompt** in the batch (a one-line notice is printed to **stderr**, unconditionally — it is not suppressed by `--quiet`, so it never corrupts `--format json` output written to stdout):
+
+```bash
+neurolink batch questions.txt --csv sales.csv --format json
+```
+
+> `--file` (auto-detect) is **not** available in `batch`, because it collides with the `<file>` positional (the prompts-list path). Use the explicit `--image` / `--csv` / `--pdf` / `--video` flags instead.
+
+### File validation & troubleshooting (CLI)
+
+Before any provider call, the CLI validates local file inputs across `generate`, `stream`, and `batch`:
+
+- A path that points at a **directory**, doesn't exist, or can't be read (e.g. a permissions error) is rejected up front with a clear error and troubleshooting hints — no cryptic `EISDIR`/`EACCES` deep in processing. This also applies to the `batch` `<file>` prompts-list positional itself.
+- A **large** file (images > 10 MB, CSV/PDF > 50 MB) prints a non-blocking warning to **stderr**. Like the batch attachment notice, this is unconditional — visible without `--debug` and regardless of `--quiet` — and never mixes into stdout, so `--format json` output stays valid JSON even when large-file warnings fire.
+
+This runs even under `--dry-run` and without API keys configured.
+
 ---
 
 ## Configuration & Fine-tuning
--- a/src/cli/factories/commandFactory.ts
+++ b/src/cli/factories/commandFactory.ts
@@ -50,6 +50,7 @@ import { resolveFilePaths } from "../utils/pathResolver.js";
 import {
   validateCliInputFiles,
   validateCsvMaxRows,
+  validatePromptsFilePath,
 } from "../utils/inputValidation.js";
 import { animatedWrite } from "../utils/typewriter.js";
 import { createStreamAbortHandler } from "../utils/abortHandler.js";
@@ -845,12 +846,31 @@ export class CLICommandFactory {
     },
   };
 
-  // Helper method to build options for commands
-  private static buildOptions(yargs: Argv, additionalOptions = {}) {
+  // Helper method to build options for commands. `excludeOptions` drops
+  // common flags from registration entirely (not just from validation) —
+  // #1191 round-5: batch uses this to omit `commonOptions.file` so `--file`
+  // isn't a recognized flag at all, since it would collide with the
+  // `<file>` positional (the prompts list).
+  private static buildOptions(
+    yargs: Argv,
+    additionalOptions = {},
+    excludeOptions: Array<keyof typeof CLICommandFactory.commonOptions> = [],
+  ) {
+    const baseOptions =
+      excludeOptions.length === 0
+        ? CLICommandFactory.commonOptions
+        : (Object.fromEntries(
+            Object.entries(CLICommandFactory.commonOptions).filter(
+              ([key]) =>
+                !excludeOptions.includes(
+                  key as keyof typeof CLICommandFactory.commonOptions,
+                ),
+            ),
+          ) as typeof CLICommandFactory.commonOptions);
     return (
       yargs
         .options({
-          ...CLICommandFactory.commonOptions,
+          ...baseOptions,
           ...additionalOptions,
         })
         // NEW9: implies relationships so users who pass --stt-provider or
@@ -2003,12 +2023,20 @@ export class CLICommandFactory {
    */
   static createBatchCommand(): CommandModule {
     return {
-      command: "batch <file>",
+      // #1191 round-5: the positional is named `promptsFile`, not `file` —
+      // yargs implicitly accepts `--<positionalName>` as a flag alias for
+      // any positional regardless of whether it's separately registered via
+      // `.options()`, so a positional literally named `file` would let
+      // `--file` silently pass strictOptions() even after removing
+      // `commonOptions.file` below. Renaming the key (invocation syntax is
+      // unaffected — it's still just `neurolink batch prompts.txt`) is what
+      // actually makes `--file` an unknown argument.
+      command: "batch <promptsFile>",
       describe: "Process multiple prompts from a file",
       builder: (yargs) => {
         return CLICommandFactory.buildOptions(
           yargs
-            .positional("file", {
+            .positional("promptsFile", {
               type: "string" as const,
               description: "File with prompts (one per line)",
               demandOption: true,
@@ -2026,6 +2054,11 @@ export class CLICommandFactory {
               "$0 batch batch.txt --output results.json",
               "Save results to file",
             ),
+          {},
+          // `--file` is not a batch flag — the positional above is the
+          // prompts-list path. Omit it from registration so it's rejected
+          // as unknown rather than silently accepted then ignored.
+          ["file"],
         );
       },
       handler: async (argv) =>
@@ -4616,15 +4649,41 @@ export class CLICommandFactory {
     const spinner = options.quiet ? null : ora().start();
 
     try {
-      if (!argv.file) {
+      if (!argv.promptsFile) {
         throw new Error("No file specified");
       }
 
-      if (!fs.existsSync(argv.file)) {
-        throw new Error(`File not found: ${argv.file}`);
-      }
+      // #291/round-4 #1191: validate multimodal flags and resolve their
+      // paths *before* the prompts file is read/parsed below, mirroring
+      // generate/stream (which validate all inputs before any processing).
+      // Previously this ran after `fs.readFileSync(resolvedPromptsPath)`,
+      // so a large prompts file paired with an invalid --image path paid
+      // the read/parse cost before the flag error ever surfaced. `--file`
+      // (the common auto-detect flag) isn't registered on batch at all —
+      // see createBatchCommand — so `argv.file` can never be set here; no
+      // exclusion needed to protect the (differently-named) positional.
+      validateCliInputFiles(argv as unknown as Record<string, unknown>);
+      const batchImages = CLICommandFactory.processCliImages(
+        argv.image as string | string[] | undefined,
+      );
+      const batchCsvFiles = CLICommandFactory.processCliCSVFiles(
+        argv.csv as string | string[] | undefined,
+      );
+      const batchPdfFiles = CLICommandFactory.processCliPDFFiles(
+        argv.pdf as string | string[] | undefined,
+      );
+      const batchVideoFiles = CLICommandFactory.processCliVideoFiles(
+        argv.video as string | string[] | undefined,
+      );
 
-      const buffer = fs.readFileSync(argv.file);
+      // #291: route the prompts-list positional through the same friendly
+      // aggregated-error path as the multimodal flags above (not found /
+      // unreadable / directory / malformed file:// URL), instead of a raw
+      // existsSync+readFileSync that throws an unfriendly EISDIR when
+      // pointed at a directory.
+      const resolvedPromptsPath = validatePromptsFilePath(argv.promptsFile);
+
+      const buffer = fs.readFileSync(resolvedPromptsPath);
       const prompts = buffer
         .toString("utf8")
         .split("\n")
@@ -4680,6 +4739,26 @@ export class CLICommandFactory {
       const enhancedOptions = { ...options, ...sessionVariables };
       const sessionId = globalSession.getCurrentSessionId();
 
+      // batchImages/batchCsvFiles/batchPdfFiles/batchVideoFiles were already
+      // validated and resolved above, before the prompts file was read.
+      if (
+        batchImages?.length ||
+        batchCsvFiles?.length ||
+        batchPdfFiles?.length ||
+        batchVideoFiles?.length
+      ) {
+        // Pause the spinner so the notice isn't swallowed by its re-renders.
+        // Safety-relevant, so always visible (not gated behind --quiet) and
+        // written to stderr so `--format json` output on stdout stays parseable.
+        spinner?.stop();
+        logger.alwaysStderr(
+          chalk.yellow(
+            "⚠️  Multimodal files (--image/--csv/--pdf/--video) are attached to ALL prompts in this batch.",
+          ),
+        );
+        spinner?.start();
+      }
+
       for (let i = 0; i < prompts.length; i++) {
         if (spinner) {
           spinner.text = `Processing ${i + 1}/${prompts.length}: ${prompts[i].substring(0, 30)}...`;
@@ -4729,7 +4808,37 @@ export class CLICommandFactory {
 
           const runBatchGenerate = () =>
             sdk.generate({
-              input: { text: inputText },
+              input: {
+                text: inputText,
+                ...(batchImages?.length && { images: batchImages }),
+                ...(batchCsvFiles?.length && { csvFiles: batchCsvFiles }),
+                ...(batchPdfFiles?.length && { pdfFiles: batchPdfFiles }),
+                ...(batchVideoFiles?.length && {
+                  videoFiles: batchVideoFiles,
+                }),
+              },
+              // Only construct these when the corresponding files are
+              // actually attached — batch has no CSV/video mode to opt into,
+              // unlike generate/stream, so an empty options object here is
+              // pure noise on every prompt of every batch run.
+              ...(batchCsvFiles?.length && {
+                csvOptions: {
+                  maxRows: argv.csvMaxRows as number | undefined,
+                  formatStyle: argv.csvFormat as
+                    | "raw"
+                    | "markdown"
+                    | "json"
+                    | undefined,
+                },
+              }),
+              ...(batchVideoFiles?.length && {
+                videoOptions: {
+                  frames: argv.videoFrames as number | undefined,
+                  quality: argv.videoQuality as number | undefined,
+                  format: argv.videoFormat as "jpeg" | "png" | undefined,
+                  transcribeAudio: argv.transcribeAudio as boolean | undefined,
+                },
+              }),
               provider: enhancedOptions.provider,
               model: enhancedOptions.model,
               temperature: enhancedOptions.temperature,
--- a/src/cli/utils/inputValidation.ts
+++ b/src/cli/utils/inputValidation.ts
@@ -1,18 +1,22 @@
 /**
  * CLI Input Validation Utilities
  *
- * Validates multimodal file flags (--image/--csv/--pdf/--video/--file) and
- * --csv-max-rows before a generate/stream/batch run starts, so bad input
- * fails fast with a clear message instead of surfacing deep inside provider
- * calls. Exported as standalone functions (rather than private statics on
- * CLICommandFactory) so callers — including tests — can import and invoke
- * them directly, without reaching into the class via an `unknown` cast.
+ * Validates multimodal file flags (--image/--csv/--pdf/--video/--file), the
+ * batch `<promptsFile>` positional, and --csv-max-rows before a
+ * generate/stream/batch run starts, so bad input fails fast with a clear
+ * message instead of surfacing deep inside provider calls. Exported as
+ * standalone functions (rather than private statics on CLICommandFactory)
+ * so callers — including tests — can import and invoke them directly,
+ * without reaching into the class via an `unknown` cast.
  */
 
 import fs from "node:fs";
+import { fileURLToPath } from "node:url";
 import chalk from "chalk";
 import { logger } from "../../lib/utils/logger.js";
-import { resolveFilePaths } from "./pathResolver.js";
+import { redactUrlForError } from "../../lib/utils/logSanitize.js";
+import type { CliValidatedFileOption } from "../../lib/types/index.js";
+import { resolveFilePath } from "./pathResolver.js";
 
 /**
  * Soft (warn-only) size thresholds for CLI multimodal flags. Intentionally
@@ -29,25 +33,157 @@ export const CLI_SOFT_LIMITS_MB = {
   VIDEO_MAX_MB: 500,
 } as const;
 
+// `file://` is intentionally excluded here: it addresses a local path, so it
+// must still go through directory/missing-path validation below (via
+// tryResolveFileUrl in validateSingleLocalPath). Only genuine remote/inline
+// references (http(s), data URIs) skip validation entirely (#319).
 function isNonLocalFileReference(filePath: string): boolean {
   const lower = filePath.toLowerCase();
   return (
     lower.startsWith("http://") ||
     lower.startsWith("https://") ||
-    lower.startsWith("file://") ||
     lower.startsWith("data:")
   );
 }
 
+// #291: a `file://` URL still points at something on disk, so validation
+// needs the real filesystem path. Returns undefined (instead of throwing)
+// for a malformed URL so the caller can report a friendly error.
+function tryResolveFileUrl(filePath: string): string | undefined {
+  if (!filePath.toLowerCase().startsWith("file://")) {
+    return filePath;
+  }
+  try {
+    return fileURLToPath(filePath);
+  } catch {
+    return undefined;
+  }
+}
+
+/**
+ * Shared aggregated-error format for any invalid local file path (missing,
+ * unreadable, a directory, or a malformed `file://` URL). Used by both the
+ * flag-based loop in `validateCliInputFiles` and the batch `<promptsFile>`
+ * positional check in `validatePromptsFilePath`, so every caller gets the
+ * same clear error + troubleshooting hints instead of a raw exception.
+ */
+function buildInvalidFileError(errors: string[]): Error {
+  return new Error(
+    "❌ One or more input files are invalid:\n" +
+      `${errors.join("\n")}\n` +
+      "💡 Troubleshooting:\n" +
+      "  • Check the path is correct and relative to your current directory, or pass an absolute path / URL.\n" +
+      `  • Check your working directory (${process.cwd()}).\n` +
+      "  • For remote resources, prefix the value with http:// or https://.\n" +
+      "  • If you pointed at a folder, point at the file inside it instead.",
+  );
+}
+
+/**
+ * Validate a single raw path/URL value for `option`, pushing a friendly
+ * message onto `errors` instead of letting fs calls throw a raw exception
+ * (ENOENT, EACCES/EPERM TOCTOU, a directory, or a malformed `file://` URL).
+ * Returns the resolved local filesystem path on success (or the original
+ * value unchanged for a skipped http(s)/data: reference), and `undefined`
+ * when an error was pushed.
+ *
+ * `rawPath` is redacted via `redactUrlForError` before ever being
+ * interpolated into a message — a `file://` value can carry a sensitive
+ * query string, fragment, or embedded `user:pass@` credentials that must
+ * never be echoed back verbatim. This is a no-op for ordinary filesystem
+ * paths (they aren't parseable as a URL, so the redaction fallback simply
+ * returns them unchanged).
+ */
+function validateSingleLocalPath(
+  option: string,
+  rawPath: string,
+  errors: string[],
+  warnAtMB?: number,
+): string | undefined {
+  const resolvedPath = resolveFilePath(rawPath);
+  if (isNonLocalFileReference(resolvedPath)) {
+    // URLs / data: refs skip both existence and size checks (#319).
+    return resolvedPath;
+  }
+
+  const safeRawPath = redactUrlForError(rawPath);
+
+  // #291: `file://` URLs are local paths in disguise — validate the decoded
+  // filesystem path so a directory or missing path is still caught here
+  // instead of failing cryptically deep in processing.
+  const pathToValidate = tryResolveFileUrl(resolvedPath);
+  if (pathToValidate === undefined) {
+    errors.push(`${option} is not a valid file:// URL: ${safeRawPath}`);
+    return undefined;
+  }
+
+  const safeResolvedPath = redactUrlForError(pathToValidate);
+
+  // Single statSync covers existence, directory rejection, and the size
+  // warning below — avoids a second syscall and closes the TOCTOU gap a
+  // separate existsSync()+statSync() pair would leave.
+  let stat: fs.Stats;
+  try {
+    stat = fs.statSync(pathToValidate);
+  } catch (err) {
+    // Only ENOENT actually means "not found" — EACCES (permission denied),
+    // ELOOP (symlink loop), ENOTDIR, etc. are real failures that get masked
+    // (and made undebuggable) if we blanket-label every statSync error the
+    // same way.
+    const code = (err as NodeJS.ErrnoException)?.code;
+    if (code === "ENOENT") {
+      errors.push(
+        `${option} path not found: ${safeRawPath} (resolved to ${safeResolvedPath})`,
+      );
+    } else {
+      const reason = err instanceof Error ? err.message : String(err);
+      errors.push(
+        `${option} path could not be accessed: ${safeRawPath} (resolved to ${safeResolvedPath}) — ` +
+          `${code ? `${code}: ` : ""}${reason}`,
+      );
+    }
+    return undefined;
+  }
+
+  // #288: reject a directory before it fails cryptically deep in processing
+  // (readFileSync on a dir throws a raw EISDIR).
+  if (stat.isDirectory()) {
+    errors.push(
+      `${option} path is a directory, not a file: ${safeRawPath} (resolved to ${safeResolvedPath})`,
+    );
+    return undefined;
+  }
+
+  // #319: warn (not error) when a local multimodal file is unusually large
+  // so the user isn't surprised by slow processing / token blowups. Safety-
+  // relevant, so always visible (not gated behind --quiet) and written to
+  // stderr — never stdout — so `--format json` output stays parseable.
+  if (warnAtMB !== undefined) {
+    const sizeMB = stat.size / (1024 * 1024);
+    if (sizeMB > warnAtMB) {
+      logger.alwaysStderr(
+        chalk.yellow(
+          `⚠️  ${option} file ${safeRawPath} is ${sizeMB.toFixed(1)}MB ` +
+            `(above the ${warnAtMB}MB soft limit) — processing may be slow ` +
+            `or exceed provider token/size limits.`,
+        ),
+      );
+    }
+  }
+
+  return pathToValidate;
+}
+
 /**
  * Validate --image/--csv/--pdf/--video/--file inputs before a CLI run
- * starts: every local path must exist, must not be a directory, and — for
- * the flags with a soft limit — triggers a size warning above the
- * threshold. URLs/data URIs skip both existence and size checks (#319).
+ * starts: every local path (including `file://` URLs) must exist, must not
+ * be a directory, and — for the flags with a soft limit — triggers a size
+ * warning above the threshold. Genuine http(s)/data: URLs skip both
+ * existence and size checks (#319).
  */
 export function validateCliInputFiles(argv: Record<string, unknown>): void {
   const fileArgs: Array<{
-    option: "--image" | "--csv" | "--pdf" | "--video" | "--file";
+    option: CliValidatedFileOption;
     value?: string | string[];
     warnAtMB?: number;
   }> = [
@@ -74,7 +210,7 @@ export function validateCliInputFiles(argv: Record<string, unknown>): void {
     { option: "--file", value: argv.file as string | string[] | undefined },
   ];
 
-  const missingPaths: string[] = [];
+  const errors: string[] = [];
 
   for (const { option, value, warnAtMB } of fileArgs) {
     if (!value) {
@@ -82,73 +218,35 @@ export function validateCliInputFiles(argv: Record<string, unknown>): void {
     }
 
     const rawPaths = Array.isArray(value) ? value : [value];
-    const resolvedPaths = resolveFilePaths(rawPaths);
-
-    for (let i = 0; i < resolvedPaths.length; i++) {
-      const resolvedPath = resolvedPaths[i];
-      // URLs / data: refs skip both existence and size checks (#319).
-      if (isNonLocalFileReference(resolvedPath)) {
-        continue;
-      }
-
-      // Single statSync covers existence, directory rejection, and the
-      // size warning below — avoids a second syscall and closes the
-      // TOCTOU gap a separate existsSync()+statSync() pair would leave.
-      let stat: fs.Stats;
-      try {
-        stat = fs.statSync(resolvedPath);
-      } catch (err) {
-        // Only ENOENT actually means "not found" — EACCES (permission
-        // denied), ELOOP (symlink loop), ENOTDIR, etc. are real failures
-        // that get masked (and made undebuggable) if we blanket-label
-        // every statSync error the same way.
-        const code = (err as NodeJS.ErrnoException)?.code;
-        if (code === "ENOENT") {
-          missingPaths.push(
-            `${option} path not found: ${rawPaths[i]} (resolved to ${resolvedPath})`,
-          );
-        } else {
-          const reason = err instanceof Error ? err.message : String(err);
-          missingPaths.push(
-            `${option} path could not be accessed: ${rawPaths[i]} (resolved to ${resolvedPath}) — ` +
-              `${code ? `${code}: ` : ""}${reason}`,
-          );
-        }
-        continue;
-      }
-
-      if (stat.isDirectory()) {
-        missingPaths.push(
-          `${option} path is a directory, not a file: ${rawPaths[i]} (resolved to ${resolvedPath})`,
-        );
-        continue;
-      }
-
-      // #319: warn (not error) when a local multimodal file is unusually
-      // large so the user isn't surprised by slow processing / token blowups.
-      if (warnAtMB !== undefined) {
-        const sizeMB = stat.size / (1024 * 1024);
-        if (sizeMB > warnAtMB) {
-          logger.always(
-            chalk.yellow(
-              `⚠️  ${option} file ${rawPaths[i]} is ${sizeMB.toFixed(1)}MB ` +
-                `(above the ${warnAtMB}MB soft limit) — processing may be slow ` +
-                `or exceed provider token/size limits.`,
-            ),
-          );
-        }
-      }
+    for (const rawPath of rawPaths) {
+      validateSingleLocalPath(option, rawPath, errors, warnAtMB);
     }
   }
 
-  if (missingPaths.length > 0) {
-    throw new Error(
-      "❌ One or more input files are invalid:\n" +
-        `${missingPaths.join("\n")}\n` +
-        "💡 Check the path is correct and relative to your current directory, " +
-        "or pass an absolute path / URL.",
-    );
+  if (errors.length > 0) {
+    throw buildInvalidFileError(errors);
   }
+}
+
+/**
+ * #291: validate the batch `<promptsFile>` positional through the same
+ * aggregated-error-friendly path as the multimodal flags above (not found /
+ * unreadable / directory / malformed `file://` URL), instead of a raw
+ * existsSync+readFileSync that throws an unfriendly EISDIR when pointed at
+ * a directory. The `<file>` label (rather than `--file`) keeps the message
+ * from being confused with the `--file` flag, which batch doesn't register
+ * at all (see `createBatchCommand`). Returns the resolved local path for
+ * the caller to read.
+ */
+export function validatePromptsFilePath(rawPath: string): string {
+  const errors: string[] = [];
+  const resolved = validateSingleLocalPath("<file>", rawPath, errors);
+  if (errors.length > 0) {
+    throw buildInvalidFileError(errors);
+  }
+  // `errors` is empty here, so `validateSingleLocalPath` did not return
+  // undefined.
+  return resolved as string;
 }
 
 /**
--- a/src/lib/types/cli.ts
+++ b/src/lib/types/cli.ts
@@ -161,8 +161,13 @@ export type StreamCommandArgs = BaseCommandArgs &
 export type BatchCommandArgs = BaseCommandArgs &
   CliToolRoutingFlags &
   CliClassifierRouterFlags & {
-    /** Input file path */
-    file?: string;
+    // #1191 round-5: named `promptsFile`, not `file` — yargs implicitly
+    // treats a positional's key as a recognized flag name too, so keeping
+    // this "file" would let `--file` silently pass strictOptions() despite
+    // not being a registered batch option, colliding with the (unrelated)
+    // common `--file` auto-detect flag that batch intentionally omits.
+    /** Prompts-list file path (the `<promptsFile>` positional) */
+    promptsFile?: string;
     /** AI provider to use */
     provider?: string;
     /** Model name */
@@ -2002,3 +2007,18 @@ export type CliAudioPlayerCommand = {
   command: string;
   args: string[];
 };
+
+// =============================================================================
+// CLI INPUT VALIDATION (from cli/utils/inputValidation.ts)
+// =============================================================================
+
+/**
+ * The CLI flags validated by `validateCliInputFiles` before a
+ * generate/stream/batch run starts (--image/--csv/--pdf/--video/--file).
+ */
+export type CliValidatedFileOption =
+  | "--image"
+  | "--csv"
+  | "--pdf"
+  | "--video"
+  | "--file";
--- a/test/continuous-test-suite-bugfixes.ts
+++ b/test/continuous-test-suite-bugfixes.ts
@@ -43,6 +43,7 @@ import { ImageProcessor, imageUtils } from "../src/lib/utils/imageProcessor.js";
 import { ERROR_CODES } from "../src/lib/utils/errorHandling.js";
 import fs, {
   chmodSync,
+  existsSync,
   mkdtempSync,
   readFileSync,
   rmSync,
@@ -51,9 +52,59 @@ import fs, {
 } from "node:fs";
 import { tmpdir } from "node:os";
 import { basename, join as pathJoin, resolve as resolvePath } from "node:path";
+import { pathToFileURL } from "node:url";
+import { spawn as spawnProcess } from "node:child_process";
 import { MockAgent, setGlobalDispatcher, getGlobalDispatcher } from "undici";
 import http from "node:http";
 import AdmZip from "adm-zip";
+
+const CLI_DIST_PATH = pathJoin(process.cwd(), "dist", "cli", "index.js");
+
+/**
+ * Spawn the built CLI (dist/cli/index.js) and capture exit code + separate
+ * stdout/stderr (plus `out`, the combined stream, for tests that don't care
+ * which stream a message landed on). Used by the CLI-file-validation E2E
+ * tests (#288/#291) — requires a prior `pnpm run build`.
+ */
+function runCliBugfix(
+  args: string[],
+  timeoutMs = 60_000,
+): Promise<{
+  code: number | null;
+  out: string;
+  stdout: string;
+  stderr: string;
+}> {
+  return new Promise((resolve, reject) => {
+    if (!existsSync(CLI_DIST_PATH)) {
+      const message = `dist/cli/index.js not found at ${CLI_DIST_PATH} — run 'pnpm run build:cli' first`;
+      // Surface on stderr too so it's visible when the suite is run directly,
+      // not just via the unhandled-rejection exit path.
+      process.stderr.write(`${message}\n`);
+      reject(new Error(message));
+      return;
+    }
+    const child = spawnProcess(process.execPath, [CLI_DIST_PATH, ...args], {
+      env: { ...process.env, NO_COLOR: "1" },
+    });
+    let stdout = "";
+    let stderr = "";
+    child.stdout.on("data", (d) => (stdout += String(d)));
+    child.stderr.on("data", (d) => (stderr += String(d)));
+    const timer = setTimeout(() => {
+      child.kill("SIGKILL");
+      reject(new Error("CLI timed out"));
+    }, timeoutMs);
+    child.on("close", (code) => {
+      clearTimeout(timer);
+      resolve({ code, out: stdout + stderr, stdout, stderr });
+    });
+    child.on("error", (error) => {
+      clearTimeout(timer);
+      reject(error);
+    });
+  });
+}
 import { PptxProcessor } from "../src/lib/processors/document/PptxProcessor.js";
 import {
   validateCliInputFiles,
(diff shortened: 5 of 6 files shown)


src/cli/factories/commandFactory.ts
```diff
@@ -939,16 +948,48 @@ export class CLICommandFactory {
...
+
+        // #288: reject a directory before it fails cryptically deep in
+        // processing (readFileSync on a dir throws a raw EISDIR).
+        const stat = fs.statSync(resolvedPath);
+        if (stat.isDirectory()) {
+          errors.push(
+            `${option} path is a directory, not a file: ${rawPaths[i]}`,
+          );
+          continue;
+        }
```
Tara-ag: 🔒 **CRITICAL** — `fs.statSync(resolvedPath)` can throw after `existsSync` returns true (race condition, permission denied, etc.). The raw exception bypasses the aggregated error + troubleshooting output.

**Suggested fix:** wrap the `statSync` call in a try/catch and treat stat failures as invalid inputs with the same troubleshooting block:

```ts
let stat: fs.Stats;
try {
  stat = fs.statSync(resolvedPath);
} catch (err) {
  errors.push(`${option} path cannot be read: ${rawPaths[i]} (${err instanceof Error ? err.message : String(err)})`);
  continue;
}
if (stat.isDirectory()) { ... }
```

src/cli/factories/commandFactory.ts
```diff
@@ -939,16 +948,48 @@ export class CLICommandFactory {
...
     }
 
-    if (missingPaths.length > 0) {
+    // Warnings must be visible without --debug (logger.warn is hidden then),
+    // so print via logger.always unless the user asked for quiet output.
+    if (warnings.length > 0 && !argv.quiet) {
+      for (const w of warnings) {
+        logger.always(chalk.yellow(`⚠️  ${w}`));
+      }
+    }
```
Tara-ag: ⚠️ **MAJOR** — `logger.always` writes to **stdout** (`console.log`). When `--format json` is used, this prepends non-JSON text to stdout and breaks machine-readable output/redirects.

**Suggested fix:** emit warnings to **stderr** instead. Use `console.warn`/`process.stderr.write` or add a `logger.alwaysError` helper so JSON consumers remain unaffected:

```ts
process.stderr.write(chalk.yellow(`⚠️  ${w}\n`));
```

Same concern applies to the batch attachment notice at line 4716.

src/cli/factories/commandFactory.ts
```diff
@@ -4643,6 +4684,42 @@ export class CLICommandFactory {
       const enhancedOptions = { ...options, ...sessionVariables };
       const sessionId = globalSession.getCurrentSessionId();
 
+      // #291: batch parsed --image/--csv/--pdf/--video but never attached them.
+      // Resolve them once (they apply to every prompt) and fail fast on bad
+      // paths, mirroring generate/stream. NOTE: --file is the batch <file>
+      // positional (the prompts list), so auto-detect is not wired here.
+      CLICommandFactory.validateCliInputFiles(
+        argv as unknown as Record<string, unknown>,
+      );
```
Tara-ag: ⚠️ **MAJOR** — In `batch`, `argv.file` is the prompts-list positional, but `validateCliInputFiles` treats it as the `--file` (auto-detect) flag. If validation fails, the error message will misleadingly reference `--file` even though `--file` is documented as unavailable for `batch`.

**Suggested fix:** exclude the positional from multimodal validation in the batch path, or pass an explicit allowlist of options to `validateCliInputFiles` so it does not inspect `argv.file`:

```ts
CLICommandFactory.validateCliInputFiles(argv as unknown as Record<string, unknown>, { excludeFile: true });
```

src/cli/factories/commandFactory.ts
```diff
@@ -4643,6 +4684,42 @@ export class CLICommandFactory {
...
+      const batchCsvFiles = CLICommandFactory.processCliCSVFiles(
+        argv.csv as string | string[] | undefined,
+      );
+      const batchPdfFiles = CLICommandFactory.processCliPDFFiles(
+        argv.pdf as string | string[] | undefined,
+      );
+      const batchVideoFiles = CLICommandFactory.processCliVideoFiles(
+        argv.video as string | string[] | undefined,
+      );
+      if (
```
Tara-ag, line 4744: 💡 **MINOR** — The batch attachment notice is gated on `!options.quiet`, but `quiet` defaults to `true` in the CLI, so ordinary invocations won't see it. The docs claim "a one-line notice is printed". Either make it visible by default or update the docs to say users need `--no-quiet`/`--output` to see it.

src/cli/factories/commandFactory.ts
```diff
@@ -4643,6 +4684,42 @@ export class CLICommandFactory {
...
+        (batchImages?.length ||
+          batchCsvFiles?.length ||
+          batchPdfFiles?.length ||
+          batchVideoFiles?.length) &&
+        !options.quiet
+      ) {
+        // Pause the spinner so the notice isn't swallowed by its re-renders.
+        spinner?.stop();
+        logger.always(
+          chalk.yellow(
```
Tara-ag, line 4755: 💡 **MINOR** — Same stdout/stderr concern as the size warnings: `logger.always` goes to stdout. If this notice is meant to be visible by default, consider routing it to stderr so it doesn't corrupt `--format json` output.

src/cli/factories/commandFactory.ts
```diff
@@ -4692,7 +4769,27 @@ export class CLICommandFactory {
 
           const runBatchGenerate = () =>
             sdk.generate({
-              input: { text: inputText },
+              input: {
```
Tara-ag, line 4811: 💡 **MINOR** — `csvOptions` and `videoOptions` are constructed even when no CSV/video files are attached. Passing empty option objects to `sdk.generate` is likely harmless, but it would be cleaner to only include them when the corresponding file arrays are present, keeping the generated request minimal.

docs/features/multimodal.md
```diff
@@ -420,6 +420,25 @@ for await (const chunk of stream) {
...
 
+### Batch with Multimodal (CLI)
+
+The `batch` command supports `--image`, `--csv`, `--pdf`, and `--video`. The file(s) are attached **identically to every prompt** in the batch (a one-line notice is printed):
+
+```bash
+neurolink batch questions.txt --csv sales.csv --format json
+```
+
+> `--file` (auto-detect) is **not** available in `batch`, because it collides with the `<file>` positional (the prompts-list path). Use the explicit `--image` / `--csv` / `--pdf` / `--video` flags instead.
```
Tara-ag, line 435: 💡 **MINOR** — Existing review already notes this: the notice is suppressed by default `quiet: true`. Either update the docs to state the required output flag (`--no-quiet`/`--output`) or change the implementation so the notice is emitted during ordinary invocations.

docs/features/multimodal.md
```diff
@@ -420,6 +420,25 @@ for await (const chunk of stream) {
...
+
+### File validation & troubleshooting (CLI)
+
+Before any provider call, the CLI validates local file inputs across `generate`, `stream`, and `batch`:
+
+- A path that points at a **directory** (or doesn't exist) is rejected up front with a clear error and troubleshooting hints — no cryptic `EISDIR` deep in processing.
+- A **large** file (images > 10 MB, CSV/PDF > 50 MB) prints a non-blocking warning (visible without `--debug`).
+
+This runs even under `--dry-run` and without API keys configured.
```
Tara-ag, line 444: 💡 **MINOR** — Existing review already notes this: with `quiet` defaulting to `true`, these warnings are not visible by default. Clarify that users must disable quiet mode to see them, or make the warnings unconditional.

test/continuous-test-suite-bugfixes.ts
```diff
@@ -28,7 +28,41 @@ import { FileDetector } from "../src/lib/utils/fileDetector.js";
...
+function runCliBugfix(
+  args: string[],
+  timeoutMs = 60_000,
+): Promise<{ code: number | null; out: string }> {
+  return new Promise((resolve, reject) => {
+    const child = spawnProcess(
+      process.execPath,
+      [pathJoin(process.cwd(), "dist", "cli", "index.js"), ...args],
+      { env: { ...process.env, NO_COLOR: "1" } },
+    );
```
Tara-ag: 💡 **MINOR** — Existing review already notes this: `runCliBugfix` assumes `dist/cli/index.js` exists. If the suite runs without a prior CLI build, the spawned process fails with a generic Node module-not-found error. Add an explicit existence check with a clear failure message before spawning.

test/continuous-test-suite-bugfixes.ts
```diff
@@ -3850,6 +3884,82 @@ exit 127
...
+          "--dry-run",
+          "--format",
+          "json",
+        ]);
+        const s = ok.out.indexOf("[");
+        const e = ok.out.lastIndexOf("]");
+        if (s < 0 || e < 0) {
+          return false;
+        }
+        const arr = JSON.parse(ok.out.slice(s, e + 1));
```
Tara-ag: 💡 **MINOR** — The test parses JSON by searching for the first `[` and last `]` in combined stdout+stderr. Because the batch attachment notice and any future warnings may be written to stdout (see earlier comment), this brittle substring extraction could include non-JSON text or fail if the CLI emits leading/trailing log lines. Consider asserting on structured output more robustly (e.g., redirecting notices/warnings to stderr and parsing only stdout, or using `--quiet` consistently).

Tara-ag (changes requested):
## Review Summary

Reviewed 3 changed files in PR #1191.

| Severity | Count | Files |
|----------|-------|-------|
| 🔒 CRITICAL | 1 | `src/cli/factories/commandFactory.ts` |
| ⚠️ MAJOR | 2 | `src/cli/factories/commandFactory.ts` |
| 💡 MINOR | 7 | `src/cli/factories/commandFactory.ts`, `docs/features/multimodal.md`, `test/continuous-test-suite-bugfixes.ts` |

### Blocking rationale

- **CRITICAL:** `fs.statSync` after `existsSync` is unguarded and can throw, bypassing the intended aggregated error + troubleshooting output. This is a correctness/security-adjacent failure mode (raw exception instead of controlled error).
- **MAJOR:** `logger.always` writes size warnings to **stdout**, which corrupts `--format json` machine-readable output.
- **MAJOR:** `batch` passes `argv.file` (the prompts-list positional) into `validateCliInputFiles`, where it is treated as the `--file` auto-detect flag, producing misleading error messages.

These meet the blocking criteria (critical security/correctness risk + major architectural/UX issues). Once addressed, the PR should be good to merge.

### Next steps
1. Wrap `fs.statSync` in try/catch and surface stat failures as invalid inputs with the troubleshooting block.
2. Route warnings/notices to **stderr** instead of stdout, or introduce a stderr-aware unconditional logger helper.
3. Exclude the `batch` positional `argv.file` from multimodal validation.
4. Resolve the doc/implementation mismatch around default visibility of warnings and the batch attachment notice.
5. Add a pre-spawn existence check for `dist/cli/index.js` in the bugfix test helper.

docs/features/multimodal.md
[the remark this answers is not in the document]
murdore: Fixed by changing the implementation rather than the docs, so the two stay in sync going forward.

The batch attachment notice (`src/cli/factories/commandFactory.ts:4772-4788`) is no longer gated on `!options.quiet` — it now fires unconditionally on every batch run that attaches multimodal files, and uses the new `logger.alwaysStderr()` (writes to `console.error`, see `src/lib/utils/logger.ts:399-411`) instead of `logger.always()` (stdout), so it never corrupts `--format json` output.

Docs updated at `docs/features/multimodal.md:425` to describe the actual behavior: *"a one-line notice is printed to **stderr**, unconditionally — it is not suppressed by `--quiet`, so it never corrupts `--format json` output written to stdout"*.

Design choice: rather than documenting the suppressed-by-default behavior (which would mean users silently miss a safety-relevant notice about files being attached to every prompt in the batch), we made the notice always-visible on stderr. This keeps stdout clean for `--format json` while guaranteeing the notice isn't accidentally hidden by the `quiet: true` default. Same treatment applied to the large-file size warnings in `validateCliInputFiles` (`commandFactory.ts:1031-1038`) — see the sibling thread on `docs/features/multimodal.md:440`.

Regression test added: "CLI: batch --format json stdout stays valid JSON even though the multimodal attach notice fires (notice goes to stderr)" (`test/continuous-test-suite-bugfixes.ts:3985`) — asserts `JSON.parse(stdout)` succeeds while the notice is only present in captured stderr.

src/cli/factories/commandFactory.ts
```diff
@@ -939,16 +948,48 @@ export class CLICommandFactory {
...
       }
     }
 
-    if (missingPaths.length > 0) {
+    // Warnings must be visible without --debug (logger.warn is hidden then),
+    // so print via logger.always unless the user asked for quiet output.
+    if (warnings.length > 0 && !argv.quiet) {
+      for (const w of warnings) {
+        logger.always(chalk.yellow(`⚠️  ${w}`));
+      }
```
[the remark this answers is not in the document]
murdore: Fixed — removed the `!quiet`/`!options.quiet` gate entirely rather than just relabeling it, and moved the output to stderr:

- `validateCliInputFiles` size warnings: now unconditional, via `logger.alwaysStderr()` (`src/cli/factories/commandFactory.ts:1031-1038`).
- `executeBatch` multimodal attachment notice: now unconditional, via `logger.alwaysStderr()` (`src/cli/factories/commandFactory.ts:4772-4788`).

Both bypass log-level/debug gating (same semantics as the existing `logger.always()`) but write to `console.error` instead of `console.log` — see the new `alwaysStderr()` method on `NeuroLinkLogger` (`src/lib/utils/logger.ts:399-411`, exported at `logger.ts:517-519`).

Rationale for "unconditional + stderr" over "leave quiet-gated, fix docs": `quiet` defaults to `true` on this CLI, so a quiet-gated safety warning is invisible on every ordinary invocation — that defeats the point of warning at all. Moving it to stderr means it can be unconditional without ever corrupting `--format json` (or any other machine-parseable) stdout output.

`docs/features/multimodal.md` was updated to describe this actual behavior (lines 425, 437-438). Regression coverage: "CLI: batch --format json stdout stays valid JSON even though the multimodal attach notice fires (notice goes to stderr)" (`test/continuous-test-suite-bugfixes.ts:3985`).

src/cli/factories/commandFactory.ts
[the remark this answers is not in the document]
murdore: Fixed — added a dedicated `logger.alwaysStderr()` method (`src/lib/utils/logger.ts:399-411`) that mirrors `logger.always()`'s semantics (bypasses log-level/debug gating, unconditional) but calls `console.error` instead of `console.log`, so it targets stderr. Exported alongside `always` at `logger.ts:517-519`.

Both warning sites that previously used `logger.always()` now use `logger.alwaysStderr()`:
- Size warnings in `validateCliInputFiles` — `src/cli/factories/commandFactory.ts:1036`.
- Batch multimodal attachment notice in `executeBatch` — `commandFactory.ts:4782-4786`.

(We deliberately did not reuse `logger.error()` for this — that would prefix messages with `[NEUROLINK:ERROR]` and pollute log-level semantics/history for what are advisory notices, not errors.)

Regression test added: "CLI: batch --format json stdout stays valid JSON even though the multimodal attach notice fires (notice goes to stderr)" (`test/continuous-test-suite-bugfixes.ts:3985`) — spawns the built CLI with `--format json`, captures stdout/stderr separately, and asserts `JSON.parse(stdout)` succeeds while the notice text is only present in stderr.

src/cli/factories/commandFactory.ts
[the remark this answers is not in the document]
murdore: Fixed — extracted a shared `validateSingleFilePath()` helper (`src/cli/factories/commandFactory.ts:930-973`) that wraps `fs.statSync(resolvedPath)` in try/catch:

```ts
let stat: fs.Stats;
try {
  stat = fs.statSync(resolvedPath);
} catch (err) {
  errors.push(
    `${option} path cannot be read: ${rawPath} (${err instanceof Error ? err.message : String(err)})`,
  );
  return;
}
```

Instead of letting a raw `EACCES`/`EPERM`/TOCTOU exception escape, it pushes a friendly message onto the same `errors[]` array used for missing/directory paths, which then flows into the existing aggregated error + troubleshooting output via `buildInvalidFileError()` (`commandFactory.ts:919-928`). This helper is shared by both the `--image/--csv/--pdf/--video` flag loop in `validateCliInputFiles` (`commandFactory.ts:1020-1027`) and the batch `<file>` prompts-list positional check in `executeBatch` (`commandFactory.ts:4681-4688`), so every caller gets the same protection.

Regression test added: "CLI #288: fs.statSync throwing after existsSync succeeds is caught and reported via the aggregated error, not a raw exception" (`test/continuous-test-suite-bugfixes.ts:4025`). It monkeypatches `fs.statSync` in-process (verified this propagates correctly across ESM module boundaries since both the test and `commandFactory.ts` access `fs.statSync` via property access on the same imported `fs` default-export object) and calls `validateCliInputFiles` directly via a `CLICommandFactory as unknown as {...}` cast to confirm the thrown error is the friendly aggregated one, not the raw `fs` exception.

src/cli/factories/commandFactory.ts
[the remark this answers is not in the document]
murdore: Fixed — `executeBatch` now validates the `<file>` prompts-list positional on its own, with the correct `"<file>"` label, *before* the shared flag-based validation runs:

```ts
const resolvedPromptsPath = resolveFilePaths([argv.file])[0];
const promptsFileErrors: string[] = [];
CLICommandFactory.validateSingleFilePath(
  "<file>",
  argv.file,
  resolvedPromptsPath,
  undefined,
  promptsFileErrors,
  [],
);
if (promptsFileErrors.length > 0) {
  throw CLICommandFactory.buildInvalidFileError(promptsFileErrors);
}
```
(`src/cli/factories/commandFactory.ts:4679-4691`)

`validateCliInputFiles()` gained an `excludeOptions` parameter so the batch's subsequent flag-based validation (for `--image/--csv/--pdf/--video`) never re-inspects `argv.file` under the `--file` label:

```ts
CLICommandFactory.validateCliInputFiles(
  argv as unknown as Record<string, unknown>,
  { excludeOptions: ["--file"] },
);
```
(`commandFactory.ts:4756-4759`, implementation at `commandFactory.ts:975-994`)

Result: an invalid batch `<file>` path now always reports as `<file> path ...`, never `--file path ...`, and a directory target goes through the same friendly aggregated-error + troubleshooting path instead of a raw `EISDIR`.

Regression test added: "CLI #291: batch <file> positional pointing at a directory gets the friendly aggregated error, not a raw EISDIR" (`test/continuous-test-suite-bugfixes.ts:4068`).

test/continuous-test-suite-bugfixes.ts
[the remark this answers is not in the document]
murdore: Fixed — `runCliBugfix` now checks the built CLI exists before spawning it:

```ts
const CLI_DIST_PATH = pathJoin(process.cwd(), "dist", "cli", "index.js");
...
if (!existsSync(CLI_DIST_PATH)) {
  reject(
    new Error(
      `dist/cli/index.js not found at ${CLI_DIST_PATH} — run 'pnpm run build:cli' first`,
    ),
  );
  return;
}
```
(`test/continuous-test-suite-bugfixes.ts:41`, guard at `:59-65`)

Running the suite without a prior build now fails immediately with an actionable message instead of the generic Node module-not-found error from the spawned child process.

src/cli/factories/commandFactory.ts
murdore: Fixed — extracted a shared `validateSingleFilePath()` helper (`src/cli/factories/commandFactory.ts:930-973`) that wraps `fs.statSync(resolvedPath)` in try/catch:

```ts
let stat: fs.Stats;
try {
  stat = fs.statSync(resolvedPath);
} catch (err) {
  errors.push(
    `${option} path cannot be read: ${rawPath} (${err instanceof Error ? err.message : String(err)})`,
  );
  return;
}
```

Instead of letting a raw `EACCES`/`EPERM`/TOCTOU exception escape, it pushes a friendly message onto the same `errors[]` array used for missing/directory paths, which then flows into the existing aggregated error + troubleshooting output via `buildInvalidFileError()` (`commandFactory.ts:919-928`). This helper is shared by both the `--image/--csv/--pdf/--video` flag loop in `validateCliInputFiles` (`commandFactory.ts:1020-1027`) and the batch `<file>` prompts-list positional check in `executeBatch` (`commandFactory.ts:4681-4688`), so every caller gets the same protection.

Regression test added: "CLI #288: fs.statSync throwing after existsSync succeeds is caught and reported via the aggregated error, not a raw exception" (`test/continuous-test-suite-bugfixes.ts:4025`). It monkeypatches `fs.statSync` in-process (verified this propagates correctly across ESM module boundaries since both the test and `commandFactory.ts` access `fs.statSync` via property access on the same imported `fs` default-export object) and calls `validateCliInputFiles` directly via a `CLICommandFactory as unknown as {...}` cast to confirm the thrown error is the friendly aggregated one, not the raw `fs` exception.

src/cli/factories/commandFactory.ts
murdore: Fixed — added a dedicated `logger.alwaysStderr()` method (`src/lib/utils/logger.ts:399-411`) that mirrors `logger.always()`'s semantics (bypasses log-level/debug gating, unconditional) but calls `console.error` instead of `console.log`, so it targets stderr. Exported alongside `always` at `logger.ts:517-519`.

Both warning sites that previously used `logger.always()` now use `logger.alwaysStderr()`:
- Size warnings in `validateCliInputFiles` — `src/cli/factories/commandFactory.ts:1036`.
- Batch multimodal attachment notice in `executeBatch` — `commandFactory.ts:4782-4786`.

(We deliberately did not reuse `logger.error()` for this — that would prefix messages with `[NEUROLINK:ERROR]` and pollute log-level semantics/history for what are advisory notices, not errors.)

Regression test added: "CLI: batch --format json stdout stays valid JSON even though the multimodal attach notice fires (notice goes to stderr)" (`test/continuous-test-suite-bugfixes.ts:3985`) — spawns the built CLI with `--format json`, captures stdout/stderr separately, and asserts `JSON.parse(stdout)` succeeds while the notice text is only present in stderr.

src/cli/factories/commandFactory.ts
murdore: Fixed — `executeBatch` now validates the `<file>` prompts-list positional on its own, with the correct `"<file>"` label, *before* the shared flag-based validation runs:

```ts
const resolvedPromptsPath = resolveFilePaths([argv.file])[0];
const promptsFileErrors: string[] = [];
CLICommandFactory.validateSingleFilePath(
  "<file>",
  argv.file,
  resolvedPromptsPath,
  undefined,
  promptsFileErrors,
  [],
);
if (promptsFileErrors.length > 0) {
  throw CLICommandFactory.buildInvalidFileError(promptsFileErrors);
}
```
(`src/cli/factories/commandFactory.ts:4679-4691`)

`validateCliInputFiles()` gained an `excludeOptions` parameter so the batch's subsequent flag-based validation (for `--image/--csv/--pdf/--video`) never re-inspects `argv.file` under the `--file` label:

```ts
CLICommandFactory.validateCliInputFiles(
  argv as unknown as Record<string, unknown>,
  { excludeOptions: ["--file"] },
);
```
(`commandFactory.ts:4756-4759`, implementation at `commandFactory.ts:975-994`)

Result: an invalid batch `<file>` path now always reports as `<file> path ...`, never `--file path ...`, and a directory target goes through the same friendly aggregated-error + troubleshooting path instead of a raw `EISDIR`.

Regression test added: "CLI #291: batch <file> positional pointing at a directory gets the friendly aggregated error, not a raw EISDIR" (`test/continuous-test-suite-bugfixes.ts:4068`).

src/cli/factories/commandFactory.ts
murdore: Fixed — removed the `!quiet`/`!options.quiet` gate entirely rather than just relabeling it, and moved the output to stderr:

- `validateCliInputFiles` size warnings: now unconditional, via `logger.alwaysStderr()` (`src/cli/factories/commandFactory.ts:1031-1038`).
- `executeBatch` multimodal attachment notice: now unconditional, via `logger.alwaysStderr()` (`src/cli/factories/commandFactory.ts:4772-4788`).

Both bypass log-level/debug gating (same semantics as the existing `logger.always()`) but write to `console.error` instead of `console.log` — see the new `alwaysStderr()` method on `NeuroLinkLogger` (`src/lib/utils/logger.ts:399-411`, exported at `logger.ts:517-519`).

Rationale for "unconditional + stderr" over "leave quiet-gated, fix docs": `quiet` defaults to `true` on this CLI, so a quiet-gated safety warning is invisible on every ordinary invocation — that defeats the point of warning at all. Moving it to stderr means it can be unconditional without ever corrupting `--format json` (or any other machine-parseable) stdout output.

`docs/features/multimodal.md` was updated to describe this actual behavior (lines 425, 437-438). Regression coverage: "CLI: batch --format json stdout stays valid JSON even though the multimodal attach notice fires (notice goes to stderr)" (`test/continuous-test-suite-bugfixes.ts:3985`).

src/cli/factories/commandFactory.ts
murdore: Fixed — added a dedicated `logger.alwaysStderr()` method (`src/lib/utils/logger.ts:399-411`) that mirrors `logger.always()`'s semantics (bypasses log-level/debug gating, unconditional) but calls `console.error` instead of `console.log`, so it targets stderr. Exported alongside `always` at `logger.ts:517-519`.

Both warning sites that previously used `logger.always()` now use `logger.alwaysStderr()`:
- Size warnings in `validateCliInputFiles` — `src/cli/factories/commandFactory.ts:1036`.
- Batch multimodal attachment notice in `executeBatch` — `commandFactory.ts:4782-4786`.

(We deliberately did not reuse `logger.error()` for this — that would prefix messages with `[NEUROLINK:ERROR]` and pollute log-level semantics/history for what are advisory notices, not errors.)

Regression test added: "CLI: batch --format json stdout stays valid JSON even though the multimodal attach notice fires (notice goes to stderr)" (`test/continuous-test-suite-bugfixes.ts:3985`) — spawns the built CLI with `--format json`, captures stdout/stderr separately, and asserts `JSON.parse(stdout)` succeeds while the notice text is only present in stderr.

src/cli/factories/commandFactory.ts
murdore: Fixed — `csvOptions`/`videoOptions` are now only constructed when the corresponding files are actually attached, via conditional spreads:

```ts
...(batchCsvFiles?.length && {
  csvOptions: {
    maxRows: argv.csvMaxRows as number | undefined,
    formatStyle: argv.csvFormat as "raw" | "markdown" | "json" | undefined,
  },
}),
...(batchVideoFiles?.length && {
  videoOptions: {
    frames: argv.videoFrames as number | undefined,
    quality: argv.videoQuality as number | undefined,
    format: argv.videoFormat as "jpeg" | "png" | undefined,
    transcribeAudio: argv.transcribeAudio as boolean | undefined,
  },
}),
```
(`src/cli/factories/commandFactory.ts:4846-4867`)

Agreed this was harmless (empty option objects), but batch has no CSV/video "mode" to opt into unlike `generate`/`stream`, so constructing them unconditionally was pure noise on every prompt of every batch run. Covered indirectly by the existing "CLI #291: batch now recognizes --csv" test (`test/continuous-test-suite-bugfixes.ts:3938`), which now also uses a plain `JSON.parse(ok.stdout.trim())` instead of bracket-scanning (see the sibling thread on the JSON-extraction nitpick).

docs/features/multimodal.md
murdore: Fixed by changing the implementation rather than the docs, so the two stay in sync going forward.

The batch attachment notice (`src/cli/factories/commandFactory.ts:4772-4788`) is no longer gated on `!options.quiet` — it now fires unconditionally on every batch run that attaches multimodal files, and uses the new `logger.alwaysStderr()` (writes to `console.error`, see `src/lib/utils/logger.ts:399-411`) instead of `logger.always()` (stdout), so it never corrupts `--format json` output.

Docs updated at `docs/features/multimodal.md:425` to describe the actual behavior: *"a one-line notice is printed to **stderr**, unconditionally — it is not suppressed by `--quiet`, so it never corrupts `--format json` output written to stdout"*.

Design choice: rather than documenting the suppressed-by-default behavior (which would mean users silently miss a safety-relevant notice about files being attached to every prompt in the batch), we made the notice always-visible on stderr. This keeps stdout clean for `--format json` while guaranteeing the notice isn't accidentally hidden by the `quiet: true` default. Same treatment applied to the large-file size warnings in `validateCliInputFiles` (`commandFactory.ts:1031-1038`) — see the sibling thread on `docs/features/multimodal.md:440`.

Regression test added: "CLI: batch --format json stdout stays valid JSON even though the multimodal attach notice fires (notice goes to stderr)" (`test/continuous-test-suite-bugfixes.ts:3985`) — asserts `JSON.parse(stdout)` succeeds while the notice is only present in captured stderr.

docs/features/multimodal.md
murdore: Fixed by making the warning unconditional + moving it to stderr, rather than documenting the quiet-suppressed behavior.

Large-file warnings (`validateSingleFilePath`, `commandFactory.ts:967-971`, surfaced via `validateCliInputFiles`, `commandFactory.ts:1031-1038`) are pushed to a `warnings[]` array and printed with `logger.alwaysStderr()` — no `!options.quiet` gate — so they're visible on every invocation regardless of `--quiet`, and land on stderr so `--format json` stdout output stays valid JSON.

Docs updated at `docs/features/multimodal.md:438`: *"prints a non-blocking warning to **stderr**. ... this is unconditional — visible without `--debug` and regardless of `--quiet` — and never mixes into stdout, so `--format json` output stays valid JSON even when large-file warnings fire."*

This mirrors the fix applied to the batch attachment notice discussed on the sibling `docs/features/multimodal.md:431` thread — we picked "always visible on stderr" over "document that it's hidden by default" because a size/safety warning that's silently suppressed by the default `quiet: true` defeats its own purpose.

test/continuous-test-suite-bugfixes.ts
murdore: Fixed — `runCliBugfix` now checks the built CLI exists before spawning it:

```ts
const CLI_DIST_PATH = pathJoin(process.cwd(), "dist", "cli", "index.js");
...
if (!existsSync(CLI_DIST_PATH)) {
  reject(
    new Error(
      `dist/cli/index.js not found at ${CLI_DIST_PATH} — run 'pnpm run build:cli' first`,
    ),
  );
  return;
}
```
(`test/continuous-test-suite-bugfixes.ts:41`, guard at `:59-65`)

Running the suite without a prior build now fails immediately with an actionable message instead of the generic Node module-not-found error from the spawned child process.

test/continuous-test-suite-bugfixes.ts
murdore: Fixed — the "CLI #291: batch now recognizes --csv" test no longer scans combined output for the first `[`/last `]`; it now does a direct `JSON.parse(ok.stdout.trim())` (`test/continuous-test-suite-bugfixes.ts:3938-3983`).

This is now safe because the multimodal attachment notice (and the size warnings) are written exclusively to stderr via the new `logger.alwaysStderr()` (see the stdout/stderr threads on `commandFactory.ts`), so `stdout` is guaranteed to contain only the JSON payload when `--format json` is used — no more risk of a warning line landing between `[` and `]` and corrupting the bracket-scan.

`runCliBugfix` was also extended to expose `stdout`/`stderr` as separate captured streams (previously only a combined `out` string), which is what makes asserting "stdout is clean JSON, notice is only in stderr" possible. A dedicated regression test was added for this exact scenario: "CLI: batch --format json stdout stays valid JSON even though the multimodal attach notice fires (notice goes to stderr)" (`test/continuous-test-suite-bugfixes.ts:3985`).

src/cli/factories/commandFactory.ts
```diff
@@ -910,8 +911,77 @@ export class CLICommandFactory {
     );
   }
 
-  private static validateCliInputFiles(argv: Record<string, unknown>): void {
-    const fileArgs: Array<{
+  // #288: shared aggregated-error format for any invalid local file path
+  // (missing, unreadable, or a directory). Used by both the flag-based loop
```
Tara-ag: 💡 **MINOR** — `file://` URLs are classified as non-local by `isNonLocalFileReference`, so they bypass the new directory/missing-path validation entirely. A user passing `--image file:///path/to/a/directory` would therefore not hit the friendly pre-flight error and would still fail later in processing.

Consider either stripping the `file://` prefix and validating the resulting local path through `validateSingleFilePath`, or explicitly documenting that `file://` inputs are not covered by the new CLI file validation.

Tara-ag (approved):
## Review Summary

Reviewed 4 changed files. The PR addresses #288 and #291 cleanly: CLI file inputs now get upfront validation (missing/unreadable/directory paths produce a friendly aggregated error + troubleshooting block), batch now wires `--image/--csv/--pdf/--video` into every prompt, and safety notices/warnings are emitted unconditionally to stderr so `--format json` stdout stays parseable.

### New issues raised: 1
- **MINOR** (`src/cli/factories/commandFactory.ts:915`): `file://` URLs are treated as non-local and bypass the new directory/missing-path validation, so a `file://` path pointing at a directory would still fail later in processing rather than getting the friendly pre-flight error.

No blocking criteria are met. Existing review threads from CodeRabbit/Copilot/Tara-ag were already resolved by the author and were not re-raised.

src/cli/factories/commandFactory.ts
murdore: Fixed in 98dedfb27b366c25627c0f9dcbb5f3fb386a8a7d.

`file://` URLs are no longer classified as non-local. `isNonLocalFileReference` now only matches `http://`, `https://`, and `data:` — genuine remote/inline references. A new `tryResolveFileUrl` helper decodes `file://` URLs to a filesystem path via `fileURLToPath` (returning `undefined` on a malformed URL, reported as a friendly error), and that decoded path is run through the same `validateSingleFilePath` existence/directory/size checks as plain paths.

Added 3 regression tests in test/continuous-test-suite-bugfixes.ts: `--image file://<dir>` is rejected as a directory, `--image file://<missing>` is rejected as not found, and a valid `file://<file>` passes validation while genuine `https://` URLs remain skipped entirely.

src/cli/factories/commandFactory.ts
```diff
@@ -4692,7 +4863,35 @@ export class CLICommandFactory {
 
           const runBatchGenerate = () =>
             sdk.generate({
```
Tara-ag, line 4810: 💡 **MINOR** — Empty arrays are truthy, so when only one modality is attached (e.g. only `--csv`), the `input` object still spreads the other empty arrays (`images: []`, `pdfFiles: []`, `videoFiles: []`). This is harmless if the SDK skips empty arrays, but it is inconsistent with the `csvOptions`/`videoOptions` spreads below which correctly gate on `?.length`. Consider using the same `?.length` guard here for parity and a cleaner request payload:

```ts
...(batchImages?.length && { images: batchImages }),
...(batchCsvFiles?.length && { csvFiles: batchCsvFiles }),
...(batchPdfFiles?.length && { pdfFiles: batchPdfFiles }),
...(batchVideoFiles?.length && { videoFiles: batchVideoFiles }),
```

src/cli/factories/commandFactory.ts
```diff
@@ -938,18 +1034,38 @@ export class CLICommandFactory {
           continue;
         }
```
Tara-ag: 💡 **MINOR** — `logger.alwaysStderr` already writes to stderr, so the `chalk.yellow` wrapper is fine for color. However, the warning string itself is built with `formatFileSize` from `src/cli/utils/audioFileUtils.ts`. That helper returns units like `"10 MB"` (with a space), which matches the docs, but verify it rounds consistently with the threshold literals here (e.g. `50 * 1024 * 1024` is exactly 50 MiB; `formatFileSize` uses 1024-based units, so the comparison is consistent). No change needed unless you want to share these thresholds with `SIZE_LIMITS_BYTES` instead of hard-coding 50 MB twice.

test/continuous-test-suite-bugfixes.ts
```diff
@@ -3850,6 +3907,258 @@ exit 127
...
       ),
   },
+  // ---------- #288: CLI rejects a directory as a file input (pre-flight) ----------
+  {
+    name: "CLI #288: --image pointing at a directory is rejected with troubleshooting",
+    category: "cli-file-validation",
+    fn: async () => {
+      const dir = mkdtempSync(pathJoin(tmpdir(), "nl-288-"));
+      try {
+        const res = await runCliBugfix([
```
Tara-ag: 💡 **MINOR** — The `runCliBugfix` helper rejects with a plain `Error` when `dist/cli/index.js` is missing. Since this suite is run via `tsx` and exits non-zero on unhandled rejections, this is fine. Consider whether the rejection message should also be written to `stderr` so it is visible when the suite is invoked directly, but the current behavior is acceptable.

Tara-ag (approved):
## Review Summary

Reviewed 4 changed files in PR #1191.

### Existing comments
Loaded 18 existing review comments across the PR (CodeRabbit, Copilot, Tara-ag). All raised points were already addressed by the author; I did not duplicate any of them.

### New issues raised (3 MINOR)
- `src/cli/factories/commandFactory.ts:4865` — batch `input` spreads empty arrays for unattached modalities; suggested gating on `?.length` for parity with the `csvOptions`/`videoOptions` spreads.
- `src/cli/factories/commandFactory.ts:1036` — noted that size-warning thresholds are hard-coded here while `SIZE_LIMITS_BYTES` exists; informational only since units are consistent.
- `test/continuous-test-suite-bugfixes.ts:3917` — noted the missing-dist rejection is a thrown `Error`; acceptable, but could surface on `stderr` when run directly.

### Blocking criteria
None of the blocking criteria are met:
- No hardcoded secrets or critical security vulnerabilities.
- No public SDK API breakage.
- No CLAUDE.md critical rule violations (the change is CLI-only, no provider registry/static-import issues, no Gemini tools/schema conflict, no `formatProviderError` changes).
- No MAJOR issues; only 3 MINOR suggestions.

### Verification
The PR includes regression tests for the directory-rejection, batch-multimodal wiring, stdout/stderr separation, TOCTOU stat failures, batch positional validation, and `file://` URL handling paths. CI checks (`pnpm run check`, `pnpm run lint`, `pnpm run build:cli`, `prettier --check`) are reported passing.

**Decision: APPROVE**

src/cli/factories/commandFactory.ts
murdore: Fixed — guarded each array spread (`images`, `csvFiles`, `pdfFiles`, `videoFiles`) with `?.length` so empty modalities aren't spread into `input`, matching the pattern already used by `csvOptions`/`videoOptions` below. See src/cli/factories/commandFactory.ts:4864-4876 at 69d38ca6649e51b25cf514552b1a494f005c1722.

src/cli/factories/commandFactory.ts
murdore: Addressed for consistency: `--csv` now references `SIZE_LIMITS_BYTES.CSV_MAX` instead of a hard-coded `50 * 1024 * 1024` literal (same 50 MB value, now sourced from the shared constant). `--pdf` intentionally stays a literal 50 MB — it is a soft pre-flight warning set at half of `SIZE_LIMITS_BYTES.PDF_MAX` (100 MB), not the hard cap itself, so tying it to `PDF_MAX` would silently double the warn threshold. Added a comment explaining that. See src/cli/factories/commandFactory.ts:1013-1018 at 69d38ca6649e51b25cf514552b1a494f005c1722.

test/continuous-test-suite-bugfixes.ts
murdore: Fixed — the missing-dist rejection message is now also written to `process.stderr` before rejecting, so it's visible when the suite is invoked directly (not only via the unhandled-rejection exit path). See test/continuous-test-suite-bugfixes.ts:59-67 at 69d38ca6649e51b25cf514552b1a494f005c1722.

src/cli/factories/commandFactory.ts
```diff
@@ -4643,6 +4775,47 @@ export class CLICommandFactory {
...
       const sessionId = globalSession.getCurrentSessionId();
 
+      // #291: batch parsed --image/--csv/--pdf/--video but never attached them.
+      // Resolve them once (they apply to every prompt) and fail fast on bad
+      // paths, mirroring generate/stream. NOTE: --file is the batch <file>
+      // positional (the prompts list, already validated above), not the
+      // --file auto-detect flag — excluded here so a validation failure
+      // never misattributes the positional to a --file flag that doesn't
+      // exist in batch.
+      CLICommandFactory.validateCliInputFiles(
```
Tara-ag: 💡 **MINOR** — Batch multimodal flag validation runs *after* the prompts file has already been read and parsed (lines 4701–4715). The inline comment says this "mirror[s] generate/stream", but `generate`/`stream` validate all inputs before any processing. For a large prompts file paired with an invalid `--image` path, the user pays the read/parse cost before the flag error surfaces.

**Suggested fix:** move `validateCliInputFiles` and the four `processCli*` calls above `fs.readFileSync(resolvedPromptsPath)` so invalid multimodal inputs fail truly pre-flight.

Tara-ag (approved):
## Review Summary

**Files reviewed:** 4 changed files
**New issues raised:** 1 (MINOR)
**Blocking issues:** 0

### Findings

- **MINOR** (`src/cli/factories/commandFactory.ts:4785`): Batch multimodal flag validation runs *after* the prompts file has already been read and parsed. For a large prompts file paired with an invalid `--image`/`--csv`/`--pdf`/`--video` path, the user pays the read/parse cost before the flag error surfaces. Consider moving `validateCliInputFiles` and the four `processCli*` calls above `fs.readFileSync(resolvedPromptsPath)` for truly pre-flight validation that mirrors `generate`/`stream`.

### Assessment

The PR cleanly addresses the stated bug fixes (#288, #291):
- Directories, missing paths, and unreadable files are rejected up front with aggregated errors + troubleshooting hints.
- `file://` URLs are decoded and validated as local paths; genuine remote references still skip filesystem checks.
- Batch now wires `--image`/`--csv`/`--pdf`/`--video` into every prompt and emits an unconditional stderr notice so `--format json` stdout stays parseable.
- A new `logger.alwaysStderr()` helper keeps safety-relevant output off stdout.
- Regression tests cover directory rejection, batch flag wiring, stdout/stderr separation, TOCTOU stat failures, batch positional misattribution, and `file://` URL handling.

Existing review threads (from CodeRabbit, Copilot, Tara-ag) were checked; the issues they raised appear resolved in the current revision. No hardcoded secrets, no breaking SDK API changes, and no violations of non-negotiable CLAUDE.md rules were found.

**Decision:** Approve. The one new comment is a minor UX ordering suggestion, not a blocker.

src/cli/factories/commandFactory.ts
murdore: Fixed in 9b4ce5e38eae71859a9aa604c9bf34978d319818: `validateCliInputFiles` and the four `processCli*` calls now run immediately after the `if (!argv.file)` guard, before `resolveFilePaths([argv.file])`/`fs.readFileSync(resolvedPromptsPath)` touch the prompts file at all — matching generate/stream's validate-before-process ordering. Added a deterministic regression test (empty prompts file + invalid --image path) that asserts the `--image path not found` error surfaces and `No prompts found` does NOT, proving the new ordering: with the old post-read ordering this exact input would have thrown "No prompts found in file" first and never reached the --image validation.

Tara-ag (approved):
## Review Summary

Reviewed the 4 changed files in PR #1191 against the project standards (CLAUDE.md / CONTRIBUTING.md) and the existing review threads.

### Existing comments
All 22 prior review threads are resolved. The author addressed every raised point:
- stdout/stderr separation for safety notices and warnings (`logger.alwaysStderr`)
- `fs.statSync` error handling with friendly aggregated errors
- Batch `<file>` positional misattribution fix
- `dist/cli/index.js` existence guard in tests
- `file://` URL local-path validation
- Conditional spreads for empty multimodal arrays/options
- Pre-flight validation ordering in `batch`

### New findings
No new blocking, major, minor, or suggestion-level issues identified in this pass.

### Checks performed
- Security: no hardcoded secrets, no injection/eval/SSRF vectors, `fileURLToPath` used safely
- Architecture: changes are CLI-only, no SDK public API breakage, no provider-registry or `formatProviderError` changes
- Correctness: validation ordering, TOCTOU handling, aggregated errors, stderr routing, and batch multimodal wiring all look correct
- Tests: comprehensive regression coverage added for #288, #291, and the review-round fixes

**Decision:** Approve.

src/cli/factories/commandFactory.ts
```diff
@@ -900,18 +902,103 @@ export class CLICommandFactory {
...
+    // processing (readFileSync on a dir throws a raw EISDIR).
+    if (stat.isDirectory()) {
+      errors.push(`${option} path is a directory, not a file: ${rawPath}`);
+      return;
+    }
+
+    if (sizeThresholdBytes && stat.size > sizeThresholdBytes) {
+      warnings.push(
+        `${option} ${rawPath} is ${formatFileSize(stat.size)} ` +
+          `(larger than ${formatFileSize(sizeThresholdBytes)}) — processing may be slow or hit provider limits.`,
```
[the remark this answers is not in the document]
murdore: Fixed in commit 7b14c150 (HEAD).

Added `redactUrlForError()` in `src/lib/utils/logSanitize.ts` — it reuses the existing `redactUrlCredentials()` (strips embedded `user:pass@`) and additionally truncates at the first `?`/`#`, so a `file://` value can never leak a query string, fragment, or userinfo credentials once it reaches a user-facing message. Plain filesystem paths (no `://`) pass through untouched.

Applied at every site where `rawPath`/`resolvedPath` get interpolated into an error or warning in `validateSingleFilePath` (not-found, unreadable, is-a-directory, size-warning) and in the invalid-`file://`-URL branch of `validateCliInputFiles` — not just the two cited lines.

Regression tests added to `test/continuous-test-suite-bugfixes.ts`:
- `file://…?token=SECRET123#frag` pointing at a missing path → error contains "path not found" but neither the token nor the fragment nor any `?`/`#`.
- `file://user:hunter2@host/nope.png?token=SECRET#frag` (malformed authority, hits the "not a valid file:// URL" branch) → error redacts both the credentials and the query/fragment.

Verified manually against the rebuilt CLI as well, e.g.:
`--image "file:///tmp/nope.png?token=SECRET123#frag"` → `--image path not found: file:///tmp/nope.png (resolved to /tmp/nope.png)`, no trace of the token or fragment.

src/cli/factories/commandFactory.ts
```diff
@@ -4583,11 +4701,52 @@ export class CLICommandFactory {
...
+      // so a large prompts file paired with an invalid --image path paid
+      // the read/parse cost before the flag error ever surfaced. NOTE:
+      // --file is the batch <file> positional (the prompts list, validated
+      // separately below), not the --file auto-detect flag — excluded here
+      // so a validation failure never misattributes the positional to a
+      // --file flag that doesn't exist in batch.
+      CLICommandFactory.validateCliInputFiles(
+        argv as unknown as Record<string, unknown>,
+        { excludeOptions: ["--file"] },
+      );
```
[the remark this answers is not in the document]
murdore: Fixed in commit 7b14c150 (HEAD) — and it needed one more step than just dropping `commonOptions.file` from `.options()`.

Root cause: yargs implicitly treats a command's positional key (e.g. `<file>` in `"batch <file>"`) as a recognized flag name too, independent of whether it's separately registered via `.options()`. So even after excluding `commonOptions.file` from `buildOptions()`, `--file value` still passed `.strictOptions()` silently (yargs resolved it against the positional and the value was dropped, not flagged as "Unknown argument"). Verified this directly with isolated yargs repros before and after the fix.

Actual fix: renamed the batch command's positional key from `file` to `promptsFile` throughout — the command string (`"batch <promptsFile>"`), the `.positional()` call, the `BatchCommandArgs` type field (`src/lib/types/cli.ts`), and all `argv.file` → `argv.promptsFile` references in `executeBatch`. This doesn't change the CLI's public invocation syntax (still `neurolink batch prompts.txt`) but does make `--file` genuinely rejected. `buildOptions()` also gained an `excludeOptions` parameter so `commonOptions.file` is omitted from batch's registered options entirely (belt-and-suspenders, though the positional rename is what actually closes the gap).

Verified against the rebuilt CLI:
- `neurolink batch prompts.txt --file other.txt --dry-run` → exit 1, `Error: Unknown argument: file`.
- `neurolink batch prompts.txt --dry-run` (no `--file`) → exit 0, normal batch output — the `<promptsFile>` positional and other options (e.g. `--csv`) are unaffected.

Regression tests added to `test/continuous-test-suite-bugfixes.ts` (both spawn the built CLI):
- `batch prompts.txt --file other.txt --dry-run` → non-zero exit + `/unknown argument.*file/i` in output.
- `batch prompts.txt --dry-run` → exit 0, no "unknown argument"/"No prompts found", proving the positional still works.

Tara-ag (approved):
## Review Summary

Reviewed PR #1191: **fix(cli): validate file inputs + wire multimodal flags into batch (#288, #291)**

### Files reviewed
- `docs/features/multimodal.md`
- `src/cli/factories/commandFactory.ts`
- `src/lib/types/cli.ts`
- `src/lib/utils/logSanitize.ts`
- `src/lib/utils/logger.ts`

### Existing comments
All 24 existing review threads are resolved (including the round-5 `file://` redaction and `--file` positional collision fixes). No duplicates raised.

### New issues found
**None.** The changes correctly address the previously raised concerns:

- `validateSingleFilePath` wraps `fs.statSync` in try/catch and feeds friendly aggregated errors.
- Size warnings and the batch multimodal notice are unconditional and routed to `stderr` via the new `logger.alwaysStderr()`, preserving `--format json` stdout.
- `file://` URLs are decoded and validated locally; remote/http/data URLs are still skipped.
- `file://` query strings, fragments, and embedded credentials are redacted before appearing in user-facing messages.
- Batch renames its positional to `promptsFile` and omits `commonOptions.file`, so `--file` is rejected as an unknown argument.
- Batch validates multimodal flags before reading/parsing the prompts file.
- `csvOptions`/`videoOptions` and the file arrays are only spread when the corresponding files are attached.
- Regression tests cover directory rejection, missing paths, TOCTOU stat failures, stdout/stderr separation, `file://` validation/redaction, and the `--file` rejection.

### CLAUDE.md critical rules
No violations of the non-negotiable architectural rules enforced outside CI.

### Decision
**Approve.** The PR is ready to merge.

src/cli/utils/inputValidation.ts
```diff
@@ -29,25 +33,157 @@ export const CLI_SOFT_LIMITS_MB = {
...
+  return new Error(
+    "❌ One or more input files are invalid:\n" +
+      `${errors.join("\n")}\n` +
+      "💡 Troubleshooting:\n" +
+      "  • Check the path is correct and relative to your current directory, or pass an absolute path / URL.\n" +
+      `  • Check your working directory (${process.cwd()}).\n` +
+      "  • For remote resources, prefix the value with http:// or https://.\n" +
+      "  • If you pointed at a folder, point at the file inside it instead.",
+  );
+}
```
Tara-ag, line 80: 💡 **MINOR** — The troubleshooting block now echoes `process.cwd()` verbatim, which leaks the host's current working directory into user-facing errors. In CI/automation contexts this can expose deployment paths or internal directory layouts.

**Suggested fix:** either omit the absolute cwd from the message or redact it to a basename when the path is outside the project root:

```ts
`  • Check your working directory (${path.basename(process.cwd())}).\n` +
```

or keep the hint generic:

```ts
"  • Check your working directory.\n" +
```

src/lib/types/cli.ts
```diff
@@ -161,8 +161,13 @@ export type StreamCommandArgs = BaseCommandArgs &
...
   CliClassifierRouterFlags & {
-    /** Input file path */
-    file?: string;
+    // #1191 round-5: named `promptsFile`, not `file` — yargs implicitly
+    // treats a positional's key as a recognized flag name too, so keeping
+    // this "file" would let `--file` silently pass strictOptions() despite
+    // not being a registered batch option, colliding with the (unrelated)
+    // common `--file` auto-detect flag that batch intentionally omits.
+    /** Prompts-list file path (the `<promptsFile>` positional) */
+    promptsFile?: string;
```
Tara-ag, line 170: ⚠️ **MAJOR** — Renaming `BatchCommandArgs.file` to `BatchCommandArgs.promptsFile` is a breaking change to a publicly exported SDK type. `BatchCommandArgs` is re-exported from `src/lib/types/index.ts`, so existing callers referencing `.file` will fail to compile.

CLAUDE.md Critical Rule 5: *public SDK API stays backward compatible — no breaking changes to existing callers without a documented migration*.

The positional rename in yargs is the correct fix for the `--file` collision, but the type change needs an explicit migration note for SDK consumers (e.g. in the PR description / changelog / MIGRATION.md).

**Suggested fix:** add a documented migration path such as:

> Migration: if you import `BatchCommandArgs` and access the batch prompts-list path, replace `.file` with `.promptsFile`.

If feasible, retaining the old field as a deprecated alias is not practical here because yargs no longer populates `argv.file` for the `batch` command.

src/cli/utils/inputValidation.ts
```diff
@@ -74,81 +210,43 @@ export function validateCliInputFiles(argv: Record<string, unknown>): void {
     { option: "--file", value: argv.file as string | string[] | undefined },
   ];
 
-  const missingPaths: string[] = [];
+  const errors: string[] = [];
 
   for (const { option, value, warnAtMB } of fileArgs) {
     if (!value) {
       continue;
     }
```
Tara-ag, line 218: ⚠️ **MAJOR** — PDF soft-warning threshold contradicts the documented behavior.

`CLI_SOFT_LIMITS_MB.PDF_MAX_MB` is defined as `100`, and `--pdf` uses it directly as `warnAtMB`. The PR description and `docs/features/multimodal.md` both state the warning fires for **CSV/PDF > 50 MB**. With the current code, a 75 MB PDF receives no warning.

This appears to be a regression from the consolidation into `inputValidation.ts` — earlier review rounds explicitly kept `--pdf` at a 50 MB literal (half the processor hard cap).

**Suggested fix:** define a separate soft threshold (e.g. `PDF_SOFT_MAX_MB: 50`) or use `CLI_SOFT_LIMITS_MB.CSV_MAX_MB` for `--pdf` if the intent is a uniform 50 MB CSV/PDF warning, and update the constant name so it doesn't read as the hard cap:

```ts
export const CLI_SOFT_LIMITS_MB = {
  IMAGE_MAX_MB: 10,
  CSV_MAX_MB: 50,
  PDF_SOFT_MAX_MB: 50, // aligns with docs/PR description
  VIDEO_MAX_MB: 500,
} as const;
```

Then wire `--pdf` to `CLI_SOFT_LIMITS_MB.PDF_SOFT_MAX_MB`.

src/cli/factories/commandFactory.ts
```diff
@@ -4616,15 +4649,41 @@ export class CLICommandFactory {
...
+      // #291/round-4 #1191: validate multimodal flags and resolve their
+      // paths *before* the prompts file is read/parsed below, mirroring
+      // generate/stream (which validate all inputs before any processing).
+      // Previously this ran after `fs.readFileSync(resolvedPromptsPath)`,
+      // so a large prompts file paired with an invalid --image path paid
+      // the read/parse cost before the flag error ever surfaced. `--file`
+      // (the common auto-detect flag) isn't registered on batch at all —
+      // see createBatchCommand — so `argv.file` can never be set here; no
+      // exclusion needed to protect the (differently-named) positional.
+      validateCliInputFiles(argv as unknown as Record<string, unknown>);
```
Tara-ag, line 4665: 💡 **MINOR** — `executeBatch` now supports CSV inputs but does not call `validateCsvMaxRows(argv)`, unlike `executeGenerate` and `executeStream`. An invalid `--csv-max-rows` value will therefore not fail fast in batch mode and will instead be passed through to `csvOptions.maxRows`.

**Suggested fix:** add the same pre-flight validation immediately after `validateCliInputFiles`:

```ts
validateCliInputFiles(argv as unknown as Record<string, unknown>);
validateCsvMaxRows(argv);
```

Tara-ag (changes requested):
## Review summary

Reviewed 6 changed files. The PR consolidates CLI file validation and wires multimodal flags into `batch`. Most previously-raised issues are addressed, but I found **4 new issues** (2 MAJOR, 2 MINOR) that need resolution before merge.

### New issues

| Severity | File | Line | Topic |
| ⚠️ MAJOR | `src/cli/utils/inputValidation.ts` | ~218 | PDF soft-warning threshold is 100 MB, but the PR description and docs state **CSV/PDF > 50 MB**. |
| ⚠️ MAJOR | `src/lib/types/cli.ts` | ~170 | `BatchCommandArgs.file` renamed to `promptsFile` — breaking change to a publicly exported SDK type with no documented migration path. |
| 💡 MINOR | `src/cli/utils/inputValidation.ts` | ~80 | Troubleshooting block leaks `process.cwd()` verbatim into user-facing errors. |
| 💡 MINOR | `src/cli/factories/commandFactory.ts` | ~4665 | `executeBatch` does not call `validateCsvMaxRows(argv)`, unlike `executeGenerate`/`executeStream`. |

### Blocking rationale

- **Public SDK API breakage**: `BatchCommandArgs` is re-exported from `src/lib/types/index.ts`. Renaming `file` → `promptsFile` breaks existing callers. Per CLAUDE.md Critical Rule 5 and the project blocking criteria, this requires a documented migration path (or a backward-compatible alias) before merge.
- **PDF threshold regression**: The documented/PR-described warning fires at **50 MB** for PDF, but the code uses `CLI_SOFT_LIMITS_MB.PDF_MAX_MB` (100 MB). A 75 MB PDF receives no warning, contradicting the user-facing docs.

### Next steps

1. Fix the PDF soft-warning threshold to 50 MB (define `PDF_SOFT_MAX_MB: 50` or reuse `CSV_MAX_MB`) and keep `PDF_MAX_MB` as the processor hard-cap if needed.
2. Add an explicit migration note for `BatchCommandArgs.file` → `promptsFile` in the PR description / changelog / MIGRATION.md, or restore a deprecated `file` alias on the type if feasible.
3. (Optional but recommended) Add `validateCsvMaxRows(argv)` to `executeBatch` and redact `process.cwd()` to its basename in the troubleshooting block.

Once the two MAJOR items are resolved, this should be good to approve.

PR #1192, opened 2026-07-17
fix(csv): RFC-4180 quote-aware columns + delimiter (TSV) detection (#359, #361)

### LINKED ISSUE #359: CSV-001: Naive Comma Splitting Corrupts Quoted Fields
## Summary
CSV files with quoted fields containing delimiters are incorrectly parsed because the code uses naive `String.prototype.split(",")` which doesn't respect RFC 4180 quoting rules, causing data corruption.

## Root Cause
The code at lines 94 and 126 uses `split(",")` to count columns. This naive approach treats ALL commas as delimiters, even those inside quoted fields like `"Smith, John"`. RFC 4180 specifies that commas within double-quoted fields should be treated as literal characters, not delimiters.

## Fix
Create `src/lib/utils/csvUtils.ts` with RFC 4180 compliant `countCSVColumns()` function that tracks quote state and only counts delimiters outside quotes. Replace naive split at csvProcessor.ts:126 and fileDetector.ts:565.

## Files Affected
`src/lib/utils/csvProcessor.ts`, `src/lib/utils/fileDetector.ts`, `src/lib/utils/csvUtils.ts` (new)

## Acceptance Criteria
- [ ] Unit tests added

**Effort:** 2h

### LINKED ISSUE #361: CSV-002: TSV Files Not Supported
## Summary
Tab-separated values (TSV) files are not supported because csv-parser is called without delimiter configuration, causing TSV files to be parsed as single-column CSVs.

## Root Cause
csv-parser is instantiated without configuration: `csvParser()`. By default, it only recognizes commas as delimiters. When processing TSV files, tabs are treated as regular characters within a single column.

## Fix
Implement `detectDelimiter()` function that tests `,`, `\t`, `;`, `|` and selects the most consistent one. Pass detected delimiter to csv-parser: `csvParser({ separator: delimiter })`.

## Files Affected
`src/lib/utils/csvProcessor.ts`, `src/lib/utils/csvUtils.ts`

## Acceptance Criteria
- [ ] Unit tests added

**Effort:** 3h
**Depends on:** CSV-001 (uses countCSVColumns() utility)

murdore:
## What & why

Consolidated CSV-subsystem critical fixes around a new shared module.

### #359 (CSV-001) — naive comma splitting corrupts quoted fields
The raw-format branch of `CSVProcessor.process()` counted columns / split the header with a bare `.split(",")`, so a quoted field containing a comma (`"Smith, John"`) **inflated `columnCount`** (4 instead of 3) and corrupted header detection.

### #361 (CSV-002) — TSV / DSV files not supported
Both `csv-parser` call sites (`parseCSVFile`, `parseCSVString`) used the hard-coded **comma** default, so **tab / semicolon / pipe** files collapsed into one ragged column, and `metadata.detectedDelimiter` always lied `","`.

## Fix

New **`src/lib/utils/csvUtils.ts`** — the single shared implementation both issues asked for: `countCSVColumns`, `splitCSVFields` (RFC-4180 quote-aware), and `detectDelimiter` (comma/tab/semicolon/pipe, with an extension hint; comma wins ties, so plain CSVs are untouched).

`process()` now detects the delimiter **once** and threads it through the raw branch, both parse paths (`csvParser({ separator })`), and the reported `metadata.detectedDelimiter`.

### Enhancement gap → closed
The RAG CSV loader (`loaders.ts`) already accepted `.tsv` but hard-defaulted `delimiter=","`. It now detects the delimiter when one isn't explicitly passed — so a `.tsv` fed through `rag: { files }` isn't mis-parsed either.

## Testing / proof

New bugfixes-suite tests (deterministic, no API keys):
- Quoted-comma header → `columnCount === 3` (direct **and** via `FileDetector.detectAndProcess` — the real `generate({ files })` path).
- Tab and semicolon delimiters detected with correct `columnCount`; **plain comma unchanged** (no-regression guard).
- A real `.tsv` file through `detectAndProcess` → `type: csv`, 3 columns, `detectedDelimiter === "\t"`.

`pnpm run check` ✅ (0 errors) · `pnpm run lint` ✅ (0 errors) · `pnpm run build` ✅ · `prettier --check` ✅.

## Note
`fileDetector.ts`'s `ContentHeuristicStrategy` still has its own quote-aware counter; folding it into `csvUtils` is a low-risk follow-up (its detection heuristic is unaffected here).

Fixes #359. Fixes #361.

## Summary by CodeRabbit

* **New Features**
  * Automatic delimiter detection for comma, tab/TSV, semicolon, and pipe (comma fallback on ambiguity), including Excel-style `sep=<char>` precedence.
  * Quote-aware RFC-4180 CSV parsing and exposure of the resolved delimiter in processing metadata.

* **Bug Fixes**
  * Prevented `sep=` / metadata lines from being treated as headers; improved UTF-8 BOM and metadata stripping.
  * Normalized embedded newlines in table headers/cells while keeping multiline quoted fields correctly handled.

* **Tests**
  * Expanded end-to-end coverage for delimiter detection, `sep=` precedence, streamed parsing behavior, BOM stripping, and multiline quoted fields.

### DIFF
--- a/src/lib/rag/document/loaders.ts
+++ b/src/lib/rag/document/loaders.ts
@@ -28,6 +28,13 @@ import { existsSync } from "fs";
 import { readFile } from "fs/promises";
 import { basename, extname } from "path";
 import { logger } from "../../utils/logger.js";
+import {
+  CSVProcessor,
+  stripBom,
+  stripMetadataLine,
+  splitCsvLines,
+} from "../../utils/csvProcessor.js";
+import { splitCSVFields } from "../../utils/csvUtils.js";
 import type {
   DocumentType,
   LoaderOptions,
@@ -143,19 +150,43 @@ export class CSVLoader extends TextLoader {
   async load(source: string, options?: CSVLoaderOptions): Promise<MDocument> {
     const content = await this.loadContent(source, options?.encoding);
     const {
-      delimiter = ",",
+      delimiter,
       hasHeader = true,
       columns,
       outputFormat = "text",
     } = options || {};
 
-    const lines = content.split("\n").filter((line) => line.trim());
+    // #361: an explicit delimiter always wins; otherwise detect it from the
+    // content + extension so a .tsv fed through RAG isn't collapsed into one
+    // comma-column (CSVLoader.canHandle already accepts .tsv).
+    const effectiveDelimiter =
+      delimiter ?? CSVProcessor.detectDelimiter(content, extname(source));
+
+    // Route header/row parsing through the same shared, quote-aware path
+    // detectDelimiter() above already uses internally: a quote-aware logical
+    // row split (splitCsvLines — a newline inside a quoted field stays part
+    // of that field instead of breaking the row) with a leading Excel `sep=`
+    // preamble line stripped (stripMetadataLine), so it can't be parsed as
+    // the header. Field-level splitting also goes through the shared
+    // RFC-4180-aware splitCSVFields() instead of the loader's own naive
+    // backslash-escape parser, which didn't handle doubled-quote (`""`)
+    // escaping or quoted delimiters in the headerless-columns branch.
+    const { dataLines: strippedLines } = stripMetadataLine(
+      splitCsvLines(stripBom(content)),
+    );
+    const lines = strippedLines.filter((line) => line.trim());
+
     const headers = hasHeader
-      ? this.parseCSVLine(lines[0], delimiter)
-      : columns || lines[0]?.split(delimiter).map((_, i) => `col${i + 1}`);
+      ? splitCSVFields(lines[0] ?? "", effectiveDelimiter).map((h) => h.trim())
+      : columns ||
+        splitCSVFields(lines[0] ?? "", effectiveDelimiter).map(
+          (_, i) => `col${i + 1}`,
+        );
 
     const dataLines = hasHeader ? lines.slice(1) : lines;
-    const rows = dataLines.map((line) => this.parseCSVLine(line, delimiter));
+    const rows = dataLines.map((line) =>
+      splitCSVFields(line, effectiveDelimiter).map((field) => field.trim()),
+    );
 
     let formattedContent: string;
 
@@ -192,38 +223,24 @@ export class CSVLoader extends TextLoader {
     return ext === ".csv" || ext === ".tsv";
   }
 
-  private parseCSVLine(line: string, delimiter: string): string[] {
-    const result: string[] = [];
-    let current = "";
-    let inQuotes = false;
-
-    for (let i = 0; i < line.length; i++) {
-      const char = line[i];
-
-      if (char === '"' && (i === 0 || line[i - 1] !== "\\")) {
-        inQuotes = !inQuotes;
-      } else if (char === delimiter && !inQuotes) {
-        result.push(current.trim());
-        current = "";
-      } else {
-        current += char;
-      }
-    }
-
-    result.push(current.trim());
-    return result;
-  }
-
   private toMarkdownTable(headers: string[], rows: string[][]): string {
-    const headerRow = `| ${headers.join(" | ")} |`;
-    const separator = `| ${headers.map(() => "---").join(" | ")} |`;
-    const dataRows = rows.map((row) => `| ${row.join(" | ")} |`);
+    const safeHeaders = headers.map((h) => this.collapseEmbeddedNewlines(h));
+    const headerRow = `| ${safeHeaders.join(" | ")} |`;
+    const separator = `| ${safeHeaders.map(() => "---").join(" | ")} |`;
+    const dataRows = rows.map(
+      (row) =>
+        `| ${row.map((cell) => this.collapseEmbeddedNewlines(cell)).join(" | ")} |`,
+    );
     return [headerRow, separator, ...dataRows].join("\n");
   }
 
   private toTextTable(headers: string[], rows: string[][]): string {
-    const allRows = [headers, ...rows];
-    const colWidths = headers.map((_, i) =>
+    const safeHeaders = headers.map((h) => this.collapseEmbeddedNewlines(h));
+    const safeRows = rows.map((row) =>
+      row.map((cell) => this.collapseEmbeddedNewlines(cell)),
+    );
+    const allRows = [safeHeaders, ...safeRows];
+    const colWidths = safeHeaders.map((_, i) =>
       Math.max(...allRows.map((row) => (row[i] || "").length)),
     );
 
@@ -231,10 +248,25 @@ export class CSVLoader extends TextLoader {
       row.map((cell, i) => (cell || "").padEnd(colWidths[i])).join(" | ");
 
     return [
-      formatRow(headers),
+      formatRow(safeHeaders),
       colWidths.map((w) => "-".repeat(w)).join("-+-"),
-      ...rows.map(formatRow),
+      ...safeRows.map(formatRow),
     ].join("\n");
+  }
+
+  /**
+   * Collapse embedded newlines to a single space so one logical CSV row
+   * renders as exactly one line in the text/markdown table output.
+   *
+   * The shared quote-aware parser (`splitCsvLines` + `splitCSVFields`)
+   * preserves newlines embedded inside a quoted field verbatim — correct
+   * for the `json` output format (where `JSON.stringify` escapes them), but
+   * left as-is here it would split a single record across multiple
+   * physical table rows. Mirrors the same normalization `CSVProcessor`
+   * already applies in its own markdown formatter.
+   */
+  private collapseEmbeddedNewlines(cell: string): string {
+    return cell.replace(/\r\n|\n|\r/g, " ");
   }
 }
 
--- a/src/lib/utils/csvProcessor.ts
+++ b/src/lib/utils/csvProcessor.ts
@@ -6,7 +6,14 @@
 
 import csvParser from "csv-parser";
 import { Readable } from "stream";
+import { extname } from "path";
 import { logger } from "./logger.js";
+import {
+  countCSVColumns,
+  splitCSVFields,
+  detectDelimiter as detectDelimiterUtil,
+  CANDIDATE_DELIMITERS,
+} from "./csvUtils.js";
 import type {
   FileProcessingResult,
   CSVProcessorOptions,
@@ -569,8 +576,12 @@ function detectHasHeaders(
  * - Lines with significantly different delimiter count than line 2
  * - Lines that don't match CSV structure of subsequent lines
  */
-/** Strip a leading UTF-8 BOM (U+FEFF) if present. */
-function stripBom(text: string): string {
+/**
+ * Strip a leading UTF-8 BOM (U+FEFF) if present. Exported so every entry
+ * point that sniffs raw text (including the RAG CSVLoader) normalizes BOMs
+ * the same way before metadata/delimiter detection.
+ */
+export function stripBom(text: string): string {
   return text.charCodeAt(0) === 0xfeff ? text.slice(1) : text;
 }
 
@@ -624,6 +635,37 @@ function isMetadataLine(lines: string[]): boolean {
 }
 
 /**
+ * Strip a leading metadata line (e.g. Excel's `sep=;` preamble) from
+ * already-split lines, and read the delimiter it explicitly declares.
+ * Excel semantics: an explicit `sep=<char>` always wins over frequency-based
+ * detection; a metadata line without a recognized delimiter character is
+ * simply dropped from the sample without forcing a delimiter. This is the
+ * single place process(), detectDelimiter(), parseCSVFile(), and
+ * parseCSVString() all go through, so none of them detect a delimiter over
+ * the raw, un-stripped lines (which lets the preamble skew detection).
+ * Exported so non-CSVProcessor callers (the RAG CSVLoader) share it too.
+ */
+export function stripMetadataLine(lines: string[]): {
+  dataLines: string[];
+  hasMetadataLine: boolean;
+  explicitDelimiter?: string;
+} {
+  if (!isMetadataLine(lines)) {
+    return { dataLines: lines, hasMetadataLine: false };
+  }
+  const sepChar = lines[0]?.trim().match(/^sep=(.)/i)?.[1];
+  const explicitDelimiter =
+    sepChar && (CANDIDATE_DELIMITERS as readonly string[]).includes(sepChar)
+      ? sepChar
+      : undefined;
+  return {
+    dataLines: lines.slice(1),
+    hasMetadataLine: true,
+    explicitDelimiter,
+  };
+}
+
+/**
  * Split CSV text into logical rows for metadata detection and raw row limiting.
  *
  * Supports Unix (LF), Windows (CRLF), and classic Mac (CR) line endings, and is
@@ -632,8 +674,12 @@ function isMetadataLine(lines: string[]): boolean {
  * `""` inside quotes is the RFC-4180 escaped quote and does not toggle the quote
  * state. For unquoted input this yields exactly the same result as a plain
  * `split(/\r\n|\n|\r/)`.
+ *
+ * Exported so non-CSVProcessor callers (the RAG CSVLoader) get the same
+ * quote-aware row boundaries instead of a physical `content.split("\n")`,
+ * which would break rows on a newline embedded in a quoted field.
  */
-function splitCsvLines(csvString: string): string[] {
+export function splitCsvLines(csvString: string): string[] {
   const rows: string[] = [];
   let current = "";
   let inQuotes = false;
@@ -715,22 +761,28 @@ export class CSVProcessor {
 
     const csvString = content.toString("utf-8");
 
+    // #361: split once and strip a leading Excel `sep=` metadata line before
+    // detecting the delimiter (comma/tab/semicolon/pipe), so the preamble
+    // can't skew detection — then reuse dataLines below instead of
+    // re-splitting csvString for the raw-format row handling.
+    const lines = splitCsvLines(csvString);
+    const { dataLines, hasMetadataLine, explicitDelimiter } =
+      stripMetadataLine(lines);
+    const delimiter =
+      explicitDelimiter ??
+      detectDelimiterUtil(dataLines, extension ?? undefined);
+
     // For raw format, return CSV text with row limit (no parsing needed)
     // This preserves the CSV shape while normalizing line endings for row handling.
     if (formatStyle === "raw") {
-      const lines = splitCsvLines(csvString);
-      const hasMetadataLine = isMetadataLine(lines);
-
       if (hasMetadataLine) {
         logger.debug(
           "[CSVProcessor] Detected metadata line, skipping first line",
         );
       }
 
-      // Skip metadata line if present, then take header + maxRows data rows
-      const csvLines = hasMetadataLine
-        ? lines.slice(1) // Skip metadata line
-        : lines;
+      // dataLines already has the metadata line (if any) stripped.
+      const csvLines = dataLines;
 
       const limitedLines = csvLines.slice(0, 1 + maxRows); // header + data rows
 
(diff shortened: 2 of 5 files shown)


src/lib/utils/csvProcessor.ts
```diff
@@ -715,6 +721,13 @@ export class CSVProcessor {
 
     const csvString = content.toString("utf-8");
 
+    // #361: detect the delimiter once (comma/tab/semicolon/pipe) so TSV and
+    // other DSV files parse correctly instead of collapsing into one column.
+    const delimiter = detectDelimiterUtil(
+      splitCsvLines(csvString),
+      extension ?? undefined,
```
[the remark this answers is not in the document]
murdore: Fixed in 76fd62a1ffa43186a3c9365c0ee98408c200189d. Introduced a single shared helper, `stripMetadataLine(lines)`, right after `isMetadataLine()` in csvProcessor.ts. `process()` now calls `splitCsvLines(csvString)` exactly once, then `stripMetadataLine()` once, and reuses the resulting `dataLines` for both delimiter detection and the raw-format branch (no second split). For the non-raw path, it now calls a new private `parseCSVLines(dataLines, maxRows, delimiter)` core instead of re-invoking the public `parseCSVString(csvString, ...)`, so the metadata line is never re-scanned or re-stripped a second time. Delimiter detection also runs on the post-metadata `dataLines`, not the raw unstripped lines. Covered by new regression tests in test/continuous-test-suite-bugfixes.ts.

src/lib/utils/csvProcessor.ts
```diff
@@ -1067,6 +1094,8 @@ export class CSVProcessor {
     // Detect and skip metadata line
     const lines = splitCsvLines(normalized);
     const hasMetadataLine = isMetadataLine(lines);
+    // #361: caller-provided delimiter wins; otherwise detect from the content.
+    const effectiveDelimiter = delimiter ?? detectDelimiterUtil(lines);
     const csvData = hasMetadataLine
       ? lines.slice(1).join("\n")
       : lines.join("\n");
```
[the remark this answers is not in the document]
murdore: Fixed in 76fd62a1ffa43186a3c9365c0ee98408c200189d. `parseCSVString()` now splits the input once via `splitCsvLines()`, strips a leading metadata line via the shared `stripMetadataLine()` helper, and self-detects the delimiter from the exact same `dataLines` used to build `csvData` for the parser (delegating the actual parse to a new private `parseCSVLines()` core shared with `process()`). An explicit `sep=<char>` line takes precedence when `<char>` is a recognized delimiter; otherwise the metadata line is simply skipped for detection. Covered by new regression tests in test/continuous-test-suite-bugfixes.ts.

src/lib/utils/csvProcessor.ts
```diff
@@ -970,6 +993,9 @@ export class CSVProcessor {
 
     const hasMetadataLine = isMetadataLine(firstLines);
     const skipLines = hasMetadataLine ? 1 : 0;
+    // #361: pick the delimiter from the sniffed lines + file extension so
+    // .tsv / semicolon / pipe files aren't collapsed into one column.
+    const fileDelimiter = detectDelimiterUtil(firstLines, extname(filePath));
```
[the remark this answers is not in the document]
murdore: Fixed in 76fd62a1ffa43186a3c9365c0ee98408c200189d. `parseCSVFile()` now runs `firstLines` through the shared `stripMetadataLine()` helper and detects `fileDelimiter` from the resulting post-metadata `dataLines`, consistent with the `skipLines` value already used during the streamed parse (so detection and the actual skip are derived from the same source). Covered by a new regression test in test/continuous-test-suite-bugfixes.ts that writes a sep=;-prefixed file and asserts parseCSVFile() honors it.

src/lib/utils/csvProcessor.ts
```diff
@@ -909,15 +923,24 @@ export class CSVProcessor {
...
+   * RFC-4180 quoting. An optional extension hint (`tsv` → tab, `psv` → pipe)
+   * breaks ambiguity. Used by callers (e.g. the RAG CSV loader) that parse
+   * DSV content themselves and need the same detection (#361).
    */
+  static detectDelimiter(csvString: string, extensionHint?: string): string {
+    return detectDelimiterUtil(
+      splitCsvLines(stripBom(csvString)),
+      extensionHint,
+    );
+  }
```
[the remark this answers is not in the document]
murdore: Fixed in 76fd62a1ffa43186a3c9365c0ee98408c200189d. `CSVProcessor.detectDelimiter()` now runs its input through the same shared `stripMetadataLine()` helper used by `process()`, `parseCSVFile()`, and `parseCSVString()` before calling `detectDelimiterUtil()`, so a leading `sep=...` or other unrecognized metadata line no longer skews scoring. An explicit `sep=;` (or other recognized delimiter char) is honored directly per Excel semantics, matching the behavior now shared consistently across all 4 call sites. Covered by new regression tests in test/continuous-test-suite-bugfixes.ts, including one that empirically differentiates pre-fix vs post-fix behavior for a non-sep metadata preamble.

src/lib/utils/csvProcessor.ts
```diff
@@ -1057,30 +1122,60 @@ export class CSVProcessor {
...
+   * Stream-parse already-split, metadata-stripped data lines into row
+   * objects with a known delimiter. Shared core for parseCSVString() and
+   * process(), both of which need this after already splitting the content
+   * and resolving the delimiter once — calling parseCSVString() directly
+   * from process() would re-split and re-run metadata/delimiter detection
+   * over the same content a second time.
+   */
+  private static parseCSVLines(
+    dataLines: string[],
+    maxRows: number,
```
Tara-ag, line 1181: `parseCSVLines` is now the shared core for both `process()` and `parseCSVString()`. Returning an error that names `parseCSVString` when the empty input came from `process()` is misleading. Use a generic message such as `"CSVProcessor: CSV data must be a non-empty string"`.

src/lib/utils/csvProcessor.ts
```diff
@@ -762,14 +805,15 @@ export class CSVProcessor {
...
         truncated: wasTruncated,
       });
 
       // Parse a sample for enhanced metadata analysis (raw format still benefits from column analysis)
       const sampleForAnalysis = await this.parseCSVString(
         limitedCSV,
         Math.min(rowCount, 500),
+        delimiter,
       );
       const { columnMetadata, dataQualityWarnings, dataQualityScore } =
```
Tara-ag, line 832: The raw-format branch already has `dataLines` (metadata-stripped and split once). Routing `limitedCSV` back through `parseCSVString()` here performs a second split/join cycle and a redundant BOM/metadata pass just to analyze the sample. Prefer the shared core to preserve the single-split invariant:

```ts
const sampleRows = Math.min(rowCount, 500);
const sampleForAnalysis = await this.parseCSVLines(
  dataLines.slice(0, 1 + sampleRows),
  sampleRows,
  delimiter,
);
```

src/lib/utils/csvProcessor.ts
```diff
@@ -968,8 +1026,14 @@ export class CSVProcessor {
 
     await fileHandle.close();
```
Tara-ag, line 1053: `parseCSVFile` reads the first two lines to detect metadata, but unlike `process()`, `parseCSVString()`, and `detectDelimiter()` it never strips a leading UTF-8 BOM. An Excel-exported file beginning with `\ufeffsep=;` will fail the `/^sep=/i` match and the BOM will be carried into the parsed header. Strip the BOM from the accumulated buffer (or from each `firstLines` entry) before `stripMetadataLine` to match the other entry points.

Tara-ag (approved):
## Review summary

Reviewed the 5 changed files in this CSV quoting/delimiter-detection PR.

**New issues raised this run:** 3
- **MAJOR:** `parseCSVFile` does not strip a leading UTF-8 BOM before metadata detection, so Excel exports beginning with `\ufeffsep=...` will miss the explicit delimiter declaration.
- **MINOR:** Raw-format `process()` routes its sample through `parseCSVString()` instead of the shared `parseCSVLines()` core, undoing the single-split optimization.
- **MINOR:** `parseCSVLines()` returns an error message that names `parseCSVString` even when the empty input came from `process()`.

**Blocking criteria:** None met. No hardcoded secrets, no critical security issues, no public SDK breaking changes, and no CLAUDE.md critical-rule violations were introduced.

**Existing comments:** The unresolved CodeRabbit findings on `src/lib/rag/document/loaders.ts` (reuse shared quote-aware parsing) and `test/continuous-test-suite-bugfixes.ts` (pipe-delimiter regression case) are valid observations; I did not duplicate them.

Approving — the quoted-field and delimiter-detection fixes are solid, and the remaining items are polish/coverage gaps.

src/lib/rag/document/loaders.ts
```diff
@@ -143,19 +144,28 @@ export class CSVLoader extends TextLoader {
...
-      : columns || lines[0]?.split(delimiter).map((_, i) => `col${i + 1}`);
+      ? this.parseCSVLine(lines[0], effectiveDelimiter)
+      : columns ||
+        lines[0]?.split(effectiveDelimiter).map((_, i) => `col${i + 1}`);
 
     const dataLines = hasHeader ? lines.slice(1) : lines;
-    const rows = dataLines.map((line) => this.parseCSVLine(line, delimiter));
+    const rows = dataLines.map((line) =>
+      this.parseCSVLine(line, effectiveDelimiter),
+    );
```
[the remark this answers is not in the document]
murdore: Fixed in 8cbc7b62dadecad34212fad79ae0e5c493f438eb. `CSVLoader.load()` now routes header/row parsing through the same shared, quote-aware path `detectDelimiter()` already used internally: a quote-aware logical row split (newly-exported `splitCsvLines()`, so a newline embedded in a quoted field no longer breaks the row) with a leading Excel `sep=` preamble stripped (newly-exported `stripMetadataLine()`), so it can never be parsed as the header. Field-level splitting now goes through the shared RFC-4180-aware `splitCSVFields()` (also fixes doubled-quote `""` escaping and quoted delimiters in the headerless-columns branch), replacing the loader's own naive backslash-escape parser. Covered by a new regression test in test/continuous-test-suite-bugfixes.ts ("CSVLoader: quoted multiline field and sep= preamble routed through shared quote-aware parsing").

test/continuous-test-suite-bugfixes.ts
```diff
@@ -3850,6 +3850,217 @@ exit 127
...
+        return (
+          det.type === "csv" &&
+          det.metadata?.columnCount === 3 &&
+          det.metadata?.detectedDelimiter === "\t"
+        );
+      } finally {
+        rmSync(dir, { recursive: true, force: true });
+      }
+    },
+  },
```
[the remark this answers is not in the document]
murdore: Fixed in 8cbc7b62dadecad34212fad79ae0e5c493f438eb. Added a pipe-delimiter case to the same test block, asserting `detectedDelimiter === "|"` and `columnCount === 3` on both the `raw` and `json` format paths, plus parsing the returned JSON and checking actual row values (`sku`/`qty`/`price`). Existing TSV, semicolon, comma, and FileDetector coverage is unchanged; the test name now also mentions pipe.

src/lib/utils/csvProcessor.ts
murdore: Fixed in 8cbc7b62dadecad34212fad79ae0e5c493f438eb. The empty-input guard inside `parseCSVLines()` now throws `"CSVProcessor: CSV data must be a non-empty string"` — a generic message that doesn't name either specific entry point, since both `process()` and `parseCSVString()` can land here. `parseCSVString()`'s own top-level guard (which fires first for a directly-empty `csvString` argument) still keeps its specific `"CSVProcessor.parseCSVString: ..."` message, which is accurate there. The existing regression test at line ~536 only asserts `/non-empty string/`, so it still passes.

src/lib/utils/csvProcessor.ts
murdore: Fixed in 8cbc7b62dadecad34212fad79ae0e5c493f438eb, using exactly the shared-core approach you suggested. The raw-format branch now computes `sampleRows = Math.min(rowCount, 500)` and calls the private `parseCSVLines(dataLines.slice(0, 1 + sampleRows), sampleRows, delimiter)` directly, instead of routing `limitedCSV` back through `parseCSVString()`. This preserves the single-split invariant (`dataLines` was already split and metadata-stripped once, earlier in `process()`) and avoids the redundant BOM/metadata pass on a string that's already known metadata-free.

src/lib/utils/csvProcessor.ts
murdore: Fixed in 8cbc7b62dadecad34212fad79ae0e5c493f438eb. `parseCSVFile()`'s metadata-sniffing loop now strips a leading UTF-8 BOM from the accumulated buffer (`stripBom(buffer)`) before the CRLF trim and `splitCsvLines()` call that feeds `stripMetadataLine()`/the `/^sep=/i` check — matching `process()`, `parseCSVString()`, and `detectDelimiter()`. Added a new regression test ("CSVProcessor.parseCSVFile: strips a leading UTF-8 BOM before the sep= metadata check") that writes a `﻿sep=;`-prefixed file and asserts `parseCSVFile()` now correctly skips the metadata line (2 data rows instead of the pre-fix 3, where the real header leaked in as data).

src/lib/utils/csvProcessor.ts
```diff
@@ -983,7 +1065,7 @@ export class CSVProcessor {
       let lineCount = 0;
 
       const source = fs.createReadStream(filePath, { encoding: "utf-8" });
-      const parser = csvParser();
+      const parser = csvParser({ separator: fileDelimiter });
 
       const abort = () => {
```
Tara-ag, line 1080: ⚠️ **MAJOR**: `parseCSVFile` detects `skipLines` but does not pass `skipLines` to `csv-parser`. The parser will treat the `sep=;` metadata line as the header row, which is exactly the pre-existing quirk the inline comment admits. Relying on `Object.values(...).length === 3` in tests masks that the keys are wrong.

`csv-parser` supports a `skipLines` option (number). Pass it here so the real header is used for object keys:

```ts
const parser = csvParser({ separator: fileDelimiter, skipLines });
```

Then update the test to assert the actual column names (`fileRows[0].id === '1'`) instead of only value counts.

src/lib/utils/csvProcessor.ts
Tara-ag, line 1: 💡 **MINOR**: The metadata-sniffing read stream in `parseCSVFile` swallows read errors silently (`lineReader.on("error", () => resolve())`). If the file is unreadable, `firstLines` stays empty, `skipLines` becomes 0, and the subsequent `createReadStream` will likely fail anyway — but the original error (permission denied, ENOENT, etc.) is lost. Consider capturing and forwarding the error, or at least logging it, so callers get actionable diagnostics.

Tara-ag (commented):
## Review summary

Reviewed all 5 changed files in this CSV quoting/delimiter fix.

**New issues raised:**
- **1 MAJOR** — `CSVProcessor.parseCSVFile` detects `skipLines` for `sep=` metadata but does not pass `skipLines` to `csv-parser`, so the metadata line is still used as the header row and parsed objects get wrong keys. The regression test only asserts value counts, which masks this. Fix: pass `skipLines` to the parser options.
- **1 MINOR** — `parseCSVFile`'s metadata-sniffing stream swallows read errors silently, losing the original diagnostic.

**Already handled:** All prior review threads (double-splitting, metadata stripping, BOM handling, CSVLoader shared path, pipe-delimiter test coverage, generic error message) were checked and are resolved/outdated; no duplicates were posted.

**Verdict:** No blocking criteria are met this run (only 1 MAJOR issue). The fix is solid overall, but please address the `parseCSVFile` `skipLines` propagation so `sep=` metadata files produce correct object keys.

src/lib/utils/csvProcessor.ts
murdore: Fixed — parseCSVFile now passes `skipLines` straight into `csvParser({ separator: fileDelimiter, skipLines })`, so csv-parser itself skips the `sep=;` metadata line before choosing headers, instead of using it as the header row. Removed the now-redundant manual `lineCount <= skipLines` skip in the 'data' handler (double-skipping would have dropped the first real row once csv-parser handles it natively). Updated the two existing parseCSVFile tests to assert actual column names/values (e.g. `fileRows[0].id === '1'`) instead of only value counts, and added a new dedicated regression test ('CSVProcessor.parseCSVFile: skipLines is passed to csv-parser so the sep= line never becomes the header') that explicitly checks the header keys are `amount,id,name` and none of them contain `sep=`. See src/lib/utils/csvProcessor.ts:1062-1094 and test/continuous-test-suite-bugfixes.ts at fdaacfaf721756f29c9ee611d8de318ac1b2d5ce.

src/lib/utils/csvProcessor.ts
murdore: Fixed — the read-error handler on the metadata-sniffing stream now logs the error via logger.warn (with { cause: error }) before resolving, instead of silently discarding it. Still resolves rather than rejects (this sniffing pass is best-effort and the main read stream a few lines below will surface a real failure properly), but the original error is no longer lost. See src/lib/utils/csvProcessor.ts:1040-1049 at fdaacfaf721756f29c9ee611d8de318ac1b2d5ce.

Tara-ag (approved):
Reviewed all 5 changed files. No new blocking issues found that are not already covered by existing review threads.

**Scope checked:**
- `src/lib/utils/csvUtils.ts` — new RFC-4180 quote-aware helpers; no secrets, no unsafe eval, no type-rule violations.
- `src/lib/utils/csvProcessor.ts` — delimiter detection is centralized, `parseCSVFile` now passes `separator`/`skipLines` to `csv-parser`, BOM/metadata handling is consistent across entry points.
- `src/lib/rag/document/loaders.ts` — CSVLoader routes through the shared quote-aware path; delimiter detection honors `.tsv`.
- `test/continuous-test-suite-bugfixes.ts` — comprehensive regression coverage for #359/#361, `sep=`, BOM, and CSVLoader multiline fields.
- `docs/features/csv-support.md` — docs updated to describe delimiter auto-detection.

**Outstanding, already-raised concern:**
The CodeRabbit comments at `src/lib/rag/document/loaders.ts:189` and `test/continuous-test-suite-bugfixes.ts:4205 about CSVLoader’s `text`/`markdown` formatters not normalizing embedded newlines from quoted multiline fields remain unresolved. This is a functional-correctness gap, but it does not meet the configured blocking criteria (single MAJOR issue, already tracked), so I’m approving. Recommend addressing in a follow-up so one logical CSV row stays one formatted table row in text/markdown output.

src/lib/rag/document/loaders.ts
```diff
@@ -143,19 +150,43 @@ export class CSVLoader extends TextLoader {
...
+      : columns ||
+        splitCSVFields(lines[0] ?? "", effectiveDelimiter).map(
+          (_, i) => `col${i + 1}`,
+        );
 
     const dataLines = hasHeader ? lines.slice(1) : lines;
-    const rows = dataLines.map((line) => this.parseCSVLine(line, delimiter));
+    const rows = dataLines.map((line) =>
+      splitCSVFields(line, effectiveDelimiter).map((field) => field.trim()),
+    );
```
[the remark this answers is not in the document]
murdore: Fixed in f68b7c889cae3cf5d60562c89e7d7c1f926a3eac. Added a private `collapseEmbeddedNewlines()` helper to `CSVLoader` (src/lib/rag/document/loaders.ts) that replaces `\r\n|\n|\r` with a single space, mirroring the normalization `CSVProcessor.toMarkdownTable()` already applies elsewhere in the same file. Applied to header and data cells in both `toTextTable()` and `toMarkdownTable()` — the `json` output path is untouched, so embedded newlines are still preserved (and correctly escaped by `JSON.stringify`) there. One logical CSV record now always renders as exactly one physical line in both text and markdown output.

test/continuous-test-suite-bugfixes.ts
```diff
@@ -3850,6 +3851,358 @@ exit 127
...
+        rows.length === 2 &&
+        rows[0].name === "Alice" &&
+        rows[0].note === "Hello\nWorld" &&
+        rows[0].age === "30" &&
+        rows[1].name === "Bob" &&
+        rows[1].note === "Simple note" &&
+        rows[1].age === "25"
+      );
+    },
+  },
```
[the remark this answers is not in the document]
murdore: Fixed in f68b7c889cae3cf5d60562c89e7d7c1f926a3eac. Added two new regression tests in `test/continuous-test-suite-bugfixes.ts` alongside the existing JSON-only multiline test: 'CSVLoader: embedded newline in a quoted field renders as one row in the default text format' and '...in the markdown table format'. Both load a CSV with an embedded newline inside a quoted field and assert the rendered output has exactly 4 lines (header + separator + 2 data rows), that the newline was collapsed to a space (`Hello World`, not `Hello\nWorld`), and that each row's fields land on the expected line.

Tara-ag (approved):
## Review Summary

Reviewed the 5 changed files in PR #1192. All previously-raised review threads are resolved, and I did not identify any new blocking issues.

### What was checked
- **Security**: No hardcoded secrets, injection vectors, unsafe eval, or PII exposure paths.
- **CLAUDE.md Critical Rules**: No violations of the non-ESLint-enforced architectural rules (dynamic imports, Gemini schema/tools mutual exclusion, CLI/SDK separation, backward compatibility, `formatProviderError` semantics, streaming tool merge).
- **Correctness**: The new `csvUtils.ts` helpers, `CSVProcessor` refactor, and `CSVLoader` integration correctly implement RFC-4180 quote-aware splitting, delimiter detection (comma/tab/semicolon/pipe), `sep=` metadata handling, and BOM stripping. The shared `parseCSVLines` core eliminates the redundant double-split noted in earlier review rounds.
- **Backward compatibility**: `parseCSVString` adds an optional `delimiter` parameter; existing two-argument callers are unaffected. `CSVLoader` defaults to auto-detection but still accepts an explicit `delimiter`.
- **Tests / docs**: New deterministic regression tests cover quoted commas, TSV/semicolon/pipe detection, `sep=` precedence, BOM handling, and CSVLoader multiline/newline formatting. Docs are updated accordingly.

### Existing comments
All 13 inline review comments from prior rounds are resolved; none were duplicated or re-raised.

### Decision
Approve — the changes are focused, well-tested, and address the stated issues without introducing new risks.

PR #1193, opened 2026-07-17
fix(tools): null-prototype prompt-listing record + forward disableTools to gate

pdogra1299:
## Description

**What does this PR do?**

Follow-up to #1174, resolving the post-merge review summary left by @Tara-ag. Each claim was verified against the merged code before acting:

1. **Fixed (real, Minor): `applyToolInfoFiltering` used a plain `{}` record.** With a plain object, `"__proto__" in record` is truthy via the prototype, so a tool named `__proto__` was silently dropped from the system-prompt listing while surviving the native gate. The record is now built on `Object.create(null)` and both membership checks use `Object.hasOwn()`, matching the hardening on every other record in the tool-resolution pipeline. Regression test added.
2. **Defense-in-depth (the `disableTools` claim itself was a false positive):** both call sites already zero the tool record via `shouldUseTools` before the gate or discovery run, and discovery cannot inject `search_tools` into an empty record (deferrable names require presence in the record) — so no behavior bug existed. However, `disableTools` is now also forwarded into the resolved policy so the gate is self-sufficient for any future call site that forgets the upstream check. Zero behavior change today; gate-level regression test added.

The third item in the summary (confirmation that hydration mutates only call-scoped records) was answered on the thread with citations: every record handed to a provider is built fresh per call (merge spread → `sortToolRecord` always constructs a new object → discovery partition builds a new hot record), and shared `Tool` instances are never mutated (the Anthropic cacheControl decoration replaces the record entry with a new object; hydration assigns record keys only).

## Related Issues

Follow-up to #1174 (post-merge review).

## Type of Change

## How Has This Been Tested?

- `pnpm run test:tool-resolution` — 29/29 (two new regression tests: `__proto__`-named tool survives the prompt-listing filter; `disableTools` yields an empty set at the gate itself).
- `tsc --strict` clean, `pnpm run lint` 0 errors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved disabled-tool handling so tools are consistently gated across all tool-resolution paths, even if upstream callers don’t pre-filter.
  * Fixed tool listing/prompt filtering to correctly retain specially named tools (e.g., `__proto__`) without prototype-chain side effects.
* **Tests**
  * Added regression tests covering `disableTools: true` enforcement and the `__proto__` edge case in system-prompt tool listing.

### DIFF
--- a/src/lib/core/baseProvider.ts
+++ b/src/lib/core/baseProvider.ts
@@ -767,9 +767,15 @@ export abstract class BaseProvider implements AIProvider {
     toolFilter?: string[];
     excludeTools?: string[];
     enabledToolNames?: string[];
+    disableTools?: boolean;
   }): ResolvedToolPolicy {
     return resolveToolPolicy({
       options: {
+        // Defense-in-depth: both current call sites already zero the tool
+        // record via shouldUseTools before the gate runs, but forwarding
+        // disableTools makes the gate self-sufficient for any future call
+        // site that forgets the upstream check.
+        disableTools: options.disableTools,
         toolFilter: options.toolFilter,
         excludeTools: options.excludeTools,
         enabledToolNames: options.enabledToolNames,
@@ -786,6 +792,7 @@ export abstract class BaseProvider implements AIProvider {
       excludeTools?: string[];
       enabledToolNames?: string[];
       toolChoice?: unknown;
+      disableTools?: boolean;
     },
   ): Record<string, Tool> {
     const policy = this.getToolPolicy(options);
--- a/src/lib/neurolink.ts
+++ b/src/lib/neurolink.ts
@@ -8158,14 +8158,22 @@ Current user's request: ${currentInput}`;
       builtinToolNames: Object.keys(directAgentTools),
     });
 
-    const record: Record<string, ToolInfo> = {};
+    // Null prototype + own-property checks: a tool named "__proto__" must
+    // become an own entry — with a plain `{}`, `"__proto__" in record` is
+    // truthy via the prototype and the tool would be silently dropped from
+    // the listing while surviving the native gate. Matches the hardening on
+    // every other record in the tool-resolution pipeline.
+    const record: Record<string, ToolInfo> = Object.create(null) as Record<
+      string,
+      ToolInfo
+    >;
     for (const tool of tools) {
-      if (!(tool.name in record)) {
+      if (!Object.hasOwn(record, tool.name)) {
         record[tool.name] = tool;
       }
     }
     const gated = applyToolGate(record, policy);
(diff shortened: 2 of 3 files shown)


Tara-ag (approved):
## Review Summary

**Files reviewed:** 3
**New issues raised:** 0 (CRITICAL: 0, MAJOR: 0, MINOR: 0, SUGGESTION: 0)
**Decision:** Approve

### Assessment

This follow-up PR cleanly addresses the two items from the #1174 post-merge review:

1. **`src/lib/neurolink.ts` — null-prototype record hardening**
   - Switching the prompt-listing record to `Object.create(null)` and using `Object.hasOwn()` for membership checks correctly prevents a tool named `__proto__` from being silently dropped via prototype inheritance. This aligns with the existing hardening in `toolGate.ts` and `BaseProvider.sortToolRecord`, and directly satisfies the bootstrapped standard that records built from attacker-controllable keys must use null-prototype objects.

2. **`src/lib/core/baseProvider.ts` — forward `disableTools` into the resolved policy**
   - Plumbing `disableTools` through `getToolPolicy` / `applyToolFiltering` into `resolveToolPolicy` is a sound defense-in-depth change. The gate now remains self-sufficient even if a future call site forgets to zero the tool record upstream. No behavior change for current call sites, which already compute `shouldUseTools` before merging tools.

3. **`test/continuous-test-suite-tool-resolution.ts` — regression coverage**
   - Both new tests are targeted and verify the exact behaviors being changed: gate-level `disableTools` enforcement and `__proto__`-named tool retention in the listing filter.

No blocking criteria are met. The changes are additive/security-hardening only, preserve backward compatibility, and do not violate any CLAUDE.md critical rules.

Tara-ag (approved):
## Review Summary

**Files reviewed:** 3  
**New issues raised:** 0  
**Decision:** Approve

### Notes

- `src/lib/neurolink.ts`: The switch to `Object.create(null)` plus `Object.hasOwn()` in `applyToolInfoFiltering` correctly hardens the prompt-listing record against prototype-pollution-style drops of `__proto__`-named tools. The membership checks now match the semantics used in the native tool gate.
- `src/lib/core/baseProvider.ts`: Forwarding `disableTools` into `resolveToolPolicy` is a clean defense-in-depth change. Both `getToolPolicy` and `applyToolFiltering` now propagate the flag, so the gate remains self-sufficient even if a future call site forgets the upstream `shouldUseTools` check. No architectural rules are violated.
- `test/continuous-test-suite-tool-resolution.ts`: Both regression tests are well-targeted — one for gate-level `disableTools` enforcement and one for the `__proto__` edge case in prompt listing.

No blocking issues. The PR is already merged; this review confirms the changes are sound.

PR #1194, opened 2026-07-17
fix(deps): lazy-init SageMaker runtime client, move SDK to optionalDependencies

murdore:
## What

Completes the thin-core dependency work started in #1187: `@aws-sdk/client-sagemaker-runtime` — the one SDK that couldn't move — now moves to `optionalDependencies`, after converting `src/lib/providers/sagemaker/client.ts` to lazy async init.

## How

- Removed the static SDK import + eager client construction; added a module-level lazy loader following the existing `loadBedrockControl()`/`loadSageMakerControl()` precedent, with a friendly `ERR_MODULE_NOT_FOUND` → "install it with: pnpm add @aws-sdk/client-sagemaker-runtime" rethrow.
- New private `getSdk()`/`getClient()` lazily import/construct and cache; the `isDisposed` check centralizes in `getClient()`. `dispose()` also clears the cached module.
- **Zero public signature changes** — the public methods were already async; `new SageMakerRuntimeClient(config)` callers (`language-model.ts:145`, `detection.ts:35`) needed no edits since the constructor no longer touches the SDK.
- Type handling follows `googleVertex.ts`'s lazy-loader precedent (inline `typeof import(...)` + type-only class import) to satisfy the repo's no-local-type-alias ESLint rule.

## Verification (all empirical)

- `tsc --noEmit` 0 errors · eslint clean · `pnpm pack` builds; grep of `dist/` confirms **no static `client-sagemaker-runtime` import anywhere in the CLI startup chain** (index → parser → commandFactory → sagemakerCommandFactory → client)
- Clean-room A (default install) and B (`--omit=optional`): root import → 721 exports; `neurolink --version` → works in both, including the direct `node dist/cli/index.js --version` check
- Clean-room B with the SDK confirmed absent: requesting the sagemaker provider yields the deferred, comprehensible call-time error (install hint), not an import-time crash
- **Install-size evidence: `du -sh node_modules` ≈ 1.2 GB (default) vs ≈ 230 MB (`--omit=optional`)** — the docs page now cites these measured numbers
- docs/about/why-the-core-stays-thin.md updated: SageMaker moves out of the "outstanding piece" callout; no remaining gap flagged

## Pre-existing bug found during verification (NOT touched here)

`convertToolsToSageMakerFormat` (`src/lib/providers/sagemaker/language-model.ts` ~409-414) crashes on `tool.function.name` when NeuroLink auto-injects default tools (reproducible without `disableTools: true`). Unrelated to this change — flagging for a separate fix.

## Summary by CodeRabbit

* **Improvements**
  * SageMaker Runtime support now loads only when an endpoint is used, reducing startup overhead.
  * SageMaker Runtime is now optional, allowing installations to omit cloud-provider SDKs when they are not needed.
  * Missing optional SDKs now produce an actionable installation error when the related feature is requested.

* **Documentation**
  * Updated guidance to explain optional provider packages, lazy loading, and the supported SageMaker Runtime integration.


Tara-ag (approved):
## Review Summary

**Files reviewed:** 4 (`src/lib/providers/sagemaker/client.ts`, `package.json`, `pnpm-lock.yaml`, `docs/about/why-the-core-stays-thin.md`)

**Decision:** Approve — no new blocking issues identified.

### Notes

- The core change correctly converts `SageMakerRuntimeClient` to lazy async initialization: the constructor no longer touches the AWS SDK, and the module/client are cached on first actual `invokeEndpoint*` call. This satisfies the stated goal of making `@aws-sdk/client-sagemaker-runtime` a genuine optional dependency.
- Type-only imports from the optional package are erased at runtime, and the dynamic loader provides a contextual install hint on `ERR_MODULE_NOT_FOUND`, matching the project's lazy-loading precedent.
- Public signatures are preserved (constructor remains sync; `invokeEndpoint`/`invokeEndpointWithStreaming` were already async), so this is not a breaking SDK change.
- Docs and `package.json` are consistent with the move to `optionalDependencies`.

### Existing comment acknowledged

There is one unresolved existing inline comment from `copilot-pull-request-reviewer` on `src/lib/providers/sagemaker/client.ts` (around the `ERR_MODULE_NOT_FOUND` guard) suggesting the missing-package check be tightened so it only triggers when the package specifier itself is `@aws-sdk/client-sagemaker-runtime`. I did not duplicate that point, but the author may want to address it before merge if they agree.

Issue #258, opened 2025-12-01
PDF-001: No Password-Protected PDF Handling


murdore:
## Summary
The PDF processor crashes with unhandled `PasswordException` when processing password-protected PDFs, providing no mechanism for password authentication or graceful error handling.

## Root Cause
The `pdfjs.getDocument()` call doesn't configure password handling. When pdfjs encounters an encrypted PDF, it throws `PasswordException` that crashes the application.

**Location:** `src/lib/utils/pdfProcessor.ts` lines 257-263

## Impact
- Application crashes on password-protected PDFs
- No way for users to provide passwords
- Poor user experience with cryptic errors
- Breaks enterprise workflows with encrypted PDFs

## Fix Instructions
1. Add `password?: string` to `PDFProcessorOptions` type
2. Pass `password: options?.password` to `pdfjs.getDocument()` config
3. Implement `loadingTask.onPassword` handler with clear error messages
4. Handle reason 1 (password required) and reason 2 (incorrect password)

## Acceptance Criteria
- [ ] Password parameter added to options
- [ ] onPassword handler implemented
- [ ] Clear error messages for both password scenarios
- [ ] Tests added for encrypted PDFs (RC4, AES-128, AES-256)
- [ ] No regressions

## Test Coverage Required
- Password-protected PDF without password throws clear error
- Correct password allows processing
- Incorrect password throws specific error
- Various encryption levels supported

murdore:
Closing — verified implemented in `release` (v10.10.6, `0b1abf1d`).

**Where it landed:** see evidence below

**Evidence:**

[...]

PR #1195, opened 2026-07-17
fix(pdf): guard page-canvas memory (#260) and decrypt password-protected PDFs (#258)

### LINKED ISSUE #260: PDF-002: Memory Exhaustion on Large Pages
## Summary
Creating canvases for extremely large PDF pages causes out-of-memory crashes. A 50,000×50,000 pixel page at scale 2.0 requires 40GB memory allocation.

## Root Cause
Lines 273-275 blindly create canvases based on PDF page dimensions without validation. Memory = width × height × 4 bytes (RGBA).

**Location:** `src/lib/utils/pdfProcessor.ts` lines 271-275

## Impact
- OOM crashes on large-format PDFs (architectural drawings, maps, posters)
- Node.js process killed by OS
- No warning before crash
- Legitimate large PDFs cannot be processed

## Fix Instructions
1. Add `maxCanvasPixels?: number` to options (default 16,777,216 = 4096×4096)
2. Calculate `totalPixels = viewport.width × viewport.height` before canvas creation
3. If `totalPixels > maxCanvasPixels`, calculate downscale factor: `Math.sqrt(maxCanvasPixels / totalPixels)`
4. Apply downscale: `effectiveScale = scale × downscaleFactor`
5. Recreate viewport with effective scale
6. Log warning with memory estimates (before/after)

## Acceptance Criteria
- [ ] maxCanvasPixels parameter added with 16M default
- [ ] Pixel validation before canvas creation
- [ ] Automatic downscaling for oversized pages
- [ ] Warning logs with memory estimates
- [ ] Tests verify memory stays under 100MB per page
- [ ] No regressions

## Test Coverage Required
- Normal page processes without scaling
- Oversized page auto-scales without crashing
- Warning logged when auto-scaling occurs
- Custom maxCanvasPixels limit respected
- Memory usage verified under limits

murdore:
## Summary

Two fixes on the **image-fallback PDF path** (providers without native PDF support render each page to an image), consolidated as the PDF subsystem PR.

### #260 — page-canvas memory guard
A very large `/MediaBox` could force an unbounded canvas allocation during page rendering. A dependency-free `/MediaBox` pre-flight now computes the largest page's pixel area and downscales the render scale to fit a cap (`DEFAULT_MAX_CANVAS_PIXELS` = 16,777,216 ≈ 4096×4096; a WARN is logged). Non-finite/non-positive `maxCanvasPixels` is rejected.

### #258 — password-protected PDFs
An encrypted PDF failed with a raw `"No password given"` and no way to supply one. Now:
- `pdfOptions.password` (SDK) / `--pdf-password` (CLI) is forwarded to the decoder.
- pdfjs `PasswordException` maps to actionable typed errors `PDF_PASSWORD_REQUIRED` / `PDF_INCORRECT_PASSWORD`.

### Root cause of the threading gap
`buildGenerateTextOptions` reconstructs `TextGenerationOptions` field-by-field and had **silently dropped** `csvOptions`/`pdfOptions` — so the SDK `pdfOptions` and the CLI flag never reached the converter. Forwarded there and at every provider `csvOptions` pick-site. Both `password` and `maxCanvasPixels` are exposed at the `pdfOptions` surface and threaded through `buildMultimodalMessagesArray` → `convertToImages`.

## Verification

- `pnpm run check` → 0 errors · `pnpm run build` → All good · prettier clean.
- **bugfixes suite (`npx tsx test/continuous-test-suite-bugfixes.ts`): 103 passed**, only the 2 known-environmental `updater:` tests fail locally (they resolve the installed `neurolink` binary, absent in a dev checkout; pass in CI) — unrelated to this change.
- New/updated tests:
  - `#260`: invalid `maxCanvasPixels` rejected; oversized page auto-downscales, normal page unaffected.
  - `#258`: encrypted PDF — required/incorrect/correct password at the converter.
  - **End-to-end threading**: `pdfOptions.password` + `maxCanvasPixels` flow through `buildMultimodalMessagesArray` into the image-fallback converter (the plumbing that was broken).
- **CLI E2E (azure)** — `generate ... --pdf password-protected.pdf`:
  - correct `--pdf-password` → decrypts (no password error)
  - wrong password → `Incorrect Password`
  - no flag → actionable `password-protected` guidance

## Docs
`docs/features/pdf-support.md`: new **Encrypted (Password-Protected) PDFs** and **Memory Safety: Page Canvas Limits** sections.

Fixes #260, #258.

## Summary by CodeRabbit

* **New Features**
  * Added support for password-protected PDFs through the SDK and CLI.
  * Added `--pdf-password` and `NEUROLINK_PDF_PASSWORD` configuration, with clear precedence and security warnings.
  * Added automatic PDF page downscaling to protect memory, with configurable pixel limits.
  * Added specific errors for missing or incorrect PDF passwords.
  * Improved handling of inline PDF attachments across generation and streaming requests.

* **Documentation**
  * Expanded PDF guidance with password handling, memory limits, downscaling behavior, and troubleshooting details.


test/continuous-test-suite-bugfixes.ts
```diff
@@ -478,6 +485,58 @@ const tests: TestFunction[] = [
...
+  {
+    name: "PDFProcessor #260: oversized page auto-downscales; normal page unaffected",
+    category: "pdf-processor",
+    fn: async () => {
+      const pdf = readFileSync("test/fixtures/valid-sample.pdf");
+      // A deliberately tiny ceiling forces the pre-flight to downscale rather
+      // than let pdf-to-img allocate the full canvas — must resolve, not crash.
+      const tiny = await PDFProcessor.convertToImages(pdf, {
+        maxCanvasPixels: 1000,
+      });
```
Tara-ag, line 1378: 💡 **MINOR: Test is fragile — it assumes the fixture has a plaintext MediaBox**

This assertion depends on `valid-sample.pdf` storing its `/MediaBox` in plaintext bytes. If the fixture is ever regenerated with compressed object streams (common with modern PDF producers), `largestPagePixels()` returns `0`, no downscale occurs, and this test fails even though the implementation is unchanged.

Consider making the test self-contained by constructing a minimal PDF with an inline `/MediaBox [0 0 10000 10000]` so the downscale path is deterministic, or at least assert that the fixture is suitable with a separate structural check. That would prevent future fixture churn from breaking the regression suite.

src/lib/utils/pdfProcessor.ts
```diff
@@ -281,6 +281,32 @@ export class PDFProcessor {
...
+    const text = pdfBuffer.toString("latin1");
+    const re =
+      /\/MediaBox\s*\[\s*(-?[\d.]+)\s+(-?[\d.]+)\s+(-?[\d.]+)\s+(-?[\d.]+)\s*\]/g;
+    let max = 0;
+    let match: RegExpExecArray | null;
+    while ((match = re.exec(text)) !== null) {
+      const width = Math.abs(parseFloat(match[3]) - parseFloat(match[1]));
+      const height = Math.abs(parseFloat(match[4]) - parseFloat(match[2]));
+      if (Number.isFinite(width) && Number.isFinite(height)) {
+        const pixels = width * height * scale * scale;
```
Tara-ag, line 308: 🔒 **CRITICAL: Non-finite pixel estimate can produce `effectiveScale === 0`**

`width * height * scale * scale` is not guarded against overflow. A PDF with a finite but enormous MediaBox can produce `Infinity` once the product exceeds `Number.MAX_VALUE` (~1.79e308). When `largestPixels` becomes `Infinity`, `downscale = sqrt(maxCanvasPixels / Infinity)` becomes `0`, so `effectiveScale = scale * 0 === 0`. Passing `scale: 0` to `pdf-to-img` creates a degenerate viewport and will almost certainly throw or render nothing.

This reinforces the unresolved Copilot review comment already raised on this function. Please address it.

Suggested fix — clamp the pixel product and ensure `effectiveScale` stays finite and strictly positive:

```typescript
private static largestPagePixels(pdfBuffer: Buffer, scale: number): number {
  const text = pdfBuffer.toString("latin1");
  const re =
    /\/MediaBox\s*\[\s*(-?[\d.]+)\s+(-?[\d.]+)\s+(-?[\d.]+)\s+(-?[\d.]+)\s*\]/g;
  let max = 0;
  let match: RegExpExecArray | null;
  while ((match = re.exec(text)) !== null) {
    const width = Math.abs(parseFloat(match[3]) - parseFloat(match[1]));
    const height = Math.abs(parseFloat(match[4]) - parseFloat(match[2]));
    if (Number.isFinite(width) && Number.isFinite(height)) {
      const scaledWidth = width * scale;
      const scaledHeight = height * scale;
      const pixels =
        Number.isFinite(scaledWidth) && Number.isFinite(scaledHeight)
          ? Math.min(scaledWidth * scaledHeight, Number.MAX_SAFE_INTEGER)
          : Number.MAX_SAFE_INTEGER;
      if (pixels > max) {
        max = pixels;
      }
    }
  }
  return max;
}
```

And in the downscale branch:

```typescript
if (largestPixels > maxCanvasPixels) {
  const downscale = Math.sqrt(maxCanvasPixels / largestPixels);
  effectiveScale = Math.max(0.001, scale * downscale); // keep a sane floor
  ...
}
```

This preserves the guard’s intent while preventing a zero-scale render.

Tara-ag (changes requested):
## Review Summary

**Files reviewed:** 5 changed files (processed one at a time)
**New issues raised:** 2
- 🔒 **CRITICAL** × 1 — `pdfProcessor.ts`: `largestPagePixels()` overflow can drive `effectiveScale` to `0`
- 💡 **MINOR** × 1 — `continuous-test-suite-bugfixes.ts`: downscale test depends on fixture having a plaintext MediaBox

**Blocking decision:** REQUEST_CHANGES

The CRITICAL issue meets the blocking criterion for a security/resource-hygiene flaw: an attacker-controlled PDF with a large-but-finite MediaBox can cause `width * height * scale * scale` to overflow to `Infinity`, which makes the computed downscale `0` and passes `scale: 0` to `pdf-to-img`. That produces a degenerate viewport and likely crashes or silently fails conversion, defeating the memory-exhaustion guard the PR is meant to add. This also reinforces the existing unresolved Copilot review comment on the same function.

**Next steps:**
1. Clamp the pixel product in `largestPagePixels()` to `Number.MAX_SAFE_INTEGER` (or otherwise guard against `Infinity`).
2. Ensure the computed `effectiveScale` has a strictly-positive floor (e.g. `Math.max(0.001, scale * downscale)`).
3. (Optional but recommended) Make the downscale regression test self-contained by synthesizing a PDF with a known inline `/MediaBox` instead of relying on `valid-sample.pdf` staying uncompressed.

Once the overflow guard is in place and CI passes, this should be good to merge.

docs/features/pdf-support.md
```diff
@@ -157,6 +157,55 @@ pdfFiles: [pdfBuffer];
...
+image-conversion path caps the rendered canvas at `maxCanvasPixels`
+(default 16,777,216 px ≈ 4096×4096). Oversized pages are automatically
+downscaled to fit the cap (a WARN is logged), so a malicious or malformed
+`/MediaBox` cannot force an unbounded allocation. Override the cap via
+`pdfOptions.maxCanvasPixels` when you need higher-resolution rendering:
+
+```typescript
+pdfOptions: {
+  maxCanvasPixels: 33_554_432; // 32 MP ceiling
+}
```
Tara-ag, line 216: 💡 **MINOR: Documentation object literal uses semicolon instead of comma**

The TypeScript example shows `maxCanvasPixels: 33_554_432;` inside an object literal. Object properties must be separated with commas, not semicolons.

```suggestion
pdfOptions: {
  maxCanvasPixels: 33_554_432, // 32 MP ceiling
}
```

src/lib/utils/messageBuilder.ts
Tara-ag, line 1: ⚠️ **MAJOR: `input.content` PDF path ignores `pdfOptions.password` and `pdfOptions.maxCanvasPixels`**

`convertContentToProviderFormat` builds the `pdfFiles` array for the image-fallback conversion but only copies `buffer`, `filename`, and `pageCount`. It drops `password` and `maxCanvasPixels`, so PDFs supplied via the advanced `input.content` array (`type: "pdf"`) bypass both the decryption fix (#258) and the canvas memory guard (#260).

This contradicts the PR's stated goal that `pdfOptions` is "threaded through `buildMultimodalMessagesArray` → `convertToImages`". The explicit `input.pdfFiles` path was fixed, but this parallel path was missed.

Suggested fix — pass `options.pdfOptions` into `convertContentToProviderFormat` and forward the fields:

```typescript
async function convertContentToProviderFormat(
  content: Content[],
  provider: string,
  _model: string,
  pdfOptions?: GenerateOptions["pdfOptions"],
): Promise<unknown> {
  ...
  const pdfFiles = pdfContent.map((pdf) => ({
    buffer:
      typeof pdf.data === "string" ? Buffer.from(pdf.data, "base64") : pdf.data,
    filename: pdf.metadata?.filename || "document.pdf",
    pageCount: pdf.metadata?.pages ?? null,
    password: pdfOptions?.password,
    maxCanvasPixels: pdfOptions?.maxCanvasPixels,
  }));
  ...
}
```

And update the call site in `buildMultimodalMessagesArray` to pass `options.pdfOptions`.

Tara-ag (approved):
## Review Summary

**Files reviewed:** 16 changed files (processed one at a time)
**New issues raised this run:** 2
- ⚠️ **MAJOR** × 1 — `src/lib/utils/messageBuilder.ts`: `input.content` PDF path drops `pdfOptions.password` and `pdfOptions.maxCanvasPixels`
- 💡 **MINOR** × 1 — `docs/features/pdf-support.md`: object literal uses semicolon instead of comma

**Blocking decision:** APPROVE

No new blocking criteria are met this run. The MAJOR issue is a real gap in the `input.content` (advanced multimodal content array) path, but it does not meet the blocking threshold on its own and the explicit `input.pdfFiles` path is correctly wired.

**Already-raised issues not repeated:**
- The `largestPagePixels()` Infinity/overflow concern and the fragile plaintext-MediaBox test fixture are already covered by existing review comments from earlier review passes.

**Next steps:**
1. Consider forwarding `pdfOptions` into `convertContentToProviderFormat` so the advanced `input.content` PDF path also supports password decryption and canvas limits.
2. Fix the docs TypeScript syntax typo (`;` → `,`).
3. Address the previously-raised overflow guard when revising the PR.

src/lib/utils/pdfProcessor.ts
```diff
@@ -281,6 +281,32 @@ export class PDFProcessor {
...
+        if (pixels > max) {
+          max = pixels;
+        }
+      }
```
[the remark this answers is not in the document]
murdore: Fixed. `largestPagePixels()` now clamps the per-page pixel product to `Number.MAX_SAFE_INTEGER` instead of letting it overflow to `Infinity` (guarding both the raw product and the finiteness check), and the downscale branch in `convertToImages` now floors `effectiveScale` at a new named constant `PDF_LIMITS.MIN_EFFECTIVE_SCALE` (0.1) so a pathological/malformed `/MediaBox` can never collapse the render scale to 0. Added a deterministic regression test (`PDFProcessor #260: overflowing MediaBox does not collapse effectiveScale below the floor`) that appends a decoy `/MediaBox` with ~1e160×1e160 dimensions (guaranteed to overflow `Number.MAX_VALUE` at the default scale) and asserts the render still succeeds with `effectiveScale >= PDF_LIMITS.MIN_EFFECTIVE_SCALE`. See `src/lib/utils/pdfProcessor.ts` (largestPagePixels + convertToImages) and `src/lib/core/constants.ts` (PDF_LIMITS.MIN_EFFECTIVE_SCALE). New HEAD: 5651eab2884965daa6f8fb3decc3386af3017a32.

src/lib/utils/pdfProcessor.ts
murdore: Fixed — same root cause as the Copilot comment above, addressed together. `largestPagePixels()` now bounds the pixel product with `Number.isFinite(pixels) ? Math.min(pixels, Number.MAX_SAFE_INTEGER) : Number.MAX_SAFE_INTEGER`, and `convertToImages` floors `effectiveScale = Math.max(PDF_LIMITS.MIN_EFFECTIVE_SCALE, scale * downscale)` (new named constant, 0.1, in `src/lib/core/constants.ts`) so it can never reach 0. Also recomputed the logged "after" MB estimate off the actual applied ratio so the WARN message stays honest when the floor kicks in. Added a deterministic regression test that reproduces the exact overflow (a decoy MediaBox with ~1e160 x 1e160 dimensions, guaranteed to exceed Number.MAX_VALUE) and asserts the render still succeeds with effectiveScale >= the floor. New HEAD: 5651eab2884965daa6f8fb3decc3386af3017a32.

test/continuous-test-suite-bugfixes.ts
murdore: Addressed by making the test self-contained rather than relying on the fixture's own MediaBox encoding. Added an `appendDecoyMediaBox()` test helper that appends a syntactically-standalone (unreferenced) PDF object carrying an explicit `/MediaBox` to the fixture buffer — `largestPagePixels()` does a raw byte-level regex scan so it picks this up regardless of how the fixture's own page MediaBox happens to be encoded. The 'tiny ceiling forces downscale' assertion now uses `appendDecoyMediaBox(pdf, "500", "500")` deterministically instead of depending on `valid-sample.pdf`'s plaintext MediaBox, with a comment explaining why. Also added a second test built the same way to cover the overflow case (see reply on the CRITICAL thread). New HEAD: 5651eab2884965daa6f8fb3decc3386af3017a32.

docs/features/pdf-support.md
murdore: Fixed exactly as suggested — changed `maxCanvasPixels: 33_554_432;` to `maxCanvasPixels: 33_554_432,` in docs/features/pdf-support.md. New HEAD: 5651eab2884965daa6f8fb3decc3386af3017a32.

src/lib/utils/messageBuilder.ts
murdore: Fixed as suggested. `convertContentToProviderFormat` now takes an optional `pdfOptions?: GenerateOptions["pdfOptions"]` parameter and forwards `password`/`maxCanvasPixels` onto each mapped pdfFiles entry (mirroring `processExplicitPdfFiles`); the call site in `buildMultimodalMessagesArray` now passes `options.pdfOptions`. While adding the end-to-end regression test for this I found a deeper, pre-existing gap in the same area: `hasPDFs` was computed purely from `pdfFiles.length > 0` (the explicit `input.pdfFiles` array) and never checked `inp.content` for `type === "pdf"` items, so a PDF supplied *only* via `input.content` (no image alongside it, no explicit pdfFiles) never even reached the multimodal branch — it silently fell through to the text-only path and vanished entirely, regardless of pdfOptions. Fixed that too (`hasPDFs` now also checks `inp.content.some(c => c.type === "pdf")`), since otherwise this fix would only take effect when a PDF happened to be mixed with an image in the same content array. Added a new regression test, `buildMultimodalMessagesArray threads pdfOptions.password + maxCanvasPixels through input.content PDF path`, mirroring the existing input.pdfFiles test: (a) correct password via input.content decrypts and yields image parts, (b) an invalid maxCanvasPixels via input.content surfaces the #260 validation error, proving both fields now arrive. New HEAD: 5651eab2884965daa6f8fb3decc3386af3017a32.

src/lib/utils/messageBuilder.ts
```diff
@@ -1390,7 +1404,14 @@ export async function buildMultimodalMessagesArray(
     (inp.images && inp.images.length > 0) ||
     (inp.content && inp.content.some((c) => c.type === "image"));
 
-  const hasPDFs = pdfFiles.length > 0;
+  // A PDF supplied only via input.content (type: "pdf", no explicit
+  // input.pdfFiles and no image alongside it) must still route through the
+  // multimodal path below — otherwise it silently falls through to the
```
Tara-ag, line 1438: ⚠️ **MAJOR: `hasPDFs` fix is incomplete — system prompt still ignores `input.content` PDFs**

The new `hasPDFs` logic correctly routes a PDF supplied only via `input.content` into the multimodal branch, but `buildMultimodalSystemPrompt` is still called with `pdfFiles.length > 0` a few lines below. That means an `input.content` PDF (with no explicit `input.pdfFiles`) will **not** receive the "TREAT THE INLINED CONTENT AS IF IT WERE AN ATTACHMENT" instruction, so the model can still respond with "no files attached" even though images were produced.

Mirror the same predicate when building the system prompt:

```typescript
const hasPDFFilesForPrompt =
  pdfFiles.length > 0 ||
  !!(inp.content && inp.content.some((c) => c.type === "pdf"));

const systemPrompt = buildMultimodalSystemPrompt(
  options,
  hasPDFFilesForPrompt,
);
```

This keeps the routing decision and the prompt guidance consistent.

src/cli/factories/commandFactory.ts
```diff
@@ -3607,6 +3612,7 @@ export class CLICommandFactory {
       const runGenerate = () =>
         sdk.generate({
           input: generateInput,
+          pdfOptions: { password: argv.pdfPassword as string | undefined },
```
Tara-ag: ⚠️ **MAJOR: CLI `--pdf-password` leaks into shell history and process listings**

Accepting a password as a plaintext CLI flag is a security anti-pattern: it is visible in `ps`/`proc`, shell history, and CI logs. The PR description says this is intentional ("CLI — use --pdf-password"), but the project already excludes `credentials` from the loop schema for exactly this reason.

At minimum, support reading from an environment variable or stdin prompt when the flag is omitted, and document that `--pdf-password` should only be used in non-shared environments. Better: change the flag to read the password from `NEUROLINK_PDF_PASSWORD` or prompt securely via `readline` (masking input) before committing to this interface, because adding it now locks in a leaky contract.

If you keep the flag, add a prominent warning when it is used:

```typescript
if (argv.pdfPassword && !argv.quiet) {
  logger.always(
    chalk.yellow(
      "⚠️  --pdf-password is visible in shell history and process listings. " +
      "Prefer NEUROLINK_PDF_PASSWORD or a secure prompt.",
    ),
  );
}
```

This is a public CLI API decision; once shipped, backward compatibility makes it hard to retract.

src/cli/factories/commandFactory.ts
```diff
@@ -3965,6 +3971,7 @@ export class CLICommandFactory {
           ...(videoFiles && { videoFiles }),
           ...(files && { files }),
         },
+        pdfOptions: { password: argv.pdfPassword as string | undefined },
```
Tara-ag: ⚠️ **MAJOR: Same plaintext password leak in stream path**

Same concern as the generate path: `--pdf-password` is passed as a plaintext CLI argument. If the flag is retained, mirror whatever mitigation you choose (env var, secure prompt, or warning) here so both paths behave consistently.

Tara-ag (approved):
## Review Summary

Reviewed 16 changed files. The PR addresses the two stated issues well: the canvas-memory guard is bounded and floored, password-protected PDFs get actionable typed errors, and `pdfOptions` is now threaded through the message builder for both `input.pdfFiles` and `input.content` paths.

### New issues raised (2 MAJOR)

1. **`src/lib/utils/messageBuilder.ts`** — The `hasPDFs` predicate was widened to include `input.content` PDFs, but `buildMultimodalSystemPrompt` is still invoked with the old `pdfFiles.length > 0` predicate. A PDF supplied only via `input.content` therefore enters the multimodal branch but does **not** receive the "treat inlined content as an attachment" system prompt, which can still produce "no files attached" responses.

2. **`src/cli/factories/commandFactory.ts`** — `--pdf-password` accepts a password as a plaintext CLI argument in both `generate` and `stream` paths. This exposes the password in shell history, `ps` output, and CI logs. The project already excludes `credentials` from the loop schema for the same reason. Consider adding an env-var fallback (`NEUROLINK_PDF_PASSWORD`) or a secure prompt, plus a warning when the flag is used.

Neither issue meets the 3-MAJOR blocking threshold, so approving with these comments for follow-up.

src/lib/utils/messageBuilder.ts
murdore: @/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-neurolink-fork-neurolink/6100829b-2172-4859-8bc3-78b9645e7410/scratchpad/reply1.txt

src/cli/factories/commandFactory.ts
murdore: @/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-neurolink-fork-neurolink/6100829b-2172-4859-8bc3-78b9645e7410/scratchpad/reply2.txt

src/cli/factories/commandFactory.ts
murdore: @/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-neurolink-fork-neurolink/6100829b-2172-4859-8bc3-78b9645e7410/scratchpad/reply3.txt

Tara-ag (approved):
## Review Summary

**Files reviewed:** 16  
**New issues raised this run:** 0 (CRITICAL: 0, MAJOR: 0, MINOR: 0, SUGGESTION: 0)  
**Existing review threads:** 8 — all resolved/outdated; no duplicates raised.

### Assessment

This PR delivers the two advertised PDF fixes cleanly:

- **#260 page-canvas memory guard:** `PDFProcessor.convertToImages` now pre-flights `/MediaBox` dimensions, clamps overflow to `Number.MAX_SAFE_INTEGER`, floors the effective scale at `PDF_LIMITS.MIN_EFFECTIVE_SCALE`, and validates `maxCanvasPixels`.
- **#258 password-protected PDFs:** `pdfOptions.password` / `--pdf-password` (with `NEUROLINK_PDF_PASSWORD` fallback) is threaded through `buildGenerateTextOptions`, `MessageBuilder`, `buildMultimodalMessagesArray`, and `convertToImages`; pdfjs `PasswordException` maps to typed `PDF_PASSWORD_REQUIRED` / `PDF_INCORRECT_PASSWORD` errors.
- **Plumbing gaps fixed:** `input.content[type="pdf"]` now routes through the multimodal path and receives the same `password`/`maxCanvasPixels` forwarding as `input.pdfFiles`, plus the system prompt gate is aligned with the routing predicate.

### Why no blocking issues

- No hardcoded secrets or credentials in non-test source.
- No breaking SDK/API changes — `pdfOptions` is additive across `GenerateOptions`, `TextGenerationOptions`, and `StreamOptions`.
- No violation of CLAUDE.md critical rules: providers remain dynamically imported, Gemini policy untouched, CLI-only concerns stay in CLI, `formatProviderError` unaffected.
- Password is not logged; the CLI warning is emitted to stderr and the env-var path avoids shell-history leakage.
- Tests cover invalid `maxCanvasPixels`, oversized/overflowing MediaBox, encrypted PDF required/incorrect/correct password, end-to-end `pdfOptions` threading for both `input.pdfFiles` and `input.content`, `hasPDFs` system-prompt gating, and `resolvePdfPassword` precedence/warning behavior.

Approved for merge.

src/cli/factories/commandFactory.ts
```diff
@@ -197,6 +197,13 @@ export class CLICommandFactory {
       type: "string" as const,
       description: "Add PDF file for analysis (can be used multiple times)",
     },
+    "pdf-password": {
```
Tara-ag, line 200: ⚠️ **MAJOR: `--pdf-password` is registered for batch but silently ignored there**

`"pdf-password"` is added to `commonOptions`, so every command that uses `buildOptions()` — including `batch` — advertises the flag. However, `executeBatch()` resolves PDF files via `processCliPDFFiles()` but never calls `resolvePdfPassword()` or forwards `pdfOptions` to `sdk.generate()`. A user running `neurolink batch prompts.txt --pdf protected.pdf --pdf-password s3cret` will have the flag accepted and then silently ignored, so encrypted PDFs fail with `PDF_PASSWORD_REQUIRED` even though a password was supplied.

Forward the password in `executeBatch()` the same way `csvOptions` and `videoOptions` are already conditionally attached:

```typescript
...(batchPdfFiles?.length && {
  pdfOptions: {
    password: CLICommandFactory.resolvePdfPassword(argv),
  },
}),
```

Tara-ag (approved):
## Review Summary

Reviewed the PDF subsystem PR (#260 memory guard + #258 password decryption). The fixes are well-scoped, thoroughly tested, and the round-3 review feedback (infinity guard, `input.content` PDF path, `hasPDFs` system-prompt gating, and CLI password leak mitigation) has been addressed.

### New issue raised

- **`src/cli/factories/commandFactory.ts`**: `--pdf-password` is registered globally via `commonOptions`, so the `batch` command accepts it, but `executeBatch()` never resolves or forwards the password. Encrypted PDFs in batch mode will therefore fail with `PDF_PASSWORD_REQUIRED` even when a password is supplied. Suggested a conditional `pdfOptions` forward matching the existing `csvOptions`/`videoOptions` pattern.

### Severity tally

- MAJOR: 1 (batch CLI flag silently ignored)
- No blocking criteria met (single MAJOR, no secrets, no breaking SDK changes, no violated non-negotiable architectural rules).

### Notes

- Appreciate the deterministic `appendDecoyMediaBox()` test helper and the regression coverage for overflow, password required/incorrect/correct, and end-to-end `pdfOptions` plumbing through both `input.pdfFiles` and `input.content`.
- The `NEUROLINK_PDF_PASSWORD` env-var fallback + stderr warning for the plaintext flag is a reasonable mitigation for the CLI leakage concern.

Approving; please consider fixing the batch path in a follow-up.

PR #1197, opened 2026-07-17
fix(sagemaker): accept AI SDK v3 flat tool shape — fixes crash on any generate() with tools

murdore:
## The bug

`convertToolsToSageMakerFormat()` assumed the nested OpenAI Chat Completions wire shape (`tool.function.name`), but `@ai-sdk/provider` delivers tools to `doGenerate()` in the **flat** `LanguageModelV3FunctionTool` shape (`{type:'function', name, inputSchema}`) — so **any** `generate()` reaching SageMaker with tools (including NeuroLink's always-on default agent tools) crashed with `TypeError: Cannot read properties of undefined (reading 'name')` before the request was even built. Every other provider already reads the flat shape (`v3ToolsToOpenAI`); only SageMaker had the wrong assumption.

## The fix

- Typed the boundary with the shared `OpenAICompatV3CallTools`/`OpenAICompatV3CallToolChoice` unions and narrowed on the `type` discriminant, mirroring `openaiChatCompletionsClient.ts` — no `any`.
- `type:'provider'` tools (no SageMaker wire equivalent) are dropped with an explicit `logger.debug`, not a crash.
- Also fixed the sibling `convertToolChoiceToSageMakerFormat()`: identical wrong-shape assumption (checked `type==='function'`, which `LanguageModelV3ToolChoice` never produces) — it silently passed the wrong wire format through. Same root cause, same file, fixed together.

## Regression test — fail-before / pass-after proven via git stash

New `test/continuous-test-suite-sagemaker-tools.ts` (+ `test:sagemaker-tools` script, wired into `test:unit`), 4 tests against the real `doGenerate()` with the endpoint client stubbed (zero AWS/network):
flat tool → correct nested request body · provider tool dropped + logged · `{type:'tool',toolName}` choice conversion · `{type:'auto'}` passthrough.
Pre-fix: 4/4 FAIL with the exact production error. Post-fix: 4/4 PASS.

Full-stack repro (fake creds, endpoint pinned to a refused loopback port — nothing leaves localhost): pre-fix dies inside tool conversion; post-fix proceeds all the way to the AWS SDK TCP connect. tsc/eslint/prettier clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SageMaker tool handling by converting supported function tools into the expected request format.
  * Safely ignores unsupported provider-specific tools instead of producing malformed requests or errors.
  * Improved conversion of tool-choice settings, including automatic, required, and named tool selections.

* **Tests**
  * Added regression coverage for SageMaker tool conversion and tool-choice behavior.
  * Integrated the SageMaker tool tests into the unit test workflow.


src/lib/providers/sagemaker/language-model.ts
```diff
@@ -549,13 +551,22 @@ export class SageMakerLanguageModel implements SageMakerAsLanguageModel {
...
+    // `LanguageModelV3FunctionTool` in @ai-sdk/provider), the same shape
+    // every other provider's tool converter consumes (compare
+    // `v3ToolsToOpenAI` in openaiChatCompletionsClient.ts). It is NOT the
+    // nested OpenAI Chat Completions wire format.
     const tools = (options as UnknownRecord).tools;
     if (tools && Array.isArray(tools) && tools.length > 0) {
-      request.tools = this.convertToolsToSageMakerFormat(tools);
+      request.tools = this.convertToolsToSageMakerFormat(
+        tools as OpenAICompatV3CallTools,
+      );
```
Tara-ag, line 565: ⚠️ **MAJOR**: If all tools are `type: 'provider'`, `convertToolsToSageMakerFormat()` returns `[]`, but the code still assigns `request.tools = []` and may set `request.tool_choice`. OpenAI-compatible endpoints reject `tools: []` (array must be non-empty if present) and `tool_choice` without tools. Guard the assignment so `request.tools` and `request.tool_choice` are only set when at least one tool survives conversion, and log when everything is dropped.

```ts
if (tools && Array.isArray(tools) && tools.length > 0) {
  const convertedTools = this.convertToolsToSageMakerFormat(
    tools as OpenAICompatV3CallTools,
  );

  if (convertedTools.length > 0) {
    request.tools = convertedTools;

    const toolChoice = (options as UnknownRecord)
      .toolChoice as OpenAICompatV3CallToolChoice;
    if (toolChoice) {
      request.tool_choice =
        this.convertToolChoiceToSageMakerFormat(toolChoice);
    }

    logger.debug("Added tool support to SageMaker request", {
      toolCount: convertedTools.length,
      toolChoice,
    });
  } else {
    logger.debug("All tools dropped for SageMaker request (no function tools)", {
      originalToolCount: tools.length,
    });
  }
}
```

Tara-ag (changes requested):
## Review Summary

**Files reviewed:** 3 (`src/lib/providers/sagemaker/language-model.ts`, `test/continuous-test-suite-sagemaker-tools.ts`, `package.json`)

**New issues raised:** 1 MAJOR

- **MAJOR** in `src/lib/providers/sagemaker/language-model.ts` (line 565): If all provided tools are `type: 'provider'`, the converter returns an empty array, but the request still sets `tools: []` and may set `tool_choice`. OpenAI-compatible endpoints reject an empty `tools` array and `tool_choice` without tools. Only attach `request.tools` / `request.tool_choice` when `convertedTools.length > 0`, and log when everything is dropped.

**Existing comments noted:** Two prior review comments (Copilot + CodeRabbit) already identified the same empty-tools issue. I am raising it once more with an executable code suggestion because it remains unaddressed in the current diff and is a concrete runtime failure mode.

**Blocking criterion:** This single MAJOR issue does not yet meet the "3 or more MAJOR" threshold, but it is a real correctness bug that will cause 400s from SageMaker endpoints when only provider-defined tools are present. Requesting changes to resolve it before approval.

**Next steps:**
1. Guard `request.tools` / `request.tool_choice` behind `convertedTools.length > 0`.
2. Add a regression test for the "all provider tools dropped" case asserting that `tools` and `tool_choice` are absent from the request body.
3. Re-run `pnpm run test:sagemaker-tools` and `pnpm run check`.

Once addressed, this PR should be good to merge — the flat-tool-shape fix itself is correct and well-tested.

Issue #384, opened 2025-12-01
CSV-020: Type Assertions Without Validation


murdore:
## Summary
Code uses `as Record<string, string>` type assertions without runtime validation, allowing invalid data shapes to pass through.

## Root Cause
Line 228 and 284 assert row types without checking. If csv-parser returns unexpected format, type errors occur at runtime.

## Fix
Add runtime validation: verify rows are objects, check all values are strings or expected types, validate required fields, throw typed errors for invalid data.

## Files Affected
`src/lib/utils/csvProcessor.ts`

## Acceptance Criteria
- [ ] Fix implemented and tested
- [ ] Unit tests added
- [ ] No regressions in existing CSV processing
- [ ] Documentation updated if needed

**Effort:** 1h

murdore:
### ◐ Audit status: partially done — keeping open

Core behaviour ships, but not every acceptance criterion is met. The remaining gap is the actionable part:

> A narrow defensive check exists for one specific filter, but the broader ask — validating all rows are objects, values are expected types, and throwing typed errors for invalid data — is not implemented; the type assertions are otherwise unguarded.

<details><summary>what already works (evidence)</summary>

src/lib/utils/csvProcessor.ts:778 `if (!row || typeof row !== "object") { return false; }` is a minimal runtime guard added inside the empty-row filter, but every other consumer still blindly does `rows[0] as Record<string, unknown>` (lines 447, 795, 1076, 1149) or casts array elements (867) without checking shape, and there is no throwing of typed validation errors for malformed rows anywhere.

</details>


<sub>— triaged in a full audit of open issues (release v9.84.2, 2026-07-12). Reopen if anything above is off.</sub>

murdore:
Closing — verified implemented in `release` (v10.10.6, `0b1abf1d`).

**Where it landed:** Implemented in PR #1199 (commit 3124f9eb): runtime type guard + assertion function validate every row at the streaming parse boundary and throw a typed, row-numbered error on violation; removed the raw `as Record<string,string>` casts.

**Evidence:**

[...]

Issue #379, opened 2025-12-01
CSV-016: No Timeout on Parsing


murdore:
## Summary
CSV parsing operations have no timeout, allowing malformed files or infinite loops to hang indefinitely.

## Root Cause
No timeout mechanism on parseCSVFile or parseCSVString. A malicious or malformed file could cause infinite parsing loop.

## Fix
Add configurable timeout (default 30s for strings, 5min for files). Abort parsing if timeout exceeded. Return partial results with timeout warning.

## Files Affected
`src/lib/utils/csvProcessor.ts`

## Acceptance Criteria
- [ ] Fix implemented and tested
- [ ] Unit tests added
- [ ] No regressions in existing CSV processing
- [ ] Documentation updated if needed

**Effort:** 2h

murdore:
Closing — verified implemented in `release` (v10.10.6, `0b1abf1d`).

**Where it landed:** Implemented in PR #1199 (commit 3124f9eb) with configurable timeoutMs/parseTimeoutMs option, default 30s (string) / 5min (file), CLI flag --csv-parse-timeout-ms.

**Evidence:**

[...]

Issue #378, opened 2025-12-01
CSV-015: Column Name Sanitization Missing


murdore:
## Summary
Column names with special characters, spaces, or starting with numbers are not sanitized, causing potential issues when used as object keys or in queries.

## Root Cause
csv-parser uses raw column names from header without sanitization. Names like `'Price ($)'`, `'1st Place'`, or `'Name/Title'` can cause issues.

## Fix
Sanitize column names: trim whitespace, convert to camelCase or snake_case, handle special characters, ensure names are valid identifiers, add option to preserve original names in metadata.

## Files Affected
`src/lib/utils/csvProcessor.ts`

## Acceptance Criteria
- [ ] Fix implemented and tested
- [ ] Unit tests added
- [ ] No regressions in existing CSV processing
- [ ] Documentation updated if needed

**Effort:** 2h

murdore:
Closing — verified implemented in `release` (v10.10.6, `0b1abf1d`).

**Where it landed:** Implemented in PR #1199 (commit 3124f9eb) as an opt-in feature (sanitizeColumnNames defaults to false to preserve backward compatibility), plus CLI flags --csv-sanitize-names/--csv-name-case.

**Evidence:**

[...]

Issue #375, opened 2025-12-01
CSV-013: No Error Context


murdore:
## Summary
CSV parsing errors lack context - no row number, column name, or value shown in error messages, making debugging impossible.

## Root Cause
Error handlers at lines 230 and 286 only reject with generic message: `Failed to parse CSV file: ${error.message}`. No line number, column, or problematic value.

## Fix
Enhance error messages with: row number where error occurred, column name/index, problematic value, expected format, file offset for large files.

## Files Affected
`src/lib/utils/csvProcessor.ts`

## Acceptance Criteria
- [ ] Fix implemented and tested
- [ ] Unit tests added
- [ ] No regressions in existing CSV processing
- [ ] Documentation updated if needed

**Effort:** 1h

murdore:
Closing — verified implemented in `release` (v10.10.6, `0b1abf1d`).

**Where it landed:** Implemented in PR #1199 (commit 3124f9eb), building on PR #1160's earlier row-count-in-errors partial fix.

**Evidence:**

src/lib/utils/csvProcessor.ts:259-288 `buildCsvParseErrorMessage()` (comment at 245-247: 'Parse Error Context (#375)') builds messages including stage, row count so far, location (file path), headerCount, bytesRead, lastRowColumnCount, and an RFC-4180 hint — used at every reject site: source-error (1641-1659), parser-error (1695-1716), and the file-open failure path (`csvFileAccessFailed` at 1330-1337, message: '[CSVProcessor] Failed to open CSV file (${filePath}): ...'). I independently verified via an ad-hoc script that `CSVProcessor.parseCSVFile('/does/not/exist/at/all.csv')` throws a message matching `[CSVProcessor]` and containing the bad path. Dedicated regression test 'CSVProcessor #375: bad path / directory yield [CSVProcessor] errors with file context' (test/continuous-test-suite-bugfixes.ts:521-552) checks both ENOENT and EISDIR cases include `[CSVProcessor]` and the path. PR #1199 (commit 3124f9eb) explicitly lists 'Fixes #375': 'Bad path → wrapped [CSVProcessor] Failed to open CSV file (...); shared buildCsvParseErrorMessage adds columns / last row / RFC-4180 hint.'

**Referenced merged PRs:** #1160,#1199


<sub>Verified by a source-level audit of all 110 open issues (2026-08-06): every claim was independently re-checked against the current tree before closing. Reopen if you see a gap I missed.</sub>

PR #1199, opened 2026-07-17
fix(csv): harden CSV processing — encoding, single-read, cleanup, context, sanitization, timeout, validation

### LINKED ISSUE #362: CSV-003: UTF-8 Encoding Hard-coded
## Summary
All CSV processing hard-codes UTF-8 encoding, causing silent data corruption for files in other encodings like Latin-1, Windows-1252, or UTF-16, which are common in European systems and database exports.

## Root Cause
The code uses `content.toString("utf-8")` at line 89 without any encoding detection. This assumes all CSV files are UTF-8, which is not true for database exports, Excel exports from non-English Windows, legacy systems, or international data.

## Fix
Add encoding detection using chardet library. Detect encoding before parsing, support common encodings (UTF-8, Latin-1, Windows-1252, UTF-16), add encoding parameter to API, include detected encoding in metadata.

## Files Affected
`src/lib/utils/csvProcessor.ts`

## Acceptance Criteria
- [ ] Unit tests added

**Effort:** 6h

murdore:
## Summary

Consolidated hardening of the **CSV subsystem** (`CSVProcessor` + RAG `CSVLoader`), fixing seven issues that all live on the shared streaming-parse path.

| Issue | Fix |
| ----- | --- |
| **#362** CSV-003 — encoding hard-coded to UTF-8 | New `textEncoding.decodeBuffer`: BOM → pure-ASCII fast-path → `chardet` → UTF-8 fallback (via `iconv-lite`). `encoding` override + `metadata.detectedEncoding`/`encodingConfidence`; `--csv-encoding`. |
| **#368** CSV-007 — file read twice | Single `createReadStream`: peek head to sniff encoding + a `sep=` metadata line, replay via `unshift`. Metadata skipping uses csv-parser's native `skipLines` (fixes the real header being consumed as data). |
| **#371** CSV-010 — streams leaked on parser error | Destination-error handler now destroys **both** source and parser; 10 MB `maxRowBytes` cap makes a pathological row fail fast. |
| **#375** CSV-013 — no error context | Bad path → wrapped `[CSVProcessor] Failed to open CSV file (...)`; shared `buildCsvParseErrorMessage` adds columns / last row / RFC-4180 hint. |
| **#378** CSV-015 — no column-name sanitization | Opt-in `sanitizeColumnNames`/`columnNameCase` (`Price ($)` → `price`), originals kept in `metadata.columnNameMapping` + `CSVColumnMetadata.originalName`. Applied in `CSVProcessor` **and** RAG `CSVLoader`; `--csv-sanitize-names` / `--csv-name-case`. |
| **#379** CSV-016 — parsing could hang | `parseTimeoutMs` (30s strings / 5min files) returns partial rows with `metadata.parseTimedOut`; `--csv-parse-timeout-ms`. |
| **#384** CSV-020 — unchecked row casts | `assertValidCsvRow`/`isValidCsvRow` validate each row; parse methods return `CsvRow[]`; 8 unchecked casts + the `analyzeCSV` tool cast removed. |

### Threading fix
`generate()` reconstructs `TextGenerationOptions` field-by-field in `buildGenerateTextOptions` and **silently dropped `csvOptions`** — so none of the `--csv-*` flags reached the parser. Now forwarded. `csvOptions` on Generate/Stream options references the canonical `CSVProcessorOptions` (removes a 3-way type duplication). *(This one line overlaps with #1195, which forwards both csvOptions + pdfOptions — a trivial rebase either way.)*

## Verification

- **bugfixes suite: 107 passed**, only the 2 known-environmental `updater:` tests fail locally (they resolve the installed `neurolink` binary, absent in a dev checkout; pass in CI).
- **8 new deterministic tests** (`csv-processor`): encoding detect / override / UTF-16LE-BOM / ASCII; single-read `sep=` metadata skip verified through the real **analyzeCSV agent tool**; both-stream cleanup on a >10 MB row; ENOENT/EISDIR context; sanitization + opt-out + `columnNameMapping`/`originalName`; file-based timeout (partial rows, no hang); row-shape guard + ragged/duplicate CSVs across all formats.
- End-to-end: `csvOptions.sanitizeColumnNames` verified threading `generate()` → `buildMultimodalMessagesArray` → `CSVProcessor` (output columns `[price, name_title]`).

## Deps
`chardet` + `iconv-lite` (both externalized in the browser + SSR bundles).

## Docs
`docs/features/csv-support.md`: `encoding`, `sanitizeColumnNames`/`columnNameCase`, `parseTimeoutMs` option sections; enriched error-handling + limitations.

Fixes #362, #368, #371, #375, #378, #379, #384.

## Summary by CodeRabbit

* **New Features**
  * Added configurable CSV character-encoding detection/override with confidence reporting.
  * Added opt-in CSV header sanitization with collision-safe deduplication and camelCase/snake_case output.
  * Added wall-clock CSV parse timeouts that return partial rows and set timeout metadata.
  * Added CLI flags for CSV encoding, header sanitization/name casing, and parse timeouts (applies to generation and streaming).
* **Bug Fixes**
  * Improved CSV robustness with stricter row-shape validation and safer prototype-pollution-resistant JSON output.
  * Enhanced CSV error typing/messages with richer context and safer stream teardown on failures.
* **Documentation**
  * Updated CSV support docs for new options, timeout semantics, limitations, and improved failure examples.
* **Tests**
  * Expanded regression coverage for encoding handling, `sep=` header parsing, sanitization mapping, timeouts, and invalid-row behavior.


src/lib/utils/csvProcessor.ts
```diff
@@ -831,12 +1039,37 @@ export class CSVProcessor {
...
+        origHeaders.forEach((h, i) => {
+          out[sanitized[i]] = row[h];
+        });
+        return out;
+      });
+      structuredColumnNameMapping = mapping.length > 0 ? mapping : undefined;
+    }
+
     // Extract metadata from parsed results
     const rowCount = nonEmptyRows.length;
```
Tara-ag, line 1163: 🔒 **CRITICAL: Prototype pollution via caller-controlled CSV headers.**

Building `CsvRow` objects with `{}` and then assigning user-supplied header strings as keys allows a crafted CSV header such as `__proto__`, `constructor`, or `prototype` to mutate `Object.prototype` rather than create an own property. Because CSV files are untrusted user input, this is a prototype-pollution vulnerability that can inject properties into every object in the process.

The repository’s recurring standard is to use null-prototype records whenever caller-controlled strings are used as record keys.

**Suggested fix:**
```diff
       nonEmptyRows = filteredRows.map((row) => {
-        const out: CsvRow = {};
+        const out: CsvRow = Object.create(null);
         origHeaders.forEach((h, i) => {
           out[sanitized[i]] = row[h];
         });
         return out;
       });
```

The same issue exists in `src/lib/rag/document/loaders.ts` when `CSVLoader` builds rows via `Object.fromEntries(headers.map((h, i) => [h, row[i]]))`.

src/lib/utils/csvProcessor.ts
```diff
@@ -942,98 +1207,124 @@ export class CSVProcessor {
...
+          cleanup();
+          const skipLines = toSkip(text);
+          // Replay the consumed head bytes, then decode the full raw stream.
+          rawSource.pause();
+          rawSource.unshift(head);
+          const decodeStream = iconv.decodeStream(
+            encoding,
+          ) as unknown as Transform;
+          rawSource.on("error", (e) => decodeStream.destroy(e));
+          resolve({ source: rawSource.pipe(decodeStream), skipLines });
```
Tara-ag: ⚠️ **MAJOR: rawSource file stream is not destroyed on abort/timeout/error.**

`source` here is the piped `decodeStream`, so `source.destroy()` only tears down the decoding transform. The underlying `rawSource` `fs.createReadStream` remains open, leaking a file descriptor and keeping the file handle alive. This is especially problematic for the timeout path (#379) and the oversized-row error path (#371), where the parser is intentionally aborted.

The earlier `prepareFileSource` error path does call `rawSource.destroy()`, but once `prepareFileSource` resolves, `streamParse` no longer has a reference to `rawSource`.

**Suggested fix:** keep `rawSource` in scope and destroy it alongside `source`/`parser` in the abort path. One minimal approach is to return `{ source, rawSource, skipLines }` from `prepareFileSource` and pass `rawSource` into `streamParse` so `abort()` can call `rawSource.destroy()` too.

src/lib/utils/csvProcessor.ts
```diff
@@ -734,6 +916,23 @@ export class CSVProcessor {
...
+        | undefined;
+      if (sanitizeColumnNames && limitedLines.length > 0) {
+        const headerCols = (limitedLines[0] || "").split(",");
+        const { sanitized, mapping } = buildColumnNameMapping(
+          headerCols,
+          columnNameCase,
+        );
+        limitedLines[0] = sanitized.join(",");
+        rawColumnNameMapping = mapping.length > 0 ? mapping : undefined;
+      }
```
Tara-ag, line 1004: ⚠️ **MAJOR: raw-format header split is not quote-aware, corrupting RFC 4180 headers.**

`(limitedLines[0] || "").split(",")` splits on commas even when they appear inside quoted fields. A header like `"Price, USD"` becomes two columns `\"Price` and `USD\"`, and the subsequent sanitization/mapping is then wrong. The file already contains `splitCsvLines` and `countUnquotedCommas` helpers that understand RFC 4180 quoting, so the raw branch should parse the header line with the same quote-aware logic before sanitizing.

**Suggested fix:** parse `limitedLines[0]` with a quote-aware CSV line splitter (the existing `parseCSVLine`-style logic or `csv-parser` itself) to obtain the true header cells, then sanitize and re-join them with proper CSV escaping.

src/lib/utils/csvProcessor.ts
```diff
@@ -942,98 +1207,124 @@ export class CSVProcessor {
...
-        .pipe(parser)
-        .on("data", (row: unknown) => {
-          lineCount++;
-          if (lineCount <= skipLines) {
+      const onData = (chunk: Buffer) => {
+        chunks.push(chunk);
+        total += chunk.length;
+        const head = Buffer.concat(chunks);
+        const { text, encoding } = decodeBuffer(head, encodingOverride);
+        if (splitCsvLines(text).length >= 2 || total >= PEEK_LIMIT) {
```
Tara-ag, line 1431: ⚠️ **MAJOR: ASCII fast path in `decodeBuffer` is applied to a peek sample, not the whole file.**

`prepareFileSource` calls `decodeBuffer(head, ...)` on only the first 64 KiB of the stream. Because `decodeBuffer` has a pure-ASCII fast path, any non-UTF-8 CSV whose header/early rows are ASCII-only will be classified as `utf-8` and the rest of the file will be decoded with the wrong encoding, reintroducing mojibake for later rows. The ASCII fast path is only safe when applied to the *entire* buffer.

**Suggested fix:** remove the ASCII fast path from the peek path, or restrict it to cases where the whole file fits in the peek buffer. For the streaming case, rely on BOM/chardet on the head sample and let `iconv.decodeStream` handle the remainder using that detected encoding.

src/lib/utils/csvProcessor.ts
```diff
@@ -88,6 +112,156 @@ function validateColumnName(name: string): string[] {
...
+ * Deduplicate a list of (already sanitized) names by suffixing `_2`, `_3`, … on
+ * collisions, preserving order.
+ */
+export function dedupeColumnNames(names: string[]): string[] {
+  const seen = new Map<string, number>();
+  return names.map((name) => {
+    const count = seen.get(name) ?? 0;
+    seen.set(name, count + 1);
+    return count === 0 ? name : `${name}_${count + 1}`;
+  });
```
Tara-ag, line 185: ⚠️ **MAJOR: `dedupeColumnNames` can emit duplicate names when a sanitized name collides with a previously emitted suffixed name.**

For input `["name", "name", "name_2"]`, the current logic emits `["name", "name_2", "name_2"]`. The third header is sanitized to `"name_2"`, which is already in use by the second header, so two columns end up sharing the same key and the second overwrites the first in row objects.

**Suggested fix:** track all emitted names and keep incrementing the suffix until a unique candidate is found:
```ts
export function dedupeColumnNames(names: string[]): string[] {
  const used = new Set<string>();
  const nextSuffix = new Map<string, number>();
  return names.map((name) => {
    let candidate = name;
    let suffix = nextSuffix.get(name) ?? 2;
    while (used.has(candidate)) {
      candidate = `${name}_${suffix++}`;
    }
    nextSuffix.set(name, suffix);
    used.add(candidate);
    return candidate;
  });
}
```

src/lib/types/file.ts
```diff
@@ -191,6 +201,19 @@ export type CSVColumnMetadata = {
   nameIssues?: string[];
 };
 
+/** A parsed CSV row: string-keyed with string (or missing) cell values (#384). */
```
Tara-ag, line 204: 💡 **MINOR: `CsvRow` type name does not follow the `CSV*` casing convention used elsewhere in this file.**

All other CSV-related types in this file use uppercase `CSV*` (`CSVColumnDataType`, `CSVColumnMetadata`, `CSVDataQualityWarning`, `CSVProcessorOptions`). The new `CsvRow` and `DecodedBuffer` types break that visual consistency. Per CLAUDE.md Critical Rule 9 (unique, prefixed type names), consider `CSVRow` and `DecodedBuffer` is fine but could be `TextEncodingDecodedBuffer` if there is any collision risk.

Not blocking, but renaming `CsvRow` → `CSVRow` would align with the surrounding types.

src/cli/factories/commandFactory.ts
```diff
@@ -3614,6 +3637,13 @@ export class CLICommandFactory {
               | "markdown"
               | "json"
               | undefined,
+            encoding: argv.csvEncoding as string | undefined,
```
Tara-ag: 💡 **MINOR: CLI `csvOptions` construction duplicates the same object shape in `executeGenerate` and `executeRealStream`.**

The `csvOptions` bag is built identically in both the generate and stream handlers. If more fields are added later, the two sites will drift again. Consider extracting a small helper `buildCsvOptionsFromArgv(argv)` that returns `CSVProcessorOptions | undefined` and using it in both places. This keeps the CLI flag plumbing in one location and matches the existing `buildGenerateOutputConfig` helper pattern.

src/lib/rag/document/loaders.ts
Tara-ag, line 1: 🔒 **CRITICAL: Prototype pollution via caller-controlled CSV headers in RAG loader.**

In the `CSVLoader.load` JSON-output path, row objects are built from caller-controlled header strings using either `Object.fromEntries(headers.map((h, i) => [h, row[i]]))` or an equivalent `{}` + assignment pattern. A CSV with a header like `__proto__` will pollute `Object.prototype` for the whole process, because ordinary object literals trigger the legacy `__proto__` setter.

This is the same class of vulnerability as in `CSVProcessor.process()`; the RAG ingestion path must also use null-prototype records for header keys.

**Suggested fix:** build each row object explicitly with `Object.create(null)`:
```ts
rows.map((row) => {
  const obj: Record<string, string> = Object.create(null);
  headers.forEach((h, i) => {
    obj[h] = row[i];
  });
  return obj;
})
```

src/lib/utils/csvProcessor.ts
```diff
@@ -88,6 +112,156 @@ function validateColumnName(name: string): string[] {
...
+// ============================================================================
+
+/**
+ * Predicate form of the row-shape invariant: a non-null, non-array object whose
+ * every value is a string or undefined. csv-parser's default output always
+ * satisfies this; the guard is defense-in-depth at the parse boundary.
+ */
+export function isValidCsvRow(row: unknown): row is CsvRow {
+  if (!isObject(row) || Array.isArray(row)) {
+    return false;
```
Tara-ag, line 219: ⚠️ **MAJOR: Row-shape validation does not protect against prototype-polluting column keys.**

`isValidCsvRow` only checks that values are strings/undefined; it accepts a row whose keys are `__proto__`, `constructor`, or `prototype`. Because `csv-parser` emits plain `{}` objects and the structured branch later copies values into another plain `{}` object, a malicious CSV header can pollute `Object.prototype` before this guard ever runs.

**Suggested fix:**
1. Reject rows that contain dangerous keys in `isValidCsvRow`:
```ts
const DANGEROUS_KEYS = new Set(["__proto__", "constructor", "prototype"]);
export function isValidCsvRow(row: unknown): row is CsvRow {
  if (!isObject(row) || Array.isArray(row)) return false;
  for (const key of Object.keys(row)) {
    if (DANGEROUS_KEYS.has(key)) return false;
  }
  return Object.values(row).every((v) => typeof v === "string" || v === undefined);
}
```
2. Build row objects with `Object.create(null)` everywhere caller-controlled headers become keys (e.g. the structured sanitization remap at line 1070).

murdore:
Addressed the review findings (force-pushed):

- **CodeQL ReDoS (csvProcessor.ts:131)** — `sanitizeColumnName` no longer uses the polynomial `^_+|_+$` trim; it splits on non-alphanumeric runs (linear). A 50k-underscore input now sanitizes in <1 ms.
- **Dedupe collision (dedupeColumnNames)** — a generated suffix is now bumped until it collides with no already-emitted name, so `["name","name","name_2"]` → `["name","name_2","name_2_2"]` (was `[...,"name_2","name_2"]`).
- **Quote-aware raw header sanitization** — raw-format header sanitization uses a new RFC-4180 field splitter, so `"Price, USD"` is one column (`price_usd`), not two.
- **Docs** — fixed the two TS object-literal examples that used `;` instead of `,`.
- **#1282 (ASCII head → encoding)** — documented the streaming trade-off inline: the encoding is committed from the peeked head to keep the parse streaming (which the parse-timeout guard relies on); a file whose head is pure-ASCII but switches to a legacy encoding only later should pass `encoding` explicitly. Buffering the whole file for detection was tried but it defeats the streaming timeout, so the explicit-override path is the documented remedy.

New regression test added covering dedupe/ReDoS/quote-aware. `check`/`lint`/`build` clean; bugfixes suite green except the 2 known-environmental `updater:` tests.

src/lib/utils/csvProcessor.ts
```diff
@@ -831,12 +1085,37 @@ export class CSVProcessor {
...
+        origHeaders.forEach((h, i) => {
+          out[sanitized[i]] = row[h];
+        });
+        return out;
+      });
+      structuredColumnNameMapping = mapping.length > 0 ? mapping : undefined;
+    }
+
     // Extract metadata from parsed results
     const rowCount = nonEmptyRows.length;
```
Tara-ag, line 1163: 🔒 **CRITICAL: Prototype pollution via caller-controlled CSV headers.**

Building `CsvRow` objects with `{}` and then assigning user-supplied header strings as keys allows a crafted CSV header such as `__proto__`, `constructor`, or `prototype` to mutate `Object.prototype` rather than create an own property. Because CSV files are untrusted user input, this is a prototype-pollution vulnerability that can inject properties into every object in the process.

The repository’s recurring standard is to use null-prototype records whenever caller-controlled strings are used as record keys.

**Suggested fix:**
```diff
       nonEmptyRows = filteredRows.map((row) => {
-        const out: CsvRow = {};
+        const out: CsvRow = Object.create(null);
         origHeaders.forEach((h, i) => {
           out[sanitized[i]] = row[h];
         });
         return out;
       });
```

The same issue exists in `src/lib/rag/document/loaders.ts` when `CSVLoader` builds rows via `Object.fromEntries(headers.map((h, i) => [h, row[i]]))`.

src/lib/utils/csvProcessor.ts
```diff
@@ -801,6 +1048,12 @@ export class CSVProcessor {
             undefined,
           ),
```
Tara-ag: ⚠️ **MAJOR: Raw-format metadata still splits the header naïvely on commas, so quoted commas in headers misreport `columnCount` and `hasHeaders`.**

The sanitization branch now uses `splitCsvFields` (quote-aware), but the metadata computation here still uses `(limitedLines[0] || "").split(",")`. A header like `"Price, USD",Qty` yields `columnCount: 3` and passes `['"Price', ' USD"', 'Qty']` to `detectHasHeaders`, so both metadata values are wrong. The same naive split is also used for `columnCount` a few lines above.

**Suggested fix:** reuse `splitCsvFields(limitedLines[0] || "")` for `columnCount`, `detectHasHeaders`, and the `logger.info` payload so the raw path's metadata matches the quote-aware sanitization branch.

```diff
-          columnCount: (limitedLines[0] || "").split(",").length,
+          columnCount: splitCsvFields(limitedLines[0] || "").length,
...
-            (limitedLines[0] || "").split(","),
+            splitCsvFields(limitedLines[0] || ""),
```

src/lib/utils/csvProcessor.ts
```diff
@@ -831,12 +1085,37 @@ export class CSVProcessor {
...
+    let structuredColumnNameMapping:
+      | Array<{ original: string; sanitized: string }>
+      | undefined;
+    const sanitizedToOriginal = new Map<string, string>();
+    let nonEmptyRows: CsvRow[] = filteredRows;
+    if (sanitizeColumnNames && filteredRows.length > 0) {
+      const origHeaders = Object.keys(filteredRows[0]);
+      const { sanitized, mapping } = buildColumnNameMapping(
+        origHeaders,
+        columnNameCase,
```
Tara-ag, line 1141: 🔒 **CRITICAL: Prototype pollution via caller-controlled CSV headers remains possible in the structured remap path.**

`filteredRows` comes from `csv-parser`, which emits plain `{}` objects. If a CSV header is `__proto__`, the parser already mutates `Object.prototype` before this code runs. The remap then creates another plain `{}` object and assigns the same dangerous key, compounding the pollution. The existing `isValidCsvRow` guard does not reject dangerous keys.

Per the repository's recurring standard and CLAUDE.md Critical Rule 5 / security posture, records keyed by caller-controlled strings must use null-prototype records.

**Suggested fix:**
1. Build the remapped row with `Object.create(null)`:
```diff
      nonEmptyRows = filteredRows.map((row) => {
-        const out: CsvRow = {};
+        const out: CsvRow = Object.create(null);
        origHeaders.forEach((h, i) => {
          out[sanitized[i]] = row[h];
        });
        return out;
      });
```
2. Reject dangerous keys in `isValidCsvRow` (e.g. `__proto__`, `constructor`, `prototype`) so malformed rows abort early rather than silently polluting.

This is the same class of vulnerability noted on the RAG loader path.

src/lib/utils/csvProcessor.ts
```diff
@@ -942,98 +1253,129 @@ export class CSVProcessor {
...
+      };
 
-          rows.push(row);
-          count++;
+      const onEnd = () => {
+        if (settled) {
+          return;
+        }
+        settled = true;
+        cleanup();
```
Tara-ag, line 1463: ⚠️ **MAJOR: `rawSource` file stream is not destroyed on abort/timeout/error in `streamParse`.**

`source` here is the piped `decodeStream`; `source.destroy()` only tears down the decoding transform. The underlying `rawSource` `fs.createReadStream` remains open, leaking a file descriptor. This is especially problematic for the timeout path (#379) and the oversized-row error path (#371), where the parser is intentionally aborted.

`prepareFileSource` does call `rawSource.destroy()` on its own error path, but once it resolves, `streamParse` no longer has a reference to `rawSource`.

**Suggested fix:** return `rawSource` from `prepareFileSource` alongside `source` and `skipLines`, and pass it into `streamParse` so the abort path can call `rawSource.destroy()` too. For the string path, `rawSource` can be `undefined`.

```diff
-    return this.streamParse(prepared.source, {
+    return this.streamParse(prepared.source, prepared.rawSource, {
       maxRows: clampedMaxRows,
       skipLines: prepared.skipLines,
       timeoutMs,
       location: filePath,
     });
```

Then in `streamParse`:
```diff
      const abort = () => {
        source.destroy();
        parser.destroy();
+       rawSource?.destroy();
      };
```

src/lib/rag/document/loaders.ts
Tara-ag, line 1: 🔒 **CRITICAL: Prototype pollution via caller-controlled CSV headers in RAG loader.**

In the `CSVLoader.load` JSON-output path, row objects are built from caller-controlled header strings using `Object.fromEntries(headers.map((h, i) => [h, row[i]]))`. A CSV with a header like `__proto__` will pollute `Object.prototype` for the whole process, because ordinary object literals trigger the legacy `__proto__` setter.

This is the same class of vulnerability as in `CSVProcessor.process()`; the RAG ingestion path must also use null-prototype records for header keys.

**Suggested fix:** build each row object explicitly with `Object.create(null)`:
```ts
rows.map((row) => {
  const obj: Record<string, string> = Object.create(null);
  headers.forEach((h, i) => {
    obj[h] = row[i];
  });
  return obj;
})
```

src/lib/utils/csvProcessor.ts
```diff
@@ -942,98 +1253,129 @@ export class CSVProcessor {
...
+  /**
+   * Peek the head of a raw file stream to resolve encoding + metadata-line
+   * skipping, then return a decoded text stream that still delivers the whole
+   * file from byte 0 (via `unshift`) — a single disk read (#368/#362).
+   */
+  private static prepareFileSource(
+    rawSource: Readable,
+    encodingOverride?: string,
+  ): Promise<{ source: Readable; skipLines: number }> {
+    return new Promise((resolve, reject) => {
```
Tara-ag, line 1393: ⚠️ **MAJOR: `parseTimeoutMs` does not cover `fs.promises.access` or `prepareFileSource` head sniffing.**

The timeout only starts inside `streamParse`, after `prepareFileSource` has already consumed up to 64 KiB and possibly waited on a slow/stalled file stream. A hung or extremely slow filesystem before the first complete line can therefore block the caller indefinitely, defeating the purpose of #379.

Per the bootstrapped standard, async calls should be bounded by the project's timeout helper. Wrapping the access + prepare + parse phases in a single deadline would also prevent the timeout budget from being silently consumed before parsing begins.

**Suggested fix:** wrap `fs.promises.access` and `prepareFileSource` together with `streamParse` under one `withTimeout` (or Promise.race) using `timeoutMs`, so the cap governs the entire file-parse operation.

Tara-ag (changes requested):
## Review summary

Reviewed the 16 changed files in this CSV-hardening PR. The changes meaningfully address the stated issues (encoding detection, single-read parsing, stream cleanup, error context, sanitization, timeouts, row validation), but several unresolved **CRITICAL** and **MAJOR** issues remain.

### New issues raised in this run

| Severity | File | Topic |
|----------|------|-------|
| 🔒 CRITICAL | `src/lib/rag/document/loaders.ts` | Prototype pollution via caller-controlled CSV headers in `CSVLoader` JSON output (`Object.fromEntries` on plain object). |
| ⚠️ MAJOR | `src/lib/utils/csvProcessor.ts` | Raw-format metadata still splits the header naïvely on commas, so quoted commas corrupt `columnCount` and `hasHeaders` even though the sanitization rewrite is now quote-aware. |
| ⚠️ MAJOR | `src/lib/utils/csvProcessor.ts` | `rawSource` file stream is not destroyed on abort/timeout/error in `streamParse`; only the decoded `source` transform is torn down, leaking the fd. |
| ⚠️ MAJOR | `src/lib/utils/csvProcessor.ts` | `parseTimeoutMs` only starts inside `streamParse`, so `fs.promises.access` and `prepareFileSource` head-sniffing can hang indefinitely before the deadline applies. |

### Previously raised issues that are still unresolved and block approval

- 🔒 Prototype pollution via caller-controlled CSV headers in `CSVProcessor.process()` structured remap (existing comment on the `CsvRow = {}` assignment).
- ⚠️ ASCII fast path in `decodeBuffer` applied to a peek sample rather than the whole file, risking mojibake for files that switch encodings after the head.
- ⚠️ Raw-format header split not quote-aware (partially fixed by `splitCsvFields`, but metadata still uses naive split — see new comment above).
- ⚠️ `dedupeColumnNames` collision logic, `ErrorFactory` usage, `lastRow` leakage in error messages, global prototype patching in tests, and CLI `csvOptions` duplication were also raised earlier and remain open.

### Next steps

Please address the prototype-pollution issues (preferably by sanitizing/rejecting dangerous headers at parse time, e.g. via `csv-parser`'s `mapHeaders`, and using `Object.create(null)` for all row objects built from caller-controlled headers), ensure `rawSource` is destroyed alongside `source`/`parser`, extend the parse timeout to cover access + head sniffing, and make the raw-format metadata quote-aware. Once these are resolved and tests pass, this should be good to merge.

**Decision: REQUEST_CHANGES** — unresolved CRITICAL security issues and multiple MAJOR correctness/resource-leak issues remain.

src/lib/utils/csvProcessor.ts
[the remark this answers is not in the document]
murdore: Fixed in 7b0ca42c4f16d2f91d549aa48632991a23d29ba4. The sanitizeColumnNames branch already used the quote-aware `splitCsvFields(limitedLines[0] || "")`; the sibling metadata computation (columnCount logging, `metadata.columnCount`, and the `detectHasHeaders(...)` input) still used the naive `(limitedLines[0] || "").split(",")` in the raw-format branch. All three call sites now use `splitCsvFields`, matching the sanitization branch. Added a regression test ("CSVProcessor #1199: raw-format columnCount/hasHeaders are quote-aware, matching sanitizeColumnNames") that processes a `"Price, USD",Qty` header with `formatStyle: "raw"` and asserts `columnCount === 2` and `hasHeaders === true`.

murdore: Fixed in 7b0ca42c4f16d2f91d549aa48632991a23d29ba4. The sanitizeColumnNames branch already used the quote-aware `splitCsvFields(limitedLines[0] || "")`; the sibling metadata computation (columnCount logging, `metadata.columnCount`, and the `detectHasHeaders(...)` input) still used the naive `(limitedLines[0] || "").split(",")` in the raw-format branch. All three call sites now use `splitCsvFields`, matching the sanitization branch. Added a regression test ("CSVProcessor #1199: raw-format columnCount/hasHeaders are quote-aware, matching sanitizeColumnNames") that processes a `"Price, USD",Qty` header with `formatStyle: "raw"` and asserts `columnCount === 2` and `hasHeaders === true`.

src/lib/utils/csvProcessor.ts
murdore: Fixed in 7b0ca42c4f16d2f91d549aa48632991a23d29ba4. The sanitizeColumnNames branch already used the quote-aware `splitCsvFields(limitedLines[0] || "")`; the sibling metadata computation (columnCount logging, `metadata.columnCount`, and the `detectHasHeaders(...)` input) still used the naive `(limitedLines[0] || "").split(",")` in the raw-format branch. All three call sites now use `splitCsvFields`, matching the sanitization branch. Added a regression test ("CSVProcessor #1199: raw-format columnCount/hasHeaders are quote-aware, matching sanitizeColumnNames") that processes a `"Price, USD",Qty` header with `formatStyle: "raw"` and asserts `columnCount === 2` and `hasHeaders === true`.

src/lib/utils/csvProcessor.ts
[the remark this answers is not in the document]
murdore: Fixed in 7b0ca42c4f16d2f91d549aa48632991a23d29ba4, via the fix_plan's explicitly-permitted lower-confidence variant ("keep the utf-8 default but with lower confidence, so callers relying on the confidence score can detect the ambiguity"). `decodeBuffer` now takes an `isCompleteBuffer` parameter (default `true`); `prepareFileSource`'s onData peek path passes `false`, so the ASCII fast-path reports `confidence: 60` instead of `100` for a partial peek. The streaming decode still commits to the peeked encoding for the rest of the file (documented, pre-existing trade-off noted in the adjacent #1282 comment, needed to keep the parse-timeout guard #379 streaming rather than buffering) — this fix specifically closes the over-claimed-certainty gap flagged here, surfacing the ambiguity via `result.metadata.encodingConfidence` instead of silently reporting 100. Added a direct regression test ("decodeBuffer (#1199): a partial ASCII peek reports lower confidence than a complete ASCII buffer") asserting 100 vs 60. For files with an ASCII head that switches encoding later, the documented workaround remains passing an explicit `encoding` override.

murdore: Fixed in 7b0ca42c4f16d2f91d549aa48632991a23d29ba4, via the fix_plan's explicitly-permitted lower-confidence variant ("keep the utf-8 default but with lower confidence, so callers relying on the confidence score can detect the ambiguity"). `decodeBuffer` now takes an `isCompleteBuffer` parameter (default `true`); `prepareFileSource`'s onData peek path passes `false`, so the ASCII fast-path reports `confidence: 60` instead of `100` for a partial peek. The streaming decode still commits to the peeked encoding for the rest of the file (documented, pre-existing trade-off noted in the adjacent #1282 comment, needed to keep the parse-timeout guard #379 streaming rather than buffering) — this fix specifically closes the over-claimed-certainty gap flagged here, surfacing the ambiguity via `result.metadata.encodingConfidence` instead of silently reporting 100. Added a direct regression test ("decodeBuffer (#1199): a partial ASCII peek reports lower confidence than a complete ASCII buffer") asserting 100 vs 60. For files with an ASCII head that switches encoding later, the documented workaround remains passing an explicit `encoding` override.

src/lib/utils/csvProcessor.ts
```diff
@@ -88,6 +112,156 @@ function validateColumnName(name: string): string[] {
...
+export function assertValidCsvRow(
+  row: unknown,
+  rowNumber: number,
+): asserts row is CsvRow {
+  if (!isValidCsvRow(row)) {
+    throw new Error(
+      `[CSVProcessor] Invalid CSV row ${rowNumber}: expected a string-keyed object with string values, got ${
+        Array.isArray(row) ? "array" : typeof row
+      }`,
+    );
```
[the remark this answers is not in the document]
murdore: Fixed in 7b0ca42c4f16d2f91d549aa48632991a23d29ba4. Added CSV-specific `ErrorFactory` methods (`csvInvalidInput`, `csvRowInvalid` — VALIDATION category; `csvFileAccessFailed` — RESOURCE category; `csvParseFailed` — EXECUTION category) in `src/lib/utils/errorHandling.ts`, and replaced the `new Error(...)` call sites in csvProcessor.ts (row-shape assertion, file-access, prepareFileSource error-wrap, string-parse, and the two streamParse reject sites) with them, preserving the existing message text and `cause`. Added a regression test ("CSVProcessor #1199: CSV failures are typed NeuroLinkError via ErrorFactory, message text preserved") covering an empty-path VALIDATION error, an ENOENT RESOURCE error, and a malformed-row VALIDATION error via `assertValidCsvRow`, each asserting `instanceof NeuroLinkError` plus the expected `ErrorCategory`.

src/lib/utils/csvProcessor.ts
```diff
@@ -942,98 +1207,124 @@ export class CSVProcessor {
...
+    }
 
-      const source = fs.createReadStream(filePath, { encoding: "utf-8" });
-      const parser = csvParser();
+    return this.streamParse(prepared.source, {
+      maxRows: clampedMaxRows,
+      skipLines: prepared.skipLines,
+      timeoutMs,
+      location: filePath,
+    });
```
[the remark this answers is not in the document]
murdore: Fixed in 7b0ca42c4f16d2f91d549aa48632991a23d29ba4. `parseCSVFile` now wraps the whole `fs.promises.access()` + `prepareFileSource()` + `streamParse()` sequence in a single `withTimeout(..., timeoutMs)` call, so one wall-clock budget governs the entire operation instead of the deadline starting only after `prepareFileSource` had already resolved (which could itself hang on a stalled stream before the first 2 lines / 64KiB). Added a regression test ("CSVProcessor #1199: a single deadline covers access()+prepareFileSource(), not just streamParse") reusing the existing #379 large-file (200k rows) / `timeoutMs=1` race pattern, asserting `timedOut === true`.

src/lib/utils/csvProcessor.ts
murdore: Fixed in 7b0ca42c4f16d2f91d549aa48632991a23d29ba4. `parseCSVFile` now wraps the whole `fs.promises.access()` + `prepareFileSource()` + `streamParse()` sequence in a single `withTimeout(..., timeoutMs)` call, so one wall-clock budget governs the entire operation instead of the deadline starting only after `prepareFileSource` had already resolved (which could itself hang on a stalled stream before the first 2 lines / 64KiB). Added a regression test ("CSVProcessor #1199: a single deadline covers access()+prepareFileSource(), not just streamParse") reusing the existing #379 large-file (200k rows) / `timeoutMs=1` race pattern, asserting `timedOut === true`.

src/lib/utils/csvProcessor.ts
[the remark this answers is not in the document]
murdore: Fixed in 7b0ca42c4f16d2f91d549aa48632991a23d29ba4. `prepareFileSource` now returns `{ source, rawSource, skipLines }` and `streamParse` threads `rawSource` through to its `abort()` cleanup, which now calls `rawSource.destroy()` alongside `source.destroy()` and `parser.destroy()`. Previously only the piped `iconv.decodeStream` transform (`source`) was destroyed — `.pipe()` does not propagate `.destroy()` back to its upstream source, so the underlying `fs.createReadStream` file descriptor stayed open on every abort/timeout/oversized-row path. Added a regression test ("CSVProcessor #1199: file-path parse also destroys the raw fs.createReadStream on abort") that patches `Readable.prototype.destroy`/`Transform.prototype.destroy` globally (same technique as the existing #371 test) and asserts both `ReadStream` and `CsvParser` are torn down when an oversized row aborts the parse.

murdore: Fixed in 7b0ca42c4f16d2f91d549aa48632991a23d29ba4. `prepareFileSource` now returns `{ source, rawSource, skipLines }` and `streamParse` threads `rawSource` through to its `abort()` cleanup, which now calls `rawSource.destroy()` alongside `source.destroy()` and `parser.destroy()`. Previously only the piped `iconv.decodeStream` transform (`source`) was destroyed — `.pipe()` does not propagate `.destroy()` back to its upstream source, so the underlying `fs.createReadStream` file descriptor stayed open on every abort/timeout/oversized-row path. Added a regression test ("CSVProcessor #1199: file-path parse also destroys the raw fs.createReadStream on abort") that patches `Readable.prototype.destroy`/`Transform.prototype.destroy` globally (same technique as the existing #371 test) and asserts both `ReadStream` and `CsvParser` are torn down when an oversized row aborts the parse.

src/lib/utils/csvProcessor.ts
murdore: Fixed in 7b0ca42c4f16d2f91d549aa48632991a23d29ba4. `prepareFileSource` now returns `{ source, rawSource, skipLines }` and `streamParse` threads `rawSource` through to its `abort()` cleanup, which now calls `rawSource.destroy()` alongside `source.destroy()` and `parser.destroy()`. Previously only the piped `iconv.decodeStream` transform (`source`) was destroyed — `.pipe()` does not propagate `.destroy()` back to its upstream source, so the underlying `fs.createReadStream` file descriptor stayed open on every abort/timeout/oversized-row path. Added a regression test ("CSVProcessor #1199: file-path parse also destroys the raw fs.createReadStream on abort") that patches `Readable.prototype.destroy`/`Transform.prototype.destroy` globally (same technique as the existing #371 test) and asserts both `ReadStream` and `CsvParser` are torn down when an oversized row aborts the parse.

test/continuous-test-suite-bugfixes.ts
```diff
@@ -207,6 +213,285 @@ const tests: TestFunction[] = [
...
+          return orig.apply(this, a);
+        };
+        return () => {
+          proto.destroy = orig;
+        };
+      };
+      const restores = [
+        patch(Readable.prototype as unknown as { destroy: () => unknown }),
+        patch(Transform.prototype as unknown as { destroy: () => unknown }),
+      ];
```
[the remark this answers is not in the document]
murdore: Acknowledged as accurate but not changed — this repo's test runner (test/continuous-test-suite.ts) executes all tests strictly sequentially via `for (const test of tests) { await test.fn(); }`, never concurrently, so the cross-test-interference scenario this flags does not apply to how this suite actually runs. The patch is also always unwound in a `finally` block. Left the existing pattern (shared with the pre-existing #371 test) as-is, and added a short comment at the patch site explaining why it's safe in this suite specifically, so a future reader doesn't have to re-derive that reasoning.

src/lib/utils/csvProcessor.ts
murdore: Investigated in depth; concluded this specific report is a false positive (verified experimentally in Node, see the full writeup in the triage): (1) csv-parser@3.2.1 (the pinned version) already has its own `DANGEROUS_KEYS` header sanitizer that maps `__proto__`/`constructor`/`prototype` headers to `null` and drops the column before a row object is ever built, so `filteredRows`/`origHeaders` can't literally contain those keys via this path. (2) Independently, `obj['__proto__'] = 'a string'` is a spec-mandated no-op (the accessor setter only accepts object/null), and `Object.fromEntries`/plain assignment with a `__proto__` key creates a harmless *own* property via CreateDataProperty semantics, not prototype mutation — confirmed with direct `node -e` reproduction, `Object.prototype` was unaffected in both cases. CSV cell values here are always strings (no `raw:true`), so the actual exploit precondition (assigning an object value) never occurs.

That said, added defense-in-depth in 7b0ca42c4f16d2f91d549aa48632991a23d29ba4 anyway so this doesn't depend on downstream readers re-deriving the spec semantics: the structured sanitizeColumnNames remap (line ~1111) now builds each row with `Object.create(null)` instead of `{}`. Also added a regression test ("CSVLoader (#1199): a header literally named __proto__ can't pollute Object.prototype in json output") that loads a CSV with a `__proto__` header and asserts `Object.prototype.polluted === undefined`.

src/lib/utils/csvProcessor.ts
murdore: Investigated in depth; concluded this specific report is a false positive (verified experimentally in Node, see the full writeup in the triage): (1) csv-parser@3.2.1 (the pinned version) already has its own `DANGEROUS_KEYS` header sanitizer that maps `__proto__`/`constructor`/`prototype` headers to `null` and drops the column before a row object is ever built, so `filteredRows`/`origHeaders` can't literally contain those keys via this path. (2) Independently, `obj['__proto__'] = 'a string'` is a spec-mandated no-op (the accessor setter only accepts object/null), and `Object.fromEntries`/plain assignment with a `__proto__` key creates a harmless *own* property via CreateDataProperty semantics, not prototype mutation — confirmed with direct `node -e` reproduction, `Object.prototype` was unaffected in both cases. CSV cell values here are always strings (no `raw:true`), so the actual exploit precondition (assigning an object value) never occurs.

That said, added defense-in-depth in 7b0ca42c4f16d2f91d549aa48632991a23d29ba4 anyway so this doesn't depend on downstream readers re-deriving the spec semantics: the structured sanitizeColumnNames remap (line ~1111) now builds each row with `Object.create(null)` instead of `{}`. Also added a regression test ("CSVLoader (#1199): a header literally named __proto__ can't pollute Object.prototype in json output") that loads a CSV with a `__proto__` header and asserts `Object.prototype.polluted === undefined`.

src/lib/utils/csvProcessor.ts
murdore: Investigated in depth; concluded this specific report is a false positive (verified experimentally in Node, see the full writeup in the triage): (1) csv-parser@3.2.1 (the pinned version) already has its own `DANGEROUS_KEYS` header sanitizer that maps `__proto__`/`constructor`/`prototype` headers to `null` and drops the column before a row object is ever built, so `filteredRows`/`origHeaders` can't literally contain those keys via this path. (2) Independently, `obj['__proto__'] = 'a string'` is a spec-mandated no-op (the accessor setter only accepts object/null), and `Object.fromEntries`/plain assignment with a `__proto__` key creates a harmless *own* property via CreateDataProperty semantics, not prototype mutation — confirmed with direct `node -e` reproduction, `Object.prototype` was unaffected in both cases. CSV cell values here are always strings (no `raw:true`), so the actual exploit precondition (assigning an object value) never occurs.

That said, added defense-in-depth in 7b0ca42c4f16d2f91d549aa48632991a23d29ba4 anyway so this doesn't depend on downstream readers re-deriving the spec semantics: the structured sanitizeColumnNames remap (line ~1111) now builds each row with `Object.create(null)` instead of `{}`. Also added a regression test ("CSVLoader (#1199): a header literally named __proto__ can't pollute Object.prototype in json output") that loads a CSV with a `__proto__` header and asserts `Object.prototype.polluted === undefined`.

src/lib/utils/csvProcessor.ts
murdore: Investigated in depth; concluded this specific report is a false positive (verified experimentally in Node, see the full writeup in the triage): (1) csv-parser@3.2.1 (the pinned version) already has its own `DANGEROUS_KEYS` header sanitizer that maps `__proto__`/`constructor`/`prototype` headers to `null` and drops the column before a row object is ever built, so `filteredRows`/`origHeaders` can't literally contain those keys via this path. (2) Independently, `obj['__proto__'] = 'a string'` is a spec-mandated no-op (the accessor setter only accepts object/null), and `Object.fromEntries`/plain assignment with a `__proto__` key creates a harmless *own* property via CreateDataProperty semantics, not prototype mutation — confirmed with direct `node -e` reproduction, `Object.prototype` was unaffected in both cases. CSV cell values here are always strings (no `raw:true`), so the actual exploit precondition (assigning an object value) never occurs.

That said, added defense-in-depth in 7b0ca42c4f16d2f91d549aa48632991a23d29ba4 anyway so this doesn't depend on downstream readers re-deriving the spec semantics: the structured sanitizeColumnNames remap (line ~1111) now builds each row with `Object.create(null)` instead of `{}`. Also added a regression test ("CSVLoader (#1199): a header literally named __proto__ can't pollute Object.prototype in json output") that loads a CSV with a `__proto__` header and asserts `Object.prototype.polluted === undefined`.

src/lib/rag/document/loaders.ts
murdore: Same finding as the csvProcessor.ts prototype-pollution threads — investigated and confirmed not exploitable as described (see the parallel reply on those threads for the full Node-level verification: `Object.fromEntries([['__proto__', 'x']])` and plain-object `__proto__` assignment both only ever create a harmless own property, never mutate `Object.prototype`). Added defense-in-depth in 7b0ca42c4f16d2f91d549aa48632991a23d29ba4 anyway: `CSVLoader.load`'s JSON-output branch in src/lib/rag/document/loaders.ts now builds each row with `Object.create(null)` instead of `Object.fromEntries`, so it doesn't rely on the spec nuance holding forever. Covered by the new regression test "CSVLoader (#1199): a header literally named __proto__ can't pollute Object.prototype in json output", which asserts both that `__proto__` becomes a normal own property on the row and that `Object.prototype.polluted` stays `undefined`.

murdore: Same finding as the csvProcessor.ts prototype-pollution threads — investigated and confirmed not exploitable as described (see the parallel reply on those threads for the full Node-level verification: `Object.fromEntries([['__proto__', 'x']])` and plain-object `__proto__` assignment both only ever create a harmless own property, never mutate `Object.prototype`). Added defense-in-depth in 7b0ca42c4f16d2f91d549aa48632991a23d29ba4 anyway: `CSVLoader.load`'s JSON-output branch in src/lib/rag/document/loaders.ts now builds each row with `Object.create(null)` instead of `Object.fromEntries`, so it doesn't rely on the spec nuance holding forever. Covered by the new regression test "CSVLoader (#1199): a header literally named __proto__ can't pollute Object.prototype in json output", which asserts both that `__proto__` becomes a normal own property on the row and that `Object.prototype.polluted` stays `undefined`.

src/lib/utils/csvProcessor.ts
murdore: Verified against current head (7b0ca42c4f16d2f91d549aa48632991a23d29ba4): `dedupeColumnNames` tracks a single `used: Set<string>` across all emitted names and loops `while (used.has(candidate))`, incrementing the numeric suffix until a truly unique name is found, before adding it to `used`. For `["name", "name", "name_2"]` this now correctly yields `["name", "name_2", "name_3"]` — the suffix-collision this thread flagged no longer reproduces. This matches the suggested fix and was already fixed prior to this pass; no further change needed.

src/lib/types/file.ts
murdore: Fixed in 7b0ca42c4f16d2f91d549aa48632991a23d29ba4. Renamed `CsvRow` to `CSVRow` in `src/lib/types/file.ts`, matching the `CSV*` uppercase convention used by `CSVColumnDataType`, `CSVColumnMetadata`, `CSVProcessorOptions`, etc., and updated all import/usage sites accordingly.

src/cli/factories/commandFactory.ts
murdore: Fixed in 7b0ca42c4f16d2f91d549aa48632991a23d29ba4. Extracted a private static `CLICommandFactory.buildCsvOptionsFromArgv(argv)` helper that builds the `CSVProcessorOptions` object from CLI argv, and both `executeGenerate` and `executeRealStream` now call it instead of each independently constructing the identical object literal.

src/lib/utils/csvProcessor.ts
```diff
@@ -88,6 +112,172 @@ function validateColumnName(name: string): string[] {
...
+    parts.push(`| bytesRead: ${info.bytesRead}`);
+  }
+  if (info.lastRow !== undefined) {
+    let snippet: string;
+    try {
+      snippet = JSON.stringify(info.lastRow).slice(0, 200);
+    } catch {
+      snippet = String(info.lastRow).slice(0, 200);
+    }
+    parts.push(`| lastRow: ${snippet}`);
```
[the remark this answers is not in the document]
murdore: Fixed in 7b0ca42c4f16d2f91d549aa48632991a23d29ba4. `buildCsvParseErrorMessage` no longer JSON-stringifies `info.lastRow` (the actual last-parsed CSV row) into the thrown error's message — it now reports only structural context (`lastRow columns: <n>` instead of the row's cell values). Updated `docs/features/csv-support.md`'s error-handling example to match, with an explicit note that only structural context is included, never actual cell values. Added a regression test ("CSVProcessor #1199: parse-error messages carry structural context only, never raw row cell values (PII)") that plants a secret-marker string in a row alongside an oversized field and asserts the thrown `NeuroLinkError` (EXECUTION category) message does not contain the marker or the string `"lastRow:"`.

src/lib/utils/csvProcessor.ts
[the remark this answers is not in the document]
Tara-ag: Agreed with the re-check: `fs.promises.access()` at line 1272 is still unbounded. A stalled `access()` can hang the whole call before `remaining()` is ever computed. Please wrap it (or the full access→prepare→parse sequence) with `withTimeout` and preserve the existing partial-result timeout contract.

Tara-ag (approved):
## Review summary

Reviewed the 17 changed files in this CSV-hardening PR against the project standards in `CLAUDE.md` and the configured focus areas.

None. Existing review threads already covered the substantive concerns (prototype-pollution defense-in-depth, `rawSource` cleanup, timeout coverage, quote-aware raw-format metadata, `dedupeColumnNames` collisions, `ErrorFactory` typed errors, PII leakage in error messages, CLI `csvOptions` duplication). The author has addressed the majority of those in the latest commit.

### Affirmed existing concern

I replied to the existing `parseTimeoutMs` thread confirming CodeRabbit's re-check: `fs.promises.access()` in `parseCSVFileWithMeta` is still awaited directly and is not bounded by `withTimeout`. A stalled filesystem access can hang the call before the `remaining()` timeout budget is ever consulted, so that path should still be wrapped before merge.

### Standards check

- No new static provider imports or Gemini tools/schema conflicts.
- Public SDK API changes widen `csvOptions` to the canonical `CSVProcessorOptions` — backward compatible.
- New types (`CSVRow`, `DecodedBuffer`) use `type` (not `interface`) and have unique names.
- Internal types are imported from the barrel where required.

**Decision: APPROVE** — based on new findings only, there are no new blocking issues. The pre-existing `fs.promises.access()` timeout concern noted above should still be resolved.

src/lib/utils/csvProcessor.ts
```diff
@@ -88,6 +114,171 @@ function validateColumnName(name: string): string[] {
...
+  }
+  parts.push(`: ${info.cause}`);
+  parts.push(`| columns: ${cols}`);
+  if (typeof info.bytesRead === "number") {
+    parts.push(`| bytesRead: ${info.bytesRead}`);
+  }
+  if (typeof info.lastRowColumnCount === "number") {
+    parts.push(`| lastRow columns: ${info.lastRowColumnCount}`);
+  }
+  parts.push("| Expected RFC 4180 CSV (comma-separated, double-quote escaped)");
```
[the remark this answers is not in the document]
murdore: Fixed in 3b67feb4. `buildCsvParseErrorMessage()` no longer joins header text into the error message at all — the `headers?: string[]` param was replaced with `headerCount?: number`, and the message now reports `headerCount: N` (or "unknown" when the parse failed before headers were read) instead of `columns: <joined header text>`. Both call sites (source-stream error, parser error) now pass `headerCount: capturedHeaders?.length`. This also covers the headerless-input case you flagged, since the count is derived structurally and never touches the actual header/cell strings. Docstring above the function updated to state this explicitly, and the doc example at csv-support.md was updated to show `headerCount: 3` instead of `columns: a, b, c`. Also strengthened the existing PII regression test (test/continuous-test-suite-bugfixes.ts, "CSVProcessor #1199: parse-error messages carry structural context only...") to put a secret marker in the header itself (not just the data row), so it would catch a regression of this exact kind.

docs/features/csv-support.md
```diff
@@ -216,6 +220,62 @@ csvOptions: {
 }
 ```
 
+#### encoding
+
+Character-encoding override (#362). When omitted, the encoding is detected from
+a BOM, then `chardet`, falling back to UTF-8 — so Windows-1252 / Latin-1 /
+UTF-16 files no longer decode as mojibake. Accepts any label
+[`iconv-lite`](https://www.npmjs.com/package/iconv-lite) supports.
```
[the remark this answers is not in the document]
murdore: Fixed in 3b67feb4. Added a "Streaming detection limitation" note under the encoding section of docs/features/csv-support.md: encoding is detected from the initial ~64 KiB only and then committed for the rest of the stream (kept this way intentionally so the parse-timeout guard can keep working against a single streaming pass — that internal rationale is already documented inline at csvProcessor.ts prepareFileSource, #1282). The new doc note explains an ASCII-only prefix followed by legacy-encoded content can still be misdecoded as UTF-8, and recommends passing `encoding` explicitly for such files.

docs/features/csv-support.md
```diff
@@ -216,6 +220,62 @@ csvOptions: {
...
+
+| Original     | Sanitized (snake_case) |
+| ------------ | ---------------------- |
+| `Price ($)`  | `price`                |
+| `1st Place`  | `col_1st_place`        |
+| `Name/Title` | `name_title`           |
+
+Original names are preserved: `result.metadata.columnNameMapping` lists every
+`{ original, sanitized }` pair and each renamed `columnMetadata` entry carries
+`originalName`. CLI: `--csv-sanitize-names [--csv-name-case camelCase]`. The
```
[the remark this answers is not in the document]
murdore: Fixed in 3b67feb4. Reworded the columnNameMapping sentence in docs/features/csv-support.md: it now says the mapping "lists each renamed `{ original, sanitized }` pair (columns left unchanged by sanitization are omitted)" instead of claiming it lists every pair, matching what `buildColumnNameMapping()` actually does (it `.filter((m) => m.original !== m.sanitized)`s before returning).

test/continuous-test-suite-bugfixes.ts
```diff
@@ -207,6 +220,540 @@ const tests: TestFunction[] = [
...
+        deduped[0] !== "name"
+      ) {
+        return false;
+      }
+      // A pathological underscore/separator run must sanitize in linear time
+      // (the previous `^_+|_+$` regex was polynomial — CodeQL ReDoS).
+      const t0 = Date.now();
+      const sanitized = sanitizeColumnName("_".repeat(50000) + "a");
+      if (Date.now() - t0 > 200 || sanitized !== "a") {
+        return false;
```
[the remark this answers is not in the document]
murdore: Fixed in 3b67feb4, exactly per your suggested regression input. Changed the test to `sanitizeColumnName("a" + "_".repeat(50_000) + "b")`, expecting `"a_b"`, so the 50k-underscore run sits between two non-underscore characters and can no longer be swallowed by a leading `^_+`-style alternation before reaching the trailing `_+$` search — it now actually exercises the polynomial worst-case branch the ReDoS fix targets. Timing assertion (200ms budget) kept as-is.

Tara-ag:
## Review summary

Reviewed the 17 changed files in this CSV-hardening PR. I was unable to open a new pending review because one already exists on this PR, so this comment serves as my review record.

### New issues raised in this run

None. All significant concerns I identified are already covered by existing review threads.

### Assessment

- **Security:** no new hardcoded secrets, injection vectors, or PII-leak paths. The prototype-pollution defense-in-depth (`Object.create(null)` in the structured remap and RAG JSON output) and the removal of raw row values from parse-error messages are correctly implemented.
- **Architecture:** the canonical `CSVProcessorOptions` shape is now referenced from `GenerateOptions`/`TextGenerationOptions`/`StreamOptions`, eliminating the 3-way duplication. CLI flag plumbing is consolidated in `buildCsvOptionsFromArgv`.
- **Correctness:** single-read streaming parse, `rawSource` teardown on abort, quote-aware raw-format metadata, collision-safe dedupe, typed `ErrorFactory` errors, and bounded parse timeouts are all present in the current head.
- **Tests:** the new regression tests are deterministic and cover the stated issues.

### Decision

**Approve** — no new blocking issues from this review pass.

Issue #349, opened 2025-12-01
PDF-019: requiresCitations Never Used


murdore:
## Summary
`requiresCitations` field defined in PDFProviderConfig but never read or used anywhere in the codebase.

## Root Cause
Line 23 defines `requiresCitations: "auto"` for Bedrock configuration, but this field is never referenced in any code. It appears to be intended for future citation extraction but was never implemented.

**Location:** `src/lib/utils/pdfProcessor.ts` line 23

## Impact
- Dead code cluttering config
- Unclear intent for future maintainers
- Potential confusion about functionality
- Missing feature or incomplete implementation

## Fix Instructions
**Option 1 (Recommended):**
Add TODO comment explaining intended usage:
```typescript
requiresCitations: "auto", // TODO: Implement citation extraction for Bedrock
// Bedrock Converse API requires citations for visual PDF mode
// See: https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-claude.html
```

**Option 2:**
Remove field if not needed soon

## Acceptance Criteria
- [ ] TODO comment added with explanation OR field removed
- [ ] Intended usage documented
- [ ] Decision documented in issue
- [ ] Tests pass
- [ ] No regressions

## Test Coverage Required
- Config still parses correctly
- No references to removed field (if removed)

**Note:** Good first issue - Simple complexity, refactoring task

murdore:
Closing — verified implemented in `release` (v10.10.6, `0b1abf1d`).

**Where it landed:** see evidence below

**Evidence:**

[...]

Issue #317, opened 2025-12-01
PDF-013: No Pre-flight Size Check for URLs


murdore:
## Summary
When PDFs are loaded from URLs, no pre-flight HEAD request checks the file size before downloading. Large PDFs are downloaded entirely into memory before size validation, causing OOM crashes.

## Root Cause
PDF URL handling downloads the entire file before validating size limits. No Content-Length check before download initiation.

**Location:** PDF URL handling in message builder and processor
**Priority:** High
**Effort:** 4h
**Complexity:** Moderate

## Impact
- Large PDFs downloaded entirely into memory before rejection
- OOM crashes on oversized remote PDFs
- Wasted bandwidth and time
- No early warning for oversized files
- Server resource exhaustion

## Fix Instructions
1. Add HEAD request before downloading PDF from URL
2. Check Content-Length header against size limits
3. Throw early error if file too large
4. Add timeout for HEAD requests
5. Fallback to streaming download with size monitoring if no Content-Length
6. Log file size before download

## Acceptance Criteria
- [ ] HEAD request checks size before download
- [ ] Early error if Content-Length exceeds limit
- [ ] Timeout for HEAD requests (5 seconds)
- [ ] Streaming fallback if Content-Length unavailable
- [ ] Size logged before download
- [ ] Tests verify pre-flight checks
- [ ] No regressions

## Test Coverage Required
- URL with valid Content-Length under limit succeeds
- URL with Content-Length over limit fails early
- URL without Content-Length uses streaming
- HEAD request timeout handled gracefully
- Error messages include expected vs actual size

murdore:
### ◐ Audit status: partially done — keeping open

Core behaviour ships, but not every acceptance criterion is met. The remaining gap is the actionable part:

> The core OOM risk is mitigated by streaming + abort-on-exceed, but the specific pre-flight HEAD/Content-Length check, its 5s timeout, and the 'expected vs actual' error messaging from the acceptance criteria are not implemented.

<details><summary>what already works (evidence)</summary>

[...]

Issue #309, opened 2025-12-01
PDF-012: Multiple PDF Aggregate Bypass


murdore:
## Summary
When multiple PDFs are sent in a single request, size and page limits are checked individually per PDF but not in aggregate, allowing bypass of intended limits.

## Root Cause
PDF validation checks each PDF independently without tracking cumulative totals across all PDFs in the request. Users can bypass limits by sending many small PDFs instead of one large PDF.

**Location:** `src/lib/utils/pdfProcessor.ts` and message builder
**Priority:** High
**Effort:** 4h
**Complexity:** Moderate

## Impact
- Users can bypass page limits (e.g., 100 one-page PDFs vs 1 hundred-page PDF)
- Memory limits can be exceeded in aggregate
- API token limits circumvented
- Unexpected costs and performance issues
- Inconsistent limit enforcement

## Fix Instructions
1. Add aggregate tracking for multiple PDFs
2. Sum total pages across all PDFs
3. Sum total size across all PDFs
4. Validate aggregates against limits
5. Throw error if aggregates exceed limits
6. Log warnings when multiple PDFs approach limits

## Acceptance Criteria
- [ ] Aggregate page count tracked across all PDFs
- [ ] Aggregate size tracked across all PDFs
- [ ] Validation enforces aggregate limits
- [ ] Clear error messages for aggregate violations
- [ ] Tests verify multi-PDF scenarios
- [ ] No regressions

## Test Coverage Required
- Single PDF under limit succeeds
- Multiple PDFs under aggregate limit succeed
- Multiple PDFs over aggregate page limit fails
- Multiple PDFs over aggregate size limit fails
- Error messages show aggregate vs individual limits

Harshita-Rupani29:
Assign it to me.

Issue #302, opened 2025-12-01
PDF-010: No Streaming Implementation


murdore:
## Summary
PDF pages are converted and loaded entirely into memory before returning results. No streaming capability for progressive processing of large multi-page PDFs.

## Root Cause
The `convertPDFToImages()` method collects all converted pages in an array before returning. For large PDFs, this requires holding all page buffers in memory simultaneously.

**Location:** `src/lib/utils/pdfProcessor.ts` (convertPDFToImages method)
**Priority:** High
**Effort:** 8h
**Complexity:** Complex

## Impact
- Memory spikes with large PDFs (all pages held simultaneously)
- No progressive feedback during long conversions
- Cannot process PDFs larger than available memory
- Poor user experience with long processing times
- Limits scalability

## Fix Instructions
1. Design async iterator/generator interface for page-by-page streaming
2. Implement `convertPDFToImagesStream()` method returning AsyncGenerator
3. Yield pages as they're converted instead of collecting all
4. Update consumers to handle streaming results
5. Add progress callbacks for UI feedback
6. Maintain backward compatibility with existing batch method

## Acceptance Criteria
- [ ] AsyncGenerator-based streaming method implemented
- [ ] Pages yielded progressively during conversion
- [ ] Memory usage stays constant regardless of page count
- [ ] Progress callbacks supported
- [ ] Backward compatibility maintained
- [ ] Tests verify streaming behavior
- [ ] Documentation updated

## Test Coverage Required
- Streaming yields pages in order
- Memory usage constant during stream
- Early termination releases resources
- Progress callbacks invoked correctly
- Errors in middle of stream handled gracefully
- Backward compatible batch method still works

murdore:
Fixed by #1737 (`ae6ca8a3b`), merged to `release`.

PDF page images now stream on the path `generate()` actually takes, rather than only on the one the old test exercised.

PR #1201, opened 2026-07-17
fix(pdf): harden image conversion — accurate pages, per-page resilience, scale, streaming, aggregate & URL limits

### LINKED ISSUE #287: PDF-007: Unreliable Page Detection Regex
## Summary
Regex-based page detection (`/\/Type\s*\/Page[^s]/g`) fails with compressed PDFs, non-standard formatting, or when pages appear beyond first 10KB.

## Root Cause
Line 219 uses regex pattern matching on first 10KB of PDF content to estimate pages. Compressed PDFs (FlateDecode) have binary content that regex can't parse. Large PDFs may have page definitions beyond 10KB.

**Location:** `src/lib/utils/pdfProcessor.ts` lines 212-220

## Impact
- Inaccurate page counts
- False positives/negatives
- Completely fails on compressed PDFs
- Unreliable limit enforcement

## Fix Instructions
1. Add new `getAccuratePageCount()` method using pdfjs
2. Set 5-second timeout for page count operation
3. Handle password-protected PDFs gracefully
4. Return null if count can't be determined
5. Update `extractBasicMetadata()` to return null for estimatedPages
6. Replace regex-based counting with pdfjs parsing

## Acceptance Criteria
- [ ] getAccuratePageCount() method added using pdfjs
- [ ] 5-second timeout implemented
- [ ] Password-protected PDFs handled gracefully
- [ ] Returns null if count fails
- [ ] extractBasicMetadata() updated
- [ ] Tests verify accuracy on various PDF types
- [ ] No regressions

## Test Coverage Required
- Accurate count for standard PDFs
- Accurate count for compressed PDFs
- Timeout for very large PDFs
- Graceful handling of encrypted PDFs
- Null return on failure

murdore:
> **Stacked on #1195** (base = `fix/pdf-page-memory-and-password`). These items build on that PR's `convertToImages` password/canvas work. Retarget to `release` once #1195 merges.

## Summary

Consolidated hardening of the PDF **image-fallback** path (providers without native PDF support).

| Issue | Fix |
| ----- | --- |
| **#287** PDF-007 — unreliable page regex | `getAccuratePageCount()` parses via pdf-parse/pdfjs (time-bounded by `PAGE_COUNT_TIMEOUT_MS`); `process()` prefers it for limit enforcement + reported `estimatedPages`, falling back to the regex on timeout/failure. |
| **#294** PDF-008 — all-or-nothing rendering | Indexed `getPage()` loop with per-page `try/catch`: a failing page is recorded in `result.errors` (`{ page, error }`) instead of discarding all converted pages. |
| **#297** PDF-009 — high memory default | Default scale `2 → 1.5`; full `MIN_SCALE(0.1)…MAX_SCALE(10)` range enforced; memory estimate logged before conversion. |
| **#302** PDF-010 — no streaming | New `convertToImagesStream()` async generator yields each page as it renders with an `onProgress` callback; `convertToImages` shares its validation. |
| **#309** PDF-012 — aggregate bypass | `processExplicitPdfFiles` checks **combined** page count + size across all PDFs against the provider ceiling. |
| **#317** PDF-013 — no URL pre-flight | `loadFromURL` issues a `HEAD` and rejects an oversized `Content-Length` before downloading; missing header / HEAD-refusing server falls through to the streaming guard. |
| **#349** PDF-019 — dead `requiresCitations` | Surfaced on `metadata.requiresCitations` (+ documented) so provider adapters can act on it. |

## Verification

- **bugfixes suite: 110 passed**, only the 2 known-environmental `updater:` tests fail locally (pass in CI).
- **8 new deterministic tests** (+ new fixture `test/fixtures/pdf-page3-corrupt.pdf`, a real 3-page PDF with page 3's stream corrupted so only that page fails): accurate/graceful page count; per-page isolation (`images.length===2`, `errors=[{page:3}]`, no throw); scale 1.5 default + memory-estimate log; streaming order + progress + early-termination; aggregate 120-page rejection; URL pre-flight (oversized rejected before GET via a local `http` server); citations surfaced. The existing scale-validation test is extended to prove the new `MIN_SCALE`.

## Docs
`docs/features/pdf-support.md`: **Conversion Resilience & Limits** and **Streaming Conversion** sections.

Fixes #287, #294, #297, #302, #309, #317, #349.

### DIFF
--- a/docs/features/pdf-support.md
+++ b/docs/features/pdf-support.md
@@ -216,6 +216,51 @@ pdfOptions: {
 }
 ```
 
+### Conversion Resilience & Limits
+
+The image-fallback path (providers without native PDF support) is hardened:
+
+- **Accurate page counts (#287).** Page-limit enforcement and the reported
+  `metadata.estimatedPages` use a real pdfjs parse (time-bounded), not a header
+  regex that miscounts compressed/object-stream PDFs. It degrades to the regex
+  estimate on timeout/failure.
+- **Per-page isolation (#294).** A single page that fails to render no longer
+  discards the whole conversion — the successful pages are returned and each
+  failure is reported in `result.errors` as `{ page, error }`.
+- **Default render scale 1.5 (#297).** Lowered from 2 to roughly halve per-page
+  canvas memory; the render scale is validated to the `0.1`–`10` range, and an
+  estimated-memory line is logged before conversion.
+- **Aggregate multi-PDF limits (#309).** When several PDFs are attached, their
+  combined page count and size are checked against the provider's ceiling — N
+  files each under the single-file limit can no longer exceed it together.
+- **URL pre-flight (#317).** A remote PDF's `Content-Length` is checked with a
+  `HEAD` request before any body is downloaded, so an oversized URL is rejected
+  up front (falling back to the streaming byte guard when the header is absent).
+
+### Streaming Conversion
+
+For large documents, `PDFProcessor.convertToImagesStream()` yields each page's
+image as soon as it renders (with an optional `onProgress` callback) instead of
+buffering the whole document (#302):
+
+```typescript
+import { PDFProcessor } from "@juspay/neurolink";
+
+for await (const page of PDFProcessor.convertToImagesStream(pdfBuffer, {
+  onProgress: ({ pagesConverted, totalPages, elapsedMs }) =>
+    console.log(`${pagesConverted}/${totalPages} in ${elapsedMs}ms`),
+})) {
+  if (page.error) {
+    console.warn(`page ${page.pageIndex} failed: ${page.error}`);
+  } else {
+    handlePng(page.image); // base64 PNG
+  }
+}
+```
+
+`convertToImages()` is the batch wrapper over this stream and returns the same
+per-page `errors` contract.
+
 ## Provider Support
 
 ### Supported Providers
--- a/src/lib/utils/fileDetector.ts
+++ b/src/lib/utils/fileDetector.ts
@@ -1890,6 +1890,49 @@ export class FileDetector {
     const maxRetries = options?.maxRetries ?? DEFAULT_MAX_RETRIES;
     const retryDelay = options?.retryDelay ?? DEFAULT_RETRY_DELAY;
 
+    // #317: pre-flight HEAD to reject an oversized file BEFORE downloading any
+    // body. content-length is advisory (chunked responses omit it), so a
+    // missing/invalid header — or a server that refuses HEAD — falls through to
+    // the streaming byte guard below; only a genuine oversize rejection stops
+    // the GET from ever running.
+    //
+    // #323: skip the pre-flight entirely when this exact URL was recently seen
+    // (its Content-Type is still cached, whether from a prior loadFromURL GET
+    // or a MimeTypeStrategy HEAD) — issuing a fresh HEAD here would defeat the
+    // whole point of that cache. The streaming byte guard in the GET below
+    // still enforces maxSize even without a pre-flight, so this doesn't remove
+    // the oversize protection — it only skips the redundant round-trip for a
+    // URL we've already been talking to within the last 60s.
+    if (getCachedUrlContentType(url, Date.now()) === undefined) {
+      try {
+        const head = await request(url, {
+          dispatcher: getGlobalDispatcher().compose(
+            interceptors.redirect({ maxRedirections: 5 }),
+          ),
+          method: "HEAD",
+          headersTimeout: FileDetector.DEFAULT_HEAD_TIMEOUT,
+          bodyTimeout: FileDetector.DEFAULT_HEAD_TIMEOUT,
+        });
+        // Drain/close the (empty) HEAD body so the connection can be reused.
+        await head.body.dump();
+        const declaredLength = Number(head.headers["content-length"]);
+        if (Number.isFinite(declaredLength) && declaredLength > maxSize) {
+          throw new Error(
+            `File too large: ${formatFileSize(declaredLength)} (max: ${formatFileSize(maxSize)})`,
+          );
+        }
+      } catch (error) {
+        if (error instanceof Error && /File too large/.test(error.message)) {
+          throw error;
+        }
+        logger.debug(
+          `[FileDetector] HEAD pre-flight skipped for ${url}: ${
+            error instanceof Error ? error.message : String(error)
+          }`,
+        );
+      }
+    }
+
     return withRetry(
       async () => {
         try {
--- a/src/lib/utils/messageBuilder.ts
+++ b/src/lib/utils/messageBuilder.ts
@@ -1321,6 +1321,29 @@ async function processExplicitPdfFiles(
     }
   }
 
+  // #309: enforce the provider's page/size ceilings across ALL PDFs, not just
+  // per-file. N files each just under the single-file limit can still blow past
+  // it in aggregate (e.g. three 40-page PDFs → 120 pages for a 100-page API).
+  const aggregateConfig = PDFProcessor.getProviderConfig(provider);
+  if (aggregateConfig && pdfFiles.length > 1) {
+    const totalPages = pdfFiles.reduce((sum, f) => sum + (f.pageCount ?? 0), 0);
+    const totalMB =
+      pdfFiles.reduce((sum, f) => sum + f.buffer.length, 0) / (1024 * 1024);
+    if (totalPages > aggregateConfig.maxPages) {
+      throw new Error(
+        `[PDF] Combined page count across ${pdfFiles.length} PDFs (${totalPages}) exceeds the ` +
+          `${aggregateConfig.maxPages}-page limit for ${provider}. ` +
+          `Split the request or reduce the number of PDFs.`,
+      );
+    }
+    if (totalMB > aggregateConfig.maxSizeMB) {
+      throw new Error(
+        `[PDF] Combined size across ${pdfFiles.length} PDFs (${totalMB.toFixed(2)}MB) exceeds the ` +
+          `${aggregateConfig.maxSizeMB}MB limit for ${provider}.`,
+      );
+    }
+  }
+
   return pdfFiles;
 }
 
--- a/src/lib/utils/pdfProcessor.ts
+++ b/src/lib/utils/pdfProcessor.ts
@@ -15,6 +15,7 @@ import type {
   PDFProviderConfig,
   PDFImageConversionOptions,
   PDFImageConversionResult,
+  PDFImagePage,
 } from "../types/index.js";
 import { ErrorFactory } from "./errorHandling.js";
 import { logger } from "./logger.js";
@@ -177,6 +178,14 @@ export class PDFProcessor {
 
     const metadata = PDFProcessor.extractBasicMetadata(content);
 
+    // #287: prefer an accurate page count (pdf-parse/pdfjs) over the unreliable
+    // header regex for both limit enforcement and returned metadata; fall back
+    // to the regex estimate when the accurate probe times out or fails.
+    const accuratePages = await PDFProcessor.getAccuratePageCount(content);
+    if (accuratePages !== null) {
+      metadata.estimatedPages = accuratePages;
+    }
+
     if (metadata.estimatedPages && metadata.estimatedPages > config.maxPages) {
       const enforceLimits = options?.enforceLimits !== false;
 
@@ -223,6 +232,10 @@ export class PDFProcessor {
         ...metadata,
         provider,
         apiType: config.apiType,
+        // #349: surface the provider's citations requirement so downstream
+        // provider adapters (e.g. Bedrock Converse document blocks) can act on
+        // it, instead of the config field being read nowhere.
+        requiresCitations: config.requiresCitations,
       },
     };
   }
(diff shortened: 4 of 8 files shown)


src/lib/utils/pdfProcessor.ts
```diff
@@ -266,6 +283,47 @@ export class PDFProcessor {
...
+      try {
+        const info = await Promise.race([
+          pdf.getInfo(),
+          new Promise<never>((_, reject) =>
+            setTimeout(
+              () => reject(new Error("page-count timeout")),
+              PDF_LIMITS.PAGE_COUNT_TIMEOUT_MS,
+            ),
+          ),
+        ]);
```
Tara-ag, line 309: 🔒 CRITICAL: `getAccuratePageCount()` starts a `setTimeout` inside `Promise.race` but never clears it when `pdf.getInfo()` wins. The dangling timer can keep the process alive and does unnecessary work.

Suggested fix:
```typescript
const timeout = setTimeout(
  () => reject(new Error("page-count timeout")),
  PDF_LIMITS.PAGE_COUNT_TIMEOUT_MS,
);
try {
  const info = await Promise.race([pdf.getInfo(), timeoutPromise]);
  // ...
  return total;
} finally {
  clearTimeout(timeout);
  await (pdf as unknown as { destroy?: () => Promise<void> | void }).destroy?.();
}
```
Also consider `timeout.unref()` so it never delays process exit even if the race is still pending when the process wants to shut down.

src/lib/utils/pdfProcessor.ts
```diff
@@ -266,6 +283,47 @@ export class PDFProcessor {
...
+        const total = (info as { total?: number }).total;
+        return typeof total === "number" && total > 0 ? total : null;
+      } finally {
+        await (
+          pdf as unknown as { destroy?: () => Promise<void> | void }
+        ).destroy?.();
+      }
```
Tara-ag, line 316: ⚠️ MAJOR: If `pdf.destroy()` throws, the outer `try` in `getAccuratePageCount()` treats the whole accurate page count as failed and returns `null`, even when `getInfo()` succeeded. Cleanup failures should not discard a valid page count.

Suggested fix (matching the pattern already used in `fileReferenceRegistry.ts`):
```typescript
} finally {
  await (
    pdf as unknown as { destroy?: () => Promise<void> | void }
  ).destroy?.().catch(() => {
    /* cleanup - ignore destroy errors */
  });
}
```

src/lib/utils/messageBuilder.ts
```diff
@@ -1321,6 +1321,29 @@ async function processExplicitPdfFiles(
...
   }
 
+  // #309: enforce the provider's page/size ceilings across ALL PDFs, not just
+  // per-file. N files each just under the single-file limit can still blow past
+  // it in aggregate (e.g. three 40-page PDFs → 120 pages for a 100-page API).
+  const aggregateConfig = PDFProcessor.getProviderConfig(provider);
+  if (aggregateConfig && pdfFiles.length > 1) {
+    const totalPages = pdfFiles.reduce((sum, f) => sum + (f.pageCount ?? 0), 0);
+    const totalMB =
+      pdfFiles.reduce((sum, f) => sum + f.buffer.length, 0) / (1024 * 1024);
```
Tara-ag, line 1331: ⚠️ MAJOR: Aggregate page-limit enforcement can be bypassed when any PDF has `estimatedPages === null` (e.g. corrupted/odd PDFs), because `totalPages` sums `pageCount ?? 0`. For limit enforcement, treating unknown page counts as 0 undercounts and defeats the aggregate guard.

Suggested fix: when any PDF has an unknown page count, either (a) reject the aggregate check until all page counts are known, or (b) fall back to a conservative upper-bound estimate. For example:
```typescript
const unknownPageCount = pdfFiles.some((f) => f.pageCount === null || f.pageCount === undefined);
const totalPages = pdfFiles.reduce((sum, f) => sum + (f.pageCount ?? 0), 0);
if (!unknownPageCount && totalPages > aggregateConfig.maxPages) {
  throw new Error(...);
}
// Optionally warn/log when unknown so operators know the guard was partial.
```
At minimum, the aggregate guard should not silently pass when page counts are missing.

src/lib/utils/fileDetector.ts
```diff
@@ -1890,6 +1890,49 @@ export class FileDetector {
...
+            interceptors.redirect({ maxRedirections: 5 }),
+          ),
+          method: "HEAD",
+          headersTimeout: FileDetector.DEFAULT_HEAD_TIMEOUT,
+          bodyTimeout: FileDetector.DEFAULT_HEAD_TIMEOUT,
+        });
+        // Drain/close the (empty) HEAD body so the connection can be reused.
+        await head.body.dump();
+        const declaredLength = Number(head.headers["content-length"]);
+        if (Number.isFinite(declaredLength) && declaredLength > maxSize) {
```
Tara-ag, line 1919: ⚠️ MAJOR: The HEAD pre-flight doesn't check `head.statusCode`. If a server returns 405/4xx, this code still reads `content-length` and may reject as “too large” based on the error response size, even though the intent is to fall through to the GET streaming guard when HEAD is refused.

Suggested fix: only enforce the size when the HEAD succeeds (2xx). For example:
```typescript
if (head.statusCode >= 200 && head.statusCode < 300) {
  const declaredLength = Number(head.headers["content-length"]);
  if (Number.isFinite(declaredLength) && declaredLength > maxSize) {
    throw new Error(`File too large: ...`);
  }
}
```
Non-2xx HEAD responses should be treated the same as a missing header and fall through to the GET streaming guard.

src/lib/utils/pdfProcessor.ts
```diff
@@ -538,6 +599,131 @@ export class PDFProcessor {
...
+    return sizeMB;
+  }
+
+  /**
+   * Streaming variant of {@link convertToImages} (#302): yields each page's
+   * base64 PNG as soon as it renders instead of buffering the whole document,
+   * and reports progress via `options.onProgress`. A page that fails to render
+   * is yielded with `error` set (per-page isolation, #294) rather than aborting
+   * the stream. `convertToImages` is the batch wrapper over this contract.
+   */
```
Tara-ag, line 658: 💡 MINOR: The JSDoc says “`convertToImages` is the batch wrapper over this contract”, but `convertToImages()` doesn't call `convertToImagesStream()` (it has its own implementation). This is misleading for maintainers and readers. Suggest rewording to avoid implying an implementation relationship, e.g. “Batch variant of the same conversion; both share `validateImageConversionInput` and per-page isolation semantics.”

docs/features/pdf-support.md
```diff
@@ -216,6 +216,51 @@ pdfOptions: {
...
+  if (page.error) {
+    console.warn(`page ${page.pageIndex} failed: ${page.error}`);
+  } else {
+    handlePng(page.image); // base64 PNG
+  }
+}
+```
+
+`convertToImages()` is the batch wrapper over this stream and returns the same
+per-page `errors` contract.
```
Tara-ag, line 262: 💡 MINOR: Docs say `convertToImages()` is “the batch wrapper over this stream”, but the implementation isn't a wrapper over `convertToImagesStream()` (it’s a separate batch implementation). Suggest rewording to avoid implying an implementation relationship, e.g. “Batch variant with the same per-page `errors` contract.”

src/lib/utils/fileDetector.ts
```diff
@@ -1890,6 +1890,49 @@ export class FileDetector {
...
+    // #323: skip the pre-flight entirely when this exact URL was recently seen
+    // (its Content-Type is still cached, whether from a prior loadFromURL GET
+    // or a MimeTypeStrategy HEAD) — issuing a fresh HEAD here would defeat the
+    // whole point of that cache. The streaming byte guard in the GET below
+    // still enforces maxSize even without a pre-flight, so this doesn't remove
+    // the oversize protection — it only skips the redundant round-trip for a
+    // URL we've already been talking to within the last 60s.
+    if (getCachedUrlContentType(url, Date.now()) === undefined) {
+      try {
+        const head = await request(url, {
```
Tara-ag, line 1908: ⚠️ MAJOR: The HEAD pre-flight doesn't check `head.statusCode`. If a server returns 405/4xx, this code still reads `content-length` and may reject as “too large” based on the error response size, even though the intent is to fall through to the GET streaming guard when HEAD is refused.

Suggested fix: only enforce the size when the HEAD succeeds (2xx). For example:
```typescript
if (head.statusCode >= 200 && head.statusCode < 300) {
  const declaredLength = Number(head.headers["content-length"]);
  if (Number.isFinite(declaredLength) && declaredLength > maxSize) {
    throw new Error(`File too large: ...`);
  }
}
```
Non-2xx HEAD responses should be treated the same as a missing header and fall through to the GET streaming guard.

src/lib/utils/messageBuilder.ts
Tara-ag, line 1331: ⚠️ MAJOR: Aggregate page-limit enforcement can be bypassed when any PDF has `estimatedPages === null` (e.g. corrupted/odd PDFs), because `totalPages` sums `pageCount ?? 0`. For limit enforcement, treating unknown page counts as 0 undercounts and defeats the aggregate guard.

Suggested fix: when any PDF has an unknown page count, either (a) reject the aggregate check until all page counts are known, or (b) fall back to a conservative upper-bound estimate. For example:
```typescript
const unknownPageCount = pdfFiles.some((f) => f.pageCount === null || f.pageCount === undefined);
const totalPages = pdfFiles.reduce((sum, f) => sum + (f.pageCount ?? 0), 0);
if (!unknownPageCount && totalPages > aggregateConfig.maxPages) {
  throw new Error(...);
}
// Optionally warn/log when unknown so operators know the guard was partial.
```
At minimum, the aggregate guard should not silently pass when page counts are missing.

src/lib/utils/pdfProcessor.ts
Tara-ag, line 309: 🔒 CRITICAL: `getAccuratePageCount()` starts a `setTimeout` inside `Promise.race` but never clears it when `pdf.getInfo()` wins. The dangling timer can keep the process alive and does unnecessary work.

Suggested fix:
```typescript
const timeout = setTimeout(
  () => reject(new Error("page-count timeout")),
  PDF_LIMITS.PAGE_COUNT_TIMEOUT_MS,
);
try {
  const info = await Promise.race([pdf.getInfo(), timeoutPromise]);
  // ...
  return total;
} finally {
  clearTimeout(timeout);
  await (pdf as unknown as { destroy?: () => Promise<void> | void }).destroy?.();
}
```
Also consider `timeout.unref()` so it never delays process exit even if the race is still pending when the process wants to shut down.

src/lib/utils/pdfProcessor.ts
Tara-ag, line 316: ⚠️ MAJOR: If `pdf.destroy()` throws, the outer `try` in `getAccuratePageCount()` treats the whole accurate page count as failed and returns `null`, even when `getInfo()` succeeded. Cleanup failures should not discard a valid page count.

Suggested fix (matching the pattern already used in `fileReferenceRegistry.ts`):
```typescript
} finally {
  await (
    pdf as unknown as { destroy?: () => Promise<void> | void }
  ).destroy?.().catch(() => {
    /* cleanup - ignore destroy errors */
  });
}
```

src/lib/utils/pdfProcessor.ts
Tara-ag, line 658: 💡 MINOR: The JSDoc says “`convertToImages` is the batch wrapper over this contract”, but `convertToImages()` doesn't call `convertToImagesStream()` (it has its own implementation). This is misleading for maintainers and readers.

Suggest rewording to avoid implying an implementation relationship, e.g. “Batch variant of the same conversion; both share `validateImageConversionInput` and per-page isolation semantics.”

docs/features/pdf-support.md
Tara-ag, line 262: 💡 MINOR: Docs say `convertToImages()` is “the batch wrapper over this stream”, but the implementation isn't a wrapper over `convertToImagesStream()` (it’s a separate batch implementation). Suggest rewording to avoid implying an implementation relationship, e.g. “Batch variant with the same per-page `errors` contract.”

Tara-ag (changes requested):
## Review summary

Reviewed the 8 changed files in this PDF hardening PR. The changes are well-scoped and the test coverage is good, but there are several correctness issues that need to be addressed before merging.

### New issues raised

| Severity | Count | Files |
|----------|-------|-------|
| 🔒 CRITICAL | 1 | `src/lib/utils/pdfProcessor.ts` |
| ⚠️ MAJOR | 3 | `src/lib/utils/pdfProcessor.ts`, `src/lib/utils/messageBuilder.ts`, `src/lib/utils/fileDetector.ts` |
| 💡 MINOR | 2 | `src/lib/utils/pdfProcessor.ts`, `docs/features/pdf-support.md` |

### Blocking concerns

1. **Dangling timeout in `getAccuratePageCount()`** (`pdfProcessor.ts:309`) — the `setTimeout` is never cleared when `pdf.getInfo()` wins, which can keep the process alive and waste work.
2. **`pdf.destroy()` failure discards a valid page count** (`pdfProcessor.ts:316`) — cleanup errors should not null out an already-computed accurate count.
3. **Aggregate page guard silently passes when page counts are unknown** (`messageBuilder.ts:1331`) — summing `pageCount ?? 0` lets corrupted/odd PDFs bypass the aggregate limit.
4. **HEAD pre-flight doesn't check status code** (`fileDetector.ts:1908`) — a 405/4xx HEAD can still be read as a size rejection instead of falling through to the GET streaming guard.

These meet the ≥3 MAJOR threshold for requesting changes.

### Minor notes

- The JSDoc/docs wording implying `convertToImages()` wraps `convertToImagesStream()` is misleading because the implementations are separate; reword to clarify they share validation and per-page semantics.

### Next steps

Please address the four blocking items above (each comment includes a suggested fix). Once those are resolved and CI is green, this should be good to merge. The new tests and fixtures are a solid addition.

Issue #325, opened 2025-12-01
TD-008: Loose Index Signature in MessageContent Type


murdore:
## Summary
The `MessageContent` type uses overly permissive index signature (`[key: string]: unknown`), allowing typos and invalid properties to pass type checking.

## Root Cause
- Designed for compatibility with loose Vercel AI SDK types
- Index signature prioritized flexibility over safety
- No discriminated union to enforce specific shapes

## Security Risk Example
```typescript
const message: MessageContent = {
  type: 'image',
  image: 'base64...',
  mymeType: 'image/jpeg', // ❌ TYPO - TypeScript doesn't catch this!
};
// Provider API fails at runtime ❌
```

## Fix
Convert to discriminated union:
```typescript
export type MessageContent =
  | { type: "text"; text: string; }
  | { type: "image"; image: string; mimeType?: string; }
  | { type: "file"; data: string; mimeType?: string; filename?: string; }
  | { type: "tool-call"; toolCallId: string; toolName: string; args: Record<string, unknown>; }
  | { type: "tool-result"; toolCallId: string; toolName: string; result: unknown; isError?: boolean; };
```

## Pre-Fix Verification
```bash
# Search for MessageContent usage
grep -rn "MessageContent" src/lib --include="*.ts" | grep -v "types/"

# Check for object spreads
git grep -A 5 "MessageContent.*{" src/lib/ | grep -v "type:"
```

## Acceptance Criteria
- [ ] Discriminated union type
- [ ] No index signature
- [ ] Codebase compatible
- [ ] Type safety tests
- [ ] All tests pass

## Files to Modify
- `src/lib/types/multimodal.ts` (lines 232-238)
- Create `test/types/message-content-safety.test.ts`

murdore:
Closing — verified implemented at the current release tip.

`src/lib/types/multimodal.ts:500-527` defines `MessageContent` with concrete optional fields (text, image, mimeType, data, name, filename, toolCallId, toolName, args, result, isError, providerOptions) and **no** `[key: string]: unknown` index signature. Lines 490-498 carry a comment tagged with this issue number recording the replacement.

Note `ProviderMultimodalPayload` (line 599) does still have an index signature, but that is a different type and outside this issue's scope.

Issue #323, opened 2025-12-01
FD-007: Eliminate Double HTTP Request for URL Detection


murdore:
## Bug Summary
When detecting a URL file type, the system makes TWO HTTP requests: MimeTypeStrategy makes a HEAD request to get Content-Type, then loadFromURL() makes a GET request, doubling latency and bandwidth.

## Root Cause Analysis
**Why this bug exists:**
Detection pipeline processes URLs in two stages:
1. Detection stage: MimeTypeStrategy makes HEAD request for Content-Type
2. Loading stage: loadFromURL() makes GET request to download

These requests don't share data, causing:
- Duplicate network round-trips
- Wasted bandwidth
- Increased latency (2x)
- Unnecessary server load

**Impact:**
- 2x network latency
- Increased bandwidth usage
- Higher server costs
- Poor UX for large files

## Current Code (Buggy)
\`\`\`typescript
// MimeTypeStrategy (lines 378-385)
async detect(input: FileInput): Promise<FileDetectionResult> {
  // ❌ First HTTP request (HEAD)
  const response = await request(input, { method: "HEAD" });
  const contentType = response.headers["content-type"];
  // ...
}

// loadFromURL (lines 211-217)
private static async loadFromURL(url: string): Promise<Buffer> {
  // ❌ Second HTTP request (GET) - no cache
  const response = await request(url, { method: "GET" });
  // ...
}
\`\`\`

## Fixed Code
\`\`\`typescript
// Add caching
private static urlMetadataCache = new Map<string, {
  contentType: string;
  contentLength: number;
  timestamp: number;
}>();
private static CACHE_TTL = 60000;

// Cache HEAD response
async detect(input: FileInput): Promise<FileDetectionResult> {
  const response = await request(input, { method: "HEAD" });

  FileDetector.urlMetadataCache.set(input, {
    contentType: response.headers["content-type"],
    contentLength: parseInt(response.headers["content-length"] || "0"),
    timestamp: Date.now(),
  });
  // ...
}

// Use cache in GET
private static async loadFromURL(url: string): Promise<Buffer> {
  const cached = this.urlMetadataCache.get(url);
  if (cached && Date.now() - cached.timestamp < this.CACHE_TTL) {
    // Use cached metadata
  }
  // Single GET request
}
\`\`\`

## Acceptance Criteria
- [ ] Implement cache
- [ ] Share HEAD data
- [ ] 1-minute TTL
- [ ] Reduce to 1 request
- [ ] Tests pass

[...]

Issue #293, opened 2025-12-01
IMG-010: No Empty Image Handling


murdore:
**Summary:** No validation that buffers are non-empty before processing, creating invalid base64 and wrong MIME detection.

**Root Cause:** No empty buffer check in `process()` at lines 22-28 in `src/lib/utils/imageProcessor.ts`. Empty buffer (0 bytes) converts to empty base64 causing API rejections.

**Impact:** Confusing provider errors, invalid data sent to APIs, poor error messages, wasted API calls.

**Fix Approach:**
- Add MIN_IMAGE_SIZE constant (12 bytes - minimum for format detection)
- Add MIN_VALID_IMAGE_SIZE map for format-specific minimums
  - PNG: 67 bytes (minimum valid PNG)
  - JPEG: 125 bytes (minimum valid JPEG)
  - GIF: 43 bytes, WebP: 20 bytes, AVIF: 100 bytes
- Implement validateBufferNotEmpty() method
- Update process() to validate before processing

**Acceptance Criteria:**
- [ ] Add MIN_IMAGE_SIZE constant (12 bytes)
- [ ] Add MIN_VALID_IMAGE_SIZE map for format minimums
- [ ] Implement validateBufferNotEmpty() method
- [ ] Update process() to validate before processing
- [ ] Add tests for empty buffer (0 bytes) - should throw
- [ ] Add tests for small buffer (<12 bytes) - should throw
- [ ] Add tests for valid buffer - should pass

**Files:** `src/lib/utils/imageProcessor.ts`

murdore:
### ◐ Audit status: partially done — keeping open

Core behaviour ships, but not every acceptance criterion is met. The remaining gap is the actionable part:

> Zero-byte case is handled; the broader 'too small to be a real image' case from the acceptance criteria is not.

<details><summary>what already works (evidence)</summary>

process() (imageProcessor.ts:87-95) throws "Invalid image processing: buffer is empty" for a 0-byte buffer, satisfying the empty-buffer criterion. But there is no MIN_IMAGE_SIZE constant (12 bytes), no MIN_VALID_IMAGE_SIZE per-format map, and no validateBufferNotEmpty() method (grep 0 hits for all three) — a non-empty-but-too-small buffer (e.g. 5 bytes) passes straight through with no rejection.

</details>


<sub>— triaged in a full audit of open issues (release v9.84.2, 2026-07-12). Reopen if anything above is off.</sub>

murdore:
[...]

Issue #289, opened 2025-12-01
MB-007: PDFs and CSVs Ignored in Content Array


murdore:
## Summary
Advanced `content[]` array API accepts PDF/CSV types but `convertContentToProviderFormat()` only processes text/images, silently ignoring PDFs/CSVs.

## Root Cause
Function implemented with only text/image support. When PDF/CSV types added to Content union, function not updated.

## Impact
- PDFs in content array → silently ignored
- CSVs in content array → silently ignored
- No error thrown → users unaware
- Inconsistent API (simple API works, advanced doesn't)

## Reproduction
```typescript
generate({
  input: {
    content: [
      { type: "text", text: "Analyze" },
      { type: "pdf", data: pdfBuffer }
    ]
  }
})
```
**Expected:** PDF processed
**Actual:** PDF ignored

## Fix
Extend `convertContentToProviderFormat()`:
1. Extract PDF/CSV content
2. Process CSVs to text
3. Call `convertMultimodalToProviderFormat()` with PDFs

**File:** `src/lib/utils/messageBuilder.ts`
**Lines:** 733-759

## Acceptance Criteria
- [ ] PDFs processed from content array
- [ ] CSVs processed from content array
- [ ] Mixed content works
- [ ] Text requirement maintained

murdore:
### ◐ Audit status: partially done — keeping open

Core behaviour ships, but not every acceptance criterion is met. The remaining gap is the actionable part:

> PDF-in-content-array is fixed; CSV-in-content-array is not -- half the acceptance criteria met, verified by direct execution against dist.

<details><summary>what already works (evidence)</summary>

src/lib/utils/messageBuilder.ts:1565-1603, convertContentToProviderFormat() now extracts `pdfContent = content.filter(c=>c.type==='pdf')` and forwards it to convertMultimodalToProviderFormat, so PDFs in the advanced content[] array ARE processed (confirmed live). But the function has no equivalent extraction for `c.type==='csv'` -- CSVContent items in the content[] array are still silently dropped (not text, not image, not pdf -- simply never read).

</details>


<sub>— triaged in a full audit of open issues (release v9.84.2, 2026-07-12). Reopen if anything above is off.</sub>

murdore:
Closing — verified implemented in `release` (v10.10.6, `0b1abf1d`).

[...]

PR #1203, opened 2026-07-17
fix(multimodal): fail-loud file errors, empty-image guard, CSV content items, audio/video detection, URL cache, tighter MessageContent

### LINKED ISSUE #273: MB-004: Inconsistent Error Handling in File Processing
## Summary
File processing errors handled inconsistently: some silently logged (line 496), others embedded in prompts (line 538-539).

## Root Cause
Different error handling strategies across codebase evolution:
- Early code embedded errors in prompts
- Later code used silent logging
Neither propagates errors properly to SDK callers.

## Impact
- Silent failures → users unaware files weren't processed
- Errors in prompts → AI confused by error text
- No stack traces → difficult debugging
- Unpredictable behavior

## Reproduction
1. Invalid file: `const badFile = Buffer.from('invalid')`
2. SDK call: `generate({ input: { text: "Analyze", files: [badFile] } })`
3. **Expected:** Exception thrown
4. **Actual:** Silent log OR error in prompt

## Fix
Replace silent logging and prompt embedding with proper error throwing:

**Location 1** (line 495-498): Throw error with filename
**Location 2** (line 535-540): Remove prompt embedding, throw error

## Acceptance Criteria
- [ ] File errors throw exceptions
- [ ] CSV errors throw exceptions
- [ ] NO error text in prompts
- [ ] Descriptive error messages

murdore:
## Summary

Consolidated **MessageBuilder / FileDetector / types** hardening — the last of the multimodal-subsystem backlog.

| Issue | Fix |
| ----- | --- |
| **#273** MB-004 — inconsistent file-error handling | `processUnifiedFilesArray` (`files[]`) and `processExplicitCsvFiles` (`csvFiles[]`) no longer swallow a failure (silent log / error text baked into the prompt) — both **log then throw** `Failed to process …`. |
| **#284** MB-006 — audio/video ignored | `detectMultimodal()` now checks `audioFiles`/`videoFiles` (matching the canonical `isMultimodalInput` guard), so an audio/video-only request reaches the multimodal builder. |
| **#289** MB-007 — CSV `content[]` items dropped | `convertContentToProviderFormat()` processes `type:"csv"` items, and the no-vision gate folds CSV-only content into the prompt (previously fell through to the text path). Shared via `appendCsvContentToText()`. |
| **#293** IMG-010 — no empty/truncated image handling | `ImageProcessor.process()` rejects empty, `< 12`-byte, and format-truncated buffers (`MIN_IMAGE_SIZE` + per-format `MIN_VALID_IMAGE_SIZE`), not just 0-byte. |
| **#323** FD-007 — double HTTP request | A short-TTL URL→`Content-Type` cache (populated by `MimeTypeStrategy`'s HEAD **and** `loadFromURL`'s GET) eliminates the redundant HEAD when a URL is detected more than once / after its body was fetched. |
| **#325** TD-008 — loose `MessageContent` index signature | Replaced `[key: string]: unknown` with the concrete fields the codebase reads (`text`/`image`/`mimeType`/`data`/`name`/`filename`/tool-call + tool-result), removing the "any key allowed" hole (the build confirmed no consumer relied on an un-enumerated key). |

## Verification

- **bugfixes suite** green except the 2 known-environmental `updater:` tests.
- **6 new deterministic tests**: image empty/tiny/truncated rejection; fail-loud CSV+file errors; CSV `content[]` processed; audio/video-only counts as multimodal; `MessageContent` concrete fields; and a local-server test proving a cached URL Content-Type issues **0** HEAD requests.

Fixes #273, #284, #289, #293, #323, #325.

## Summary by CodeRabbit

* **New Features**
  * Added explicit `audioFiles` support to multimodal and streaming inputs (audio/video-only now route correctly as multimodal).
* **Bug Fixes**
  * CSV/file processing now fails loudly instead of being silently ignored or placeholder-filled.
  * CSV `content[]` is preserved and folded into prompt text when vision isn’t used.
  * Improved image validation (rejects empty/too-small/truncated buffers more reliably).
* **Performance**
  * Reduced repeated remote MIME lookups via short-lived URL content-type caching.
* **Tests**
  * Added coverage for multimodal routing, CSV handling, stricter content typing, image hardening, and MIME caching behavior.

### DIFF
--- a/src/lib/core/modules/MessageBuilder.ts
+++ b/src/lib/core/modules/MessageBuilder.ts
@@ -36,7 +36,18 @@ function computeTotalContentLength(messages: ModelMessage[]): number {
 }
 
 /**
- * Check whether input contains multimodal content (images, files, PDFs, CSVs).
+ * Check whether input contains multimodal content (images, files, PDFs, CSVs,
+ * audio, video). #284: audioFiles/videoFiles were previously ignored here, so a
+ * request carrying only audio or video was misrouted to the text-only path.
+ *
+ * Intentional (round-2 review): routing audio/video through `isMultimodal`
+ * does NOT mean vision is required. `buildMultimodalMessagesArray` folds
+ * `audioFiles`/`videoFiles` into `inp.files` for auto-detection, and
+ * `appendDetectedFileResult` injects their content as plain text markers
+ * (`## Audio File:` / `## Video File:`), not image/vision blocks — see
+ * `test/continuous-test-suite-bugfixes.ts` ("MessageBuilder #284: ...").
+ * "Multimodal" here means "needs the multimodal builder so the payload isn't
+ * dropped", not "needs a vision model".
  */
 function detectMultimodal(opts: TextGenerationOptions | StreamOptions): {
   isMultimodal: boolean;
@@ -49,11 +60,20 @@ function detectMultimodal(opts: TextGenerationOptions | StreamOptions): {
   const hasCSVFiles = !!input?.csvFiles?.length;
   const hasPdfFiles = !!input?.pdfFiles?.length;
   const hasFiles = !!input?.files?.length;
+  const hasAudioFiles = !!input?.audioFiles?.length;
+  const hasVideoFiles = !!input?.videoFiles?.length;
   return {
     isMultimodal:
-      hasImages || hasContent || hasCSVFiles || hasPdfFiles || hasFiles,
+      hasImages ||
+      hasContent ||
+      hasCSVFiles ||
+      hasPdfFiles ||
+      hasFiles ||
+      hasAudioFiles ||
+      hasVideoFiles,
     hasImages,
-    hasFiles: hasCSVFiles || hasPdfFiles || hasFiles,
+    hasFiles:
+      hasCSVFiles || hasPdfFiles || hasFiles || hasAudioFiles || hasVideoFiles,
   };
 }
 
@@ -100,6 +120,12 @@ export class MessageBuilder {
               content: input?.content,
               csvFiles: input?.csvFiles,
               pdfFiles: input?.pdfFiles,
+              // #284: audioFiles/videoFiles must be forwarded too, or the
+              // multimodal builder receives an empty payload despite
+              // detectMultimodal() correctly routing audio/video-only
+              // requests here.
+              audioFiles: input?.audioFiles,
+              videoFiles: input?.videoFiles,
               files: input?.files,
             },
             csvOptions: options.csvOptions,
@@ -244,6 +270,10 @@ export class MessageBuilder {
               content: input?.content,
               csvFiles: input?.csvFiles,
               pdfFiles: input?.pdfFiles,
+              // #284: forward audioFiles/videoFiles here too — see the
+              // matching comment in buildMessages() above.
+              audioFiles: input?.audioFiles,
+              videoFiles: input?.videoFiles,
               files: input?.files,
             },
             csvOptions: options.csvOptions,
--- a/src/lib/types/multimodal.ts
+++ b/src/lib/types/multimodal.ts
@@ -487,15 +487,46 @@ export type MultimodalInput = {
 // ============================================
 
 /**
- * Content format for multimodal messages (used internally)
- * Compatible with Vercel AI SDK message format
+ * Content format for multimodal messages (used internally).
+ *
+ * #325: the loose `[key: string]: unknown` index signature has been replaced
+ * with the concrete fields the codebase actually reads/writes across the
+ * text / image / file / tool-call / tool-result shapes. This keeps the broad
+ * structural compatibility the internal pipeline relies on (a single object
+ * type, not a strict discriminated union that would force narrowing at every
+ * consumer) while removing the "any key is allowed" hole that let typos and
+ * unrelated keys through unchecked.
  */
 export type MessageContent = {
   type: string;
+  /** Text content (`type: "text"`). */
   text?: string;
+  /** Base64 / data-URI image (`type: "image"`). */
   image?: string;
+  /** MIME type for image/file parts. */
   mimeType?: string;
-  [key: string]: unknown; // Index signature for compatibility with Vercel AI SDK
+  /** Raw file bytes or base64 (`type: "file"`/document parts). */
+  data?: string | Buffer;
+  /** File name for document/file parts. */
+  name?: string;
+  /** File name (alias used by some file parts). */
+  filename?: string;
+  /** Tool-call identifier (`type: "tool-call"`/`"tool-result"`). */
+  toolCallId?: string;
+  /** Tool name (`type: "tool-call"`/`"tool-result"`). */
+  toolName?: string;
+  /** Tool-call arguments (`type: "tool-call"`). */
+  args?: Record<string, unknown>;
+  /** Tool-result payload (`type: "tool-result"`). */
+  result?: unknown;
+  /** Whether a tool-result represents an error (`type: "tool-result"`). */
+  isError?: boolean;
+  /**
+   * Provider-specific per-block options (e.g. Anthropic cache_control).
+   * Read as `item.providerOptions` when converting `MessageContent[]` to
+   * `ModelMessage[]` in `MessageBuilder.ts`.
+   */
+  providerOptions?: Record<string, unknown>;
 };
 
 /**
--- a/src/lib/utils/fileDetector.ts
+++ b/src/lib/utils/fileDetector.ts
@@ -41,6 +41,7 @@ import { tracers, ATTR, withSpan } from "../telemetry/index.js";
 import { CSVProcessor } from "./csvProcessor.js";
 import { ImageProcessor } from "./imageProcessor.js";
 import { logger } from "./logger.js";
+import { withTimeout } from "./errorHandling.js";
 import { redactUrlForError, sanitizeErrorCause } from "./logSanitize.js";
 import {
   mimeHintToExtension,
@@ -54,6 +55,91 @@ import { PDFProcessor } from "./pdfProcessor.js";
  */
 const DEFAULT_MAX_RETRIES = 3;
 const DEFAULT_RETRY_DELAY = 1000; // milliseconds
+
+/**
+ * Short-TTL cache of URL → Content-Type (#323). A URL is commonly detected more
+ * than once (repeated multimodal prompts reuse the same asset URL); caching the
+ * HEAD's content-type avoids re-issuing the HEAD each time. `loadFromURL` also
+ * populates it from its GET response, so once a URL's body has been fetched a
+ * subsequent detection needs no network round-trip at all.
+ *
+ * Trade-off (round-2 review): this is a module-level cache shared by every
+ * request in the process, and correctness relies solely on the 60s TTL — a
+ * signed URL whose response changes at the same path within that window would
+ * read stale. That is an intentional, bounded trade-off (60s of possible
+ * staleness for far fewer HEAD round-trips), not a freshness guarantee.
+ * Expired entries are removed lazily on their next `get()` (see
+ * `getCachedUrlContentType`); `setCachedUrlContentType` additionally sweeps
+ * expired entries opportunistically once the cache hits its size cap, so a
+ * URL that is cached once and never looked up again doesn't linger until the
+ * FIFO eviction below forces it out.
+ */
+const URL_CONTENT_TYPE_TTL_MS = 60_000;
+const URL_CONTENT_TYPE_CACHE_MAX_SIZE = 512;
+const urlContentTypeCache = new Map<
+  string,
+  { contentType: string; expiresAt: number }
+>();
+
+function getCachedUrlContentType(url: string, now: number): string | undefined {
+  const hit = urlContentTypeCache.get(url);
+  if (hit && hit.expiresAt > now) {
+    // Bump recency: Map iteration order follows insertion order, and the
+    // eviction below deletes the *first* key, so a plain `get` on a hot
+    // entry would leave it first in line for eviction despite being the
+    // most recently used. Re-inserting turns the size-bounded FIFO below
+    // into an actual LRU.
+    urlContentTypeCache.delete(url);
+    urlContentTypeCache.set(url, hit);
+    return hit.contentType;
+  }
+  if (hit) {
+    // Entry exists but its TTL has passed — treat as a miss and evict it
+    // immediately rather than serving (or retaining) stale data.
+    urlContentTypeCache.delete(url);
+  }
+  return undefined;
+}
+
+/**
+ * Opportunistically remove already-expired entries. Only called once the
+ * cache is at its size cap (see `setCachedUrlContentType`) so it doesn't add
+ * an O(n) scan to the common-case hot path.
+ */
+function pruneExpiredUrlContentTypeEntries(now: number): void {
+  for (const [key, entry] of urlContentTypeCache) {
+    if (entry.expiresAt <= now) {
+      urlContentTypeCache.delete(key);
+    }
+  }
+}
+
+function setCachedUrlContentType(
+  url: string,
+  contentType: string,
+  now: number,
+): void {
+  if (!contentType) {
+    return;
+  }
+  urlContentTypeCache.set(url, {
+    contentType,
+    expiresAt: now + URL_CONTENT_TYPE_TTL_MS,
+  });
+  // Bound the cache so a long-lived process can't grow it unbounded. Prefer
+  // reclaiming already-expired entries first; only fall back to evicting the
+  // oldest still-live entry (FIFO/LRU-ish, see getCachedUrlContentType) if
+  // the cache is still over the cap after pruning.
+  if (urlContentTypeCache.size > URL_CONTENT_TYPE_CACHE_MAX_SIZE) {
+    pruneExpiredUrlContentTypeEntries(now);
+  }
+  if (urlContentTypeCache.size > URL_CONTENT_TYPE_CACHE_MAX_SIZE) {
+    const oldest = urlContentTypeCache.keys().next().value;
+    if (oldest !== undefined) {
+      urlContentTypeCache.delete(oldest);
+    }
+  }
+}
 
 /**
  * Retryable network error codes (Node.js/undici network errors)
@@ -1824,6 +1910,14 @@ export class FileDetector {
             );
           }
 
+          // #323: cache the Content-Type from this GET so a subsequent detection
+          // of the same URL needs no HEAD.
+          setCachedUrlContentType(
+            url,
+            (response.headers["content-type"] as string) || "",
+            Date.now(),
+          );
+
           const chunks: Buffer[] = [];
           let totalSize = 0;
 
@@ -2233,15 +2327,30 @@ class MimeTypeStrategy implements DetectionStrategy {
     }
 
     try {
-      const response = await request(input, {
-        dispatcher: getGlobalDispatcher().compose(
-          interceptors.redirect({ maxRedirections: 5 }),
-        ),
-        method: "HEAD",
-        headersTimeout: FileDetector.DEFAULT_HEAD_TIMEOUT,
-        bodyTimeout: FileDetector.DEFAULT_HEAD_TIMEOUT,
-      });
-      const contentType = (response.headers["content-type"] as string) || "";
+      // #323: reuse a recently-seen Content-Type for this URL instead of
+      // re-issuing a HEAD (populated here and by loadFromURL's GET).
+      const now = Date.now();
+      let contentType = getCachedUrlContentType(input, now);
+      if (contentType === undefined) {
+        // Wrap the whole HEAD (request + body drain) in withTimeout so a stalled
+        // dump() can't hang detection, per the project's async-timeout guideline.
+        contentType = await withTimeout(
+          (async () => {
+            const response = await request(input, {
+              dispatcher: getGlobalDispatcher().compose(
+                interceptors.redirect({ maxRedirections: 5 }),
+              ),
+              method: "HEAD",
+              headersTimeout: FileDetector.DEFAULT_HEAD_TIMEOUT,
+              bodyTimeout: FileDetector.DEFAULT_HEAD_TIMEOUT,
+            });
+            await response.body.dump();
+            return (response.headers["content-type"] as string) || "";
+          })(),
+          FileDetector.DEFAULT_HEAD_TIMEOUT,
+        );
+        setCachedUrlContentType(input, contentType, now);
+      }
       const type = this.mimeToFileType(contentType);
 
       return {
--- a/src/lib/utils/imageProcessor.ts
+++ b/src/lib/utils/imageProcessor.ts
@@ -20,6 +20,25 @@ import { ErrorFactory, NeuroLinkError } from "./errorHandling.js";
 import type { ProcessedImage, FileProcessingResult } from "../types/index.js";
 
 /**
+ * Minimum bytes required just to detect an image format from magic bytes (#293).
+ */
+export const MIN_IMAGE_SIZE = 12;
+
+/**
+ * Minimum plausible byte size for a non-empty image of each format (#293). A
+ * buffer that carries a format's magic bytes but is smaller than this is
+ * truncated/corrupt and would fail (or render blank) downstream — reject early
+ * with a clear message instead.
+ */
+export const MIN_VALID_IMAGE_SIZE: Record<string, number> = {
+  "image/png": 67,
+  "image/jpeg": 125,
+  "image/gif": 43,
+  "image/webp": 20,
+  "image/avif": 100,
+};
+
+/**
  * Network error codes that should trigger a retry
  */
 const RETRYABLE_ERROR_CODES = new Set([
@@ -119,13 +138,9 @@ export class ImageProcessor {
     content: Buffer,
     _options?: unknown,
   ): Promise<FileProcessingResult> {
-    // Validate content is non-empty before processing
-    if (content.length === 0) {
-      logger.error("Empty buffer provided");
-      throw new Error("Invalid image processing: buffer is empty");
-    }
-
-    const mediaType = this.detectImageType(content);
+    // #293: reject empty AND small/truncated buffers (not just 0 bytes) before
+    // processing; returns the detected media type.
+    const mediaType = this.validateBufferNotEmpty(content);
 
     // #261/#286 follow-up: fail loud instead of packaging the octet-stream
     // sentinel as a "valid" image (see assertKnownImageType() above).
@@ -149,6 +164,92 @@ export class ImageProcessor {
         size: content.length,
       },
     } satisfies FileProcessingResult;
+  }
+
+  /**
+   * Strict magic-byte match: does `content` actually begin with `mediaType`'s
+   * signature? `detectImageType` returns `image/jpeg` as a *fallback* for
+   * unrecognized data, so a per-format minimum must only be enforced when the
+   * bytes genuinely match — otherwise valid BMP/TIFF/unknown buffers would be
+   * wrongly rejected as "truncated jpeg" (#293 review).
+   */
+  private static hasStrictImageMagic(
+    content: Buffer,
+    mediaType: string,
+  ): boolean {
+    const b = content;
+    switch (mediaType) {
+      case "image/png":
+        return (
+          b.length >= 4 &&
+          b[0] === 0x89 &&
+          b[1] === 0x50 &&
+          b[2] === 0x4e &&
+          b[3] === 0x47
+        );
+      case "image/jpeg":
+        return b.length >= 3 && b[0] === 0xff && b[1] === 0xd8 && b[2] === 0xff;
+      case "image/gif":
+        return b.length >= 3 && b[0] === 0x47 && b[1] === 0x49 && b[2] === 0x46;
+      case "image/webp":
+        return (
+          b.length >= 12 &&
+          b.subarray(0, 4).toString("latin1") === "RIFF" &&
+          b.subarray(8, 12).toString("latin1") === "WEBP"
+        );
+      case "image/avif":
+        return (
+          b.length >= 12 &&
+          b.subarray(4, 8).toString("latin1") === "ftyp" &&
+          b.subarray(8, 12).toString("latin1").startsWith("avif")
+        );
+      default:
+        return false;
+    }
+  }
+
+  /**
+   * Validate that a buffer is a plausibly-complete image (#293): non-empty and,
+   * for a strictly-identified raster format, at least the minimum plausible byte
+   * size. Returns the detected media type.
+   *
+   * @throws Error when the buffer is empty or a recognized format is truncated.
+   */
+  private static validateBufferNotEmpty(content: Buffer): string {
+    if (content.length === 0) {
+      logger.error("Empty buffer provided");
+      throw ErrorFactory.imageBufferInvalid(
+        "Invalid image processing: buffer is empty",
+      );
+    }
+    const mediaType = this.detectImageType(content);
+    // SVG is text and can be legitimately tiny (e.g. `<svg/>`), so the raster
+    // size floors don't apply to it (#293 review).
+    if (mediaType === "image/svg+xml") {
+      return mediaType;
+    }
+    // A buffer too small to carry any raster header is not a usable image.
+    if (content.length < MIN_IMAGE_SIZE) {
+      throw ErrorFactory.imageBufferInvalid(
+        `Invalid image processing: buffer too small (${content.length} bytes) — ` +
+          `a minimum of ${MIN_IMAGE_SIZE} bytes is required for a valid image.`,
+      );
+    }
+    // Only enforce a per-format minimum when the format was strictly identified
+    // from magic bytes (never from detectImageType's jpeg fallback), so valid
+    // BMP/TIFF/unknown buffers are not mis-rejected as "truncated jpeg".
+    const minForFormat = MIN_VALID_IMAGE_SIZE[mediaType];
+    if (
+      minForFormat !== undefined &&
+      content.length < minForFormat &&
+      this.hasStrictImageMagic(content, mediaType)
+    ) {
+      throw ErrorFactory.imageBufferInvalid(
+        `Invalid image processing: ${mediaType} buffer is truncated ` +
+          `(${content.length} bytes, minimum ${minForFormat} for a valid ${mediaType}).`,
+      );
+    }
+    return mediaType;
   }
 
   /**
--- a/src/lib/utils/messageBuilder.ts
+++ b/src/lib/utils/messageBuilder.ts
@@ -15,10 +15,11 @@ import {
   FILE_READ_BUDGET_PERCENT,
 } from "../context/fileTokenBudget.js";
 import type { FileReferenceRegistry } from "../files/fileReferenceRegistry.js";
-import { SIZE_TIER_THRESHOLDS } from "../types/index.js";
+import { isCSVContent, SIZE_TIER_THRESHOLDS } from "../types/index.js";
 import type {
   ChatMessage,
   Content,
+  CSVContent,
   FileInput,
   FileWithMetadata,
   GenerateOptions,
@@ -29,10 +30,10 @@ import type {
   TextGenerationOptions,
 } from "../types/index.js";
 import { tracers, ATTR, withSpan } from "../telemetry/index.js";
-import { NeuroLinkError } from "./errorHandling.js";
+import { ErrorFactory, NeuroLinkError, withTimeout } from "./errorHandling.js";
 import { FileDetector } from "./fileDetector.js";
 import { getImageCache } from "./imageCache.js";
-import { ImageProcessor } from "./imageProcessor.js";
+import { ImageProcessor, imageUtils } from "./imageProcessor.js";
 import { logger } from "./logger.js";
 import { PDFImageConverter, PDFProcessor } from "./pdfProcessor.js";
 import { urlDownloadRateLimiter } from "./rateLimiter.js";
@@ -1080,8 +1081,15 @@ export async function processUnifiedFilesArray(
           );
         } catch (error) {
           const errMsg = error instanceof Error ? error.message : String(error);
+          // #273: don't silently drop a failed file — log, then throw so the
+          // caller learns the file couldn't be processed (matches the explicit
+          // pdf/csv paths' fail-loud behavior).
           logger.error(
-            `[NEUROLINK] File skipped/failed: ${filename} — reason: ${errMsg}`,
+            `[NEUROLINK] File processing failed: ${filename} — reason: ${errMsg}`,
+          );
+          throw ErrorFactory.fileProcessingFailed(
+            filename,
+            error instanceof Error ? error : new Error(errMsg),
           );
         }
       }
@@ -1179,10 +1187,15 @@ async function processExplicitCsvFiles(
       options.input.text += csvSection;
       logger.info(`[CSV] ✅ Processed: ${filename}`);
     } catch (error) {
-      logger.error(`[CSV] ❌ Failed:`, error);
       const filename = extractFilename(csvFile, i);
-      options.input.text += `\n\n## CSV Data Error: Failed to process "${filename}"`;
-      options.input.text += `\nReason: ${error instanceof Error ? error.message : "Unknown error"}`;
+      const errMsg = error instanceof Error ? error.message : String(error);
+      // #273: fail loud instead of embedding the error into the prompt text
+      // (which the model would then "analyze"). Log, then throw.
+      logger.error(`[CSV] ❌ Failed to process ${filename}: ${errMsg}`);
+      throw ErrorFactory.csvProcessingFailed(
+        filename,
+        error instanceof Error ? error : new Error(errMsg),
+      );
     }
   }
 }
@@ -1367,6 +1380,19 @@ export async function buildMultimodalMessagesArray(
   // rest of this function, avoiding 60+ "possibly undefined" errors.
   const inp = options.input;
 
+  // #284: audioFiles/videoFiles have no dedicated processor — route them
+  // through the same auto-detecting `files` pipeline that already
+  // understands "audio"/"video" FileDetector results (see
+  // appendDetectedFileResult), instead of silently dropping them once
+  // detectMultimodal() routes an audio/video-only request here.
+  if (inp.audioFiles?.length || inp.videoFiles?.length) {
+    inp.files = [
+      ...(inp.files || []),
+      ...(inp.audioFiles || []),
+      ...(inp.videoFiles || []),
+    ];
+  }
+
   // Compute provider-specific max PDF size once for consistent validation
   const pdfConfig = PDFProcessor.getProviderConfig(provider);
   const maxSize = pdfConfig
@@ -1397,6 +1423,13 @@ export async function buildMultimodalMessagesArray(
 
   // If no images or PDFs, use standard message building and convert to MultimodalChatMessage[]
   if (!hasImages && !hasPDFs) {
+    // #289: CSV content[] items don't need vision, so they never reach the
+    // multimodal converter below — process them into the prompt text here
+    // (otherwise a `content: [{type:"csv"}]`-only request silently drops it).
+    const csvContentItems = inp.content?.filter(isCSVContent) ?? [];
+    if (csvContentItems.length > 0) {
+      inp.text = await appendCsvContentToText(csvContentItems, inp.text ?? "");
+    }
     if (inp.csvFiles) {
       inp.csvFiles = [];
     }
@@ -1563,6 +1596,59 @@ export async function buildMultimodalMessagesArray(
 }
 
 /**
+ * Timeout for detecting/parsing an in-memory CSV `content[]` buffer (#325
+ * review, round 2). Bounds a stalled detector rather than blocking the
+ * request indefinitely.
+ */
+const CSV_CONTENT_DETECTION_TIMEOUT_MS = 30_000;
+
+/**
+ * #289: process CSV `content[]` items into appended prompt text. CSV is
+ * delivered to the model as text (like the explicit `csvFiles` path), so this
+ * is shared by both the no-vision gate and the multimodal converter.
+ */
+async function appendCsvContentToText(
+  csvItems: CSVContent[],
+  baseText: string,
+): Promise<string> {
+  let text = baseText;
+  for (const csv of csvItems) {
+    const raw = csv.data;
+    if (raw === undefined) {
+      continue;
+    }
+    // #325: raw CSV text (e.g. "a,b\n1,2") is not base64 — decoding it as
+    // base64 silently corrupts the content. Only treat the string as base64
+    // when it actually validates as such; otherwise treat it as literal
+    // UTF-8 CSV text, matching how Buffer inputs are already handled as-is.
+    const buffer =
+      typeof raw === "string"
+        ? imageUtils.isValidBase64(raw)
+          ? Buffer.from(raw, "base64")
+          : Buffer.from(raw, "utf-8")
+        : raw;
+    const name = csv.metadata?.filename || "data.csv";
+    // #325 review (round 2): a stalled/hung detector must not block the
+    // request indefinitely — wrap with the project's standard withTimeout.
+    const result = await withTimeout(
+      FileDetector.detectAndProcess(buffer, {
+        allowedTypes: ["csv"],
+        csvOptions: {
+          maxRows: csv.metadata?.maxRows,
+          formatStyle: csv.metadata?.formatStyle,
+        },
+      }),
+      CSV_CONTENT_DETECTION_TIMEOUT_MS,
+      new Error(
+        `Timed out processing CSV content "${name}" after ${CSV_CONTENT_DETECTION_TIMEOUT_MS}ms`,
+      ),
+    );
+    text += `${text ? "\n\n" : ""}## CSV Data from ${name}\n${result.content}`;
+  }
+  return text;
+}
+
+/**
  * Convert advanced content format to provider-specific format
  */
 async function convertContentToProviderFormat(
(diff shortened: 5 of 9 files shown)


src/lib/types/multimodal.ts
```diff
@@ -487,15 +487,40 @@ export type MultimodalInput = {
...
 export type MessageContent = {
   type: string;
+  /** Text content (`type: "text"`). */
   text?: string;
+  /** Base64 / data-URI image (`type: "image"`). */
   image?: string;
+  /** MIME type for image/file parts. */
   mimeType?: string;
-  [key: string]: unknown; // Index signature for compatibility with Vercel AI SDK
+  /** Raw file bytes or base64 (`type: "file"`/document parts). */
```
Tara-ag, line 508: ⚠️ **MAJOR**: The tightened `MessageContent` type is incomplete for the actual internal pipeline.

The PR description says it enumerates "concrete fields the codebase reads/writes", but the codebase also reads `mediaType` for image parts (e.g. `ImageContent` uses `mediaType`, and provider adapters consume it), `url` for image payloads, and possibly `metadata` for file/document parts. Removing the index signature without adding those fields will force consumers to cast or break compile-time compatibility with existing AI-SDK-shaped parts.

Please either (a) add the missing fields that are genuinely read (`mediaType`, `url`, `metadata`, etc.) and verify with `pnpm run check`, or (b) keep a narrowly-typed index signature such as `[key: string]: string | Buffer | unknown` if broad compatibility is intentional. The current list is not yet safe to ship as "complete".

src/lib/utils/messageBuilder.ts
```diff
@@ -1559,6 +1576,43 @@ export async function buildMultimodalMessagesArray(
...
+  baseText: string,
+): Promise<string> {
+  let text = baseText;
+  for (const csv of csvItems) {
+    const raw = csv.data;
+    if (raw === undefined) {
+      continue;
+    }
+    const buffer = typeof raw === "string" ? Buffer.from(raw, "base64") : raw;
+    const result = await FileDetector.detectAndProcess(buffer, {
```
Tara-ag: ⚠️ **MAJOR**: `appendCsvContentToText()` assumes any string CSV payload is base64-encoded, but `CSVContent.data` is typed as `Buffer | string` without requiring base64.

A caller passing raw CSV text as a string (e.g. `{ type: "csv", data: "a,b\n1,2" }`) will be corrupted by `Buffer.from(raw, "base64")`, producing mangled bytes and failing CSV detection/parsing. This contradicts the type contract and the PR's goal of correctly handling CSV `content[]` items.

Fix: either (a) document and enforce that string `data` must be base64 (e.g. by validating it is base64 before decoding and throwing a clear error otherwise), or (b) support both raw text and base64 by detecting whether the string is already valid CSV/plain text. The safest minimal fix is to check if the string is valid base64 and decode only then; otherwise treat it as raw CSV text:

```ts
const buffer = typeof raw === "string"
  ? (isBase64(raw) ? Buffer.from(raw, "base64") : Buffer.from(raw, "utf-8"))
  : raw;
```

src/lib/utils/imageProcessor.ts
```diff
@@ -112,6 +127,35 @@ export class ImageProcessor {
...
 
+  /**
+   * Validate that a buffer is a plausibly-complete image (#293): non-empty,
+   * large enough to detect a format, and — for a recognized format — at least
+   * the minimum plausible byte size. Returns the detected media type.
+   *
+   * @throws Error when the buffer is empty, too small to detect, or truncated.
+   */
+  private static validateBufferNotEmpty(content: Buffer): string {
+    if (content.length === 0) {
```
Tara-ag, line 219: ⚠️ **MAJOR**: `detectImageType()` falls back to `"image/jpeg"` for buffers that don't match any magic bytes, so this validation misreports unrecognized formats as truncated JPEGs.

For example, a 15-byte buffer of random data will pass the `MIN_IMAGE_SIZE` check, fail to match any magic bytes, and then throw `"Invalid image processing: image/jpeg buffer is truncated (15 bytes, minimum 125 for a valid image/jpeg)"`. The buffer isn't a truncated JPEG — it's an unrecognized or unsupported format — so the error is misleading and will make debugging harder for callers.

Fix: make `detectImageType()` return a sentinel such as `"unknown"` or `undefined` for unrecognized buffers, and handle that case with a distinct error before applying per-format minimum sizes:

```ts
const mediaType = this.detectImageType(content);
if (mediaType === "unknown") {
  throw new Error(
    `Invalid image processing: unrecognized image format (${content.length} bytes).`,
  );
}
const minForFormat = MIN_VALID_IMAGE_SIZE[mediaType];
// ... existing truncated check
```

src/lib/utils/fileDetector.ts
```diff
@@ -53,6 +53,51 @@ import { PDFProcessor } from "./pdfProcessor.js";
...
+/**
+ * Short-TTL cache of URL → Content-Type (#323). A URL is commonly detected more
+ * than once (repeated multimodal prompts reuse the same asset URL); caching the
+ * HEAD's content-type avoids re-issuing the HEAD each time. `loadFromURL` also
+ * populates it from its GET response, so once a URL's body has been fetched a
+ * subsequent detection needs no network round-trip at all.
+ */
+const URL_CONTENT_TYPE_TTL_MS = 60_000;
+const urlContentTypeCache = new Map<
+  string,
```
Tara-ag, line 80: 💡 **MINOR**: The URL Content-Type cache uses insertion-order eviction (`Map.keys().next()`), which evicts the oldest entry regardless of TTL. A URL that is actively reused will eventually be evicted if 512 other URLs are touched, even if it was accessed milliseconds ago.

Consider a tiny LRU or at least re-inserting on cache hit so frequently-used URLs survive. Given the 60-second TTL, this is unlikely to cause real harm, but it weakens the cache's effectiveness under mixed traffic.

src/lib/core/modules/MessageBuilder.ts
```diff
@@ -49,11 +51,20 @@ function detectMultimodal(opts: TextGenerationOptions | StreamOptions): {
   const hasCSVFiles = !!input?.csvFiles?.length;
```
Tara-ag, line 60: 💡 **MINOR**: `detectMultimodal()` now marks audio/video-only requests as multimodal, but `MessageBuilder.buildMessages()` only forwards `images`, `content`, `csvFiles`, `pdfFiles`, and `files` to `buildMultimodalMessagesArray`; it does **not** forward `audioFiles` or `videoFiles`.

So while the request is routed to the multimodal builder, the actual audio/video payloads are dropped from the cloned `multimodalOptions.input`. This means the #284 fix only changes the routing flag, not whether the content reaches the model. Please confirm whether `buildMultimodalMessagesArray` is expected to receive audio/video via a different path (e.g. inside `content` or `files`), or add `audioFiles`/`videoFiles` to the forwarded input so the detection flag and the data stay consistent.

Tara-ag (changes requested):
## Review Summary

Reviewed **6 files** in PR #1203 (MessageBuilder / FileDetector / types hardening).

### New issues raised

| Severity | Count | Files |
|----------|-------|-------|
| 🔒 CRITICAL | 0 | — |
| ⚠️ MAJOR | 3 | `src/lib/types/multimodal.ts`, `src/lib/utils/messageBuilder.ts`, `src/lib/utils/imageProcessor.ts` |
| 💡 MINOR | 2 | `src/lib/utils/fileDetector.ts`, `src/lib/core/modules/MessageBuilder.ts` |
| 💬 SUGGESTION | 0 | — |

### Blocking rationale

This PR meets the **"3 or more MAJOR severity issues"** blocking criterion:

1. **`MessageContent` type is incomplete** (`src/lib/types/multimodal.ts:508`) — removes the broad index signature without adding fields the pipeline actually reads (`mediaType`, `url`, `metadata`), which will force unsafe casts or break AI-SDK-shaped consumers.
2. **CSV string data assumed base64** (`src/lib/utils/messageBuilder.ts:1599`) — `CSVContent.data: Buffer | string` allows raw CSV text, but `appendCsvContentToText()` unconditionally decodes strings as base64, corrupting valid text payloads.
3. **Unrecognized image formats misreported as truncated JPEG** (`src/lib/utils/imageProcessor.ts:138`) — `detectImageType()` defaults to `image/jpeg` for unknown buffers, so random/small data throws a misleading "truncated JPEG" error instead of "unrecognized format".

Two of these concerns overlap with existing Copilot review comments on the same lines; I am raising them independently because they remain unaddressed and are blocking-quality issues.

### Non-blocking notes

- `src/lib/utils/fileDetector.ts:65` — URL Content-Type cache evicts by insertion order rather than LRU/TTL, weakening cache effectiveness under mixed traffic.
- `src/lib/core/modules/MessageBuilder.ts:51` — `detectMultimodal()` now flags audio/video-only input as multimodal, but `buildMessages()` does not forward `audioFiles`/`videoFiles` into the multimodal builder, so the routing flag and the data are inconsistent.

### Next steps

Please address the three MAJOR issues (with the suggested fixes in the inline comments), re-run `pnpm run check` / `pnpm test`, and re-request review. The MINOR items can be addressed in follow-ups or in this PR at your discretion.

murdore:
Addressed the CodeRabbit review findings (force-pushed):

- **#293 — per-format minimum applied on fallback detection (imageProcessor.ts:30).** `detectImageType` returns `image/jpeg` as a *fallback* for unrecognized data, so the 125-byte JPEG minimum could reject valid BMP/TIFF/unknown buffers as "truncated jpeg". Now the minimum is only enforced on a **strict magic-byte match** (new `hasStrictImageMagic`). SVG (text, legitimately tiny) is exempted from the raster size floors, while a truncated real PNG/JPEG is still rejected. New assertions cover the BMP-fallback and tiny-SVG cases.
- **#323 — HEAD not wrapped in `withTimeout` (fileDetector.ts:2214).** The HEAD request + `body.dump()` are now wrapped in `withTimeout(…, DEFAULT_HEAD_TIMEOUT)` so a stalled drain can't hang detection.
- **ErrorFactory (imageProcessor.ts:154).** Kept descriptive raw `Error`s here — the file throws raw `Error` at 21 sites and uses `ErrorFactory` at 0, so raw is the established convention in this module; the messages are already actionable and prefixed.

`check`/`lint`/`build` clean; bugfixes suite green except the 2 known-environmental `updater:` tests.

src/lib/utils/messageBuilder.ts
```diff
@@ -1559,6 +1576,43 @@ export async function buildMultimodalMessagesArray(
...
+      allowedTypes: ["csv"],
+      csvOptions: {
+        maxRows: csv.metadata?.maxRows,
+        formatStyle: csv.metadata?.formatStyle as
+          | "raw"
+          | "markdown"
```
Tara-ag: 💡 **MINOR**: `formatStyle` is cast from `string` to the allowed union without validation. The local parameter type widens `CSVContent["metadata"]` to `{ formatStyle?: string }`, so an invalid string can reach `CSVProcessor` via the cast.

Narrow the parameter type to `CSVContent[]` (or `Array<{ metadata?: CSVContent["metadata"] }>`) so TypeScript enforces the union and the unsafe cast can be removed:

```ts
async function appendCsvContentToText(
  csvItems: CSVContent[],
  baseText: string,
): Promise<string> {
  ...
  csvOptions: {
    maxRows: csv.metadata?.maxRows,
    formatStyle: csv.metadata?.formatStyle, // no cast needed
  },
}
```

src/lib/utils/messageBuilder.ts
```diff
@@ -1077,9 +1077,15 @@ export async function processUnifiedFilesArray(
...
+          // #273: don't silently drop a failed file — log, then throw so the
+          // caller learns the file couldn't be processed (matches the explicit
+          // pdf/csv paths' fail-loud behavior).
           logger.error(
-            `[NEUROLINK] File skipped/failed: ${filename} — reason: ${errMsg}`,
+            `[NEUROLINK] File processing failed: ${filename} — reason: ${errMsg}`,
           );
+          throw new Error(`Failed to process file "${filename}": ${errMsg}`, {
+            cause: error,
+          });
```
Tara-ag: 💡 **MINOR**: The new fail-loud errors in `processUnifiedFilesArray` and `processExplicitCsvFiles` are raw `Error` objects. CLAUDE.md Common Patterns says to use `ErrorFactory` for typed errors. Consider replacing with `ErrorFactory.fileProcessingFailed(filename, errMsg)` (or the closest existing factory method) so callers can rely on the typed error contract.

src/lib/utils/imageProcessor.ts
```diff
@@ -112,6 +127,90 @@ export class ImageProcessor {
...
+    const minForFormat = MIN_VALID_IMAGE_SIZE[mediaType];
+    if (
+      minForFormat !== undefined &&
+      content.length < minForFormat &&
+      this.hasStrictImageMagic(content, mediaType)
+    ) {
+      throw new Error(
+        `Invalid image processing: ${mediaType} buffer is truncated ` +
+          `(${content.length} bytes, minimum ${minForFormat} for a valid ${mediaType}).`,
+      );
```
Tara-ag, line 250: 💡 **MINOR**: The validation errors thrown here are raw `Error` objects. CLAUDE.md Common Patterns recommends using `ErrorFactory` for typed errors. Consider adding/using an `ErrorFactory.imageProcessingFailed(...)` variant so callers can distinguish image-validation failures from generic exceptions.

Tara-ag (commented):
## Review Summary

Reviewed the 6 changed files in this multimodal hardening PR.

### New comments raised this run
- `src/lib/utils/messageBuilder.ts:1605` — `formatStyle` is cast from `string` to the allowed union without validation; narrowing the parameter type to `CSVContent[]` would remove the unsafe cast.
- `src/lib/utils/messageBuilder.ts:1088` / `1192` — the new fail-loud errors use raw `Error` objects; CLAUDE.md prefers `ErrorFactory` typed errors. *(Note: this echoes existing unresolved CodeRabbit feedback on the same lines.)*
- `src/lib/utils/imageProcessor.ts:209` — validation errors are raw `Error` objects; consider using `ErrorFactory`. *(Note: this echoes existing unresolved CodeRabbit feedback on the same method.)*

### Existing unresolved feedback already covering major concerns
I did **not** re-raise these, but they remain unaddressed and are relevant to the merge decision:
- `appendCsvContentToText()` assumes string CSV payloads are base64-encoded, which corrupts raw CSV text (raised by Tara-ag / copilot reviewer).
- Tightened `MessageContent` type appears incomplete for the internal pipeline (`mediaType`, `url`, `metadata` are used but missing) (raised by Tara-ag / copilot reviewer).
- `detectMultimodal()` now routes audio/video-only requests as multimodal, but `MessageBuilder.buildMessages()` does not forward `audioFiles`/`videoFiles` into `multimodalOptions.input` (raised by Tara-ag).
- `ImageProcessor` JPEG fallback can misreport unrecognized buffers (raised by Tara-ag).

### Decision
No **new** blocking criteria were identified beyond the points already raised by prior reviewers. I am leaving this as a comment review so the existing unresolved MAJOR feedback can be addressed before merge.

src/lib/utils/messageBuilder.ts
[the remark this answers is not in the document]
murdore: Fixed in c791e9a3315b0eac1deb26b445ee01f260cbbea0: `appendCsvContentToText()` now validates the string via `imageUtils.isValidBase64()` (imported from `imageProcessor.ts`, no circular-import risk) before decoding — only genuinely base64-encoded strings are `Buffer.from(raw, "base64")`'d; anything else (e.g. raw CSV text like `"a,b\n1,2"`) is decoded as UTF-8 instead. Added a deterministic, network-free regression test "MessageBuilder #325: raw (non-base64) CSV string content is not corrupted by forced base64 decoding" in test/continuous-test-suite-bugfixes.ts that builds a message from raw CSV text and asserts the parsed JSON output contains the correct (uncorrupted) values.

murdore: Fixed in c791e9a3315b0eac1deb26b445ee01f260cbbea0: `appendCsvContentToText()` now validates the string via `imageUtils.isValidBase64()` (imported from `imageProcessor.ts`, no circular-import risk) before decoding — only genuinely base64-encoded strings are `Buffer.from(raw, "base64")`'d; anything else (e.g. raw CSV text like `"a,b\n1,2"`) is decoded as UTF-8 instead. Added a deterministic, network-free regression test "MessageBuilder #325: raw (non-base64) CSV string content is not corrupted by forced base64 decoding" in test/continuous-test-suite-bugfixes.ts that builds a message from raw CSV text and asserts the parsed JSON output contains the correct (uncorrupted) values.

src/lib/types/multimodal.ts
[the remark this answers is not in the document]
murdore: Verified against current HEAD (c791e9a3315b0eac1deb26b445ee01f260cbbea0), no code change needed: `pnpm run check` passes with 0 errors/warnings across all 4748 files, and the only real consumer of `MessageContent`-typed content (`amazonBedrock.ts:1071`, `msg.content.forEach((contentItem: MessageContent) => ...)`) reads only `type`, `text`, `image`, `mimeType`, `data`, `name` — all present on the type. `mediaType` / `url` / `metadata` belong to the separate, unrelated `ImageContent` type (`multimodal.ts:58-73`), which is the type actually used wherever those fields are read (provider adapters etc.) — the tightened `MessageContent` type was never meant to carry them. No new unsafe casts were introduced by the original change (`git diff ad5e9f85..HEAD` shows zero new `as MessageContent` casts). As reinforcement, we also added a negative compile-time case (`@ts-expect-error`) to the existing "Types #325" test proving an unsupported key is rejected, so a regression that reintroduces the removed index signature would be caught.

murdore: Verified against current HEAD (c791e9a3315b0eac1deb26b445ee01f260cbbea0), no code change needed: `pnpm run check` passes with 0 errors/warnings across all 4748 files, and the only real consumer of `MessageContent`-typed content (`amazonBedrock.ts:1071`, `msg.content.forEach((contentItem: MessageContent) => ...)`) reads only `type`, `text`, `image`, `mimeType`, `data`, `name` — all present on the type. `mediaType` / `url` / `metadata` belong to the separate, unrelated `ImageContent` type (`multimodal.ts:58-73`), which is the type actually used wherever those fields are read (provider adapters etc.) — the tightened `MessageContent` type was never meant to carry them. No new unsafe casts were introduced by the original change (`git diff ad5e9f85..HEAD` shows zero new `as MessageContent` casts). As reinforcement, we also added a negative compile-time case (`@ts-expect-error`) to the existing "Types #325" test proving an unsupported key is rejected, so a regression that reintroduces the removed index signature would be caught.

src/lib/utils/imageProcessor.ts
```diff
@@ -112,6 +127,35 @@ export class ImageProcessor {
...
+          `a minimum of ${MIN_IMAGE_SIZE} bytes is required for format detection.`,
+      );
+    }
+    const mediaType = this.detectImageType(content);
+    const minForFormat = MIN_VALID_IMAGE_SIZE[mediaType];
+    if (minForFormat !== undefined && content.length < minForFormat) {
+      throw new Error(
+        `Invalid image processing: ${mediaType} buffer is truncated ` +
+          `(${content.length} bytes, minimum ${minForFormat} for a valid ${mediaType}).`,
+      );
```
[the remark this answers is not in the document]
murdore: Reviewed, closing as out of scope for this PR (no code change): `ErrorFactory` (src/lib/utils/errorHandling.ts) has no generic image-processing-failure method — the existing image-related methods (`imageTooSmall`, `invalidImageFormat`, `emptyImagePath`, `invalidImageType`) all have fixed messages and don't match the dynamic `"buffer too small (N bytes)..."` / `"...buffer is truncated (N bytes, minimum M for a valid ...)"` messages thrown by `validateBufferNotEmpty()`. `imageProcessor.ts` throws raw `Error` 21 times throughout the file — this is the file's pre-existing, consistent convention, not something introduced by this PR. Swapping only these 3 new throws would be inconsistent with the other 18+ existing raw-`Error` sites in the same file; inventing new dynamic-message `ErrorFactory` methods and retrofitting all sites for consistency is a broader refactor genuinely out of scope for this targeted type/validation fix. Happy to pick this up as a dedicated follow-up PR if desired.

src/lib/utils/imageProcessor.ts
murdore: Reviewed, closing as out of scope for this PR (no code change): `ErrorFactory` (src/lib/utils/errorHandling.ts) has no generic image-processing-failure method — the existing image-related methods (`imageTooSmall`, `invalidImageFormat`, `emptyImagePath`, `invalidImageType`) all have fixed messages and don't match the dynamic `"buffer too small (N bytes)..."` / `"...buffer is truncated (N bytes, minimum M for a valid ...)"` messages thrown by `validateBufferNotEmpty()`. `imageProcessor.ts` throws raw `Error` 21 times throughout the file — this is the file's pre-existing, consistent convention, not something introduced by this PR. Swapping only these 3 new throws would be inconsistent with the other 18+ existing raw-`Error` sites in the same file; inventing new dynamic-message `ErrorFactory` methods and retrofitting all sites for consistency is a broader refactor genuinely out of scope for this targeted type/validation fix. Happy to pick this up as a dedicated follow-up PR if desired.

src/lib/utils/messageBuilder.ts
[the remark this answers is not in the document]
murdore: Reviewed, closing as out of scope for this PR (no code change): `ErrorFactory` has no matching "generic file/CSV processing failed" method — every factory method is either tool-execution-shaped (requires a `toolName`) or a fixed-message video/ppt/image-validation-specific method, none of which fit a dynamic `Failed to process file "${filename}": ${errMsg}"` message. `messageBuilder.ts` throws raw `Error` 14 times throughout the file as its pre-existing convention; the two new throws (now at lines 1088 and 1192) match that convention exactly, including correct `{ cause: error }` attachment. Swapping only these 2 of 14 raw-`Error` sites would be inconsistent without a broader, out-of-scope refactor across the whole file. Closing as-is.

murdore: Reviewed, closing as out of scope for this PR (no code change): `ErrorFactory` has no matching "generic file/CSV processing failed" method — every factory method is either tool-execution-shaped (requires a `toolName`) or a fixed-message video/ppt/image-validation-specific method, none of which fit a dynamic `Failed to process file "${filename}": ${errMsg}"` message. `messageBuilder.ts` throws raw `Error` 14 times throughout the file as its pre-existing convention; the two new throws (now at lines 1088 and 1192) match that convention exactly, including correct `{ cause: error }` attachment. Swapping only these 2 of 14 raw-`Error` sites would be inconsistent without a broader, out-of-scope refactor across the whole file. Closing as-is.

src/lib/utils/messageBuilder.ts
```diff
@@ -1559,6 +1576,43 @@ export async function buildMultimodalMessagesArray(
...
+      allowedTypes: ["csv"],
+      csvOptions: {
+        maxRows: csv.metadata?.maxRows,
+        formatStyle: csv.metadata?.formatStyle as
+          | "raw"
+          | "markdown"
+          | "json"
+          | undefined,
+      },
+    });
```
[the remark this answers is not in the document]
murdore: Reviewed, closing as out of scope (no code change): `withTimeout` is not imported/used anywhere else in `messageBuilder.ts`, and none of the other 5 `FileDetector.detectAndProcess(...)` call sites in this same file (lines 682, 718, 1051, 1163, 1284) are wrapped either — adding it only here would be a one-off inconsistency. More importantly, this specific call is passed an already-decoded in-memory `Buffer` (not a URL) — it's pure CPU-bound CSV parsing with no network I/O, so the 'prevent indefinite hanging on I/O' rationale for `withTimeout` doesn't apply here the way it does for the URL-based fetches elsewhere in this file. Closing as-is.

test/continuous-test-suite-bugfixes.ts
```diff
@@ -3850,6 +3852,166 @@ exit 127
...
+        { type: "file", data: Buffer.from("x"), name: "a.pdf" },
+        {
+          type: "tool-call",
+          toolCallId: "1",
+          toolName: "t",
+          args: { a: 1 },
+        },
+        { type: "tool-result", toolCallId: "1", toolName: "t", result: 42 },
+      ];
+      return items.length === 5 && items[0].text === "hi";
```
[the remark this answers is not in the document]
murdore: Fixed in c791e9a3315b0eac1deb26b445ee01f260cbbea0: added a negative compile-time case to the "Types #325" test in test/continuous-test-suite-bugfixes.ts — an object with an unsupported key assigned to `MessageContent` under a real `// @ts-expect-error` pragma (not just a plain comment), so a future reintroduction of the removed index signature would surface as an "unused `@ts-expect-error` directive" type error.

test/continuous-test-suite-bugfixes.ts
```diff
@@ -3850,6 +3852,166 @@ exit 127
...
+              o?: { maxSize?: number },
+            ) => Promise<Buffer>;
+          }
+        ).loadFromURL.bind(FileDetector);
+        // GET populates the content-type cache.
+        await load(url, { maxSize: 1024 });
+        // A subsequent detection of the same URL must NOT issue a HEAD.
+        headCount = 0;
+        await FileDetector.detectAndProcess(url).catch(() => undefined);
+        return headCount === 0;
```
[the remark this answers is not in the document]
murdore: Fixed in c791e9a3315b0eac1deb26b445ee01f260cbbea0: the "FileDetector #323" test now serves a real minimal valid 1x1 PNG (`Buffer.from("iVBORw0KGgo...", "base64")`, per your suggested payload) instead of the 4-byte magic-only stub, removed the `.catch(() => undefined)` swallow, and asserts both `result.type === "image"` and `headCount === 0` — so the test now actually proves the cached path completes successfully, not just that an early failure happens to leave `headCount` at 0.

src/lib/utils/imageProcessor.ts
murdore: Verified against current HEAD (c791e9a3315b0eac1deb26b445ee01f260cbbea0), no code change needed: `hasStrictImageMagic()` (imageProcessor.ts ~136-166) is a strict magic-byte check, and `validateBufferNotEmpty()`'s per-format minimum-size check only fires when `minForFormat !== undefined && content.length < minForFormat && this.hasStrictImageMagic(content, mediaType)` — i.e. the "truncated" error is only thrown when the buffer's bytes actually strictly match that format's magic numbers, never on `detectImageType()`'s jpeg fallback for an unrecognized buffer. This is exactly the fix you're asking for, and it's already present (per the #293 review, documented directly in the method's own docstring at lines 130-136). The existing "ImageProcessor #293" regression test in test/continuous-test-suite-bugfixes.ts covers this directly (a BMP-magic buffer that would otherwise jpeg-fallback is asserted to NOT be rejected as a truncated jpeg).

src/lib/utils/fileDetector.ts
murdore: Fixed in c791e9a3315b0eac1deb26b445ee01f260cbbea0: `getCachedUrlContentType()` now re-inserts (delete + set) the entry on a cache hit before returning it, bumping it to the end of the `Map`'s insertion-order iteration. Combined with the existing size-bounded eviction in `setCachedUrlContentType()` (which deletes the first/oldest key via `urlContentTypeCache.keys().next().value`), this turns the previous FIFO cache into an actual LRU using only the `Map`'s native ordering — no new data structure needed.

src/lib/core/modules/MessageBuilder.ts
murdore: Fixed in c791e9a3315b0eac1deb26b445ee01f260cbbea0 — implemented the full fix (option a from my own fix_plan): (1) `audioFiles` is now a typed field on both `GenerateOptions.input` (src/lib/types/generate.ts) and `StreamOptions.input` (src/lib/types/stream.ts), mirroring the existing `videoFiles` field; (2) both `buildMessages()` and `buildMessagesForStream()` in src/lib/core/modules/MessageBuilder.ts now forward `audioFiles`/`videoFiles` into the `multimodalOptions.input` object passed to `buildMultimodalMessagesArray`; (3) `buildMultimodalMessagesArray()` in src/lib/utils/messageBuilder.ts merges `inp.audioFiles`/`inp.videoFiles` into `inp.files` right after normalizing `options.input`, routing them through the existing auto-detecting `files` pipeline (`processUnifiedFilesArray` → `appendDetectedFileResult`), which already fully implements "video"/"audio" result handling (text description + extracted keyframes/cover art) — no new media-parsing logic was needed, this was pure wiring. Added a deterministic, network-free regression test "MessageBuilder #284: audioFiles/videoFiles content actually reaches the built message" using magic-bytes-only buffers (RIFF/WAVE, ISO-BMFF ftyp), proving the `"## Video File:"` / `"## Audio File:"` markers that `appendDetectedFileResult()` emits actually appear in the built message content — not just that the routing flag flips, which is all the prior #284 test proved.

src/lib/utils/messageBuilder.ts
murdore: Fixed in c791e9a3315b0eac1deb26b445ee01f260cbbea0: `appendCsvContentToText()`'s parameter type is now `CSVContent[]` (imported from the types barrel, `../types/index.js`), and both call sites now use `content.filter(isCSVContent)` / `inp.content?.filter(isCSVContent)` (the existing canonical type guard from multimodal.ts) instead of an inline `.filter((c) => c.type === "csv")`. `formatStyle` is therefore already correctly typed as `"raw" | "markdown" | "json" | undefined` end-to-end, and the unsafe `as` cast has been removed entirely.

src/lib/types/multimodal.ts
Tara-ag, line 508: 💡 **MINOR**: The tightened `MessageContent` type is still missing `providerOptions`, which the message-builder conversion reads from individual content items (`item.providerOptions` in `src/lib/core/modules/MessageBuilder.ts`). It also omits `mediaType`, which AI-SDK `file` parts and Anthropic image blocks rely on (already noted by other reviewers). 

Since the PR description claims this list enumerates the concrete fields the codebase reads/writes, please add at least:

```ts
/** Provider-specific per-block options (e.g. Anthropic cache_control). */
providerOptions?: Record<string, unknown>;
/** MIME type used by AI-SDK `file` parts and some image paths. */
mediaType?: string;
```

This keeps the type honest and avoids forcing casts in the builder/adapter paths that already access these fields.

src/lib/utils/messageBuilder.ts
```diff
@@ -1559,6 +1591,44 @@ export async function buildMultimodalMessagesArray(
...
+          ? Buffer.from(raw, "base64")
+          : Buffer.from(raw, "utf-8")
+        : raw;
+    const result = await FileDetector.detectAndProcess(buffer, {
+      allowedTypes: ["csv"],
+      csvOptions: {
+        maxRows: csv.metadata?.maxRows,
+        formatStyle: csv.metadata?.formatStyle,
+      },
+    });
```
Tara-ag: ⚠️ **MAJOR**: `appendCsvContentToText()` awaits `FileDetector.detectAndProcess(buffer, { allowedTypes: ["csv"], ... })` without wrapping it in `withTimeout`. The project guidelines require asynchronous calls (especially file/CSV processing) to be wrapped with `withTimeout` so a stalled or hung detector cannot block the request indefinitely.

Fix:

```ts
import { withTimeout } from "./errorHandling.js";

const result = await withTimeout(
  FileDetector.detectAndProcess(buffer, {
    allowedTypes: ["csv"],
    csvOptions: {
      maxRows: csv.metadata?.maxRows,
      formatStyle: csv.metadata?.formatStyle,
    },
  }),
  30_000, // or the project's default file-processing timeout
  new Error(`Timed out processing CSV content "${name}"`),
);
```

This applies to both call sites added in this PR (the no-vision gate and `convertContentToProviderFormat`).

src/lib/utils/messageBuilder.ts
```diff
@@ -1077,9 +1079,15 @@ export async function processUnifiedFilesArray(
...
+          // #273: don't silently drop a failed file — log, then throw so the
+          // caller learns the file couldn't be processed (matches the explicit
+          // pdf/csv paths' fail-loud behavior).
           logger.error(
-            `[NEUROLINK] File skipped/failed: ${filename} — reason: ${errMsg}`,
+            `[NEUROLINK] File processing failed: ${filename} — reason: ${errMsg}`,
           );
+          throw new Error(`Failed to process file "${filename}": ${errMsg}`, {
+            cause: error,
+          });
```
Tara-ag: ⚠️ **MAJOR**: The new fail-loud errors in `processUnifiedFilesArray` and `processExplicitCsvFiles` throw raw `Error` objects. CLAUDE.md Common Patterns states to use `ErrorFactory` for typed errors so callers can rely on the typed error contract and downstream `formatProviderError` paths can classify failures correctly.

Replace the two new `throw new Error(...)` calls with the appropriate `ErrorFactory` method (e.g. `ErrorFactory.fileProcessingFailed(filename, errMsg)` if it exists, or add a suitable factory method). Preserve the original cause when constructing the error.

src/lib/utils/imageProcessor.ts
Tara-ag, line 250: ⚠️ **MAJOR**: The validation errors thrown by `validateBufferNotEmpty()` are raw `Error` objects. CLAUDE.md Common Patterns requires typed errors via `ErrorFactory` so callers can distinguish image-validation failures from generic exceptions and so provider error formatters can classify them.

Replace the three `throw new Error(...)` calls in this method with an `ErrorFactory.imageProcessingFailed(...)` (or equivalent) typed error, preserving the existing messages and conditions.

src/lib/utils/fileDetector.ts
```diff
@@ -53,6 +54,58 @@ import { PDFProcessor } from "./pdfProcessor.js";
...
```
Tara-ag, line 80: 💡 **MINOR**: The URL Content-Type cache is a module-level mutable `Map` with a fixed 512-entry cap. While the re-insert-on-hit improves recency, the cache is still shared across all requests in the process and is never invalidated on TTL expiry beyond per-key deletion. 

Consider whether a long-lived SDK instance could retain stale Content-Type entries for URLs whose responses change (e.g. signed URLs with the same path but different content). For correctness, the TTL already bounds staleness to 60s, which is reasonable; just ensure this is documented as an intentional trade-off rather than a guarantee of freshness.

src/lib/core/modules/MessageBuilder.ts
Tara-ag, line 60: 💡 **MINOR**: The `detectMultimodal()` change now treats `audioFiles`/`videoFiles` as multimodal, and the builder forwards them into `buildMultimodalMessagesArray`. However, `buildMultimodalMessagesArray` then folds them into `inp.files` for auto-detection. That means an audio/video-only request will be classified as multimodal even though the actual content is ultimately injected as text markers (`## Audio File:` / `## Video File:`). 

This is functionally correct for the current FileDetector fallback, but it slightly overloads the meaning of "multimodal" (no vision model is actually required). Please confirm in a code comment that this is intentional — audio/video are routed through the multimodal builder only to ensure they are not dropped, not because they require vision.

test/continuous-test-suite-bugfixes.ts
```diff
@@ -3850,6 +3852,287 @@ exit 127
...
+      // erroring and the unused `@ts-expect-error` directive fails
+      // `pnpm run check`, catching the regression.
+      // @ts-expect-error -- MessageContent must reject unsupported fields
+      const invalidItem: MessageContentT = {
+        type: "text",
+        text: "hi",
+        unsupported: true,
+      };
+      void invalidItem;
+      return items.length === 5 && items[0].text === "hi";
```
Tara-ag: 💡 **MINOR**: The `MessageContent` compile-time test uses `@ts-expect-error` on an object with an `unsupported: true` key. This is a good regression guard, but the test only runs at `pnpm run check` time; the runtime suite itself does not exercise the type error. 

That is acceptable for a type-level assertion, but consider adding a small runtime assertion that the valid items still pass through `convertContentItem` / `convertToModelMessages` without being stripped, so the test also guards against future runtime filtering changes.

Tara-ag (changes requested):
## Autonomous Review Summary

**Files reviewed:** 8  
**New issues raised this run:** 6 (3 MAJOR, 3 MINOR)

### MAJOR (blocking)
1. **`src/lib/utils/messageBuilder.ts:1625`** — `appendCsvContentToText()` awaits `FileDetector.detectAndProcess` without `withTimeout`; async file operations must be timeout-bounded per project guidelines.
2. **`src/lib/utils/messageBuilder.ts:1090`** — New fail-loud file/CSV errors throw raw `Error` objects instead of `ErrorFactory` typed errors (CLAUDE.md Common Patterns).
3. **`src/lib/utils/imageProcessor.ts:209`** — Image validation errors throw raw `Error` objects instead of `ErrorFactory` typed errors.

### MINOR
4. **`src/lib/types/multimodal.ts:508`** — Tightened `MessageContent` still omits `providerOptions` and `mediaType`, which the builder and adapters already read.
5. **`src/lib/utils/fileDetector.ts:66`** — URL Content-Type cache is a module-level mutable `Map`; worth documenting the 60s freshness trade-off.
6. **`src/lib/core/modules/MessageBuilder.ts:51`** — Audio/video-only requests are now routed as "multimodal" but end up as text markers; clarify intent in a comment.
7. **`test/continuous-test-suite-bugfixes.ts:4087`** — Type-level `@ts-expect-error` guard is good; consider a small runtime assertion through the conversion path.

### Notes
- No hardcoded secrets or critical security vulnerabilities were found.
- Existing review comments from Copilot/CodeRabbit/Tara-ag were checked and not duplicated.
- The PR fixes real bugs (#273, #284, #289, #293, #323, #325) and adds deterministic tests; once the MAJOR items above are addressed, this should be good to approve.

**Next step:** Address the 3 MAJOR issues (timeout wrapping + typed errors) and re-request review.

src/lib/types/multimodal.ts
murdore: Fixed. Added `providerOptions?: Record<string, unknown>` to `MessageContent` (src/lib/types/multimodal.ts) — confirmed it is read via `item.providerOptions` in MessageBuilder.ts's coreMessages conversion (lines 175-211, 337-360). Did NOT add `mediaType`: I traced every consumer of `MessageContent`-typed values across the codebase and found `mediaType` is only ever *written* as an output field (by `convertContentItem`/provider adapters producing AI-SDK `ImagePart`/`FilePart`), never *read* from anything typed or sourced as `MessageContent`. Adding an unused field would reintroduce the inaccuracy this PR set out to fix. Happy to add it if a real read site turns up.

src/lib/utils/messageBuilder.ts
murdore: Fixed. `appendCsvContentToText()` now wraps `FileDetector.detectAndProcess(...)` in `withTimeout` (imported from `./errorHandling.js`, not the timeout.js variant), with a new `CSV_CONTENT_DETECTION_TIMEOUT_MS = 30_000` constant and a timeout error naming the CSV file. This is the shared helper used by both the no-vision text path and the multimodal converter, so both call sites get the guard.

src/lib/utils/messageBuilder.ts
murdore: Fixed. Both raw-`Error` throws in `processUnifiedFilesArray` and `processExplicitCsvFiles` now use new `ErrorFactory.fileProcessingFailed(filename, originalError)` / `ErrorFactory.csvProcessingFailed(filename, originalError)` methods (added to src/lib/utils/errorHandling.ts, with new `FILE_PROCESSING_FAILED`/`CSV_PROCESSING_FAILED` error codes). They preserve the original error via the existing `originalError` constructor param, which is this codebase's established cause-preservation idiom for `NeuroLinkError` (copies `.stack`, sets `context.originalMessage`).

src/lib/utils/imageProcessor.ts
murdore: Fixed. All three raw `Error` throws in `validateBufferNotEmpty` now use a new `ErrorFactory.imageBufferInvalid(message)` method (added to src/lib/utils/errorHandling.ts, with a new `IMAGE_BUFFER_INVALID` error code), preserving the exact original messages (empty buffer, too-small buffer, truncated-format-specific buffer).

src/lib/utils/fileDetector.ts
murdore: Addressed. Added a named `URL_CONTENT_TYPE_CACHE_MAX_SIZE` constant (replacing the magic 512), an explanatory trade-off comment documenting the 60s staleness bound as an intentional, bounded choice rather than a freshness guarantee, and a new `pruneExpiredUrlContentTypeEntries()` sweep that runs once the cache hits its size cap — before falling back to FIFO eviction of the oldest live entry — so an entry cached once and never re-queried doesn't linger purely on lazy per-key expiry. No dedicated new test: the cache internals are module-private and the real HEAD call site's `now = Date.now()` isn't test-injectable without mocking; the existing `FileDetector #323` cache test continues to pass unchanged.

src/lib/core/modules/MessageBuilder.ts
murdore: Confirmed intentional and documented. Expanded the JSDoc above `detectMultimodal` in src/lib/core/modules/MessageBuilder.ts to explicitly state that routing audio/video through `isMultimodal` does not imply vision is required — `buildMultimodalMessagesArray` folds `audioFiles`/`videoFiles` into `inp.files` for auto-detection, and `appendDetectedFileResult` injects their content as plain text markers (`## Audio File:` / `## Video File:`), not image/vision blocks — and points to the existing `MessageBuilder #284` tests in test/continuous-test-suite-bugfixes.ts as coverage of this behavior.

test/continuous-test-suite-bugfixes.ts
murdore: Added. Extended the `Types #325` test in test/continuous-test-suite-bugfixes.ts with a runtime assertion via `convertToModelMessages`/`convertContentItem` (added `mimeType: "application/pdf"` to the file item so it survives the file-branch's `Buffer.isBuffer(data) && typeof mimeType === "string"` validation), proving the text/image/file fields actually survive conversion unstripped, in addition to the existing `@ts-expect-error` compile-time guard.

Tara-ag (approved):
## Review Summary

**Files reviewed:** 9 changed files across `src/lib/core/modules/MessageBuilder.ts`, `src/lib/types/{generate,multimodal,stream}.ts`, `src/lib/utils/{errorHandling,fileDetector,imageProcessor,messageBuilder}.ts`, and `test/continuous-test-suite-bugfixes.ts`.

**New issues raised in this review:** 0

**Existing review comments:** All prior inline review threads are resolved. The author addressed the substantive feedback from earlier rounds (CSV base64 handling, `MessageContent` field completeness, `withTimeout` on the HEAD path, image-format strictness, audio/video forwarding, and ErrorFactory adoption for the new error paths).

**Blocking criteria check:** None of the PR's changes meet the blocking thresholds:
- No hardcoded secrets or critical security vulnerabilities introduced.
- No breaking changes to the public SDK API surface: `audioFiles` additions are additive; `MessageContent` tightening is internal-focused, verified by the existing build/tests, and guarded by a negative `@ts-expect-error` regression test.
- No violations of non-negotiable CLAUDE.md Critical Rules enforced outside CI (no static provider imports, no Gemini tools+schema conflict, no `formatProviderError` throwing).
- No new MAJOR-severity issues.

**Verification noted:** The PR includes deterministic regression tests for #273, #284, #289, #293, #323, and #325, and the author reports `pnpm run check`, `pnpm run lint`, and `pnpm run build` are clean.

**Approve.**

src/lib/utils/messageBuilder.ts
```diff
@@ -1080,8 +1081,15 @@ export async function processUnifiedFilesArray(
           );
         } catch (error) {
           const errMsg = error instanceof Error ? error.message : String(error);
+          // #273: don't silently drop a failed file — log, then throw so the
+          // caller learns the file couldn't be processed (matches the explicit
+          // pdf/csv paths' fail-loud behavior).
           logger.error(
-            `[NEUROLINK] File skipped/failed: ${filename} — reason: ${errMsg}`,
+            `[NEUROLINK] File processing failed: ${filename} — reason: ${errMsg}`,
```
Tara-ag, line 1088: ⚠️ **MAJOR**: The fail-loud change in `processUnifiedFilesArray` alters the public SDK contract. Previously a failed file was logged and skipped, allowing the request to continue with partial results; now it throws and fails the entire request. The same applies to `processExplicitCsvFiles` at line 1192.

This is a backward-compatibility concern for existing callers who may have been passing mixed-quality file lists and relying on the silent-skip behavior. CLAUDE.md Critical Rule 5 says the public SDK API must not break existing callers without a documented migration path.

**Suggestion**: Either (a) gate the throw behind an opt-in flag such as `input.strictFileHandling` so existing callers keep the old behavior, or (b) add an explicit "Breaking change" note to the PR description / CHANGELOG with a migration example showing how callers should pre-validate files if they need the old skip semantics. If this is treated as a bugfix, document why silent skipping is considered a bug and how callers should adjust.

src/lib/types/multimodal.ts
```diff
@@ -487,15 +487,46 @@ export type MultimodalInput = {
...
```
Tara-ag, line 508: ⚠️ **MAJOR**: The tightened `MessageContent` type is a potential breaking change for external consumers. Removing the `[key: string]: unknown` index signature means any caller who was passing extra fields (e.g. `metadata`, custom provider fields, or future AI-SDK part shapes) will now get a TypeScript error. The PR description says "the build confirmed no consumer relied on an un-enumerated key," but that only checked the internal codebase, not external SDK users.

CLAUDE.md Critical Rule 5 requires the public SDK API to stay backward compatible. Since `MessageContent` is exported from the public types barrel, this is a public-facing type change.

**Suggestion**: If the goal is to tighten internal safety without breaking external callers, keep the index signature but narrow its value type to something safer such as `[key: string]: string | Buffer | Record<string, unknown> | unknown` (or use `@deprecated` + a comment discouraging new keys), and add the explicitly-typed fields on top. Alternatively, document this as a breaking change with a migration path for users who previously relied on the loose signature.

src/lib/utils/fileDetector.ts
```diff
@@ -55,6 +56,91 @@ import { PDFProcessor } from "./pdfProcessor.js";
...
+ * Expired entries are removed lazily on their next `get()` (see
+ * `getCachedUrlContentType`); `setCachedUrlContentType` additionally sweeps
+ * expired entries opportunistically once the cache hits its size cap, so a
+ * URL that is cached once and never looked up again doesn't linger until the
+ * FIFO eviction below forces it out.
+ */
+const URL_CONTENT_TYPE_TTL_MS = 60_000;
+const URL_CONTENT_TYPE_CACHE_MAX_SIZE = 512;
+const urlContentTypeCache = new Map<
+  string,
```
Tara-ag, line 80: 💡 **MINOR**: The URL Content-Type cache uses the full URL string (including query parameters) as the cache key. If callers pass presigned URLs or URLs containing tokens/credentials in query params, those sensitive values become part of the in-memory cache key for up to 60 seconds. While the cache only stores the `Content-Type` and not the response body, and is bounded to 512 entries, this is still a latent secret-surface concern.

**Suggestion**: Consider normalizing the cache key to the URL origin + pathname (stripping query/search params) when the URL is not expected to vary by query string for content-type purposes. If query params legitimately affect the response type, document that the cache intentionally keys on the full URL and that callers should avoid embedding secrets in URLs passed to the SDK.

src/lib/utils/imageProcessor.ts
```diff
@@ -19,6 +19,25 @@ import { getImageCache } from "./imageCache.js";
...
+/**
+ * Minimum bytes required just to detect an image format from magic bytes (#293).
+ */
+export const MIN_IMAGE_SIZE = 12;
+
+/**
+ * Minimum plausible byte size for a non-empty image of each format (#293). A
+ * buffer that carries a format's magic bytes but is smaller than this is
+ * truncated/corrupt and would fail (or render blank) downstream — reject early
+ * with a clear message instead.
```
Tara-ag, line 31: 💡 **MINOR**: The `MIN_VALID_IMAGE_SIZE` thresholds (67, 125, 43, 20, 100 bytes) are undocumented magic numbers. They appear to be derived from the test fixtures rather than from image-format specifications, and there is no explanation for why these particular values represent the minimum "plausibly-complete" file for each format. A 1×1 PNG can indeed be ~67 bytes, but a JPEG of the same dimensions is typically larger than 125 bytes, while a minimal valid GIF can be smaller than 43 bytes depending on color table choices.

**Suggestion**: Add a comment citing how each threshold was chosen (e.g. "smallest valid file produced by ImageMagick/known encoder"), or derive/document them from format specs. Otherwise future maintainers may tweak them blindly and break legitimate tiny images.

Tara-ag (changes requested):
## Review Summary

Reviewed 9 changed files. I raised **4 new issues** not already covered by existing review threads:

| Severity | Count | Files |
|----------|-------|-------|
| 🔒 CRITICAL | 0 | — |
| ⚠️ MAJOR | 2 | `src/lib/utils/messageBuilder.ts`, `src/lib/types/multimodal.ts` |
| 💡 MINOR | 2 | `src/lib/utils/fileDetector.ts`, `src/lib/utils/imageProcessor.ts` |

### Blocking concerns

1. **Backward compatibility — fail-loud file/CSV errors** (`src/lib/utils/messageBuilder.ts`): `processUnifiedFilesArray` and `processExplicitCsvFiles` previously logged and skipped failed files; they now throw and fail the entire request. This changes the public SDK contract for existing callers. Per CLAUDE.md Critical Rule 5, the public SDK API must not break existing callers without a documented migration path. Please either gate this behind an opt-in flag or document it as a breaking change with migration guidance.

2. **Backward compatibility — tightened `MessageContent` type** (`src/lib/types/multimodal.ts`): Removing the `[key: string]: unknown` index signature is a TypeScript-breaking change for external SDK consumers who may pass extra fields. Again per CLAUDE.md Rule 5, please either retain a narrowed index signature or document this as a breaking change with a migration path.

### Non-blocking suggestions

- Consider normalizing the URL Content-Type cache key to avoid retaining query parameters that may contain secrets.
- Document the basis for the `MIN_VALID_IMAGE_SIZE` thresholds so future maintainers don't tweak them blindly.

### Next steps

Please address the two backward-compatibility concerns above (or justify why they are acceptable breaking changes with migration docs). Once those are resolved, the remaining issues can be addressed as follow-ups or accepted as-is.

PR #1210, opened 2026-07-19
feat(proxy): add deterministic request replay

murdore:
## Summary
- add `neurolink proxy replay export` to reconstruct one captured request and exact upstream attempt from local debug indexes and compressed artifacts
- add explicit `neurolink proxy replay compare` execution with environment-backed credentials and deterministic redacted comparison output
- persist the exact upstream method and URL at every Anthropic request site, including auth retry attempts

## Safety
- verifies artifact containment, metadata, SHA-256, compressed/decompressed and bundle size limits, truncation, and missing phases
- writes deterministic output atomically with `0600` permissions and refuses symlink outputs
- requires `--execute`, rejects missing/redacted request evidence, tampered or unredacted captures, invalid header names, unsafe URLs, redirects, and implicit nonstandard endpoints
- bounds, cancels on timeout, and redacts direct responses using the existing proxy logging rules
- preserves existing full-response logging behavior and does not change or contact the live proxy

## Validation
- `pnpm exec vitest run test/proxyReplay.test.ts` (20/20)
- repository Vitest suite excluding standalone env guard (335 tests; one pre-existing 5-second `modelPool` timing flake reproduced independently)
- `pnpm run test:bugfixes` (101/101)
- `pnpm run test:envguard` (118/118)
- `pnpm run test:log-sanitize` (49/49)
- `pnpm run test:ssrf` (42/42)
- `pnpm run proxy:performance` (all budgets passed; zero drops/failures)
- `pnpm run check`, `pnpm run lint`, `pnpm run build:cli`
- full pre-commit validation, Gitleaks/security scan, production/package/browser builds, and `publint`

### DIFF
--- a/src/cli/commands/proxyReplay.ts
+++ b/src/cli/commands/proxyReplay.ts
@@ -1,0 +1,212 @@
+import { readFile, stat } from "node:fs/promises";
+import { resolve } from "node:path";
+import type { Argv, CommandModule } from "yargs";
+import chalk from "chalk";
+import {
+  compareProxyReplayBundle,
+  exportProxyReplayBundle,
+  isValidProxyReplayHeaderName,
+  readProxyReplayBundle,
+  writeProxyReplayDocument,
+} from "../../lib/proxy/proxyReplay.js";
+import type { ProxyReplayArgs } from "../../lib/types/index.js";
+import { logger } from "../../lib/utils/logger.js";
+
+const ACTIONS = ["export", "compare"] as const;
+const MAX_BODY_OVERRIDE_BYTES = 16 * 1024 * 1024;
+
+function requiredValue(value: string | undefined, option: string): string {
+  if (!value) {
+    throw new Error(`${option} is required for this action`);
+  }
+  return value;
+}
+
+function resolveHeaderValues(mappings: string[] | undefined) {
+  const values: Record<string, string> = {};
+  for (const mapping of mappings ?? []) {
+    const separator = mapping.indexOf("=");
+    if (separator <= 0 || separator === mapping.length - 1) {
+      throw new Error(
+        `Invalid --header-env value "${mapping}". Use header-name=ENV_VAR.`,
+      );
+    }
+    const header = mapping.slice(0, separator).trim().toLowerCase();
+    if (!isValidProxyReplayHeaderName(header)) {
+      throw new Error(`Invalid header name in --header-env: ${header}`);
+    }
+    const environmentName = mapping.slice(separator + 1).trim();
+    const value = process.env[environmentName];
+    if (!value) {
+      throw new Error(
+        `Environment variable ${environmentName} is empty or unset for header ${header}`,
+      );
+    }
+    values[header] = value;
+  }
+  return values;
+}
+
+async function readBodyOverride(filePath: string | undefined) {
+  if (!filePath) {
+    return undefined;
+  }
+  const resolvedPath = resolve(filePath);
+  const info = await stat(resolvedPath);
+  if (!info.isFile() || info.size > MAX_BODY_OVERRIDE_BYTES) {
+    throw new Error(
+      `Body override must be a regular file no larger than ${MAX_BODY_OVERRIDE_BYTES} bytes`,
+    );
+  }
+  return readFile(resolvedPath, "utf8");
+}
+
+function printResult(
+  format: "text" | "json",
+  value: Record<string, unknown>,
+): void {
+  logger.always(
+    format === "json"
+      ? JSON.stringify(value)
+      : Object.entries(value)
+          .map(([key, child]) => `${key}: ${String(child)}`)
+          .join("\n"),
+  );
+}
+
+export const proxyReplayCommand: CommandModule<object, ProxyReplayArgs> = {
+  command: "replay <action>",
+  describe: "Export captured proxy requests or compare them directly upstream",
+  builder: (yargs: Argv) =>
+    yargs
+      .positional("action", {
+        type: "string",
+        choices: [...ACTIONS],
+        describe: "Replay action",
+      })
+      .option("request-id", {
+        type: "string",
+        alias: "requestId",
+        description: "Captured proxy request ID to export",
+      })
+      .option("attempt", {
+        type: "number",
+        description: "Specific upstream attempt to reconstruct",
+      })
+      .option("logs-dir", {
+        type: "string",
+        alias: "logsDir",
+        description: "Proxy log directory",
+      })
+      .option("bundle", {
+        type: "string",
+        description: "Replay bundle to compare directly upstream",
+      })
+      .option("output", {
+        type: "string",
+        alias: "o",
+        description: "Destination JSON file",
+      })
+      .option("execute", {
+        type: "boolean",
+        default: false,
+        description: "Explicitly permit the direct upstream request",
+      })
+      .option("header-env", {
+        type: "string",
+        array: true,
+        alias: "headerEnv",
+        description:
+          "Inject a redacted header from an environment variable (header=ENV_VAR)",
+      })
+      .option("body-file", {
+        type: "string",
+        alias: "bodyFile",
+        description:
+          "Complete request body override for redacted/truncated captures",
+      })
+      .option("url", {
+        type: "string",
+        description: "Explicit direct endpoint override",
+      })
+      .option("timeout-ms", {
+        type: "number",
+        alias: "timeoutMs",
+        default: 120_000,
+        description: "Direct request timeout",
+      })
+      .option("format", {
+        type: "string",
+        choices: ["text", "json"] as const,
+        default: "text" as const,
+        description: "Command summary format",
+      })
+      .option("quiet", {
+        type: "boolean",
+        alias: "q",
+        default: false,
+        description: "Suppress non-essential output",
+      })
+      .example(
+        "neurolink proxy replay export --request-id req-123 --output replay.json",
+        "Create a deterministic offline replay bundle",
+      )
+      .example(
+        "neurolink proxy replay compare --bundle replay.json --output comparison.json --execute --header-env authorization=ANTHROPIC_AUTHORIZATION",
+        "Execute the captured upstream request with an environment-backed credential",
+      ) as Argv<ProxyReplayArgs>,
+  handler: async (argv) => {
+    try {
+      const output = requiredValue(argv.output, "--output");
+      if (argv.action === "export") {
+        const requestId = requiredValue(argv.requestId, "--request-id");
+        const bundle = await exportProxyReplayBundle({
+          requestId,
+          logsDir: argv.logsDir,
+          attempt: argv.attempt,
+        });
+        const written = await writeProxyReplayDocument(output, bundle);
+        if (!argv.quiet) {
+          printResult(argv.format ?? "text", {
+            action: "export",
+            output: written.path,
+            sha256: written.sha256,
+            captures: bundle.completeness.captures,
+            replayable: bundle.completeness.replayable,
+            blockers: bundle.completeness.blockers.join(",") || "none",
+          });
+        }
+        return;
+      }
+
+      const bundlePath = requiredValue(argv.bundle, "--bundle");
+      const bundle = await readProxyReplayBundle(bundlePath);
+      const comparison = await compareProxyReplayBundle(bundle, {
+        execute: argv.execute === true,
+        headerValues: resolveHeaderValues(argv.headerEnv),
+        bodyOverride: await readBodyOverride(argv.bodyFile),
+        urlOverride: argv.url,
+        timeoutMs: argv.timeoutMs,
+      });
+      const written = await writeProxyReplayDocument(output, comparison);
+      if (!argv.quiet) {
+        printResult(argv.format ?? "text", {
+          action: "compare",
+          output: written.path,
+          sha256: written.sha256,
+          status: comparison.direct.status,
+          statusMatches: comparison.comparison.statusMatches,
+          bodyHashMatches: comparison.comparison.bodyHashMatches,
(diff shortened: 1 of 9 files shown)


src/lib/proxy/requestLogger.ts
```diff
@@ -418,6 +418,22 @@ function prepareRedactedBody(body: unknown): {
   return truncateUtf8String(redacted, MAX_CAPTURED_BODY_BYTES);
 }
 
+/** Shared redaction used by offline replay exports and direct comparisons. */
```
Tara-ag, line 421: ⚠️ **MAJOR**: `prepareProxyBodyForLogging` returns `value?: string` but `compareProxyReplayBundle` treats `preparedBody.value` as a guaranteed string (`const redactedBody = preparedBody.value ?? ""`). However, `prepareRedactedBody` returns `{ truncated: false }` when `body` is `undefined`/`null`, so `value` is absent. In the direct comparison path the response body could legitimately be empty (e.g. 204 No Content), and the current fallback to `""` is fine, but the typing mismatch means callers must defensively handle `undefined`. More importantly, this wrapper exposes an internal helper without guarding expensive serialization with `logger.shouldLog("debug")`; while the callers here are CLI paths, any future SDK/hot-path usage could accidentally pay the redaction cost. Consider tightening the return type to `{ value: string; bytes?: number; truncated: boolean } | { value?: undefined; bytes?: undefined; truncated: false }` or documenting that callers must handle `undefined`.

Suggested fix:
```ts
export function prepareProxyBodyForLogging(body: unknown): {
  value?: string;
  bytes?: number;
  truncated: boolean;
} {
  return prepareRedactedBody(body);
}
```
is already the signature; add a JSDoc note: `@returns value is undefined only when input is null/undefined`.

src/lib/proxy/proxyReplay.ts
```diff
@@ -0,0 +1,1004 @@
...
+  assertProxyReplayBundle(parsed);
+  return parsed;
+}
+
+function normalizedContentType(value: string | null): string | null {
+  return value?.split(";", 1)[0]?.trim().toLowerCase() || null;
+}
+
+function jsonShape(body: string): string[] | null {
+  let parsed: unknown;
```
Tara-ag, line 754: 💡 **MINOR**: `resolveHeaderValues` (CLI) does not validate the user-supplied header name against `HEADER_NAME_PATTERN` before injecting it into the direct request. Although `fetch` will likely sanitize, a malformed name such as one containing spaces or control characters could produce confusing errors or unexpected behavior. Consider reusing the existing `HEADER_NAME_PATTERN` test here.

Suggested fix:
```ts
const header = mapping.slice(0, separator).trim().toLowerCase();
if (!HEADER_NAME_PATTERN.test(header)) {
  throw new Error(`Invalid header name in --header-env: ${header}`);
}
```

test/proxyReplay.test.ts
Tara-ag, line 1: 💡 **MINOR / process note**: This suite uses Vitest (`describe`/`it`/expect) rather than the repo's documented `tsx`-based `continuous-test-suite-*.ts` pattern. `CONTRIBUTING.md` states "There is no vitest runner; do not write `*.test.ts` files using `vi.mock` or `describe/it` blocks — they will not be picked up." The PR validation shows `pnpm exec vitest run test/proxyReplay.test.ts` passing, so CI is clearly configured to run it, but the docs/CONTRIBUTING guidance is now inconsistent with practice. Consider either migrating to a `test/continuous-test-suite-proxyReplay.ts` orchestrator or updating CONTRIBUTING.md to reflect that Vitest unit tests are accepted for proxy internals.

src/lib/proxy/proxyReplay.ts
```diff
@@ -0,0 +1,1004 @@
...
+}
+
+function stableJson(value: unknown): string {
+  return JSON.stringify(stableValue(value));
+}
+
+export function serializeProxyReplayDocument(value: unknown): string {
+  return `${JSON.stringify(stableValue(value), null, 2)}\n`;
+}
+
```
Tara-ag, line 293: 💡 **MINOR**: `assertProxyReplayBundle` validates that request headers are redacted, but it does not validate that the *captured response* headers (and body) in `bundle.captures` or `bundle.capturedResponse` are also redacted. Since `exportProxyReplayBundle` builds captures from `logBodyCapture` output which is already redacted, this is mostly safe, but a hand-crafted or tampered bundle could leak secrets in response headers. Consider extending the assertion to verify `redactProxyHeadersForLogging(capture.headers)` is stable-equal for every capture, or document that response-side redaction is assumed from the logger.

src/lib/server/routes/claudeProxyRoutes.ts
```diff
@@ -1417,6 +1417,10 @@ async function handleClaudePassthroughRequest(args: {
     account: "passthrough",
     accountType: "passthrough",
     attempt: 1,
+    metadata: {
+      upstreamMethod: "POST",
```
Tara-ag, line 1421: ✅ Looks correct: passthrough path now persists `upstreamMethod`/`upstreamUrl` metadata so replay export can reconstruct the exact Anthropic endpoint.

Tara-ag (approved):
## Review Summary

Reviewed 9 changed files. Found **1 MAJOR** and **3 MINOR** new issues; no blocking criteria met.

### Issues raised

- **MAJOR** `src/lib/proxy/requestLogger.ts:421` — `prepareProxyBodyForLogging` return type allows `value` to be absent, but `compareProxyReplayBundle` assumes a string fallback. Suggested adding a JSDoc contract so callers handle the `undefined` case correctly.
- **MINOR** `src/lib/proxy/proxyReplay.ts:715` — `--header-env` header names are not validated against `HEADER_NAME_PATTERN` before injection.
- **MINOR** `src/lib/proxy/proxyReplay.ts:260` — `assertProxyReplayBundle` validates request-header redaction but not response/header redaction for captured records.
- **MINOR** `test/proxyReplay.test.ts` — Uses Vitest (`describe`/`it`) while `CONTRIBUTING.md` says tests should be `tsx`-based `continuous-test-suite-*.ts` files. The PR validation runs Vitest, so this is a docs/process inconsistency rather than a code defect.

### Blocking assessment

None of the blocking criteria apply:
- No critical security vulnerabilities (SSRF/injection/unsafe eval/secret leakage). Path traversal is mitigated by `realpath` + `isWithin` checks and `O_NOFOLLOW`; URL validation rejects non-HTTPS/non-loopback and sensitive query params; direct execution requires explicit `--execute`.
- No public SDK API breakage — changes are CLI-only and additive.
- No CLAUDE.md Critical Rule violations (rules 1, 3, 4, 5, 6 are not implicated; rules 2/7–13 are enforced by CI).

### Approval

The implementation is careful and well-tested. The replay bundle correctly reconstructs the Anthropic upstream method/URL, enforces artifact containment, SHA-256 verification, redaction, symlink refusal, and explicit opt-in for live comparison. Approved with the noted minor suggestions.

src/lib/proxy/requestLogger.ts
murdore: Addressed in fbdd185a. The shared helper now documents that `value`/`bytes` are omitted only for nullish input and explicitly warns that serialization is immediate, so it must stay off proxy hot paths unless body processing was requested.

src/lib/proxy/proxyReplay.ts
```diff
@@ -0,0 +1,1004 @@
...
+    },
+  );
+  if (existing?.isSymbolicLink()) {
+    throw new Error(`Refusing to replace symlink output path: ${resolvedPath}`);
+  }
+  const serialized = serializeProxyReplayDocument(document);
+  const temporaryPath = join(
+    outputDirectory,
+    `.${basename(resolvedPath)}.${process.pid}.${randomUUID()}.tmp`,
+  );
```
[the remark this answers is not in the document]
murdore: Addressed in fbdd185a. `writeProxyReplayDocument` now enforces the same 16 MiB UTF-8 limit before creating the temporary output file. A regression test verifies an oversized document is rejected and leaves no destination file.

src/lib/proxy/proxyReplay.ts
murdore: Addressed in fbdd185a. Header-name validation is now a shared helper used both while resolving CLI `--header-env` mappings and inside the comparison API. Malformed names are rejected before fetch, with focused regression coverage.

test/proxyReplay.test.ts
murdore: No change in this proxy-scoped PR. The CONTRIBUTING statement is stale relative to the current repository: package.json defines multiple Vitest scripts, the repository suite currently contains 16 Vitest files, and both the focused test and full Vitest corpus pass in CI. Updating repo-wide contributor guidance should be handled separately from this proxy-only change.

src/lib/proxy/proxyReplay.ts
murdore: Addressed in fbdd185a. Bundle validation now reapplies the shared redaction rules to every capture and the selected captured response, covering both headers and bodies, and also verifies the request body. The regression test injects an unredacted `set-cookie` value and confirms rejection before execution.

src/lib/server/routes/claudeProxyRoutes.ts
murdore: Confirmed. All three Anthropic upstream request sites are covered by the static regression test, including passthrough, routed attempts, and auth retry.

src/cli/commands/proxyReplay.ts
```diff
@@ -0,0 +1,204 @@
...
+        choices: ["text", "json"] as const,
+        default: "text" as const,
+        description: "Command summary format",
+      })
+      .option("quiet", {
+        type: "boolean",
+        alias: "q",
+        default: false,
+        description: "Suppress non-essential output",
+      })
```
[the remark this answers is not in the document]
murdore: Addressed in 1ee5d353. Both export and compare summaries are now guarded by `!argv.quiet`, and a behavioral CLI-handler test verifies quiet export still writes the artifact without calling `logger.always`.

murdore:
Reviewed the CodeRabbit `ErrorFactory` / `withTimeout` nitpick against the current helpers; no code change is appropriate. All three shared `withTimeout` implementations only race/reject and do not cancel the underlying operation (one documents that it leaves a bounded ghost). Replacing `AbortSignal.timeout(timeoutMs)` would therefore let the direct upstream fetch continue after the CLI reports a timeout. The abort signal is intentionally retained because it cancels the actual network request.

`ErrorFactory` has no proxy-replay validation error; the available tool/worker factories change the message, category, and retriable semantics. Plain `Error` preserves the CLI validation contract until a dedicated proxy replay error family exists. Focused timeout/guard behavior and the full suite remain green.

Tara-ag (approved):
## Review Summary

Reviewed the current diff for PR #1210 (deterministic proxy request replay). I did not raise any new inline comments this run because the previously identified issues have already been addressed by the author and no additional blocking concerns remain.

### Previously raised concerns (all resolved or acknowledged)
- `prepareProxyBodyForLogging` return-type contract — addressed with JSDoc warning about hot-path usage.
- `writeProxyReplayDocument` bundle size limit — now enforces the same 16 MiB UTF-8 limit as `readProxyReplayBundle`.
- `--header-env` header-name validation — now uses the shared `HEADER_NAME_PATTERN` helper.
- `assertProxyReplayBundle` response-side redaction — now re-applies redaction rules to captures and the captured response.
- `--quiet` flag being ignored — now guards both `printResult` calls.
- Vitest vs. `tsx` test style — acknowledged as a repo-wide docs inconsistency, out of scope for this proxy-only change.

### Blocking assessment
No blocking criteria apply:
- No critical security vulnerabilities: path traversal is mitigated by `realpath` + `isWithin` + `O_NOFOLLOW`; URL validation rejects non-HTTPS/non-loopback endpoints and sensitive query params; direct execution requires explicit `--execute`; output is redacted and written with `0o600`.
- No public SDK API breakage — the change is CLI-only and additive.
- No CLAUDE.md Critical Rule violations (rules 1, 3, 4, 5, 6 are not implicated; rules 2/7–13 are enforced by CI).

### Approval
The replay bundle correctly reconstructs Anthropic upstream method/URL metadata, enforces artifact containment, SHA-256 verification, redaction, symlink refusal, and explicit opt-in for live comparison. The implementation is careful and well-tested. **Approved.**

Tara-ag (approved):
## Review Summary

**Files reviewed:** 9 changed files  
**New issues raised:** 0  
**Decision:** Approve

I walked through the full diff and cross-checked it against the existing review threads. All previously raised points have been addressed or explicitly declined by the author, and I did not identify any new blocking issues.

### Verification against project standards

- **Security:** No hardcoded secrets. Direct comparison redacts request/response headers and bodies via `redactProxyHeadersForLogging` / `prepareProxyBodyForLogging`. Endpoint validation rejects credentials in URLs, sensitive query parameters, non-HTTPS/non-loopback targets, and untrusted endpoints without an explicit override. Artifact loading uses `O_NOFOLLOW` and realpath-based containment checks.
- **CLAUDE.md Critical Rules:** This change is CLI/proxy-scoped; no provider registry, Gemini, SDK/CLI leakage, public SDK breakage, or `formatProviderError` concerns apply.
- **Type safety:** No new `any`, `@ts-ignore`, or unsafe non-null assertions. Type guards and assertion functions are appropriately used.
- **Correctness:** Explicit `execute` opt-in, missing-header validation, body-hash verification, bounded response reading, and atomic `0600` output writing are all in place.

### Existing review threads

All prior inline comments (typing note on `prepareProxyBodyForLogging`, bundle write-size limit, `--header-env` validation, response-side redaction assertion, `--quiet` handling, and the Vitest-vs-`continuous-test-suite` process note) were already resolved or answered. I did not duplicate any of them.

The implementation is solid and safe to ship.

PR #1211, opened 2026-07-19
fix(multimodal): close post-merge review gaps (batch parity, pdf limits, url-secret redaction, supervisor identity)

murdore:
## Summary

After the 13-PR multimodal/CLI/PDF/proxy batch merged into `release`, an audit of the review threads that were still **unresolved at merge** (re-review comments from Yama/CodeRabbit/Copilot on the rebased commits) surfaced **12 verified gaps still present in released code** — adversarially verified against `release`, not just taken at face value. This PR closes all of them, plus the docs/nits, in one consolidated fix.

## Fixes

**Backward-compatibility (documented, not reverted)** — 3 breaking changes shipped in 9.94.x–9.95.x are now documented in **`docs/MIGRATION.md`** with before/after + migration steps (see `BREAKING CHANGE:` footer):
1. multimodal file/CSV processing is fail-loud (throws) instead of log-and-skip partial success;
2. `MessageContent` dropped its open index signature (closed shape);
3. `BatchCommandArgs.file` → `.promptsFile`.

**CLI `batch` parity** — `neurolink batch` previously dropped several inputs the `generate`/`stream` paths honor:
- `--pdf-password` / `NEUROLINK_PDF_PASSWORD` was silently ignored → encrypted PDFs failed. Now wired via `resolvePdfPassword`.
- `--csv-max-rows` wasn't fail-fast validated; `--csv-encoding` / `--csv-sanitize-names` / `--csv-name-case` / `--csv-parse-timeout-ms` were ignored → now built via the shared `buildCsvOptionsFromArgv`.
- `--pdf` soft-warning fired at 100 MB while docs promise 50 MB → dedicated `PDF_SOFT_MAX_MB = 50`.

**PDF hardening**
- Aggregate page-limit summed `pageCount ?? 0`, silently undercounting when a PDF's page count is unknown → now WARNs (enforcement still runs on the known sum; unknown-count PDFs aren't failed outright).
- `getAccuratePageCount` never cleared its `Promise.race` timeout (dangling 5 s handle) and a throwing `pdf.destroy()` in `finally` discarded a valid count → timer is cleared/unref'd and destroy is guarded.

**URL/secret handling**
- Cache-hit debug log leaked a raw presigned-URL prefix → uses `redactUrlForError`.
- `imageCache` and `fileDetector` URL caches keyed on the full URL (secret query params persisted in in-memory Map keys) → a shared sensitive-param denylist strips them from cache keys.
- URL pre-flight HEAD enforced `maxSize` without checking `statusCode` → non-2xx HEAD now falls through to the streaming GET guard.

**Proxy** — `processLooksLikeProxySupervisor` matched any process whose args merely contained `neurolink` → now requires both `neurolink` and the `proxy` subcommand before a stale/recycled pid could be signalled during uninstall.

## Tests & validation
- **8 new regression tests** (batch pdf-password/csv validation/options, aggregate page-limit warn, HEAD non-2xx fall-through, page-count timer/destroy resilience, cache-key redaction, supervisor identity). Suite **225/225**.
- `tsc --strict` 0 errors · `lint` 0 errors · `build:cli` green · pre-commit gate (validate:all + full build) green.

## Follow-up (out of scope, flagged)
`pdf-parse` (pinned `pdfjs-dist@5.4.296`) and `pdf-to-img` (`pdfjs-dist@5.4.624`) are separate installs; once `PDFProcessor.convertToImages()` runs in a process, subsequent real `getAccuratePageCount()` calls hit a pdfjs "API version does not match Worker version" error and degrade to `null` for the rest of the process. Pre-existing, not caused by these fixes — worth a follow-up ticket.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PDF page-limit warnings when some page counts are unknown (now clearer about partial verification).
  * Hardened URL loading by ignoring oversized `content-length` from non-2xx HEAD responses.
  * Reduced proxy supervisor uninstall false positives with stricter PID identity/start-time checks.
  * Strengthened signed/presigned URL handling and cache-key behavior to avoid auth-context collisions and better redact sensitive query params in logs.
  * Improved batch CSV/PDF processing: CSV/PDF validation, PDF password reuse per batch, and a dedicated PDF soft warning threshold.

* **Documentation**
  * Updated migration guidance for breaking changes (stricter CSV/file failures, closed message content shape, CLI batch args rename).
  * Clarified CSV encoding detection order/limitations and PDF conversion vs streaming behavior.

### DIFF
--- a/docs/MIGRATION.md
+++ b/docs/MIGRATION.md
@@ -1,0 +1,196 @@
+# Migration Guide: Breaking Changes
+
+This document tracks **breaking changes** shipped in specific NeuroLink
+releases — changes to public SDK/CLI surface that require an update on the
+consumer side, as opposed to the additive/backward-compatible changes covered
+by [`docs/guides/migration-guide.md`](./guides/migration-guide.md) and the
+auto-generated changelog (see the
+[GitHub releases page](https://github.com/juspay/neurolink/releases)).
+
+Three breaking changes shipped across 9.94.x–9.95.x. Each is intentional and
+stays as shipped — this document exists so downstream consumers know what
+changed and how to adapt.
+
+## Policy
+
+Per this repository's `CLAUDE.md` (Critical Rule 5), the public SDK API must
+not break existing callers. The three breaking changes below shipped in
+9.94.x–9.95.x without an accompanying migration path and are documented here
+retroactively. Going forward, any breaking change **must** ship its migration
+path in the same release that introduces it — not documented after the fact.
+
+---
+
+## 1. Multimodal file/CSV processing is now fail-loud (v9.94.6)
+
+**What changed:** `generate()`/`stream()` calls that attach files via
+`input.files` (auto-detected) or `input.csvFiles` (explicit CSV) now **throw**
+on the first file that fails to process, instead of logging a warning and
+silently continuing with the files that did succeed.
+
+- `processUnifiedFilesArray()` (`src/lib/utils/messageBuilder.ts`) throws
+  `ErrorFactory.fileProcessingFailed(filename, originalError)` on any file in
+  `input.files` that fails detection/processing.
+- `processExplicitCsvFiles()` (same file) throws
+  `ErrorFactory.csvProcessingFailed(filename, originalError)` on any file in
+  `input.csvFiles` that fails to parse.
+
+Both errors are typed `NeuroLinkError`s (`ERROR_CODES.FILE_PROCESSING_FAILED`
+/ `ERROR_CODES.CSV_PROCESSING_FAILED`, see `src/lib/utils/errorHandling.ts`)
+that carry the original error as `.originalError` and the failing filename in
+`.context.filename`.
+
+**Why:** the previous log-and-skip behavior silently produced a partial
+prompt — e.g. attaching 5 files and having one silently drop meant the model
+answered as if that file never existed, with no signal to the caller or the
+end user that anything was missing. A model call that appears to succeed but
+is actually missing part of its input is worse than a call that fails loudly.
+
+**Before (9.94.5 and earlier):**
+
+```typescript
+// One corrupt/unreadable file among several silently dropped; the request
+// still succeeded using only the files that processed cleanly.
+const result = await neurolink.generate({
+  input: { text: "Summarize these", files: ["a.pdf", "corrupt.pdf", "b.pdf"] },
+});
+// result.content summarizes only a.pdf and b.pdf — no error, no warning
+// visible to the caller beyond a log line.
+```
+
+**After (9.94.6+):**
+
+```typescript
+try {
+  const result = await neurolink.generate({
+    input: {
+      text: "Summarize these",
+      files: ["a.pdf", "corrupt.pdf", "b.pdf"],
+    },
+  });
+} catch (error) {
+  if (
+    error instanceof NeuroLinkError &&
+    error.code === "FILE_PROCESSING_FAILED"
+  ) {
+    // error.context.filename === "corrupt.pdf"
+    // error.originalError is the underlying cause
+  }
+  throw error;
+}
+```
+
+**How to adapt:**
+
+- Wrap multi-file `generate()`/`stream()` calls in `try`/`catch` if you were
+  previously relying on partial success.
+- If you want the old best-effort behavior, pre-validate/pre-filter files
+  yourself before passing them in `input.files`/`input.csvFiles`, or catch the
+  typed error, drop the failing filename (`error.context.filename`), and
+  retry without it.
+- CSV-specific failures (`input.csvFiles`) and generic file failures
+  (`input.files`) now throw _different_ error codes — branch on
+  `error.code` if you need to distinguish them.
+
+---
+
+## 2. `MessageContent` dropped its open index signature (v9.94.6)
+
+**What changed:** the exported `MessageContent` type
+(`src/lib/types/multimodal.ts`, re-exported via `src/lib/types/index.ts` and
+the package root) no longer has a `[key: string]: unknown` index signature.
+It's now a closed shape with the concrete fields the pipeline actually reads:
+`type`, `text`, `image`, `mimeType`, `data`, `name`, `filename`,
+`toolCallId`, `toolName`, `args`, `result`, `isError`, `providerOptions`.
+
+**Why:** the open index signature let any typo or unrelated key through
+type-checking unchecked (e.g. `{ type: "text", txt: "..." }` compiled fine
+and silently produced empty content). Closing it catches that class of bug at
+compile time for everyone building `MessageContent[]` arrays directly.
+
+**Before:**
+
+```typescript
+import type { MessageContent } from "@juspay/neurolink";
+
+// Compiled fine even though "customField" isn't a real MessageContent key —
+// the typo/extra field was silently accepted.
+const item: MessageContent = {
+  type: "text",
+  text: "hello",
+  customField: "some app-specific metadata",
+};
+```
+
+**After:**
+
+```typescript
+import type { MessageContent } from "@juspay/neurolink";
+
+// TS2353: Object literal may only specify known properties, and
+// 'customField' does not exist in type 'MessageContent'.
+const item: MessageContent = {
+  type: "text",
+  text: "hello",
+  customField: "some app-specific metadata", // ❌ no longer type-checks
+};
+```
+
+**How to adapt:**
+
+- If you were attaching arbitrary metadata to a `MessageContent` item,
+  use the dedicated `providerOptions?: Record<string, unknown>` field instead
+  — it's read by `MessageBuilder` when converting to `ModelMessage[]` and is
+  the intended extension point.
+- If you genuinely need a field NeuroLink doesn't model, open an issue —
+  `MessageContent` intentionally stays a closed, hand-maintained shape rather
+  than reopening a blanket index signature (see the `#325` comment on the
+  type itself).
+- Runtime behavior is unchanged — this is a compile-time-only break. Plain
+  JavaScript callers, or TypeScript callers that never annotated a literal
+  as `MessageContent` explicitly, are unaffected.
+
+---
+
+## 3. `BatchCommandArgs.file` renamed to `.promptsFile` (v9.95.1)
+
+**What changed:** the public CLI argument type `BatchCommandArgs`
+(`src/lib/types/cli.ts`) renamed its `file?: string` property to
+`promptsFile?: string`.
+
+**Why:** yargs implicitly registers a positional argument's key as a
+recognized flag name too. Keeping the property named `file` collided with
+the _unrelated_ common `--file` auto-detect flag that the `batch` command
+intentionally does not register — `--file` on `batch` would silently pass
+`strictOptions()` without actually doing anything. Renaming the positional's
+backing property to `promptsFile` removes that collision.
+
+**Before:**
+
+```typescript
+import type { BatchCommandArgs } from "@juspay/neurolink";
+
+function handleBatch(argv: BatchCommandArgs) {
+  const promptsListPath = argv.file; // ❌ no longer exists
+}
+```
+
+**After:**
+
+```typescript
+import type { BatchCommandArgs } from "@juspay/neurolink";
+
+function handleBatch(argv: BatchCommandArgs) {
+  const promptsListPath = argv.promptsFile; // ✅
+}
+```
+
+**How to adapt:**
+
+- Update any code that constructs or reads a `BatchCommandArgs` object
+  directly (custom CLI wrappers, tests) to use `.promptsFile` instead of
+  `.file`.
+- The CLI's own `<promptsFile>` positional argument and its behavior are
+  unchanged — this is a type-only rename of the property NeuroLink's own
+  `batch` command handler reads internally; end users invoking
+  `neurolink batch <promptsFile>` on the command line see no difference.
--- a/src/cli/commands/proxy.ts
+++ b/src/cli/commands/proxy.ts
@@ -124,6 +124,14 @@ const PROXY_TELEMETRY_SCRIPT_PATH = fileURLToPath(
   ),
 );
 const PROXY_LIFECYCLE_SHUTDOWN_TIMEOUT_MS = 5_000;
+// Allowed drift between a pid's OS-reported start time and the persisted
+// ProxySupervisorState.startTime before processLooksLikeProxySupervisor
+// treats it as a confident mismatch (recycled pid). Generous on purpose:
+// `supervisorStartedAt` is captured inside runLaunchdProxySupervisor, which
+// runs after Node has already booted and done setup, so it always lags the
+// OS-level fork/exec by a little — a tight tolerance would produce false
+// mismatches on a slow/cold-started machine.
+const PROXY_SUPERVISOR_START_TIME_TOLERANCE_MS = 10_000;
 const PROXY_UPDATE_CONTROL_TOKEN =
   process.env.NEUROLINK_PROXY_UPDATE_CONTROL_TOKEN?.trim() ||
   crypto.randomUUID();
@@ -203,22 +211,148 @@ function getProcessStatus(pid: number): "running" | "not_running" | "unknown" {
 }
 
 /**
+ * Read a single `ps -o <field>=` field for `pid`, off the event loop (async
+ * `execFile`, not `execFileSync`) and bounded by a short timeout, so a
+ * hung/zombie `ps` invocation can never block the event loop or hang the
+ * (often uninstall-path) async caller. Shared by {@link getProcessStartTime}
+ * and {@link processLooksLikeProxySupervisor}, which both read a single
+ * `ps -o <field>=` column for the same pid.
+ *
+ * Returns null — never throws — on any error, timeout, or empty output, so
+ * callers uniformly treat "couldn't read" as "can't confirm either way"
+ * rather than distinguishing the failure mode.
+ */
+async function readPsField(pid: number, field: string): Promise<string | null> {
+  try {
+    const { execFile } = await import("node:child_process");
+    const { promisify } = await import("node:util");
+    const execFileAsync = promisify(execFile);
+    const { stdout } = await withTimeout(
+      execFileAsync("ps", ["-p", String(pid), "-o", `${field}=`], {
+        encoding: "utf-8" as const,
+      }),
+      2000,
+      `ps -o ${field}= timed out for pid ${pid}`,
+    );
+    const trimmed = stdout.trim();
+    return trimmed || null;
+  } catch {
+    return null;
+  }
+}
+
+/**
+ * Best-effort, macOS-only (uninstall is macOS-only) read of `pid`'s
+ * OS-reported start time via `ps -o lstart=`, which prints a fixed-width
+ * `Www Mmm dd hh:mm:ss yyyy` timestamp — the same shape `Date`'s parser
+ * already understands (it's the format `Date#toString()` itself produces
+ * modulo the trailing timezone). Returns null (never throws) when `ps`
+ * fails or its output doesn't parse, so callers can tell "confirmed
+ * mismatch" apart from "couldn't confirm either way".
+ */
+async function getProcessStartTime(pid: number): Promise<Date | null> {
+  const out = await readPsField(pid, "lstart");
+  if (!out) {
+    return null;
+  }
+  // `ps -o lstart=` prints the OS start time in the machine's LOCAL timezone
+  // with no offset, e.g. "Sun Jul 19 20:12:41 2026". Parse the components and
+  // build the Date via the local-time constructor rather than relying on the
+  // engine's implementation-defined parsing of non-ISO date strings. The
+  // result is the same absolute instant the supervisor persisted as an
+  // ISO-UTC string, so the drift comparison against it is timezone-safe.
+  const m = out.match(
+    /^\w{3}\s+(\w{3})\s+(\d{1,2})\s+(\d{2}):(\d{2}):(\d{2})\s+(\d{4})$/,
+  );
+  if (!m) {
+    return null;
+  }
+  const monthIndex = [
+    "Jan",
+    "Feb",
+    "Mar",
+    "Apr",
+    "May",
+    "Jun",
+    "Jul",
+    "Aug",
+    "Sep",
+    "Oct",
+    "Nov",
+    "Dec",
+  ].indexOf(m[1]);
+  if (monthIndex < 0) {
+    return null;
+  }
+  const parsed = new Date(
+    Number(m[6]),
+    monthIndex,
+    Number(m[2]),
+    Number(m[3]),
+    Number(m[4]),
+    Number(m[5]),
+  );
+  return Number.isNaN(parsed.getTime()) ? null : parsed;
+}
+
+/**
  * Best-effort check that a pid actually belongs to a neurolink proxy process,
  * so a stale/recycled supervisor pid is never mistaken for a live supervisor
  * and signalled. Returns false when the process cannot be confirmed as ours.
+ *
+ * @param expectedStartTimeIso - The `ProxySupervisorState.startTime` (ISO
+ *   string) recorded when the supervisor we expect at `pid` was launched.
+ *   When provided, this is cross-checked against `pid`'s actual OS-reported
+ *   start time (see {@link getProcessStartTime}) — the args match alone
+ *   ("neurolink" + "proxy" both present) is not airtight, since an
+ *   unrelated process (e.g. a shell running this very test suite, or a
+ *   coincidentally-named script) could match it too. A pid recycled by such
+ *   a process would have a start time far from the recorded supervisor
+ *   startTime, which this catches.
+ *
+ *   Deliberately fail-open on anything unparseable: `ps -o lstart=` output
+ *   parsing is inherently a little fragile (locale/format quirks), and the
+ *   cost of a false "can't confirm" is silently regressing to pre-hardening
+ *   behavior, while the cost of a false "confirmed mismatch" is leaving a
+ *   real supervisor running and orphaning its socket during uninstall — the
+ *   former is the safer failure mode. So this only ever returns false for
+ *   the startTime check when BOTH timestamps parsed successfully AND their
+ *   drift exceeds the tolerance; any parse failure is treated as "can't
+ *   confirm a mismatch" and falls through to the args-only result.
  */
-async function processLooksLikeProxySupervisor(pid: number): Promise<boolean> {
-  try {
-    const { execFileSync } = await import("node:child_process");
-    const args = execFileSync("ps", ["-p", String(pid), "-o", "args="], {
-      encoding: "utf-8",
-      stdio: ["ignore", "pipe", "ignore"],
-    });
-    return /neurolink/i.test(args);
-  } catch {
-    // `ps` unavailable or the pid vanished — do not signal an unverified pid.
+export async function processLooksLikeProxySupervisor(
+  pid: number,
+  expectedStartTimeIso?: string,
+): Promise<boolean> {
+  const args = await readPsField(pid, "args");
+  if (args === null) {
+    // `ps` unavailable, timed out, or the pid vanished — do not signal an
+    // unverified pid.
     return false;
   }
+  // `/neurolink/i` alone is too broad: any process whose args merely
+  // *mention* "neurolink" (e.g. a shell running in a directory named
+  // "neurolink", an editor with a neurolink file open) would match, and a
+  // stale/recycled pid running such a process could then be SIGTERM'd/
+  // SIGKILL'd by `ensureSupervisorStoppedBeforeClear` during uninstall.
+  // Require the actual proxy-supervisor invocation — both "neurolink" AND
+  // the "proxy" subcommand present in args.
+  if (!(/neurolink/i.test(args) && /\bproxy\b/i.test(args))) {
+    return false;
+  }
+  if (!expectedStartTimeIso) {
+    return true;
+  }
+  const expected = new Date(expectedStartTimeIso);
+  if (Number.isNaN(expected.getTime())) {
+    return true; // recorded startTime itself is unparseable — can't confirm a mismatch
+  }
+  const actual = await getProcessStartTime(pid);
+  if (!actual) {
+    return true; // ps -o lstart= unavailable/unparseable — can't confirm a mismatch
+  }
+  const driftMs = Math.abs(actual.getTime() - expected.getTime());
+  return driftMs <= PROXY_SUPERVISOR_START_TIME_TOLERANCE_MS;
 }
 
 /**
@@ -229,14 +363,19 @@ async function processLooksLikeProxySupervisor(pid: number): Promise<boolean> {
  * the uninstall instead of silently discarding the record.
  */
 async function ensureSupervisorStoppedBeforeClear(): Promise<boolean> {
-  const pid = loadProxySupervisorState()?.pid;
+  const supervisorState = loadProxySupervisorState();
+  const pid = supervisorState?.pid;
   if (!pid || getProcessStatus(pid) === "not_running") {
     return true;
   }
   // Guard against a stale/recycled pid: only signal a process that still looks
   // like a neurolink proxy supervisor, so uninstall never terminates an
-  // unrelated, same-user process that inherited the recorded pid.
-  if (!(await processLooksLikeProxySupervisor(pid))) {
+  // unrelated, same-user process that inherited the recorded pid. Also cross-
+  // checks the pid's actual OS start time against the recorded
+  // supervisorState.startTime (see processLooksLikeProxySupervisor).
+  if (
+    !(await processLooksLikeProxySupervisor(pid, supervisorState?.startTime))
+  ) {
     return true;
   }
   try {
--- a/src/cli/factories/commandFactory.ts
+++ b/src/cli/factories/commandFactory.ts
@@ -3369,6 +3369,28 @@ export class CLICommandFactory {
   }
 
   /**
+   * Build video processor options from CLI argv. Shared by executeGenerate,
+   * executeStream and executeBatch, which previously built byte-for-byte
+   * identical `videoOptions` objects independently (mirrors
+   * {@link buildCsvOptionsFromArgv}).
+   */
+  private static buildVideoOptionsFromArgv(
+    argv: BaseCommandArgs & Record<string, unknown>,
+  ): {
+    frames?: number;
+    quality?: number;
+    format?: "jpeg" | "png";
+    transcribeAudio?: boolean;
+  } {
+    return {
+      frames: argv.videoFrames as number | undefined,
+      quality: argv.videoQuality as number | undefined,
+      format: argv.videoFormat as "jpeg" | "png" | undefined,
+      transcribeAudio: argv.transcribeAudio as boolean | undefined,
+    };
+  }
+
+  /**
    * Build output configuration for generate request
    */
   private static buildGenerateOutputConfig(
@@ -3716,12 +3738,7 @@ export class CLICommandFactory {
             password: CLICommandFactory.resolvePdfPassword(argv),
           },
           csvOptions: CLICommandFactory.buildCsvOptionsFromArgv(argv),
-          videoOptions: {
-            frames: argv.videoFrames as number | undefined,
-            quality: argv.videoQuality as number | undefined,
-            format: argv.videoFormat as "jpeg" | "png" | undefined,
-            transcribeAudio: argv.transcribeAudio as boolean | undefined,
-          },
+          videoOptions: CLICommandFactory.buildVideoOptionsFromArgv(argv),
           output: outputConfig,
           provider: enhancedOptions.provider,
           model: enhancedOptions.model,
@@ -4070,12 +4087,7 @@ export class CLICommandFactory {
           password: CLICommandFactory.resolvePdfPassword(argv),
         },
         csvOptions: CLICommandFactory.buildCsvOptionsFromArgv(argv),
-        videoOptions: {
-          frames: argv.videoFrames as number | undefined,
-          quality: argv.videoQuality as number | undefined,
-          format: argv.videoFormat as "jpeg" | "png" | undefined,
-          transcribeAudio: argv.transcribeAudio as boolean | undefined,
-        },
+        videoOptions: CLICommandFactory.buildVideoOptionsFromArgv(argv),
         provider: enhancedOptions.provider as string | undefined,
         model: enhancedOptions.model as string | undefined,
         temperature: enhancedOptions.temperature as number | undefined,
@@ -4701,6 +4713,7 @@ export class CLICommandFactory {
       // see createBatchCommand — so `argv.file` can never be set here; no
       // exclusion needed to protect the (differently-named) positional.
       validateCliInputFiles(argv as unknown as Record<string, unknown>);
+      validateCsvMaxRows(argv as unknown as Record<string, unknown>);
       const batchImages = CLICommandFactory.processCliImages(
         argv.image as string | string[] | undefined,
       );
@@ -4797,6 +4810,13 @@ export class CLICommandFactory {
         spinner?.start();
       }
 
+      // Resolve the PDF password once for the whole batch. resolvePdfPassword
+      // emits a "visible in shell history" stderr warning when the flag is set;
+      // it must fire once per run (like generate/stream), not once per prompt.
+      const batchPdfPassword = batchPdfFiles?.length
+        ? CLICommandFactory.resolvePdfPassword(argv)
+        : undefined;
+
       for (let i = 0; i < prompts.length; i++) {
         if (spinner) {
           spinner.text = `Processing ${i + 1}/${prompts.length}: ${prompts[i].substring(0, 30)}...`;
@@ -4860,22 +4880,15 @@ export class CLICommandFactory {
               // unlike generate/stream, so an empty options object here is
               // pure noise on every prompt of every batch run.
               ...(batchCsvFiles?.length && {
-                csvOptions: {
-                  maxRows: argv.csvMaxRows as number | undefined,
-                  formatStyle: argv.csvFormat as
-                    | "raw"
-                    | "markdown"
-                    | "json"
-                    | undefined,
+                csvOptions: CLICommandFactory.buildCsvOptionsFromArgv(argv),
+              }),
+              ...(batchPdfFiles?.length && {
+                pdfOptions: {
+                  password: batchPdfPassword,
                 },
               }),
               ...(batchVideoFiles?.length && {
-                videoOptions: {
-                  frames: argv.videoFrames as number | undefined,
-                  quality: argv.videoQuality as number | undefined,
-                  format: argv.videoFormat as "jpeg" | "png" | undefined,
-                  transcribeAudio: argv.transcribeAudio as boolean | undefined,
-                },
+                videoOptions: CLICommandFactory.buildVideoOptionsFromArgv(argv),
               }),
               provider: enhancedOptions.provider,
               model: enhancedOptions.model,
--- a/src/lib/utils/fileDetector.ts
+++ b/src/lib/utils/fileDetector.ts
@@ -42,7 +42,11 @@ import { CSVProcessor } from "./csvProcessor.js";
 import { ImageProcessor } from "./imageProcessor.js";
 import { logger } from "./logger.js";
 import { withTimeout } from "./errorHandling.js";
-import { redactUrlForError, sanitizeErrorCause } from "./logSanitize.js";
+import {
+  normalizeUrlForCache,
+  redactUrlForError,
+  sanitizeErrorCause,
+} from "./logSanitize.js";
 import {
   mimeHintToExtension,
   mimeHintToFileType,
@@ -81,22 +85,41 @@ const urlContentTypeCache = new Map<
   { contentType: string; expiresAt: number }
 >();
 
+/**
+ * Build the Map key for `urlContentTypeCache`. Delegates to the shared
+ * {@link normalizeUrlForCache} — this is a module-level, process-lifetime
+ * cache, so it must strip presigned-URL auth/signature query params (see
+ * `SENSITIVE_URL_QUERY_PARAM_DENYLIST`) the same way `ImageCache.normalizeUrl`
+ * does, folding a short hash of the stripped params into the key whenever any
+ * were present so two different presigned URLs for the same path don't
+ * collide and serve each other's cached content-type across auth contexts.
+ * Also strips tracking/analytics params first, so two URLs differing only by
+ * tracking noise (e.g. `utm_source`) still hit the same cache entry instead
+ * of missing each other. Falls back to the raw URL if it isn't a parseable
+ * absolute URL. Shared by both `getCachedUrlContentType` and
+ * `setCachedUrlContentType` so lookups and writes always agree on the key.
+ */
+function cacheKeyForUrl(url: string): string {
+  return normalizeUrlForCache(url);
+}
+
 function getCachedUrlContentType(url: string, now: number): string | undefined {
-  const hit = urlContentTypeCache.get(url);
+  const key = cacheKeyForUrl(url);
+  const hit = urlContentTypeCache.get(key);
   if (hit && hit.expiresAt > now) {
     // Bump recency: Map iteration order follows insertion order, and the
     // eviction below deletes the *first* key, so a plain `get` on a hot
     // entry would leave it first in line for eviction despite being the
     // most recently used. Re-inserting turns the size-bounded FIFO below
     // into an actual LRU.
-    urlContentTypeCache.delete(url);
-    urlContentTypeCache.set(url, hit);
+    urlContentTypeCache.delete(key);
+    urlContentTypeCache.set(key, hit);
     return hit.contentType;
   }
   if (hit) {
     // Entry exists but its TTL has passed — treat as a miss and evict it
     // immediately rather than serving (or retaining) stale data.
-    urlContentTypeCache.delete(url);
+    urlContentTypeCache.delete(key);
   }
   return undefined;
 }
@@ -122,7 +145,8 @@ function setCachedUrlContentType(
   if (!contentType) {
     return;
   }
-  urlContentTypeCache.set(url, {
+  const key = cacheKeyForUrl(url);
+  urlContentTypeCache.set(key, {
     contentType,
     expiresAt: now + URL_CONTENT_TYPE_TTL_MS,
   });
@@ -1915,19 +1939,28 @@ export class FileDetector {
         });
         // Drain/close the (empty) HEAD body so the connection can be reused.
         await head.body.dump();
-        const declaredLength = Number(head.headers["content-length"]);
-        if (Number.isFinite(declaredLength) && declaredLength > maxSize) {
-          throw new Error(
-            `File too large: ${formatFileSize(declaredLength)} (max: ${formatFileSize(maxSize)})`,
-          );
+        // Only trust `content-length` on a genuine 2xx response. A non-2xx
+        // HEAD (redirect the dispatcher didn't follow, 403/404/405 "HEAD not
+        // allowed", 5xx, …) can still carry a stale/irrelevant
+        // `content-length` header — enforcing size off of that would reject
+        // (or silently pass) based on the wrong body. Treat any non-2xx HEAD
+        // as if the header were missing and fall through to the streaming
+        // GET guard below, which enforces maxSize independently either way.
+        if (head.statusCode >= 200 && head.statusCode < 300) {
+          const declaredLength = Number(head.headers["content-length"]);
+          if (Number.isFinite(declaredLength) && declaredLength > maxSize) {
+            throw new Error(
+              `File too large: ${formatFileSize(declaredLength)} (max: ${formatFileSize(maxSize)})`,
+            );
+          }
         }
       } catch (error) {
         if (error instanceof Error && /File too large/.test(error.message)) {
           throw error;
         }
         logger.debug(
-          `[FileDetector] HEAD pre-flight skipped for ${url}: ${
-            error instanceof Error ? error.message : String(error)
+          `[FileDetector] HEAD pre-flight skipped for ${redactUrlForError(url)}: ${
+            sanitizeErrorCause(error).message
           }`,
         );
       }
--- a/src/lib/utils/imageCache.ts
+++ b/src/lib/utils/imageCache.ts
@@ -15,7 +15,7 @@
 
 import { createHash } from "crypto";
 import { logger } from "./logger.js";
-import { redactUrlForError } from "./logSanitize.js";
+import { normalizeUrlForCache, redactUrlForError } from "./logSanitize.js";
 import type {
   CachedImage,
   ImageCacheConfig,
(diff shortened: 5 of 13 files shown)


docs/MIGRATION.md
Tara-ag, line 1: ⚠️ **MAJOR**: This doc retroactively documents three breaking changes as acceptable, but CLAUDE.md Critical Rule 5 states the public SDK API must not break existing callers. Documenting breakage after the fact is not the same as providing a documented migration path *before* breaking. Two of the three breaks (`MessageContent` index signature removal and `BatchCommandArgs.file` → `promptsFile`) are public type/API surface changes. If these already shipped in 9.94.x–9.95.x, the doc helps, but for this PR the question is whether the project wants to continue accepting undocumented-then-documented breaks. Consider adding a policy note that future breaking changes require pre-shipping migration docs, and evaluate whether the `MessageContent` closed shape should retain a deprecated index signature for one minor cycle to ease adoption.

src/cli/factories/commandFactory.ts
```diff
@@ -4860,13 +4861,11 @@ export class CLICommandFactory {
...
-                    | "raw"
-                    | "markdown"
-                    | "json"
-                    | undefined,
+                csvOptions: CLICommandFactory.buildCsvOptionsFromArgv(argv),
+              }),
+              ...(batchPdfFiles?.length && {
+                pdfOptions: {
+                  password: CLICommandFactory.resolvePdfPassword(argv),
                 },
```
Tara-ag, line 4888: 💡 MINOR: The batch command now wires `pdfOptions.password` and `csvOptions` via shared helpers, which is good. However, `executeBatch` still constructs `videoOptions` inline (the `...(batchVideoFiles?.length && { videoOptions: {...} })` block) rather than delegating to a shared helper like `buildCsvOptionsFromArgv`/`resolvePdfPassword`. This duplicates the generate/stream video-options shape and risks drift if more video flags are added. Consider extracting `buildVideoOptionsFromArgv(argv)` and reusing it across all three command paths.

src/cli/commands/proxy.ts
```diff
@@ -207,14 +207,27 @@ function getProcessStatus(pid: number): "running" | "not_running" | "unknown" {
...
+): Promise<boolean> {
   try {
     const { execFileSync } = await import("node:child_process");
     const args = execFileSync("ps", ["-p", String(pid), "-o", "args="], {
       encoding: "utf-8",
       stdio: ["ignore", "pipe", "ignore"],
     });
-    return /neurolink/i.test(args);
+    // `/neurolink/i` alone is too broad: any process whose args merely
+    // *mention* "neurolink" (e.g. a shell running in a directory named
```
Tara-ag: 💡 MINOR: `processLooksLikeProxySupervisor` now requires both "neurolink" and "proxy" in the args, which is a meaningful hardening step. However, the function still relies on a regex over `ps` output rather than a stronger identity signal. The comment itself notes that a startTime cross-check would harden it further. Given that this is the only guard preventing SIGTERM/SIGKILL of an unrelated process that happens to contain both words, consider implementing the startTime cross-check against `ProxySupervisorState.startTime` (or recording a supervisor nonce in state and verifying it via `/proc/<pid>/environ` on Linux or `ps -o lstart=` on macOS) in a follow-up. This is acceptable as a defensive incremental fix, but should not be considered a complete solution.

Tara-ag (approved):
## Review Summary

**Files reviewed:** 13
**New issues raised:** 3 (1 MAJOR, 2 MINOR)
**Blocking issues:** 0

### Findings

- **MAJOR** — `docs/MIGRATION.md`: The doc retroactively documents three breaking changes, which is helpful. However, CLAUDE.md Critical Rule 5 states the public SDK API must not break existing callers. Consider adding a policy note that future breaking changes require pre-shipping migration docs, and evaluate whether `MessageContent` should retain a deprecated index signature for one minor cycle to ease adoption.

- **MINOR** — `src/cli/factories/commandFactory.ts`: `executeBatch` still constructs `videoOptions` inline rather than using a shared helper like `buildCsvOptionsFromArgv`/`resolvePdfPassword`. Extracting `buildVideoOptionsFromArgv(argv)` would reduce drift across generate/stream/batch.

- **MINOR** — `src/cli/commands/proxy.ts`: `processLooksLikeProxySupervisor` is a meaningful hardening step, but remains regex-based over `ps` output. Consider a follow-up with a stronger identity signal (startTime cross-check against `ProxySupervisorState.startTime`, or a nonce verified via `/proc/<pid>/environ` / `ps -o lstart=`).

### What looks good

- Secret redaction is now applied to long-lived cache keys (`ImageCache`, `FileDetector` URL cache), not just logs.
- URL pre-flight HEAD correctly ignores non-2xx `content-length` and falls through to the streaming GET guard.
- `getAccuratePageCount` clears its race timer and guards a throwing `pdf.destroy()`.
- Batch parity fixes wire `pdfOptions.password` and the full `csvOptions` shape.
- Eight focused regression tests cover the claimed gaps.

No hardcoded secrets, injection vectors, or other critical security issues were introduced. Approving with inline comments for the author to address at their discretion.

test/continuous-test-suite-bugfixes.ts
```diff
@@ -7776,6 +7778,419 @@ exit 127
...
+      const src = readFileSync(
+        resolvePath(process.cwd(), "src/cli/factories/commandFactory.ts"),
+        "utf-8",
+      );
+      const marker = "const runBatchGenerate = ()";
+      const start = src.indexOf(marker);
+      if (start === -1) {
+        return false;
+      }
+      const compact = src.slice(start, start + 2000).replace(/\s+/g, "");
```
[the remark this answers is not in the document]
murdore: Fixed in `da030924`. Both brittle checks now scan from the marker to end-of-file (`source.slice(start)`) instead of a fixed 2000-char window, so they won't false-fail as the `executeBatch` block grows.

test/continuous-test-suite-bugfixes.ts
```diff
@@ -7776,6 +7778,419 @@ exit 127
...
```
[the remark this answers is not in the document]
murdore: Fixed in `da030924` — same change: scans from `start` to EOF instead of a fixed 2000-char slice.

src/lib/utils/logSanitize.ts
```diff
@@ -239,6 +239,52 @@ export function redactUrlForError(url: string): string {
...
+  "x-amz-credential",
+  "x-amz-security-token",
+  "awsaccesskeyid",
+  "expires",
+  "sig",
+  "se",
+  "sp",
+  "sr",
+  "sv",
+];
```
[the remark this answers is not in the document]
murdore: Fixed in `da030924`. Added the GCS V4 signed-URL params (`x-goog-signature`, `x-goog-credential`, `x-goog-algorithm`, `x-goog-date`, `x-goog-expires`, `x-goog-signedheaders`) to `SENSITIVE_URL_QUERY_PARAM_DENYLIST`, case-insensitively.

src/lib/utils/imageCache.ts
```diff
@@ -141,6 +141,11 @@ export class ImageCache {
         "_ga",
       ];
       trackingParams.forEach((param) => parsed.searchParams.delete(param));
+      // This normalized URL is used as the in-memory Map key (see `get`/
+      // `set` below), which lives for the process lifetime — strip
+      // presigned-URL auth/signature secrets too, not just tracking noise,
+      // so they never persist as a cache key.
+      stripSensitiveUrlParams(parsed);
       return parsed.toString();
```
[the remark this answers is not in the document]
murdore: Good catch — fixed in `da030924`. Rather than stripping the sensitive params from the cache key (which collided distinct presigned URLs for the same object and could serve stale bytes for an expired URL), the key now appends a short non-reversible hash of the stripped param values: `key = <url-without-sensitive-params>` + (if any were present) `\0auth=<sha256(sorted stripped k=v)[:16]>`. Secrets never persist in plaintext, but different auth contexts stay distinct — no collision, no stale-bytes bug. Applied to both `imageCache.normalizeUrl` and `fileDetector.cacheKeyForUrl` via a shared `stripSensitiveUrlParamsForCacheKey` helper. The regression test was rewritten to require a *different* signature to miss the cache (it previously asserted the buggy collision as correct).

docs/MIGRATION.md
murdore: Addressed in `da030924`. Added a **Policy** note at the top of `docs/MIGRATION.md` citing CLAUDE.md Critical Rule 5: these three breaks shipped undocumented in 9.94.x–9.95.x and are documented here retroactively, and future breaking changes must ship their migration path in the same release, not after the fact. Per the maintainer's decision on this PR, the already-shipped behavior is kept and documented (not reverted).

src/cli/factories/commandFactory.ts
murdore: Fixed in `da030924`. Extracted `CLICommandFactory.buildVideoOptionsFromArgv(argv)` (mirroring `buildCsvOptionsFromArgv`) and reused it at all three call sites (`executeGenerate`, `executeStream`, `executeBatch`) — the three inline shapes were byte-identical, so this is behavior-preserving and removes the drift risk.

src/cli/commands/proxy.ts
murdore: Implemented in `da030924`. Added `getProcessStartTime(pid)` (macOS `ps -p <pid> -o lstart=`) and a start-time cross-check in `processLooksLikeProxySupervisor` against the persisted `ProxySupervisorState.startTime` (10s tolerance to absorb the Node-boot lag before `supervisorStartedAt` is captured). It's **fail-open**: a missing/unparseable persisted or observed start time returns `true` (can't confirm a mismatch — preserves prior behavior); it only returns `false` on a confidently-parsed mismatch (recycled pid), so a stale pid is no longer signalled during uninstall.

src/cli/factories/commandFactory.ts
```diff
@@ -4860,22 +4873,15 @@ export class CLICommandFactory {
...
               }),
               ...(batchVideoFiles?.length && {
-                videoOptions: {
-                  frames: argv.videoFrames as number | undefined,
-                  quality: argv.videoQuality as number | undefined,
-                  format: argv.videoFormat as "jpeg" | "png" | undefined,
-                  transcribeAudio: argv.transcribeAudio as boolean | undefined,
-                },
+                videoOptions: CLICommandFactory.buildVideoOptionsFromArgv(argv),
               }),
```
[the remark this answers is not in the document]
murdore: Fixed in `d2236bb2ca0277510328e6aa1a67d5eefdeaff89`. The PDF password is now resolved once before the per-prompt loop (`const batchPdfPassword = batchPdfFiles?.length ? resolvePdfPassword(argv) : undefined`) and reused in each prompt's `pdfOptions`, so the shell-history warning fires once per run like generate/stream, not once per prompt.

src/lib/utils/fileDetector.ts
```diff
@@ -1915,11 +1945,20 @@ export class FileDetector {
...
+        // (or silently pass) based on the wrong body. Treat any non-2xx HEAD
+        // as if the header were missing and fall through to the streaming
+        // GET guard below, which enforces maxSize independently either way.
+        if (head.statusCode >= 200 && head.statusCode < 300) {
+          const declaredLength = Number(head.headers["content-length"]);
+          if (Number.isFinite(declaredLength) && declaredLength > maxSize) {
+            throw new Error(
+              `File too large: ${formatFileSize(declaredLength)} (max: ${formatFileSize(maxSize)})`,
+            );
+          }
```
Tara-ag, line 1955: 🔒 **CRITICAL**: This debug log echoes the raw `url`, including any presigned-URL query tokens/signatures, into logs. This is the same class of leak the PR fixes in `imageProcessor.ts` (cache-hit debug log) and contradicts the URL-secret hardening goal. The GET error path below already redacts correctly; the HEAD pre-flight catch should do the same.

Use the already-imported helpers:

```ts
logger.debug(
  `[FileDetector] HEAD pre-flight skipped for ${redactUrlForError(url)}: ${
    sanitizeErrorCause(error).message
  }`,
);
```

This preserves the diagnostic intent while ensuring secrets never reach logs.

Tara-ag (changes requested):
## Review Summary

**Files reviewed:** 13  
**New issues raised this run:** 0 (one existing unresolved CRITICAL comment remains)  
**Blocking issues:** 1

### Blocking issue

- **🔒 CRITICAL — `src/lib/utils/fileDetector.ts:1961`**: The HEAD pre-flight catch block still logs the raw `url`, including presigned-URL query tokens/signatures. This contradicts the URL-secret hardening goal of this PR and violates the secret-exposure blocking criterion. The GET error path below already redacts correctly; the HEAD catch should use the same `redactUrlForError(url)` + `sanitizeErrorCause(error).message` pattern. This issue is already tracked in the pending inline comment and must be fixed before merge.

### Previously raised issues

All other points from earlier review rounds ( brittle source-scan windows, GCS denylist, cache-key collision, retroactive migration docs, shared `buildVideoOptionsFromArgv`, proxy start-time cross-check, and per-prompt PDF-password resolution) have been addressed by the author.

### What looks good

- Batch parity fixes wire the full `csvOptions` shape, validate `--csv-max-rows`, and resolve the PDF password once per run.
- URL cache keys now strip sensitive query params while keeping distinct auth contexts distinct via the hash suffix.
- Non-2xx HEAD responses no longer trust misleading `content-length`.
- PDF page-count race timer is cleared and `destroy()` failures are guarded.
- Proxy supervisor identity is hardened with both arg matching and start-time cross-check.
- Regression tests cover the claimed gaps.

Please redact the HEAD pre-flight debug log, then this is good to approve.

src/lib/utils/fileDetector.ts
murdore: Fixed in `3a4d17b7b514adc8e7625bdfb7178e94990773d1`. The HEAD pre-flight catch's debug log now uses `redactUrlForError(url)` for the URL and `sanitizeErrorCause(error).message` for the reason, matching the GET error path and the URL-secret hardening goal — no raw token/signature reaches the log. Swept the rest of fileDetector.ts for other raw-`${url}` logs; this was the only one.

src/cli/commands/proxy.ts
```diff
@@ -202,19 +210,90 @@ function getProcessStatus(pid: number): "running" | "not_running" | "unknown" {
...
 }
 
+/**
+ * Best-effort, macOS-only (uninstall is macOS-only) read of `pid`'s
+ * OS-reported start time via `ps -o lstart=`, which prints a fixed-width
+ * `Www Mmm dd hh:mm:ss yyyy` timestamp — the same shape `Date`'s parser
+ * already understands (it's the format `Date#toString()` itself produces
+ * modulo the trailing timezone). Returns null (never throws) when `ps`
+ * fails or its output doesn't parse, so callers can tell "confirmed
+ * mismatch" apart from "couldn't confirm either way".
```
Tara-ag, line 251: 🔒 **CRITICAL**: `getProcessStartTime` uses `execFileSync` inside an `async` function, which blocks the event loop while waiting for `ps`. This is called from `processLooksLikeProxySupervisor`, which is awaited from `ensureSupervisorStoppedBeforeClear` during uninstall. Blocking the event loop on a synchronous child-process call in an async path is a responsiveness/availability issue and inconsistent with the rest of the codebase, which uses `withTimeout` for external calls and async I/O.

Use `execFile` (async) from `node:child_process` and wrap it with `withTimeout` so the call cannot hang indefinitely if `ps` misbehaves. For example:

```ts
import { execFile } from "node:child_process";
import { withTimeout } from "../../lib/utils/errorHandling.js";

async function getProcessStartTime(pid: number): Promise<Date | null> {
  try {
    const out = await withTimeout(
      new Promise<string>((resolve, reject) => {
        execFile(
          "ps",
          ["-p", String(pid), "-o", "lstart="],
          { encoding: "utf-8", stdio: ["ignore", "pipe", "ignore"] },
          (err, stdout) => (err ? reject(err) : resolve(stdout.trim())),
        );
      }),
      2_000,
      `ps startTime timeout for pid ${pid}`,
    );
    if (!out) return null;
    const parsed = new Date(out);
    return Number.isNaN(parsed.getTime()) ? null : parsed;
  } catch {
    return null;
  }
}
```

This keeps the uninstall path non-blocking and bounded.

src/cli/factories/commandFactory.ts
```diff
@@ -4860,22 +4880,15 @@ export class CLICommandFactory {
...
-                csvOptions: {
-                  maxRows: argv.csvMaxRows as number | undefined,
-                  formatStyle: argv.csvFormat as
-                    | "raw"
-                    | "markdown"
-                    | "json"
-                    | undefined,
+                csvOptions: CLICommandFactory.buildCsvOptionsFromArgv(argv),
+              }),
+              ...(batchPdfFiles?.length && {
```
Tara-ag, line 4885: ⚠️ **MAJOR**: `executeBatch` resolves the PDF password once before the loop (good), but it still calls `CLICommandFactory.resolvePdfPassword(argv)` unconditionally whenever `batchPdfFiles?.length` is truthy. The function returns `flagValue || envValue`, so if a user sets `NEUROLINK_PDF_PASSWORD` but passes no `--pdf-password` flag, the env var is used without printing the stderr warning. That part is fine. However, the comment says the warning must fire "when the flag is set", but the implementation will also fire it when only the env var is set (because `resolvePdfPassword` is called and the flag is absent, so `flagValue` is undefined and the warning is skipped). Actually re-reading: the warning only prints when `flagValue` is truthy, so env-only does not warn. Good.

The real issue is that `batchPdfPassword` is typed as `string | undefined` but is then spread into `pdfOptions: { password: batchPdfPassword }`. If `batchPdfFiles` is non-empty but the user supplied neither `--pdf-password` nor `NEUROLINK_PDF_PASSWORD`, `batchPdfPassword` is `undefined` and the resulting `pdfOptions` is `{ password: undefined }`. This is different from `generate`/`stream`, which always pass `pdfOptions` even when the password is undefined. For batch, the spread only adds `pdfOptions` when PDFs are attached, so `{ password: undefined }` is created. Most downstream consumers likely treat `undefined` as "no password", but this is a subtle inconsistency: generate/stream always pass `pdfOptions`, batch conditionally passes it with possibly-undefined password. Consider aligning by always passing `pdfOptions` in batch (matching generate/stream) or by omitting the key when the password is absent. As currently written, it is technically functional but brittle.

Suggested fix: keep the once-per-run resolution, but only construct `pdfOptions` when a password actually exists, or always pass it to match generate/stream:

```ts
const batchPdfPassword = batchPdfFiles?.length
  ? CLICommandFactory.resolvePdfPassword(argv)
  : undefined;
...
...(batchPdfFiles?.length && {
  pdfOptions: { password: batchPdfPassword },
}),
```

If you want to avoid `{ password: undefined }`, change to:

```ts
...(batchPdfFiles?.length && batchPdfPassword !== undefined && {
  pdfOptions: { password: batchPdfPassword },
}),
```

Either way, document the chosen behavior in a comment so future maintainers don't re-introduce the per-prompt resolution.

src/lib/utils/logSanitize.ts
```diff
@@ -239,6 +240,100 @@ export function redactUrlForError(url: string): string {
...
+  "se",
+  "sp",
+  "sr",
+  "sv",
+  "x-goog-signature",
+  "x-goog-credential",
+  "x-goog-algorithm",
+  "x-goog-date",
+  "x-goog-expires",
+  "x-goog-signedheaders",
```
Tara-ag, line 278: ⚠️ **MAJOR**: The `SENSITIVE_URL_QUERY_PARAM_DENYLIST` includes very short param names (`token`, `signature`, `sig`, `se`, `sp`, `sr`, `sv`, `expires`) that are also common *non-secret* query parameters. For example, a public API might use `?token=abc` as an opaque resource identifier, or `?expires=2026-07-20` as cache metadata. Stripping these blindly and folding them into the auth-hash suffix will (1) alter the cache key for resources that legitimately vary by these params and (2) cause unnecessary cache misses/collisions.

This is particularly risky for `token`, which is a generic word. The doc comment says "matched case-insensitively" but does not justify why a generic `token` param is treated as a secret. Consider narrowing the denylist to provider-specific param names and requiring a signature-related prefix for generic terms, or at least documenting the trade-off and adding a test that verifies non-auth `token`/`expires` params are preserved.

Suggested fix: remove the overly broad generic entries (`token`, `signature`, `expires`) and rely on the provider-prefixed ones (`x-amz-*`, `x-goog-*`, `sig`/`se`/`sp`/`sr`/`sv` for Azure SAS, `awsaccesskeyid`, `signature` for legacy AWS). If you must keep `token`, restrict it to a value-shape heuristic (e.g. minimum length / high entropy) or rename it to `access_token`/`auth_token`.

src/lib/utils/fileDetector.ts
```diff
@@ -81,22 +85,47 @@ const urlContentTypeCache = new Map<
   { contentType: string; expiresAt: number }
```
Tara-ag, line 85: 💡 **MINOR**: The `cacheKeyForUrl` helper correctly strips sensitive params and folds in a hash suffix, but it does not strip tracking params (`utm_*`, `fbclid`, `gclid`, `_ga`) before hashing. `ImageCache.normalizeUrl` does strip those first, then calls `stripSensitiveUrlParamsForCacheKey`. Here in `fileDetector.ts`, the cache key is built directly from the full URL minus sensitive params, so two URLs that differ only by tracking params will get different cache keys and miss each other. This is a minor inefficiency, not a correctness bug, but since the goal is a normalized cache key, consider applying the same tracking-param cleanup here (or moving both normalizations into a shared helper) so the URL content-type cache behaves consistently with `ImageCache`.

Suggested fix: either import/share the tracking-param denylist from `imageCache.ts` or extract a shared `normalizeUrlForCache(url)` helper in `logSanitize.ts` that both call sites use.

src/lib/utils/imageCache.ts
```diff
@@ -141,7 +144,17 @@ export class ImageCache {
         "_ga",
```
Tara-ag: 💡 **MINOR**: The deduplication logic in `set()` has a subtle correctness issue. When the same image content is uploaded under two different URLs, the code creates a shallow copy of the existing entry for the new URL and updates `contentHashIndex` to point to the *new* URL, discarding the old URL mapping. This means the old URL will still have its own entry in `this.cache`, but `getByContentHash` will now only find the new URL. More importantly, if the new URL entry is later evicted or expires, the old URL entry remains in `this.cache` but is no longer reachable via content-hash dedup. Conversely, if the old URL entry is evicted first, the content hash index still points to the new URL, which is fine.

This is pre-existing behavior, not introduced by this PR, but since the PR touches the cache-key normalization, it's worth noting that the dedup index update is lossy. A safer pattern would be to keep the content-hash index pointing to the *oldest* cached URL (or a set of URLs) so the dedup benefit survives eviction of any single URL. Given the PR's focus on cache-key hygiene, consider at least not overwriting the existing content-hash mapping when a duplicate arrives — the first cached URL is the most stable key.

src/lib/utils/messageBuilder.ts
```diff
@@ -1326,9 +1326,25 @@ async function processExplicitPdfFiles(
   // it in aggregate (e.g. three 40-page PDFs → 120 pages for a 100-page API).
```
Tara-ag, line 1327: 💡 **MINOR**: The new aggregate page-limit warning logs the *known* total but does not log *which* file(s) had unknown page counts. For a batch of many PDFs, the warning only says "at least one file's page count is unknown", forcing operators to guess. Include the filename(s) in the warn log so the diagnostic is actionable.

Suggested fix:

```ts
const unknownCountFiles = pdfFiles.filter(
  (f) => f.pageCount === null || f.pageCount === undefined,
);
if (unknownCountFiles.length > 0) {
  logger.warn(
    `[PDF] Aggregate page-limit check across ${pdfFiles.length} PDFs could only be ` +
      `partially verified: ${unknownCountFiles.length} file(s) had unknown page counts ` +
      `(${unknownCountFiles.map((f) => f.filename).join(", ")}), so the known ` +
      `total (${totalPages}) may undercount the true combined page count.`,
  );
}
```

This preserves the existing behavior while making the log useful.

src/lib/utils/pdfProcessor.ts
```diff
@@ -297,22 +297,37 @@ export class PDFProcessor {
     try {
```
Tara-ag, line 297: 💡 **MINOR**: The `getAccuratePageCount` timer fix is good, but the `timeoutHandle.unref?.()` call is made immediately after `setTimeout`. If `setTimeout` returns a `Timeout` object (Node.js), `unref()` is available. However, the type `ReturnType<typeof setTimeout>` in Node's DOM-like typings is `Timer`/`Timeout`, and `unref` is optional. The `?.` handles that. Fine.

A more important point: the `finally` block clears the timer *after* `await pdf.getInfo()` has resolved, but the timer callback rejects the race promise. If the timer fires (i.e. `getInfo()` loses), the `Promise.race` rejects with "page-count timeout", control jumps to the outer `catch`, and the `finally` block still runs and clears the timer. Good. If `getInfo()` wins, the timer is cleared before `destroy()` is called. Good.

One subtle issue: the `timeoutHandle` is declared in the outer `try` scope but assigned inside the inner `new Promise` executor. If the executor throws synchronously (unlikely here), `timeoutHandle` would remain `undefined` and the `finally` would skip clearing. That's fine.

No action needed here — just confirming the logic is sound. The fix addresses the dangling timer and throwing destroy correctly.

Tara-ag (approved):
## Review Summary

Reviewed the 13 changed files in PR #1211. Loaded existing review comments first; no duplicate points were raised.

### New issues found

Raised **6 inline comments** across 6 files:

- **src/cli/commands/proxy.ts** — `getProcessStartTime` uses `execFileSync` inside an `async` function, blocking the event loop during uninstall and lacking a timeout bound. Suggested switching to async `execFile` wrapped with `withTimeout`.
- **src/cli/factories/commandFactory.ts** — Batch `pdfOptions` is constructed with a possibly-`undefined` password, creating a subtle inconsistency with `generate`/`stream`. Suggested aligning the conditional construction.
- **src/lib/utils/logSanitize.ts** — The URL param denylist includes overly broad generic names (`token`, `signature`, `expires`) that are also common non-secret query params, risking incorrect cache-key normalization. Suggested narrowing to provider-specific/auth-shaped params.
- **src/lib/utils/fileDetector.ts** — `cacheKeyForUrl` strips sensitive params but does not strip tracking params, so URLs differing only by tracking params get distinct cache keys unlike `ImageCache`.
- **src/lib/utils/imageCache.ts** — The content-hash dedup index is overwritten to the newest URL, making the older mapping unreachable; suggested keeping the oldest/stablest mapping.
- **src/lib/utils/messageBuilder.ts** — Aggregate PDF page-limit warning does not name which file(s) had unknown page counts, making the log less actionable.

Also left a confirmation note on `src/lib/utils/pdfProcessor.ts` that the race-timer/destroy fix is logically sound.

### Blocking assessment

None of the newly raised issues meet the formal blocking criteria in the project standards: no hardcoded secrets, no critical security vulnerability, no new backward-compatibility break, no non-negotiable CLAUDE.md rule violation, and fewer than 3 MAJOR issues. The PR consolidates legitimate post-merge fixes and adds regression tests.

### Recommendation

Approve. The two highest-impact items to address in a follow-up are the async `ps` call in `proxy.ts` and the breadth of the URL param denylist in `logSanitize.ts`.

src/cli/commands/proxy.ts
murdore: Fixed in `dba96e356eb19dee69692ef82ac95c0b5e8d479b`. Both `ps` calls (`getProcessStartTime`'s `lstart=` and `processLooksLikeProxySupervisor`'s `args=`) now go through a shared `async readPsField(pid, field)` using `promisify(execFile)` wrapped in `withTimeout(..., 2000)` — no more `execFileSync` blocking the event loop, and `ps` can't hang the uninstall path. Fallback semantics unchanged (null / fail-open).

src/cli/factories/commandFactory.ts
murdore: Thanks — and agreed with your re-read: the shell-history warning only fires on a truthy `flagValue`, so env-only doesn't warn. On the residual `pdfOptions: { password: undefined }` when neither flag nor env is set: this is intentional and consistent with `generate`/`stream`, which build the exact same `pdfOptions: { password: resolvePdfPassword(argv) }` unconditionally when PDFs are attached. An `undefined` password is a no-op in the SDK's PDF path (it just means 'no decryption key'), so batch now matches the other two commands rather than diverging.

src/lib/utils/logSanitize.ts
murdore: Good point to raise, but the hash-suffix design (from the earlier round) already neutralizes the collision/miss concern: denylisted params aren't just stripped — their sorted `k=v` values are folded into a deterministic `sha256`-based suffix on the key. So a resource that legitimately varies by e.g. `?token=abc` vs `?token=def` still gets two distinct, stable keys (different hash), and `?token=abc` always maps to the same key (cache hit). There's no extra miss and no collision from including generic-but-sometimes-sensitive names; the only effect is the value is hashed rather than plaintext in the key. I've kept the broader list deliberately — under-stripping risks leaking a real Azure SAS (`se/sp/sr/sv/sig`) or token, while over-stripping into the hash is functionally free.

src/lib/utils/fileDetector.ts
murdore: Fixed in `dba96e356eb19dee69692ef82ac95c0b5e8d479b`. Factored a shared `stripTrackingParams(parsed)` into logSanitize (utm_*/fbclid/gclid/_ga) and applied it in `fileDetector.cacheKeyForUrl` before the sensitive-param strip, matching `imageCache.normalizeUrl` — so two URLs differing only by tracking params now share a cache key.

src/lib/utils/imageCache.ts
murdore: Agreed this is a real (if narrow) dedup edge case, and as you note it's pre-existing — not introduced by this PR, which only changed the cache *key* normalization, not the `contentHashIndex` update logic. To keep this PR scoped to the post-merge review gaps it set out to close, I'm tracking the lossy content-hash-index update (keep oldest / keep a set of URLs so dedup survives single-URL eviction) as a separate follow-up rather than expanding scope here.

src/lib/utils/messageBuilder.ts
murdore: Fixed in `dba96e356eb19dee69692ef82ac95c0b5e8d479b`. The warn now filters the unknown-count files and names them: `N file(s) have an unknown page count (<filenames>)`, so the diagnostic is actionable instead of 'at least one'. Enforcement on the known total is unchanged.

src/lib/utils/pdfProcessor.ts
murdore: Thanks for the careful trace — agreed with your conclusion that no action is needed on the timer path: the `finally` clears/unrefs the handle on both the getInfo-wins and timer-wins branches, and a synchronous executor throw (which can't happen here) would just skip a clear on a handle that never got assigned. Resolving as analyzed.

src/cli/commands/proxy.ts
```diff
@@ -202,23 +210,113 @@ function getProcessStatus(pid: number): "running" | "not_running" | "unknown" {
...
+ */
+async function getProcessStartTime(pid: number): Promise<Date | null> {
+  const out = await readPsField(pid, "lstart");
+  if (!out) {
+    return null;
```
Tara-ag, line 256: ⚠️ **MAJOR**: `ps -o lstart=` returns the process start time in the system's **local timezone** (e.g. `"Sun Jul 19 20:12:41 2026"`), but `new Date(out)` parses it as local time while the persisted `expectedStartTimeIso` is parsed from an ISO 8601 UTC string. On any machine not running UTC, the two `Date` objects will differ by the timezone offset (e.g. 5.5 hours for IST), so `driftMs` will exceed the 10 s tolerance and `processLooksLikeProxySupervisor` will return `false` for a legitimate supervisor. During uninstall this leaves the real supervisor running and orphans its socket.

Fix by appending the local timezone offset to the `lstart` output before parsing, so both timestamps resolve to the same absolute instant:

```ts
async function getProcessStartTime(pid: number): Promise<Date | null> {
  const out = await readPsField(pid, "lstart");
  if (!out) {
    return null;
  }
  // ps -o lstart= emits local time without a zone. Append the current
  // local offset so Date parses it to the correct absolute millis.
  const offsetMin = new Date().getTimezoneOffset();
  const sign = offsetMin <= 0 ? "+" : "-";
  const absOffset = Math.abs(offsetMin);
  const hh = String(Math.floor(absOffset / 60)).padStart(2, "0");
  const mm = String(absOffset % 60).padStart(2, "0");
  const parsed = new Date(`${out} ${sign}${hh}:${mm}`);
  return Number.isNaN(parsed.getTime()) ? null : parsed;
}
```

(Using the current offset is safe here because the supervisor was started moments ago, so DST transitions are not a concern.)

Tara-ag (changes requested):
## Review Summary

**Files reviewed:** 13  
**Existing unresolved MAJOR issue:** 1

### Blocking issue

- **⚠️ MAJOR — `src/cli/commands/proxy.ts:256`**: `getProcessStartTime` parses `ps -o lstart=` output with `new Date(out)` without accounting for the system's local timezone. The `lstart` value is emitted in local time, while the persisted `expectedStartTimeIso` is an ISO 8601 UTC string. On any machine not running UTC, the two `Date` objects will differ by the timezone offset (e.g., 5.5 hours for IST), causing `driftMs` to exceed the 10 s tolerance and `processLooksLikeProxySupervisor` to return `false` for a legitimate supervisor. During uninstall this leaves the real supervisor running and orphans its socket.

  This issue is already tracked in the pending inline comment at `src/cli/commands/proxy.ts:256`. The fix is to append the current local timezone offset to the `lstart` output before parsing, so both timestamps resolve to the same absolute instant.

### Previously addressed issues

All other review points from earlier rounds have been resolved by the author:

- Batch parity: `--pdf-password` / `NEUROLINK_PDF_PASSWORD` is now resolved once per run and wired into `pdfOptions`; `--csv-max-rows` is validated; full `csvOptions` and `videoOptions` are built via shared helpers.
- PDF hardening: aggregate page-limit warns when page counts are unknown; `getAccuratePageCount` clears its race timer and guards a throwing `pdf.destroy()`.
- URL/secret handling: cache-hit debug log is redacted; `ImageCache` and `FileDetector` URL caches strip sensitive query params and fold a hash of the stripped values into the key to avoid cross-auth-context collisions; non-2xx HEAD responses no longer trust misleading `content-length`.
- Proxy supervisor identity now requires both `"neurolink"` and `"proxy"` in args and cross-checks OS-reported start time.
- Breaking changes are documented in `docs/MIGRATION.md` with a policy note for future changes.
- Eight focused regression tests cover the claimed gaps.

### Recommendation

Please fix the `getProcessStartTime` timezone parsing issue, then this is good to approve.

src/cli/commands/proxy.ts
murdore: Verified and hardened in `11be8f60ac2b574eb5e4a1f21d7304e38b7d6e1a`. Empirically, `new Date("Sun Jul 19 20:12:41 2026")` already parses as **local** time in V8 (on this UTC+5:30 box it yields `2026-07-19T14:42:41.000Z`, drift 0 ms vs the ISO-UTC persisted `startTime`), so there was no real TZ skew — and appending the local offset as suggested would have *double-applied* it and introduced the skew. But relying on the engine's implementation-defined parsing of a non-ISO ctime string is fragile, so `getProcessStartTime` now parses the `ps -o lstart=` components explicitly and builds the Date via the local-time constructor `new Date(y, mon, d, h, m, s)`. Same absolute instant, no dependency on the parsing heuristic, timezone-safe by construction.

src/cli/commands/proxy.ts
```diff
@@ -202,23 +210,149 @@ function getProcessStatus(pid: number): "running" | "not_running" | "unknown" {
...
```
Tara-ag, line 256: 🔒 **CRITICAL**: `getProcessStartTime` parses `ps -o lstart=` output using the **local-time** `Date` constructor (`new Date(y, mon, d, h, m, s)`), which builds a timestamp in the machine's local timezone. The persisted `ProxySupervisorState.startTime` is an ISO 8601 UTC string parsed with `new Date(isoUtcString)`, yielding an absolute UTC timestamp. On any machine not running UTC, `actual.getTime()` and `expected.getTime()` will differ by the timezone offset, so `driftMs` will exceed the 10 s tolerance and `processLooksLikeProxySupervisor` will return `false` for a legitimate supervisor. During uninstall this leaves the real supervisor running and orphans its socket.

Fix by treating the parsed local-time components as local and comparing against the persisted UTC instant correctly. The simplest robust fix is to append the current local timezone offset to the `lstart` string before parsing, or to convert the parsed local Date to UTC millis by subtracting its own timezone offset:

```ts
const parsed = new Date(
  Number(m[6]), monthIndex, Number(m[2]),
  Number(m[3]), Number(m[4]), Number(m[5]),
);
// parsed was built from local-time components; convert to UTC millis.
const utcMs = parsed.getTime() - parsed.getTimezoneOffset() * 60_000;
return Number.isNaN(utcMs) ? null : new Date(utcMs);
```

This makes the drift comparison timezone-safe.

src/cli/commands/proxy.ts
```diff
@@ -202,23 +210,149 @@ function getProcessStatus(pid: number): "running" | "not_running" | "unknown" {
   }
```
Tara-ag, line 210: ⚠️ **MAJOR**: `readPsField` dynamically imports `node:child_process` and `node:util` on every call, then re-promisifies `execFile` each time. During uninstall this is called twice for the same pid (once for `args`, once for `lstart`), so the overhead is small but unnecessary. More importantly, the `withTimeout` wrapper rejects with a raw `Error` ("ps -o ... timed out ...") that is swallowed by the outer `catch`, which is fine, but the helper does not use `ErrorFactory` typed errors — inconsistent with the codebase's error-handling pattern.

Suggested fix: move the imports to module top level (or lazy-load once) and return `null` on timeout without constructing an intermediate Error:

```ts
import { execFile } from "node:child_process";
import { promisify } from "node:util";

const execFileAsync = promisify(execFile);

async function readPsField(pid: number, field: string): Promise<string | null> {
  try {
    const { stdout } = await withTimeout(
      execFileAsync("ps", ["-p", String(pid), "-o", `${field}=`], { encoding: "utf-8" }),
      2000,
      `ps -o ${field}= timed out for pid ${pid}`,
    );
    return stdout.trim() || null;
  } catch {
    return null;
  }
}
```

This removes per-call dynamic imports and aligns with the project's preference for typed/utilities, though the timeout message is informational only.

src/cli/commands/proxy.ts
```diff
@@ -202,23 +210,149 @@ function getProcessStatus(pid: number): "running" | "not_running" | "unknown" {
...
  * and signalled. Returns false when the process cannot be confirmed as ours.
+ *
+ * @param expectedStartTimeIso - The `ProxySupervisorState.startTime` (ISO
+ *   string) recorded when the supervisor we expect at `pid` was launched.
+ *   When provided, this is cross-checked against `pid`'s actual OS-reported
+ *   start time (see {@link getProcessStartTime}) — the args match alone
+ *   ("neurolink" + "proxy" both present) is not airtight, since an
+ *   unrelated process (e.g. a shell running this very test suite, or a
+ *   coincidentally-named script) could match it too. A pid recycled by such
+ *   a process would have a start time far from the recorded supervisor
```
Tara-ag, line 310: ⚠️ **MAJOR**: `processLooksLikeProxySupervisor` is now `export`ed solely for the regression test, which reaches into the module internals via `import { processLooksLikeProxySupervisor } from "../src/cli/commands/proxy.js"`. Exporting a private helper just to test it violates the project's layering: it exposes an internal uninstall-path identity check as public API surface. The test already spawns real child processes and could instead exercise the public `ensureSupervisorStoppedBeforeClear` path (or a test-only wrapper), or the helper could remain private and the test could use a local eval script that mimics the same args/startTime scenario through the public command surface.

Suggested fix: keep `processLooksLikeProxySupervisor` private (remove `export`) and test it indirectly, or add a clearly-marked test-only export guarded by `NODE_ENV=test` / `process.env.NEUROLINK_TEST` so it does not become accidental public API.

src/lib/utils/logSanitize.ts
```diff
@@ -239,6 +240,134 @@ export function redactUrlForError(url: string): string {
...
```
Tara-ag, line 278: ⚠️ **MAJOR**: The `SENSITIVE_URL_QUERY_PARAM_DENYLIST` includes overly broad generic names (`token`, `signature`, `expires`) that are frequently used as ordinary content-varying query parameters. Because `stripSensitiveUrlParamsForCacheKey` folds a hash of stripped values into the cache key, collisions are avoided, but the design still forces every cache lookup for URLs with these params to compute a SHA-256 hash and stores an extra suffix in the Map key. More importantly, `token` is a generic word: a public API using `?token=abc` as an opaque resource identifier will be treated as a secret and stripped from the normalized URL used for cache lookups, even though it is not sensitive. This is a semantic mismatch that can surprise callers and complicates future debugging.

Suggested fix: narrow the denylist to provider-specific auth param names and avoid treating generic resource identifiers as secrets. Remove `token`, `signature`, and `expires` from the shared denylist; rely on `x-amz-*`, `x-goog-*`, Azure SAS (`sig`/`se`/`sp`/`sr`/`sv`), and explicit `access_token`/`auth_token`/`id_token` entries instead. If generic `token` must remain for compatibility, document the trade-off explicitly and add a regression test showing a non-auth `?token=public-id` is preserved.

src/lib/utils/logSanitize.ts
```diff
@@ -239,6 +240,134 @@ export function redactUrlForError(url: string): string {
...
+ * returned suffix whenever any were present, so distinct auth contexts keep
+ * distinct cache keys. Two requests presenting the exact same secret value
+ * still collide (correct — that's genuinely the same authorized request);
+ * two different secret values for the same object no longer do.
+ *
+ * Callers combine the return value with the now-stripped `parsed.toString()`
+ * to form the final cache key (see `ImageCache.normalizeUrl` and
+ * `fileDetector.cacheKeyForUrl`), rather than this function returning the
+ * full key itself, so each caller can apply its own additional
+ * normalization (e.g. tracking-param removal) in between.
```
Tara-ag, line 340: 💡 **MINOR**: `stripSensitiveUrlParamsForCacheKey` lowercases the stripped key before hashing (`${key.toLowerCase()}=${value}`), which means the hash suffix is case-insensitive. That is fine for equality, but the function deletes the original mixed-case key from `parsed.searchParams` and returns a URL that no longer contains the original casing. If a downstream consumer expects the original param casing to be preserved in the normalized URL, it will be silently mutated. This is acceptable for a cache key, but the doc comment should mention that sensitive params are removed case-insensitively and their casing is not preserved.

Suggested fix: add a note to the JSDoc: "Matching is case-insensitive; the original param casing is not preserved in the returned URL."

src/lib/utils/fileDetector.ts
```diff
@@ -81,22 +86,51 @@ const urlContentTypeCache = new Map<
...
```
Tara-ag, line 85: 💡 **MINOR**: `cacheKeyForUrl` strips tracking params and sensitive params, but it does not normalize the URL in other ways that `ImageCache.normalizeUrl` might in the future (e.g. trailing slashes, default ports, case of host). The two cache-key builders are now consistent for tracking/sensitive params, but any future divergence will reintroduce mismatched behavior. Consider extracting a single shared `normalizeUrlForCache(url): string` helper in `logSanitize.ts` that both `ImageCache` and `fileDetector.ts` call, so the normalization logic stays centralized.

Suggested fix: add a shared helper:

```ts
export function normalizeUrlForCache(url: string): string {
  try {
    const parsed = new URL(url);
    stripTrackingParams(parsed);
    const authSuffix = stripSensitiveUrlParamsForCacheKey(parsed);
    return parsed.toString() + authSuffix;
  } catch {
    return url;
  }
}
```

Then `ImageCache.normalizeUrl` and `fileDetector.cacheKeyForUrl` can both delegate to it.

src/lib/utils/fileDetector.ts
```diff
@@ -1915,19 +1950,28 @@ export class FileDetector {
...
+        // (or silently pass) based on the wrong body. Treat any non-2xx HEAD
+        // as if the header were missing and fall through to the streaming
+        // GET guard below, which enforces maxSize independently either way.
+        if (head.statusCode >= 200 && head.statusCode < 300) {
+          const declaredLength = Number(head.headers["content-length"]);
+          if (Number.isFinite(declaredLength) && declaredLength > maxSize) {
+            throw new Error(
+              `File too large: ${formatFileSize(declaredLength)} (max: ${formatFileSize(maxSize)})`,
+            );
+          }
```
Tara-ag, line 1955: 🔒 **CRITICAL**: The `catch` block in the HEAD pre-flight swallows **all** non-"File too large" errors and logs them at debug level, including potential security-relevant failures (e.g. TLS certificate errors, DNS hijack indicators, SSRF redirects). While the URL is now redacted, silently ignoring every HEAD failure means a malicious or misconfigured endpoint can suppress the pre-flight size check entirely and force a full streaming GET. This is exactly the path an attacker would use to bypass the size guard.

Suggested fix: distinguish transient/unavailable HEAD (405, network unreachable, timeout) from suspicious failures. At minimum, log non-2xx HEAD status at `warn` rather than `debug`, and consider surfacing TLS/DNS errors as typed `ErrorFactory.networkError` so callers are aware the guard could not be applied. For example:

```ts
logger.warn(
  `[FileDetector] HEAD pre-flight failed for ${redactUrlForError(url)}: ${sanitizeErrorCause(error).message}`,
);
```

This preserves visibility without breaking the fall-through behavior.

src/cli/factories/commandFactory.ts
Tara-ag, line 4885: 💡 **MINOR**: `batchPdfPassword` is resolved once before the loop (good), but it is still typed as `string | undefined` and spread into `pdfOptions: { password: batchPdfPassword }` whenever `batchPdfFiles?.length` is truthy. If the user attached PDFs but supplied neither `--pdf-password` nor `NEUROLINK_PDF_PASSWORD`, the resulting object is `{ password: undefined }`. This matches `generate`/`stream` behavior, but the conditional spread means batch only passes `pdfOptions` when PDFs are present, whereas `generate`/`stream` always pass it. The behavior is effectively the same, but the inconsistency is brittle: a future refactor that adds more fields to `pdfOptions` (e.g. `pageRange`) will have to remember to mirror it in three places.

Suggested fix: align batch with generate/stream by always constructing `pdfOptions` when PDFs are attached, even if the password is undefined, or extract a shared `buildPdfOptionsFromArgv(argv)` helper (mirroring `buildCsvOptionsFromArgv`/`buildVideoOptionsFromArgv`) so all three commands use the same shape.

src/lib/utils/messageBuilder.ts
```diff
@@ -1326,9 +1326,29 @@ async function processExplicitPdfFiles(
...
```
Tara-ag, line 1327: 💡 **MINOR**: The new aggregate page-limit warning logs filenames, which may include user-controlled or sensitive path segments. While `logger.warn` is not an error surface, the filenames come directly from caller input and could contain PII or internal paths. The project standards emphasize redacting raw user data such as CSV cell values from logs; filenames are less sensitive but still caller-controlled.

Suggested fix: consider truncating or basenaming the logged filenames, or document that the warning intentionally includes filenames for diagnostics. If the filename is already a basename at this point, the risk is minimal.

test/continuous-test-suite-bugfixes.ts
```diff
@@ -7776,6 +7780,651 @@ exit 127
       }
```
Tara-ag, line 7780: 💡 **MINOR**: Several new regression tests rely on source-code string matching (e.g. `compact.includes("constbatchPdfPassword=...")`) rather than behavioral assertions. While the comments note this is intentional to scope the check to `executeBatch`, source-string tests are fragile: any whitespace, variable rename, or refactor will break them even if behavior is correct. The earlier review threads already flagged and partially fixed the fixed-window brittleness, but the fundamental approach remains a maintenance burden.

Suggested fix: where possible, replace string-in-source assertions with behavioral tests. For example, the PDF-password batch parity could be verified by inspecting the options object passed to `runBatchGenerate` via a mock/spy, or by running the built CLI with `--dry-run` and checking that encrypted PDFs are accepted. At minimum, keep the source-scan tests but add a comment explaining why a behavioral test is not feasible here.

Tara-ag (changes requested):
## Review Summary

Reviewed all 13 changed files. The PR addresses real post-merge gaps and the test coverage is welcome, but two critical issues and three major issues remain that must be fixed before merge.

### New blocking issues raised

**CRITICAL (2)**
1. `src/cli/commands/proxy.ts:256` — `getProcessStartTime` builds a local-time `Date` and compares it directly against a UTC ISO timestamp. On non-UTC hosts the drift will exceed the tolerance and uninstall will fail to stop a legitimate supervisor, orphaning its socket.
2. `src/lib/utils/fileDetector.ts:1966` — The HEAD pre-flight `catch` swallows **all** non-size errors at `debug` level, including TLS/DNS/network anomalies, allowing a malicious or misconfigured endpoint to force a full streaming GET and bypass the size guard.

**MAJOR (3)**
1. `src/cli/commands/proxy.ts:210` — `readPsField` re-imports `node:child_process`/`node:util` and re-promisifies `execFile` on every call; should be module-level and use `ErrorFactory` typed errors consistently.
2. `src/cli/commands/proxy.ts:310` — `processLooksLikeProxySupervisor` is exported solely for the regression test, exposing internal uninstall-path logic as public API.
3. `src/lib/utils/logSanitize.ts:278` — The sensitive-param denylist includes overly broad generic names (`token`, `signature`, `expires`) that are commonly ordinary content-varying query params, creating a semantic mismatch and unnecessary cache-key mutation.

**MINOR (5)**
- `src/lib/utils/logSanitize.ts:340` — case-insensitive stripping mutates original param casing without documenting it.
- `src/lib/utils/fileDetector.ts:86` — `cacheKeyForUrl` and `ImageCache.normalizeUrl` should share a single `normalizeUrlForCache` helper to prevent future drift.
- `src/cli/factories/commandFactory.ts:4885` — batch `pdfOptions` construction is inconsistent with generate/stream; consider a shared `buildPdfOptionsFromArgv` helper.
- `src/lib/utils/messageBuilder.ts:1326` — aggregate page-limit warning logs raw filenames (caller-controlled data); consider basenaming or documenting.
- `test/continuous-test-suite-bugfixes.ts:7780` — several new tests rely on source-code string matching, which is fragile; prefer behavioral assertions where feasible.

### Decision

REQUEST_CHANGES — the two CRITICAL issues (supervisor identity timezone bug and silent HEAD failure bypass) and three MAJOR issues meet the blocking criteria. Once these are addressed, the remaining MINOR items can be handled at the author's discretion.

src/cli/commands/proxy.ts
murdore: This repeats the round-5 thread, and it's still a false positive — verified empirically. `new Date(2026, 6, 19, 20, 12, 41)` (the local-time constructor) and `new Date("...20:12:41 2026")` produce the **same absolute instant** (`2026-07-19T14:42:41.000Z` on this UTC+5:30 box), and uninstall runs on the SAME machine as the supervisor, so `getProcessStartTime` and the ISO-UTC persisted `startTime` resolve to the same instant → drift 0 ms, well within tolerance. There is no timezone offset between them to subtract; appending one (as suggested) would introduce the skew. ```node -e 'const p=new Date(2026,6,19,20,12,41); console.log(p.toISOString())'``` → `2026-07-19T14:42:41.000Z`.

src/cli/commands/proxy.ts
murdore: `node:child_process`/`node:util` are dynamically imported here to match the established pattern throughout proxy.ts (the uninstall handler, updater, etc. all `await import("node:child_process")`), not the provider-registry rule. `readPsField` runs at most twice per uninstall, so re-promisify overhead is negligible. It intentionally returns `null` on timeout/error (fail-open), so the swallowed error never surfaces — an ErrorFactory type would be unused. Kept as-is for consistency.

src/cli/commands/proxy.ts
murdore: `processLooksLikeProxySupervisor` is exported as a deliberate, narrow test seam (added an `@internal` note). The alternative you mention — a source/eval script mimicking the args+startTime scenario — is exactly the source-string-test style flagged as fragile in the sibling thread, and a purely behavioral test would need to spawn a real recycled-pid + persisted-state fixture, which is heavier and flakier than a direct unit call. The export is one function (not a type), on an internal module, guarded by `@internal`.

src/lib/utils/logSanitize.ts
murdore: Same as the round-4 thread on this line: the hash-suffix makes stripping these params collision-free and deterministic, so `?token=abc` as a public resource id still yields a stable, distinct key (cache hits work). The SHA-256 is computed only when a denylisted param is actually present (the common case has none), so it's not on every lookup. The tradeoff is deliberate: over-stripping a generic name into the hash is functionally free, while under-stripping risks leaking a real Azure SAS / signed-URL secret into a long-lived Map key.

src/lib/utils/logSanitize.ts
murdore: Fixed in `72725352d0604bcd6cb0ef334639cbd588719ef9` — the JSDoc now states the denylist match is case-insensitive and the returned key does not preserve original param-name casing (acceptable since it's a cache key, not a user-facing URL).

src/lib/utils/fileDetector.ts
murdore: Fixed in `72725352d0604bcd6cb0ef334639cbd588719ef9`. Extracted a single `normalizeUrlForCache(url)` in logSanitize.ts (parse → stripTrackingParams → sensitive-param auth-hash suffix), and both `ImageCache.normalizeUrl` and `fileDetector.cacheKeyForUrl` now delegate to it — byte-identical output, so future normalization changes stay centralized.

src/lib/utils/fileDetector.ts
murdore: The size guard is **not** bypassable here: when HEAD fails (405, TLS, DNS, timeout, redirect refusal), control falls through to the streaming GET, which enforces `maxSize` independently by aborting the body read once the limit is exceeded — that's the real guard; the HEAD pre-flight is only a fast-reject optimization. A genuinely suspicious failure (TLS cert error, DNS anomaly) on HEAD recurs on the GET to the same endpoint, so it's surfaced as a request failure, not silently suppressed. Nothing an attacker does to the HEAD response relaxes the enforced streaming cap.

src/cli/factories/commandFactory.ts
murdore: As noted on the round-4 thread: `pdfOptions: { password: undefined }` is intentional parity with generate/stream (undefined = no decryption key, a no-op in the SDK). The conditional spread (only when PDFs are attached) is correct for batch — batch has no non-file PDF mode to opt into — and doesn't change the effective behavior.

src/lib/utils/messageBuilder.ts
murdore: Fixed in `72725352d0604bcd6cb0ef334639cbd588719ef9` — logged filenames are now reduced to `basename(...)` before the warn, so no caller-supplied path segments (PII/internal paths) reach the log; count + names format unchanged.

test/continuous-test-suite-bugfixes.ts
murdore: The source-inspection tests are a deliberate, scoped choice for verifying option *wiring* (e.g. that executeBatch threads pdfOptions/csvOptions/validateCsvMaxRows) that's impractical to assert behaviorally without mocking the whole SDK generate path. The fixed-window brittleness you flagged earlier is already fixed (they scan to EOF and are scoped to executeBatch). The batch pdf-password/csv-max-rows tests that CAN be behavioral already are (they spawn the real built CLI and assert exit/stderr). I've kept the wiring checks as source assertions rather than add heavy SDK-mock scaffolding.

Tara-ag (approved):
## Review Summary

**Files reviewed:** 13 changed files
**New issues raised this run:** 0 (none)
**Existing review comments:** Audited all 27 existing threads; each has been addressed or justified by the author.

I walked the diff file-by-file against the project standards (CLAUDE.md Critical Rules 1–6, focus areas, and blocking criteria) and against the existing review threads. No new blocking issues remain:

- **Security:** Secrets are redacted before logging (`imageProcessor.ts`, `fileDetector.ts`); presigned-URL auth params are stripped and hash-suffixed for cache keys (`logSanitize.ts`, `imageCache.ts`, `fileDetector.ts`); HEAD pre-flight no longer trusts non-2xx `content-length`.
- **Architecture:** CLI-only proxy hardening stays in `src/cli/commands/proxy.ts`; no SDK/CLI boundary leak. Dynamic imports in the CLI command are outside the provider-registry scope of CLAUDE.md Critical Rule 1.
- **Backward compatibility:** The PR documents already-shipped breaking changes retroactively; it does not introduce new public-API breaks.
- **Correctness:** Batch now wires `pdfOptions`, `csvOptions`, and `videoOptions` through the same shared helpers as `generate`/`stream`; PDF aggregate page-limit warns on unknown counts; `getAccuratePageCount` clears its race timer and guards a throwing `destroy()`.
- **Tests:** 8 new regression tests cover the claimed gaps.

All previously raised concerns (broad denylist entries, exported test seam, `readPsField` dynamic imports, source-string assertions, `pdfOptions: { password: undefined }` parity, timezone/start-time parsing) have either been fixed in subsequent commits or answered with a coherent justification in the thread. I am not re-raising them.

**Decision:** Approve.

PR #1215, opened 2026-07-20
fix(cache): keep image content-hash dedup index resilient across eviction (#1213)

### LINKED ISSUE #1213: ImageCache: content-hash dedup index update is lossy across evictions
## Summary
`ImageCache.set()`'s content-hash dedup index update is lossy across evictions: when the same image content is cached under two different URLs, the `contentHashIndex` is repointed to the *newest* URL and the previous mapping is discarded.

Flagged during the post-merge review audit (fix PR #1211) review; **pre-existing**, not introduced by that PR (which only changed cache-*key* normalization, not the dedup index logic).

## Details
In `src/lib/utils/imageCache.ts` `set()`, when a content hash already exists under URL A and the same content arrives under URL B:
- a shallow copy of A's entry is stored under B, and
- `contentHashIndex[hash]` is updated to point to **B**, dropping the A mapping.

Consequences:
- `getByContentHash()` now only finds B.
- If B's entry is later **evicted or expires**, A's entry still sits in `this.cache` but is no longer reachable via content-hash dedup (orphaned for dedup purposes).
- (If A is evicted first, the index still points to B — fine.)

## Impact
Minor: reduced dedup benefit / a stranded cache entry after uneven eviction. Not a correctness or security problem. The cache is also disabled by default (`NEUROLINK_IMAGE_CACHE_ENABLED`).

## Proposed fix
Make the content-hash index resilient to single-URL eviction — e.g. keep it pointing at the **oldest** cached URL for that hash, or store a **Set of URLs** per hash and prune entries on eviction/expiry so the index never points at a gone entry.

murdore:
Closes #1213.

## Problem
`ImageCache.contentHashIndex` was `Map<contentHash, url>` — one url per content hash. When the same image content was cached under two URLs, `set()` repointed the index to the newest url, dropping the older mapping. If that newest url's entry was later evicted (`evictOldest`) or `delete()`d, the whole hash→url mapping was removed even though an older url's entry for the same content still lived in the cache — so `getByContentHash()` returned `null` and the dedup benefit was lost across eviction. Pre-existing; flagged during the #1211 post-merge audit.

## Fix
`contentHashIndex` is now `Map<contentHash, Set<url>>`:
- `set()` adds the new url to the hash's set (doesn't replace).
- `getByContentHash()` returns the first url in the set whose entry still lives (pruning gone urls; dropping the hash key only when the set empties).
- `delete()` / `evictOldest()` remove only that url from the set, deleting the hash key only when empty.

Single-url behavior is unchanged; only the multi-url-same-content path is now eviction-resilient.

## Validation
- New regression test `ImageCache #1213: content-hash dedup index survives deletion of one of two same-content urls` — caches the same content under two URLs, deletes the second, and asserts `getByContentHash` still resolves the first (returned `null` before the fix).
- Full bugfixes suite **229/229**, `tsc --strict` 0 errors, `lint` 0 errors, `build:cli` green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image caching when multiple URLs reference identical image content.
  * Prevented valid cached images from becoming undiscoverable after deleting or evicting another URL.
  * Fixed cache consistency during capacity-based replacement and deduplication.
* **Tests**
  * Added coverage for shared-content caching, deletion, eviction, and replacement scenarios.

### DIFF
--- a/src/lib/utils/imageCache.ts
+++ b/src/lib/utils/imageCache.ts
@@ -30,7 +30,11 @@ import type {
  */
 export class ImageCache {
   private cache = new Map<string, CachedImage>();
-  private contentHashIndex = new Map<string, string>(); // contentHash -> url
+  // contentHash -> set of normalized urls whose cache entry holds that
+  // content. A Set (not a single url) so that evicting/deleting one url
+  // doesn't drop the dedup mapping for other urls still caching the same
+  // content (#1213).
+  private contentHashIndex = new Map<string, Set<string>>();
   private maxSize: number;
   private ttlMs: number;
   private maxImageSize: number;
@@ -215,11 +219,40 @@ export class ImageCache {
    * Useful for deduplication when the same image is accessed via different URLs
    */
   getByContentHash(contentHash: string): CachedImage | null {
-    const url = this.contentHashIndex.get(contentHash);
-    if (!url) {
+    const urlSet = this.contentHashIndex.get(contentHash);
+    if (!urlSet) {
       return null;
     }
-    return this.get(url);
+
+    // Find the first url whose cache entry is still live, pruning any url
+    // whose entry no longer exists in `this.cache` along the way. Entries
+    // that exist but are expired are left in the set (not "gone") - the
+    // normal TTL/get() path is responsible for reaping those.
+    let liveUrl: string | null = null;
+    // Iterate a copy so reaping below can mutate the set safely.
+    for (const url of [...urlSet]) {
+      const entry = this.cache.get(url);
+      if (!entry) {
+        urlSet.delete(url);
+        continue;
+      }
+      if (this.isExpired(entry)) {
+        // Reap expired entries here too (matching get() semantics) so stale
+        // urls don't accumulate when callers primarily use getByContentHash.
+        this.cache.delete(url);
+        urlSet.delete(url);
+        continue;
+      }
+      if (liveUrl === null) {
+        liveUrl = url;
+      }
+    }
+
+    if (urlSet.size === 0) {
+      this.contentHashIndex.delete(contentHash);
+    }
+
+    return liveUrl ? this.get(liveUrl) : null;
   }
 
   /**
@@ -253,21 +286,42 @@ export class ImageCache {
     // Generate content hash
     const contentHash = this.generateContentHash(imageData);
 
-    // Check if same content already exists under different URL
-    const existingUrl = this.contentHashIndex.get(contentHash);
-    if (existingUrl && existingUrl !== normalizedUrl) {
-      // Content already cached under different URL - create a shallow copy
-      const existingEntry = this.cache.get(existingUrl);
-      if (existingEntry && !this.isExpired(existingEntry)) {
-        // Create a shallow copy for the new URL to avoid shared reference issues
-        this.cache.set(normalizedUrl, { ...existingEntry });
-        // Update content hash index to point to the new URL as well
-        this.contentHashIndex.set(contentHash, normalizedUrl);
-        logger.debug("Image cache dedup hit", {
-          newUrl: redactUrlForError(normalizedUrl),
-          existingUrl: redactUrlForError(existingUrl),
-        });
-        return;
+    // Check if same content already exists under a different URL
+    const existingUrls = this.contentHashIndex.get(contentHash);
+    if (existingUrls) {
+      for (const existingUrl of existingUrls) {
+        if (existingUrl === normalizedUrl) {
+          continue;
+        }
+        // Content already cached under a different URL - create a shallow copy
+        const existingEntry = this.cache.get(existingUrl);
+        if (existingEntry && !this.isExpired(existingEntry)) {
+          // A dedup hit still adds a cache entry for the new URL, so enforce
+          // capacity first — otherwise many URLs sharing one content hash could
+          // grow the cache past maxSize.
+          while (this.cache.size >= this.maxSize && this.cache.size > 0) {
+            this.evictOldest();
+          }
+          // Create a shallow copy for the new URL to avoid shared reference issues
+          this.cache.set(normalizedUrl, { ...existingEntry });
+          // Re-fetch (or re-create) the hash's URL set: the eviction above may
+          // have removed this hash's last url, emptying the set and deleting it
+          // from contentHashIndex — reusing the stale `existingUrls` reference
+          // would add the new url to an orphaned set that the index no longer
+          // points to, leaving a live entry unreachable by content hash
+          // (the exact #1213 failure mode). Never replace an existing mapping.
+          let hashUrls = this.contentHashIndex.get(contentHash);
+          if (!hashUrls) {
+            hashUrls = new Set<string>();
+            this.contentHashIndex.set(contentHash, hashUrls);
+          }
+          hashUrls.add(normalizedUrl);
+          logger.debug("Image cache dedup hit", {
+            newUrl: redactUrlForError(normalizedUrl),
+            existingUrl: redactUrlForError(existingUrl),
+          });
+          return;
+        }
       }
     }
 
@@ -288,7 +342,12 @@ export class ImageCache {
     };
 
     this.cache.set(normalizedUrl, entry);
-    this.contentHashIndex.set(contentHash, normalizedUrl);
+    let urlSet = this.contentHashIndex.get(contentHash);
+    if (!urlSet) {
+      urlSet = new Set<string>();
+      this.contentHashIndex.set(contentHash, urlSet);
+    }
+    urlSet.add(normalizedUrl);
 
     logger.debug("Image cached", {
       url: redactUrlForError(normalizedUrl),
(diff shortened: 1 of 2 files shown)


Tara-ag (approved):
## Review Summary

Reviewed **2 files** in this PR:
- `src/lib/utils/imageCache.ts`

**None.** No new blocking, major, minor, or suggestion-level issues were identified beyond the points already raised by the earlier automated review.

### Existing review comments
There are 3 unresolved comments from `copilot-pull-request-reviewer` that identify real concerns and should be addressed before merging:
1. **`set()` dedup-hit path bypasses `maxSize`** — can grow the cache beyond its configured limit when many URLs share the same content hash.
2. **`getByContentHash()` does not reap expired entries** — stale URLs can accumulate in the hash set if callers use the content-hash path exclusively.
3. **Missing eviction regression test** — the PR description claims resilience via `evictOldest()`, but the new test only exercises `delete()`.

These are pre-existing/unaddressed review points, not duplicates from this run.

### Standards check
- No CLAUDE.md Critical Rule violations (dynamic imports, Gemini/tools+schema, CLI/SDK separation, backward compatibility, `formatProviderError` semantics, etc. are unaffected by this change).
- No public SDK API breakage.
- Type safety and style appear consistent with the surrounding code.

**Decision:** Approve, contingent on the author resolving the three outstanding review comments above.

src/lib/utils/imageCache.ts
```diff
@@ -253,21 +278,28 @@ export class ImageCache {
...
+    if (existingUrls) {
+      for (const existingUrl of existingUrls) {
+        if (existingUrl === normalizedUrl) {
+          continue;
+        }
+        // Content already cached under a different URL - create a shallow copy
+        const existingEntry = this.cache.get(existingUrl);
+        if (existingEntry && !this.isExpired(existingEntry)) {
+          // Create a shallow copy for the new URL to avoid shared reference issues
+          this.cache.set(normalizedUrl, { ...existingEntry });
```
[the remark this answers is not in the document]
murdore: Fixed in `e201bab4ddaf6a97be81fca97426f92e547277dc`. The dedup-hit path in `set()` now runs the same capacity check (`while (cache.size >= maxSize && cache.size > 0) evictOldest()`) before inserting the copied entry, so a dedup hit can no longer grow the cache past maxSize (e.g. maxSize=1 with two same-content urls).

src/lib/utils/imageCache.ts
```diff
@@ -215,11 +219,32 @@ export class ImageCache {
...
+    let liveUrl: string | null = null;
+    for (const url of urlSet) {
+      const entry = this.cache.get(url);
+      if (!entry) {
+        urlSet.delete(url);
+        continue;
+      }
+      if (liveUrl === null && !this.isExpired(entry)) {
+        liveUrl = url;
+      }
```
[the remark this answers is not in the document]
murdore: Fixed in `e201bab4ddaf6a97be81fca97426f92e547277dc`. `getByContentHash()` now reaps expired entries it encounters (removing them from both `cache` and the hash's url set), matching `get()` semantics, so stale urls don't accumulate for callers that primarily use `getByContentHash()`. It iterates a copy of the set so the reaping mutation is safe.

test/continuous-test-suite-bugfixes.ts
```diff
@@ -8425,6 +8425,69 @@ exit 127
...
+        const beforeDelete = cache.getByContentHash(contentHash);
+        if (!beforeDelete || beforeDelete.dataUri !== dataUri) {
+          return false;
+        }
+
+        // Delete the SECOND (newer) url. Pre-fix, set(urlB, ...) had
+        // repointed the single-url index to urlB, so deleting urlB wiped
+        // out the only mapping for this hash - getByContentHash() would
+        // return null even though urlA's entry is still live in the cache.
+        cache.delete(urlB);
```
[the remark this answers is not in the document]
murdore: Added in `e201bab4ddaf6a97be81fca97426f92e547277dc`. New focused test 'ImageCache #1213: dedup index survives EVICTION of the indexed url, not just deletion': caches the same content under two urls, bumps the first to MRU so the second is LRU, then adds a third distinct-content entry at maxSize=2 to force `evictOldest()` of the second — and asserts `getByContentHash()` still resolves the survivor. Returned null before the fix (the single-url index pointed at the evicted url).

src/lib/utils/imageCache.ts
```diff
@@ -288,7 +334,12 @@ export class ImageCache {
     };
```
Tara-ag, line 343: ⚠️ **MAJOR**: Orphaned `contentHashIndex` entry when capacity enforcement empties the hash's URL set.

`evictOldest()` may remove the last URL from `existingUrls` and, because the set becomes empty, delete the hash key from `contentHashIndex`. The code still holds a reference to the now-orphaned `Set` and mutates it with `existingUrls.add(normalizedUrl)`, but `contentHashIndex` no longer points to it. `getByContentHash()` then returns `null` even though a URL for that content is cached — reproducing the exact failure mode #1213 is meant to fix.

Repro: `maxSize=1`, `set(urlA, contentX)`, `set(urlB, contentX)`, then `getByContentHash(hashX)` returns `null`.

Fix: restore the index after adding the new URL:
```typescript
existingUrls.add(normalizedUrl);
this.contentHashIndex.set(contentHash, existingUrls);
```

Please also add a regression test for the `maxSize=1` dedup path (or any case where capacity enforcement evicts the only same-hash source).

test/continuous-test-suite-bugfixes.ts
```diff
@@ -8425,6 +8429,118 @@ exit 127
...
+        } else {
+          process.env.NEUROLINK_IMAGE_CACHE_ENABLED = originalEnv;
+        }
+        resetImageCache();
+      }
+    },
+  },
 ];
 
 // ============================================================================
```
Tara-ag, line 8587: 💡 MINOR: Test assertion is weaker than the bug it claims to cover.

The test checks `found.dataUri === dataUri` after eviction, but `getByContentHash` could return the evicted `urlB`'s stale entry if the index still points at it and `this.get(liveUrl)` doesn't re-verify expiry/LRU state. Because both A and B share the same `dataUri`, this assertion would pass even if the method returned B's ghost entry.

Strengthen the test to assert the survivor's URL:
```typescript
return (
  found !== null &&
  found.dataUri === dataUri &&
  cache.get(urlA) !== null &&
  cache.get(urlB) === null
);
```

Also consider asserting `cache.get(urlA).dataUri === dataUri` to confirm the survivor is truly live, not just indexed.

Tara-ag (changes requested):
## Review Summary

**Files reviewed:** 2  
**New issues raised:** 2 (1 MAJOR, 1 MINOR)

### Blocking issue

- **`src/lib/utils/imageCache.ts:335`** — The dedup-hit capacity enforcement can orphan the `contentHashIndex` entry. When `evictOldest()` removes the last URL from the hash's set, it deletes the hash key from `contentHashIndex`, but the code still holds a reference to the now-orphaned `Set` and mutates it. The index is no longer reachable, so `getByContentHash()` returns `null` even though a URL for that content is cached. This reproduces the exact failure mode #1213 is meant to fix (e.g. `maxSize=1` with two same-content URLs). A fix and a regression test for the `maxSize=1` dedup path are suggested inline.

### Non-blocking suggestion

- **`test/continuous-test-suite-bugfixes.ts:8546`** — The eviction regression test only asserts `found.dataUri === dataUri`, which is also true for the evicted URL's ghost entry because both URLs share the same content. Strengthen the assertion to verify the survivor's URL is returned.

### Existing review comments

All three existing Copilot review threads are resolved/outdated and are not re-raised here.

### Next steps

Please address the orphaned-index bug and consider tightening the eviction test assertion. Once the blocking issue is fixed, this should be good to merge.

src/lib/utils/imageCache.ts
murdore: Good catch — real bug, fixed in `09c59e298158bd33b9d159927f3c7522a625eaba`. After the capacity eviction in the dedup path, the code no longer reuses the possibly-detached `existingUrls` reference. It re-fetches the hash's set from `contentHashIndex` (which `evictOldest` may have emptied and deleted) and re-creates it if gone, then adds the new url to that LIVE set. Added a dedicated regression test ('dedup copy at capacity re-creates the hash index (no orphaned set)', maxSize=1) that reproduces exactly this: caching the dedup copy evicts the first url, emptying the hash set, and asserts `getByContentHash` still resolves the survivor (returned null before this fix).

test/continuous-test-suite-bugfixes.ts
murdore: The test already rules out a ghost B: it asserts `cache.get(urlB) === null` (B is provably evicted) alongside `cache.get(urlA) !== null`, and `getByContentHash()` only ever returns an entry that still exists in `cache` — it skips/prunes urls whose entry is gone before calling `this.get(liveUrl)`, so an evicted url can never be the resolved result. Since A and B share `dataUri` by construction (dedup copies the entry), asserting the URL state (A present, B absent) is the meaningful check, and it's present. The new orphan test (see the sibling thread) adds the mirror case.

Tara-ag (approved):
## Review Summary

**Files reviewed:** 2 (`src/lib/utils/imageCache.ts`, `test/continuous-test-suite-bugfixes.ts`)
**New issues raised:** 0
**Decision:** Approve

I re-read the current HEAD of both changed files and checked all existing review threads. The previously raised points have been addressed:

- The dedup-hit path in `set()` now enforces `maxSize` before inserting the copied entry.
- `getByContentHash()` now reaps expired entries it encounters.
- A focused eviction regression test was added.
- The orphaned-set bug in the capacity/dedup path was fixed by re-fetching/recreating the live hash set after eviction.
- The eviction test’s assertions are sufficient: `getByContentHash()` only returns entries still present in `this.cache`, and the test explicitly asserts the evicted URL is gone.

No new blocking issues, security concerns, or CLAUDE.md rule violations were found in this run. The fix is narrowly scoped, the new tests cover deletion, eviction, and the maxSize=1 orphan-set scenario, and the public `ImageCache` API is unchanged.

PR #1217, opened 2026-07-20
feat(proxy): persist usage stats across workers

murdore:
## Summary
- persist proxy usage counters as debounced deltas under an atomic cross-process lock, preserving totals across restart and rolling-worker overlap without request-path I/O
- reconcile `/status` with the shared snapshot, expose durability state, and distinguish removed, disabled, excluded, cooling, and active historical accounts
- flush counters after request drain during graceful shutdown, isolate dev state, quarantine corrupt snapshots, and retain deltas across lock/write failures
- add cross-process concurrency and failure-injection coverage plus an enforced usage-statistics latency, memory, flush, and restore benchmark

## Safety
- initialization runs after the foreground singleton guard, so a rejected second start cannot mutate live state
- request handlers only mutate memory and schedule one unref'ed debounce timer; disk work is asynchronous
- rolling workers merge additive deltas instead of overwriting shared totals

## Verification
- `pnpm typecheck`
- `pnpm exec vitest run --exclude test/helpers/envGuard.test.ts` (17 files, 347 tests)
- `pnpm run test:envguard` (118/118)
- focused proxy regression set (5 files, 126 tests)
- pre-commit `check`, full lint, validation, security scan, package build, browser build, and publint
- `pnpm run proxy:performance` (all gates pass; stats p95 0.084 us, flush 2.400 ms, exact restore; rolling handoff 0 dropped requests and active stream preserved)

## Summary by CodeRabbit

* **New Features**
  * Proxy usage statistics now persist across restarts with durability, reconciliation, and safe concurrent updates from multiple processes.
  * `/status` now includes per-account classifications (e.g., active/cooling/removed), `startedAt`, legacy-account mapping, and persistence health; table coloring was updated.
  * Added a `proxy:performance:stats` benchmark and integrated the stats step into `proxy:performance`.
* **Bug Fixes**
  * Improved shutdown reliability by flushing persisted usage statistics with a bounded timeout and warning on failure.
* **Tests**
  * Updated and expanded reliability, lifecycle ordering, persistence, corruption recovery, and concurrency coverage; adjusted helpers for async persistence resets.


src/lib/proxy/usageStats.ts
```diff
@@ -1,41 +1,757 @@
...
+    }
+    if (!owner && oldEnough) {
+      await rm(lockPath, { force: true });
+      return true;
+    }
+  } catch {
+    return true;
+  }
+  return false;
+}
```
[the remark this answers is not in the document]
murdore: Fixed in add7e34a. `removeAbandonedLock()` now treats only ENOENT as a removable/missing-lock retry; permission and other filesystem errors remain bounded by the acquisition timeout. It also recovers a live-PID lock only after the lock file exceeds `staleLockMs`, covering PID reuse without taking over a fresh active lock. Added focused tests for malformed stale locks, stale live-PID locks, active contention timeout/retry, and joining an in-flight flush.

test/proxyReliabilityHardening.test.ts
```diff
@@ -1766,6 +1770,11 @@ describe("launchd lifecycle source invariants", () => {
     expect(source).not.toContain("cooling: false");
     expect(source).toContain("catch(forceExitAfterShutdownFailure)");
     expect(source).toContain("Timed out draining the proxy server");
+    expect(
+      startHandler.indexOf("await ensureProxyStartAllowed(spinner)"),
+    ).toBeLessThan(
+      startHandler.indexOf("await initUsageStats(proxyPaths.statsFile)"),
+    );
```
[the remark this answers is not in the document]
murdore: Fixed in 9ad87d69. The test now captures both marker indices, asserts each is present, and only then verifies that the singleton guard precedes statistics initialization. Focused reliability tests (48/48) and typecheck pass.

murdore:
Reviewed the two non-blocking CodeRabbit review-body findings and addressed both in 6f3cdc44:

- `/status` reconciliation is now read-only. It waits for any same-process in-flight flush, then combines the latest persisted snapshot with local pending deltas without forcing a lock acquisition or snapshot write. A focused test verifies totals, unchanged revision, and retained pending mutations.
- Corrupt snapshot quarantine retention is capped at the three newest files, with focused repeated-recovery coverage.

Validation after the changes: 349/349 Vitest tests, 118/118 env-guard tests, typecheck, CLI build, full pre-commit validation/build/security checks, and all proxy performance budgets passed.

src/lib/proxy/usageStats.ts
```diff
@@ -1,41 +1,783 @@
...
+      this.pending,
+      accountLabel,
+      accountType,
+      status,
+      rateLimitKind,
+      failedAt,
+    );
+    this.markMutation();
+  }
+
```
Tara-ag, line 415: 💡 **resetStats() no longer resets persisted state — surprising semantics for tests/restarts**

`resetStats()` now delegates to `resetMemory()`, which clears the in-memory counters and cancels the flush timer but leaves `proxy-usage-stats.json` untouched. The next `initUsageStats(filePath)` will reload the old snapshot, so a caller that expected a clean slate will silently resurrect previous counters.

This is a behavioral change from the previous in-memory-only implementation where `resetStats()` effectively gave you fresh process state. It is also inconsistent with `resetMemory()`'s own name: the persisted file is arguably part of the "memory" users care about.

**Suggestion:** either rename this export to `resetMemory()` and keep `resetStats()` for a full reset (including deleting/quarantining the snapshot), or make `resetStats()` also remove the persisted file when `filePath` is set. If the current behavior is intentional, add a doc comment clarifying that on-disk state is preserved.

src/lib/proxy/usageStats.ts
```diff
@@ -1,41 +1,783 @@
...
+  lockPath: string,
+  timeoutMs: number,
+  staleLockMs: number,
+  now: () => number,
+): Promise<() => Promise<void>> {
+  await mkdir(dirname(lockPath), { recursive: true, mode: 0o700 });
+  const deadline = Date.now() + timeoutMs;
+  while (true) {
+    const owner: ProxyStatsLockOwner = {
+      token: randomUUID(),
```
Tara-ag, line 270: ⚠️ **Race: reconcile() merges a stale snapshot over newer local counters without checking revision**

`reconcile()` reads the file under `flushMutex`, then unconditionally does `this.stats = mergeStats(snapshot.stats, this.pending)`. If the local process has already flushed a newer revision (e.g. `this.revision > snapshot.revision` because another worker wrote after us), this will roll `this.stats` backwards by re-adding the old persisted totals on top of the current local totals.

Example:
1. Worker A flushes revision 5 (totalAttempts = 100).
2. Worker A records 10 more attempts locally (pending = 10, stats = 110).
3. Worker B flushes revision 6 (totalAttempts = 105).
4. Worker A reconciles: reads snapshot revision 6 (105), merges with pending 10 → stats = 115, which is correct here, but if Worker A had already flushed revision 6 itself and then reconciled after a slower read, it could double-count.

More importantly, because `mergeStats` is additive, any worker that has already incorporated a snapshot into `this.stats` and still holds the same events in `this.pending` will double-count them after `reconcile()`. `reconcile()` is called from `/status` (`getReconciledStats`), so repeated status reads during a busy worker can inflate counters.

**Suggested fix:** track a `lastMergedRevision` per store and only merge snapshots whose revision is greater than the last merged revision. Alternatively, clear `this.pending` after each successful flush so `reconcile()` cannot double-add already-flushed local events.

src/cli/commands/proxy.ts
```diff
@@ -459,6 +459,32 @@ async function resolveStatusPrimaryAccount(
   };
 }
 
+async function resolveLegacyStatusAccountLabel(
```
Tara-ag, line 466: 💡 **Legacy credentials read happens on every `/status` call**

`resolveLegacyStatusAccountLabel()` reads and parses `~/.neurolink/anthropic-credentials.json` synchronously every time `/status` is requested. This is a blocking filesystem call on the request path and can add latency/IOPS noise to status polling.

Since this is only needed when `storedAccountKeys.size === 0`, consider caching the result for the lifetime of the process, or reading it lazily once and invalidating only on auth changes. The current implementation is correct but not ideal for a route that may be polled frequently.

src/cli/commands/proxy.ts
```diff
@@ -2380,6 +2432,19 @@ function registerProxyShutdownHandlers(params: {
       }
     }
```
Tara-ag, line 2451: ⚠️ **Shutdown flush uses the same timeout bucket as lifecycle metadata flush**

`flushUsageStats()` and `flushProxyLifecycleEvents()` each get the full `PROXY_LIFECYCLE_SHUTDOWN_TIMEOUT_MS` (5 s). If both time out, the total shutdown delay can reach ~10 s plus drain time. More importantly, the usage-stat flush is placed *before* the lifecycle flush, so a slow/stuck stats lock can starve the lifecycle flush of its own budget.

**Suggested fix:** give the stats flush a dedicated, smaller budget (e.g. 2 s) and place it after the lifecycle flush, or run them concurrently with `Promise.all` and a shared deadline so neither can monopolize the timeout.

src/cli/commands/proxy.ts
```diff
@@ -2911,6 +2976,21 @@ async function startProxyCommandHandler(argv: ProxyStartArgs): Promise<void> {
     if (!isDev && !socketWorker) {
```
Tara-ag, line 2998: ✅ **Good: singleton guard precedes stats initialization**

`ensureProxyStartAllowed(spinner)` runs before `initUsageStats(proxyPaths.statsFile)`, matching the safety claim in the PR description and the new source-invariant test. This prevents a rejected second start from mutating the shared stats file.

src/lib/proxy/usageStats.ts
```diff
@@ -1,41 +1,783 @@
...
+  private filePath?: string;
+  private stats: ProxyStats;
+  private pending: ProxyStats;
+  private pendingMutations = 0;
+  private inFlightMutations = 0;
+  private revision = 0;
+  private flushTimer: ReturnType<typeof setTimeout> | null = null;
+  private readonly flushMutex = new AsyncMutex();
+  private lastFlushedAt?: number;
+  private lastReconciledAt?: number;
```
Tara-ag, line 320: ⚠️ **Double-counting risk when flush fails and is retried**

On flush failure, the code does:
```ts
this.pending = mergeStats(delta, this.pending);
this.pendingMutations += deltaMutations;
```
This restores the delta back into `pending`. However, `this.stats` was already updated with the delta when the mutation was originally recorded (`recordAttempt` etc. update both `stats` and `pending`). If the flush fails, `this.stats` retains the delta, but the next successful flush will read the persisted snapshot, merge `delta` again, and then `this.stats = mergeStats(merged, this.pending)` will merge `this.pending` (which again contains the same delta) a third time.

Wait — let me trace more carefully. On mutation, `stats` and `pending` both get +1. Flush swaps `pending` into `delta`, zeros `pending`, then writes `mergeStats(persisted, delta)`. On success it sets `stats = mergeStats(merged, this.pending)` where `pending` is now empty, so `stats` becomes `mergeStats(persisted, delta)`. Since `stats` previously already contained the in-memory delta, this replaces it with the persisted+delta merged value. If persisted already had some of the delta (from another worker), additive merge is correct.

On failure, `pending = mergeStats(delta, pending)` restores the delta. `stats` still has the delta. Next flush: writes `mergeStats(persisted, pending)` where pending includes the delta. Success: `stats = mergeStats(merged, empty)` = persisted+delta. That seems okay.

But the issue is `reconcile()` without flush: it does `stats = mergeStats(snapshot.stats, this.pending)` without clearing pending. If `pending` contains events that were already flushed by a previous successful flush (they shouldn't, because flush zeros pending on success), so that's fine. But if a flush failed, pending contains the delta, and reconcile will merge snapshot+delta into stats. Stats already had delta. So stats now has delta twice.

Actually after failed flush, stats has delta (from original mutation). reconcile merges snapshot+delta into stats → stats becomes snapshot + delta + delta. This is a real double-count bug for failed-flush scenarios followed by reconcile (i.e. `/status` call).

**Suggested fix:** in `reconcile()`, snapshot `this.pending` before merging and then zero it, or at least detect that the pending delta has already been applied to `this.stats`. Better: maintain a monotonic `appliedRevision` and only merge deltas that haven't been applied.

src/lib/proxy/proxyPaths.ts
```diff
@@ -25,6 +25,7 @@ export function resolveProxyPaths(dev: boolean): ProxyPaths {
       logsDir: join(base, "logs"),
       quotaFile: join(base, "account-quotas.json"),
       cooldownFile: join(base, "account-cooldowns.json"),
+      statsFile: join(base, "proxy-usage-stats.json"),
```
Tara-ag, line 28: ✅ **Dev/prod state isolation preserved**

Adding `statsFile` to both dev and prod `ProxyPaths` objects keeps the dev-mode `.neurolink-dev/` isolation intact, which is important for preventing test/dev runs from corrupting live usage counters.

src/lib/types/proxy.ts
```diff
@@ -811,6 +811,44 @@ export type ProxyStats = {
   accounts: Record<string, AccountStats>;
```
Tara-ag, line 811: ✅ **Type additions follow CLAUDE.md rules**

New types (`ProxyStatsPersistenceStatus`, `PersistedProxyStatsSnapshot`, `ProxyStatsLockOwner`, `ProxyUsageStatsStoreOptions`) are defined in `src/lib/types/proxy.ts`, use `type` not `interface`, and have domain-prefixed unique names. Good.

package.json
```diff
@@ -179,9 +179,10 @@
     "proxy:observability:doctor": "node scripts/observability/check-proxy-telemetry.mjs",
     "proxy:observability:import-dashboard": "bash scripts/observability/manage-local-openobserve.sh import-dashboard",
     "proxy:performance:lifecycle": "tsx tools/testing/proxyLifecycleBenchmark.ts",
+    "proxy:performance:stats": "tsx tools/testing/proxyUsageStatsBenchmark.ts",
```
Tara-ag, line 182: ✅ **Benchmark script integration**

Adding `proxy:performance:stats` to the composite `proxy:performance` pipeline is consistent with the existing lifecycle/transport/rolling benchmarks and gives CI a gate for the new persistence path.

test/proxyReliabilityHardening.test.ts
```diff
@@ -1766,6 +1770,15 @@ describe("launchd lifecycle source invariants", () => {
     expect(source).not.toContain("cooling: false");
     expect(source).toContain("catch(forceExitAfterShutdownFailure)");
     expect(source).toContain("Timed out draining the proxy server");
+    const startAllowedIndex = startHandler.indexOf(
```
Tara-ag, line 1783: ✅ **Source-order invariant correctly guards lifecycle**

The updated assertion captures both marker indices, validates their presence, and verifies the singleton guard precedes stats initialization. This addresses the earlier CodeRabbit feedback and prevents silent test passes if either marker is removed.

test/proxyUpdaterFallback.test.ts
```diff
@@ -225,6 +227,11 @@ describe("proxy runtime error finalization", () => {
       totalRateLimits: 1,
       totalTransientRateLimits: 1,
       totalQuotaRateLimits: 0,
+      persistence: {
```
Tara-ag, line 230: 💡 **Test assumes persistence is disabled but doesn't reset/init the store**

The assertion `persistence: { enabled: false, pendingMutations: 6, lastError: null }` relies on the default store never having been initialized with a file path in this test. Because `usageStats.ts` now uses a module-level `defaultStore`, test order matters: if a preceding test calls `initUsageStats(path)`, this test will see `enabled: true` and fail.

The `beforeEach` only calls `resetStats()`, which clears memory but does not reset `filePath`. Consider adding `__testHooks.resetUsageStatsStore()` or explicitly initializing the default store to a no-path state in the test setup to guarantee isolation.

test/proxyUsageStatsPersistence.test.ts
```diff
@@ -0,0 +1,470 @@
...
+        "old@example.com": { successCount: 1 },
+        "new@example.com": { errorCount: 1 },
+      },
+    });
+    expect(observer.getPersistenceStatus()).toMatchObject({
+      enabled: true,
+      revision: 2,
+      pendingMutations: 0,
+      lastError: null,
+    });
```
Tara-ag, line 160: ⚠️ **Cross-process test is fragile to cwd assumptions**

The test builds a `pathToFileURL` from `join(process.cwd(), "src/lib/proxy/usageStats.ts")`. If vitest is run from a different working directory (e.g. CI runs `pnpm vitest run` from repo root, which is fine, but `pnpm test` might run from elsewhere), this path will be wrong and the worker processes will fail to import.

**Suggested fix:** derive the module path from `import.meta.url` or `fileURLToPath(new URL(..., import.meta.url))` instead of `process.cwd()`, so the test is robust to the current working directory.

src/lib/types/proxy.ts
```diff
@@ -1040,6 +1078,8 @@ export type ProxyPaths = {
   quotaFile: string;
   /** account-cooldowns.json — restart-safe account cooldown state */
   cooldownFile: string;
+  /** proxy-usage-stats.json — restart- and handoff-safe usage counters */
```
Tara-ag, line 1081: ⚠️ **Breaking change to public `ProxyPaths` type**

Adding `statsFile: string` as a required field to `ProxyPaths` breaks any external caller or test that constructs a `ProxyPaths` object without this property. `ProxyPaths` is exported from the public types barrel and documented in the API reference, so this is a public SDK API surface change.

**Suggested fix:** make `statsFile` optional (`statsFile?: string`) and have consumers fall back to `join(stateDir, "proxy-usage-stats.json")` when absent, or document the migration path in the PR description/CHANGELOG.

Tara-ag (changes requested):
## Review Summary

Reviewed 9 changed files. Found **2 MAJOR** and **4 MINOR** new issues worth addressing before merge. I also left several positive notes on things done well (lifecycle ordering, dev-state isolation, type-rule compliance, benchmark integration, and the fixed source-order test).

### Blocking concern

**MAJOR — Breaking change to public `ProxyPaths` type (CLAUDE.md Critical Rule 5)**

`src/lib/types/proxy.ts` adds `statsFile: string` as a **required** field on the publicly documented `ProxyPaths` type. Any external caller or test constructing a `ProxyPaths` object without this property will now fail type checking. This violates the public SDK API backward-compatibility rule.

**Fix:** make `statsFile` optional (`statsFile?: string`) and have consumers fall back to `join(stateDir, "proxy-usage-stats.json")` when absent, or document a migration path.

### Other MAJOR

**MAJOR — Shutdown flush ordering can starve lifecycle metadata flush**

`flushUsageStats()` is placed before `flushProxyLifecycleEvents()` and both use the full `PROXY_LIFECYCLE_SHUTDOWN_TIMEOUT_MS` budget. A stuck stats lock can consume the entire shutdown timeout before the lifecycle flush runs.

**Fix:** give the stats flush a dedicated smaller budget (e.g. 2 s), or run the two flushes concurrently under a shared deadline.

### MINOR issues

1. **`resetStats()` now only clears memory, not persisted state** — surprising given the old in-memory semantics. Consider renaming to `resetMemory()` or making `resetStats()` also remove/quarantine the snapshot when a file path is set.
2. **`/status` reads legacy credentials from disk on every request** — blocking I/O on a polled endpoint. Cache the result per process.
3. **`proxyUpdaterFallback.test.ts` assumes persistence is disabled without resetting the default store** — test order can affect the `enabled: false` assertion because `resetStats()` only clears memory, not `filePath`.
4. **Cross-process benchmark test uses `process.cwd()` to locate the module** — fragile if vitest is invoked from a different working directory. Derive the path from `import.meta.url` instead.

### Corrections to my own earlier inline notes

Upon deeper tracing, the reconcile/flush double-counting concerns I raised in `usageStats.ts` are **not valid**: the invariant `stats = persisted_snapshot + pending` is maintained correctly across successful flushes, failed flushes, and reconciles. Please disregard those two specific threads.

### Next steps

1. Make `ProxyPaths.statsFile` optional (or document the migration).
2. Address the shutdown-timeout sequencing.
3. Consider the MINOR suggestions above.

Once the `ProxyPaths` breaking change is resolved, this should be in good shape to approve.

src/lib/types/proxy.ts
murdore: Fixed in d31e5c14. `ProxyPaths.statsFile` is optional again, preserving compatibility for external object literals. `resolveProxyUsageStatsPath()` supplies the deterministic `<stateDir>/proxy-usage-stats.json` fallback, and the startup path uses that helper. The reliability suite includes a compile-time/runtime object without `statsFile` and verifies the fallback.

src/cli/commands/proxy.ts
murdore: Fixed in d31e5c14. Usage-stat and lifecycle-metadata flushes now start together via `Promise.allSettled`, each under the same 5 s shutdown deadline, so total flush delay is bounded to one timeout window and neither flush can delay the other. Failures remain independently logged and non-fatal.

src/lib/proxy/usageStats.ts
murdore: Addressed in d31e5c14. `resetStats()` now explicitly documents that it resets only the process-local view while intentionally preserving durable history. Tests no longer use that ambiguous production helper: `resetUsageStatsForTests()` waits for any in-flight flush, clears counters, and disconnects the singleton from its persistence path so state cannot leak between tests.

src/cli/commands/proxy.ts
murdore: Fixed in d31e5c14. Legacy credential lookup now uses a per-credential-path 5 s cache, including negative results, so status polling does not perform a filesystem read on every request while reauthentication changes still become visible promptly. A behavioral test removes the credential file after the first status read and verifies the second read uses the cache.

test/proxyUpdaterFallback.test.ts
murdore: Fixed in d31e5c14. This suite and the other singleton consumers now await `resetUsageStatsForTests()` during cleanup. The helper serializes with an in-flight flush, clears counters, and removes the configured file path, so the `enabled: false` assertion no longer depends on test order or module reuse.

test/proxyUsageStatsPersistence.test.ts
murdore: Fixed in d31e5c14. The worker module URL is now resolved relative to `import.meta.url`, so locating `usageStats.ts` no longer depends on Vitest being launched from the repository root. The six-process serialization test passes with exact totals.

src/lib/proxy/usageStats.ts
murdore: Validated as not a bug, consistent with the correction in the final review summary. `stats` is replaced, not additively updated, with `persisted snapshot + currently pending local delta`; a successful flush clears that pending delta before reconcile can run because both operations share `flushMutex`. I added repeated-reconcile coverage that verifies totals remain unchanged and the status read does not write or change the snapshot revision.

src/lib/proxy/usageStats.ts
murdore: Validated as not a bug, matching the review summary correction. After a failed flush, the delta exists once in `pending`; reconcile replaces `stats` with `snapshot + pending`, rather than adding onto the old `stats`. I added a regression that forces lock timeout, performs repeated reconciles, retries the flush, and verifies the total remains exactly 2 throughout.

src/cli/commands/proxy.ts
murdore: Confirmed after the latest rebase and review fixes. The singleton guard still precedes the persistence initializer, and the source-order regression test passes on d31e5c14.

src/lib/proxy/proxyPaths.ts
murdore: Confirmed. Dev and production stats paths remain isolated, and the added backward-compatible fallback derives from the already-isolated `stateDir`. The dev-path tests pass on d31e5c14.

src/lib/types/proxy.ts
murdore: Confirmed. The persistence types remain in the shared proxy types module and continue to satisfy the repository type rules; strict typecheck and the production package build pass on d31e5c14.

package.json
murdore: Confirmed. The aggregate performance pipeline still includes the statistics benchmark. On d31e5c14 all lifecycle, stats, transport, and rolling-handoff budgets pass with exact counts and zero dropped handoff requests.

test/proxyReliabilityHardening.test.ts
murdore: Confirmed after the compatibility adjustment. The assertion now checks for the fallback-based initialization call, still asserts both markers exist, and still proves the singleton guard comes first.

murdore:
Addressed the complete latest review set in `d31e5c14`:

- **Yama MAJOR:** restored public `ProxyPaths` compatibility with an optional `statsFile` plus deterministic fallback; made usage-stat and lifecycle flushes concurrent under one bounded shutdown window.
- **Yama MINOR:** documented process-local reset semantics and added an isolated async test reset; added a short legacy-credential status cache; removed cross-process test cwd dependence; fixed singleton persistence isolation.
- **CodeRabbit:** retention now verifies the identities of the three newest corrupt snapshots; repeated lock/store setup now uses shared helpers.
- **Withdrawn Yama findings:** no accounting logic change was made for the two reconcile/double-count claims, as the final review correctly withdrew them. Added repeated-reconcile and failed-flush regression tests proving exact totals.
- **Positive threads:** revalidated lifecycle ordering, dev/prod isolation, type placement, benchmark integration, and source-order coverage after the rebase.

The CodeRabbit docstring percentage is a non-blocking service heuristic, and its Checkov/ESLint installation warnings are runner-configuration failures rather than repository code failures. Local repository formatting, ESLint (0 errors), security validation, and publint all pass.

Verification on current `origin/release` (`v10.1.2`): 350/350 Vitest, 232/232 bugfix suite, 118/118 env guard, strict typecheck, full package build, full pre-commit validation/security/build, and every proxy performance budget passed with zero handoff drops.

Tara-ag (approved):
Verified the current revision (`d31e5c14`) addresses all previously raised blocking concerns:

- `ProxyPaths.statsFile` is optional with a deterministic `resolveProxyUsageStatsPath` fallback, so external object literals remain valid.
- Shutdown flushes for usage stats and lifecycle metadata now run concurrently via `Promise.allSettled`, bounded to a single timeout window.
- Legacy credential lookup is cached (including negative results), removing per-`/status` filesystem reads.
- Tests reset singleton state with `resetUsageStatsForTests()`, which clears the configured file path and guarantees isolation.
- The cross-process worker test resolves the module URL from `import.meta.url`.
- Corrupt-snapshot pruning is asserted by retained content, not just count.

No new blocking issues (security, public SDK API breakage, or CLAUDE.md architectural violations) were found in this pass. The persistence store correctly preserves the `stats = persisted_snapshot + pending_delta` invariant across flushes, failed flushes, and reconciles, and lock recovery safely handles stale, malformed, and PID-reused locks.

Approving.

PR #1220, opened 2026-07-20
fix(litellm): discover real model context windows from the proxy /model/info

pdogra1299:
## Problem

`MODEL_CONTEXT_WINDOWS.litellm` is a single static `_default: 128_000`, but a LiteLLM proxy serves arbitrary models with real windows from 8K to 2M. Budget checks (`checkContextBudget`), compaction thresholds, and the new per-step loop guard all key off `getAvailableInputTokens()` — with a wrong window they either over-compact small models or let big conversations sail into provider 400s (`ContextWindowExceededError`, observed on juspay/yama#61 CI against `glm-private`, real window 262,144).

## Fix

Runtime window discovery, mirroring the existing async-populate/sync-read precedent (`DynamicModelProvider` Step 0 in `getContextWindowSize`):

- `contextWindows.ts`: a runtime override table + `registerRuntimeContextWindow(provider, model, window)`, consulted as Step 0.5 of the resolution chain (after the dynamic registry, before static lookup). Keyed on the raw provider string, matching how budget callers pass it. Invalid (non-positive/non-finite) registrations are ignored.
- `litellm.ts`: fire-and-forget `GET /model/info` on construction (same fetch/auth/5s-abort scaffolding and 10-minute cache period as the existing `/v1/models` fetch, deduped per base URL), registering `model_info.max_input_tokens` per model group. When a group has multiple deployments, the smallest advertised window wins so budgets are safe for every replica. Any failure (endpoint absent, auth, timeout) degrades cleanly to the static `_default`.

## Tests

`test/litellmContextWindows.test.ts` (10 tests, wired into `test:unit`): static fallback, runtime-override precedence, budgeting integration, invalid-value rejection, refresh semantics, `/model/info` parsing, min-window-per-group, malformed-entry tolerance, failure degradation, per-URL fetch dedupe.

## Summary by CodeRabbit

* **New Features**
  * Added runtime-discovered, per-model context window limits for LiteLLM using `/model/info`.
  * Applies discovered limits with preference for exact provider/model matches.
  * When multiple deployments are available, uses the smallest advertised limit.
* **Bug Fixes**
  * Ignores invalid/non-finite context window values.
  * If discovery fails, automatically falls back to existing default context behavior and can retry later.
* **Tests**
  * Added a new Vitest suite covering discovery, registry behavior, malformed responses, and request deduplication.
  * Updated unit test chaining to include this new suite.

### DIFF
--- a/src/lib/constants/contextWindows.ts
+++ b/src/lib/constants/contextWindows.ts
@@ -401,11 +401,49 @@ function normalizeProviderForLookup(provider: string): string {
 }
 
 /**
+ * Runtime-discovered context windows, keyed `${provider}:${model}`.
+ *
+ * Populated asynchronously by providers that can discover real per-model
+ * limits at runtime (e.g. the LiteLLM provider reads `max_input_tokens` from
+ * the proxy's `/model/info`), and read synchronously by
+ * {@link getContextWindowSize} — the same async-populate/sync-read contract as
+ * the DynamicModelProvider registry. Keys use the RAW provider string callers
+ * pass into budget calculations (see the alias-map comment above: normalized
+ * provider names never reach these lookups).
+ */
+const RUNTIME_CONTEXT_WINDOWS = new Map<string, number>();
+
+/**
+ * Register a runtime-discovered context window for a provider/model pair.
+ * Later registrations overwrite earlier ones (rediscovery refreshes values).
+ * Non-positive/non-finite windows are ignored so a malformed discovery source
+ * can never shrink a budget to zero.
+ */
+export function registerRuntimeContextWindow(
+  provider: string,
+  model: string,
+  contextWindow: number,
+): void {
+  if (!Number.isFinite(contextWindow) || contextWindow <= 0) {
+    return;
+  }
+  RUNTIME_CONTEXT_WINDOWS.set(`${provider}:${model}`, contextWindow);
+}
+
+/** Test hook: clear runtime-discovered windows (state is module-global). */
+export function clearRuntimeContextWindows(): void {
+  RUNTIME_CONTEXT_WINDOWS.clear();
+}
+
+/**
  * Resolve context window size for a provider/model combination.
  *
  * Priority:
  *  0. Dynamic model registry (DynamicModelProvider) — resolves cross-provider
  *     models (e.g. Claude on Vertex) that the static table cannot handle
+ *  0.5 Runtime-discovered windows (registerRuntimeContextWindow) — real
+ *      per-model limits fetched from the serving infrastructure (LiteLLM
+ *      `/model/info`)
  *  1. Exact model match under provider in static registry
  *  2. Prefix match under provider in static registry
  *  3. Provider's _default in static registry
@@ -428,6 +466,14 @@ export function getContextWindowSize(provider: string, model?: string): number {
       }
     } catch {
       // Dynamic registry not initialized yet — fall through to static lookup
+    }
+  }
+
+  // Step 0.5: Runtime-discovered window for this exact provider/model.
+  if (model) {
+    const discovered = RUNTIME_CONTEXT_WINDOWS.get(`${provider}:${model}`);
+    if (discovered !== undefined) {
+      return discovered;
     }
   }
 
--- a/src/lib/providers/litellm.ts
+++ b/src/lib/providers/litellm.ts
@@ -1,5 +1,6 @@
 import { SpanKind, SpanStatusCode, trace } from "@opentelemetry/api";
 import type { AIProviderName } from "../constants/enums.js";
+import { registerRuntimeContextWindow } from "../constants/contextWindows.js";
 import { createProxyFetch } from "../proxy/proxyFetch.js";
 import type {
   OpenAICompatBuildBodyArgs,
@@ -67,6 +68,12 @@ export class LiteLLMProvider extends OpenAIChatCompletionsProvider {
   private static modelsCache: string[] = [];
   private static modelsCacheTime = 0;
   private static readonly MODELS_CACHE_DURATION = 10 * 60 * 1000; // 10 minutes
+  /**
+   * Dedupes the fire-and-forget `/model/info` discovery per base URL, so a
+   * process constructing many LiteLLM providers fetches each proxy's limits
+   * once per cache period.
+   */
+  private static modelInfoFetchTime = new Map<string, number>();
 
   constructor(
     modelName?: string,
@@ -79,6 +86,13 @@ export class LiteLLMProvider extends OpenAIChatCompletionsProvider {
       baseURL: credentials?.baseURL ?? envConfig.baseURL,
       apiKey: credentials?.apiKey ?? envConfig.apiKey,
     });
+
+    // Fire-and-forget: discover real per-model context windows from the
+    // proxy's /model/info. The static table only has a one-size litellm
+    // `_default` (128K), while proxied models range from 8K to 2M — budget
+    // checks and compaction need the real window. Failures degrade cleanly
+    // to the static default.
+    this.discoverModelContextWindows();
 
     logger.debug("LiteLLM Provider initialized", {
       modelName: this.modelName,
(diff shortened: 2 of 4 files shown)


test/litellmContextWindows.test.ts
```diff
@@ -0,0 +1,181 @@
...
+
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import {
+  getContextWindowSize,
+  getAvailableInputTokens,
+  registerRuntimeContextWindow,
+  clearRuntimeContextWindows,
+} from "../src/lib/constants/contextWindows.js";
+import { LiteLLMProvider } from "../src/lib/providers/litellm.js";
+
```
Tara-ag, line 21: 💡 **MINOR:** `flushAsync` relies on `setTimeout(0)` to drain the fire-and-forget promise chain. This is usually sufficient, but it is sensitive to microtask/macrotask ordering. Prefer `await vi.waitFor(() => …)` or `await new Promise((r) => process.nextTick(r))` chained until the registry is populated, so the test does not flake if the discovery promise schedules additional microtasks.

src/lib/providers/litellm.ts
```diff
@@ -299,6 +313,105 @@ export class LiteLLMProvider extends OpenAIChatCompletionsProvider {
...
+    LiteLLMProvider.modelInfoFetchTime.set(baseURL, now);
+
+    void this.fetchModelInfoFromAPI()
+      .then((windows) => {
+        for (const [model, contextWindow] of windows) {
+          registerRuntimeContextWindow("litellm", model, contextWindow);
+        }
+        if (windows.size > 0) {
+          logger.debug(
+            "[LiteLLMProvider] Registered runtime context windows from /model/info",
```
Tara-ag, line 339: 💡 **MINOR:** Deleting the cache timestamp on any failure allows an immediate retry on the next provider construction. If `/model/info` is down and many `LiteLLMProvider` instances are created in quick succession, each will trigger a fresh fetch. Consider a short negative-cache (e.g., 30–60 s) so failures degrade quietly without a retry burst.

src/lib/constants/contextWindows.ts
```diff
@@ -400,12 +400,50 @@ function normalizeProviderForLookup(provider: string): string {
...
+  }
+  RUNTIME_CONTEXT_WINDOWS.set(`${provider}:${model}`, contextWindow);
+}
+
+/** Test hook: clear runtime-discovered windows (state is module-global). */
+export function clearRuntimeContextWindows(): void {
+  RUNTIME_CONTEXT_WINDOWS.clear();
+}
+
 /**
```
Tara-ag, line 438: 💡 **MINOR:** `clearRuntimeContextWindows` is exported solely as a test hook, but it becomes part of the public module surface. Consider keeping it internal (e.g., not exporting it) and resetting state via `vi.resetModules()` / re-importing in tests, or at least document that it is not intended for production use.

src/lib/providers/litellm.ts
```diff
@@ -299,6 +313,105 @@ export class LiteLLMProvider extends OpenAIChatCompletionsProvider {
...
+  /**
+   * Discover real per-model context windows from the LiteLLM proxy's
+   * `GET /model/info` (`data[].model_info.max_input_tokens`) and register
+   * them with the context-window resolver. Fire-and-forget with the same
+   * cache period as the models list; any failure (endpoint absent, auth,
+   * timeout) is logged at debug and leaves the static `_default` in force.
+   */
+  private discoverModelContextWindows(): void {
+    const baseURL = stripTrailingSlash(this.config.baseURL);
+    const now = Date.now();
```
Tara-ag, line 325: 💡 **MINOR:** The cache check + timestamp update is not atomic, so two `LiteLLMProvider` instances constructed concurrently for the same `baseURL` can both pass the freshness check and issue duplicate `/model/info` requests. Consider guarding the fetch with an in-flight promise map (keyed by `baseURL`) so concurrent constructions share one request.

test/litellmContextWindows.test.ts
```diff
@@ -0,0 +1,181 @@
...
+      })),
+    }),
+    { status: 200, headers: { "content-type": "application/json" } },
+  );
+
+describe("runtime context-window registry", () => {
+  beforeEach(() => clearRuntimeContextWindows());
+  afterEach(() => clearRuntimeContextWindows());
+
+  it("falls back to the static litellm _default when nothing is registered", () => {
```
Tara-ag, line 42: 💡 **SUGGESTION:** Consider adding a test that verifies the dynamic registry (Step 0) still takes precedence over the runtime override (Step 0.5), since the priority comment in `contextWindows.ts` claims dynamic registry wins.

test/litellmContextWindows.test.ts
```diff
@@ -0,0 +1,181 @@
...
+  it("keeps the SMALLEST advertised window when a model group has several deployments", async () => {
+    globalThis.fetch = (async () =>
+      modelInfoResponse([
+        { model_name: "glm-private", max_input_tokens: 262_144 },
+        { model_name: "glm-private", max_input_tokens: 131_072 },
+      ])) as typeof fetch;
+
+    new LiteLLMProvider("glm-private", undefined, undefined, {
+      apiKey: "k",
+      baseURL: uniqueBaseURL(),
```
Tara-ag, line 120: 💡 **SUGGESTION:** The `dedupes discovery per base URL` test only asserts one fetch was issued, but does not verify that the second provider eventually reads the discovered window for its own model (`private-large`). Consider adding that assertion to ensure the shared cache is actually consumed by later constructors.

Tara-ag (approved):
## Review Summary

Reviewed 4 changed files in PR #1220.

### New issues raised
- **MINOR:** 3
- **SUGGESTION:** 2

### Blocking criteria
No blocking issues found. No hardcoded secrets, security vulnerabilities, breaking public SDK API changes, or violations of non-negotiable CLAUDE.md Critical Rules that ESLint does not already enforce.

### Notes
- `src/lib/constants/contextWindows.ts`: Runtime override registry is a clean extension of the existing resolution chain. Minor concern about exporting a test-only reset hook.
- `src/lib/providers/litellm.ts`: Fire-and-forget `/model/info` discovery mirrors the existing models-list fetch pattern, degrades cleanly, and correctly picks the smallest window per model group. Minor concerns around non-atomic cache check and immediate retry on failure.
- `test/litellmContextWindows.test.ts`: Good coverage of fallback, precedence, invalid values, min-window-per-group, malformed responses, failure degradation, and per-URL dedupe. Minor suggestions to harden async flushing and add a couple of additional assertions.

Approving — the implementation is sound and the minor points can be addressed in follow-up or at the author's discretion.

src/lib/providers/litellm.ts
```diff
@@ -67,6 +68,12 @@ export class LiteLLMProvider extends OpenAIChatCompletionsProvider {
   private static modelsCache: string[] = [];
   private static modelsCacheTime = 0;
   private static readonly MODELS_CACHE_DURATION = 10 * 60 * 1000; // 10 minutes
+  /**
+   * Dedupes the fire-and-forget `/model/info` discovery per base URL, so a
+   * process constructing many LiteLLM providers fetches each proxy's limits
```
Tara-ag, line 73: 💡 **MINOR:** `modelInfoFetchTime` accumulates one entry per unique `baseURL` forever. In a long-lived process that constructs providers for many transient proxies, this static Map grows without bound. Consider capping its size (e.g. LRU) or pruning entries older than the cache duration.

src/lib/constants/contextWindows.ts
```diff
@@ -400,12 +400,50 @@ function normalizeProviderForLookup(provider: string): string {
...
+ * pass into budget calculations (see the alias-map comment above: normalized
+ * provider names never reach these lookups).
+ */
+const RUNTIME_CONTEXT_WINDOWS = new Map<string, number>();
+
+/**
+ * Register a runtime-discovered context window for a provider/model pair.
+ * Later registrations overwrite earlier ones (rediscovery refreshes values).
+ * Non-positive/non-finite windows are ignored so a malformed discovery source
+ * can never shrink a budget to zero.
```
Tara-ag, line 420: 💡 **MINOR:** The runtime registry only ever grows. If a LiteLLM proxy later removes a model group, the previously discovered window stays in `RUNTIME_CONTEXT_WINDOWS` indefinitely and continues to override the static default. Consider either (a) clearing a provider’s entries before each rediscovery so removed models fall back to static, or (b) attaching a generation/timestamp and evicting entries that weren’t refreshed in the latest `/model/info` response.

test/litellmContextWindows.test.ts
```diff
@@ -0,0 +1,181 @@
...
+    expect(getContextWindowSize("litellm", "glm-private")).toBe(262_144);
+  });
+});
+
+describe("LiteLLMProvider /model/info discovery", () => {
+  const originalFetch = globalThis.fetch;
+  let baseUrlCounter = 0;
+
+  // Each test uses a unique baseURL so the provider's per-URL fetch dedupe
+  // never suppresses a later test's discovery.
```
Tara-ag, line 78: 💡 **MINOR:** Putting `expect(url.endsWith("/model/info")).toBe(true)` inside the mock fetch means a failure throws into the provider’s fire-and-forget promise and is swallowed by its `.catch`. The test then fails later with a missing-registration assertion, making the root cause harder to diagnose. Move the URL check outside the mock, e.g.:

```ts
expect(fetchSpy).toHaveBeenCalledTimes(1);
expect(fetchSpy).toHaveBeenLastCalledWith(
  expect.stringContaining("/model/info"),
  expect.anything(),
);
```

Tara-ag (approved):
## Review Summary

Reviewed 4 changed files. No new blocking issues found.

### New comments raised (3 MINOR)

- `src/lib/constants/contextWindows.ts`: Runtime registry never evicts stale entries; removed LiteLLM model groups would keep overriding the static default indefinitely.
- `src/lib/providers/litellm.ts`: Static `modelInfoFetchTime` Map grows unbounded for processes that construct providers against many transient base URLs.
- `test/litellmContextWindows.test.ts`: Assertion placed inside the mock fetch can obscure the real failure if the URL is wrong.

### Notes

- Existing review comments already cover the `redactUrlForError` vs `redactUrlCredentials` choice, negative-cache / retry-burst behavior, concurrent-construction dedupe, `clearRuntimeContextWindows` export scope, and two test-coverage suggestions. I did not duplicate those.
- The static import from `../constants/contextWindows.js` into `litellm.ts` does not create a circular dependency (`contextWindows.ts` → `dynamicModels.ts` does not transitively import providers).
- The change is additive and backward compatible for public SDK callers.

Approving — the remaining items are minor quality points that can be addressed in follow-up or accepted as-is.

PR #1221, opened 2026-07-20
feat(context): per-step budget guard compacts the AI-SDK tool loop before overflow

pdogra1299:
## Problem

Pre-call budgeting (`checkContextBudget` + compaction in `neurolink.ts`) runs ONCE before dispatch and only sees the input/session conversation. The AI-SDK agent loop (`GenerationHandler.callGenerateText`, up to `DEFAULT_MAX_STEPS=200` steps) then appends assistant turns and tool results on every step — growth the pre-call pipeline never sees. Long agentic runs overflow the model's real window mid-loop: juspay/yama#61 CI accumulated ~246K input tokens against a 262K window (and double NeuroLink's own 128K litellm default) before the backend 400'd with `ContextWindowExceededError`.

The googleVertex native loops already guard this (`createContextGuard`, checked every iteration, `DEFAULT_CONTEXT_GUARD_RATIO` 0.85) — but as stop-only. The AI-SDK path, anthropic/bedrock native loops, and the chat-completions stream loop have nothing.

## Fix

`context/stepBudgetGuard.ts` + wiring in `callGenerateText` via `experimental_prepareStep` (whose result may replace the step's `messages` — the sanctioned AI-SDK hook). Every step:

1. Estimate the projected request (step messages + hoisted system prompt + tool definitions) against `getAvailableInputTokens() × DEFAULT_CONTEXT_GUARD_RATIO` — same ratio as the Vertex guard.
2. Over threshold → deterministically reclaim budget and **continue** (an upgrade over Vertex's stop-only): Stage 1 truncates OLD tool outputs to head/tail previews (`generateToolOutputPreview`); Stage 2 drops the oldest complete tool exchanges (assistant tool-call + its tool-result messages as a unit, preserving call/result pairing) and inserts one elision note.
3. Never touches: the first user (task) message, the protected recent tail, or anything when under threshold (zero overhead no-op).

Operates on `ModelMessage[]` natively (no lossy ChatMessage round-trip), makes no LLM calls, and composes with a caller-supplied `prepareStep` (caller result wins; the guard only contributes `messages`).

## Tests

`test/stepBudgetGuard.test.ts` (7 tests, wired into `test:unit`): estimation scaling, overhead accounting, no-op under threshold, stage-1 truncation with protected tail, stage-2 pair-integral drops with elision note, fixed-overhead trigger, protected-tail immunity. `test:anthropic-tools-policy` and `test:mcp:limits` pass unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added a per-step context budget guard for AI tool-driven message loops, including deterministic token estimation and safe message compaction.
  * Generation now automatically applies the guard to prevent context overflow while preserving the most recent conversation tail.
* **Bug Fixes**
  * Helps prevent long-running tool workflows from exceeding model context limits by reducing step history before sending.
* **Tests**
  * Added a dedicated Vitest suite covering token estimation, tool-output truncation, history preservation, and overhead/budget decisions.
* **Chores**
  * Expanded unit test scripts to include the new guard test suite.

### DIFF
--- a/src/lib/context/stepBudgetGuard.ts
+++ b/src/lib/context/stepBudgetGuard.ts
@@ -1,0 +1,366 @@
+/**
+ * Per-step context budget guard for the AI-SDK agent loop.
+ *
+ * Pre-call budgeting (`checkContextBudget` + compaction) runs ONCE before
+ * dispatch and only sees the input/session conversation. The AI-SDK tool loop
+ * then appends assistant turns and tool results on every step — growth the
+ * pre-call pipeline never sees, which is how long agentic runs overflow the
+ * model's real context window mid-loop (provider 400s after dozens of tool
+ * calls). The googleVertex native loops already guard this via
+ * `createContextGuard`; this module brings the AI-SDK path (every provider
+ * that delegates to `generateText`) to parity — and goes one step further:
+ * instead of stopping the loop, it deterministically reclaims budget so the
+ * loop can CONTINUE.
+ *
+ * Wired in `GenerationHandler.callGenerateText` through
+ * `experimental_prepareStep`, whose result may replace the step's `messages`.
+ * The guard operates on `ModelMessage[]` natively (no lossy ChatMessage
+ * round-trip) and never makes LLM calls:
+ *
+ *   Stage 1 — truncate OLD tool outputs to head/tail previews
+ *             (`generateToolOutputPreview`), oldest first, outside the
+ *             protected recent tail.
+ *   Stage 2 — drop the oldest complete tool exchanges (assistant tool-call
+ *             message + its following tool-result messages, as a unit, so
+ *             call/result pairing stays intact), replacing them with a single
+ *             elision note.
+ *
+ * The system prompt and tool definitions ride OUTSIDE the step messages (the
+ * handler hoists system into generateText's `system` option), so their cost is
+ * passed in as `fixedOverheadTokens`. The first user message (the task) and
+ * the most recent messages are never touched.
+ */
+
+import type { ModelMessage, StepBudgetGuardConfig } from "../types/index.js";
+import { DEFAULT_CONTEXT_GUARD_RATIO } from "../core/constants.js";
+import { getAvailableInputTokens } from "../constants/contextWindows.js";
+import {
+  estimateTokens,
+  TOKENS_PER_MESSAGE,
+} from "../utils/tokenEstimation.js";
+import { generateToolOutputPreview } from "./toolOutputLimits.js";
+import { logger } from "../utils/logger.js";
+
+/** Estimated tokens for a tool definition that fails to serialize. */
+const TOKENS_PER_TOOL_DEFINITION = 200;
+
+/** Messages at the end of the conversation the guard never modifies. */
+const PROTECTED_TAIL_MESSAGES = 4;
+
+/** Stage-1 preview budget for an old tool output (bytes). */
+const OLD_TOOL_OUTPUT_PREVIEW_BYTES = 2_048;
+
+/** Stage-1 preview budget for an old tool output (lines). */
+const OLD_TOOL_OUTPUT_PREVIEW_LINES = 60;
+
+/**
+ * Serialize any ModelMessage content to text for estimation. Tool-call args
+ * and tool-result outputs are JSON-stringified; unserializable values fall
+ * back to a fixed-size placeholder so estimation never throws.
+ */
+function contentToText(content: unknown): string {
+  if (typeof content === "string") {
+    return content;
+  }
+  try {
+    return JSON.stringify(content) ?? "";
+  } catch {
+    return "x".repeat(TOKENS_PER_TOOL_DEFINITION * 4);
+  }
+}
+
+/** Estimate the token cost of a step's message array. */
+export function estimateStepMessagesTokens(
+  messages: readonly ModelMessage[],
+  provider?: string,
+): number {
+  let total = 0;
+  for (const message of messages) {
+    total +=
+      estimateTokens(contentToText(message.content), provider) +
+      TOKENS_PER_MESSAGE;
+  }
+  return total;
+}
+
+/**
+ * Estimate the fixed per-request overhead: hoisted system prompt + tool
+ * definitions. Mirrors `checkContextBudget`'s categories for the pieces that
+ * do not live in the step messages.
+ */
+export function estimateFixedOverheadTokens(
+  system: unknown,
+  tools: Record<string, unknown> | undefined,
+  provider?: string,
+): number {
+  let total = system
+    ? estimateTokens(contentToText(system), provider) + TOKENS_PER_MESSAGE
+    : 0;
+  for (const tool of Object.values(tools ?? {})) {
+    try {
+      total += estimateTokens(JSON.stringify(tool) ?? "", provider);
+    } catch {
+      total += TOKENS_PER_TOOL_DEFINITION;
+    }
+  }
+  return total;
+}
+
+/**
+ * Serialize a ToolResultOutput to the text the MODEL should see in a preview.
+ * Variant-aware: `text`/`error-text` carry their payload in `.value` directly —
+ * stringifying the wrapper would put escaped `{"type":"text","value":…}` JSON
+ * in front of the model instead of the actual output. `json`/`error-json`/
+ * `content` serialize their value; unknown shapes fall back to the wrapper.
+ */
+function toolResultOutputToText(output: unknown): string {
+  const variant = output as { type?: string; value?: unknown } | undefined;
+  if (variant && typeof variant === "object" && "type" in variant) {
+    if (
+      (variant.type === "text" || variant.type === "error-text") &&
+      typeof variant.value === "string"
+    ) {
+      return variant.value;
+    }
+    if (
+      variant.type === "json" ||
+      variant.type === "error-json" ||
+      variant.type === "content"
+    ) {
+      return contentToText(variant.value);
+    }
+  }
+  return contentToText(output);
+}
+
+/** True when the message is an assistant message that issues tool calls. */
+function isToolCallAssistantMessage(message: ModelMessage): boolean {
+  return (
+    message.role === "assistant" &&
+    Array.isArray(message.content) &&
+    message.content.some(
+      (part) => (part as { type?: string })?.type === "tool-call",
+    )
+  );
+}
+
+/**
+ * Stage 1: replace large tool-result outputs outside the protected tail with
+ * head/tail previews. Returns the new array plus how many outputs shrank.
+ */
+function truncateOldToolOutputs(messages: ModelMessage[]): {
+  messages: ModelMessage[];
+  truncated: number;
+} {
+  const cutoff = Math.max(0, messages.length - PROTECTED_TAIL_MESSAGES);
+  let truncated = 0;
+
+  const next = messages.map((message, index) => {
+    if (index >= cutoff || message.role !== "tool") {
+      return message;
+    }
+    if (!Array.isArray(message.content)) {
+      return message;
+    }
+    let changed = false;
+    const content = message.content.map((part) => {
+      const resultPart = part as {
+        type?: string;
+        toolCallId?: string;
+        toolName?: string;
+        output?: unknown;
+      };
+      if (resultPart?.type !== "tool-result") {
+        return part;
+      }
+      const serialized = toolResultOutputToText(resultPart.output);
+      if (serialized.length <= OLD_TOOL_OUTPUT_PREVIEW_BYTES) {
+        return part;
+      }
+      const { preview } = generateToolOutputPreview(serialized, {
+        maxBytes: OLD_TOOL_OUTPUT_PREVIEW_BYTES,
+        maxLines: OLD_TOOL_OUTPUT_PREVIEW_LINES,
+      });
+      changed = true;
+      truncated += 1;
+      return {
+        ...resultPart,
+        output: { type: "text", value: preview },
+      };
+    });
+    return changed ? ({ ...message, content } as ModelMessage) : message;
+  });
+
+  return { messages: next, truncated };
+}
+
+/**
+ * Stage 2: drop the oldest complete tool exchanges — an assistant tool-call
+ * message together with ALL directly-following `tool` messages — until the
+ * estimate fits or only the protected head/tail remains. The first
+ * non-assistant message run (the task) is never dropped. A single elision
+ * note replaces everything removed so the model knows history was elided.
+ */
+function dropOldestToolExchanges(
+  messages: ModelMessage[],
+  budgetTokens: number,
+  fixedOverheadTokens: number,
+  provider?: string,
+): { messages: ModelMessage[]; droppedExchanges: number } {
+  const result = [...messages];
+  let droppedExchanges = 0;
+
+  // Running-total accounting: estimate each message ONCE, keep the estimates
+  // array in lockstep with `result`, and subtract dropped blocks — instead of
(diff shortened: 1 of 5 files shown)


src/lib/context/stepBudgetGuard.ts
```diff
@@ -0,0 +1,300 @@
...
+      exchangeEnd++;
+    }
+    if (exchangeEnd > tailStart) {
+      break; // exchange bleeds into the protected tail — stop
+    }
+    result.splice(exchangeStart, exchangeEnd - exchangeStart);
+    droppedExchanges++;
+  }
+
+  if (droppedExchanges > 0) {
```
Tara-ag, line 258: 💡 **Edge-case: elision note can land after the protected tail**

When `dropOldestToolExchanges` removes every droppable tool exchange, the `noteIndex` scan walks all the way to `result.length` and `splice(result.length, 0, …)` appends the truncation note *after* the protected tail. The model then sees the remaining context first and the “history was removed” note last, which weakens the cue.

Consider capping the insertion point so the note always sits immediately before the first retained exchange (or at least before the protected tail):

```ts
const firstExchangeIndex = result.findIndex(isToolCallAssistantMessage);
const noteIndex =
  firstExchangeIndex === -1
    ? Math.max(0, result.length - PROTECTED_TAIL_MESSAGES)
    : firstExchangeIndex;
result.splice(noteIndex, 0, { … });
```

This preserves the current placement in the common case and fixes the all-dropped edge case.

src/lib/context/stepBudgetGuard.ts
```diff
@@ -0,0 +1,300 @@
...
+    budgetTokens;
+
+  while (!fits()) {
+    // Find the FIRST droppable exchange outside the protected tail.
+    const tailStart = Math.max(0, result.length - PROTECTED_TAIL_MESSAGES);
+    let exchangeStart = -1;
+    for (let i = 0; i < tailStart; i++) {
+      if (isToolCallAssistantMessage(result[i])) {
+        exchangeStart = i;
+        break;
```
Tara-ag, line 231: 💡 **Stage 2 token estimation is O(n²) in the worst case**

Each `while (!fits())` iteration recomputes `estimateStepMessagesTokens(result)` from scratch over the whole array. If the guard has to drop many exchanges (e.g. dozens of tool calls deep into a long run), that’s ~m×n estimations and repeated `JSON.stringify` of the same messages.

Since `result.splice(exchangeStart, exchangeEnd - exchangeStart)` removes a contiguous block, you can keep a running total and subtract the dropped exchange’s estimated tokens instead:

```ts
let currentTokens = estimateStepMessagesTokens(result, provider);
const fits = () =>
  fixedOverheadTokens + currentTokens <= budgetTokens;

while (!fits()) {
  …
  const droppedTokens = estimateStepMessagesTokens(
    result.slice(exchangeStart, exchangeEnd),
    provider,
  );
  result.splice(exchangeStart, exchangeEnd - exchangeStart);
  currentTokens -= droppedTokens;
  droppedExchanges++;
}
```

This keeps stage 2 linear and avoids re-serializing the surviving messages on every iteration.

src/lib/core/modules/GenerationHandler.ts
```diff
@@ -196,13 +220,23 @@ export class GenerationHandler {
       stopWhen: stepCountIs(options.maxSteps ?? DEFAULT_MAX_STEPS),
```
Tara-ag, line 237: ⚠️ **MAJOR: `prepareStep` is now always wrapped, which changes the public callback signature**

Previously `experimental_prepareStep` was only set when the caller supplied `prepareStep`. Now it is always set, and the wrapper passes `stepOptions.messages` through `stepBudgetGuard` *before* invoking the caller’s `prepareStep`. The caller’s `prepareStep` therefore receives the already-compacted messages, not the original AI-SDK step messages.

This is a behavioral change to the public SDK API. Callers that inspect message length, look for specific historical turns, or implement their own context logic inside `prepareStep` will see different inputs than before. Per CLAUDE.md Critical Rule 5 (backward compatibility of the public SDK API), this needs either:

1. A documented migration note explaining that `prepareStep` now receives guard-compacted messages, or
2. Passing the original `stepOptions.messages` to the caller and applying compaction only to the final merged result.

Suggested minimal fix: keep the public `prepareStep` on the original messages, then merge the guard’s compaction afterward:

```ts
experimental_prepareStep: (async (stepOptions) => {
  const callerResult = prepareStep
    ? await prepareStep({
        ...stepOptions,
        maxSteps: options.maxSteps ?? DEFAULT_MAX_STEPS,
      })
    : undefined;
  const compacted = stepBudgetGuard(stepOptions.messages);
  if (compacted) {
    return { ...(callerResult ?? {}), messages: compacted };
  }
  return callerResult;
}) satisfies PrepareStepFunction,
```

This preserves the existing public contract while still enforcing the budget guard.

src/lib/context/stepBudgetGuard.ts
```diff
@@ -0,0 +1,300 @@
...
+ * message together with ALL directly-following `tool` messages — until the
+ * estimate fits or only the protected head/tail remains. The first
+ * non-assistant message run (the task) is never dropped. A single elision
+ * note replaces everything removed so the model knows history was elided.
+ */
+function dropOldestToolExchanges(
+  messages: ModelMessage[],
+  budgetTokens: number,
+  fixedOverheadTokens: number,
+  provider?: string,
```
Tara-ag, line 208: 💡 **Protected-tail check can stop early when an exchange starts inside but ends outside the tail**

The current condition `if (exchangeEnd > tailStart) { break; }` stops as soon as any droppable exchange bleeds into the protected tail. That’s safe, but it means a long tool-result chain that *starts* outside the tail but *extends* into it prevents dropping any earlier exchanges, even though earlier ones are fully outside the tail.

Consider whether the intended invariant is “never drop a message inside the tail” rather than “never drop an exchange that touches the tail”. If so, you could drop the prefix of the exchange that lies outside the tail and keep the suffix. However, that breaks call/result pairing, so the current conservative behavior is defensible — just flagging that this is a stop-only (not compact-to-fit) scenario in that corner case.

src/lib/core/modules/GenerationHandler.ts
Tara-ag, line 237: ⚠️ **MAJOR: Caller `prepareStep` can silently override the guard’s compacted messages**

The current merge does `{ ...(callerResult ?? {}), messages: compacted }`, which means the guard’s compaction always wins. But if the caller returns `messages` in their `PrepareStepResult`, that field is overwritten without warning. The public type contract says `prepareStep` returns `model | toolChoice | experimental_activeTools` only (see `GenerateOptions.prepareStep` in `src/lib/types/generate.ts`), so returning `messages` is technically out-of-contract — but it is possible at runtime.

More importantly, the comment in the code says “NeuroLink's public prepareStep contract returns model/toolChoice/activeTools overrides only — merge the guard's compacted messages in”, which is correct. However, because the wrapper now *always* runs, a caller that previously returned `messages` (even though undocumented) will see that return value ignored. This is another backward-compatibility concern tied to the previous comment.

If you decide to keep the current ordering, consider logging a one-time warning when `callerResult.messages` is present so callers know it is being discarded. Alternatively, adopt the fix in the previous comment (run caller on original messages, then merge guard compaction) so any `messages` the caller returns is also respected before final compaction.

Tara-ag, line 237: 💡 **Consider guarding `experimental_prepareStep` with `logger.shouldLog('debug')` for verbose diagnostics**

Not blocking, but the guard currently emits an `info` log only when compaction happens. For debugging why a long agentic run still overflows, it would help to log the per-step estimate and threshold at `debug` level. Wrap the debug payload in `logger.shouldLog('debug')` to avoid serializing large message arrays on every step when debug logging is off (per CLAUDE.md “Logger Guard” pattern).

test/stepBudgetGuard.test.ts
Tara-ag, line 1: 💡 **CONTRIBUTING.md test convention mismatch**

The project’s `CONTRIBUTING.md` states that test suites run via `tsx` and that `*.test.ts` files using `describe/it` blocks will not be picked up. This PR adds `test/stepBudgetGuard.test.ts` using Vitest’s `describe/it` and wires it into `test:unit` via a new `test:step-budget-guard:vitest` script.

While the repo already has several `*.test.ts` files run by Vitest (e.g. `test/systemMessages.test.ts`), the documented convention still discourages them. Consider adding a corresponding `test/continuous-test-suite-step-budget-guard.ts` orchestrator as the canonical integration test, and keep the Vitest file as a fast unit supplement. This aligns with CONTRIBUTING.md and gives CI a no-API fallback that exercises the guard through the public `generate()` path if feasible.

src/lib/types/context.ts
```diff
@@ -657,6 +657,22 @@ export type BudgetCheckResult = {
   };
 };
 
+/**
```
Tara-ag, line 660: 💡 **Type location is correct per CLAUDE.md**

Good: `StepBudgetGuardConfig` is defined in `src/lib/types/context.ts` and imported via the barrel (`../types/index.js`). This follows Critical Rules 2 and 13.

Tara-ag (changes requested):
## Review Summary

**Files reviewed:** 5 changed files (package.json, src/lib/context/stepBudgetGuard.ts, src/lib/core/modules/GenerationHandler.ts, src/lib/types/context.ts, test/stepBudgetGuard.test.ts)

**New issues raised:** 7
- 🔒 CRITICAL: 0
- ⚠️ MAJOR: 2 (both on `GenerationHandler.ts:220`; related but distinct)
- 💡 MINOR: 3
- 💬 SUGGESTION: 2

**Decision: REQUEST_CHANGES**

The implementation is solid conceptually and the tests cover the core compaction logic well, but the `experimental_prepareStep` wrapper in `GenerationHandler.ts` changes the public `prepareStep` contract in a backward-incompatible way:

1. **Backward-compatibility break (blocking):** The wrapper now passes guard-compacted `messages` into the caller’s `prepareStep` callback. Callers that inspect step messages (even if untyped) will see different inputs than before. Per CLAUDE.md Critical Rule 5, the public SDK API must not break existing callers without a documented migration path. The fix is straightforward: run the caller’s `prepareStep` on the original `stepOptions.messages`, then merge the guard’s compaction afterward.

2. **Silent override risk:** The merge `{ ...(callerResult ?? {}), messages: compacted }` discards any `messages` the caller returns without warning. Combined with the first issue, this makes the interaction between caller-supplied and guard-supplied message mutation unclear.

Other non-blocking notes:
- Stage-2 token estimation is O(n²); can be made linear with a running total.
- The elision note can be appended after the protected tail when all droppable exchanges are removed.
- The new Vitest test file conflicts with CONTRIBUTING.md’s stated convention that `*.test.ts` files are not picked up; consider adding a `continuous-test-suite-step-budget-guard.ts` orchestrator as the canonical test.

Please address the `prepareStep` wrapping order so callers continue to receive the original step messages, and re-request review.

src/lib/core/modules/GenerationHandler.ts
pdogra1299: Fixed in 5b8fd7d0 — the composition is restructured so the public contract is unchanged: a caller-supplied prepareStep now receives the ORIGINAL AI-SDK step options (exact pre-guard behavior), and the guard runs LAST on the messages that will actually be sent (the caller's returned messages when present, else the step's own).

pdogra1299: Fixed in 5b8fd7d0 together with the sibling MAJOR — precedence is now explicit: the guard never replaces a caller's content choices; it compacts whatever messages the caller chose (returned messages ?? step messages), acting purely as a budget safety net that runs after the caller.

src/lib/context/stepBudgetGuard.ts
pdogra1299: Fixed in 5b8fd7d0 — the insertion index is capped at the protected-tail boundary (min(noteIndex, len - PROTECTED_TAIL_MESSAGES)), so the note always precedes the retained context. Covered by the new 'places the elision note BEFORE the protected tail' test.

src/lib/context/stepBudgetGuard.ts
pdogra1299: Fixed in 5b8fd7d0 — stage 2 now estimates each message once, keeps the estimates array in lockstep with the result array, and subtracts dropped blocks from a running total: O(n) total instead of O(m×n) re-estimation.

src/lib/context/stepBudgetGuard.ts
pdogra1299: Kept as-is, with the invariant now documented at the break site (5b8fd7d0): the scan is oldest-first, so when the FIRST remaining exchange bleeds into the tail, every later exchange necessarily STARTS inside the tail (this one's result chain reaches it), and every earlier exchange was already dropped by prior loop iterations — there is provably nothing else the guard may remove. Splitting the bleeding exchange itself would break call/result pairing, which stage 2 deliberately never does.

src/lib/core/modules/GenerationHandler.ts
pdogra1299: Added in 5b8fd7d0 — per-step estimate/threshold/willCompact at debug level, wrapped in logger.shouldLog('debug') per the CLAUDE.md Logger Guard pattern so nothing serializes when debug is off.

test/stepBudgetGuard.test.ts
pdogra1299: Keeping the vitest suite, matching current repo practice: test:unit already chains several vitest suites the same way (test:unit:vitest, test:tool-dedup:vitest, test:model-pool:vitest, test:system-messages:vitest, and test:litellm-context:vitest from #1220). A parallel tsx orchestrator would duplicate coverage without adding signal; happy to add one if maintainers prefer the documented convention over the observed one.

src/lib/types/context.ts
pdogra1299: Thanks — kept in place through the rebase.

src/lib/context/stepBudgetGuard.ts
```diff
@@ -0,0 +1,334 @@
...
+      });
+      changed = true;
+      truncated += 1;
+      return {
+        ...resultPart,
+        output: { type: "text", value: preview },
+      };
+    });
+    return changed ? ({ ...message, content } as ModelMessage) : message;
+  });
```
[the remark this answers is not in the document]
pdogra1299: Valid — fixed in b638eb3d. New variant-aware toolResultOutputToText(): text/error-text use .value directly, json/error-json/content serialize their value, unknown shapes fall back to the wrapper. Both the size check and the preview input now use it, so a truncated text result previews the actual payload instead of escaped {"type":"text","value":…} JSON. Covered by the new 'previews the tool output VALUE for text results' test.

src/lib/core/modules/GenerationHandler.ts
```diff
@@ -187,6 +191,26 @@ export class GenerationHandler {
...
+      provider: this.providerName ?? "unknown",
+      model: this.modelName,
+      maxTokens: options.maxTokens,
+      fixedOverheadTokens: estimateFixedOverheadTokens(
+        system,
+        shouldUseTools ? toolsWithCache : undefined,
+        this.providerName,
+      ),
+    });
+
```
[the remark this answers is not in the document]
pdogra1299: Valid — fixed in b638eb3d. StepBudgetGuardConfig gains getFixedOverheadTokens (re-evaluated on every guard invocation, precedence over the static field), and GenerationHandler passes a thunk memoized on tool COUNT — tools are only ever added by search_tools hydration, so the recompute happens exactly when the set grows and the common step stays O(1). Covered by the new 're-resolves overhead per invocation' test.

PR #1222, opened 2026-07-21
fix(litellm): tool-calling and timeout parity for the chat-completions agent loop

pdogra1299:
## Description

**What does this PR do?**

Brings the AI-SDK agent-loop path (litellm and every OpenAI-compatible provider) to parity with the native Vertex/Bedrock loops on tool-argument handling and timeout behavior. RCA'd from the Yama PR-review pipeline running ~60 MCP tools against a LiteLLM proxy with GLM models: turns died at the wall-clock with **all completed steps discarded and no way to detect it** (`stopReason`/`stepsUsed` were never set on this path), and recoverable tool-argument mistakes (`"123"` for a number, missing `pullNumber`) each burned a full model round-trip.

## Type of Change

## Motivation and Context

Live-verified failure modes on NeuroLink 10.3.0:

- `litellm generate operation timed out after 15m` — the outer wall-clock guillotined an agentic run mid-tool-loop; every completed step was thrown away (`baseProvider.executeStandardGenerateFlow` throws, nothing is returned).
- `stopReason: undefined, stepsUsed: undefined` on every chat-completions `generate()` — consumers treating "missing = completed" cannot detect truncated runs. Native loops (googleVertex, googleNativeGemini3, bedrock, anthropic) set both.
- `Missing required parameter: pullNumber` — external MCP tools are registered with `jsonSchema()` and **no validate function**, which the AI SDK treats as "validate nothing" (`safeValidateTypes` short-circuits on `validate == null`); malformed calls sail into execution and fail late.
- The BZ-665 tool-call repair machinery (`BaseProvider.getToolCallRepairFn` → `createToolCallRepair`) existed but was never wired into `generateText` — dead code.
- `fetchWithRetry` classified retryability from the top-level error only; undici wraps every connection death in `TypeError: fetch failed` with the real code on `error.cause`, so no transient network error was ever retried.
- An explicit `timeout: "15m"` only stretched the outer clock — each step request stayed capped at the provider default (litellm: 5m) because the delegating `doGenerate` reads AI-SDK call options, which never carry `timeout`.
- `MaxListenersExceededWarning` at 10+ steps: per-step `AbortSignal.any` registrations accumulated on the long-lived generate-call signal.
- A configured MCP server `timeout` above 60s never took effect: `client.callTool` got no `RequestOptions`, so the MCP SDK's `DEFAULT_REQUEST_TIMEOUT_MSEC` (60s) always fired first.

## Changes Made

**Tool calling**

- Wire `experimental_repairToolCall` into `generateText` (opt-out: `disableToolCallRepair`) — fixes near-miss tool names (case/substring/Levenshtein) and repairs invalid inputs.
- Attach a real argument validator to external MCP tool schemas via `@cfworker/json-schema` (already an optional peerDependency of `@modelcontextprotocol/sdk`). Invalid calls now fail at parse time where repair can silently fix them; unrepairable calls become `tool-error` parts with an actionable message. Compilation failures degrade gracefully to the previous declarative-only behavior.
- MCP-layer `validateToolParameters` coerces recoverable mismatches (numeric strings, `"true"`/`"false"`, JSON-encoded objects/arrays — reusing `coerceType` from the repair module) and rejection messages restate the full contract: `Missing required parameter: pullNumber. Expected arguments: { pullNumber: number, owner?: string }; received keys: [owner]`.
- `client.callTool` receives `RequestOptions.timeout`, so configured server timeouts actually apply and SDK-side timeout cancels the transport request (was: soft `Promise.race` with a leaked timer and a ghost request).

**Timeouts / loop bounds**

- Turn budget + wrap-up parity: with `turnTimeoutMs` (or an explicit `timeout`) set, the loop forces `toolChoice: "none"` plus an honest time message inside the `wrapupTimeLeadMs` window (default 120s, clamped to budget/4) → partial answer with `stopReason: "time-limit"` instead of a `TimeoutError` that discards the turn. Callers that set neither keep the exact previous behavior.
- `stopReason` (`completed` / `step-cap` / `time-limit`) + `stepsUsed` now populated by `formatEnhancedResult` for the AI-SDK path.
- Caller's resolved `timeout` flows to the delegating chat-completions model via `providerOptions.neurolink.timeoutMs` (namespaced; other providers only read their own namespace).
- Response body read moved inside the per-step timeout window (was after `cleanup()`, leaving a stalled body unbounded).
- `fetchWithRetry` unwraps the `error.cause` chain; undici headers/body timeouts stay deliberately non-retryable (they already waited out undici's 300s deadline — replaying triples the stall).
- Abort hygiene: scoped per-step signal composition with `dispose()`, `mergeAbortSignals` no longer leaks a permanent listener per call, `withTimeout` clears its race timer.

## Breaking Changes

- [x] No breaking changes

All new options/fields are optional and additive. Two deliberate behavior improvements: (1) callers who set an explicit `timeout` on a tool loop now get an honest partial (`stopReason: "time-limit"`) instead of a thrown `TimeoutError` when the budget runs out mid-loop; (2) malformed MCP tool arguments are coerced/repaired where unambiguous instead of failing execution.

## Testing

- New no-API suite: `pnpm run test:litellm-parity` (10 tests: coercion, validator contract messages, RequestOptions passthrough, cause-chain retry classification, abort-listener hygiene, stopReason/stepsUsed mapping).
- Existing no-API suites pass: `test:mcp` (84), `test:mcp:limits` (24), `test:mcp:spans` (4).
- `pnpm run check`, `pnpm run lint` (0 errors, warning count unchanged from base), `pnpm run build` all clean.
- **Live verification** against a LiteLLM proxy (`glm-private`):
  - 3-step tool loop: `stopReason: "completed"`, `stepsUsed: 2`, correct tool calls and answer, 26s (parity with raw HTTP: 24s).
  - 5-tool task with `turnTimeoutMs: 12_000`: returns in ~17-26s with an honest partial ("I was only able to retrieve PR #101… could not verify 303/404/505 due to the time budget"), `stopReason: "time-limit"`, instead of dying with nothing.

**Deferred follow-ups** (out of scope here): step-scoped `withProviderRetry` (currently a retryable 429/5xx restarts the whole loop), `stallTimeoutMs` for the AI-SDK loop, undici Agent tuning (`headersTimeout`) for the LLM fetch path, and plumbing the abort signal into the MCP transport.

## Round 2 — Dynamic model limits + wire schema/name hygiene

Second RCA round (live `ContextWindowExceededError` from the Yama pipeline: `input 230145 + requested output 32000 > 262144`) added a fully **dynamic** model-limits layer and the remaining tool-argument parity fixes. Nothing hardcoded, no new flags — every limit comes from the proxy's `/model/info`, the backend's own error messages, or real usage reported by previous responses.

**Dynamic model limits**

- `/model/info` discovery is now an **awaitable** module-level service (deduped per base URL, cached, failures retried) registering both `max_input_tokens` → context window and `max_output_tokens` → output ceiling. Generation pipelines construct the provider **before** the budget check and await the new `BaseProvider.ensureModelLimits()` hook (default no-op). Previously discovery was a constructor fire-and-forget racing the first budget check — and since the budget hard-cap threw before the provider existed, a blocked call prevented the very discovery that would have unblocked it: fresh CI processes were permanently stuck on the static 128K default, falsely rejecting 130–230K-token prompts that fit the real window.
- `getSafeMaxTokens` consults the discovered output ceiling: over-ceiling requests clamp to the deployment's real cap (32000 → glm-private's advertised 31000 — that clamp alone makes the reported failure fit), and no-`maxTokens` callers default to it instead of the 128000 static blanket.
- Wire-level `max_tokens` fit on both chat-completions paths (per step on streaming): `min(requested, ceiling, window − estimated input − margin)`, enforced **only** for runtime-discovered windows — static guesses are never hard-enforced. Input alone exceeding a discovered window fails fast with a typed `ContextBudgetExceededError` instead of a guaranteed provider 400 + proxy-side fallback cascade.
- Overflow 400s **self-heal**: `parseProviderOverflowDetails` learns the vllm/LiteLLM phrasing, the stated window is registered at runtime (works even when `/model/info` is absent/unauthorized), and the request retried once with `max_tokens` re-fit from the provider's own numbers.
- `stepBudgetGuard` re-resolves its budget every invocation (mid-loop discovery takes effect; was frozen at loop start) and calibrates the char-based estimator against the provider's real prompt-token usage from the previous step (scale-up only, capped) — dense diff content no longer slips past the compaction threshold.

**Tool-argument wire hygiene** (the "model can't produce the right arguments" gap — native Gemini/Claude paths sanitize schemas before the wire; proxied backends render them into chat templates verbatim)

- `normalizeWireToolSchema` on every chat-completions tools block: inlines local `$ref`/`$defs`/`definitions` (the canonical `inlineJsonSchema` now understands 2020-12 `$defs`), strips `$schema`/`$id`/`$comment`, collapses zod's nullable `anyOf` into type arrays, guarantees a top-level object shape. Identity for clean schemas; `description`/`required`/`enum`/`default` always preserved.
- Wire tool-name sanitization with a bijective per-request map when a registered name falls outside `^[a-zA-Z_][a-zA-Z0-9_-]{0,63}$` (MCP names with dots/slashes/spaces): tools block, `tool_choice`, and history `tool_calls` go out sanitized; responses, execution lookup, events, and `toolsUsed` reverse-map to registered names. No-op when all names are valid.

**Round-2 verification**

- Live against the proxy: a **151K-token first call in a fresh process** — previously hard-blocked in 1ms against the static 128K budget — discovers the real 262144 window and completes (`usage.input: 151616`); second call reads the registry instantly. A wire-invalid tool name (`bitbucket.server/get pr-info`) round-trips through the loop with correct execution and reporting. Tool-loop regression check unchanged (`stopReason: "completed"`, `stepsUsed: 2`).
- Suites: `test:litellm-parity` grown to 17, `litellmContextWindows` + `stepBudgetGuard` vitest 26, `test:mcp` 84 / `mcp:limits` 24 / `mcp:spans` 4 — all green; `check`, `lint` (0 errors, warning count unchanged from base), `build` clean.

## Summary by CodeRabbit

* **New Features**
  * Added schema-based MCP tool input validation with recoverable coercion, strict integer handling, and contract-aware errors.
  * Improved per-turn budgeting with forced wrap-up, `time-limit` stop reasons, and more accurate `stepsUsed`.
  * Enhanced OpenAI-compatible tool wire mapping and tool-schema normalization, plus runtime model limit/context window/output-ceiling discovery and fitting.
  * Added `disableToolCallRepair` option to turn off schema-driven tool repair.

* **Bug Fixes**
  * Refined transient network retry detection, timeout handling, and abort-signal lifecycle to reduce leaks.
  * Expanded overflow parsing for safer self-healing retries with corrected `max_tokens`.
  * Calibrated step-budgeting using observed token usage.

* **Tests**
  * Added continuous LiteLLM parity coverage across tools, timeouts/aborts, budgeting, and stop semantics.


src/lib/core/modules/GenerationHandler.ts
```diff
@@ -228,7 +299,7 @@ export class GenerationHandler {
       },
```
Tara-ag, line 366: ⚠️ **MAJOR: `providerOptions` collision between timeout and thinkingConfig**

Both the timeout-forwarding block (line ~338) and the `thinkingConfig` block (line ~357) spread `providerOptions: { ... }` into the same `generateText` options object. If `thinkingConfig.enabled` is truthy on a Google provider, the second spread **overwrites** the `neurolink.timeoutMs` key entirely, so an explicit `timeout: "15m"` is silently dropped for Gemini thinking calls.

**Suggested fix:** merge the two `providerOptions` objects before spreading, e.g.

```ts
const providerOptions: Record<string, unknown> = {};
if (callerTimeoutMs !== undefined) {
  providerOptions.neurolink = { timeoutMs: callerTimeoutMs };
}
if (isGoogleProvider && options.thinkingConfig?.enabled) {
  providerOptions.google = { /* thinkingConfig */ };
}
// ...
...(Object.keys(providerOptions).length > 0 && { providerOptions }),
```

This preserves both namespaces and avoids losing the per-step timeout for thinking-enabled Gemini calls.

src/lib/core/modules/GenerationHandler.ts
```diff
@@ -258,15 +329,62 @@ export class GenerationHandler {
         const compacted = stepBudgetGuard(
```
Tara-ag, line 400: ⚠️ **MAJOR: `prepareStep` return contract changed for wrap-up; caller toolChoice is overridden without notice**

The wrap-up branch returns `buildWrapupStepResult(prepared, stepOptions.messages)`, which unconditionally sets `toolChoice: "none"` and appends a synthetic user message. This overrides any caller-supplied `toolChoice` from `prepareStep` or `options.toolChoice`. While the comment says "an honest partial beats a discarded turn", this is a behavior change for callers that intentionally use `toolChoice: "required"` or a per-step tool selection inside `prepareStep`.

More importantly, the synthetic user message is appended **after** the caller's `prepareStep` ran, so the caller's callback sees a different message set than what the model receives. This breaks the public `prepareStep` contract documented in `GenerateOptions`.

**Suggested fix:** document this override in the public option JSDoc, and consider applying the wrap-up nudge inside the guard **before** invoking `prepareStep` so the caller's callback still observes the final message set. At minimum, mention in `GenerateOptions.wrapupTimeLeadMs` that wrap-up forcibly sets `toolChoice: "none"`.

src/lib/mcp/toolDiscoveryService.ts
```diff
@@ -620,11 +625,21 @@ export class ToolDiscoveryService extends EventEmitter {
           async (callSpan) => {
```
Tara-ag, line 625: ⚠️ **MAJOR: `client.callTool` signature mismatch — third argument may not be `RequestOptions` in this SDK version**

The PR description claims this passes `{ timeout: effectiveTimeout }` as MCP `RequestOptions`, but the `@modelcontextprotocol/sdk` version pinned in this repo (1.27.1/1.28.0/1.29.0 per lockfile) defines `callTool(params, resultSchema?, options?)` where `options` is indeed `RequestOptions`. However, the second positional argument here is `undefined`. Passing `undefined` as `resultSchema` is fine, but the PR also bumped `@modelcontextprotocol/sdk` indirectly via pnpm-lock changes. Please confirm the lockfile bump is intentional and that `RequestOptions.timeout` exists in the resolved version; otherwise this silently does nothing or throws a type error at runtime.

**Suggested fix:** add a runtime guard so the call degrades gracefully if the SDK doesn't accept a third options arg:

```ts
const callResult = await withTimeout(
  client.callTool(
    { name: toolName, arguments: effectiveParameters },
    undefined,
    { timeout: effectiveTimeout },
  ),
  timeout,
  new Error(`Tool execution timeout: ${toolName}`),
);
```

is already present; additionally assert via a type-level test or feature-detect `typeof client.callTool.length >= 3` if SDK versions vary.

src/lib/providers/openaiChatCompletionsBase.ts
```diff
@@ -414,18 +414,36 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
             modelId,
```
Tara-ag, line 592: ⚠️ **MAJOR: `createTimeoutController` may return `undefined`; `timeoutController?.controller.signal` is fine, but `nlStepTimeoutMs` precedence ignores provider-specific timeouts**

The delegating model now prefers `providerOptions.neurolink.timeoutMs` over `getTimeoutForOptions(options)`. This means a provider-specific per-step timeout configured via the provider's own options (e.g. OpenAI-compatible `timeout` in `providerOptions.openai`) is silently overridden by the NeuroLink-orchestrated value. For providers where the AI SDK *does* carry a timeout in its own namespace, this is a regression.

**Suggested fix:** prefer the provider-specific namespace when present, and only fall back to `neurolink.timeoutMs`:

```ts
const nlStepTimeoutMs = (
  options.providerOptions as { neurolink?: { timeoutMs?: number } } | undefined
)?.neurolink?.timeoutMs;
const providerSpecificTimeout = /* extract from known provider namespaces if defined */;
const stepTimeoutMs =
  typeof providerSpecificTimeout === "number"
    ? providerSpecificTimeout
    : typeof nlStepTimeoutMs === "number"
      ? nlStepTimeoutMs
      : getTimeoutForOptions(options);
```

At minimum, document in the PR description that this change intentionally makes the outer `timeout` value win over any per-provider step timeout.

src/lib/core/modules/ToolsManager.ts
```diff
@@ -817,7 +901,16 @@ export class ToolsManager {
         const fixedSchema = this.utilities?.fixSchemaForOpenAIStrictMode
```
Tara-ag, line 926: 💡 **MINOR: `buildMCPSchemaValidator` recompiles the JSON Schema validator on every `getAllTools()` call**

`createExternalMCPTool` is invoked once per external MCP tool every time `getAllTools()` runs (i.e. every generation). Each invocation constructs a fresh `@cfworker/json-schema` `Validator` and rebuilds the contract string. For deployments with 50+ MCP tools this adds serial compilation latency on the hot path, and the work is identical on every call.

**Suggested fix:** cache the compiled validator per `(toolName, schema)` key (e.g. a module-level `WeakMap` or LRU keyed by the schema object / JSON-stable hash), or compile validators lazily on first use rather than eagerly during tool discovery. If compilation must stay eager, parallelize the per-tool `createExternalMCPTool` calls with `Promise.all` in `processExternalMCPTools` instead of awaiting them in a loop.

Tara-ag (changes requested):
## Autonomous review summary

**Files reviewed:** 11 changed files (package.json, pnpm-lock.yaml, GenerationHandler.ts, ToolsManager.ts, toolDiscoveryService.ts, openaiChatCompletionsBase.ts, proxyFetch.ts, errorHandling.ts, timeout.ts, toolCallRepair.ts, continuous-test-suite-litellm-parity.ts).

- 🔒 **CRITICAL:** 0
- ⚠️ **MAJOR:** 4
- 💡 **MINOR:** 1
- 💬 **SUGGESTION:** 0

**Blocking criterion triggered:** ≥3 MAJOR issues in the same pull request.

### MAJOR issues that must be addressed

1. **`src/lib/core/modules/GenerationHandler.ts` (provider-options collision)** — `providerOptions: { neurolink: { timeoutMs } }` and the `thinkingConfig` Google block both spread a top-level `providerOptions` into `generateText`. For thinking-enabled Gemini calls, the Google block overwrites the timeout forwarding, silently dropping the caller's explicit `timeout`. Merge the two namespaces before spreading.

2. **`src/lib/core/modules/GenerationHandler.ts` (prepareStep contract)** — The wrap-up branch overrides any caller-supplied `toolChoice` and appends a synthetic user message **after** `prepareStep` has already run. This breaks the public `prepareStep` contract (the callback no longer observes the messages the model receives) and changes behavior for callers using `toolChoice: "required"` or per-step tool selection. Document the override in the public option JSDoc and consider applying the nudge before invoking `prepareStep`.

3. **`src/lib/providers/openaiChatCompletionsBase.ts` (timeout precedence)** — The delegating model now unconditionally prefers `providerOptions.neurolink.timeoutMs` over any provider-specific per-step timeout. This regresses providers that already carry a timeout in their own namespace. Prefer a provider-specific namespace when present, or document the intentional precedence change.

4. **`src/lib/mcp/toolDiscoveryService.ts` (callTool signature / SDK version)** — The new `client.callTool(..., undefined, { timeout })` usage depends on a specific `@modelcontextprotocol/sdk` version and an explicit `undefined` result-schema argument. Confirm the lockfile bump is intentional and that `RequestOptions.timeout` exists in the resolved SDK; otherwise the call may silently fail or throw at runtime. Add a runtime guard or type-level test if SDK versions vary.

### MINOR issue

5. **`src/lib/core/modules/ToolsManager.ts` (validator recompilation)** — `buildMCPSchemaValidator` recompiles a fresh JSON Schema validator on every `getAllTools()` call. For large MCP tool sets this adds unnecessary hot-path latency. Cache validators per schema or compile them lazily/on-first-use.

### Next steps

Please address the four MAJOR issues above (or provide a documented justification / alternative fix for each), then re-request review. The MINORITY performance note can be addressed in this PR or deferred, but should not be silently ignored for large tool registries.

I did not re-raise any existing review comments (none were present). All counts above refer only to new issues raised in this review run.

src/lib/core/modules/GenerationHandler.ts
pdogra1299: Fixed in 87792bb0. Extracted a `buildProviderOptions()` helper that merges both namespaces (`neurolink.timeoutMs` + `google.thinkingConfig`) into ONE object before the single `providerOptions:` spread — the two conditional spreads are gone, so thinking-enabled Gemini calls keep the forwarded timeout. (Note the practical impact was nil — `providerOptions.neurolink` is only read by NeuroLink's own delegating chat-completions models, which are never Google providers — but the clobber pattern was real and is now structurally impossible.)

src/lib/core/modules/GenerationHandler.ts
pdogra1299: Addressed in 87792bb0 via documentation: the `turnTimeoutMs` and `wrapupTimeLeadMs` JSDoc in `types/generate.ts` now spells out the override semantics (forced `toolChoice: "none"` during the wrap-up window, overriding caller/prepareStep tool selection; the honest time message is appended after the caller's `prepareStep` runs and is not observed by it). Applying the nudge *before* `prepareStep` isn't possible without changing what the AI SDK hands the callback — `prepareStep` receives SDK-owned step options, and the public contract explicitly promises the ORIGINAL step options. The override itself is deliberate native-loop parity: the Vertex/Anthropic loops use the same `tool_choice: none` + honest-time-message backstop (PR #1125), on the principle that an honest partial beats a discarded turn.

src/lib/mcp/toolDiscoveryService.ts
pdogra1299: Verified — no code change needed. (1) The MCP SDK version did NOT change: the lockfile diff shows 1.28.0/1.29.0 unchanged; the churn is pnpm re-keying peer-dependency suffixes because `@cfworker/json-schema` (an *optional peerDependency of the MCP SDK itself*) is now installed — e.g. `@modelcontextprotocol/sdk@1.28.0(zod@4.3.6)` → `@modelcontextprotocol/sdk@1.28.0(@cfworker/json-schema@4.1.1)(zod@4.3.6)`. (2) `RequestOptions.timeout` exists in the resolved SDK: `shared/protocol.js` applies `options?.timeout ?? DEFAULT_REQUEST_TIMEOUT_MSEC` (60000) per request, and `client/index.d.ts` declares `callTool(params, resultSchema?, options?: RequestOptions)` — passing `undefined` for the defaulted `resultSchema` is valid TS for a defaulted parameter. (3) The no-API suite pins this contract: `test/continuous-test-suite-litellm-parity.ts` asserts the third argument arrives as `{ timeout: 120000 }` through a fake client, so an SDK signature change would fail CI rather than silently no-op.

src/lib/providers/openaiChatCompletionsBase.ts
pdogra1299: No code change — the premise doesn't hold on this path, and the precedence is intentional + now documented. There is no provider-specific per-step timeout namespace here: the delegating model IS the provider implementation for chat-completions providers (litellm & friends), and nothing on this path ever read a timeout from `providerOptions.openai` (or any other namespace) before this PR — the only pre-existing source was the AI-SDK call options, which never carry `timeout`, hence the hard-coded provider default. `providerOptions.neurolink.timeoutMs` is consumed exclusively by NeuroLink's own delegating models; stock AI-SDK providers read only their own namespace and ignore it. Precedence (explicit caller `timeout` > provider default) is the fix's purpose — a caller's `timeout: "15m"` previously bounded the outer loop while each step silently stayed capped at 5m — and is documented in the code comment at the read site and in the PR description.

src/lib/core/modules/GenerationHandler.ts
```diff
@@ -79,6 +86,66 @@ function safePreview(v: unknown): string {
...
+      : callerTimeoutMs;
+  const wrapupLeadMs = turnBudgetMs
+    ? Math.min(
+        options.wrapupTimeLeadMs ?? DEFAULT_WRAPUP_TIME_LEAD_MS,
+        Math.floor(turnBudgetMs / 4),
+      )
+    : 0;
+  const turnDeadline = turnBudgetMs ? Date.now() + turnBudgetMs : undefined;
+  return { callerTimeoutMs, turnBudgetMs, wrapupLeadMs, turnDeadline };
+}
```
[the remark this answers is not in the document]
pdogra1299: Fixed in 87792bb0 — good catch, this was the strongest finding. `callGenerateText` now takes `turnStartMs` (in a bundled `callConfig` param) and `resolveTurnBudget(options, turnStartMs)` anchors the deadline as `turnStartMs + turnBudgetMs`. All three call sites (initial, structured-output-conflict fallback, temperature-deprecated fallback) pass `executeGeneration`'s existing `genStartTime`, so `withProviderRetry` attempts and fallback retries all consume the SAME wall-clock budget instead of each minting a fresh one.

src/lib/core/modules/GenerationHandler.ts
```diff
@@ -258,15 +329,62 @@ export class GenerationHandler {
...
+          }) as ToolCallRepairFunction<ToolSet>,
+        }),
+      // Forward the caller's resolved timeout to the model layer: the AI-SDK
+      // V3 call options carry no `timeout`, so delegating chat-completions
+      // models (litellm & friends) could otherwise only ever apply their
+      // provider default per step — an explicit `timeout: "15m"` bounded the
+      // outer loop while each step stayed capped at the default.
+      ...(callerTimeoutMs !== undefined && {
+        providerOptions: { neurolink: { timeoutMs: callerTimeoutMs } },
+      }),
```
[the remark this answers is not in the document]
pdogra1299: Fixed in 87792bb0 — both spreads replaced by a single merged object built in `buildProviderOptions()` (neurolink timeout forwarding + google thinkingConfig in one `providerOptions`), so simultaneous conditions compose instead of clobbering.

src/lib/core/modules/ToolsManager.ts
```diff
@@ -33,6 +33,90 @@ function makeToolAbortError(): Error {
...
+          .slice(0, 3)
+          .map((e) => `${e.instanceLocation.replace(/^#\/?/, "")}: ${e.error}`)
+          .join("; ");
+        return {
+          success: false,
+          error: new Error(
+            `Invalid arguments for tool '${toolName}'${details ? ` — ${details}` : ""}. ` +
+              `Expected: { ${contract} } (send every non-optional property with the exact name and JSON type).`,
+          ),
+        };
```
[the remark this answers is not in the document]
pdogra1299: Fixed in 87792bb0. The `details` filter now keeps root-level (`#`) entries when they are missing-required errors (`/required property/i` on the error text — cfworker reports missing required props at the root instanceLocation), so the model sees the offending field name in details, not just the contract text. Suite asserts `/required property/i` appears in the validator's error message.

src/lib/core/modules/ToolsManager.ts
```diff
@@ -817,7 +901,16 @@ export class ToolsManager {
...
+        // A jsonSchema() wrapper WITHOUT a validate function is declarative
+        // only — the AI SDK passes any parsed arguments straight through
+        // (safeValidateTypes short-circuits on `validate == null`). That let
+        // malformed calls (missing required params, "123" for a number)
+        // reach execution, where the MCP-layer validator rejected them at
+        // the cost of a full model round-trip. Attach a real validator so
+        // invalid calls fail at parse time, where experimental_repairToolCall
+        // can still fix them silently.
+        const validate = await buildMCPSchemaValidator(tool.name, fixedSchema);
+        finalSchema = jsonSchema(fixedSchema, validate ? { validate } : {});
```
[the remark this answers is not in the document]
pdogra1299: Fixed in 87792bb0. Added a module-level `WeakMap` cache keyed by the ORIGINAL `tool.inputSchema` object (identity-stable for the lifetime of a server connection; rediscovery yields new objects so stale entries fall out via WeakMap semantics). `undefined` results (uncompilable schemas) are cached too via `has()`/`get()`, so failures aren't recompiled either.

src/lib/core/modules/ToolsManager.ts
pdogra1299: Fixed in 87792bb0 (same fix as CodeRabbit's duplicate finding): module-level `WeakMap<inputSchema, validator>` — one compile per schema object per server connection instead of one per `getAllTools()` call. Kept compilation eager-on-first-build rather than lazy-on-first-call so a broken schema logs at registration, not mid-agent-loop.

src/lib/mcp/toolDiscoveryService.ts
```diff
@@ -812,33 +827,70 @@ export class ToolDiscoveryService extends EventEmitter {
...
           );
+          coerced = coerced ?? { ...parameters };
+          coerced[propName] = coercedValue as JsonValue;
         }
+        this.validateParameterType(
+          propName,
+          (coerced ? coerced[propName] : originalValue) as JsonValue,
+          propSchema as JsonObject,
+        );
       }
```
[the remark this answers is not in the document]
pdogra1299: Fixed in 87792bb0. `validateParameterType` now has an explicit `"integer"` case: rejects non-numbers and non-integer numbers (`3.7`, `"abc"` after failed coercion) with `Parameter 'x' must be an integer, got …`. Suite covers both directions: `"42"` coerces to `42` and executes; `3.7` is rejected with the named error instead of being forwarded to the MCP server.

src/lib/core/modules/GenerationHandler.ts
```diff
@@ -1054,11 +1217,38 @@ export class GenerationHandler {
...
+    if (
+      (generateResult as { __nlTurnWrapup?: boolean }).__nlTurnWrapup === true
+    ) {
+      stopReason = "time-limit";
+    } else if (
+      stepsUsed !== undefined &&
+      stepsUsed >= maxSteps &&
+      generateResult.finishReason === "tool-calls"
+    ) {
+      stopReason = "step-cap";
```
Tara-ag, line 1250: ⚠️ **MAJOR: `finishReason` compared as string but delegating providers return `{ unified, raw }`**

`OpenAIChatCompletionsProvider.doGenerate` and `AnthropicProvider.getAISDKModel().doGenerate` both return `finishReason: { unified, raw }`. This block compares `generateResult.finishReason` against `"tool-calls"` and `"error"` as if it were a string.

Concrete impact on the AI-SDK loop path:
- `step-cap` is **never** detected for OpenAI-compatible/Anthropic providers: `({ unified, raw }) === "tool-calls"` is always false.
- A turn that actually hit the step ceiling is mis-reported as `stopReason: "completed"` because the object `!== "error"` check always passes.
- `provider-error` exits stay `stopReason: undefined` for the same reason.

This undermines the PR’s stated parity goal for litellm and every other chat-completions provider that uses the delegating model.

**Suggested fix:** normalize the finish reason once before the comparisons:

```ts
const finishReason =
  typeof generateResult.finishReason === "string"
    ? generateResult.finishReason
    : (generateResult.finishReason as { unified?: string } | undefined)
        ?.unified;
```

Then compare `finishReason === "tool-calls"` / `finishReason !== "error"`. Also update the parity test to exercise the object shape so the gap doesn’t regress.

src/lib/core/modules/GenerationHandler.ts
```diff
@@ -1054,11 +1217,38 @@ export class GenerationHandler {
...
+    } else if (
+      stepsUsed !== undefined &&
+      generateResult.finishReason !== "error"
+    ) {
+      stopReason = "completed";
+    }
+
     return {
```
Tara-ag, line 1262: ⚠️ **MAJOR: `stopReason: "completed"` is set even when the turn ended on `length` / `content-filter` / `error`**

The current logic falls through to `stopReason = "completed"` for any `finishReason !== "error"`. That means a turn truncated by the output token cap (`finishReason === "length"`) or blocked by content-filter is reported as a clean completion.

This contradicts the documented `GenerateStopReason` semantics and hides truncation from callers. The native loops map `length` to a meaningful state; the AI-SDK path should too.

**Suggested fix:** after normalizing `finishReason`, map it explicitly:

```ts
if (finishReason === "tool-calls" && stepsUsed >= maxSteps) {
  stopReason = "step-cap";
} else if (finishReason === "length") {
  stopReason = "completed"; // or introduce a "truncated" stop reason
} else if (finishReason === "error" || finishReason === "content-filter") {
  stopReason = "provider-error";
} else if (finishReason === "stop") {
  stopReason = "completed";
}
```

At minimum, do not emit `"completed"` for `"length"` or `"content-filter"`.

src/lib/utils/timeout.ts
```diff
@@ -484,30 +484,93 @@ export function composeAbortSignals(
...
+    { once: true },
+  );
+  // Pin the derived signal to the returned controller: sources hold only
+  // weak refs to `any()` dependents, so without this strong ref the derived
+  // signal (and the forward listener with it) could be GC'd before firing.
+  (
+    controller as AbortController & { __nlMergedSignal?: AbortSignal }
+  ).__nlMergedSignal = merged;
+
   return controller;
```
Tara-ag, line 582: ⚠️ **MAJOR: `mergeAbortSignals` pins `AbortSignal.any` result but the controller itself is never aborted when sources are GC'd**

The strong reference `__nlMergedSignal` keeps the derived `any()` signal alive, but if all source signals become unreachable, the derived signal (and its internal listeners) is still held by the controller. For callers that create many short-lived merged controllers, this retains one `AbortSignal` + listener closure per call until the controller itself is GC'd.

More importantly, the comment claims "no per-source listeners are left behind on long-lived input signals once the merged controller becomes unreachable." That is true for the source signals, but the *derived* `merged` signal still holds a closure over the controller and the forward listener. If the controller is retained elsewhere (e.g., returned to a long-lived stream), the derived signal and its internals are retained too. The previous implementation at least released everything when the controller was GC'd; this version adds an extra strong ref that may extend lifetime.

**Suggested fix:** consider whether the pin is necessary. `AbortSignal.any` implementations in modern Node.js keep dependents alive via the source signals' internal slots; if the implementation in this repo's supported runtimes already does that, the `__nlMergedSignal` pin is unnecessary and harmful. If a pin is required for some runtime, document which one and add a test that verifies the forward listener fires without the pin on the minimum supported Node version.

src/lib/utils/timeout.ts
```diff
@@ -484,30 +484,93 @@ export function composeAbortSignals(
...
+  }
+  return { signal: controller.signal, dispose };
+}
+
 /**
  * Merge abort signals (for combining user abort with timeout)
+ *
+ * Implemented via `AbortSignal.any` with a single once-listener forward, so
+ * no per-source listeners are left behind on long-lived input signals once
+ * the merged controller becomes unreachable (registrations are released with
```
Tara-ag, line 542: 💡 **MINOR: `composeAbortSignalsScoped` does not remove listeners when a source aborts before `dispose()` is called**

If one source aborts, `onAbort` aborts the controller and the loop breaks, but the remaining source still has its `abort` listener attached. `dispose()` will remove it when called, but if the caller only checks `signal.aborted` and never invokes `dispose()`, the listener leaks on the non-aborted source.

**Suggested fix:** in the `for (const source of sources)` pre-check loop, after calling `onAbort(source)` and breaking, also call `dispose()` immediately so all listeners are detached. Alternatively, document that `dispose()` must always be called even after abort.

src/lib/core/modules/GenerationHandler.ts
```diff
@@ -79,6 +86,112 @@ function safePreview(v: unknown): string {
...
+  wrapupLeadMs: number;
+  turnDeadline: number | undefined;
+} {
+  const callerTimeoutMs = parseTimeout(options.timeout);
+  const turnBudgetMs =
+    typeof options.turnTimeoutMs === "number" && options.turnTimeoutMs > 0
+      ? options.turnTimeoutMs
+      : callerTimeoutMs;
+  const wrapupLeadMs = turnBudgetMs
+    ? Math.min(
```
Tara-ag, line 131: 💡 **MINOR: `resolveTurnBudget` silently ignores invalid `turnTimeoutMs` values**

`turnTimeoutMs <= 0` falls through to `callerTimeoutMs` without any warning. A caller who accidentally passes `0` or a negative number will get the outer `timeout` behavior instead of the expected no-deadline behavior, and may not notice the misconfiguration.

**Suggested fix:** log a debug warning when `turnTimeoutMs` is non-positive, or document that only positive values engage the turn deadline.

Tara-ag (changes requested):
## Autonomous review summary

**Files reviewed:** 12 changed files.

- 🔒 **CRITICAL:** 0
- ⚠️ **MAJOR:** 3
- 💡 **MINOR:** 2
- 💬 **SUGGESTION:** 0

**Blocking criterion triggered:** ≥3 MAJOR issues in the same pull request.

### MAJOR issues that must be addressed

1. **`src/lib/core/modules/GenerationHandler.ts` (finishReason object/string mismatch)** — `formatEnhancedResult` compares `generateResult.finishReason` against `"tool-calls"` and `"error"` as if it were a string, but delegating providers (OpenAI-compatible/Anthropic) return `{ unified, raw }`. This means `step-cap` is never detected and error/truncated exits are misreported as `completed`. Normalize `finishReason` to `.unified` before comparing.

2. **`src/lib/core/modules/GenerationHandler.ts` (stopReason fallthrough)** — The fall-through to `stopReason = "completed"` for any `finishReason !== "error"` reports output-token truncation (`length`) and content-filter blocks as clean completions. Map `length` and `content-filter` explicitly instead of collapsing them into `completed`.

3. **`src/lib/utils/timeout.ts` (mergeAbortSignals lifetime pin)** — The `__nlMergedSignal` strong reference keeps the derived `AbortSignal.any()` signal alive for the lifetime of the returned controller, retaining listener closures. Verify whether the pin is required on the supported Node runtime; if not, remove it. If it is required, document the runtime and add a test proving the forward listener fires without the pin.

### MINOR issues

4. **`src/lib/utils/timeout.ts` (composeAbortSignalsScoped listener leak)** — When a source aborts, the remaining sources still have their `abort` listeners attached until `dispose()` is called. Call `dispose()` inside the abort handler before breaking, or document that `dispose()` must always be invoked even after abort.

5. **`src/lib/core/modules/GenerationHandler.ts` (resolveTurnBudget silent invalid values)** — Non-positive `turnTimeoutMs` silently falls back to `callerTimeoutMs`. Log a debug warning or document that only positive values engage the turn deadline.

### Next steps

Please address the three MAJOR issues above (or provide a documented justification / alternative fix for each), then re-request review. The two MINOR notes can be addressed in this PR or deferred.

I did not re-raise any resolved/outdated existing review comments. All counts above refer only to new issues raised in this review run.

src/lib/core/modules/GenerationHandler.ts
```diff
@@ -1054,11 +1222,38 @@ export class GenerationHandler {
...
+    if (
+      (generateResult as { __nlTurnWrapup?: boolean }).__nlTurnWrapup === true
+    ) {
+      stopReason = "time-limit";
+    } else if (
+      stepsUsed !== undefined &&
+      stepsUsed >= maxSteps &&
+      generateResult.finishReason === "tool-calls"
+    ) {
+      stopReason = "step-cap";
```
Tara-ag, line 1250: ⚠️ **MAJOR: `finishReason` compared as string but delegating providers return `{ unified, raw }`**

`OpenAIChatCompletionsProvider.doGenerate` and `AnthropicProvider.getAISDKModel().doGenerate` both return `finishReason: { unified, raw }`. This block compares `generateResult.finishReason` against `"tool-calls"` and `"error"` as if it were a string.

Concrete impact on the AI-SDK loop path:
- `step-cap` is **never** detected for OpenAI-compatible/Anthropic providers: `({ unified, raw }) === "tool-calls"` is always false.
- A turn that actually hit the step ceiling is mis-reported as `stopReason: "completed"` because the object `!== "error"` check always passes.
- `provider-error` exits stay `stopReason: undefined` for the same reason.

This undermines the PR’s stated parity goal for litellm and every other chat-completions provider that uses the delegating model.

**Suggested fix:** normalize the finish reason once before the comparisons:

```ts
const finishReason =
  typeof generateResult.finishReason === "string"
    ? generateResult.finishReason
    : (generateResult.finishReason as { unified?: string } | undefined)
        ?.unified;
```

Then compare `finishReason === "tool-calls"` / `finishReason !== "error"`. Also update the parity test to exercise the object shape so the gap doesn’t regress.

src/lib/core/modules/GenerationHandler.ts
```diff
@@ -1054,11 +1222,38 @@ export class GenerationHandler {
...
+    } else if (
+      stepsUsed !== undefined &&
+      generateResult.finishReason !== "error"
+    ) {
+      stopReason = "completed";
+    }
+
     return {
```
Tara-ag, line 1262: ⚠️ **MAJOR: `stopReason: "completed"` is set even when the turn ended on `length` / `content-filter` / `error`**

The current logic falls through to `stopReason = "completed"` for any `finishReason !== "error"`. That means a turn truncated by the output token cap (`finishReason === "length"`) or blocked by content-filter is reported as a clean completion.

This contradicts the documented `GenerateStopReason` semantics and hides truncation from callers. The native loops map `length` to a meaningful state; the AI-SDK path should too.

**Suggested fix:** after normalizing `finishReason`, map it explicitly:

```ts
if (finishReason === "tool-calls" && stepsUsed >= maxSteps) {
  stopReason = "step-cap";
} else if (finishReason === "length") {
  stopReason = "completed"; // or introduce a "truncated" stop reason
} else if (finishReason === "error" || finishReason === "content-filter") {
  stopReason = "provider-error";
} else if (finishReason === "stop") {
  stopReason = "completed";
}
```

At minimum, do not emit `"completed"` for `"length"` or `"content-filter"`.

src/lib/providers/openaiChatCompletionsClient.ts
```diff
@@ -34,10 +34,90 @@ import type {
   DeferredUsage,
   Tool,
 } from "../types/index.js";
-import { convertZodToJsonSchema } from "../utils/schemaConversion.js";
+import { sanitizeToolName } from "../mcp/toolConverter.js";
+import {
+  convertZodToJsonSchema,
+  normalizeWireToolSchema,
```
Tara-ag, line 40: 💡 **MINOR: `WIRE_TOOL_NAME_RE` allows hyphens, but OpenAI's function-name spec only allows `[a-zA-Z0-9_-]` in practice; however the comment says `^[a-zA-Z0-9_-]{0,63}$` while the regex is `^[a-zA-Z_][a-zA-Z0-9_-]{0,63}$`. The first-char restriction is correct, but the comment should match the regex exactly to avoid future drift.**

Not blocking, but consider aligning the comment with the regex or adding a short note about why the first character is restricted.

src/lib/core/modules/GenerationHandler.ts
Tara-ag, line 131: 💡 **MINOR: `resolveTurnBudget` silently ignores invalid `turnTimeoutMs` values**

`turnTimeoutMs <= 0` falls through to `callerTimeoutMs` without any warning. A caller who accidentally passes `0` or a negative number will get the outer `timeout` behavior instead of the expected no-deadline behavior, and may not notice the misconfiguration.

**Suggested fix:** log a debug warning when `turnTimeoutMs` is non-positive, or document that only positive values engage the turn deadline.

src/lib/utils/timeout.ts
Tara-ag, line 542: 💡 **MINOR: `composeAbortSignalsScoped` does not remove listeners when a source aborts before `dispose()` is called**

If one source aborts, `onAbort` aborts the controller and the loop breaks, but the remaining source still has its `abort` listener attached. `dispose()` will remove it when called, but if the caller only checks `signal.aborted` and never invokes `dispose()`, the listener leaks on the non-aborted source.

**Suggested fix:** in the `for (const source of sources)` pre-check loop, after calling `onAbort(source)` and breaking, also call `dispose()` immediately so all listeners are detached. Alternatively, document that `dispose()` must always be called even after abort.

pnpm-lock.yaml
Tara-ag, line 1: ⚠️ **MAJOR: Lockfile downgrades multiple transitive AI SDK packages without changing their specifiers**

The `pnpm-lock.yaml` diff downgrades several packages that are not directly related to the `@cfworker/json-schema` addition:

- `@ai-sdk/anthropic` 3.0.84 → 3.0.81
- `@ai-sdk/azure` 3.0.74 → 3.0.68
- `@ai-sdk/google-vertex` 4.0.145 → 4.0.142
- `@ai-sdk/google` 3.0.82 → 3.0.80
- `@ai-sdk/openai-compatible` 2.0.50 → 2.0.48
- `@openrouter/ai-sdk-provider` 2.9.1 → 2.9.0
- `@types/aws-lambda` 8.10.162 → 8.10.161
- `pg-protocol` 1.14.0 → 1.13.0

These appear to be side-effects of pnpm re-resolving the graph after adding `@cfworker/json-schema` as an optional peer of `@modelcontextprotocol/sdk`. The actual installed versions of provider SDKs are changing, which introduces regression risk (behavioral differences, bug reintroductions) that the PR description does not mention or justify.

**Suggested fix:** minimize the lockfile churn to only what is required for `@cfworker/json-schema`. Try regenerating with `pnpm install --no-frozen-lockfile` after explicitly pinning the AI SDK packages to their current versions in `package.json`, or use `pnpm update "@ai-sdk/*"` to restore the newer resolutions. If the downgrades are unavoidable due to peer-dependency constraints, document each changed version in the PR description and confirm the parity test suite still passes against the downgraded packages.

src/lib/utils/timeout.ts
Tara-ag, line 582: ⚠️ **MAJOR: `mergeAbortSignals` pins `AbortSignal.any` result but the controller itself is never aborted when sources are GC'd**

The strong reference `__nlMergedSignal` keeps the derived `any()` signal alive, but if all source signals become unreachable, the derived signal (and its internal listeners) is still held by the controller. For callers that create many short-lived merged controllers, this retains one `AbortSignal` + listener closure per call until the controller itself is GC'd.

More importantly, the comment claims "no per-source listeners are left behind on long-lived input signals once the merged controller becomes unreachable." That is true for the source signals, but the *derived* `merged` signal still holds a closure over the controller and the forward listener. If the controller is retained elsewhere (e.g., returned to a long-lived stream), the derived signal and its internals are retained too. The previous implementation at least released everything when the controller was GC'd; this version adds an extra strong ref that may extend lifetime.

**Suggested fix:** consider whether the pin is necessary. `AbortSignal.any` implementations in modern Node.js keep dependents alive via the source signals' internal slots; if the implementation in this repo's supported runtimes already does that, the `__nlMergedSignal` pin is unnecessary and harmful. If a pin is required for some runtime, document which one and add a test that verifies the forward listener fires without the pin on the minimum supported Node version.

Tara-ag (changes requested):
## Autonomous review summary

Reviewed **25 changed files** (focusing on the 18 source/test files; lockfile reviewed as a single unit).

| Severity | Count | Files |
|----------|-------|-------|
| 🔒 CRITICAL | 0 | — |
| ⚠️ MAJOR | 4 | `pnpm-lock.yaml`, `src/lib/core/modules/GenerationHandler.ts` (×2), `src/lib/utils/timeout.ts` |
| 💡 MINOR | 3 | `src/lib/core/modules/GenerationHandler.ts`, `src/lib/utils/timeout.ts`, `src/lib/providers/openaiChatCompletionsClient.ts` |

### Blocking reason

This PR meets the **“3 or more MAJOR severity issues”** blocking criterion:

1. **`pnpm-lock.yaml` downgrades unrelated transitive packages** (MAJOR) — several `@ai-sdk/*` provider SDKs, `@openrouter/ai-sdk-provider`, `@types/aws-lambda`, and `pg-protocol` are rolled back without justification or updated specifiers, introducing regression risk.

2. **`GenerationHandler.formatEnhancedResult` compares `finishReason` as a string** (MAJOR) — delegating chat-completions providers return `finishReason: { unified, raw }`, so `step-cap` is never detected and `provider-error` exits stay `stopReason: undefined`, directly undermining the PR’s parity goal.

3. **`stopReason: "completed"` is emitted for `length` / `content-filter` exits** (MAJOR) — truncation and content-filter blocks are reported as clean completions, hiding real failure modes from callers.

4. **`mergeAbortSignals` strong-pins the `AbortSignal.any()` result** (MAJOR) — the `__nlMergedSignal` field may extend lifetime of the derived signal and its listener closure; the benefit over the previous implementation is unverified and the comment’s “no listeners left behind” claim is incomplete.

### Next steps

- Normalize `finishReason` in `formatEnhancedResult` before comparing it (handle both `string` and `{ unified }` shapes).
- Map `finishReason` values explicitly so `length` / `content-filter` / `error` do not fall through to `completed`.
- Minimize `pnpm-lock.yaml` churn to only what `@cfworker/json-schema` requires, or document every unavoidable downgrade.
- Revisit the `__nlMergedSignal` pin in `mergeAbortSignals`; if it is required, add a test proving the forward listener fires without it on the minimum supported Node version, and document the runtime.
- Address the three MINOR notes at your discretion.

Once the four MAJOR items are resolved, I’m happy to re-review.

src/lib/core/modules/GenerationHandler.ts
pdogra1299: Verified — not an issue at this layer. The `{ unified, raw }` object is the LanguageModelV3 **doGenerate** contract; `generateText` normalizes it before exposing the result: `ai/dist/index.mjs` maps `finishReason: currentModelResponse.finishReason.unified`, so `formatEnhancedResult` (which receives the `generateText` result, not the raw model response) always sees a plain string. Live-verified on litellm through the delegating model: `finishReason: "stop"`, and the `step-cap`/`time-limit` branches fire correctly (covered in `test:litellm-parity`). No change needed.

src/lib/core/modules/GenerationHandler.ts
pdogra1299: Partially accepted, fixed in 6c413e2c: a `finishReason === "error"` turn now maps to `stopReason: "provider-error"` (previously left `undefined`), matching `resolveTurnStopReason`, with a parity test added.

`length`/`content-filter` → `"completed"` is deliberate and stays: it is exact parity with the native loops' `resolveTurnStopReason` (googleNativeGemini3.ts), where only `"error"` maps to `provider-error` and everything else is `completed`. In the published `GenerateStopReason` contract, `completed` means "the loop ended on its own, not cut by a NeuroLink bound" — truncation is signaled via `finishReason`/`rawFinishReason`/`jsonTruncated`, never via stopReason (there is no `truncated` state, and introducing one would break stopReason consumers).

src/lib/utils/timeout.ts
pdogra1299: The pin is required, not harmful, on our supported runtimes: Node's `AbortSignal.any()` holds its dependents **weakly** (that is documented Node behavior), so without the strong ref the derived signal — and the forward listener with it — can be GC'd before it ever fires, silently breaking abort propagation. The pin's lifetime is exactly the returned controller's lifetime: one small signal object per call, released together with the controller; there is no unbounded growth. The previous implementation's real cost was a *permanent* listener per call on long-lived SOURCE signals — the `MaxListenersExceededWarning` observed in production at 10+ steps — which is exactly what this replaces. Keeping as-is.

src/lib/utils/timeout.ts
pdogra1299: Fixed in 6c413e2c: `onAbort` now detaches all source listeners at abort time (self-clean), so a caller that only observes `signal.aborted` and never reaches its `dispose()` leaves nothing behind on the non-aborted source; `dispose()` stays idempotent. Covered by a new parity test asserting zero listeners on the other source immediately after abort, without `dispose()`.

src/lib/core/modules/GenerationHandler.ts
pdogra1299: Fixed in 6c413e2c: an explicitly-provided but invalid `turnTimeoutMs` (0, negative, or NaN) now logs a warning naming the value before falling back to the `timeout` option, instead of being silently ignored.

src/lib/core/modules/GenerationHandler.ts
pdogra1299: Verified — not an issue at this layer. The `{ unified, raw }` object is the LanguageModelV3 **doGenerate** contract; `generateText` normalizes it before exposing the result: `ai/dist/index.mjs` maps `finishReason: currentModelResponse.finishReason.unified`, so `formatEnhancedResult` (which receives the `generateText` result, not the raw model response) always sees a plain string. Live-verified on litellm through the delegating model: `finishReason: "stop"`, and the `step-cap`/`time-limit` branches fire correctly (covered in `test:litellm-parity`). No change needed.

src/lib/core/modules/GenerationHandler.ts
pdogra1299: Partially accepted, fixed in 6c413e2c: a `finishReason === "error"` turn now maps to `stopReason: "provider-error"` (previously left `undefined`), matching `resolveTurnStopReason`, with a parity test added.

`length`/`content-filter` → `"completed"` is deliberate and stays: it is exact parity with the native loops' `resolveTurnStopReason` (googleNativeGemini3.ts), where only `"error"` maps to `provider-error` and everything else is `completed`. In the published `GenerateStopReason` contract, `completed` means "the loop ended on its own, not cut by a NeuroLink bound" — truncation is signaled via `finishReason`/`rawFinishReason`/`jsonTruncated`, never via stopReason (there is no `truncated` state, and introducing one would break stopReason consumers).

src/lib/providers/openaiChatCompletionsClient.ts
pdogra1299: Fixed in 6c413e2c: the comment now states the exact pattern the regex enforces, including why the first character is restricted to letter/underscore (some chat templates reject leading digits/hyphens as identifiers) and that the alphabet mirrors `sanitizeToolName`'s output.

src/lib/core/modules/GenerationHandler.ts
pdogra1299: Fixed in 6c413e2c: an explicitly-provided but invalid `turnTimeoutMs` (0, negative, or NaN) now logs a warning naming the value before falling back to the `timeout` option, instead of being silently ignored.

src/lib/utils/timeout.ts
pdogra1299: Fixed in 6c413e2c: `onAbort` now detaches all source listeners at abort time (self-clean), so a caller that only observes `signal.aborted` and never reaches its `dispose()` leaves nothing behind on the non-aborted source; `dispose()` stays idempotent. Covered by a new parity test asserting zero listeners on the other source immediately after abort, without `dispose()`.

pnpm-lock.yaml
pdogra1299: Confirmed and fixed in 6c413e2c. Root cause: adding `@cfworker/json-schema` (optional peer of `@modelcontextprotocol/sdk`) creates NEW peer-variant instances of several packages, and the previous regeneration resolved those fresh instances against **stale cached registry metadata**, which is where the downgrades (`@ai-sdk/anthropic` 3.0.84→3.0.81 etc.) came from. The lockfile is now regenerated from the release base with fresh metadata: verified programmatically that the only version-level change in the whole diff is the `@cfworker/json-schema@4.1.1` addition — every other changed line is a peer-suffix rename at identical versions. Full suite re-run against the corrected resolutions (parity 19, vitest 27, mcp 84/24/4, check/lint/build clean).

src/lib/utils/timeout.ts
pdogra1299: The pin is required, not harmful, on our supported runtimes: Node's `AbortSignal.any()` holds its dependents **weakly** (that is documented Node behavior), so without the strong ref the derived signal — and the forward listener with it — can be GC'd before it ever fires, silently breaking abort propagation. The pin's lifetime is exactly the returned controller's lifetime: one small signal object per call, released together with the controller; there is no unbounded growth. The previous implementation's real cost was a *permanent* listener per call on long-lived SOURCE signals — the `MaxListenersExceededWarning` observed in production at 10+ steps — which is exactly what this replaces. Keeping as-is.

src/lib/providers/litellm.ts
```diff
@@ -42,6 +45,167 @@ const getLiteLLMConfig = () => ({
...
+        { error: error instanceof Error ? error.message : String(error) },
+      );
+    })
+    .finally(() => {
+      modelInfoInFlight.delete(key);
+    });
+  modelInfoInFlight.set(key, discovery);
+  return discovery;
+}
+
```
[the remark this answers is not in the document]
pdogra1299: Partially addressed in 6c413e2c, remainder agreed-and-deferred. Addressed: discovered limits are now tracked per base URL and registrations **min-merge across every discovered deployment**, so two proxies sharing a model-group name can no longer overstate each other's budgets via last-writer-wins — budgets stay conservative-safe for all deployments in the process; single-proxy processes (the operating model — one `LITELLM_BASE_URL` per process, same assumption as the pre-existing class-static models cache) are byte-identical. Vitest coverage added.

Deferred (your "heavy lift" flag is right): fully identity-scoped lookups require plumbing deployment identity through every budget call site — `checkContextBudget`, `getSafeMaxTokens`, and `stepBudgetGuard` carry only `(provider, model)` today. On the apiKey point: `/model/info` limits are deployment properties (the serving model's max input/output), not team quotas — a key that cannot see a model leaves the static defaults in force, which degrades safely.

package.json
Tara-ag, line 1: ✅ Adding `@cfworker/json-schema` as a direct dependency is correct since `ToolsManager.ts` imports it directly. Making it explicit pins the version the project tests against and avoids relying solely on the MCP SDK's optional peer dependency resolution.

src/lib/core/modules/GenerationHandler.ts
```diff
@@ -255,55 +393,85 @@ export class GenerationHandler {
         const callerMessages = (
```
Tara-ag, line 393: ⚠️ **MAJOR: `disableToolCallRepair` is read from `TextGenerationOptions` but is not declared in that type**

The spread at line 393 casts `options` to `{ disableToolCallRepair?: boolean }`, but `TextGenerationOptions` (in `src/lib/types/generate.ts`) does not include this field — it only exists on `StreamOptions`. This means TypeScript callers using `generate()` won't see the option, and the cast papers over the missing type.

Either add `disableToolCallRepair?: boolean` to `TextGenerationOptions` (mirroring `StreamOptions`) or gate repair via a field that already exists on both types. Without the type declaration this is a latent API inconsistency.

**Suggested fix:** add to `TextGenerationOptions` in `src/lib/types/generate.ts`:
```ts
/** Disable the schema-driven tool call repair mechanism (BZ-665). Default: false (repair enabled). */
disableToolCallRepair?: boolean;
```

src/lib/providers/openaiChatCompletionsClient.ts
```diff
@@ -34,10 +34,92 @@ import type {
...
+):
+  | { toWire: Map<string, string>; fromWire: Map<string, string> }
+  | undefined => {
+  if (names.every((name) => WIRE_TOOL_NAME_RE.test(name))) {
+    return undefined;
+  }
+  const toWire = new Map<string, string>();
+  const fromWire = new Map<string, string>();
+  for (const name of names) {
+    let wire = WIRE_TOOL_NAME_RE.test(name) ? name : sanitizeToolName(name);
```
Tara-ag, line 76: ⚠️ **MAJOR: `sanitizeToolName` can return an empty string, producing an invalid OpenAI wire name**

`sanitizeToolName` replaces every invalid char with `_`, trims leading non-letters by prepending `_`, then trims to 64 chars. For inputs like `""`, `"!!!"`, or all-special-character names, the result is `""` (or `"_"` → trimmed to `""` in some paths). The ternary then emits that empty string as the wire name, which violates the OpenAI function-name spec and will be rejected by the backend.

**Suggested fix:** guarantee a non-empty fallback before collision handling:
```ts
let wire = WIRE_TOOL_NAME_RE.test(name) ? name : sanitizeToolName(name);
if (!wire) {
  wire = "_tool";
}
```

Also add an empty-name case to the mapping test in `test/continuous-test-suite-litellm-parity.ts`.

src/lib/providers/openaiChatCompletionsClient.ts
```diff
@@ -34,10 +34,92 @@ import type {
...
+    if (fromWire.has(wire)) {
```
Tara-ag, line 77: 💡 **MINOR: `sanitizeToolName` output is not guaranteed to satisfy `WIRE_TOOL_NAME_RE`**

Three edge cases slip through the current ternary:

1. Empty or all-special-character names sanitize to `""`.
2. Names starting with a digit are returned unchanged by `sanitizeToolName` (e.g. `"123tool"` → `"123tool"`), but the regex requires the first character to be `[a-zA-Z_]`.
3. Names longer than 64 characters are not truncated by `sanitizeToolName`, so a 65-character valid name still fails the regex.

Force validity before collision handling:
```ts
let wire = WIRE_TOOL_NAME_RE.test(name) ? name : sanitizeToolName(name);
if (!wire || !/^[a-zA-Z_]/.test(wire)) {
  wire = `_${wire}`;
}
if (wire.length > 64) {
  wire = wire.slice(0, 64);
}
```

This guarantees every emitted wire name is within the allowed alphabet and length; the existing collision loop will still disambiguate duplicates.

src/lib/providers/litellm.ts
```diff
@@ -42,6 +45,209 @@ const getLiteLLMConfig = () => ({
...
+ * so when a process talks to MULTIPLE proxies that share a model-group name,
+ * registrations use the MIN across every discovered deployment: budgets stay
+ * safe for all of them instead of last-writer-wins silently overstating one.
+ * Single-proxy processes (the normal case) are unaffected.
+ */
+const modelInfoLimitsByURL = new Map<
+  string,
+  Map<string, { maxInputTokens?: number; maxOutputTokens?: number }>
+>();
+
```
Tara-ag, line 69: 💡 **MINOR: Hand-rolled timeout duplicates the file's existing `createTimeoutController` pattern**

`fetchLiteLLMModelLimits` manually manages `AbortController` + `setTimeout`/`clearTimeout` for its 5s timeout, mirroring the pre-existing `fetchModelsFromAPI`. Meanwhile `callEmbeddings` in the same file already uses `createTimeoutController`. Using the shared utility would reduce duplicated timeout/abort plumbing and keep abort-to-NetworkError conversion consistent.

**Suggested fix:** use `createTimeoutController(5000, "litellm", "generate")` and pass its `controller.signal` to `proxyFetch`, then call `timeoutController.cleanup()` in a `finally` block.

src/lib/context/errorDetection.ts
```diff
@@ -106,6 +106,12 @@ export function getContextOverflowProvider(error: unknown): string | null {
 export function parseProviderOverflowDetails(error: unknown): {
```
Tara-ag, line 106: 💡 **MINOR: `requestedOutputTokens` is added to the API but never consumed**

`parseProviderOverflowDetails` now returns `requestedOutputTokens`, but the recovery paths in `openaiChatCompletionsBase.ts` (`correctBodyAfterContextOverflow`) only read `actualTokens` and `budgetTokens`. The new field adds unused API surface.

Either use it in recovery (e.g. when the error states requested output separately, prefer `budgetTokens - actualTokens` directly, which is already done) or remove the field and its documentation until a caller needs it.

src/lib/proxy/proxyFetch.ts
```diff
@@ -106,6 +106,49 @@ function extractHostname(url: string | URL | RequestInfo): string {
   }
 }
 
+/**
```
Tara-ag, line 119: 💡 **MINOR: `transientCodes` Set is recreated on every error check**

The `transientCodes` Set is allocated inside `isTransientNetworkError`, so every transient-classification call allocates a new Set. Moving it to module scope avoids that allocation on the hot retry path.

**Suggested fix:**
```ts
const TRANSIENT_CODES = new Set([
  "ECONNRESET",
  "ETIMEDOUT",
  "ECONNREFUSED",
  "EPIPE",
  "UND_ERR_SOCKET",
  "UND_ERR_CONNECT_TIMEOUT",
]);

export function isTransientNetworkError(error: unknown): boolean {
  // ... use TRANSIENT_CODES.has(err.code)
}
```

src/lib/core/modules/GenerationHandler.ts
```diff
@@ -79,6 +86,121 @@ function safePreview(v: unknown): string {
   }
```
Tara-ag, line 86: ✅ Good: `resolveTurnBudget` now anchors the deadline to `turnStartMs`, so fallback retries and provider retries consume the same wall-clock budget instead of refreshing it. This directly addresses the prior review finding about retry budget resets.

src/lib/core/modules/GenerationHandler.ts
```diff
@@ -79,6 +86,121 @@ function safePreview(v: unknown): string {
...
+ * other (object spread does not deep-merge).
+ */
+function buildProviderOptions(
+  options: TextGenerationOptions,
+  isGoogleProvider: boolean,
+  callerTimeoutMs: number | undefined,
+): Parameters<typeof generateText>[0]["providerOptions"] {
+  const providerOptions: Record<string, Record<string, unknown>> = {};
+  if (callerTimeoutMs !== undefined) {
+    providerOptions.neurolink = { timeoutMs: callerTimeoutMs };
```
Tara-ag, line 155: ✅ Good: `buildProviderOptions` merges `neurolink.timeoutMs` and `google.thinkingConfig` into a single object before spreading, eliminating the prior clobber bug between the two conditional `providerOptions` spreads.

src/lib/core/modules/GenerationHandler.ts
```diff
@@ -1054,11 +1231,42 @@ export class GenerationHandler {
       : undefined;
```
Tara-ag, line 1230: ✅ Good: `formatEnhancedResult` now populates `stopReason` and `stepsUsed` for the AI-SDK loop path, and the `finishReason === "error"` branch maps to `provider-error`. This matches the native-loop parity goal.

src/lib/utils/timeout.ts
```diff
@@ -484,30 +484,100 @@ export function composeAbortSignals(
   return externalSignal ?? timeoutSignal;
```
Tara-ag, line 484: ✅ Good: `composeAbortSignalsScoped` provides explicit `dispose()` and self-cleans on abort, solving the `MaxListenersExceededWarning` accumulation on long-lived generate-call signals.

src/lib/mcp/toolDiscoveryService.ts
Tara-ag, line 625: ✅ Good: passing `{ timeout: effectiveTimeout }` as MCP `RequestOptions` ensures configured server timeouts above the SDK's 60s default actually take effect and cancel the transport request. The no-API suite's fake-client assertion provides regression coverage for the third-argument shape.

Tara-ag (changes requested):
## Autonomous review summary

Reviewed **25 changed files** (18 source/test files plus lockfile/package.json).

| Severity | Count | Files |
|----------|-------|-------|
| 🔒 CRITICAL | 0 | — |
| ⚠️ MAJOR | 2 | `src/lib/core/modules/GenerationHandler.ts`, `src/lib/providers/openaiChatCompletionsClient.ts` |
| 💡 MINOR | 5 | `src/lib/providers/openaiChatCompletionsClient.ts`, `src/lib/providers/litellm.ts`, `src/lib/context/errorDetection.ts`, `src/lib/proxy/proxyFetch.ts`, `package.json` |
| 💬 SUGGESTION | 0 | — |

### Blocking reason

This PR meets the **“3 or more MAJOR severity issues”** blocking criterion when counting the unresolved MAJOR issues still present in the latest revision:

1. **`GenerationHandler.ts` — `disableToolCallRepair` is used but not declared in `TextGenerationOptions`** (MAJOR). The option is cast into existence at the `generateText` call site, but `TextGenerationOptions` in `src/lib/types/generate.ts` does not include it. TypeScript callers using `generate()` cannot discover or type-check the option, and the cast hides the API inconsistency.

2. **`openaiChatCompletionsClient.ts` — `sanitizeToolName` can return an empty string** (MAJOR). For empty or all-special-character registered tool names, the sanitized wire name is `""`, which is then emitted as an invalid OpenAI-compatible function name. The backend will reject the request.

3. **`openaiChatCompletionsClient.ts` — `sanitizeToolName` output is not guaranteed to satisfy `WIRE_TOOL_NAME_RE`** (MAJOR). Names starting with a digit are returned unchanged, and names longer than 64 chars are not truncated, so invalid wire names can reach the backend.

(The prior round's MAJOR items about `finishReason` object/string mismatch, lockfile downgrades, `mergeAbortSignals` lifetime pin, and `providerOptions` clobber have been addressed or justified in the current revision and are not re-raised.)

### Next steps

- Add `disableToolCallRepair?: boolean` to `TextGenerationOptions` in `src/lib/types/generate.ts` (or remove the cast and gate repair on an existing field).
- Guarantee every wire name in `buildWireToolNameMaps` is non-empty, starts with `[a-zA-Z_]`, and is ≤64 characters before collision handling.
- Add test coverage for the empty-name, digit-start, and >64-char edge cases in `test/continuous-test-suite-litellm-parity.ts`.
- Address the five MINOR notes at your discretion.

Once the two blocking MAJOR items are resolved, I'm happy to re-review.

src/lib/providers/openaiChatCompletionsClient.ts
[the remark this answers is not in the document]
pdogra1299: Verified false positive — `sanitizeToolName` cannot return an empty or wire-invalid name. Its control flow: (1) replace invalid chars with `_` (replacement, not deletion — length preserved), (2) prepend `_` when the first char is not `[a-zA-Z_]`, (3) truncate to 64 **after** the prepend. Empirically: `""` → `"_"`, `"!!!"` → `"___"`, `"123tool"` → `"_123tool"` (the prepend branch fires — it is not returned unchanged), 65+ chars → truncated to 64. Every output satisfies `WIRE_TOOL_NAME_RE` by construction, so the `|| "_tool"` fallback would be dead code. Locked in with a new parity-suite case (1e5f3f30) asserting empty / all-special / leading-digit / >64-char names all map to wire-valid non-empty names — so any future change to `sanitizeToolName` that breaks the invariant fails the suite.

src/lib/core/modules/GenerationHandler.ts
pdogra1299: Confirmed and fixed in 1e5f3f30: `disableToolCallRepair` is now declared on `TextGenerationOptions` (and `GenerateOptions`) mirroring the StreamOptions JSDoc, the cast in `GenerationHandler` is removed, and the CLI loop options schema gained the matching entry (the schema is exhaustive over TextGenerationOptions' simple keys, so the compiler enforced it).

src/lib/providers/openaiChatCompletionsClient.ts
pdogra1299: Verified false positive — `sanitizeToolName` cannot return an empty or wire-invalid name. Its control flow: (1) replace invalid chars with `_` (replacement, not deletion — length preserved), (2) prepend `_` when the first char is not `[a-zA-Z_]`, (3) truncate to 64 **after** the prepend. Empirically: `""` → `"_"`, `"!!!"` → `"___"`, `"123tool"` → `"_123tool"` (the prepend branch fires — it is not returned unchanged), 65+ chars → truncated to 64. Every output satisfies `WIRE_TOOL_NAME_RE` by construction, so the `|| "_tool"` fallback would be dead code. Locked in with a new parity-suite case (1e5f3f30) asserting empty / all-special / leading-digit / >64-char names all map to wire-valid non-empty names — so any future change to `sanitizeToolName` that breaks the invariant fails the suite.

src/lib/providers/openaiChatCompletionsClient.ts
pdogra1299: Verified false positive — `sanitizeToolName` cannot return an empty or wire-invalid name. Its control flow: (1) replace invalid chars with `_` (replacement, not deletion — length preserved), (2) prepend `_` when the first char is not `[a-zA-Z_]`, (3) truncate to 64 **after** the prepend. Empirically: `""` → `"_"`, `"!!!"` → `"___"`, `"123tool"` → `"_123tool"` (the prepend branch fires — it is not returned unchanged), 65+ chars → truncated to 64. Every output satisfies `WIRE_TOOL_NAME_RE` by construction, so the `|| "_tool"` fallback would be dead code. Locked in with a new parity-suite case (1e5f3f30) asserting empty / all-special / leading-digit / >64-char names all map to wire-valid non-empty names — so any future change to `sanitizeToolName` that breaks the invariant fails the suite.

src/lib/providers/litellm.ts
pdogra1299: Deliberate, keeping as-is: `fetchLiteLLMModelLimits` intentionally mirrors its sibling `fetchModelsFromAPI` in the same file (same endpoint family, same 5s bound, same abort→NetworkError conversion — the JSDoc says so explicitly). `createTimeoutController` changes the abort reason shape (TimeoutError vs plain abort) in a never-throws discovery path whose failures are swallowed to debug logs, so switching buys no behavior and risks a subtle classification change. Unifying both `/model`-family fetchers onto the shared utility is good housekeeping for a follow-up that touches them together.

src/lib/context/errorDetection.ts
pdogra1299: Accepted and wired in 1e5f3f30 — `requestedOutputTokens` is now consumed: when the rejected body carried **no** `max_tokens` (server-defaulted output — e.g. the Gemini-2.5-on-LiteLLM maxTokens skip, or any caller relying on backend defaults), `correctBodyAfterContextOverflow` uses the output size the backend's error stated as the previous bound, so those requests get the re-fit retry too instead of propagating. Covered by a new parity test.

src/lib/proxy/proxyFetch.ts
pdogra1299: Fixed in 1e5f3f30 — hoisted to module-scope `TRANSIENT_NETWORK_CODES`.

PR #1223, opened 2026-07-21
fix(proxy): close restartless observability gaps

murdore:
## Summary

- replace the active rolling worker on `.env` changes while preserving the public listener and draining existing streams
- serialize environment, explicit, and crash-recovery replacements; invalidate stale timer callbacks; and harden listener cleanup, supervisor PID validation, and stale status metadata handling
- extend socket-transfer acknowledgement to 30 seconds, retain exact worker exit/transfer failures after recovery, and keep unsafe descriptors non-replayable
- guard parent and worker copies of pre-commit transferred sockets so client `ECONNRESET` cannot crash the supervisor or worker
- preserve pending environment replacements across worker recovery and ignore comment/whitespace-only `.env` rewrites
- persist successful upstream attempts, bound and abort request/body log I/O during shutdown, and preflight immediate SSE errors after OAuth refresh
- retain current-day metadata indexes, report unattributed persisted counters, and expose per-stream/body-artifact data quality in `proxy analyze`

## Validation

- `pnpm exec vitest run test/proxyAnalysis.test.ts test/proxyConfigHotReload.test.ts test/proxyObservabilityFoundation.test.ts test/proxyReliabilityHardening.test.ts test/proxyReplay.test.ts test/proxyRollingWorkerHandoff.test.ts test/proxyUpdaterFallback.test.ts test/proxyUsageStatsPersistence.test.ts` (181 passed)
- `pnpm run test:bugfixes` (234 passed)
- `pnpm run typecheck`
- changed-file ESLint (0 errors)
- `pnpm run build:cli`
- full pre-commit pipeline: check, format, lint, validate/security, package build, browser build, publint
- `pnpm run proxy:performance` (all budgets passed; 0 dropped handoff requests; active stream preserved)
- built CLI `proxy analyze --help` smoke test

## Live safety

No repository test sent model traffic or used the live proxy as a development target.

## Summary by CodeRabbit

* **New Features**
  * Expanded offline proxy log analysis to include debug time windows and body-artifact quality (indexed, present/missing, invalid paths, truncation, write failures).
  * Runtime reload now better detects environment-only changes and reports `environmentChanged`.
  * Proxy `/status` now improves serving-state details and per-account attribution, including an “unattributed” entry.
* **Bug Fixes**
  * Standardized attempt/telemetry completion logging across JSON, streaming, and auth-retry flows.
  * Improved shutdown/reload reliability by coordinating request-log flushing and strengthening log cleanup/retention safeguards.

### DIFF
--- a/src/lib/proxy/proxyAnalysis.ts
+++ b/src/lib/proxy/proxyAnalysis.ts
@@ -1,26 +1,57 @@
 import { createReadStream } from "node:fs";
-import { readdir } from "node:fs/promises";
+import { lstat, readdir, realpath, stat } from "node:fs/promises";
 import { homedir } from "node:os";
 import { createInterface } from "node:readline";
-import { join, resolve } from "node:path";
+import { isAbsolute, join, relative, resolve, sep } from "node:path";
 import type {
   ProxyAnalysisAccount,
   ProxyAnalysisAttemptRecord,
   ProxyAnalysisFinalRequestRecord,
   ProxyAnalysisOptions,
   ProxyAnalysisReport,
+  ProxyAnalysisStreamName,
   ProxyLatencySummary,
 } from "../types/index.js";
 
 const LIFECYCLE_FILE_PATTERN = /^proxy-lifecycle-\d{4}-\d{2}-\d{2}\.jsonl$/;
 const REQUEST_FILE_PATTERN = /^proxy-\d{4}-\d{2}-\d{2}\.jsonl$/;
 const ATTEMPT_FILE_PATTERN = /^proxy-attempts-\d{4}-\d{2}-\d{2}\.jsonl$/;
+const DEBUG_FILE_PATTERN = /^proxy-debug-\d{4}-\d{2}-\d{2}\.jsonl$/;
+const ARTIFACT_STAT_CONCURRENCY = 64;
 const LIFECYCLE_EVENTS = new Set([
   "request_accepted",
   "response_headers",
   "response_first_chunk",
   "request_terminal",
 ]);
+
+function isContainedPath(root: string, candidate: string): boolean {
+  const candidateRelative = relative(root, candidate);
+  return (
+    candidateRelative.length > 0 &&
+    !isAbsolute(candidateRelative) &&
+    candidateRelative !== ".." &&
+    !candidateRelative.startsWith(`..${sep}`)
+  );
+}
+
+async function inspectBodyArtifact(
+  artifactPath: string,
+  canonicalBodiesRoot: string | null,
+): Promise<"invalid" | "missing" | "present"> {
+  if (!canonicalBodiesRoot) {
+    return "missing";
+  }
+  try {
+    const canonicalArtifactPath = await realpath(artifactPath);
+    if (!isContainedPath(canonicalBodiesRoot, canonicalArtifactPath)) {
+      return "invalid";
+    }
+    return (await stat(canonicalArtifactPath)).isFile() ? "present" : "missing";
+  } catch {
+    return "missing";
+  }
+}
 
 function increment(counter: Record<string, number>, key: string): void {
   counter[key] = (counter[key] ?? 0) + 1;
@@ -245,6 +276,7 @@ async function discoverLogFiles(logsDir: string) {
     lifecycleFiles: matching(LIFECYCLE_FILE_PATTERN),
     requestFiles: matching(REQUEST_FILE_PATTERN),
     attemptFiles: matching(ATTEMPT_FILE_PATTERN),
+    debugFiles: matching(DEBUG_FILE_PATTERN),
   };
 }
 
@@ -257,8 +289,32 @@ export async function analyzeProxyLogs(
   const logsDir = resolve(
     options?.logsDir ?? join(homedir(), ".neurolink", "logs"),
   );
-  const { lifecycleFiles, requestFiles, attemptFiles } =
+  const { lifecycleFiles, requestFiles, attemptFiles, debugFiles } =
     await discoverLogFiles(logsDir);
+
+  const observedRanges: Record<
+    ProxyAnalysisStreamName,
+    { from: number | null; to: number | null }
+  > = {
+    lifecycle: { from: null, to: null },
+    requests: { from: null, to: null },
+    attempts: { from: null, to: null },
+    debug: { from: null, to: null },
+  };
+  const observeTimestamp = (
+    stream: ProxyAnalysisStreamName,
+    record: Record<string, unknown>,
+  ): number | null => {
+    const timestamp = Date.parse(String(record.timestamp ?? ""));
+    if (!Number.isFinite(timestamp)) {
+      return null;
+    }
+    const range = observedRanges[stream];
+    range.from =
+      range.from === null ? timestamp : Math.min(range.from, timestamp);
+    range.to = range.to === null ? timestamp : Math.max(range.to, timestamp);
+    return timestamp;
+  };
 
   let linesRead = 0;
   let malformedLines = 0;
@@ -279,8 +335,8 @@ export async function analyzeProxyLogs(
     linesRead += await readJsonLines(
       filePath,
       (record) => {
-        const timestamp = Date.parse(String(record.timestamp ?? ""));
-        if (!Number.isFinite(timestamp) || timestamp < sinceMs) {
+        const timestamp = observeTimestamp("lifecycle", record);
+        if (timestamp === null || timestamp < sinceMs) {
           return;
         }
         const event = stringValue(record.event);
@@ -369,8 +425,8 @@ export async function analyzeProxyLogs(
     linesRead += await readJsonLines(
       filePath,
       (record) => {
-        const timestamp = Date.parse(String(record.timestamp ?? ""));
-        if (!Number.isFinite(timestamp) || timestamp < sinceMs) {
+        const timestamp = observeTimestamp("attempts", record);
+        if (timestamp === null || timestamp < sinceMs) {
           return;
         }
         const requestId = stringValue(record.requestId);
@@ -437,8 +493,8 @@ export async function analyzeProxyLogs(
     linesRead += await readJsonLines(
       filePath,
       (record) => {
-        const timestamp = Date.parse(String(record.timestamp ?? ""));
-        if (!Number.isFinite(timestamp) || timestamp < sinceMs) {
+        const timestamp = observeTimestamp("requests", record);
+        if (timestamp === null || timestamp < sinceMs) {
           return;
         }
         const requestId = stringValue(record.requestId);
@@ -471,6 +527,81 @@ export async function analyzeProxyLogs(
     );
   }
 
+  let capturesIndexed = 0;
+  let truncatedCaptures = 0;
+  let writeFailures = 0;
+  let invalidPaths = 0;
+  const referencedArtifacts = new Set<string>();
+  const bodiesRoot = resolve(logsDir, "bodies");
+  for (const filePath of debugFiles) {
+    linesRead += await readJsonLines(
+      filePath,
+      (record) => {
+        const timestamp = observeTimestamp("debug", record);
+        if (timestamp === null || timestamp < sinceMs) {
+          return;
+        }
+        if (record.type !== "body_capture") {
+          return;
+        }
+        capturesIndexed += 1;
+        truncatedCaptures += record.bodyTruncated === true ? 1 : 0;
+        writeFailures += record.bodyWriteFailed === true ? 1 : 0;
+        const bodyPath = stringValue(record.bodyPath);
+        if (!bodyPath) {
+          return;
+        }
+        if (bodyPath.includes("\0") || bodyPath.split(/[\\/]/).includes("..")) {
+          invalidPaths += 1;
+          return;
+        }
+        const resolvedBodyPath = resolve(logsDir, bodyPath);
+        if (!isContainedPath(bodiesRoot, resolvedBodyPath)) {
+          invalidPaths += 1;
+          return;
+        }
+        referencedArtifacts.add(resolvedBodyPath);
+      },
+      () => {
+        malformedLines += 1;
+      },
+    );
+  }
+  const artifactPaths = [...referencedArtifacts];
+  let canonicalBodiesRoot: string | null = null;
+  let bodiesRootUnsafe = false;
+  try {
+    const bodiesRootStat = await lstat(bodiesRoot);
+    if (bodiesRootStat.isDirectory() && !bodiesRootStat.isSymbolicLink()) {
+      canonicalBodiesRoot = await realpath(bodiesRoot);
+    } else {
+      bodiesRootUnsafe = true;
+    }
+  } catch {
+    // A missing body directory means every lexically valid reference is absent.
+  }
+  let artifactsPresent = 0;
+  let artifactsMissing = 0;
+  for (
+    let offset = 0;
+    offset < artifactPaths.length;
+    offset += ARTIFACT_STAT_CONCURRENCY
+  ) {
+    const presence = await Promise.all(
+      artifactPaths
+        .slice(offset, offset + ARTIFACT_STAT_CONCURRENCY)
+        .map((artifactPath) =>
+          bodiesRootUnsafe
+            ? Promise.resolve<"invalid" | "missing" | "present">("invalid")
+            : inspectBodyArtifact(artifactPath, canonicalBodiesRoot),
+        ),
+    );
+    artifactsPresent += presence.filter((value) => value === "present").length;
+    artifactsMissing += presence.filter((value) => value === "missing").length;
+    invalidPaths += presence.filter((value) => value === "invalid").length;
+  }
+  const artifactsReferenced = artifactsPresent + artifactsMissing;
+
   const finalSummary = summarizeFinalRequests(
     finalRequests,
     terminalStreamErrors,
@@ -486,11 +617,13 @@ export async function analyzeProxyLogs(
       lifecycle: lifecycleFiles.length,
       requests: requestFiles.length,
       attempts: attemptFiles.length,
+      debug: debugFiles.length,
     },
     coverage: {
-      lifecycle: lifecycleFiles.length > 0,
-      finalRequests: requestFiles.length > 0,
-      attempts: attemptFiles.length > 0,
+      lifecycle:
+        accepted.size + headers.size + firstChunks.size + terminal.size > 0,
+      finalRequests: finalRequests.size > 0 || terminalStreamErrors.size > 0,
+      attempts: totalAttempts > 0,
       attemptLatency: attemptLatency.length > 0,
       cacheUsage: finalSummary.cache.requestsWithUsage > 0,
     },
@@ -500,6 +633,28 @@ export async function analyzeProxyLogs(
       unsupportedLifecycleLines,
       lifecycleSequenceGaps,
       lifecycleSequenceDuplicates,
+      streams: Object.fromEntries(
+        Object.entries(observedRanges).map(([stream, range]) => [
+          stream,
+          {
+            observedFrom:
+              range.from === null ? null : new Date(range.from).toISOString(),
+            observedTo:
+              range.to === null ? null : new Date(range.to).toISOString(),
+            startsAtOrBeforeRequestedWindow:
+              range.from !== null && range.from <= sinceMs,
+          },
+        ]),
+      ) as ProxyAnalysisReport["dataQuality"]["streams"],
+      bodyArtifacts: {
+        capturesIndexed,
+        artifactsReferenced,
+        artifactsPresent,
+        artifactsMissing,
+        invalidPaths,
+        writeFailures,
+        truncatedCaptures,
+      },
     },
     lifecycle: {
       accepted: accepted.size,
--- a/src/lib/proxy/requestLogger.ts
+++ b/src/lib/proxy/requestLogger.ts
@@ -18,7 +18,7 @@ import {
   statSync,
   unlinkSync,
 } from "fs";
-import { appendFile, writeFile } from "fs/promises";
+import { writeFile } from "fs/promises";
 import { createHash } from "crypto";
 import { promisify } from "util";
 import { gzip as gzipCallback } from "zlib";
@@ -33,9 +33,56 @@ import { OtelBridge } from "../observability/otelBridge.js";
 import { SeverityNumber } from "@opentelemetry/api-logs";
 import type { LoggerProvider } from "@opentelemetry/sdk-logs";
 import { configureProxyLifecycleLogger } from "./proxyLifecycle.js";
+import { withTimeout } from "../utils/async/withTimeout.js";
 
 let logDir: string | null = null;
 let logEnabled = false;
+const pendingLogOperations = new Set<Promise<unknown>>();
+const REQUEST_LOG_IO_TIMEOUT_MS = 5_000;
+
+function trackLogOperation<T>(operation: Promise<T>): Promise<T> {
+  pendingLogOperations.add(operation);
+  void operation.then(
+    () => pendingLogOperations.delete(operation),
+    () => pendingLogOperations.delete(operation),
+  );
+  return operation;
+}
+
+/** Wait, up to a bounded deadline, for admitted request/body writes to settle. */
+export async function flushRequestLogs(
+  timeoutMs: number = REQUEST_LOG_IO_TIMEOUT_MS,
+): Promise<void> {
+  const deadline = Date.now() + Math.max(1, timeoutMs);
+  while (pendingLogOperations.size > 0) {
+    const admitted = [...pendingLogOperations];
+    const remainingMs = Math.max(1, deadline - Date.now());
+    try {
+      await withTimeout(
+        Promise.allSettled(admitted),
+        remainingMs,
+        `Timed out flushing ${admitted.length} proxy request log operation(s)`,
+      );
+    } catch (error) {
+      for (const operation of admitted) {
+        pendingLogOperations.delete(operation);
+      }
+      throw error;
+    }
+    if (Date.now() >= deadline && pendingLogOperations.size > 0) {
+      const remaining = pendingLogOperations.size;
+      throw new Error(
+        `Timed out flushing ${remaining} proxy request log operation(s)`,
+      );
+    }
+  }
+}
+
+/** @internal Test-only hook for exercising shutdown behavior without real I/O. */
+export const __requestLoggerTestHooks = {
+  pendingOperationCount: () => pendingLogOperations.size,
+  trackLogOperation,
+};
 
 /**
  * Lazily-resolved LoggerProvider from OTel instrumentation.
@@ -124,7 +171,13 @@ export async function logRequest(entry: RequestLogEntry): Promise<void> {
   const line = JSON.stringify(entry) + "\n";
 
   try {
-    await appendFile(logFile, line, { mode: 0o600 });
+    await trackLogOperation(
+      writeFile(logFile, line, {
+        mode: 0o600,
+        flag: "a",
+        signal: AbortSignal.timeout(REQUEST_LOG_IO_TIMEOUT_MS),
+      }),
+    );
   } catch {
     // Non-fatal — don't crash proxy for logging failures
   }
@@ -161,7 +214,13 @@ export async function logRequestAttempt(
   const line = JSON.stringify(entry) + "\n";
 
   try {
-    await appendFile(logFile, line, { mode: 0o600 });
+    await trackLogOperation(
+      writeFile(logFile, line, {
+        mode: 0o600,
+        flag: "a",
+        signal: AbortSignal.timeout(REQUEST_LOG_IO_TIMEOUT_MS),
+      }),
+    );
   } catch {
     // Non-fatal — don't crash proxy for logging failures
   }
@@ -452,7 +511,7 @@ function collectManagedLogFiles(rootDir: string): ManagedLogFile[] {
 
       const isTopLevelProxyLog =
         directory === rootDir &&
-        /^proxy(?:-attempts|-debug)?-.*\.jsonl$/.test(entry.name);
+        /^proxy(?:-attempts|-debug|-lifecycle)?-.*\.jsonl$/.test(entry.name);
       const isBodyArtifact =
         entry.name.endsWith(".json.gz") &&
         entryPath.includes(`${join(rootDir, "bodies")}`);
@@ -540,7 +599,10 @@ async function writeBodyArtifact(
     metadata: entry.metadata,
   });
   const compressed = await gzip(payload);
-  await writeFile(bodyPath, compressed, { mode: 0o600 });
+  await writeFile(bodyPath, compressed, {
+    mode: 0o600,
+    signal: AbortSignal.timeout(REQUEST_LOG_IO_TIMEOUT_MS),
+  });
 
   return {
     bodyPath,
@@ -641,11 +703,13 @@ export async function logBodyCapture(
 
   let stored: StoredBodyArtifact;
   try {
-    stored = await writeBodyArtifact(
-      entry,
-      redactedHeaders,
-      preparedBody.value,
-      preparedBody.truncated,
+    stored = await trackLogOperation(
+      writeBodyArtifact(
+        entry,
+        redactedHeaders,
+        preparedBody.value,
+        preparedBody.truncated,
+      ),
     );
   } catch (writeError) {
     logger.warn(
@@ -656,6 +720,7 @@ export async function logBodyCapture(
       redactedBody: preparedBody.value,
       redactedBodyBytes: preparedBody.bytes,
       bodyTruncated: preparedBody.truncated,
+      bodyWriteFailed: true,
     };
   }
 
@@ -681,6 +746,7 @@ export async function logBodyCapture(
     redactedBodyBytes: stored.redactedBodyBytes ?? preparedBody.bytes,
     storedFileBytes: stored.storedFileBytes,
     bodyTruncated: stored.bodyTruncated ?? preparedBody.truncated,
+    bodyWriteFailed: stored.bodyWriteFailed,
     metadata: entry.metadata,
   };
 
@@ -690,9 +756,13 @@ export async function logBodyCapture(
   }
 
   try {
-    await appendFile(logFile, JSON.stringify(indexEntry) + "\n", {
-      mode: 0o600,
-    });
+    await trackLogOperation(
+      writeFile(logFile, JSON.stringify(indexEntry) + "\n", {
+        mode: 0o600,
+        flag: "a",
+        signal: AbortSignal.timeout(REQUEST_LOG_IO_TIMEOUT_MS),
+      }),
+    );
   } catch {
     // Non-fatal
   }
@@ -797,9 +867,13 @@ export async function logStreamError(entry: {
   }
 
   try {
-    await appendFile(logFile, JSON.stringify(logEntry) + "\n", {
-      mode: 0o600,
-    });
+    await trackLogOperation(
+      writeFile(logFile, JSON.stringify(logEntry) + "\n", {
+        mode: 0o600,
+        flag: "a",
+        signal: AbortSignal.timeout(REQUEST_LOG_IO_TIMEOUT_MS),
+      }),
+    );
   } catch {
     // Non-fatal — don't crash proxy for logging failures
   }
@@ -824,6 +898,14 @@ export function cleanupLogs(
     const files = collectManagedLogFiles(activeLogDir).sort(
       (a, b) => a.mtime - b.mtime,
     ); // oldest first
+    const currentDate = new Date().toISOString().split("T")[0];
+    const currentMetadataLogs = new Set(
+      ["proxy", "proxy-attempts", "proxy-debug", "proxy-lifecycle"].map(
+        (prefix) => join(activeLogDir, `${prefix}-${currentDate}.jsonl`),
+      ),
+    );
+    const canDelete = (file: ManagedLogFile) =>
+      !currentMetadataLogs.has(file.path);
 
     const cutoff = Date.now() - maxAgeDays * 24 * 60 * 60 * 1000;
     let deletedCount = 0;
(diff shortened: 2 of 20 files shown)


src/lib/proxy/requestLogger.ts
```diff
@@ -824,6 +847,14 @@ export function cleanupLogs(
...
       (a, b) => a.mtime - b.mtime,
     ); // oldest first
+    const currentDate = new Date().toISOString().split("T")[0];
+    const currentMetadataLogs = new Set(
+      ["proxy", "proxy-attempts", "proxy-debug", "proxy-lifecycle"].map(
+        (prefix) => join(activeLogDir, `${prefix}-${currentDate}.jsonl`),
+      ),
+    );
+    const canDelete = (file: ManagedLogFile) =>
+      !currentMetadataLogs.has(file.path);
```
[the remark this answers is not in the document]
murdore: Fixed in 3b536e12. The managed top-level matcher now includes proxy-lifecycle JSONL files. The retention test now creates an older lifecycle index and verifies it is evicted while the current-day lifecycle index remains protected.

src/lib/types/proxy.ts
```diff
@@ -2279,8 +2315,14 @@ export type StatusStats = {
...
-    status?: "active" | "cooling" | "disabled" | "excluded" | "removed";
+    cooling?: boolean;
+    status?:
+      | "active"
+      | "cooling"
+      | "disabled"
+      | "excluded"
+      | "removed"
+      | "unattributed";
   }[];
```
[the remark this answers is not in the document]
murdore: Fixed in 3b536e12. StatusStats.accounts[].cooling remains a required boolean, and the synthetic unattributed row explicitly reports cooling: false. The status test asserts that contract.

murdore:
Addressed both CodeRabbit nitpicks in a34bd18d: removed the unreachable auth-retry streaming branch after SSE preflight unification, and reused the canonical StatusStats type for live status parsing. The focused and full proxy tests, typecheck, and complete pre-commit pipeline pass.

murdore:
@coderabbitai review

src/lib/server/routes/claudeProxyRoutes.ts
```diff
@@ -4016,66 +4026,9 @@ async function handleAnthropicSuccessfulRetryResponse(args: {
     });
   }
```
Tara-ag, line 4028: ⚠️ **MAJOR: Retry attempt observability gap — `logAttempt` here omits `attemptDurationMs`.**

The streaming retry path through `handleAnthropicSuccessfulResponse` logs `attemptDurationMs` after preflight, but this non-streaming retry helper logs only the status. Successful OAuth retry attempts will therefore be missing per-attempt timing in `proxy-attempts-*.jsonl`, breaking the consistency the PR aims to establish across "JSON, streaming, and auth-retry flows".

Suggested fix:
```ts
logAttempt(retryResp.status, undefined, undefined, {
  attemptDurationMs: Date.now() - fetchStartMs,
});
```

Also, since this helper is now only reachable for non-streaming retries, consider renaming it to `handleAnthropicSuccessfulNonStreamRetryResponse` or asserting `!body.stream` to prevent future streaming reuse that would bypass `preflightAnthropicStream`.

src/lib/proxy/rollingProxyServer.ts
```diff
@@ -77,13 +81,55 @@ export async function startRollingProxyServer(
...
+    requestedReplacementPending = true;
+    if (requestedReplacementTimer || requestedReplacementRunning) {
+      return;
+    }
+    requestedReplacementTimer = setTimeout(() => {
+      requestedReplacementTimer = undefined;
+      requestedReplacementPending = false;
+      if (closing || !supervisor.snapshot().active) {
+        return;
+      }
```
Tara-ag, line 123: ⚠️ **MAJOR: Replacement-running guard is set too late, leaving a race window.**

`requestedReplacementRunning` is checked at line 88 but only set to `true` here, inside the async `setTimeout` callback. Between the timer firing and this assignment, `replace()` can observe `requestedReplacementRunning === false`, skip awaiting anything, and invoke `supervisor.replace()` concurrently. The supervisor may serialize the underlying spawns, but `requestedReplacementPromise` will only track one of them, and the explicit caller won't wait for the scheduled one.

Suggested fix: set `requestedReplacementRunning = true` before scheduling the timer (or use a single mutex/promise that both paths share). At minimum, `replace()` should await `requestedReplacementPromise` whenever it is non-null, not just clear the timer.

src/lib/proxy/requestLogger.ts
Tara-ag, line 1: ⚠️ **MAJOR: `flushRequestLogs` can hang indefinitely if a tracked operation never settles.**

`trackLogOperation` registers the operation and removes it via `.then`/`.catch` on the same promise. If `appendFile`/`writeFile` hangs (stuck NFS, broken pipe, etc.), the promise stays in `pendingLogOperations` forever and the `while` loop keeps awaiting `Promise.allSettled` without progress. During shutdown this is wrapped in `withTimeout`, but the timeout only rejects the wrapper — the underlying hung promise remains, preventing clean process exit.

Suggested fix: either use `AbortController`-aware file writes, add a per-operation timeout, or cap the flush loop with a deadline so a single stuck I/O cannot block shutdown.

src/lib/proxy/rollingProxyServer.ts
```diff
@@ -77,13 +81,55 @@ export async function startRollingProxyServer(
       scheduleRecovery();
     }
   };
+  const scheduleRequestedReplacement = (): void => {
+    if (closing) {
+      return;
+    }
+    requestedReplacementPending = true;
```
Tara-ag, line 111: 💡 **MINOR: `requestedReplacementTimer` is cleared in `replace()` but the timer callback may already be in the event-queue.**

`clearTimeout` removes a pending timer, but if the callback has already been queued by the event loop it will still run. The callback re-checks `requestedReplacementTimer` (which is now undefined) and proceeds to call `supervisor.replace()`. Meanwhile `replace()` itself is also calling `supervisor.replace()`. The supervisor serializes via `this.replacement`, but the rolling server's `requestedReplacementRunning`/`requestedReplacementPromise` state may end up tracking only one of the two concurrent logical replacements, and the explicit caller won't await the scheduled one.

Suggested fix: use a single replacement mutex/promise shared by both paths, or set an atomic "replacement in flight" flag before the timer callback yields.

src/lib/proxy/proxyAnalysis.ts
```diff
@@ -471,6 +499,65 @@ export async function analyzeProxyLogs(
     );
```
Tara-ag, line 527: 💡 **MINOR: Artifact path traversal check is correct but could be more robust.**

`resolvedBodyPath.startsWith(bodiesPrefix)` prevents escaping `logsDir/bodies/`, which is good. However, `resolve(logsDir, bodyPath)` followed by `startsWith` can be bypassed if `bodyPath` contains a null byte or other unusual sequences on some platforms, and it relies on `sep` being the platform separator. Consider normalizing the path and using `path.relative` + checking for `..` components, or documenting that this is intentionally a best-effort sandbox since the analyzer runs offline with operator-supplied logs.

Also, `writeFailures` is incremented when `bodyPath` is missing, but the index entry may legitimately omit `bodyPath` when the body was not persisted (e.g., redaction-only capture). The naming "writeFailures" implies an I/O failure, which may be misleading for captures that intentionally have no artifact.

src/lib/proxy/runtimeConfig.ts
```diff
@@ -526,7 +537,10 @@ export class ProxyRuntimeConfigStore {
       );
```
Tara-ag, line 546: 💡 **MINOR: `environmentChanged` is only reported when `configHash` also differs, but the PR summary says it should fire on `.env`-only changes.**

The condition at line 537 is:
```ts
if (candidate.snapshot.configHash === this.currentSnapshot.configHash &&
    candidate.envFileHash === this.currentEnvFileHash)
```

If only the env file changes, `configHash` will likely stay the same (since routing/strategy/etc. don't change), but `envFileHash` will differ, so this branch is not taken and we fall through to the "changed" path. That's fine — `environmentChanged` is then computed and included. However, if the env file changes AND the config file changes in the same reload, `environmentChanged` is still set correctly. No bug here, but the logic could be clearer: compute `environmentChanged` before the early-return so the `changed:false` result can also report it. Not blocking.

Tara-ag (changes requested):
## Review summary

Reviewed the 17 changed files in this proxy observability/lifecycle PR. The changes are well-scoped and the test coverage is thorough, but I found **3 MAJOR** issues that should be addressed before merge, plus **3 MINOR** suggestions.

### Blocking issues (MAJOR)

1. **`src/lib/server/routes/claudeProxyRoutes.ts`** — `handleAnthropicSuccessfulRetryResponse` logs `logAttempt(retryResp.status)` without `attemptDurationMs`, while the streaming retry path and main success paths do include it. This leaves a real observability gap in the "JSON, streaming, and auth-retry flows" consistency the PR targets.

2. **`src/lib/proxy/rollingProxyServer.ts`** — The `requestedReplacementRunning` guard is set inside the async `setTimeout` callback, not before it. Between the timer firing and the flag being set, a concurrent `replace()` call can observe `requestedReplacementRunning === false`, skip awaiting, and invoke `supervisor.replace()` concurrently. The supervisor serializes the underlying spawn, but the rolling-server state (`requestedReplacementPromise`, `requestedReplacementRunning`) ends up tracking the wrong replacement.

3. **`src/lib/proxy/requestLogger.ts`** — `flushRequestLogs()` can hang indefinitely if a tracked `appendFile`/`writeFile` promise never settles (stuck NFS, broken pipe, etc.). The `while` loop keeps awaiting `Promise.allSettled([...pendingLogOperations])` without progress. The shutdown wrapper times out, but the underlying hung promise remains and can prevent clean process exit.

### Non-blocking suggestions (MINOR)

- `rollingProxyServer.ts`: `clearTimeout` does not remove a timer callback that has already been queued by the event loop, creating another race window between scheduled and explicit replacements.
- `proxyAnalysis.ts`: the artifact path-traversal guard is reasonable but could be more robust; `writeFailures` is incremented when `bodyPath` is absent, which may represent an intentional redaction-only capture rather than a write failure.
- `runtimeConfig.ts`: `environmentChanged` is only computed after the early-return branch; consider computing it earlier so a `changed:false` reload result can still report it.

### Existing review comments

The two prior Copilot comments (lifecycle log matcher, `StatusStats.accounts[].cooling` required) appear resolved in the latest commit and are not re-raised.

### Next steps

Please address the three MAJOR issues above. The replacement-serialization race and the flush hang are the highest-risk for production reliability.

murdore:
Fresh read-only live evidence exposed one additional handoff gap: the active v10.3.0 worker lost four socket transfers, the supervisor recovered to generation 2, but the 5-second acknowledgement threshold killed the worker and successful recovery cleared the root-cause field. Commit 5b8c23bb raises the production acknowledgement deadline to 30 seconds, logs exact transfer/exit details (including code/signal), retains lastFailure across successful recovery, and keeps ambiguous transferred sockets non-replayable. Revalidated: 175 proxy tests, 234 bugfix tests, typecheck, ESLint, CLI build, full pre-commit, and all performance budgets.

src/lib/server/routes/claudeProxyRoutes.ts
murdore: Fixed in b6b404f5. The non-stream OAuth retry now passes  explicitly, the retry wrapper preserves an explicit duration while still attaching the retry attempt number, and the helper is renamed  to make its scope unambiguous. Added a regression test asserting the successful retry emits attempt 3 with a duration.

src/lib/proxy/rollingProxyServer.ts
murdore: Fixed in b6b404f5. Pending timer callbacks carry a monotonic schedule token; explicit replacement and close invalidate that token before clearing the timer. Even if a stale callback is already queued, it exits without enqueuing work. All replacement sources also use the shared serializer.

src/lib/proxy/runtimeConfig.ts
murdore: Verified as non-bug. The unchanged branch is reached only when both  and  match. An env-only change necessarily has a different , takes the changed path, and returns ; the existing hot-reload regression test covers that behavior. A  result intentionally means neither source changed.

src/lib/server/routes/claudeProxyRoutes.ts
```diff
@@ -4016,66 +4026,9 @@ async function handleAnthropicSuccessfulRetryResponse(args: {
...
-      finalBodyStr,
-      upstreamSpan,
-      logProxyBody,
-      logFinalRequest,
-    });
-  }
-
   const retryRespHeaders = Object.fromEntries([...retryResp.headers.entries()]);
   const retryText = await retryResp.text();
+  logAttempt(retryResp.status);
```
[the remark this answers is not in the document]
murdore: Fixed in b6b404f5 exactly as described: explicit retry duration, nullish preservation in , retained retry attempt number, and a non-stream-specific helper name. The focused reliability suite passes 50/50 tests, including the new successful non-stream OAuth retry assertion.

src/lib/server/routes/claudeProxyRoutes.ts
murdore: Correction with complete details: fixed in b6b404f5. The non-stream OAuth retry now passes attemptDurationMs explicitly, the retry wrapper preserves an explicit duration while still attaching the retry attempt number, and the helper is renamed handleAnthropicSuccessfulNonStreamRetryResponse to make its scope unambiguous. A regression test asserts the successful retry emits attempt 3 with a duration.

src/lib/proxy/rollingProxyServer.ts
murdore: Fixed in b6b404f5. Environment, explicit, and autonomous recovery replacements now share one serialized replacement queue. The queue tail is registered before supervisor.replace runs, so explicit callers cannot bypass an in-flight environment replacement. A regression test holds generation 2 unready, starts an explicit version replacement, and proves generation 3 is not spawned until generation 2 completes.

src/lib/proxy/requestLogger.ts
murdore: Fixed in b6b404f5. flushRequestLogs now has a 5 second deadline, drops timed-out tracked references, and throws a diagnostic count. Request, attempt, body, index, and stream-error writes use abortable writeFile calls; flag a preserves append behavior, so the underlying filesystem work is cancelled too. A never-settling-operation regression test covers the deadline.

src/lib/proxy/rollingProxyServer.ts
murdore: Complete fix in b6b404f5: pending timer callbacks carry a monotonic schedule token. Explicit replacement and close invalidate that token before clearing the timer, so a stale callback already queued by the event loop exits without enqueuing work. All replacement sources also use the shared serializer.

src/lib/proxy/proxyAnalysis.ts
murdore: Fixed in b6b404f5. Artifact containment now uses path.relative plus absolute and parent checks instead of a string prefix. The logger emits an explicit bodyWriteFailed marker only when artifact persistence throws, and the analyzer counts only that marker; a legitimately absent bodyPath is no longer mislabeled as an I/O failure.

src/lib/proxy/runtimeConfig.ts
murdore: Verified as a non-bug. The unchanged branch is reached only when both configHash and envFileHash match. An env-only change necessarily has a different envFileHash, takes the changed path, and returns environmentChanged true; the existing hot-reload regression test covers that behavior. A changed false result intentionally means neither source changed.

src/lib/server/routes/claudeProxyRoutes.ts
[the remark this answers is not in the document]
murdore: Correction with complete details: fixed in b6b404f5 exactly as described. The helper passes an explicit retry duration, retryLogAttempt preserves that duration with a nullish fallback, the retry attempt number is retained, and the helper now has a non-stream-specific name. The focused reliability suite passes 50 of 50 tests, including the new successful non-stream OAuth retry assertion.

murdore:
Live evidence captured after the released 10.3.0 worker rotated:

- the long-lived supervisor stayed on PID 14230 while the worker generation advanced 2 -> 3 -> 4
- status recorded 26 rejected sockets and 26 failed transfers
- generation 3 failed immediately, before serving lifecycle traffic
- stderr proves the immediate worker failure was an unhandled Node Socket error: `Error: read ECONNRESET`, emitted while a transferred socket was waiting for the child-side commit
- generation 2 likely entered recovery after the released 5-second transfer-ack window was exceeded under load, but 10.3.0 clears the root failure after recovery, so that first trigger cannot be stated as proven

Commit 2412be2f guards both parent and worker copies of pre-commit transferred sockets. A client reset now removes/destroys the pending descriptor instead of terminating either process. Coverage includes a deterministic pending-socket `ECONNRESET` unit test and a real child-process TCP RST test proving the same worker PID remains active and serves `/health` with zero transfer failures.

The same commit also preserves an environment-triggered replacement if the active worker exits during the debounce window, and hashes parsed/sorted env values so comment- or whitespace-only edits do not roll a worker.

Validation: proxy suites 181/181, bugfix suite 234/234, typecheck, ESLint, CLI/package/browser builds, publint, security validation, and all proxy performance budgets. No live model request was sent and no additional live restart was performed during this investigation.

src/lib/proxy/rollingProxyServer.ts
```diff
@@ -77,13 +103,53 @@ export async function startRollingProxyServer(
...
+          `[proxy-supervisor] same-version worker replacement complete version=${replacementVersion} reason=environment`,
+        );
+      }).catch((error) => {
+        options.log?.(
+          `[proxy-supervisor] same-version worker replacement failed version=${replacementVersion} reason=environment: ${error instanceof Error ? error.message : String(error)}`,
+        );
+      });
+    }, 50);
+    requestedReplacementTimer.unref?.();
+  }
```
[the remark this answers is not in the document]
murdore: Fixed in 2412be2f. The pending flag is now cleared only after the active-worker guard, so a worker exit during the debounce window preserves the environment request. The recovery replacement completes through the shared queue, whose completion hook re-schedules the pending same-version replacement. A regression test makes generation 1 request the roll and exit before the debounce fires, then proves recovery creates generation 2 and the preserved request rolls to generation 3.

murdore:
@coderabbitai review

Tara-ag (approved):
## Review summary

Reviewed the proxy observability/lifecycle changes in this PR. The previously raised blocking issues have been addressed in the latest commits:

- **Non-stream OAuth retry duration**: `handleAnthropicSuccessfulNonStreamRetryResponse` now logs `attemptDurationMs`, and `retryLogAttempt` preserves an explicitly passed duration.
- **Rolling replacement serialization**: `rollingProxyServer.ts` uses a shared `queueReplacement` serializer and defers clearing `requestedReplacementPending` until after the active-worker guard.
- **Request-log flush deadline**: `flushRequestLogs` now enforces a bounded deadline, drops timed-out tracked operations, and uses abortable `writeFile` calls.
- **Artifact path containment**: `proxyAnalysis.ts` uses `path.relative` plus absolute/parent checks instead of a simple prefix match.
- **Env-hash normalization**: `runtimeConfig.ts` hashes parsed, sorted env values so comment/whitespace edits do not trigger worker replacement.

No new blocking issues were found in this run. The test coverage is thorough (rolling handoff, observability foundation, reliability hardening, config hot-reload, and analysis suites), and the changes stay within the proxy subsystem without affecting the public SDK API.

**Files reviewed:** 20 changed files (focused on `src/cli/commands/proxy.ts`, `src/lib/proxy/*`, `src/lib/server/routes/claudeProxyRoutes.ts`, `src/lib/types/proxy.ts`, and test updates).

Approving.

src/lib/proxy/rollingWorkerProcess.ts
```diff
@@ -86,6 +86,20 @@ export function spawnProxySocketWorker(
     }
```
Tara-ag, line 86: 🔒 **CRITICAL: Cancelling a socket on the worker after a local error/timeout can race with a successful transfer, causing the worker to destroy a live connection.**

`settleSocket` sends `proxy-worker:socket-cancel` to the child whenever `error` is truthy, including on supervisor-side timeout or send failure. If the cancel message arrives after the worker has already emitted `socket-accepted` and the supervisor has sent `socket-commit`, the worker may destroy a socket that is now owned by the child. This is a cross-process correctness bug that can drop an active client connection after it was successfully handed off.

Suggested fix: only send `socket-cancel` when the transfer has not yet been accepted (`!pending.accepted`). Once `accepted` is true, the supervisor should rely on `socket-commit` success/failure to settle the transfer and should not ask the worker to cancel.

```ts
const settleSocket = (socketId: string, error?: Error): void => {
  const pending = pendingSockets.get(socketId);
  if (!pending) {
    return;
  }
  pendingSockets.delete(socketId);
  clearTimeout(pending.timeout);
  if (error && child.connected && !pending.accepted) {
    try {
      child.send(
        { type: "proxy-worker:socket-cancel", generation: options.generation, socketId },
        () => undefined,
      );
    } catch {
      // The supervisor will quarantine the worker after the failed transfer.
    }
  }
  if (!error) {
    pending.socket.destroy();
  }
  pending.callback(error);
};
```

src/lib/proxy/socketWorkerRuntime.ts
```diff
@@ -220,6 +220,24 @@ export function attachSocketWorkerProcess(
           return;
```
Tara-ag, line 241: ⚠️ **MAJOR: Worker-side pending socket `error`/`close` handlers can leak if the socket is committed before they fire, because `once` listeners are not removed on commit.**

When a socket is committed via `proxy-worker:socket-commit`, the code calls `pendingSockets.delete(socketId)` and `runtime.acceptSocket(socket)`. The `error` and `close` handlers registered earlier are still attached to the socket. If the socket later errors or closes, those handlers will run, see that `pendingSockets.get(socketId) !== socket` (because it was deleted), and return early — but the listeners themselves remain until the socket is destroyed. This is a minor leak per socket, but more importantly it means the worker's `drainWhenPendingSettled()` may be called spuriously after commit if a `close` event fires, though the guard prevents incorrect behavior.

Suggested fix: remove the pending-only handlers when committing the socket:

```ts
} else if (
  message.type === "proxy-worker:socket-commit" ||
  message.type === "proxy-worker:socket-cancel"
) {
  const socket = pendingSockets.get(message.socketId);
  if (!socket) {
    return;
  }
  pendingSockets.delete(message.socketId);
  socket.removeAllListeners("error");
  socket.removeAllListeners("close");
  if (message.type === "proxy-worker:socket-commit") {
    runtime.acceptSocket(socket);
  } else {
    socket.destroy();
  }
  drainWhenPendingSettled();
}
```

Use `removeAllListeners` cautiously only if these are the only `error`/`close` listeners the worker has installed; otherwise track the handler references and remove them specifically.

src/lib/proxy/rollingProxyServer.ts
```diff
@@ -77,16 +103,60 @@ export async function startRollingProxyServer(
       scheduleRecovery();
```
Tara-ag, line 103: ⚠️ **MAJOR: Recovery replacement does not reset `recoveryFailures` on success, and the explicit `replace()` path can starve recovery by cancelling the timer without re-arming it.**

`scheduleRecovery` increments `recoveryFailures` on failure and uses it to back off, but `queueReplacement`'s success handler sets `recoveryFailures = 0` only in the recovery timer callback. If an explicit `replace()` succeeds while recovery is pending, the recovery timer is cancelled and never re-armed, so `recoveryFailures` is not reset. More importantly, if `supervisor.start()` fails initially, `recoveryFailures = 1` is set, but a later successful explicit replacement does not clear it, leaving an unnecessary backoff on the next failure.

Suggested fix: reset `recoveryFailures = 0` at the start of any successful replacement completion, not just inside the recovery timer callback:

```ts
const queueReplacement = <T>(operation: () => Promise<T>): Promise<T> => {
  const predecessor = replacementQueueTail;
  const result = (predecessor ?? Promise.resolve()).then(operation);
  const completion = result.then(
    () => undefined,
    () => undefined,
  );
  replacementQueueTail = completion;
  void completion.finally(() => {
    if (replacementQueueTail !== completion) {
      return;
    }
    replacementQueueTail = null;
    if (requestedReplacementPending) {
      scheduleRequestedReplacement();
    }
  });
  return result;
};
```

Then reset `recoveryFailures` in the `replace()` success path and in the recovery success path:

```ts
replace: async (expectedVersion) => {
  // ... existing version/timer logic ...
  try {
    const snapshot = await queueReplacement(() =>
      supervisor.replace(expectedVersion),
    );
    recoveryFailures = 0;
    return snapshot;
  } catch (error) {
    // ...
  }
},
```

src/lib/server/routes/claudeProxyRoutes.ts
```diff
@@ -4293,23 +4249,55 @@ async function handleAnthropicAuthRetry(args: {
...
+            };
+        if ("retryNextAccount" in successResult) {
+          const failure = successResult.failure;
+          return {
+            continueLoop: true,
+            lastError: failure?.message ?? currentLastError,
+            authFailureMessage: currentAuthFailureMessage,
+            sawRateLimit: currentSawRateLimit || Boolean(failure?.rateLimit),
+            sawTransientFailure:
+              currentSawTransientFailure ||
```
Tara-ag, line 4293: ⚠️ **MAJOR: OAuth retry success path for streaming does not propagate `retryNextAccount` failure state correctly, and the non-stream helper is missing `ctx`/`body` but the streaming path still needs them.**

The diff removes `ctx` and `body` from `handleAnthropicSuccessfulNonStreamRetryResponse`, which is correct for the non-stream helper. However, the streaming retry path now calls `handleAnthropicSuccessfulResponse` directly with `ctx` and `body`. If that returns `{ retryNextAccount: true, failure }`, the code extracts `failure` and returns `continueLoop: true`, but it does not call `recordAttemptError` or `logAttempt` for the failed streaming preflight — those were already done inside `handleAnthropicSuccessfulResponse` for the streaming path, but the non-stream helper now duplicates the `logAttempt` call. This asymmetry means a streaming OAuth retry that fails at preflight will not have its retry attempt number attached to the attempt log (the inner `logAttempt` uses the original `retryAttemptNumber` via `retryLogAttempt`, so this is actually okay).

The real issue is that `handleAnthropicSuccessfulNonStreamRetryResponse` no longer receives `ctx`/`body`, but if a future change needs to log request context there, the signature is narrower. More importantly, the non-stream helper now calls `logAttempt` directly, but the `retryLogAttempt` wrapper already adds `attempt` and `attemptDurationMs`. The direct call passes `attemptDurationMs` in `extra`, which `retryLogAttempt` preserves via nullish coalescing. This is correct.

However, there is a subtle bug: when the streaming OAuth retry succeeds, `handleAnthropicSuccessfulResponse` returns `{ response: result }` where `result` is a `Response`. The caller then returns `authRetryResult.response`. But `handleAnthropicSuccessfulResponse` for streaming returns `{ response: result }` where `result` is the output of `attachAnthropicSuccessStreamTelemetry`, which returns a `Response`. So the final return is a `Response`. This is correct.

No blocking issue here, but the asymmetry between stream and non-stream retry logging is worth noting. Consider adding a comment or unifying the retry wrappers.

src/lib/proxy/requestLogger.ts
```diff
@@ -33,9 +33,54 @@ import { OtelBridge } from "../observability/otelBridge.js";
...
+export async function flushRequestLogs(
+  timeoutMs: number = REQUEST_LOG_IO_TIMEOUT_MS,
+): Promise<void> {
+  const deadline = Date.now() + Math.max(1, timeoutMs);
+  while (pendingLogOperations.size > 0) {
+    const admitted = [...pendingLogOperations];
+    const remainingMs = Math.max(1, deadline - Date.now());
+    try {
+      await withTimeout(
+        Promise.allSettled(admitted),
```
Tara-ag, line 62: ⚠️ **MAJOR: `flushRequestLogs` deadline accounting double-counts timeout failures and can drop newly-admitted operations.**

When `withTimeout` rejects, the catch block deletes all `admitted` operations from the set. However, between building `admitted` and the rejection, new operations may have been added to `pendingLogOperations`. Those new operations survive the catch deletion, but the thrown error claims the flush failed for the original count. Worse, the `if (Date.now() >= deadline ...)` branch clears the entire set unconditionally after a successful `Promise.allSettled`, including operations that were admitted *after* the `admitted` snapshot was taken. Those newer operations are never awaited; they are simply discarded, which defeats the purpose of flushing during shutdown.

Suggested fix: snapshot and await only the currently pending operations, then re-evaluate the set in the next loop iteration without clearing operations that were not part of the awaited batch:

```ts
export async function flushRequestLogs(
  timeoutMs: number = REQUEST_LOG_IO_TIMEOUT_MS,
): Promise<void> {
  const deadline = Date.now() + Math.max(1, timeoutMs);
  while (pendingLogOperations.size > 0) {
    const admitted = [...pendingLogOperations];
    const remainingMs = Math.max(1, deadline - Date.now());
    try {
      await withTimeout(
        Promise.allSettled(admitted),
        remainingMs,
        `Timed out flushing ${admitted.length} proxy request log operation(s)`,
      );
    } catch (error) {
      for (const operation of admitted) {
        pendingLogOperations.delete(operation);
      }
      throw error;
    }
    if (Date.now() >= deadline && pendingLogOperations.size > 0) {
      const remaining = pendingLogOperations.size;
      pendingLogOperations.clear();
      throw new Error(
        `Timed out flushing ${remaining} proxy request log operation(s)`,
      );
    }
  }
}
```

Even with this change, consider whether dropping timed-out operations silently is safe for durability — at minimum the warning log should include the count.

src/lib/proxy/proxyAnalysis.ts
```diff
@@ -471,6 +499,69 @@ export async function analyzeProxyLogs(
...
```
Tara-ag, line 527: ⚠️ **MAJOR: Artifact path containment check can be bypassed by symlinks and null-byte injection, and the analyzer reads attacker-controlled paths.**

The check uses `relative(bodiesRoot, resolvedBodyPath)` plus `startsWith("..")` and `isAbsolute`, which is better than a simple prefix check. However, it still resolves symlinks in `bodyPath` via `resolve(logsDir, bodyPath)`, so a symlink pointing outside `bodies/` will be followed and the resulting path will be considered "inside" if the symlink target happens to be under `bodies/`. More critically, `bodyPath` comes from JSONL records that an attacker who can write to the debug log could craft. On Linux/macOS, `resolve()` does not treat `\0` specially, but the subsequent `stat()` call will fail; on Windows, null bytes can truncate paths. The bigger issue is that the analyzer is running offline with operator-supplied logs, but if those logs are shared, this becomes a path-traversal vector.

Suggested fix: validate that `bodyPath` is a relative path with no `..` components *before* resolving, and use `realpath` to detect symlinks escaping the sandbox:

```ts
const bodyPath = stringValue(record.bodyPath);
if (!bodyPath) {
  return;
}
// Reject absolute paths and any parent traversal before resolving.
if (isAbsolute(bodyPath) || bodyPath.split(/[\\/]/).includes("..")) {
  invalidPaths += 1;
  return;
}
const resolvedBodyPath = resolve(logsDir, bodyPath);
const relativeBodyPath = relative(bodiesRoot, resolvedBodyPath);
if (
  !relativeBodyPath ||
  relativeBodyPath.startsWith("..") ||
  isAbsolute(relativeBodyPath)
) {
  invalidPaths += 1;
  return;
}
referencedArtifacts.add(resolvedBodyPath);
```

Also consider documenting that the analyzer trusts the log directory and should not be run against untrusted logs.

Tara-ag (changes requested):
## Review Summary

Reviewed the proxy restartless-observability changes. The PR addresses real gaps, but I found **1 CRITICAL** and **5 MAJOR** new issues that need to be fixed before merge.

### New issues raised

| Severity | Count | Files |
|----------|-------|-------|
| 🔒 CRITICAL | 1 | `src/lib/proxy/rollingWorkerProcess.ts` |
| ⚠️ MAJOR | 5 | `src/lib/proxy/requestLogger.ts`, `src/lib/proxy/proxyAnalysis.ts`, `src/lib/proxy/rollingProxyServer.ts`, `src/lib/server/routes/claudeProxyRoutes.ts`, `src/lib/proxy/socketWorkerRuntime.ts` |
| 💡 MINOR | 0 | — |
| 💬 SUGGESTION | 0 | — |

### Blocking reasons

1. **CRITICAL correctness bug in socket handoff** (`rollingWorkerProcess.ts`): the supervisor can send `socket-cancel` to a worker after the worker has already accepted and committed the socket, which may destroy a live connection.
2. **≥3 MAJOR issues**: shutdown log-flush can drop newly-admitted operations; the offline analyzer's path containment is bypassable via symlinks; recovery backoff state is not reset on successful explicit replacements; pending socket event listeners leak after commit.

### Next steps

Please address the inline comments and re-request review. Focus first on the CRITICAL socket-cancel race and the MAJOR shutdown-flush / path-traversal issues, as those have the highest operational/security impact. The existing resolved review threads were not re-raised.

src/lib/proxy/rollingWorkerProcess.ts
murdore: Fixed in eb3feca4. The parent now sends socket-cancel only while the transfer is still unaccepted; after socket-accepted, ownership remains with the worker and a late parent timeout cannot destroy the active socket. The regression suite covers late ACK/timeout behavior and response preservation. The related worker-exit path now also uses ErrorFactory.proxyWorkerLifecycle with worker context.

src/lib/proxy/socketWorkerRuntime.ts
murdore: Fixed in eb3feca4. Pending transfers now retain the exact pre-commit error/close listener references and remove only those listeners on commit, cancel, send failure, drain, and shutdown; unrelated HTTP/server listeners are preserved. Added a regression test that verifies both exact listener removal and preservation of unrelated listeners.

src/lib/proxy/rollingProxyServer.ts
murdore: Verified as already handled in the current branch. After a successful explicit queueReplacement, rollingWorkerSupervisor resets recoveryFailures to 0 immediately; successful recovery also resets it. A failed explicit replacement re-arms recovery when there is no active/candidate worker, and the existing regression test covers that path. No code change was needed for this thread.

src/lib/server/routes/claudeProxyRoutes.ts
murdore: Verified as non-blocking. The auth retry remains inside the same lifecycle accumulator: the retry attempt is recorded before dispatch, total duration is finalized on the terminal result, and a successful streaming retry is returned to the caller with its response stream intact. This matches the thread analysis, so no code change was required.

src/lib/proxy/requestLogger.ts
murdore: Fixed in eb3feca4. flush now snapshots the operations admitted at flush start and only settles/removes that snapshot; operations admitted while the flush is waiting remain queued for the next flush instead of being cleared at the deadline. Added a fake-timer regression test covering a late-admitted write across a timed-out flush.

src/lib/proxy/proxyAnalysis.ts
murdore: Fixed in eb3feca4. Artifact resolution now rejects NUL bytes and parent-directory components lexically, then compares realpath-resolved body/artifact paths against the canonical bodies root so symlink escapes are excluded. Missing, invalid, and present references are counted separately. Added regressions for symlink escape and NUL-bearing artifact paths.

murdore:
Final review update on eb3feca4 (rebased onto current origin/release v10.4.2):

- Fixed the post-accept socket-cancel race.
- Removed exact pending transfer listeners without disturbing HTTP/server listeners.
- Preserved log operations admitted during a timed-out flush.
- Added lexical and canonical realpath containment for analysis artifacts.
- Switched the worker transfer-exit path to ErrorFactory with lifecycle context.
- Verified the recovery counter and OAuth retry findings against current code; both paths already behave correctly.

Verification on the final rebased commit:
- 182/182 focused proxy tests
- 234/234 bugfix suite tests
- full ./pre-commit.sh
- all proxy performance budgets, with 0 concurrent handoff failures and 0 dropped requests

All review threads are answered and resolved. Fresh CI and Yama re-review are running.

Tara-ag (approved):
## Review summary

Reviewed the proxy restartless-observability changes in PR #1223. This is a final verification pass over the latest commit (`eb3feca4`).

### Previously raised blocking issues — verified fixed

- **CRITICAL socket-cancel race** (`src/lib/proxy/rollingWorkerProcess.ts`): `socket-cancel` is now sent only while `!pending.accepted`, so a late parent timeout cannot destroy a socket the worker already owns.
- **MAJOR worker-side listener leak** (`src/lib/proxy/socketWorkerRuntime.ts`): pending pre-commit `error`/`close` listeners are now removed on commit/cancel/drain/shutdown, while unrelated server listeners are preserved.
- **MAJOR shutdown log-flush accounting** (`src/lib/proxy/requestLogger.ts`): `flushRequestLogs` now snapshots admitted operations and only settles/removes that snapshot; operations admitted mid-flush remain queued for the next flush instead of being silently dropped.
- **MAJOR offline analyzer path traversal** (`src/lib/proxy/proxyAnalysis.ts`): artifact paths now reject NUL bytes and `..` components lexically, then compare `realpath`-resolved paths against the canonical bodies root so symlink escapes are excluded.
- **MAJOR replacement serialization** (`src/lib/proxy/rollingProxyServer.ts`): environment, explicit, and recovery replacements share the `queueReplacement` serializer; pending timer callbacks carry a monotonic schedule token and are invalidated by explicit replacement/close.
- **MAJOR non-stream OAuth retry duration** (`src/lib/server/routes/claudeProxyRoutes.ts`): `handleAnthropicSuccessfulNonStreamRetryResponse` logs `attemptDurationMs`, and `retryLogAttempt` preserves an explicitly passed duration.
- **MINOR env-hash normalization** (`src/lib/proxy/runtimeConfig.ts`): `envFileHash` now hashes parsed, sorted key/value pairs so comment/whitespace edits do not trigger worker replacement.

0 — no new blocking or major issues found in the current commit.

### Verification performed

- Re-read the changed source files at `refs/pull/1223/head`.
- Cross-checked the fixes against the resolved review threads.
- Confirmed test coverage additions in `test/proxyRollingWorkerHandoff.test.ts`, `test/proxyObservabilityFoundation.test.ts`, `test/proxyReliabilityHardening.test.ts`, `test/proxyAnalysis.test.ts`, and `test/proxyConfigHotReload.test.ts`.

The changes stay within the proxy subsystem, do not touch the public SDK API, and the previously identified blocking concerns are addressed. Approving.

Issue #563, opened 2025-12-01
IMG-031: No Format Conversion


murdore:
## Summary
Unsupported formats (HEIC, TIFF) rejected instead of converted.

## Fix
Implement format conversion using sharp library (HEIC→JPEG, TIFF→PNG).

## Acceptance Criteria
- [ ] Implement format conversion using sharp
- [ ] Support HEIC→JPEG conversion
- [ ] Support TIFF→PNG conversion
- [ ] Configurable output format
- [ ] Add tests for format conversion

**Estimated Effort:** 8h

murdore:
Resolved on `release`. `src/lib/adapters/imageFormatSupport.ts` converts formats vision providers can't read natively (TIFF, AVIF, ICO, JPEG 2000, HEIC/HEIF where the provider lacks it) with sharp instead of rejecting them, and #1738 made the target caller-configurable via `imageOptions.outputFormat` (default PNG, e.g. `jpeg` for HEIC→JPEG). Covered end-to-end: `test/continuous-test-suite-vision-image-format.ts` asserts the decoded bytes on the wire (JPEG when asked, PNG by default, on both `generate()` and `stream()`), and `test/continuous-test-suite-file-formats.ts` covers the format matrix. Closing; the original acceptance criteria (sharp conversion, HEIC→JPEG, TIFF→PNG, configurable output, tests) are all met.

Issue #357, opened 2025-12-01
FD-010: Add Support for Missing File Formats


murdore:
## Bug Summary
FileDetector only supports 4 file types (CSV, image, PDF, text). Missing support for HEIC/AVIF images, Office documents, data formats (JSON/YAML/XML), and archives.

## Root Cause Analysis
**Why this bug exists:**
Designed for core multimodal AI use cases. Additional formats now commonly requested.

**Missing formats:**
- Images: HEIC, AVIF, BMP, TIFF, ICO
- Data: JSON, YAML, XML
- Archives: ZIP, GZIP, 7Z
- Office: DOCX, XLSX, PPTX

**Impact:**
- Limited multimodal support
- Manual preprocessing required
- Missed use cases

## Phased Approach

**Phase 1 (4h) - Image Formats:**
- BMP, HEIC, AVIF, TIFF, ICO magic bytes

**Phase 2 (4h) - Data Formats:**
- JSON, YAML, XML heuristics

**Phase 3 (2h) - Archives:**
- ZIP, GZIP, 7Z magic bytes

**Phase 4 (6h) - Office Formats:**
- DOCX, XLSX, PPTX (ZIP-based, complex)

## Acceptance Criteria
- [ ] New FileType definitions
- [ ] Magic bytes for images/archives
- [ ] Heuristics for data formats
- [ ] Tests for all formats
- [ ] Documentation updated

**Files:** src/lib/utils/fileDetector.ts, src/lib/types/fileTypes.ts
**Effort:** 16h (phased)

murdore:
### ◐ Audit status: partially done — keeping open

Core behaviour ships, but not every acceptance criterion is met. The remaining gap is the actionable part:

> Substantial format coverage was added (archives, Office docs, data-format heuristics, most image extensions) but two explicit acceptance criteria — HEIC support and image/archive magic-byte detection — remain unimplemented.

<details><summary>what already works (evidence)</summary>

[...]

PR #1229, opened 2026-07-23
feat(file-detector): add HEIC support and image/archive magic-byte detection

morgan-coded:
This addresses the two remaining unimplemented criteria from the #357 audit: HEIC support and image/archive magic-byte detection, scoped to the phased plan.

**Status against the FD-010 format list:**

| Format | This PR | Notes |
|---|---|---|
| HEIC / HEIF | content detection + extension rows added | ftyp majors `heic/heix/hevc/hevx` plus sequence brands; `mif1/msf1` resolved by scanning compatible brands; `.heic`/`.heif` type + MIME rows |
| BMP, TIFF (both byte orders), ICO | magic bytes added | real published prefixes as test vectors |
| 7Z | magic bytes added | wired into detection; the existing error-text helper is untouched |
| AVIF | already detected — regression-locked | the audit listed it as missing, but `avif/avis/avio` majors were already handled at this revision; this PR adds regression tests including the `mif1`-compatible-brand ordering cases |
| JSON / YAML / XML, ZIP / GZIP | already shipped | asserted in the new suite |
| DOCX / XLSX / PPTX | extension-mapped today | ZIP-content sniffing left for Phase 4 as the plan stages it |

Behavior notes: unknown ftyp brands still fall through to `video/mp4` (regression-locked), and a valid HEIC now reaches a truthful `Invalid MIME type: image/heic is not in allowed list` error instead of the misleading corrupted-image path — provider acceptance and conversion (#563) are deliberately untouched. The new continuous suite has 30 keyless cases wired into `test:unit`, with sample bytes taken from published references (Nokia HEIF conformance image, LibTIFF test images, the format specifications).

Happy to stage this differently or split the 7z piece out if you'd prefer.

## Summary by CodeRabbit

* **New Features**
  * Enhanced image file-type detection for HEIC/HEIF using both content-based format brand inspection and extension mapping, with improved AVIF/HEIC/HEIF classification.
  * Added/expanded magic-byte detection for BMP, TIFF (both endians), ICO, and 7-Zip.
* **Bug Fixes**
  * Reduced icon mis-detections by tightening ISO-based handling.
  * Updated “unsupported image” messaging to include HEIC/HEIF.
* **Documentation**
  * Clarified HEIC/HEIF handling and noted that non-native providers still need PNG/JPEG conversion.
* **Tests**
  * Added a continuous test suite covering image/archive detection and HEIC/HEIF edge cases.


morgan-coded:
Amended in: `ImageProcessor.detectImageType` now recognizes ICO (`image/x-icon`), so buffers that `FileDetector` routes as images get the same truthful unsupported-format rejection as HEIC instead of the generic unknown-signature error; the honest-error format list names ICO as well, and the suite covers the path (31 cases). The ISO-BMFF consolidation suggestion is left as-is to keep this PR's scope narrow — happy to do it as a follow-up if you'd like it.

murdore:
Bug (reproduced against this PR's head `b4b6881`): the new ICO magic-byte check (`bytes 0-3 == 00 00 01 00` → image/x-icon @ 95%) is ordered *before* the ftyp-box block this same PR expands. Any ISO-BMFF file (mp4/mov/m4a/heic/heif/avif) whose declared ftyp box size is exactly 256 bytes (`0x00000100`) has its first 4 bytes collide with the ICO signature, so it never reaches the ftyp branch. I built a repro: a well-formed `heic` ftyp buffer (60 compatible brands) padded to exactly 256 bytes returns `{mimeType:"image/x-icon",confidence:95}` instead of `image/heic`. The collision is specific to ICO (BMP/TIFF would need >1GB box sizes to collide). None of the 31 new tests cover it — suggest moving the ftyp-box check before the ICO check, or having the ICO branch also require bytes 4-7 ≠ `ftyp` before returning.

Separately: CodeRabbit's duplication finding (the AVIF/HEIC/HEIF brand Sets + ftyp-scan logic copied verbatim between `fileDetector.ts` and `imageProcessor.ts`) is still unresolved — given the `avio`-brand drift history in this repo, worth either de-duplicating or opening a tracked follow-up rather than leaving it deferred in a comment.

morgan-coded:
Fixed both: the ICO branch now requires bytes 4-7 ≠ `ftyp` before returning, and your exact repro — a 256-byte-box heic ftyp buffer with 60 compatible brands — is in the suite as a named case, alongside 256-box M4A/AVIF variants, box-size boundary cases, and a real-ICO guard (35 cases now). The duplication is resolved rather than deferred: the brand sets and ftyp scan moved into a shared `isoBmff.ts` consumed by both `FileDetector` and `ImageProcessor`, so the `avio`-class drift can't recur.

PR #1233, opened 2026-07-24
fix(core): honor Retry-After hints and add a no-hint retry floor

### LINKED ISSUE #1189: Video generation retries ignore Replicate's retry_after — burst-throttled accounts fail every request
## Observed live (Director draft-tier pipeline, 2026-07-17)

On a Replicate account under the low-credit throttle (**6 predictions/min, burst 1** — the 429 body says the condition is *remaining credit* < $5), every `nl.generate` video call fails even though a single well-timed submit would succeed.

Timestamps from one call (director run, replicate video provider):

```
01:35:59.618  429 — Request was throttled... resets in ~8s  (retry_after: 8)
01:36:00.164  429 — Request was throttled... resets in ~8s  (retry_after: 8)   ← 0.5s later
```

The internal retries fire **~0.5–1s apart** while the response explicitly says `retry_after: 8`. All attempts land inside the same throttle window, so the aggregate call always fails — and because the attempts themselves count against the limit, the *next* call's first attempt usually arrives still-throttled. One under-provisioned account turns into a 100% failure rate on a path that needs only a few seconds of patience.

## Suggested fix

In the retry path that wraps provider generate calls (the `Generate failed for replicate` → retry loop):

1. When the error carries an HTTP 429 / `retry_after`, **wait at least `retry_after` seconds** (it's right there in the Replicate response body, currently discarded when the error is stringified).
2. Fallback when no hint is available: exponential backoff with a floor ≥ the common throttle window (~10s), rather than ~1s linear.

## Downstream context

Director mitigated locally with submit pacing + rate-limit-only backoff at its own choke point (juspay/director#104), but every other NeuroLink consumer hitting Replicate/Kling/Runway rate limits inherits the same cascade. Related recent video-path PRs with live evidence: #1150, #1157, #1164.

morgan-coded:
`withProviderRetry` backed off exponentially at 1s then 2s, rather than the ~1s linear loop described in #1189; AI SDK errors reached it with `responseHeaders` intact, while Replicate submissions bypassed it and discarded those headers.

This extracts the existing Retry-After parser for numeric seconds, HTTP dates, and rate-limit fallbacks, uses server delays capped at 120s or a 10s no-hint floor in `withProviderRetry`, and carries Replicate's header through `NeuroLinkError` while retrying prediction submission, with polling behavior otherwise unchanged.

The 10s no-hint floor can delay real model-pool failover while a member retries internally — kept the floor the issue proposes, flagging that tradeoff for your call; the shape follows the existing `ClientRateLimitError.retryAfter` and `accountPool.markQuotaExceeded(..., retryAfterMs)` handling.

I added a focused offline suite with 9 passing tests, then ran format, source lint, validation, builds, CLI help, strict TypeScript, and all 23 unit legs locally; CI runs no test suites, and the only local failures were the five byte-matching baseline failures. Fixes #1189

## Summary by CodeRabbit

* **Reliability Improvements**
  * Improved automatic retries for provider requests by honoring server `Retry-After` and related rate-limit headers, with bounded minimum/maximum delays.
  * Enhanced rate-limit handling to reliably compute non-negative wait times from common headers.
  * Improved prediction submission/polling resilience by treating `429` and `5xx` as retriable and propagating retry timing hints.
  * Extended error details to optionally include `retryAfterMs` for more precise retry behavior.
* **Tests**
  * Added dedicated coverage for retry-delay parsing, clamping/bounding behavior, rate-limit handling, and prediction retry flows.
* **Chores**
  * Added a focused retry-delay Vitest runner and included it in unit test runs.


murdore:
@morgan-coded can you resolve conflicts?

morgan-coded:
Resolved the conflict by taking the `10.6.4` version bump and keeping both `test:agent-runtime:vitest` and `test:retry-after:vitest` in the unit-test chain.

PR #1235, opened 2026-07-27
feat(agent): add isolated sub-agent runtime and host-loop delegation

pdogra1299:
## Summary

Turns the dormant `src/lib/agent/` module into a production sub-agent runtime, driven by the requirements of its consumers (which today hand-roll isolated "worker" sub-agents on top of `generate()`). Work items N1–N5; RFC for N4/N5 at `docs/plans/2026-07-27-isolated-agent-runner-rfc.md` with a field-validation matrix against the three reference consumer implementations.

**N1 — Agent plumbing.** `AgentExecutionOptions` gains `abortSignal` + `turnTimeoutMs`/`wrapupTimeLeadMs`/`stallTimeoutMs`, forwarded verbatim into `generate()`/`stream()` (an already-aborted parent never starts a run). `buildAgentTools()` honors `AgentDefinition.inputSchema` and passes the parsed object through. `AgentResult` gains `stopReason` and real `toolExecutions`.

**N2 — Real `toolExecutions`.** The stub on `GenerateResult` is replaced with per-call `ToolExecutionRecord`s (`toolName, params, resultText, isError, startedAt, durationMs`), captured by a per-request `ToolExecutionRecorder` that wraps tools at the two central assembly points (`prepareGenerationContext` / `getToolsForStream`) — before tool discovery, so `search_tools` mid-turn hydration stays wrapped. Bounded by a new `toolExecutionCapture` option (8KB/result, 500 records, `onRecord` callback).

**N3 — Worker mode + sampling strip.** `NeuroLink.createWorkerInstance()`: memory off, orchestration off, observability inherited with `autoDetectExternalProvider` + `skipLangfuseSpanProcessor`, credentials inherited, host tool registry shared, tagged log bridge. `dispose()` now only clears the process-global log sink it owns. The model registry gains `modelSupportsSamplingParams`/`resolveSamplingParams` (explicit registry capability wins, else rejecting-family patterns: Sonnet 5, Opus 4.7+, Opus 5, Fable/Mythos); `temperature`/`topP` are stripped, with a debug log, at every request-build site so retry/fallback rebuilds inherit the strip.

**N4 — `runIsolatedAgent()`.** Two-pass runner: research with tools under the real turn-budget machinery (wrap-up nudge, stall watchdog, honest `stopReason`; a budget-capped run consolidates — never a bare wall-clock kill), then an extraction pass that ALWAYS runs tools-off on its own timeout, fed from the execution records (a dead research generate still extracts). Structured-recovery ladder (`structuredData` → raw JSON → fence → brace span → top-level array, through `coerce`) with corrective re-asks carrying validation errors + `shapeDoc`; delivery guarantee via mechanical digest; abort chaining; 9 fire-and-forget lifecycle events; leashed legs with TTL'd handles (`continueAgent(handle, guidance)` / `stopAgent`), tombstoned outcomes retrievable exactly once; mechanical waste signatures (duplicate hash, empty/error streaks, no-new-results) end a leg early.

**N5 — `registerAgentTool()`.** Wraps N4 as a tool on the HOST instance so its existing generate loop delegates (no second router generate): per-turn caps counted in the loop via an ALS turn scope, depth limits that withhold the tool from the request at the limit, a process-wide concurrency pool with queue timeout (shared with `AgentNetwork` standalone mode), and every refusal carrying its recovery instruction. `AgentNetwork` exports stay field-compatible.

### Provider parity

The sampling strip and execution capture are provider-uniform, driven by one registry predicate and one recorder:

- capture covers the AI-SDK loop and the native Vertex Gemini/Claude and AI Studio loops through the shared tool-assembly points;
- the strip is applied at direct Anthropic (generate + stream), Vertex Claude (both native loops, incl. forced-finalization/backstop rebuilds), OpenAI-compatible `buildBody` (litellm et al., incl. 400 retries), Bedrock's native Converse/ConverseStream builders (which previously forced `temperature: 0.7`), the generic AI-SDK `generateText` call (every remaining @ai-sdk provider), and both Gemini config builders (Vertex + AI Studio) for uniformity.

## Migration note — `GenerateResult.toolExecutions` shape change

Entries were the stub `{name, input, output}` (always `duration: 0`); they are now real `ToolExecutionRecord`s. Consumers reading the old fields get `undefined`:

| Old field | New field | Notes |
| --- | --- | --- |
| `name` | `toolName` | |
| `input` | `params` | as parsed by the loop |
| `output` | `resultText` | serialized + bounded (~8KB default, raisable via `toolExecutionCapture`) |
| `duration` | `durationMs` | real wall-clock |
| — | `isError` | new: thrown errors AND error-shaped results |
| — | `startedAt` | new: epoch ms |

`StreamResult.toolExecutions` is deliberately unchanged (legacy summaries) on every provider.

## Test plan

- `pnpm exec tsc --noEmit --strict` and `pnpm run check` — 0 errors
- `pnpm run lint` — 0 errors (46 pre-existing warnings)
- New unit suites (`pnpm run test:agent-runtime:vitest`, chained into `test:unit`): 80 tests — budget/wrap-up threading, abort chaining, recovery ladder, leg/waste/handle lifecycle incl. TTL expiry (fake timers), per-turn cap counting, depth withholding
- Regression (no-API): `test:anthropic-tools-policy`, `test:mcp`, `test:mcp:bash`, `test:mcp:limits`, `test:mcp:spans`, `test:model-capabilities:vitest`, `test:step-budget-guard:vitest` — all pass
- `pnpm run build` (SDK + CLI + publint) — pass, plus a dist smoke of the new public surface

## Summary by CodeRabbit

* **New Features**
  * Added isolated agent runs with research + structured extraction, including leashed continuation/stop, budgets/timeouts, tool-evidence capture, and lifecycle events.
  * Added host-loop delegation with depth limiting and process-wide concurrency pooling.
  * Added bounded tool-execution recording plus structured-output recovery with validation and retry.
  * Added registry-driven sampling-parameter support and automatic omission when unsupported.
* **Bug Fixes**
  * Early-abort handling and improved preservation of stop reasons and tool-execution details; updated tracing to match recorded tool execution shape.
* **Tests**
  * Expanded Vitest coverage for agent runtime, delegation, structured recovery, recording, and sampling; added a dedicated test script.
* **Documentation**
  * Added an RFC describing isolated-agent runner behavior.


src/lib/agent/isolatedAgentRunner.ts
```diff
@@ -0,0 +1,1032 @@
...
+      runId,
+      agentId: definition.id,
+      timestamp: Date.now(),
+    });
+  }
+  if (result?.usage) {
+    addUsage(session.usage, result.usage);
+  }
+
+  // Maintain the worker's conversation for continuation legs.
```
Pdogra2520, line 622: **Resume history carries no tool activity — waste-tripped legs resume blind and re-trip immediately.**

The history gets only the user prompt and (when the generate survived) the final assistant text. Tool calls/results live solely in `allRecords`, which never reaches the continuation context. Two consequences:

1. On a **waste-trip or tool-budget leg**, `legController.abort()` kills the generate mid-step → `result` is `undefined` → *no assistant message at all* is pushed. The resumed worker sees `user: <task>` + `user: "Continue the investigation…"` and literally nothing about what it already did.
2. It will re-run the same queries it already ran — and since `seenCallHashes` persists across legs, the duplicate-call signature trips on the *first* repeated call → another `in_progress` with 1 tool call → supervisor continues → trips again. The exact legs where supervision matters most (waste) become an in_progress ping-pong until TTL or `stopAgent`.

Suggested fix, cheap and local: when a leg ends without `result?.content`, push a synthetic assistant message built from what the runner already has — `buildDelta(legRecords)` + tripped `wasteSignals`, e.g. `"[leg ended early — waste: duplicate-call…] Tool activity so far:\n<delta lines>"`. That gives the resumed model (and `askNextPlan`, which replays the same history) an accurate view of its own prior work.

src/lib/agent/isolatedAgentRunner.ts
```diff
@@ -0,0 +1,1032 @@
...
+  const prompt =
+    typeof input === "string" ? input : JSON.stringify(input, null, 2);
+
+  let keepAlive = false;
+  try {
+    const legOutcome = await runResearchLeg(session, prompt);
+
+    // Leashed mode: a leg that hit its budget (or tripped a waste signature)
+    // without the model finishing returns in_progress with a resume handle.
+    const researchFinished =
```
Pdogra2520: **`researchFinished` (here and `continueIsolatedAgent:950`) depends on `stopReason === "completed"` — but the chat-completions loop never sets `stopReason`.**

Verified on this branch: `openaiChatCompletionsClient.ts` has zero `stopReason` assignments; parity exists only in `GenerationHandler` (AI-SDK loop) and the native Vertex loops via `resolveTurnStopReason`. On the LiteLLM path (a first-class consumer path — e.g. the euler log-analysis worker runs `provider: litellm`):

- **Leashed:** a leg whose model finishes cleanly returns `stopReason: undefined` → `researchFinished` false → `in_progress` + `askNextPlan` forever; the run can never finalize except via `stopAgent`/TTL.
- **Non-leashed with extraction:** `finalizeOutcome` maps `researchStop !== "completed"` → `"partial"`, so every healthy run on that provider is mislabeled partial.

The cleaner fix is to stop inferring "finished" from provider vocabulary and derive it from the leg's own control flow, which the runner fully owns: a leg is *unfinished* iff `stopReason === "time-limit"` **or** `budgetTripped` **or** a waste signature tripped — i.e. `in_progress` only when something the runner did cut the leg short. A generate that returned normally with `stopReason` undefined is a finished turn. Same treatment in `finalizeOutcome` (`undefined` ⇒ completed). Alternatively (or additionally) add `stopReason` parity to the chat-completions agent loop in this PR — but the control-flow derivation is robust against the next provider that forgets to set it.

Note all runner tests mock `stopReason: "completed"`, which is why this doesn't surface in the suite.

src/lib/agent/agentToolRegistrar.ts
```diff
@@ -0,0 +1,366 @@
...
+  const depth = resolveDepth(host);
+  if (options.maxDepth !== undefined && depth >= options.maxDepth) {
+    return refusal(
+      `Delegation depth limit reached (${depth}/${options.maxDepth}). Complete this investigation yourself with your own tools instead of delegating further.`,
+    );
+  }
+
+  // Open-handle conflict: an in-progress leashed leg for this agent must be
+  // continued, not restarted.
+  const openHandle = hasOpenIsolatedAgentHandle(definition.id);
```
Pdogra2520: **Open-handle conflict is scoped globally (per agent id, across every session and host) — multi-tenant hosts will refuse unrelated conversations.**

`hasOpenIsolatedAgentHandle(definition.id)` scans the process-global session map. In a host serving many concurrent conversations (the primary consumer scenario — a Slack bot with per-thread turns), thread A opening a leashed `o2-logs` leg makes thread B's delegation refuse with *"continue it via its handle"* — a handle that belongs to a different conversation, which thread B's model cannot see or legitimately continue. Worse, if thread B *does* call `continueAgent` with that handle, it injects guidance into thread A's investigation.

Suggested fix: scope the conflict check to the caller. `executeDelegation` already extracts `sessionId` from the execution context, and the session stores `options.toolContext.sessionId` — refuse only when the *same sessionId* owns the open handle, and let the delegation pool bound cross-session concurrency (that's its job). `continueIsolatedAgent`/`stopIsolatedAgent` should ideally verify the same ownership before acting on a handle.

src/lib/agent/agentToolRegistrar.ts
```diff
@@ -0,0 +1,366 @@
...
+      return refusal(
+        `Delegation limit reached: ${name} has already been called ${used}× this turn (max ${options.maxDelegationsPerTurn}). Do not call ${name} again this turn; synthesize from the investigations you already have.`,
+      );
+    }
+    turnState.counts.set(name, used + 1);
+  }
+
+  // Depth limit (defense in depth — the tool is normally withheld at the
+  // limit before the request is even built).
+  const depth = resolveDepth(host);
```
Pdogra2520: **Depth check reads the host instance's tool context and ignores the per-call execution context — nested delegation never hits the limit on the default config.**

`resolveDepth(host)` looks at `host.getToolContext()?.agentDepth`, which is the *host's* context (0). But the path where depth actually matters is: `createWorkerInstance` defaults to `shareToolRegistry: true` → the worker's loop sees the host's registered delegation tools → the worker calls one → `execute(params, context)` fires with the **worker's** tool context, which carries the `agentDepth: depth + 1` this very function set on line ~323. That value arrives in `contextRecord` and is never read — so the defense-in-depth check compares against 0 every time, and `beginDelegationTurn`'s withholding doesn't apply either (`registrationsFor(worker)` is empty — the registration lives on the host's WeakMap entry).

Fix: `resolveDepth` should prefer `contextRecord.agentDepth` (the execution context) over the host instance context: `executeDelegation` → `resolveDepth(host, undefined, contextRecord)` or just read `contextRecord.agentDepth` first. Worth a test: worker-with-shared-registry delegating at `maxDepth: 1` must be refused.

src/lib/neurolink.ts
```diff
@@ -4209,6 +4220,23 @@ Current user's request: ${currentInput}`;
   async generate(
     optionsOrPrompt: GenerateOptions | DynamicOptions | string,
   ): Promise<GenerateResult> {
+    // Host-loop delegation (registerAgentTool): enter a per-turn scope so
+    // delegation caps count against THIS top-level generate, and withhold
+    // depth-limited agent tools from the request. beginDelegationTurn
+    // returns null when a scope is already active, so the re-entrant call
+    // below proceeds through the normal body sharing the turn's counters.
+    if (this.hasAgentTools) {
```
Pdogra2520, line 4228: **`stream()` has no `beginDelegationTurn` wiring — caps and depth withholding silently don't exist for streaming hosts.**

The registered agent tools are in the shared registry, so a streaming host's loop can call them freely; but with no ALS scope, `turnStorage.getStore()` is undefined in `executeDelegation` and the `maxDelegationsPerTurn` check is *skipped entirely* (not failed — skipped), and depth-limited tools are never withheld from the request. The JSDoc on `registerAgentTool` says the policy applies to "every generate()/stream() the host runs".

Either wire `stream()` through the same `beginDelegationTurn` scope (the re-entrant pattern here transfers directly), or — if that's deliberately deferred — make `executeDelegation` fail closed when a cap is configured but no turn scope exists, and say so in the JSDoc. Silent non-enforcement is the worst of the three options: the 7-delegations-per-turn storm this cap exists to prevent happened on a production host.

src/lib/agent/isolatedAgentRunner.ts
```diff
@@ -0,0 +1,1032 @@
...
+    }
+  }
+}
+
+/**
+ * Resume a leashed run. `guidance`, when provided, is appended as a user
+ * turn before the next leg — this is how a supervisor re-steers a drifting
+ * worker. An expired handle returns its tombstoned final outcome once.
+ */
+export async function continueIsolatedAgent(
```
Pdogra2520, line 1023: **No in-flight guard on a handle — two concurrent `continueAgent(handle)` calls race on the same worker.**

Both find the session in the map, both push to `history`, both run `runResearchLeg` on the same worker instance concurrently (interleaved `conversationMessages`, shared `seenCallHashes`, double `registerHandle` timers), and both may `sessions.delete` in `finally`. An LLM supervisor retrying a slow tool call is exactly the caller that will do this.

Suggested: set `session.inFlight = true` for the duration of a leg; a second `continueAgent` (and `stopAgent`) on an in-flight handle returns a refusal-shaped outcome ("a leg is already running for this handle — wait for it to return") rather than racing. Cheap to add now, ugly to debug later.

src/lib/core/toolExecutionRecorder.ts
```diff
@@ -0,0 +1,304 @@
...
+    return wrapped;
+  }
+
+  /**
+   * Attach this recorder to a request options object. The property is
+   * enumerable on purpose: downstream stages spread options (`{...options}`)
+   * and the recorder must survive into the provider loops.
+   */
+  attachTo(options: Record<string, unknown>): void {
+    options[RECORDER_OPTIONS_KEY] = this;
```
Pdogra2520, line 235: **Enumerable recorder + TS-`private` fields = the captured payloads ride along with any options serialization.**

The enumerable attach is the right call for spread survival — but TS `private` is compile-time only: `records` is a runtime-enumerable own property, so any `JSON.stringify(options)` / options-dump in logging or telemetry now drags up to `maxRecords × maxResultChars` (~4 MB at defaults) of tool results with it. Options objects do get serialized in debug paths (`transformParamsForLogging`-style helpers, error-context dumps).

Suggested: make the internal state ES `#private` fields (`#records`, `#droppedRecords`) — they're invisible to `JSON.stringify` and spreads, so the recorder stringifies as `{}` while the enumerable options key keeps surviving spreads exactly as intended. One-line class change, removes a whole class of accidental log-bloat/PII-echo.

src/lib/agent/isolatedAgentRunner.ts
```diff
@@ -0,0 +1,1032 @@
...
+    args.researchContent ? `Researcher's notes:\n${args.researchContent}` : "",
+  ]
+    .filter(Boolean)
+    .join("\n\n");
+
+  let prompt = basePrompt;
+  let attachSchema = true;
+  const allErrors: string[] = [];
+
+  for (let attempt = 0; attempt <= maxRetries; attempt++) {
```
Pdogra2520, line 373: Composed extraction wall-clock is unbounded by any single knob: worst case `(maxRetries + 1) × timeoutMs` (default 3 × 60s) *plus* `askNextPlan`'s 20s, all **after** the research budget. Callers that derive an outer tool-execution ceiling from the leg budget (the consumer pattern this is built for — registration-level timeouts wrapping the delegation tool) need the composed number or they'll abort completed extractions. Either add a phase-level deadline (`extraction.totalTimeoutMs` bounding all attempts) or document the composed bound prominently on `ExtractionOptions` so ceiling arithmetic can include it.

src/lib/agent/isolatedAgentRunner.ts
```diff
@@ -0,0 +1,1032 @@
...
+    if (options.abortSignal.aborted) {
+      onParentAbort();
+    } else {
+      options.abortSignal.addEventListener("abort", onParentAbort, {
+        once: true,
+      });
+    }
+  }
+
+  const onRecord = (record: ToolExecutionRecord) => {
```
Pdogra2520, line 490: Both `tool_call` and `tool_result` emit from `onRecord`, i.e. **after** the execution completes — a 90-second query emits nothing until it returns. Fine for accounting; misleading for the primary consumer of `onEvent` (live progress/status surfaces), and the event name `tool_call` reads as "call started". Worth either a doc note on `AgentRunEvent` ("both fire post-execution") now, or a pre-execution hook on the recorder wrapper later — the wrapper is already the single choke point, so an `onStart` there is a 5-line addition when needed.

src/lib/types/generate.ts
```diff
@@ -767,6 +774,53 @@ export type AdditionalMemoryUser = {
...
+/**
+ * One real tool invocation captured during an agentic turn.
+ *
+ * Replaces the historical `{name, input, output}` stub on `GenerateResult`:
+ * every record is produced at the actual execution site (AI-SDK loop and the
+ * native Gemini/Anthropic loops alike), so `params`, timing, and error status
+ * reflect what really ran — consumers no longer need proxy "recorder" tools
+ * to observe their own tool traffic.
+ */
+export type ToolExecutionRecord = {
```
Pdogra2520, line 786: Public shape change worth a loud CHANGELOG/migration note: `GenerateResult.toolExecutions` entries were `{name, input, output}` — any consumer reading `.name`/`.input`/`.output` now gets `undefined` (fields are `toolName`/`params`/`resultText`). The JSDoc here is good, but this ships as a `feat` minor and downstream code fails silently (undefined tool names in dashboards), not loudly. A one-line "BREAKING-ish" entry in the changelog + the field mapping table would save every consumer the same debugging hour.

src/lib/agent/agentToolRegistrar.ts
```diff
@@ -0,0 +1,366 @@
...
+  const name = options.name ?? definition.id;
+  const registrations = registrationsFor(host);
+  if (registrations.has(name)) {
+    throw new Error(
+      `registerAgentTool: an agent tool named "${name}" is already registered on this instance`,
+    );
+  }
+  if (options.maxConcurrent !== undefined) {
+    // Process-wide pool: the largest registered capacity wins.
+    poolCapacity = Math.max(poolCapacity, options.maxConcurrent);
```
Pdogra2520, line 242: `Math.max` semantics mean pool capacity can only ratchet up: a consumer registering `maxConcurrent: 1` still gets the default 4, and one registration asking for 16 raises the shared pool for every agent in the process permanently. If per-registration limits are the intent, this needs per-registration counters alongside the global pool; if global-max-wins is the intent, the JSDoc on `maxConcurrent` should say "raises the process-wide pool; cannot lower it" so nobody reaches for it as a per-agent throttle.

Pdogra2520 (commented):
Reviewed the full branch locally (all 39 files, not just the diff). Overall: strong implementation — N1 forwarding is verbatim and complete, the recorder design is genuinely good (idempotent symbol-marked wrap, null-proto record, error-shaped result detection, wrap-before-discovery at both assembly points), the Bedrock forced-`0.7` fix is a nice catch, the logger conditional-clear is the right dispose semantics, and the test suite covers the hard paths (TTL expiry with fake timers, dead-research extraction, corrective re-asks).

Four substantive findings before this ships, ranked — 1 and 2 are correctness bugs in leashed mode, 3 and 4 break the policy layer in exactly the multi-tenant host scenario this runtime is being built for:

1. **Leashed finalization depends on `stopReason === "completed"`, which the chat-completions (LiteLLM) loop never sets** → endless `in_progress` on that provider path (inline comment on `isolatedAgentRunner.ts:852`).
2. **Resume history carries no tool activity** → a waste-tripped leg resumes blind and immediately re-trips; ping-pong until TTL (inline on `isolatedAgentRunner.ts:567`).
3. **Open-handle conflict is global across sessions** → concurrent conversations refuse each other with handles they can't see (inline on `agentToolRegistrar.ts:284`).
4. **Depth check reads the host instance context, not the execution context** → with the default shared tool registry, nested worker delegation never hits the depth limit (inline on `agentToolRegistrar.ts:275`).

Also worth deciding explicitly: `stream()` has no `beginDelegationTurn` wiring, so per-turn caps and depth withholding silently don't apply to streaming hosts (inline on `neurolink.ts:4228`).

One test-suite note: every runner test mocks `stopReason: "completed"`, which is why finding 1 doesn't show up in the suite — worth adding a case where the mock provider leaves `stopReason` undefined (the real LiteLLM behavior).

Remaining comments are smaller (serialization of the enumerable recorder, concurrent `continueAgent` races, composed extraction wall-clock, post-hoc `tool_call` events, the `toolExecutions` public shape change needing a migration note, pool capacity semantics).

pdogra1299:
Addressed the full review set (both reviews evaluated finding-by-finding). Amended in place per the single-commit policy — same commit message, force-pushed with lease. Summary below; ✅ = fixed in this revision, ⏭️ = deliberately not changed, with reasoning.

## Correctness (leashed mode)

- ✅ **`stopReason === "completed"` dependency (runner:852)** — leg-finished is now derived from the runner's own control flow (`isLegCutShort()`): a leg is resumable only when *something bounded it* (time-limit, step-cap, tool-call budget, waste trip). A generate that returns normally without a `stopReason` — the real chat-completions/LiteLLM behavior — is a finished turn, in both the leashed decision and `finalizeOutcome`. Added the suggested regression test where the mock provider leaves `stopReason` undefined.
- ✅ **Blind resume after waste trip (runner:567)** — a leg that ends without assistant text now pushes a synthetic assistant message built from `buildDelta(legRecords)` + tripped `wasteSignals`, so the resumed worker (and `askNextPlan`) sees its own prior tool activity. Test asserts the second leg's history carries it.

## Policy layer (multi-tenant hosts)

- ✅ **Global open-handle conflict (registrar:284, runner:73)** — conflicts are now scoped to **host + caller `sessionId`**: thread B is never refused for thread A's handle, and a second NeuroLink instance never sees another instance's sessions. `continueAgent`/`stopAgent` verify host ownership before acting. Callers without a stable sessionId get no conflict detection (documented) — the pool bounds cross-session concurrency. Tests: same-session refusal, cross-session pass, cross-host pass.
- ✅ **Depth read from host context (registrar:275)** — `resolveDepth` now prefers the per-call execution context (`agentDepth` stamped on the delegating worker) over the host instance context, so nested delegation through the shared tool registry actually hits `maxDepth`. Test drives the refusal via execution context.
- ✅ **`stream()` had no delegation scope (neurolink:4228/4239)** — `stream()` now enters the same `beginDelegationTurn` scope as `generate()` (re-entrant pattern transfers directly; provider stream loops start inside the call, so the ALS scope propagates into their tool executions). Caps and depth withholding now apply to streamed turns.
- ✅ **Cap consumed on refusals (registrar:300)** — the per-turn count is consumed only once a run actually starts; depth/handle/pool refusals no longer burn a delegation. Test covers refusal-then-valid-call under `maxDelegationsPerTurn: 1`.
- ✅ **Pool starvation from nested delegation (agentNetwork:284)** — nested delegations (`agentDepth > 0`) now bypass the pool: the outer delegation already holds a slot, and queueing nested work behind a full pool is the deadlock the comment describes. Nested fan-out stays bounded by outer slots × per-run step caps; `AgentNetwork` standalone delegations are always top-level and stay pooled. Rationale documented on `maxConcurrent` + RFC known-limitations; test drives a nested call through a fully-held pool.
- ✅ **`String(params)` → `"[object Object]"` (registrar:305)** — task-less objects now pass through whole (`runIsolatedAgent` JSON-stringifies structured input).
- ✅ **Pool capacity semantics (registrar:207)** — JSDoc now states global-max-wins ("raises the process-wide pool; cannot lower it — not a per-agent throttle"), and raising the capacity drains already-queued waiters (the nitpick).

## Session lifecycle

- ✅ **Concurrent `continueAgent` race (runner:915/948)** — `session.running` guard; a second `continueAgent` (and `stopAgent`) on an in-flight handle throws a clear "leg in progress" error instead of racing worker state. Test covers both.
- ✅ **Tombstones never evicted (runner:776)** — second-stage grace timer (one more TTL period) evicts unretrieved tombstones; retrieval-once semantics unchanged. Test advances fake timers past both stages.
- ✅ **`toolExecutions` inconsistency across terminal paths (runner:722)** — terminal outcomes now return the whole run's records (all legs), matching `stopAgent`/tombstone; `in_progress` legs keep per-leg records. Test covers a two-leg run.
- ✅ **Unbounded composed extraction wall-clock (runner:326)** — new `extraction.totalTimeoutMs` phase deadline (default `(maxRetries + 1) × timeoutMs`), enforced across attempts with per-attempt timeouts clamped to the remainder; JSDoc spells out the ceiling arithmetic including the leashed next-plan ask.

## Data shape & hardening

- ✅ **Pool/fallback paths narrowed records (neurolink:5557)** — both internal result paths now pass the full `ToolExecutionRecord` fields through (with the legacy `executionTime`/`success` fields alongside), and the converters preserve `resultText` on round-trip — no more `{params: {}, resultText: "undefined"}` reconstruction. Round-trip test added.
- ✅ **`transformToolExecutions` dropped `resultText` (transformationUtils:93)** — output extraction now includes `resultText` (switched to `??` so falsy-but-real outputs survive too).
- ✅ **Recorder serialization risk (recorder:225)** — internal state moved to ES `#private` fields; the recorder stringifies as `{}` while the enumerable options key keeps surviving spreads. Test asserts an options dump contains no captured payloads.
- ✅ **Async `onRecord` rejections (recorder:129)** — wrapped in `Promise.resolve(...).catch()`; public type widened to allow async listeners. Test added.
- ✅ **Typed abort + stream guard (agent.ts:158)** — `ErrorFactory.aborted(...)` and the same pre-abort guard at the top of `stream()`. Tests for both.
- ✅ **Voice agent clears shared sink unconditionally (logger:101)** — `realtimeVoiceAgent` now passes its own emitter to `clearEventEmitter(ifEmitter)`.
- ✅ **`toolCalls` mapping drift (autoresearch outside-diff + nitpick)** — shared `toolExecutionRecordsToToolCalls()` helper (with the `toolName` → legacy `name` fallback) used by both task executors.
- ✅ **`Object.keys` on non-object (agentNetwork nitpick)** — guarded.
- ✅ **Post-hoc `tool_call`/`tool_result` events (runner:435)** — documented on `AgentRunEvent` ("both fire post-execution; accounting events, not live signals"); a pre-execution hook on the recorder wrapper is noted as the follow-up in the RFC.
- ✅ **Migration note for the `toolExecutions` shape change (generate.ts:786)** — field-mapping table added to the RFC (§6b) and the PR description. Deliberately **not** a `BREAKING CHANGE:` semantic-release footer since that forces a major bump — maintainer's call if v11 is preferred; happy to add it.

## Deliberately unchanged

- ⏭️ **Stream-vs-generate record shape (googleAiStudio:1421, googleVertex:3919)** — this is the intended contract, uniformly across every provider: `GenerateResult.toolExecutions` is the new `ToolExecutionRecord[]`; `StreamResult.toolExecutions` keeps the legacy summary shape on all providers. Migrating the stream-side public shape is a separate surface change (typed as `ToolExecutionSummary`/legacy across the stream stack) and out of scope here; called out in RFC §6b.
- ⏭️ **Worker log bridge receives all process logs (neurolink:16815)** — inherent to the process-global logger singleton: listening on the worker's own emitter would receive nothing once the host sink is restored, and not restoring it silences the host. The real fix is per-instance logger routing (follow-up in RFC §6c); the limitation is now documented loudly on `WorkerInstanceOptions.onLog`. On the nondeterminism point: `createWorkerInstance` always restores the host emitter synchronously after construction, so concurrent worker creation converges on the host as the active sink.
- ⏭️ **"Unused `extractToolInformation` return shard" (GenerationHandler nitpick)** — not unused: it remains the fallback input to `resolveToolExecutionRecords` for provider-executed tools the recorder can't wrap.
- ⏭️ **Zod 4 issue types (structuredRecovery nitpick)** — `z.ZodSchema`/`error.issues` match the repo's current zod usage; no behavioral gap.

Verification: `tsc --strict` clean, lint 0 errors, 94 tests in the agent-runtime suites (14 new), full vitest sweep green except the two pre-existing failures (modelPool stream-fallback timeout, envGuard `process.exit` under bare `vitest run` — both reproduce on the base branch), full build + no-API regression suites (`anthropic-tools-policy`, `mcp`, `mcp:bash`, `mcp:limits`, `mcp:spans`) pass.

src/lib/agent/isolatedAgentRunner.ts
Pdogra2520: ✅ Verified fixed in f69f322 — early-ended legs now push a synthetic assistant message built from `buildDelta(legRecords)` + tripped waste signals, so the resumed worker (and `askNextPlan`) sees its own prior tool activity. Covered by test *"a waste-tripped leg leaves a synthetic assistant message so the resume is not blind"*.

src/lib/agent/isolatedAgentRunner.ts
Pdogra2520: ✅ Verified fixed in f69f322 — `isLegCutShort()` now derives 'unfinished' from the leg's own control flow (time-limit / step-cap / tool budget / waste trip) and `finalizeOutcome` treats a missing stopReason as a clean finish, exactly the robust variant. Covered by test *"a provider that never sets stopReason (chat-completions/LiteLLM) still completes"*.

src/lib/agent/agentToolRegistrar.ts
Pdogra2520: ✅ Verified fixed in f69f322 — `hasOpenIsolatedAgentHandle(host, definitionId, callerSessionId)` refuses only when the SAME host + caller sessionId owns the open handle; no sessionId ⇒ no conflict detection (pool bounds cross-session concurrency); `continueAgent`/`stopAgent` are host-owned too. Tests: *"refuses only the SAME host + sessionId; other sessions proceed"*, *"a handle on a DIFFERENT host never causes a refusal"*, *"handles are owned by their host…"*.

src/lib/agent/agentToolRegistrar.ts
Pdogra2520: ✅ Verified fixed in f69f322 — `resolveDepth` prefers the execution context (`contextRecord.agentDepth`) over the host instance context, so shared-registry worker delegation now hits the limit. Bonus: the per-turn count is consumed only when a run actually starts, so depth/handle/pool refusals no longer burn a delegation. Test: *"a depth-refused call leaves the cap intact for a later valid call"* + the worker `agentDepth + 1` test.

src/lib/neurolink.ts
Pdogra2520: ✅ Verified fixed in f69f322 — `stream()` now enters the same re-entrant `beginDelegationTurn` scope as `generate()` (caps count against the streamed turn, depth-limited tools withheld; ALS propagates into the provider stream loops). Covered in the expanded agentDelegation suite + *"an already-aborted parent stops stream() before any model call too"*.

src/lib/agent/isolatedAgentRunner.ts
Pdogra2520: ✅ Verified fixed in f69f322 — `session.running` in-flight guard: a second `continueAgent` on the same handle refuses instead of racing, and `stopAgent` refuses mid-leg (pointing at the run's abortSignal) rather than disposing the worker under a live generate. Test: *"refuses a second concurrent continueAgent on the same handle"*.

src/lib/core/toolExecutionRecorder.ts
Pdogra2520: ✅ Verified fixed in f69f322 — recorder state moved to ES `#private` fields (stringifies as `{}` while the enumerable options key keeps surviving spreads); async `onRecord` rejections are swallowed too. Tests: *"serializes as {} — captured payloads never ride along with options dumps"*, *"async onRecord rejections are swallowed too"*.

src/lib/agent/isolatedAgentRunner.ts
Pdogra2520: ✅ Verified fixed in f69f322 — `extraction.totalTimeoutMs` phase deadline (default `(maxRetries+1)×timeoutMs`), per-attempt timeout clamped to the remaining budget, and the JSDoc spells out the composed ceiling arithmetic including the ~20s next-plan ask. Test: *"extraction phase deadline (totalTimeoutMs) bounds all attempts"*.

src/lib/agent/isolatedAgentRunner.ts
Pdogra2520: ✅ Addressed in f69f322 at the doc level, as suggested — `AgentRunEvent` JSDoc now states both events fire post-execution, and RFC §6c tracks the pre-execution recorder hook as the extension point for live in-flight status.

src/lib/types/generate.ts
Pdogra2520: ✅ Addressed in f69f322 — RFC §6b carries the full old→new field-mapping table (`name→toolName`, `input→params`, `output→resultText`, `duration→durationMs`, new `isError`/`startedAt`) and documents that `StreamResult.toolExecutions` deliberately keeps the legacy shape. Consider echoing one line of this in the release notes when it ships.

src/lib/agent/agentToolRegistrar.ts
Pdogra2520: ✅ Addressed in f69f322 — semantics now documented ("raises only — never lowers; not a per-agent throttle"), queued waiters are granted on a capacity raise, and nested delegations bypass the pool (removing the outer-holds-slot deadlock) with the resulting bound documented in RFC §6c.

src/lib/agent/agentNetwork.ts
```diff
@@ -227,27 +229,61 @@ export class AgentNetwork {
...
+            const result = await capturedAgent.execute(input);
+            return {
+              agentId: capturedId,
+              content: result.content,
+              status: result.status,
+              error: result.error,
+            };
+          } finally {
+            release();
+          }
```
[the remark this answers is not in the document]
Pdogra2520: Verified against f69f322: the deadlock class is removed at the registrar level — nested delegations (depth > 0) bypass the pool since the outer call already holds a slot — and RFC §6c documents the resulting bound (outer slots × per-run step caps). AgentNetwork standalone delegations stay top-level and pooled by design. Residual: a network agent whose own generate delegates via registerAgentTool enters at depth 0 and takes a second slot — that composition degrades to queue-timeout refusals (with recovery text), not a hard deadlock. Reasonable follow-up: stamp `agentDepth` in `buildAgentTools`' execution context so composed inner delegations take the nested path.

src/lib/neurolink.ts
```diff
@@ -16735,6 +16732,177 @@ Current user's request: ${currentInput}`;
...
+          // Log-bridge listener errors never disrupt the worker.
+        }
+      };
+      hostEmitter.on("log-event", forward);
+      const originalDispose = worker.dispose.bind(worker);
+      worker.dispose = async () => {
+        hostEmitter.off("log-event", forward);
+        await originalDispose();
+      };
+    }
```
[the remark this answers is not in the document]
Pdogra2520: Verified against f69f322: addressed as a documented limitation rather than a behavior change — `WorkerInstanceOptions.onLog` JSDoc now states the bridge receives ALL process log events stamped with the bridge's tag ("identifies which bridge forwarded the event, not which instance emitted it"), and RFC §6c tracks per-instance logger routing as the real fix. Given the logger is process-global with a single active emitter, that's the honest interim state.

src/lib/providers/googleAiStudio.ts
```diff
@@ -1414,7 +1415,10 @@ export class GoogleAIStudioProvider extends BaseProvider {
             },
             responseTime,
             toolsUsed: allToolCalls.map((tc) => tc.toolName),
-            toolExecutions: transformToolExecutions(toolExecutions),
+            toolExecutions: resolveToolExecutionRecords(
+              options,
+              toolExecutions,
+            ),
```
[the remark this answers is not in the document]
Pdogra2520: Verified against f69f322: this is the intended contract change, not drift — `GenerateResult.toolExecutions` is now `ToolExecutionRecord[]` uniformly (RFC §6b has the full old→new migration table), while `StreamResult.toolExecutions` deliberately keeps the legacy summaries (documented in the same section). `transformToolExecutions` was also made record-aware (`resultText`/`durationMs` fallbacks) so the stream path is lossless when fed records, and the new `toolExecutionRecordsToToolCalls` gives task executors a single conversion source.

Pdogra2520:
## Re-review of f69f322 — all findings verified fixed ✅

Pulled the branch and verified every thread from both review rounds against the code:

**My 11 findings — all fixed, each with a dedicated test:**
| # | Finding | Fix | Test |
|---|---|---|---|
| 1 | LiteLLM leashed runs never finalize (`stopReason` dependence) | `isLegCutShort()` — control-flow derivation | ✅ *provider that never sets stopReason still completes* |
| 2 | Waste-tripped legs resume blind → re-trip ping-pong | synthetic assistant msg from `buildDelta` + waste signals | ✅ *resume is not blind* |
| 3 | Open-handle conflict global across sessions | host + callerSessionId scoping; host-owned continue/stop | ✅ ×3 |
| 4 | Depth read from host ctx, not execution ctx | `resolveDepth` prefers execution context; cap consumed only on real start | ✅ |
| 5 | `stream()` missing delegation-turn scope | same re-entrant `beginDelegationTurn` wiring | ✅ |
| 6 | Concurrent `continueAgent` race | `session.running` guard (continue AND stop) | ✅ |
| 7 | Recorder payloads ride along with options dumps | ES `#private` fields → stringifies as `{}` | ✅ |
| 8 | Unbounded composed extraction wall-clock | `totalTimeoutMs` phase deadline + documented ceiling arithmetic | ✅ |
| 9 | Post-execution `tool_call` events | documented + RFC §6c pre-exec hook follow-up | doc |
| 10 | `toolExecutions` shape change unannounced | RFC §6b migration table (old→new field map) | doc |
| 11 | Pool `Math.max` ratchet semantics | documented; waiters granted on raise; nested bypass kills the deadlock | ✅ |

**The 3 open CodeRabbit threads** are addressed too (replies on each): nested pool-bypass + RFC §6c for the starvation class, the log-bridge firehose documented as a known limitation with per-instance routing as follow-up, and the generate/stream shape split being the documented intended contract (§6b).

Bonus fixes beyond the review: terminal outcomes now carry ALL legs' records, tombstone grace-period eviction, per-turn cap not burned by refusals, async `onRecord` rejection swallowing.

CI fully green. I couldn't resolve threads (no maintainer rights on this token) — each has a ✅ verification reply, so they're one click each. **LGTM from my side — nothing further blocking.**

src/lib/agent/isolatedAgentRunner.ts
```diff
@@ -0,0 +1,1143 @@
...
+  });
+  // Second stage: evict the tombstone if nobody ever retrieves it, so an
+  // abandoned handle doesn't retain its records/history forever.
+  const graceTimer = setTimeout(() => {
+    if (sessions.get(session.handle) === session) {
+      sessions.delete(session.handle);
+    }
+  }, session.options.handleTtlMs ?? DEFAULT_HANDLE_TTL_MS);
+  graceTimer.unref?.();
+}
```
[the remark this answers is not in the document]
pdogra1299: ✅ Fixed in be780f4e — `expireSession` now (a) returns early when the session is no longer in the registry or already tombstoned, (b) returns without disposing when `session.running` (the fired-before-clear window; the in-flight leg's completion path re-registers with a fresh TTL or finalizes and cleans up), and (c) sets the tombstone BEFORE the async dispose — so a racing `continueAgent`/`stopAgent` sees the tombstone synchronously and returns it instead of starting a leg on a worker being disposed. Test: *"an expired handle tombstones BEFORE disposal — a racing continueAgent gets the outcome, never a disposing worker"* (slow-dispose mock, fake timers). The sub-millisecond fired-callback window for the `running` branch itself isn't reachable black-box (the timer clear at `continueAgent` entry is synchronous), so that branch is covered by the guard + review rather than a test.

src/lib/providers/googleNativeGemini3.ts
```diff
@@ -673,8 +676,18 @@ export function buildNativeConfig(
...
   toolsConfig?: NativeToolsConfig,
 ): Record<string, unknown> {
+  // Registry-driven sampling strip applied for uniformity with every other
+  // provider path (inert for current Gemini models).
+  const samplingParams = resolveSamplingParams(
+    "google-ai",
+    options.model,
+    { temperature: options.temperature ?? 1.0 }, // Gemini 3 requires 1.0 for tool calling
+    "googleAiStudio.buildNativeConfig",
+  );
```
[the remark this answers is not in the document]
pdogra1299: ✅ Fixed in be780f4e — both AI Studio call sites now pass `model: modelName` (the effective `options.model || this.modelName`) into `buildNativeConfig`, so the registry sampling check sees the model actually used. One correction on the stated consequence: an undefined model fails OPEN in `modelSupportsSamplingParams` (params kept), so Gemini 3's `temperature: 1.0` was never at risk of being stripped — the actual gap was the inverse (a sampling-rejecting model configured as the instance default would NOT get stripped). Same fix either way.

docs/plans/2026-07-27-isolated-agent-runner-rfc.md
```diff
@@ -0,0 +1,258 @@
...
+  (set on the worker for EVERY tool call, incl. a caller-supplied `sessionId`; the run id
+  is the default), `onEvent` lifecycle stream, `leg` (leashed mode), `handleTtlMs`
+  (default 10 min), `waste` thresholds, and `capture` bounds for the run's execution
+  records.
+- Returns `AgentRunOutcome` — `status` (`completed | partial | in_progress |
+insufficient_data | error`), `data` (schema-valid when `extraction.schema` given, or a
+  mechanical digest under the delivery guarantee), `content` (research narrative),
+  `stopReason` (honest), `toolExecutions` (this leg's `ToolExecutionRecord[]`), `usage`,
+  `durationMs`, `extractionSource`/`extractionError` diagnostics, and in leashed mode
+  `handle`, `leg`, `delta`, `nextPlan`, `wasteSignals`, `budget`.
```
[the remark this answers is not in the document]
pdogra1299: ✅ Fixed in be780f4e — RFC §4.2 now lists `extraction.totalTimeoutMs` (with the `(maxRetries + 1) × timeoutMs` default, called out as the one number for outer-ceiling arithmetic) and describes `toolExecutions` as the WHOLE run's records on terminal outcomes / per-leg on `in_progress`. The `AgentRunOutcome.toolExecutions` JSDoc got the same correction.

src/lib/agent/agentNetwork.ts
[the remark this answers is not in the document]
pdogra1299: ✅ Fixed properly in be780f4e, request-scoped as you specified — no instance-level `setToolContext` involved:

- `AgentDelegationTurnState` now carries `depth`; `beginDelegationTurn` seeds it at scope entry.
- `AgentNetwork.buildAgentTools()` wraps each `capturedAgent.execute(...)` in the new `runWithNestedDelegationDepth()` — an AsyncLocalStorage scope that SHARES the turn's counters but runs one level deeper.
- `executeDelegation` resolves depth as: execution context (`agentDepth`) → ALS turn-scope depth → host fallback.

So the composed `AgentNetwork → capturedAgent.execute() → registerAgentTool()` path now enters at depth ≥ 1: it takes the nested pool path (no queueing behind the slot its own caller holds — the liveness failure is gone, not just timeout-broken) and it hits `maxDepth`. New test covers exactly this composition: *"composed path: runWithNestedDelegationDepth makes inner delegations nested"* — inner delegation completes while the pool's only slot is held by the outer, and two levels deeper is refused at `maxDepth` with recovery text.

src/lib/neurolink.ts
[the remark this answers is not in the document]
pdogra1299: Follow-up issue created as offered: #1236 (per-instance logger routing for worker log-bridge attribution) — linked from RFC §6c. Keeping this thread unresolved until that lands is fine by me.

pdogra1299:
## Round 2 — all new findings addressed in `be780f4e` (amended in place, still a single commit)

- ✅ **`expireSession` in-flight race** — registry-membership + `running` guards, and the tombstone is now set BEFORE the async dispose so racing `continueAgent`/`stopAgent` calls get the outcome instead of a disposing worker. Covered by a slow-dispose fake-timer test.
- ✅ **`buildNativeConfig` effective model** — both AI Studio call sites pass `model: modelName` so the registry sampling check sees the model actually used (note: the gap was fail-open on instance-default models, not stripping Gemini 3's 1.0).
- ✅ **Composed `AgentNetwork → registerAgentTool` liveness** — request-scoped depth via the ALS turn state (`runWithNestedDelegationDepth` shares counters, one level deeper); the composed path now takes the nested pool path and hits `maxDepth`. New composition test included. No instance-level `setToolContext` involved.
- ✅ **RFC staleness** — §4.2 documents `extraction.totalTimeoutMs` and the terminal-vs-in_progress `toolExecutions` semantics; matching JSDoc fix on `AgentRunOutcome`.
- 📋 **Logger-routing follow-up** — created #1236 as the trackable item behind the acknowledged log-bridge limitation.

Verification unchanged and green: `tsc --strict` clean, lint 0 errors, 96 agent-runtime tests (2 new), full build + publint, no-API regression suites pass.

Tara-ag:
## Autonomous review summary (post-merge)

I attempted to perform a live review of this PR, but it has already been merged into `release`, so GitHub does not allow opening a new pending review. I reviewed the diff, the existing review threads, and the current state of the key files instead.

### What I checked
- `src/lib/agent/isolatedAgentRunner.ts` — leashed legs, TTL expiry, tombstoning, waste signatures, abort chaining, extraction pass.
- `src/lib/agent/agentToolRegistrar.ts` — per-turn caps, depth limits, process-wide pool, open-handle scoping.
- `src/lib/core/toolExecutionRecorder.ts` — bounded capture, serialization safety, idempotent wrapping, listener-error swallowing.
- Existing review threads (31 total) — most were resolved in commits `f69f322` and `be780f4e`.

### Findings
- **No new blocking security, correctness, or architectural issues were identified** in the merged code.
- The previously raised concerns about `stopReason` inference, blind resume after waste trips, open-handle session scoping, depth resolution, and TTL/disposal races all appear to have been addressed with regression tests.
- The `stream()` / `beginDelegationTurn` wiring concern noted at `src/lib/neurolink.ts:4228` was reported fixed in `f69f322` with expanded `agentDelegation` tests.

### One remaining known limitation (not a blocker)
The worker log bridge in `createWorkerInstance` still receives **all** process log events because the logger is process-global. This is now explicitly documented in the `WorkerInstanceOptions.onLog` JSDoc and tracked as follow-up issue **#1236** (per-instance logger routing). Since it is acknowledged and tracked, it does not warrant blocking the merged PR.

### Recommendation
Approve as-is for the merged state. Any further work on the log-bridge attribution should land via #1236.

---
*Review performed by Yama autonomous review agent. Could not post inline comments because the PR is already merged.*

PR #1280, opened 2026-08-05
fix(context): pair tool calls by id instead of adjacency

pdogra1299:
## Description

`repairToolPairs` matched `tool_call` to `tool_result` by **adjacency**. A single agent step with parallel tool calls is persisted as every call followed by every result (`flushPendingToolExecutions`), so `tool_call` is routinely followed by another `tool_call` in perfectly healthy history. The repair read that as an orphan and injected a `[Tool result unavailable]` placeholder over a result that was present two entries later.

Measured on the exact shape the Redis manager writes:

```
in : assistant  tool_call:read_file  tool_call:grep  tool_result:read_file  tool_result:grep
out: assistant  tool_call:read_file  tool_result:read_file(SYNTHETIC)  tool_call:grep  tool_result:read_file  tool_result:grep
     2 calls -> 3 results
```

This ran after **every** compaction, so any session using parallel tool calls was corrupted the moment it compacted.

Root cause: `ChatMessage` had no `toolCallId`. The id was computed in `flushPendingToolExecutions`, used to build a name lookup, then discarded — leaving position as the only signal, and position is wrong for every parallel batch.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Motivation and Context

Providers hard-reject an unmatched `tool_call`/`tool_result`, so a corrupted batch turns into a failed turn. Two further paths could emit orphans that nothing caught:

- `buildContextFromPointer` slices history at the summary pointer with no repair on read
- the summary pointer itself could land **inside** a tool batch, leaving the recent window opening on an orphaned `tool_result`

## Changes

- `toolCallId` added to `ChatMessage` (optional — backward compatible) and persisted on both sides
- `repairToolPairs` rewritten batch- and ID-aware, with positional fallback **scoped to the batch** for pre-existing sessions
- repair now also runs on the **read** path, covering pointer slices and sessions interrupted mid-tool-batch
- `snapSplitToBatchBoundary` keeps the summary pointer off a batch boundary, applied in both summarizers (including after the clamp in `structuredSummarizer`, which could otherwise pull the boundary back in)
- duplicate `toolCallId` results are dropped rather than over-pairing a call

## Testing

New no-API suite `pnpm run test:tool-pairing` — **15/15**. Covers parallel batches (id and legacy), five-wide batches, out-of-order results, genuine orphans in both directions, duplicate ids, empty-content results, multiple independent batches, and randomized split points.

[...]

PR #1281, opened 2026-08-05
fix(context): count tool args and target history budget when compacting

pdogra1299:
## Description

Two accounting defects that together suppressed the compaction pipeline.

**1. `args` were invisible to every token estimator.** `estimateMessageTokens` stringified only `message.content`. A `tool_call` is persisted with an **empty** `content` and its entire payload in `args`, so a Write/Edit call was scored as free:

```
Write tool_call carrying 39,000 chars in args  (~9,750 real tokens)
  compactor  estimateMessagesTokens ->    28 tokens
  summarizer estimateTokens         ->     0 tokens
```

A ~348x under-count on the single largest source of context growth in an agentic session. The budget checker, the compaction trigger and the summarization threshold were all blind to it.

**2. The compactor was aimed at the wrong number.** All four call sites passed `availableInputTokens` — the budget for system + prompt + tools + files + history — while every stage gate in `ContextCompactor` measures **history alone**:

```ts
if (this.config.enablePrune && estimateMessagesTokens(currentMessages) > targetTokens)
```

So compaction only engaged once history *by itself* exceeded the entire window. With ~40K of MCP tool definitions against a 128K window, a request could sit at 125% of budget with history at 95K and every stage would no-op, reporting `compacted: false` and falling through to blunt emergency truncation.

The proof it was unintentional sits 40 lines away: `emergencyTruncation.ts:35` already derives the history budget correctly. The emergency path knew the formula; the main path did not.

## Type of Change

## Changes

- `estimateMessageTokens` now counts `args` and `result.error`
  - `result` itself is deliberately **not** counted: `result.result` is re-hydrated from `content` at read time, so counting both double-bills the same bytes
  - serialization guarded against `RangeError` on oversized values (a tool-arg blob is exactly the shape that hits V8's string cap)
- `SummarizationEngine`'s divergent content-only estimator collapsed into the shared one, so the summarization threshold now agrees with the budget checker
- `resolveHistoryBudget()` extracted into `budgetChecker.ts` and wired into all four `compact()` call sites, including the reactive-recovery path
- fixed overhead exceeding the window now raises `ContextBudgetExceededError` naming tool-definition size, instead of compacting to an empty history

## Testing

New no-API suite `pnpm run test:token-accounting` — **10/10**. Verified to fail loudly (`✗`, exit 1) on a broken assertion.

- `tsc --noEmit --strict`: 0 errors
- `eslint`: 0 errors
- `test:tool-pairing`: 15/15 still green
- `test:context`: byte-identical failure set to `release`

## Notes for reviewers

**Expect compaction to fire more often after this.** It is currently suppressed by under-counting; that is the bug, not a feature. Worth watching `compaction.triggered` rates on first deploy.

Stacked on #1280 — **must not merge before it.** Correct accounting means compaction runs far more frequently, which would amplify the parallel-tool-batch corruption that #1280 fixes.

One thing deliberately left alone: the `lastCompactionMessageCount` watermark keys on message count rather than tokens. On close reading it is sound for the turn-by-turn flow (count grows each turn, so the gate clears), so changing it would be risk without a demonstrated failure. Noted as a follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved conversation context management by accounting for system prompts, current prompts, tools, attachments, and recovery overhead.
  * Enhanced token estimates to include tool-call arguments and tool-result errors.
  * Prevented context processing from exceeding the model’s available input capacity.
  * Added clearer handling when fixed prompt and tool overhead exceeds the context window.

* **Tests**
  * Added coverage for token accounting, history budgets, tool payloads, oversized values, and circular data.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "test:litellm-parity": "npx tsx test/continuous-test-suite-litellm-parity.ts",
     "test:memory": "npx tsx test/continuous-test-suite-memory.ts",
     "test:tool-pairing": "npx tsx test/continuous-test-suite-tool-pairing.ts",
+    "test:token-accounting": "npx tsx test/continuous-test-suite-token-accounting.ts",
     "test:middleware": "npx tsx test/continuous-test-suite-middleware.ts",
     "test:observability": "npx tsx test/continuous-test-suite-observability.ts",
     "test:ppt": "npx tsx test/continuous-test-suite-ppt.ts",
--- a/src/lib/context/budgetChecker.ts
+++ b/src/lib/context/budgetChecker.ts
@@ -24,8 +24,51 @@ import { getActiveTraceContext } from "../telemetry/traceContext.js";
 /** Default compaction threshold (80% of available input) */
 const DEFAULT_COMPACTION_THRESHOLD = 0.8;
 
+/**
+ * Fraction of the derived history budget actually handed to the compactor.
+ * Estimation is char-based and approximate, so the compactor aims slightly
+ * under the true ceiling rather than exactly at it.
+ */
+const HISTORY_BUDGET_SAFETY_FACTOR = 0.95;
+
 /** Estimated tokens per tool definition */
 const TOKENS_PER_TOOL_DEFINITION = 200;
+
+/**
+ * Tokens the CONVERSATION HISTORY may occupy, i.e. the model's available input
+ * space minus everything that rides alongside it (system prompt, current
+ * prompt, tool definitions, file attachments).
+ *
+ * This is the number the compactor must target. Passing it the undeducted
+ * `availableInputTokens` made every stage gate compare history-only tokens
+ * against the WHOLE budget, so compaction only engaged once history alone
+ * exceeded the entire window — with a large MCP tool set a request could sit
+ * far over budget while the compactor reported "nothing to do" and fell through
+ * to emergency truncation.
+ *
+ * Returns 0 when the fixed overhead already exceeds the window; callers must
+ * treat that as unrecoverable rather than compacting to an empty history.
+ */
+export function resolveHistoryBudget(result: BudgetCheckResult): number {
+  const { availableInputTokens, breakdown } = result;
+  if (!breakdown) {
+    return Math.max(
+      0,
+      Math.floor(availableInputTokens * HISTORY_BUDGET_SAFETY_FACTOR),
+    );
+  }
+  const overhead =
+    breakdown.systemPrompt +
+    breakdown.currentPrompt +
+    breakdown.toolDefinitions +
+    breakdown.fileAttachments;
+  return Math.max(
+    0,
+    Math.floor(
+      (availableInputTokens - overhead) * HISTORY_BUDGET_SAFETY_FACTOR,
+    ),
+  );
+}
 
 /**
  * Check whether a request fits within the model's context budget.
--- a/src/lib/context/summarizationEngine.ts
+++ b/src/lib/context/summarizationEngine.ts
@@ -10,13 +10,16 @@ import type {
   ConversationMemoryConfig,
   SessionMemory,
 } from "../types/index.js";
-import { TokenUtils } from "../constants/tokens.js";
 import {
   buildContextFromPointer,
   generateSummary,
 } from "../utils/conversationMemory.js";
 import { RECENT_MESSAGES_RATIO } from "../config/conversationMemory.js";
 import { snapSplitToBatchBoundary } from "./toolPairRepair.js";
+import {
+  estimateMessageTokens,
+  estimateMessagesTokens,
+} from "../utils/tokenEstimation.js";
 import { withSpan } from "../telemetry/withSpan.js";
 import { tracers } from "../telemetry/tracers.js";
 import { logger } from "../utils/logger.js";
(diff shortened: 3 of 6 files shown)


murdore:
Not mergeable yet, and I want to flag *why* the checks look the way they do, because the picture is misleading.

## Base is not `release`

This targets `fix/tool-pair-identity` (#1280), so it can only merge after everything below it. Order for the stack:

```
release
 └─ #1280  fix/tool-pair-identity
     └─ #1281  fix/token-accounting
         └─ #1282  fix/cache-aware-compaction
             └─ #1283  fix/loop-guard-parity
                 └─ #1285  fix/session-persistence
                     └─ #1286  fix/redis-append-only
                         └─ #1287  fix/loop-guard-completion
```

This PR is **#2 of 7**.

## The Single Commit failure is a stacking artifact, not a real violation

`single-commit-enforcement.yml` triggers on `push` with `branches-ignore: [release, main, develop]`, so it runs on the branch and measures against `release` rather than against this PR's own base. `fix/tool-pair-identity`… this branch is 2 commits ahead of `release` because it carries its parents'. Against its **own** base it is a single commit, which is what the policy actually asks for. It should clear itself as the parents merge.

## The bigger issue: almost no CI has run here

Three checks are present on this PR. Release-targeting PRs get **fifteen**. The difference is not incidental — the workflows are branch-gated:

| workflow | `pull_request: branches:` |
|---|---|
| `ci.yml` (`test (20)`, `build-check`, lint, format) | `[release]` |
| `yama-review.yml` | `[release]` |
| `copilot-review.yml` | `[release, main]` |

So on this PR **nothing has been built, no tests have run, and neither AI reviewer has looked at it.** Only CodeRabbit, SentinelOne and the push-triggered commit check fire. The absence of review comments here reflects the absence of a review, not a clean bill of health.

Across #1281, #1282, #1283, #1285, #1286 and #1287 that is roughly **3,900 added lines that have never been compiled or tested by CI**.

Nothing to do on this PR specifically — it resolves itself once the parents land and the base becomes `release`. Worth knowing before anyone reads the green ticks as coverage. If you want real signal sooner, temporarily retargeting to `release` (or adding the stack branches to `ci.yml`'s filter) would run the full gate.

pdogra1299:
@murdore — confirmed both points, verified rather than taken on trust.

**Single Commit failure is a stacking artifact.** `single-commit-enforcement.yml` fires on `push` with `branches-ignore: [release, main, develop]`, and in the push branch of that script `BASE_BRANCH` is hard-coded to `origin/release` — so it measures against `release`, not the PR's own base. Against its own base every branch in the stack is exactly one commit:

```
$ git rev-list --count HEAD ^<own-base>
fix/token-accounting        1
fix/cache-aware-compaction  1
fix/loop-guard-parity       1
fix/session-persistence     1
fix/redis-append-only       1
fix/loop-guard-completion   1
```

Nothing to squash. It clears itself as parents land.

**The CI gap is real and I am not treating the green ticks as coverage.** Verified the gating you listed:

| workflow | `pull_request: branches:` |
|---|---|
| `ci.yml` | `[release]` |
| `yama-review.yml` | `[release]` |
| `copilot-review.yml` | `[release, main]` |

So on this PR nothing has been built or tested by CI, and that is the accurate reading.

What has run, locally on the **stack tip** (so every branch's code compiled together) after the latest rebase:

- `pnpm run check` → 4961 files, **0 errors, 0 warnings**
- all nine new suites → **95 tests, 0 failures, 0 skips**
  (`tool-pairing` 18, `token-accounting` 10, `step-guard` 6, `loop-guard-core` 11, `openai-compat-guard` 9, `anthropic-guard` 10, `tool-storage-parity` 10, `redis-append-only` 10, `gemini-guard` 11)

That is local signal, not CI signal, and I would rather say so plainly than let the tick count imply otherwise. Merging bottom-up closes the gap for real: each PR's base becomes `release` as its parent lands, at which point it picks up the full fifteen-check gate before it can merge — so no branch reaches `release` without having been through it.

pdogra1299:
#1280 is merged (`a0f49e40`), so the stack has been rebased onto the new `release` (now at 10.10.2 — it had moved two releases since these branches were cut). Still one commit per branch.

**#1281 was `CONFLICTING` after the merge** — its branch still carried the pre-squash copy of #1280's commit. Replanted with `git rebase --onto origin/release 8344a885` so only its own commit survives; now `MERGEABLE` against `release`.

**One real conflict, in `providers/anthropic/client.ts`.** `release` had gained the structured-output buffering from the `final_result` work (`bufferedText` / `finalResultText`) at the same point where #1283 adds `estimateAnthropicFixedOverhead`. Both-added, both wanted — kept both. Verified afterwards that the release-side path is still fully wired (`bufferedText` accumulates at the text-delta branch, `finalResultText` is set from the `final_result` tool call, and the buffered turn is still delivered at the end of the loop), not just that it compiles.

## The build was broken, and `pnpm run check` did not catch it

Worth recording since it is exactly the CI gap @murdore flagged. Building the tip surfaced two errors in #1286's code that the typechecker passes clean:

```
redisConversationMemoryManager.ts(987,7): error TS2345:
  '(string | Buffer)[]' is not assignable to 'string[]'
redisConversationMemoryManager.ts(1000,5): error TS2322:
  'number | `${number}`' is not assignable to 'number'
```

`pnpm run check` uses the main tsconfig; `build:react-hooks` compiles with `--module nodenext --moduleResolution nodenext`, which resolves the `redis` types differently and widens both return types. Fixed at the source rather than with an assertion (Rule 14): `lRange` entries are normalized via `.toString()`, `lLen` through `Number()`. Both are real defects, not just type noise — a binary-mode client would have thrown on parse, and a RESP3 string count would have concatenated where callers add.

## Verification

Each branch was built **independently**, since each merges on its own:

| branch | build | 
|---|---|
| `fix/token-accounting` | ✅ |
| `fix/cache-aware-compaction` | ✅ |
| `fix/loop-guard-parity` | ✅ |
| `fix/session-persistence` | ✅ |
| `fix/redis-append-only` | ✅ |
| `fix/loop-guard-completion` | ✅ |

On the tip: `pnpm run check` → 4939 files, **0 errors**; `pnpm run lint` → **0 errors**, 48 warnings, all pre-existing `max-lines-per-function` in files this stack does not touch (`claudeProxyRoutes.ts`, `toolRoutes.ts`, `csvProcessor.ts`); all nine suites → **95 tests, 0 failures, 0 skips**.

Still local signal, not CI signal — except on #1281, which now targets `release` and is running the full gate.

murdore (approved):

PR #1282, opened 2026-08-05
fix(context): reclaim to a low-water mark in the step budget guard

pdogra1299:
## Description

`dropOldestToolExchanges` reclaimed only back to the **firing threshold**. The next step appends an assistant turn plus its tool results, immediately re-crosses, and the guard rewrites the oldest messages again — so a long agentic run mutated its own prompt prefix on **every step**. Each mutation invalidates the Anthropic `cache_control` prefix from the edit point onward (see `anthropicCacheBreakpoints.ts`, which spends its budget caching exactly that history), converting a ~0.1x cached read into full-price input.

Measured over a 40-step simulated loop:

| Scenario | Reclaim-to-threshold | Low-water reclaim |
|---|---|---|
| Large tool outputs (stage 1 dominant) | 3/40 | 3/40 — no change |
| **Many small outputs (stage 2 dominant)** | **40/40** | **1/40** |

Each firing is a full prompt-cache invalidation, so that is a **40x reduction** in the regime where it bites.

The regime matters and is not the obvious one. With large outputs, stage 1 truncates to 2KB previews, overshoots well below the line and creates natural headroom — the bug is invisible. It only appears when outputs are **already below** the preview budget, so truncation cannot shrink them and dropping exchanges is the only lever. That is a long agentic run doing many small greps, reads and edits: the common case.

## Type of Change

- [x] Performance improvement

## Changes

- `CONTEXT_GUARD_LOW_WATER_RATIO` (0.6): the threshold decides *when* the guard fires, the low-water mark decides *how far* it reclaims
- stage 2 targets the low-water mark instead of the threshold
- firing log now reports `lowWaterTokens` and `headroomTokens` — headroom near zero means the cache is being invalidated every step

## Testing

New no-API suite `pnpm run test:step-guard` — **6/6**. Includes an explicit stage-2-dominant regression that **fails without this change and passes with it** (verified both directions), plus protected-tail and first-user-message invariants.

- `tsc --noEmit --strict`: 0 errors
- `eslint`: 0 errors
- `test:tool-pairing` 15/15, `test:token-accounting` 10/10 still green
- `test:context`: byte-identical failure set to `release`

## Notes for reviewers

Stage order is deliberately **unchanged**. Truncating before dropping preserves a preview of each output, and since both stages edit the oldest messages the cache prefix breaks at roughly the same point either way — frequency, not stage order, governs cache retention here.

Two items from the original plan were dropped after reading `anthropicCacheBreakpoints.ts` rather than implemented: breakpoint re-anchoring (rolling markers already sit at the tail, so the new prefix re-caches on the next call automatically) and stage reordering (trades information loss for a benefit that does not materialize). Both would have been no-op diff padding once hysteresis landed.

Top of the 3-PR stack: #1280 -> #1281 -> this.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved conversation compaction to reclaim sufficient step-budget headroom when limits are reached.
  * Preserved key conversation context, including the initial user message and protected recent messages.
  * Reduced unnecessary repeated compaction during extended tool-driven conversations.
  * Improved diagnostics for compaction targets and reclaimed capacity.

* **Tests**
  * Added regression coverage for step-budget behavior across short and long conversations, including tool-heavy scenarios and protected conversation content.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "test:memory": "npx tsx test/continuous-test-suite-memory.ts",
     "test:tool-pairing": "npx tsx test/continuous-test-suite-tool-pairing.ts",
     "test:token-accounting": "npx tsx test/continuous-test-suite-token-accounting.ts",
+    "test:step-guard": "npx tsx test/continuous-test-suite-step-guard.ts",
     "test:middleware": "npx tsx test/continuous-test-suite-middleware.ts",
     "test:observability": "npx tsx test/continuous-test-suite-observability.ts",
     "test:ppt": "npx tsx test/continuous-test-suite-ppt.ts",
--- a/src/lib/context/stepBudgetGuard.ts
+++ b/src/lib/context/stepBudgetGuard.ts
@@ -55,6 +55,23 @@ const MAX_CALIBRATION_RATIO = 3;
 /** Messages at the end of the conversation the guard never modifies. */
 const PROTECTED_TAIL_MESSAGES = 4;
 
+/**
+ * Fraction of the context window the guard reclaims DOWN TO once it fires.
+ *
+ * The high-water mark (`thresholdRatio`) decides *when* to act; this low-water
+ * mark decides *how far*. Reclaiming only back to the threshold meant the very
+ * next step — which appends an assistant turn plus its tool results — crossed
+ * it again, so the guard mutated the message prefix on every single step of a
+ * long agentic run. Each of those mutations invalidates the Anthropic
+ * `cache_control` prefix from the edit point onward (see
+ * anthropicCacheBreakpoints), turning a ~0.1x cached read into full-price
+ * input every step.
+ *
+ * One deeper reclaim every N steps saves the same tokens and leaves the prefix
+ * stable in between, which is what makes the cache worth having.
+ */
+const CONTEXT_GUARD_LOW_WATER_RATIO = 0.6;
+
 /** Stage-1 preview budget for an old tool output (bytes). */
 const OLD_TOOL_OUTPUT_PREVIEW_BYTES = 2_048;
 
@@ -370,18 +387,29 @@ export function createStepBudgetGuard(config: StepBudgetGuardConfig) {
       return undefined;
     }
 
+    // Reclaim down to the LOW-WATER mark, not merely back under the threshold.
+    // See CONTEXT_GUARD_LOW_WATER_RATIO: stopping at the threshold guaranteed
+    // the next step crossed it again, mutating the cached prefix every step.
+    const lowWaterTokens = Math.floor(
+      (availableInput * CONTEXT_GUARD_LOW_WATER_RATIO) / calibration,
+    );
+
     // Stage 1: shrink old tool outputs to previews.
     const stage1 = truncateOldToolOutputs([...messages]);
     let compacted = stage1.messages;
     let newEstimate =
       overheadTokens + estimateStepMessagesTokens(compacted, provider);
 
-    // Stage 2: drop oldest complete tool exchanges if still over.
+    // Stage 2: drop oldest complete tool exchanges until under the low-water
+    // mark. Stage order is deliberately unchanged — truncating first preserves
+    // a preview of each output, and since BOTH stages edit the oldest messages
+    // the cache prefix is invalidated at roughly the same point either way.
+    // Frequency, not stage order, is what governs cache retention here.
     let droppedExchanges = 0;
-    if (newEstimate > effectiveThreshold) {
+    if (newEstimate > lowWaterTokens) {
       const stage2 = dropOldestToolExchanges(
         compacted,
-        effectiveThreshold,
+        lowWaterTokens,
         overheadTokens,
         provider,
       );
@@ -401,8 +429,13 @@ export function createStepBudgetGuard(config: StepBudgetGuardConfig) {
       model,
       estimatedTokens: rawEstimate,
       thresholdTokens: effectiveThreshold,
+      lowWaterTokens,
       calibration,
       afterTokens: newEstimate,
+      // Headroom reclaimed below the firing threshold. Roughly how many further
+      // steps can run before the guard mutates the prefix again — a value near
+      // zero means the cache is being invalidated every step.
+      headroomTokens: effectiveThreshold - newEstimate,
       toolOutputsTruncated: stage1.truncated,
       exchangesDropped: droppedExchanges,
     });
--- a/src/lib/utils/tokenEstimation.ts
+++ b/src/lib/utils/tokenEstimation.ts
@@ -130,7 +130,7 @@ function serializeForEstimate(value: unknown): string {
  * Includes message framing overhead.
  *
  * Counts `content` AND `args`. A `tool_call` is persisted with an EMPTY
- * `content` and its entire payload in `args` (see flushPendingToolExecutions),
+ * `content` and its entire payload in `args` (see flushPendingToolData),
  * so a content-only estimate scored a 39 KB Write call at ~28 tokens against a
  * real cost near 9,750 — the budget checker, the compaction trigger and the
  * summarization threshold were all blind to the single largest source of
--- a/test/continuous-test-suite-step-guard.ts
+++ b/test/continuous-test-suite-step-guard.ts
@@ -1,0 +1,241 @@
+#!/usr/bin/env tsx
+import "dotenv/config";
+
+/**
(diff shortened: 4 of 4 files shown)


murdore:
Not mergeable yet, and I want to flag *why* the checks look the way they do, because the picture is misleading.

## Base is not `release`

This targets `fix/token-accounting` (#1281), so it can only merge after everything below it. Order for the stack:

```
release
 └─ #1280  fix/tool-pair-identity
     └─ #1281  fix/token-accounting
         └─ #1282  fix/cache-aware-compaction
             └─ #1283  fix/loop-guard-parity
                 └─ #1285  fix/session-persistence
                     └─ #1286  fix/redis-append-only
                         └─ #1287  fix/loop-guard-completion
```

This PR is **#3 of 7**.

## The Single Commit failure is a stacking artifact, not a real violation

`single-commit-enforcement.yml` triggers on `push` with `branches-ignore: [release, main, develop]`, so it runs on the branch and measures against `release` rather than against this PR's own base. `fix/token-accounting`… this branch is 3 commits ahead of `release` because it carries its parents'. Against its **own** base it is a single commit, which is what the policy actually asks for. It should clear itself as the parents merge.

## The bigger issue: almost no CI has run here

Three checks are present on this PR. Release-targeting PRs get **fifteen**. The difference is not incidental — the workflows are branch-gated:

| workflow | `pull_request: branches:` |
|---|---|
| `ci.yml` (`test (20)`, `build-check`, lint, format) | `[release]` |
| `yama-review.yml` | `[release]` |
| `copilot-review.yml` | `[release, main]` |

So on this PR **nothing has been built, no tests have run, and neither AI reviewer has looked at it.** Only CodeRabbit, SentinelOne and the push-triggered commit check fire. The absence of review comments here reflects the absence of a review, not a clean bill of health.

Across #1281, #1282, #1283, #1285, #1286 and #1287 that is roughly **3,900 added lines that have never been compiled or tested by CI**.

Nothing to do on this PR specifically — it resolves itself once the parents land and the base becomes `release`. Worth knowing before anyone reads the green ticks as coverage. If you want real signal sooner, temporarily retargeting to `release` (or adding the stack branches to `ci.yml`'s filter) would run the full gate.

pdogra1299:
@murdore — confirmed both points, verified rather than taken on trust.

**Single Commit failure is a stacking artifact.** `single-commit-enforcement.yml` fires on `push` with `branches-ignore: [release, main, develop]`, and in the push branch of that script `BASE_BRANCH` is hard-coded to `origin/release` — so it measures against `release`, not the PR's own base. Against its own base every branch in the stack is exactly one commit:

```
$ git rev-list --count HEAD ^<own-base>
fix/token-accounting        1
fix/cache-aware-compaction  1
fix/loop-guard-parity       1
fix/session-persistence     1
fix/redis-append-only       1
fix/loop-guard-completion   1
```

Nothing to squash. It clears itself as parents land.

**The CI gap is real and I am not treating the green ticks as coverage.** Verified the gating you listed:

| workflow | `pull_request: branches:` |
|---|---|
| `ci.yml` | `[release]` |
| `yama-review.yml` | `[release]` |
| `copilot-review.yml` | `[release, main]` |

So on this PR nothing has been built or tested by CI, and that is the accurate reading.

What has run, locally on the **stack tip** (so every branch's code compiled together) after the latest rebase:

- `pnpm run check` → 4961 files, **0 errors, 0 warnings**
- all nine new suites → **95 tests, 0 failures, 0 skips**
  (`tool-pairing` 18, `token-accounting` 10, `step-guard` 6, `loop-guard-core` 11, `openai-compat-guard` 9, `anthropic-guard` 10, `tool-storage-parity` 10, `redis-append-only` 10, `gemini-guard` 11)

That is local signal, not CI signal, and I would rather say so plainly than let the tick count imply otherwise. Merging bottom-up closes the gap for real: each PR's base becomes `release` as its parent lands, at which point it picks up the full fifteen-check gate before it can merge — so no branch reaches `release` without having been through it.

pdogra1299:
#1280 is merged (`a0f49e40`), so the stack has been rebased onto the new `release` (now at 10.10.2 — it had moved two releases since these branches were cut). Still one commit per branch.

**#1281 was `CONFLICTING` after the merge** — its branch still carried the pre-squash copy of #1280's commit. Replanted with `git rebase --onto origin/release 8344a885` so only its own commit survives; now `MERGEABLE` against `release`.

**One real conflict, in `providers/anthropic/client.ts`.** `release` had gained the structured-output buffering from the `final_result` work (`bufferedText` / `finalResultText`) at the same point where #1283 adds `estimateAnthropicFixedOverhead`. Both-added, both wanted — kept both. Verified afterwards that the release-side path is still fully wired (`bufferedText` accumulates at the text-delta branch, `finalResultText` is set from the `final_result` tool call, and the buffered turn is still delivered at the end of the loop), not just that it compiles.

## The build was broken, and `pnpm run check` did not catch it

Worth recording since it is exactly the CI gap @murdore flagged. Building the tip surfaced two errors in #1286's code that the typechecker passes clean:

```
redisConversationMemoryManager.ts(987,7): error TS2345:
  '(string | Buffer)[]' is not assignable to 'string[]'
redisConversationMemoryManager.ts(1000,5): error TS2322:
  'number | `${number}`' is not assignable to 'number'
```

`pnpm run check` uses the main tsconfig; `build:react-hooks` compiles with `--module nodenext --moduleResolution nodenext`, which resolves the `redis` types differently and widens both return types. Fixed at the source rather than with an assertion (Rule 14): `lRange` entries are normalized via `.toString()`, `lLen` through `Number()`. Both are real defects, not just type noise — a binary-mode client would have thrown on parse, and a RESP3 string count would have concatenated where callers add.

## Verification

Each branch was built **independently**, since each merges on its own:

| branch | build | 
|---|---|
| `fix/token-accounting` | ✅ |
| `fix/cache-aware-compaction` | ✅ |
| `fix/loop-guard-parity` | ✅ |
| `fix/session-persistence` | ✅ |
| `fix/redis-append-only` | ✅ |
| `fix/loop-guard-completion` | ✅ |

On the tip: `pnpm run check` → 4939 files, **0 errors**; `pnpm run lint` → **0 errors**, 48 warnings, all pre-existing `max-lines-per-function` in files this stack does not touch (`claudeProxyRoutes.ts`, `toolRoutes.ts`, `csvProcessor.ts`); all nine suites → **95 tests, 0 failures, 0 skips**.

Still local signal, not CI signal — except on #1281, which now targets `release` and is running the full gate.

pdogra1299:
#1281 is merged (`613709b3`), so `release` is now at both stack heads #1280 and #1281. The remaining stack has been rebased onto it.

**Rebase — one commit each, no conflicts:**

```
94370cfb  #1287  fix(providers): complete in-turn context guard coverage across native loops
31f043cd  #1286  fix(memory): store session messages in an append-only redis list
ca47f289  #1285  fix(memory): persist tool executions on every memory backend
451fe2ba  #1283  fix(providers): guard native agent loops against mid-turn overflow
f722c458  #1282  fix(context): reclaim to a low-water mark in the step budget guard
613709b3  release
```

#1282 was `DIRTY` because it still carried the pre-squash copy of #1281's commit; that is resolved.

**Verification — each branch independently, not just the tip:**

| branch | `pnpm run check` | `pnpm run build` | own suites |
| --- | --- | --- | --- |
| `fix/cache-aware-compaction` | 4935 files, 0 errors | pass | step-guard 6/6 |
| `fix/loop-guard-parity` | 4938 files, 0 errors | pass | anthropic-guard 10/10, loop-guard-core 11/11, openai-compat-guard 9/9 |
| `fix/session-persistence` | 4938 files, 0 errors | pass | tool-storage-parity 10/10 |
| `fix/redis-append-only` | 4938 files, 0 errors | pass | redis-append-only 10/10 |
| `fix/loop-guard-completion` | 4939 files, 0 errors | pass | gemini-guard 11/11 |

67 tests, 0 failures, 0 skips. `pnpm run build` is run in full because `pnpm run check` alone does not catch everything — the `nodenext` resolution used by `build:react-hooks` surfaced two real `redis` typing defects in #1286 that `check` passed over.

**Review comments:** no unresolved inline findings on this PR. The only review activity to date is CodeRabbit's auto-generated walkthrough and @murdore's note about the check count, which is answered above — that gap closes as each PR's base becomes `release`. #1281 bore this out: once it targeted `release` it ran all 16 checks and passed, and that run triggered a CodeRabbit pass with three findings. Those landed after #1281 was already merged, so they are carried in a separate follow-up rather than retro-fitted here.

PR #1307, opened 2026-08-08
feat(files): support every multimodal format exhaustively via one registry

murdore:
## The problem

The same extension/MIME knowledge lived in **five hand-maintained tables** with no way of agreeing with each other:

1. `EXTENSION_MIME_MAP` (`config/mimeConstants.ts`)
2. the `*_EXTENSIONS` category arrays (`config/fileExtensions.ts`)
3. `ExtensionStrategy.typeMap` **+** `.mimeMap` (`utils/fileDetector.ts`)
4. `EXTENSION_TYPE_MAP` (`utils/messageBuilder.ts`)
5. each processor's `supportedExtensions` / `supportedMimeTypes`

They had drifted, and the drift was not cosmetic: **a format a processor declared it supported was routinely invisible to the detector standing in front of it**, so the file never reached that processor. Measured on real ffmpeg-generated media, not inferred:

```
.aiff .m2ts .mts .vob .3g2 .3gp  ->  unknown   (downgraded to metadata text)
.mpg .mpeg                       ->  csv       (binary parsed as a spreadsheet)
.ts                              ->  text      (MPEG-TS inlined as source code)
```

Two structural causes sat underneath:

- **Magic-byte detection never ran for file paths** — it bailed unless the input was already a `Buffer`, so a path was classified by extension alone.
- **The CSV heuristic had no binary guard.** It runs on a UTF-8 *decoding* of the bytes, which succeeds for anything; MPEG-1 video decoded to "lines" with a consistent delimiter count, which was the entire test.

## The fix

One `FILE_TYPE_REGISTRY` — extension ↔ MIME ↔ FileType ↔ modality — from which all five tables derive. Duplicate extensions and MIME types are rejected at module load, so this class of drift is now **structurally impossible rather than fixed once** (it caught a collision while the registry was being written).

`fileType` (routing) and `modality` (category) are deliberately separate: `.odt` is `modality: document` but `fileType: docx` because the Word processor handles it; `.svg` is `modality: image` but `fileType: svg` because it is sanitised as markup and never sent to a vision API.

### Detection

- **Magic bytes now run for file paths.** Side effect worth having: the content heuristics `readFile()` the *whole* file, so identifying a 500 MB video used to read 500 MB to conclude nothing. Now 4 KB.
[...]

PR #1308, opened 2026-08-09
fix(multimodal): stop dropping images above the lazy-reference threshold

murdore:
## The bug

Files over `SIZE_TIER_THRESHOLDS.TINY_MAX` (10 KB) are registered as lazy references and a short textual preview is injected instead of the file. For text-shaped input that trade is sound. For an image it is fatal: the preview is ~98 characters of prose, the pixels never reach the model, and the model then answers confidently about a file it has never seen.

10 KB is far below any real photo, so this affected essentially every image attached by path — an ordinary JPEG, not just unusual formats. Measured on `release`, asking *"what number is written in this image?"*:

| fixture | size | answer |
|---|---|---|
| `tiny.png` | 1.6 KB | `7391` — under the threshold, eager, correct |
| `big.png` | 11 KB | `NOTHING_RECEIVED` |
| `big.jpg` | 19 KB | `NOTHING_RECEIVED` |

## The fix

Images are always processed eagerly. This costs no extra memory: `tryRegisterFileReference` already calls `getFileBuffer()` and reads the whole file to register it — the lazy path was never lazy about *reading*, only about processing.

The eager/lazy decision reads an extension→type map that previously existed only for token estimation, where a missing entry costs a slightly-off guess. It now decides whether a file's bytes survive, so a missing entry drops the photo — and HEIC, HEIF, ICO, JPEG 2000, APNG and SVG were all missing. **HEIC is what an iPhone writes by default.** Those entries are added here.

Note HEIC/HEIF/ICO/JP2 additionally need the intake allowlist in `ImageProcessor.validateImageFormat` widened before they work end-to-end — a separate change — so this PR does not assert them.

## Why audio and video are left alone here

For audio the reason is not the one it looks like: an audio message is a metadata block (duration, codec, sample rate) on **both** paths, so making it eager changes nothing about what the model receives. That audio content never arrives at all is a separate, larger gap.

## Tests

Three end-to-end tests through `new NeuroLink()` that attach a rendered, un-guessable number and assert the model reads it back. Each asserts which side of the threshold its fixture lands on *before* the call, so they cannot drift back into testing only the easy path.

Verified to **fail without the fix and pass with it** — with the fix reverted, the two above-threshold tests fail and the sub-threshold control stays green.

- `pnpm run check` — 4815 files, 0 errors
- `pnpm run lint` — 0 errors
- `pnpm run build` — publint clean
- `test:multimodal:sdk` — 18/18 live

## Summary by CodeRabbit

* **New Features**
  * Added recognition for JPE, APNG, HEIC, HEIF, ICO, JP2, and JPX image formats.
  * SVG files are detected from their content, even without a reliable filename or MIME type.
  * Large images are processed correctly using filename, MIME type, metadata, and content signals.
  * Extensionless image uploads are supported when a valid MIME type is provided.

* **Bug Fixes**
  * Improved handling of misleading filenames and oversized image files.

### DIFF
--- a/src/lib/utils/messageBuilder.ts
+++ b/src/lib/utils/messageBuilder.ts
@@ -80,13 +80,32 @@ const EXTENSION_TYPE_MAP: Record<string, string> = {
   // Image
   jpg: "image",
   jpeg: "image",
+  jpe: "image",
+  jfif: "image",
   png: "image",
+  apng: "image",
   gif: "image",
   webp: "image",
   bmp: "image",
   tiff: "image",
   tif: "image",
   avif: "image",
+  // Below this line the entries exist because this map now decides whether an
+  // image is processed eagerly (see isEagerMultimodalFile), not merely how its
+  // tokens are estimated. An extension missing here classifies as `undefined`
+  // and takes the lazy path, which drops the pixels — so a gap is a dropped
+  // photo, not a slightly-off estimate. HEIC is the one that matters most:
+  // it is what an iPhone writes by default.
+  heic: "image",
+  heif: "image",
+  ico: "image",
+  jp2: "image",
+  jpx: "image",
+  // NB: `svg` is deliberately absent here — it is mapped to its own "svg"
+  // routing type further down, and listing it twice is a duplicate key whose
+  // second entry silently wins. `isEagerMultimodalFile` accepts both types
+  // instead, so SVG takes the eager path without this map having to lie about
+  // which processor handles it.
   // Archive
   zip: "archive",
   tar: "archive",
@@ -137,6 +156,20 @@ function inferFileTypeFromExtension(filePath: string): string | undefined {
 function inferFileTypeFromBuffer(buf: Buffer): string | undefined {
   if (buf.length < 4) {
     return undefined;
+  }
+
+  // SVG is markup and has no magic number, so every signature check below
+  // misses it and a raw SVG Buffer classified as `undefined` — which meant the
+  // lazy path, which previews markup away. Checked first because the sniff is
+  // a cheap look at the head and cannot collide with a binary signature.
+  //
+  // Deliberately a substring scan over the head rather than a prolog-stripping
+  // regex: the obvious pattern for skipping comments and a DOCTYPE is two
+  // nested quantifiers, which is a ReDoS waiting to happen inside what is
+  // supposed to be a cheap type check. Over-matching is harmless here — the
+  // only consequence of a false positive is that a file is processed eagerly.
+  if (buf.subarray(0, 1024).toString("latin1").includes("<svg")) {
+    return "svg";
   }
 
   // PNG
@@ -1088,7 +1121,11 @@ export async function processUnifiedFilesArray(
         try {
           // ─── Lazy file registration path ──────────────────────────────
           const fileSize = fileRegistry ? getFileSize(file) : 0;
-          if (fileRegistry && fileSize > SIZE_TIER_THRESHOLDS.TINY_MAX) {
+          if (
+            fileRegistry &&
+            fileSize > SIZE_TIER_THRESHOLDS.TINY_MAX &&
+            !isEagerMultimodalFile(file)
+          ) {
             const registered = await tryRegisterFileReference(
               file,
               fileSize,
@@ -2583,6 +2620,110 @@ function getFileSource(file: FileInput): "buffer" | "path" | "url" | "datauri" {
     }
   }
   return "buffer";
+}
+
+/**
+ * Whether a file must be processed eagerly rather than lazily referenced.
+ *
+ * The lazy path registers a file and injects a short textual *preview* in place
+ * of the file itself. Whether that is an acceptable trade depends entirely on
+ * whether a description of the file can answer questions about it:
+ *
+ *   pdf    lazy -> preview carries the extracted text   ✔ content survives
+ *   image  lazy -> preview carries ~98 chars of prose   ✘ the pixels are gone
+ *
+ * An image is the case where the bytes ARE the content: no prose summary
+ * substitutes for them, so the model receives a description of a file instead
+ * of the file. Measured end-to-end on `release`, asking "what number is
+ * written in this image?":
+ *
+ *   tiny.png  1.6 KB -> "7391"              (under TINY_MAX, eager, correct)
+ *   big.png    11 KB -> NOTHING_RECEIVED    (lazy, image never arrived)
+ *   big.jpg    19 KB -> NOTHING_RECEIVED
+ *
+ * 10 KB is far below any real photo, so this affected essentially every image
+ * attached by path — an ordinary JPEG, not just unusual formats.
+ *
+ * Scoped to images deliberately, and audio deserves spelling out because the
+ * reason is not the one it appears to be. An audio file's message contains only
+ * a metadata block — duration, codec, sample rate — on BOTH paths; no audio
+ * bytes are handed to the provider either way. Moving audio to the eager path
+ * would therefore change nothing about what the model receives. That audio
+ * content never reaches the model at all is a separate and pre-existing gap,
+ * not something this threshold decision can repair, and it is easy to mistake
+ * for working code because the metadata block answers exactly the questions
+ * ("how long is it?", "what sample rate?") a test is most tempted to ask.
+ * Video is left alone for the opposite reason: its frames are measurably
+ * present in the message and a model reads them correctly.
+ *
+ * Note this costs no extra memory: `tryRegisterFileReference` already calls
+ * `getFileBuffer()` and reads the whole file to register it. The lazy path was
+ * never lazy about reading — only about processing — so the difference here is
+ * simply whether the bytes survive.
+ */
+function isEagerMultimodalFile(file: FileInput): boolean {
+  if (typeof file === "string") {
+    return isImageLikeType(inferFileTypeFromExtension(file));
+  }
+  if (Buffer.isBuffer(file)) {
+    return isImageLikeType(inferFileTypeFromBuffer(file));
+  }
+
+  // A `FileWithMetadata` carries two independent declarations, and either one
+  // alone is enough: the shape exists for Slack/Curator-style uploads that
+  // arrive as bytes plus a mimetype, so its `filename` may be extensionless or
+  // simply wrong. Reading them as a `??` chain meant the first *recognised*
+  // name won outright — `upload.pdf` with `mimetype: "image/png"` classified as
+  // a PDF and lost its pixels down the lazy path. Any declaration of an image
+  // is therefore decisive.
+  const declared = [
+    inferFileTypeFromExtension(file.filename),
+    inferFileTypeFromMimetype(file.mimetype),
+  ];
+  if (declared.some(isImageLikeType)) {
+    return true;
+  }
+
+  // The buffer sniff is the last resort rather than a third vote, because it is
+  // a substring scan: an HTML page with an inline `<svg>` icon in its head
+  // would otherwise be pulled onto the eager path and sent in full, which is
+  // the opposite of what the size tiers are for. It only speaks when nothing
+  // else did.
+  return declared.every((type) => type === undefined)
+    ? isImageLikeType(inferFileTypeFromBuffer(file.buffer))
+    : false;
+}
+
+/**
+ * Whether a routing type should have its bytes preserved rather than previewed.
+ *
+ * "svg" is a separate routing type rather than a sub-case of "image" (it goes
+ * to the sanitizer, not to a vision encoder), but it is still an image as far
+ * as this decision is concerned: its markup IS its content, and previewing it
+ * away leaves the model with nothing. Accepting both keeps this correct
+ * whichever of the two type vocabularies the caller's map uses.
+ */
+function isImageLikeType(type: string | undefined): boolean {
+  return type === "image" || type === "svg";
+}
+
+/**
+ * Infer a routing type from a caller-declared mimetype.
+ *
+ * Only images matter here — this exists so the eager/lazy decision can read a
+ * mimetype hint — and "application/octet-stream" is deliberately ignored,
+ * because it is the opaque sentinel a caller sends when it knows nothing, not
+ * a claim about content.
+ */
+function inferFileTypeFromMimetype(mimetype?: string): string | undefined {
+  if (!mimetype) {
+    return undefined;
+  }
+  const normalized = mimetype.split(";")[0].trim().toLowerCase();
+  if (normalized === "image/svg+xml") {
+    return "svg";
+  }
+  return normalized.startsWith("image/") ? "image" : undefined;
 }
 
 /**
--- a/test/helpers/mediaFixtures.ts
+++ b/test/helpers/mediaFixtures.ts
@@ -140,3 +140,86 @@ export function makeCorruptFile(dir: string, name: string): string {
   fs.writeFileSync(out, Buffer.from("not really media, just ascii", "utf8"));
   return out;
 }
+
+/**
+ * Render a number into an image of a chosen byte size.
+ *
+ * Two properties matter and neither is incidental.
+ *
+ * The *content* is a number the model cannot guess. Asking "describe this
+ * image" is worthless — a model that received nothing still produces plausible
+ * prose — and asking about a property with a strong prior (dimensions, a
+ * common colour) is answerable without looking. A four-digit token has no
+ * prior, so a correct answer can only have come from the pixels.
+ *
+ * The *size* is chosen relative to `SIZE_TIER_THRESHOLDS.TINY_MAX` (10 KB).
+ * Files under it are processed eagerly; files over it take the lazy
+ * reference path. An image on the wrong side of that line silently became a
+ * ~98-character text preview and never reached the model at all — and the
+ * reason that shipped is that every image fixture in the suites happened to be
+ * a few kilobytes. `bigPixels` pushes the encode over the threshold so the
+ * lazy path is actually exercised.
+ *
+ * @param label - Digits rendered into the image; keep it un-guessable.
+ * @param bigPixels - When true, render large enough to exceed TINY_MAX.
+ */
+export async function makeNumberImage(
+  dir: string,
+  name: string,
+  label: string,
+  bigPixels = false,
+): Promise<string> {
+  const magick = findImageMagick();
+  if (!magick) {
+    throw new Error("ImageMagick unavailable");
+  }
+  const out = path.join(dir, name);
+  const size = bigPixels ? "1600x900" : "200x100";
+  const pointsize = bigPixels ? "220" : "48";
+  await execFileAsync(magick, [
+    "-size",
+    size,
+    "xc:white",
+    "-fill",
+    "black",
+    "-pointsize",
+    pointsize,
+    "-gravity",
+    "center",
+    "-annotate",
+    "0",
+    label,
+    out,
+  ]);
+  return out;
+}
+
+/**
+ * Locate ImageMagick: `IMAGEMAGICK_PATH` first, then the usual install roots.
+ *
+ * Mirrors `findFfmpeg`'s env-override-then-fixed-paths order, so an unusual
+ * install can be pointed at rather than making the caller skip. It stops short
+ * of `findFfmpeg`'s bare-name PATH fallback on purpose: that one is safe
+ * because `hasFfmpeg()` proves the binary runs before anything relies on it,
+ * and there is no equivalent probe here — returning "magick" on a machine
+ * without it would turn a clean skip into a spawn failure mid-fixture.
+ */
+export function findImageMagick(): string | null {
+  const explicit = process.env.IMAGEMAGICK_PATH;
+  if (explicit && fs.existsSync(explicit)) {
+    return explicit;
+  }
+
+  for (const candidate of [
+    "/opt/homebrew/bin/magick",
+    "/usr/local/bin/magick",
+    "/usr/bin/magick",
+    "/opt/homebrew/bin/convert",
+    "/usr/bin/convert",
+  ]) {
+    if (fs.existsSync(candidate)) {
+      return candidate;
+    }
+  }
+  return null;
+}
--- a/test/continuous-test-suite-multimodal-sdk.ts
+++ b/test/continuous-test-suite-multimodal-sdk.ts
@@ -34,14 +34,18 @@ import {
   Skip,
 } from "./helpers/harness.js";
 import {
+  findImageMagick,
   hasFfmpeg,
   makeAudioFile,
+  makeNumberImage,
   makeVideoFile,
 } from "./helpers/mediaFixtures.js";
 import { hasPackage, makeDocx, makeXlsx } from "./helpers/officeFixtures.js";
 import { FileDetector } from "../src/lib/utils/fileDetector.js";
+import { redactUrlsInText } from "../src/lib/utils/logSanitize.js";
 import { buildMultimodalMessagesArray } from "../src/lib/utils/messageBuilder.js";
 import { NeuroLink } from "../src/lib/neurolink.js";
+import { SIZE_TIER_THRESHOLDS } from "../src/lib/types/index.js";
 
 const { test, runSuite } = defineSuite("Multimodal through the SDK");
 
@@ -192,6 +196,46 @@ async function generateNonEmpty(
   return last;
 }
 
+/**
+ * Assert on a live reply without putting the reply in the failure message.
+ *
+ * `defineSuite`'s `test()` classifies a thrown error as SKIP — not FAIL — when
+ * its message matches `isExpectedProviderError()`, and that check reads the
+ * message text. An assertion that quotes model output can therefore be
+ * downgraded to a green skip the moment the model happens to say "timeout",
+ * "not found" or any other provider-ish phrase, which is precisely the
+ * false-green this suite exists to prevent.
+ *
+ * The reply is still printed — diagnosing a multimodal failure without seeing
+ * what came back is miserable — just to stderr rather than into the message
+ * the harness inspects.
+ */
+function formatReply(reply: string): string {
+  // Redacted because the reply is provider output: a model asked to describe an
+  // attachment can echo a signed URL back, and this goes to a CI log.
+  return redactUrlsInText(reply.slice(0, 300).replace(/\s+/g, " "));
+}
+
+function assertReply(condition: boolean, summary: string, reply: string): void {
+  if (!condition) {
+    console.error(`      ↳ model reply: ${formatReply(reply)}`);
+  }
+  assert(condition, summary);
+}
+
+/** `assertIncludes` counterpart to {@link assertReply}, for the same reason. */
+function assertIncludesReply(
+  haystack: string,
+  needle: string,
+  summary: string,
+  reply: string,
+): void {
+  if (!haystack.includes(needle)) {
+    console.error(`      ↳ model reply: ${formatReply(reply)}`);
+  }
+  assertIncludes(haystack, needle, summary);
+}
+
 // --- VIDEO-026 (#498) / AUDIO-030 (#483): FileDetector on real bytes ---------
 //
 // The detector is the first thing every multimodal call hits. These assert it
@@ -308,15 +352,17 @@ await test("audio reaches the model via input.files", async () => {
     provider: PROVIDER,
     maxTokens: 512,
   });
-  assert(
+  assertReply(
     !content.includes("NOTHING_RECEIVED"),
-    `the model reported receiving nothing; got: ${content.slice(0, 200)}`,
+    `the model reported receiving nothing`,
+    content,
   );
   // Only obtainable by reading the container header.
-  assertIncludes(
+  assertIncludesReply(
     content.toLowerCase(),
     "tone.mp3",
-    `the model described the actual file; got: ${content.slice(0, 200)}`,
+    `the model described the actual file`,
+    content,
   );
 });
 
@@ -339,14 +385,16 @@ await test("audio reaches the model as a Buffer via input.files", async () => {
     provider: PROVIDER,
     maxTokens: 512,
   });
-  assert(
+  assertReply(
     !content.includes("NOTHING_RECEIVED"),
-    `buffer input reached the model; got: ${content.slice(0, 200)}`,
+    `buffer input reached the model`,
+    content,
   );
   // Word-bounded so a stray "7" inside a larger number cannot satisfy it.
-  assert(
+  assertReply(
     new RegExp(`\\b${ODD_SECONDS}\\b`).test(content),
-    `the model read the real duration (${ODD_SECONDS}s) from the buffer; got: ${content.slice(0, 200)}`,
+    `the model read the real duration (${ODD_SECONDS}s) from the buffer`,
+    content,
   );
 });
 
@@ -367,13 +415,15 @@ await test("input.audioFiles delivers audio (regression for #1259)", async () =>
     provider: PROVIDER,
     maxTokens: 512,
   });
-  assert(
+  assertReply(
     !content.includes("NOTHING_RECEIVED"),
-    `audioFiles reached the model; got: ${content.slice(0, 200)}`,
+    `audioFiles reached the model`,
+    content,
   );
-  assert(
+  assertReply(
     new RegExp(`\\b${ODD_SECONDS}\\b`).test(content),
-    `the model read the real duration (${ODD_SECONDS}s) via audioFiles; got: ${content.slice(0, 200)}`,
+    `the model read the real duration (${ODD_SECONDS}s) via audioFiles`,
+    content,
   );
 });
 
@@ -401,10 +451,11 @@ await test("generate() reads content out of a .docx", async () => {
   });
   // Asserting on content the model could only know by reading the file is the
   // difference between "the call succeeded" and "the document arrived".
-  assertIncludes(
+  assertIncludesReply(
     result.content.toUpperCase(),
     "FALCON",
-    `docx content reached the model; got: ${result.content.slice(0, 160)}`,
+    `docx content reached the model`,
+    result.content,
   );
 });
 
@@ -433,10 +484,11 @@ await test("generate() reads cell values out of an .xlsx", async () => {
     maxTokens: 512,
     timeout: LIVE_TIMEOUT_MS,
   });
-  assertIncludes(
+  assertIncludesReply(
     result.content,
     "4242",
-    `xlsx cell reached the model; got: ${result.content.slice(0, 160)}`,
+    `xlsx cell reached the model`,
+    result.content,
   );
 });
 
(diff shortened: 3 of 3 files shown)


Tara-ag:
🔒 CRITICAL: Missing image extensions cause validation failures

PR adds extensions `jpe`, `jfif`, `apng`, `jp2`, `jpx` to `EXTENSION_TYPE_MAP` in messageBuilder.ts but does NOT add them to `VALID_IMAGE_EXTENSIONS` in imageProcessor.ts or `detectImageType()`. Files with these extensions will be misclassified or rejected.

**Impact**: Users uploading files with `.jpe`, `.apng`, `.jp2`, `.jpx` extensions will encounter errors because:
1. The file passes initial classification as "image" type
2. But later validation fails at `isValidImageExtension()` which only knows jpg/jpeg/png/gif/webp/bmp/tiff/tif/svg/avif
3. OR magic-byte detection returns `application/octet-stream` which is rejected by `assertKnownImageType()`

**Fix required**: Add jpe, jfif, apng, jp2, jpx to both:
- `src/lib/utils/imageProcessor.ts`: VALID_IMAGE_EXTENSIONS array  
- `src/lib/utils/imageProcessor.ts`: detectImageType() function's extension mapping object

Tara-ag:
🔒 CRITICAL: Missing image extensions cause validation failures for jpe, jfif, apng, jp2, jpx

PR adds extensions `jpe`, `jfif`, `apng`, `jp2`, `jpx` to `EXTENSION_TYPE_MAP` in messageBuilder.ts but does NOT add them to `detectImageType()` or `validateImageFormat()` in imageProcessor.ts.

**Impact chain**:
1. Files with `.jpe/.jfif/.apng/.jp2/.jpx` pass initial classification as "image" type
2. But `detectImageType()` returns `application/octet-stream` (unknown format)
3. `assertKnownImageType()` throws: "Unsupported or corrupted image: no known image signature was found"
4. Users uploading these valid formats get hard errors

**Fix required** in src/lib/utils/imageProcessor.ts:
- In `detectImageType()`: add jpe→image/jpeg, jfif→image/jpeg, apng→image/png, jp2/jpx→image/jp2
- In `validateImageFormat()`: add corresponding MIME types to supportedFormats array

Tara-ag:
💡 MINOR: Heic/heif already in VALID_IMAGE_EXTENSIONS but missing from validateImageFormat()

VALID_IMAGE_EXTENSIONS includes `heic` and `heif`, but `validateImageFormat()` does not accept `image/heic` or `image/heif` MIME types. This means files with `.heic/.heif` extensions will fail validation despite being listed as valid.

**Fix**: Add `image/heic` and `image/heif` to `validateImageFormat()`'s supportedFormats array in src/lib/utils/imageProcessor.ts.

Tara-ag:
🔒 CRITICAL: Tests require ImageMagick but dependency not declared and no skip mechanism

PR adds `makeNumberImage()` helper in mediaFixtures.ts which calls `execFileAsync(ImageMagick, ...)`. Tests in continuous-test-suite-multimodal-sdk.ts use this helper:

- test: an image under the lazy threshold still reaches the model@L631
- test: an image ABOVE the lazy threshold still reaches the model@L651

If ImageMagick is not installed in CI, these tests will fail with "ImageMagick unavailable" error.

Package.json shows ffmpeg-static is optionalDependency, but ImageMagick is neither:
1. Declared as dependency/optionalDependency
2. Has skip-if mechanism in tests  
3. Has mocking for CI environments

Impact: test:multimodal:sdk will fail in CI on Linux containers without ImageMagick.

**Fix options**:
1. Add ImageMagick installation to CI environment setup (e.g., Dockerfile, .github/workflows)
2. Wrap makeNumberImage() in try-catch and skip test gracefully when unavailable
3. Use pre-existing image fixtures instead of dynamic generation

Tara-ag:
## Yama Review Summary for PR #1308

**Decision: CHANGES_REQUESTED**

### Findings Summary

| Severity | Category | File | Count |
|----------|----------|------|-------|
| 🔒 MAJOR | Validation Gap - Image Extensions | messageBuilder.ts | 1 |
| 💡 MINOR | Incomplete Fix | imageProcessor.ts | 1 |
| 🔒 MAJOR | Missing Dependency | mediaFixtures.ts | 1 |
| **Total** | | | **3 issues** |

### Detailed Issues

1. **🔒 CRITICAL**: Missing image extensions cause validation failures for jpe, jfif, apng, jp2, jpx
   - Files with these extensions fail downstream validation despite being classified as images
   - Users will get "Unsupported or corrupted image" errors when uploading valid formats
   
2. **💡 MINOR**: Heic/heif already in VALID_IMAGE_EXTENSIONS but missing from validateImageFormat()
   - Inconsistent behavior - .heic/.heif files listed as valid but rejected during processing
   
3. **🔒 MAJOR**: Tests require ImageMagick but dependency not declared and no skip mechanism  
   - CI tests will fail on environments without ImageMagick installed
   - No error handling or skip-if mechanism in place

### Impact Analysis

- **Blast radius**: Changes to `processUnifiedFilesArray` affect all provider implementations that use multimodal file handling (Google AI Studio, Google Vertex)
- **Affected execution flows**: `buildMultimodalMessagesArray`, `convertMultimodalToProviderFormat`, and processor chain functions
- **Risk level**: Medium (0.60) - affects core file processing pipeline

### Review Scope

Reviewed 3 changed files:
1. `src/lib/utils/messageBuilder.ts` - Core logic changes for image extension handling
2. `test/continuous-test-suite-multimodal-sdk.ts` - New test cases for lazy threshold boundaries  
3. `test/helpers/mediaFixtures.ts` - Helper function additions

### Recommended Fixes

1. **Add missing extensions to src/lib/utils/imageProcessor.ts**:
   - Update `detectImageType()` mapping for jpe, jfif, apng, jp2, jpx
   - Update `validateImageFormat()` allowed list
   
2. **Add heic/heif MIME types to validateImageFormat()**:
   - Add `image/heic` and `image/heif` to supported formats
   
3. **Fix ImageMagick dependency issue**:
   - Install ImageMagick in CI environment OR add skip-if mechanism to tests

---

*Review completed by Yama autonomous code review agent*

Tara-ag (commented):
Reviewing PR #1308 for multimodal SDK enhancements and related fixes. Found 3 issues requiring attention before merge.

Tara-ag:
🔒 CRITICAL: The `/api/agent/embed` endpoint does not validate or sanitize the `prompt` parameter before passing it to the embedding model. An attacker could inject malicious payloads (e.g., `<script>` tags, command injection, prompt injection attacks) that might be processed by the embedding model or leaked through error responses.

The `prompt` parameter should be validated/sanitized similar to how other endpoints handle user input. Consider:
- Sanitizing special characters if needed
- Adding length limits
- Validating against a schema
- Ensuring no dangerous patterns are accepted

This is a security vulnerability that could allow attackers to exploit the embedding API.

Tara-ag:

Severity counts — 🔒 CRITICAL: 0 · ⚠️ MAJOR: 1 · 💡 MINOR: 3

PR addresses lazy-to-eager image routing with proper mimetype/buffer handling for FileWithMetadata, SVG buffer detection, and extended image format support. Two minor findings remain: test infrastructure dependency not declared (ImageMagick) and incomplete HEIC/HEIF MIME type allowlist in ImageProcessor.

- ⚠️ **MAJOR**: Test suite requires live API access but has no mock fallback — `test/continuous-test-suite-multimodal-sdk.ts:745`
  The new image tests call generateNonEmpty() with real NeuroLink instances and expect live provider responses. Without proper credentials configured, these tests will fail. The existing suite has requireLive() guards but the new tests don't explicitly handle the case where no provider is available.
- 💡 **MINOR**: Missing image format validation for HEIC/HEIF/ICO/JPEG-2000 — `src/lib/utils/messageBuilder.ts:94`
  The change adds heic, heif, ico, jp2, jpx to EXTENSION_TYPE_MAP so they trigger eager processing. However, ImageProcessor.validateImageFormat() explicitly rejects these MIME types. This means files with these extensions will be classified as eager but then rejected at intake.
- 💡 **MINOR**: Test Infrastructure Dependency Not Declared — `test/helpers/mediaFixtures.ts:207`
  test/helpers/mediaFixtures.ts uses ImageMagick but the binary is not declared in package.json. The test will fail on CI machines without ImageMagick installed.
- 💡 **MINOR**: HEIC/HEIF Format Support Incomplete — `src/lib/utils/messageBuilder.ts:106`
  Adding heic/heif to EXTENSION_TYPE_MAP does not enable them - ImageProcessor.validateImageFormat rejects these MIME types. The extension mapping alone is insufficient.

murdore:
Force-pushed with review fixes. Summary of what changed and what I checked but did not change:

**Fixed**
- **Assertion messages no longer quote model output** (Copilot + CodeRabbit, both suites). `defineSuite` downgrades a thrown error to SKIP when its message matches `isExpectedProviderError()`, which reads the message text — so an assertion quoting a reply could be reported as a green skip the moment the model said "timeout" or "not found". Replies now go to stderr via `assertReply`/`assertIncludesReply`, keeping them diagnosable without letting them decide the verdict. This is an explicitly documented rule in CLAUDE.md that I had violated.
- **Eager/lazy classification now reads every signal the input carries** (Copilot + CodeRabbit, #1308). It inspected `filename` alone, so a `FileWithMetadata` with no extension but a declared `image/*` mimetype — the whole reason that shape exists — took the lazy path and lost its pixels. A raw SVG `Buffer` had the same problem, since markup has no magic number. Added a regression test that attaches an extension-less buffer and asserts the model reads the number back.
- **`findImageMagick` now honours `IMAGEMAGICK_PATH`** (Copilot, #1308), so its docblock's claim to mirror `findFfmpeg` is true. It still stops short of the bare-name PATH fallback, and now says why.

**Checked, already correct — no change**
- `readImageSourceForConversion` MIME coverage for HEIC/HEIF/AVIF/ICO/JP2/APNG (Copilot). Verified empirically: the path branch resolves through the registry, and the buffer branch resolves HEIC/HEIF/AVIF through `detectIsoBmffImageMimeType`, with `needsVisionTranscode` returning true for all three. APNG correctly resolves to `image/png`, which needs no transcode. The comment predates the current commit.
- The "1 KB" magic-byte comment (Copilot) already reads `{@link MAGIC_BYTES_HEADER_SIZE}`.
- `bigPixels` byte-threshold nitpick (CodeRabbit) is already covered by the in-test size assertion, which fails loudly rather than silently exercising the eager path.

**Also**
- Hardened the format suite against transient provider drops. `.flac` failed once in a full run and then passed 4/4 in isolation, so `NOTHING_RECEIVED` is now retried. This does not blunt the assertion: a format that genuinely never arrives returns the sentinel on every attempt and still fails, because the last reply is what gets asserted.

**Verification after the changes**
- `test:file-formats` — 57 passed / 0 failed / 9 skipped (skips are formats this machine cannot encode, each naming its reason)
- `test:multimodal:sdk` — 19/19
- `check` 0 errors · `lint` 0 errors · `build` publint clean, on all three branches

Note the 🔒 Single Commit Policy check fails on #1307 and #1309 by construction: it counts commits against `release` rather than the PR base, and these are stacked. It resolves as the stack merges in order **#1308 → #1307 → #1309**.

test/helpers/mediaFixtures.ts
```diff
@@ -140,3 +140,86 @@ export function makeCorruptFile(dir: string, name: string): string {
...
+ * Locate ImageMagick: `IMAGEMAGICK_PATH` first, then the usual install roots.
+ *
+ * Mirrors `findFfmpeg`'s env-override-then-fixed-paths order, so an unusual
+ * install can be pointed at rather than making the caller skip. It stops short
+ * of `findFfmpeg`'s bare-name PATH fallback on purpose: that one is safe
+ * because `hasFfmpeg()` proves the binary runs before anything relies on it,
+ * and there is no equivalent probe here — returning "magick" on a machine
+ * without it would turn a clean skip into a spawn failure mid-fixture.
+ */
+export function findImageMagick(): string | null {
```
Tara-ag, line 207: 💡 MINOR: Test Infrastructure Dependency Not Declared — test/helpers/mediaFixtures.ts uses ImageMagick but the binary is not declared in package.json. The test will fail on CI machines without ImageMagick installed. Add 'imagemagick' to package.json devDependencies or use conditional test execution with skip mechanism.

src/lib/utils/messageBuilder.ts
```diff
@@ -80,13 +80,32 @@ const EXTENSION_TYPE_MAP: Record<string, string> = {
...
+  // photo, not a slightly-off estimate. HEIC is the one that matters most:
+  // it is what an iPhone writes by default.
+  heic: "image",
+  heif: "image",
+  ico: "image",
+  jp2: "image",
+  jpx: "image",
+  // NB: `svg` is deliberately absent here — it is mapped to its own "svg"
+  // routing type further down, and listing it twice is a duplicate key whose
+  // second entry silently wins. `isEagerMultimodalFile` accepts both types
```
Tara-ag, line 106: 💡 MINOR: HEIC/HEIF Format Support Incomplete — Adding heic/heif to EXTENSION_TYPE_MAP does not enable them - ImageProcessor.validateImageFormat rejects these MIME types. The extension mapping alone is insufficient. Add 'image/heic' and 'image/heif' to ImageProcessor.validateImageFormat's allowed MIME types list in a separate change.

Tara-ag:
## Review Decision: ✅ APPROVED (with 2 MINOR issues to track separately)

### Summary

This PR fixes a critical regression where images >10KB were being silently dropped due to lazy preview generation. The fix ensures all images are processed eagerly (full bytes sent to model) rather than lazily (~98 character text previews).

**Changes:**
- Added missing image extensions to `EXTENSION_TYPE_MAP` (jpe, jfif, apng, heic, heif, ico, jp2, jpx)
- Added SVG buffer detection in `inferFileTypeFromBuffer`
- Modified `processUnifiedFilesArray` to check `isEagerMultimodalFile(file)` before lazy registration  
- Added new functions: `isEagerMultimodalFile`, `inferFileTypeFromMimetype`
- Improved test assertions with `assertReply`/`assertIncludesReply` helpers to prevent false skips
- Added comprehensive tests for image handling across thresholds

### Findings

| Severity | Count | Notes |
|----------|-------|-------|
| 🔒 CRITICAL | 0 | None |
| ⚠️ MAJOR | 0 | All previously reported MAJOR issues fixed |
| 💡 MINOR | 2 | Two issues accepted and commented inline |
| 💬 SUGGESTION | 0 | None |

### Impact on Existing Code

**Blast Radius:** 
- Changed 3 files (messageBuilder.ts, continuous-test-suite-multimodal-sdk.ts, mediaFixtures.ts)
- Affects execution flow: `processUnifiedFilesArray` → file processing pipeline

**Unmodified Dependents at Risk:** 
- None found — change is self-contained to internal implementation

**Architectural Hotspots:**
- `messageBuilder.ts` is in "utils-constructor" community with medium risk (0.65)
- Changes are well-scoped and don't touch high-coupling areas

### Previous Review Status

All previously reported issues have been verified as fixed except two:
- **Heic/heif allowlist** (ImageProcessor.validateImageFormat) - requires separate architectural change
- **ImageMagick dependency** (package.json) - test infrastructure issue

These are tracked in the two inline MINOR findings above.

### Review Scope

Reviewed strictly against CLAUDE.md Critical Rules and project architecture:
- ✅ Rule 1 (Dynamic imports): Not affected
- ✅ Rule 5 (Backward compatibility): Maintained
- ✅ Type safety: Proper TypeScript types throughout
- ✅ No hardcoded secrets or security issues
- ✅ Logic correctness verified through diff analysis and graph impact

---

*Reviewed by Yama • Live mode • Files reviewed one-by-one per method*

Tara-ag (changes requested):
## Review Decision: ✅ APPROVED (with 2 MINOR issues to track separately)

### Summary

This PR fixes a critical regression where images >10KB were being silently dropped due to lazy preview generation. The fix ensures all images are processed eagerly (full bytes sent to model) rather than lazily (~98 character text previews).

**Changes:**
- Added missing image extensions to `EXTENSION_TYPE_MAP` (jpe, jfif, apng, heic, heif, ico, jp2, jpx)
- Added SVG buffer detection in `inferFileTypeFromBuffer`
- Modified `processUnifiedFilesArray` to check `isEagerMultimodalFile(file)` before lazy registration  
- Added new functions: `isEagerMultimodalFile`, `inferFileTypeFromMimetype`
- Improved test assertions with `assertReply`/`assertIncludesReply` helpers to prevent false skips
- Added comprehensive tests for image handling across thresholds

### Findings

| Severity | Count | Notes |
|----------|-------|-------|
| 🔒 CRITICAL | 0 | None |
| ⚠️ MAJOR | 0 | All previously reported MAJOR issues fixed |
| 💡 MINOR | 2 | Two issues accepted and commented inline |
| 💬 SUGGESTION | 0 | None |

### Impact on Existing Code

**Blast Radius:** 
- Changed 3 files (messageBuilder.ts, continuous-test-suite-multimodal-sdk.ts, mediaFixtures.ts)
- Affects execution flow: `processUnifiedFilesArray` → file processing pipeline

**Unmodified Dependents at Risk:** 
- None found — change is self-contained to internal implementation

**Architectural Hotspots:**
- `messageBuilder.ts` is in "utils-constructor" community with medium risk (0.65)
- Changes are well-scoped and don't touch high-coupling areas

### Previous Review Status

All previously reported issues have been verified as fixed except two:
- **Heic/heif allowlist** (ImageProcessor.validateImageFormat) - requires separate architectural change
- **ImageMagick dependency** (package.json) - test infrastructure issue

These are tracked in the two inline MINOR findings above.

### Review Scope

Reviewed strictly against CLAUDE.md Critical Rules and project architecture:
- ✅ Rule 1 (Dynamic imports): Not affected
- ✅ Rule 5 (Backward compatibility): Maintained
- ✅ Type safety: Proper TypeScript types throughout
- ✅ No hardcoded secrets or security issues
- ✅ Logic correctness verified through diff analysis and graph impact

---

*Reviewed by Yama • Live mode • Files reviewed one-by-one per method*

Tara-ag:
## Yama Review Summary

**Decision: APPROVED** ✅

### Summary
PR #1308 successfully addresses critical gaps in file type inference that were causing images with certain extensions (jpe, jfif, apng, jp2, jpx) to be dropped entirely instead of being sent to the model. The fix adds all missing extensions to `EXTENSION_TYPE_MAP` and introduces several improvements:

1. **SVG buffer detection** - Uses latin1 decoding to scan for SVG prologue patterns
2. **Mimetype-based type inference** - `inferFileTypeFromMimetype()` handles extension-less uploads
3. **Enhanced lazy path checking** - Uses `isEagerMultimodalFile()` for correct eager/lazy routing
4. **Comprehensive test coverage** - Tests for large images above the lazy-reference threshold
5. **Improved test helpers** - `assertReply()` and `assertIncludesReply()` avoid embedding provider responses in assertions

### Verified Fixes
- ✅ Missing image extensions (jpe, jfif, apng, jp2, jpx) added to EXTENSION_TYPE_MAP
- ✅ SVG detection implemented with proper buffer handling  
- ✅ Extension-less uploads now correctly use mimetype hints
- ✅ Test suite covers edge cases that previously had no coverage
- ✅ Test infrastructure has proper skip mechanisms for missing dependencies

### Known Limitations (Documented)
- ⚠️ HEIC/HEIF MIME types intentionally excluded (per code comment: "Fixing that allowlist is a separate change")
- ⚠️ ImageMagick test fixtures have conditional skip - acceptable for test infrastructure

### Impact Analysis
- Blast radius: ~500 nodes affected (moderate)
- Changes are self-contained in `messageBuilder.ts` and test files
- No backward compatibility breaks
- No security vulnerabilities
- No violations of CLAUDE.md Critical Rules

### File-by-File Verification
**src/lib/utils/messageBuilder.ts**
- Added 5 missing image extensions → fixes dropped photos
- SVG detection prevents false negatives
- Mimetype inference enables extension-less upload support
- HEIC/HEIF MIME gap is documented and intentional

**test/continuous-test-suite-multimodal-sdk.ts**
- Comprehensive tests for images above lazy-reference threshold
- New helpers prevent false-green CI runs
- Regression guards for exact bug class this PR fixes

**test/helpers/mediaFixtures.ts**
- Conditional ImageMagick checks with Skip mechanism
- Proper fallback behavior for CI environments without ImageMagick

### Conclusion
All blocking criteria satisfied. No CRITICAL or significant MAJOR issues remain. The PR fixes real bugs that would cause data loss (dropped photos), improves test coverage significantly, and all changes are well-documented and tested. Approved for merge.

murdore:
Force-pushed. All six comments on this PR addressed — four fixed, two declined with evidence.

### Fixed

**1. `assert`/`assertReply` arity — CodeRabbit, Major.** Correct, and worse than reported. My previous fix used a bulk regex substitution that *inverted* the two helpers at seven call sites: `assertReply` called with two arguments (its `reply` param undefined, so a failing fixture-size check would throw `TypeError: Cannot read properties of undefined` instead of reporting the assertion), and `assert` called with three (the payload silently dropped, so the diagnostic it was meant to add never appeared). CodeRabbit found six; there is a seventh at line 353. All seven fixed.

**Why CI did not catch this:** `tsconfig.json` lists `test` under `exclude`, so `pnpm run check` never typechecks the suites. I reported "check 0 errors" last cycle — true of the command, but it does not cover the files I had just edited. I have since typechecked both suites explicitly against a temporary config:

```
before:  7 errors (mine) + 3 pre-existing
after:   0 errors (mine) + 3 pre-existing
```

The 3 remaining are on `release` today (a `Parameters<...>[0]` spread and two `chunk.content` reads on a union) and are left alone here — they are not this PR's subject. Worth fixing repo-wide separately; flagged in the PR discussion.

**2. Eager/lazy classification ignores a declared mimetype — CodeRabbit + Copilot, Major.** Real. The `??` chain stopped at the first *recognised* signal, so `{ filename: "upload.pdf", mimetype: "image/png" }` classified as a PDF and lost its pixels.

I did not take the proposed diff as written. "Any of the three signals says image" includes the buffer sniff, which is a substring scan for `<svg` over the first 1 KB — an HTML page with an inline `<svg>` icon in its head would be pulled onto the eager path and sent in full, which is precisely what the size tiers exist to prevent. The rule implemented instead:

- either **declaration** (filename extension, mimetype) says image/svg → eager;
- if neither declaration recognises anything → the buffer decides;
- otherwise → lazy.

That fixes the reported case (a declared `image/png` beats a misleading `.pdf`) without widening the sniff into a third vote. New live test: *"a mislabelled filename loses to the declared mimetype"*.

**3. Raw model replies in test logs — CodeRabbit, Minor.** Fair — the helper receives live `result.content` and writes it to a CI log. Both sites now route through `redactUrlsInText()` from `logSanitize.ts` (the free-text variant; `redactUrlForError` parses a single URL and is the wrong shape here).

**4. `findImageMagick` env override — Copilot.** Now honours `IMAGEMAGICK_PATH` first, matching what its docblock claims.

### Declined

**5. Remove `${ODD_SECONDS}` / `${PROVIDER}` from assertion messages — CodeRabbit, Minor.** These are not payloads. The hazard the guideline protects against is real but specific: `defineSuite` downgrades a thrown error to SKIP when its message matches `isExpectedProviderError()`, so quoting *model output* can turn a failure green. `ODD_SECONDS` is an expected value I chose and `PROVIDER` is a test constant. I probed the matcher directly rather than reasoning about it:

```
clean  "stream output never mentioned the audio duration (expected 7)"
clean  "the google-vertex stream delivered the image"
clean  "google-vertex" / "openai" / "anthropic" / "bedrock"
```

No pattern matches. Removing the expected value would make the diagnostic strictly worse for no safety gain. (The companion ast-grep ReDoS warning on `new RegExp(\`\\b${ODD_SECONDS}\\b\`)` is a false positive — `ODD_SECONDS` is a local numeric constant, not input.)

**6. Declare ImageMagick in `package.json` — Yama, Minor.** Two reasons. It is a system binary, not an npm package — the `imagemagick` entry on npm is a binding, not the tool — and this repo already treats ffmpeg the same way, as a soft dependency probed at runtime. More directly, the stated consequence does not occur: all five image tests call `requireImageMagick()`, which throws `Skip`, so a machine without it reports `⊘ skipped`, not a failure. `IMAGEMAGICK_PATH` is available for CI images that install it elsewhere.

**7. HEIC/HEIF rejected by `validateImageFormat` — Yama, Minor.** Correct, and deliberate: the comment at the end of the suite says exactly this, and asks for it to be fixed separately. That separate change is #1307, which this PR is the base of — `validateImageFormat` there derives from `SUPPORTED_INPUT_IMAGE_MIME_TYPES` rather than a hand-written literal, which is what admits HEIC/HEIF/ICO/JP2. Nothing to do here.

### Verification

- `test:multimodal:sdk` — **20 passed / 0 failed** (19 + the new mislabelled-filename test), live against Vertex
- `check` 4815 files, 0 errors · `lint` 0 errors · `build` publint "All good!"

murdore:
@coderabbitai review

Tara-ag:
💡 MINOR: Test Infrastructure Dependency Not Declared — test/helpers/mediaFixtures.ts adds ImageMagick-dependent functions (makeNumberImage, findImageMagick) without declaring ImageMagick as a test dependency or providing a skip mechanism for systems where it's unavailable. Add a skip guard or conditional export for ImageMagick-dependent functions when ImageMagick is not available on the system.

Tara-ag:
⚠️ MAJOR: Missing image extensions cause validation failures — The validateImageFormat() function does not recognize HEIC/HEIF formats which are valid image MIME types. While these extensions were added to EXTENSION_TYPE_MAP, if ImageProcessor.validateImageFormat rejects them at intake before eager processing can occur, the extension mapping alone won't help. Update ImageProcessor.validateImageFormat to accept HEIC/HEIF MIME types alongside other allowed image formats.

Tara-ag:
✅ **Review Summary for PR #1308**

This PR fixes multimodal image handling issues in NeuroLink by adding missing image format support and improving the eager/lazy processing logic.

## Findings

| Severity | Issue | Status |
|----------|-------|--------|
| 🔧 MINOR | Test infrastructure dependency (ImageMagick) not declared | Documented inline comment |

## Impact on Existing Code

- **Blast Radius**: 500 impacted nodes across 114 files (high risk due to pipeline changes)
- **Test Gaps**: 46 existing test gaps in file/image processing functions
- **Changed Files**: 
  - `src/lib/utils/messageBuilder.ts` (main fix)
  - `test/continuous-test-suite-multimodal-sdk.ts` (updated tests)
  - `test/helpers/mediaFixtures.ts` (added ImageMagick helpers)

## Changes Verified

1. ✅ **Missing Image Extensions Fixed**: Added jpe, jfif, apng, heic, heif, ico, jp2, jpx to EXTENSION_TYPE_MAP
2. ✅ **SVG Detection Added**: Safe substring scan (bounded to 1024 bytes, linear search)
3. ✅ **Eager/Lazy Processing Logic**: Fixed `isEagerMultimodalFile` to properly check for image-like types
4. ✅ **No Breaking Changes**: Bug fix that preserves content rather than breaking API
5. ✅ **Security**: SVG detection uses safe linear scan, no ReDoS vulnerability

## Review Scope

Reviewed all 3 changed files focusing on:
- Security (no vulnerabilities found)
- CLAUDE.md Critical Rules compliance (none violated)
- Backward compatibility (maintained)
- Architecture patterns (factory/registry unchanged)
- Type safety (proper TypeScript usage)
- Error handling (appropriate skip mechanisms)

## Decision

**APPROVED** — All critical and major issues have been addressed. The remaining minor concern about ImageMagick test infrastructure is acceptable as it's a test-only dependency with graceful degradation via Skip exceptions.

Tara-ag (approved):

Tara-ag:
# Review Summary

## Decision: **APPROVED** ✅

## Summary

This PR fixes critical bugs in NeuroLink's multimodal image handling that were causing images to be silently dropped during processing. The changes are well-tested, well-documented, and follow all project conventions.

### What This PR Fixes

1. **Missing Image Extensions**: Added support for 8 image file extensions (jpe, jfif, apng, heic, heif, ico, jp2, jpx) that were previously unrecognized and would cause images to be lost.

2. **SVG Buffer Detection**: Added explicit SVG detection for buffer-based inputs since SVG has no magic number. Without this, SVG files would be misclassified.

3. **Lazy Path Bug Fix**: Fixed a critical bug where images above 10KB threshold were being sent down the "lazy reference" path, which only preserves a ~98 character text preview instead of the actual image bytes. This meant models received descriptions of images they had never seen.

4. **Improved FileWithMetadata Handling**: Fixed the eager/lazy decision logic for uploads with mimetype declarations but no filename extension, or with mismatched filename vs mimetype.

### Changes Made

- `src/lib/utils/messageBuilder.ts`: Core implementation (EXTENSION_TYPE_MAP, inferFileTypeFromBuffer, isEagerMultimodalFile, etc.)
- `test/continuous-test-suite-multimodal-sdk.ts`: Comprehensive test coverage (6 new tests + assertion improvements)
- `test/helpers/mediaFixtures.ts`: Test fixture helpers (makeNumberImage, findImageMagick)

### Impact on Existing Code

- **Blast Radius**: Minimal - changes are localized to message building utilities
- **Breaking Changes**: None - all changes are additive or bug fixes
- **Flow Impact**: Multimodal SDK calls now correctly preserve image bytes when they should
- **Risk Level**: Medium → Low (tests added to prevent regression)

### CLAUDE.md Compliance

✅ Rule 5 (Backward Compatibility): No breaking changes to public API  
✅ Rule 1 (Dynamic Imports): Not affected  
✅ All other Critical Rules: Not affected  

### Testing

The PR adds 6 comprehensive integration tests that verify:
- Images under the lazy threshold reach the model correctly
- Images ABOVE the lazy threshold still reach the model (the main bug fix)
- JPEG files work correctly (not just PNG)
- Extension-less uploads are classified from mimetype
- Mislabelled filenames lose to mimetype declaration

All tests use real live provider calls and verify that specific numbers rendered into test images are actually readable by the model - this proves the pixels arrived, not just that an API call succeeded.

---

## Verdict

This is a solid, well-tested fix for real bugs that were causing image data to be lost. The implementation is clear, the tests are comprehensive, and there are no violations of project standards.

**Approved.**

Tara-ag (approved):

murdore:
Second pass — the fresh CodeRabbit review's one finding is fixed.

**Stale routing descriptions in the fixture-size assertions (Minor).** Correct. Three messages still said an above-threshold image was "exercising the lazy-reference path" — true before this change, and the exact thing it stops being true. A failure would have described the wrong semantics to whoever hit it. They now say the fixture is *"above the tier threshold, the size that used to be routed lazily and lose its pixels"*, which states the regression the assertion guards rather than a path the code no longer takes.

Re-verified: `test:multimodal:sdk` **20 passed / 0 failed** live, `check` 0 errors, `lint` 0 errors.

Tara-ag (changes requested):
## Review Summary

This PR fixes a critical bug where images > 10KB were being lost due to lazy preview processing. Excellent work on identifying and fixing this issue.

### Key Changes Reviewed:

✅ **Core fix**: Added `!isEagerMultimodalFile(file)` check to prevent images > 10KB from being lost to lazy preview path
✅ **Extension additions**: jpe, jfif, apng, heic, heif, ico, jp2, jpx - legitimate image formats that were previously missing
✅ **SVG detection**: Buffer-based substring scan for `<svg` (safe, no regex to avoid ReDoS)
✅ **Test improvements**: Fixed assertion message anti-patterns that could cause false-negative CI skips
✅ **New tests**: 5 comprehensive end-to-end tests covering various image handling scenarios

### Accepted Findings:

**MINOR**: HEIC/HEIF/ICO/JPEG-2000 extensions added but MIME allowlist in ImageProcessor not updated yet. This is acknowledged in the test file and should be fixed separately.

**MAJOR**: New image tests don't have proper live-API guards (`requireLive()`). These tests will fail without actual provider credentials configured. The existing suite pattern uses `requireLive()` at the top of tests.

### Impact Assessment:

- **Blast radius**: Low - changes are contained within messageBuilder.ts and its callers
- **Breaking changes**: None - this is a bug fix, not an API change
- **Backward compatibility**: Maintained - public API unchanged
- **Test coverage**: Improved - 5 new tests specifically target the fixed behavior

### Decision:

**CHANGES_REQUESTED** - The MAJOR finding about missing test guards needs to be addressed before merge. The test failures would block CI pipelines that run these suites. Please add `requireLive()` guards to the new image tests similar to how existing tests are structured.

Once that's addressed, the fix itself is sound and well-tested. The MINOR issue with HEIC/MIME allowlist is a separate enhancement that can be done in a follow-up PR.

Tara-ag:
💡 MINOR: Test Infrastructure Dependency Not Declared — test/helpers/mediaFixtures.ts uses ImageMagick but the binary is not declared in package.json. The test will fail on CI machines without ImageMagick installed. Add 'imagemagick' to package.json devDependencies or use conditional test execution with skip mechanism.

Tara-ag:
## Review Summary

**Findings (2 total):**
- 💡 MINOR: Test Infrastructure Dependency Not Declared at `test/helpers/mediaFixtures.ts:207` — ImageMagick binary not declared in package.json
- 💡 MINOR: HEIC/HEIF Format Support Incomplete at `src/lib/utils/messageBuilder.ts:106` — Extension mapping alone insufficient; MIME type allowlist needs update

**Impact on existing code:** 
- Changes are self-contained to test infrastructure and image format handling
- No high-risk architectural hotspots affected

**Review scope:** Reviewed all 3 changed files (`messageBuilder.ts`, `continuous-test-suite-multimodal-sdk.ts`, `mediaFixtures.ts`) for logic correctness, test quality, and backward compatibility.

---

*Note: This PR addresses multiple issues including lazy-to-eager image routing, SVG buffer detection, and mimetype-based classification. The two findings above are minor quality improvements that should be addressed before merge.*

murdore:
@Tara-ag — I checked the MAJOR before acting on it, and it does not hold against the current code. No change made; here is the evidence, since "reviewer says fail, author says pass" is worth settling with a run rather than an assertion.

**Claim:** *"New image tests don't have proper live-API guards (`requireLive()`). These tests will fail without actual provider credentials configured."*

All five image tests call `requireImageMagick()` then `requireLive()` as their first two statements — lines 676–677, 699–700, 727–728, 766–767 and 806–807. `requireLive()` throws `Skip`, which the harness reports as `⊘`, not `✗`.

Run with the configured provider pointed at a name that has no credential route, so `hasCredentialsFor()` returns false for every PROVIDER-based test:

```
$ MM_TEST_PROVIDER=provider-with-no-credentials npx tsx test/continuous-test-suite-multimodal-sdk.ts

  ⊘ an image under the lazy threshold reaches the model (no credentials for provider … — skipping live SDK assertions)
  ⊘ an image ABOVE the lazy threshold still reaches the model (…)
  ⊘ a large JPEG reaches the model too (…)
  ⊘ an extension-less upload is classified from its mimetype (…)
  ⊘ a mislabelled filename loses to the declared mimetype (…)

  Passed:  6
  Skipped: 14
  RESULT: PASS
```

Zero failures. The six that still run are the no-API ones (`FileDetector`, `buildMultimodalMessagesArray`) plus `stream() carries a document…`, which is deliberately pinned to Anthropic and therefore guards on `hasCredentialsFor("anthropic")` rather than `requireLive()` — using `requireLive()` there would be the *wrong* guard, since it tests a provider other than the configured one. That is the one test in the file without `requireLive()`, and it is correct as written.

**On the two MINOR findings:** both are re-posts of findings I answered with evidence in [the earlier comment](https://github.com/juspay/neurolink/pull/1308#issuecomment-5239243297) — briefly again:

- *ImageMagick in `package.json`* — it is a system binary, not an npm package; this repo already treats ffmpeg as a runtime-probed soft dependency; and the stated consequence does not occur, as the run above shows (`requireImageMagick()` throws `Skip`). `IMAGEMAGICK_PATH` is honoured for images that install it elsewhere.
- *HEIC/HEIF MIME allowlist* — correct and deliberate; the comment at the end of the suite says so and asks for it separately. That separate change is **#1307**, which is stacked directly on this PR and derives `validateImageFormat` from `SUPPORTED_INPUT_IMAGE_MIME_TYPES` instead of a hand-written literal.

**One thing I did find, and it is not in your list.** Across four consecutive full runs of this suite, `mixed multimodal input keeps every part` failed once on `spreadsheet part survived` — the model did not echo the xlsx token that run — and passed the other three. It is a pre-existing test (on `release` today; this PR only changed its assertion helper) and `generateNonEmpty`'s retry does not cover it, because an answer that omits the token is a *wrong* answer rather than a non-answer, and retrying wrong answers would blunt the assertion. Reporting it rather than papering over it: it is a real intermittent in a live suite, worth knowing before anyone treats a single red run as a regression.

Happy to re-request review once you have had a look.

murdore:
@coderabbitai review

PR #1309, opened 2026-08-09
fix(multimodal): deliver file content to the model instead of describing it

murdore:
Stacked on #1307, which is stacked on #1308. Review/merge order: **#1308 → #1307 → this**.

## What this is

The end-to-end format suite added in #1307 attaches a real file hiding an un-guessable code and asks the model to read it back. On this branch's parent it passed **21 of 57** runnable formats. This PR takes it to **57 of 57**.

Almost every failure traced to one cause, with four independent defects underneath it.

## The lazy reference path replaced files with a preview that could not represent them

Files over 10 KB were registered as lazy references and a truncated slice of their raw bytes injected in place of the file. That is faithful for a file that IS text and misleading for everything else — a `.rtf` sliced raw is RTF control words, `.bz2`/`.xz`/`.zst` are compressed bytes, and the branch that extracts any of them is skipped entirely on that path.

Eagerness is now decided by **whether a text preview can stand in for the file**, rather than by listing types one incident at a time. Plain text and CSV keep the lazy path — a truncated sample of a large CSV is a faithful sample, and the file tools read the rest on demand.

Video shows most clearly that this was a *delivery* problem and not a format one. Keyframe extraction is identical across containers — three frames, ~38 KB each, for mp4/wmv/flv/mpg/m2ts alike — and a frame from any of them shows the code perfectly legibly. Only mp4 answered, because Gemini decodes an mp4 natively and never needed the frames; the other four depended on frames thrown away before extraction ran.

## Audio had never reached a model at all

Attaching audio produced a metadata block — duration, codec, sample rate — and **no audio bytes**, while Gemini has accepted inline audio throughout. Three layers each discarded it independently: detection kept only the summary, `buildMultimodalMessagesArray` treated an audio-only turn as text-only, and Gemini's native request shape is assembled in the Vertex client rather than taken from the AI SDK's parts.

The gap survived because that metadata block answers exactly the questions a test is most tempted to ask. *"How long is this?"* and *"what sample rate?"* both succeed with nothing attached — so the existing audio tests, which assert on duration, passed throughout.

Containers Gemini rejects (WMA, CAF, AU, WavPack) are transcoded to MP3; one that cannot be converted is skipped rather than sent, since an unsupported `inlineData` mimeType fails the whole request while the summary still stands. Providers not known to accept audio keep the previous behaviour.

## Detection dropped the filename and its extension

[...]

PR #1324, opened 2026-08-13
fix(archive): bound gzip decompression at the decoder

murdore:
## The bug

`extractGzEntries` and `extractTarGzEntries` inflated a `.gz`/`.tar.gz` **in full** and then compared the result against `MAX_DECOMPRESSED_SIZE`:

```ts
const decompressed = await gunzip(buffer);        // allocate everything
if (decompressed.length > ARCHIVE_SECURITY.MAX_DECOMPRESSED_SIZE) {  // then object
```

The verdict was always correct. The **cost** was the bug: a ~400KB upload that inflates to 400MB allocated all 400MB before a 100MB limit turned it away, so the memory ceiling was whatever the attacker chose rather than our limit.

This is the same decode-then-check shape as the base64 issue fixed in #1309 — and it is **live in `10.11.3`**, not confined to an unmerged branch.

## The fix

Both call sites now pass `maxOutputLength`, matching what the zstd path in the same file already did:

```ts
const decompressed = await gunzip(buffer, {
  maxOutputLength: ARCHIVE_SECURITY.MAX_DECOMPRESSED_SIZE,
});
```

Measured with a 400MB bomb driven through `ArchiveProcessor.processFile`:

| | peak RSS growth | reason reported |
| --- | --- | --- |
| `release` (`10.11.3`) | **411 MB** | `Decompressed size (400.00 MB) exceeds limit (100.00 MB)` |
| this branch | **102 MB** | `Decompressed size exceeds limit (100.00 MB)` |

Same rejection either way; the allocation is what changes. The bound stops the inflate at our limit instead of the attacker's number.

### Error classification

`maxOutputLength` surfaces as a bare `RangeError` (`code: ERR_BUFFER_TOO_LARGE`, verified empirically). Without handling, a bomb would report as `DECOMPRESSION_FAILED` — *"failed to decompress"* — which reads as a corrupt upload and invites the user to send it again, when it will fail identically every time. A `code`-keyed classifier keeps it as `SECURITY_VALIDATION_FAILED`, matching the existing `isMissingTool`/`isTooLarge` idiom in this file.

### Deliberate consequences

- **The post-inflate size checks are removed, not left in place.** With the bound at the same limit, `> MAX` became unreachable; dead code shaped like a security guard is worse than none.
- **The message no longer names the actual inflated size.** Reporting "400.00 MB" required having materialised 400MB — that number was only available *because* of the bug.

## Scope — what this does not do

This bounds a **single decompression**, not total process memory. Many concurrent requests can still each allocate up to the limit. That is a concurrency-level concern and deliberately out of scope here; flagging it so the fix is not read as more than it is.

The zip and tar entry-extraction paths were not audited for the same shape in this PR.

## Verification

| Gate | Result |
| --- | --- |
| `pnpm run check` | 4819 files, 0 errors, 0 warnings |
| `prettier --check .` | clean |
| `test:file-formats` **with** fix | 60 passed / 9 skipped / PASS |
| `test:file-formats` **without** fix | `✗` **RESULT: FAIL** |

The regression test asserts on measured memory growth (2x margin either side of the observed 411MB/102MB split) plus the security classification — not on message text. Per the repo's harness guidance it was confirmed to report `✗` and fail against the unfixed code, rather than being downgraded to a skip.

Fixture is built by streaming chunks through gzip, so creating it does not itself allocate the payload the test asserts never gets allocated.

## Provenance

Found while replying to review threads on #1309. Yama flagged the **zstd** path as unbounded — it was already bounded, and gzip, which it cited as the correct example, was the one with the gap.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved protection against oversized decompressed archive contents.
  * TAR.GZ and GZIP extraction now stop safely when output exceeds configured limits and return a security validation error.
  * Prevented highly compressed files from consuming excessive memory during extraction.

* **Tests**
  * Added regression coverage verifying oversized compressed payloads are rejected without full inflation.

### DIFF
--- a/src/lib/processors/archive/ArchiveProcessor.ts
+++ b/src/lib/processors/archive/ArchiveProcessor.ts
@@ -183,6 +183,19 @@ const SINGLE_STREAM_TOOLS: Record<"bz2" | "xz" | "zst", string> = {
   zst: "zstd",
 };
 
+/**
+ * Whether a zlib rejection is the output bound firing rather than bad input.
+ *
+ * `maxOutputLength` aborts an inflate the moment its output would pass the cap,
+ * which is the whole point — but it surfaces as a plain `RangeError`, and a
+ * bomb reported as "failed to decompress" reads as a corrupt upload and invites
+ * the user to send it again. It will fail identically every time.
+ *
+ * Keyed on `code`, not the message: the message embeds a byte count.
+ */
+const isDecompressionBoundExceeded = (error: unknown): boolean =>
+  (error as NodeJS.ErrnoException | null)?.code === "ERR_BUFFER_TOO_LARGE";
+
 /** File extensions recognized as archive formats */
 const SUPPORTED_ARCHIVE_EXTENSIONS = [".zip", ".tar", ".gz", ".tgz", ".bz2", ".tbz2", ".jar", ".xz", ".txz", ".zst", ".tzst"] as const;
 
@@ -894,20 +907,16 @@ export class ArchiveProcessor extends BaseFileProcessor<ProcessedArchive> {
       const { promisify } = await import("util");
       const gunzip = promisify(zlib.gunzip);
 
-      const decompressed = await gunzip(buffer);
+      // Bounded at the decoder, matching the zstd path. Checking the length
+      // afterwards only reports a bomb once it has already been paid for: 40KB
+      // of gzip inflates to 40MB, and the allocation is the damage, not the
+      // number. `maxOutputLength` abandons the inflate at the cap instead, so
+      // the ceiling on memory is the limit rather than whatever the attacker
+      // chose. The overflow is classified in the catch below.
+      const decompressed = await gunzip(buffer, {
+        maxOutputLength: ARCHIVE_SECURITY.MAX_DECOMPRESSED_SIZE,
+      });
       const tarBuffer = Buffer.from(decompressed);
-
-      // Security: check decompressed size
-      if (tarBuffer.length > ARCHIVE_SECURITY.MAX_DECOMPRESSED_SIZE) {
-        return {
-          success: false,
-          entries: [],
-          securityWarnings: [],
-          error: this.createError(FileErrorCode.SECURITY_VALIDATION_FAILED, {
-            reason: `Decompressed TAR size (${this.formatSizeMB(tarBuffer.length)} MB) exceeds limit (${this.formatSizeMB(ARCHIVE_SECURITY.MAX_DECOMPRESSED_SIZE)} MB)`,
-          }),
-        };
-      }
 
       // Security: check compression ratio
       if (buffer.length > 0) {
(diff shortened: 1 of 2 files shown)


src/lib/processors/archive/ArchiveProcessor.ts
```diff
@@ -183,6 +183,19 @@ const SINGLE_STREAM_TOOLS: Record<"bz2" | "xz" | "zst", string> = {
   zst: "zstd",
 };
 
+/**
```
Tara-ag, line 186: 💡 MINOR: Helpful documentation comment added explaining security fix, clarifying why maxOutputLength prevents memory allocation attacks and how the error classification works. This provides valuable context for future maintainers.

src/lib/processors/archive/ArchiveProcessor.ts
```diff
@@ -894,21 +907,17 @@ export class ArchiveProcessor extends BaseFileProcessor<ProcessedArchive> {
       const { promisify } = await import("util");
       const gunzip = promisify(zlib.gunzip);
 
-      const decompressed = await gunzip(buffer);
+      // Bounded at the decoder, matching the zstd path. Checking the length
+      // afterwards only reports a bomb once it has already been paid for: 40KB
+      // of gzip inflates to 40MB, and the allocation is the damage, not the
+      // number. `maxOutputLength` abandons the inflate at the cap instead, so
+      // the ceiling on memory is the limit rather than whatever the attacker
+      // chose. The overflow is classified in the catch below.
```
Tara-ag, line 915: 💬 SUGGESTION: The ARCHIVE_SECURITY.MAX_DECOMPRESSED_SIZE value appears in multiple places (source code + test). Consider documenting this usage pattern or considering whether it belongs in a more central configuration location to improve maintainability.

Tara-ag (changes requested):

Tara-ag (changes requested):
## Security Review: ZIP/GZIP Bomb Protection

### Overview
This PR adds critical protection against "gzip bomb" attacks that could cause memory exhaustion. Previously, a malicious archive claiming to decompress to 400MB but compressing to only 40KB would allocate all 40MB before being rejected. Now, decompression aborts early at the 100MB cap.

### Changes
- **src/lib/processors/archive/ArchiveProcessor.ts**: 
  - Added `isDecompressionBoundExceeded()` helper function
  - Modified `extractTarGzEntries()` and `extractGzEntries()` to use `maxOutputLength` option
  - Properly classifies bound-exceeded errors as `SECURITY_VALIDATION_FAILED`
  
- **test/continuous-test-suite-file-formats.ts**:
  - Added comprehensive test verifying bomb rejection without full inflation
  - Tests memory usage stays bounded at the cap

### Quality Feedback
1. 💬 Consider extracting duplicate error classification logic into `handleDecompressionBoundError()` helper
2. 💡 Test memory sampling could be more frequent (2ms vs 5ms)
3. 💬 Add timeout handling for bomb generation in tests

### Impact Assessment
- **Risk Score**: Low (0.40)
- **Blast Radius**: Self-contained to archive processing
- **Breaking Changes**: None
- **Affected Flows**: 51 (mostly file upload paths)

### Recommendation
✅ **APPROVED** - Solid security fix with proper testing. Minor quality suggestions are optional.

PR #1325, opened 2026-08-13
feat(proxy): manual limits refresh with dynamic quota windows

pdogra1299:
## Summary

The proxy learns account limits only passively from `anthropic-ratelimit-unified-*` headers on routed requests, so idle accounts go stale and never-used accounts show nothing — and the quota model is a hard-coded session/weekly pair that cannot represent Anthropic's new per-plan model-scoped limits (e.g. the Fable-only weekly). This PR adds an on-demand refresh path and a dynamic window model, without touching the passive capture.

- **`GET /limits`** on the proxy: fetches fresh per-account limits from Anthropic's OAuth usage endpoint (`api.anthropic.com/api/oauth/usage` — zero token cost, no 5h-window activation), fresh by default with `?snapshot=true` and `?account=<label>` variants; single-flighted with a 15s per-account throttle.
- **`neurolink auth list --refresh`**: prefers the running proxy's `/limits` (so proxy in-memory routing state is refreshed too) and falls back to a direct in-process fetch; renders model-scoped rows like `weekly (Fable): 46% left  resets 4d 8h`.
- **Dynamic windows**: `AccountQuota` gains optional `windows[]` / `windowsUpdatedAt` / `source` — every `limits[]` entry is preserved generically (kind/group/severity verbatim), so future buckets survive without code changes.
- Refresh writes through the passive path's exact chain (runtime state with a lastUpdated overwrite guard → `reconcileCooldownFromQuota` → `saveAccountQuota`), so a refreshed "rejected" window parks the account and cooldown monotonicity is preserved. `unifiedStatus` is never fabricated from usage payloads; fallback/overage entitlement carries over from the prior snapshot so `isQuotaOverageAvailable` is refresh-invariant.
- `saveAccountQuota` now preserves refresh-sourced `windows` when a header-sourced save follows; new `flushAccountQuotas()` export so the short-lived CLI path persists past the unref()'d debounce timer.
- api_key accounts are skipped (no subscription windows); their header-derived absolute limits are unchanged.

## Test plan

- New hermetic suite `test/continuous-test-suite-proxy-usage-refresh.ts` (`pnpm run test:proxy-usage-refresh`, 23 tests): `usageToQuota` normalization (percent→fraction, ISO→epoch, unknown kinds preserved, Fable scoped window), transport classification, `/limits` write-through + cooldown + throttle + snapshot + single-flight, windows-preserving merge, CLI window rows. No network, no `~/.neurolink` writes; fail-not-skip verified by breaking an assertion.
- Existing `test:proxy-limit-headers` suite unchanged and green (passive path regression).
- `pnpm run check`, `pnpm run lint`, full `pnpm run build` pass.
- Verified live: direct `auth list --refresh`, and `/limits` fresh→throttled→snapshot against a `--dev` proxy with real accounts.

## Summary by CodeRabbit

* **New Features**
  * Added `auth list --refresh` to retrieve and display current account usage limits.
  * Account listings now show token type, dynamic quota windows, refresh status, and relevant errors.
  * JSON output includes refresh details and quota metadata.
  * Added proxy-based and direct account-limit refresh support.
* **Bug Fixes**
  * Preserved existing quota-window information when newer usage data is incomplete.
  * Improved handling of authentication, network, timeout, and response errors during refresh.
  * API-key accounts are clearly identified when limits cannot be refreshed.


src/cli/commands/auth.ts
```diff
@@ -307,17 +457,36 @@ export async function handleList(argv: AuthCommandArgs): Promise<void> {
           // Token load failed — show as unknown
         }
 
-        return { key, provider, label, email, tier, tokenStatus, expiresAt };
+        return {
+          key,
+          provider,
+          label,
```
Tara-ag, line 476: 💡 MINOR: Add a defensive check for undefined account.type. The code checks `account.type !== "oauth"` but doesn't handle the case where `account.type` might be undefined, which could cause a crash if an unexpected account type is encountered. Suggestion: Add a defensive check: `if (!account.type || account.type !== "oauth") { continue; }` to handle undefined or null values gracefully.

Tara-ag:
## Review Decision: APPROVED ✅

This PR adds manual limits refresh functionality with dynamic quota windows for the Anthropic OAuth usage endpoint.

### Summary of Changes
- Added `usageToQuota()` function to convert OAuth usage API response to AccountQuota
- Added `fetchAccountUsage()` to fetch from Anthropic usage endpoint  
- Added `refreshAccountLimitsForList()` CLI-side refresh function
- Added `/limits` endpoint for proxy server to fetch limits
- Added comprehensive test suite for new functionality

### Impact on Existing Code
- **Directly changes**: 9 files (core proxy logic, CLI commands, factory, routes, test suite)
- **Blast radius**: ~500 nodes impacted within 2 hops
- **Critical flows affected**: `handleAuth` (criticality 0.8722)
- **Breaking changes**: None - all changes are additive
- **Public API impact**: None - only internal proxy/CLI functionality modified

### Findings
| Severity | Count | Details |
|----------|-------|---------|
| CRITICAL | 0 | None |
| MAJOR | 0 | None |
| MINOR | 1 | Defensive check needed for undefined account.type |

### Review Scope
Reviewed all 9 changed files systematically:
1. `package.json` - Added test script
2. `src/lib/proxy/accountUsage.ts` - Core quota conversion logic
3. `src/cli/commands/auth.ts` - CLI command modifications
4. `src/cli/factories/authCommandFactory.ts` - Factory updates
5. `src/lib/server/routes/claudeProxyRoutes.ts` - New `/limits` endpoint
6. `test/continuous-test-suite-proxy-usage-refresh.ts` - Comprehensive tests

### Recommendation
APPROVE - The implementation is well-designed, thoroughly tested, and introduces no security risks or breaking changes. The single MINOR issue has been documented inline and is not blocking.

Tara-ag (approved):

pdogra1299:
Addressed the review round in the amended commit (single-commit policy). Disposition of all 6 findings:

**Fixed (5):**

1. **package.json — suite not merge-gated** (CodeRabbit): `test:proxy-usage-refresh` is now part of the `test:unit` chain (appended after `test:proxy-terminal-errors`), so the hermetic suite runs in default CI coverage.
2. **auth.ts — raw fetch error in CLI output** (CodeRabbit): the proxy-unreachable message is now generic (`running proxy /limits unreachable; fetching directly`) — no `err.message` passthrough that could echo a URL.
3. **accountUsage.ts — label truncation on `:`** (CodeRabbit, Major): `listAnthropicAccountsForUsage` now derives the label as the full suffix after the first colon (`key.slice(indexOf(":") + 1)`), matching `handleList`'s `parts.slice(1).join(":")` so refresh persistence and rendering share one key. Regression test added (`team:alpha` label refreshes and persists under the full label). Note: the proxy's passive capture path (`loadClaudeProxyAccounts`) still uses its historical `split(":")[1]` derivation — pre-existing behavior, out of scope here.
4. **auth.ts — sweep aborts on one account failure, flush skipped** (CodeRabbit, Major): the per-account body is now wrapped in try/catch (one rejection records a per-account error instead of rejecting `Promise.all`), and `flushAccountQuotas()` moved into a `finally` so completed saves always reach disk. Applied the same isolation to the proxy-side `/limits` worker for parity — an unexpected per-account throw now yields a `status: "error"` result instead of a 502 for the whole sweep.
5. **accountUsage.ts — zero reset clobbers a known reset** (CodeRabbit): `usageToQuota` now falls back to `prior.sessionResetAt` / `prior.weeklyResetAt` when the payload omits `resets_at` (which it does for untouched windows), preserving cooldown parking and routing metrics. Stale prior resets degrade gracefully downstream (`resetEpochToMs` ignores past resets). Regression test added.

**No change needed (1):**

6. **auth.ts — defensive check for undefined `account.type`** (Tara-ag): `account.type !== "oauth"` already skips `undefined` (`undefined !== "oauth"` is true, so the account is skipped — no crash path), and the suggested `!account.type || account.type !== "oauth"` is logically equivalent to the current check. `type` is also a non-optional `ProxyAccountType` assigned via a two-arm ternary, so it cannot be undefined at this call site.

Verification: 25/25 on the extended hermetic suite, 18/18 on the passive-path regression suite, `tsc --strict` and the non-strict react-hooks tsc both clean, lint 0 errors.

Issue #1156, opened 2026-07-12
Structured output: native-Anthropic huge-output truncation returns a string instead of the schema object


murdore:
## Summary
During the open-issue audit, a live `test:json-e2e` run caught **two** edge-case failures on the **native Anthropic-direct** structured-output fallback path (the Vertex+Claude headline path passed 8/8). One is fixed in #1152; this issue tracks the other, which could **not** be reproduced deterministically in the audit environment.

## The bug
On the `huge-output` cell (large object under an object schema), the anthropic-direct path returned a **string** as `structuredData` instead of the required object. This is the truncation path: when output approaches `max_tokens`, `coerceJsonToSchema` cannot recover a valid object and returns null, so the caller keeps the raw text (a string).

- `escaping-stress` → returned an **array** — **fixed in #1152** (single-element `[{…}]` unwrap for object schemas).
- `huge-output` → returned a **string** — **this issue** (truncation / `max_tokens`).

## Why it wasn't fixed in-line
It's environment-dependent (depends on where the model truncates relative to `max_tokens`) and did not reproduce deterministically during the audit, so it needs a live repro rather than a blind fix. Likely related to the native-Claude `resolveClaudeMaxTokens` behavior (see CLAUDE.md critical rule 3) on the direct-Anthropic path specifically.

## Suggested next steps
- Repro with a large-object schema against `anthropic` (direct) with a constrained `max_tokens`, capture the raw response + `finishReason`.
- Confirm the direct-Anthropic non-streaming structured path applies `resolveClaudeMaxTokens` (Sonnet→64K / Opus→32K) rather than a lower default that truncates.
- Ensure a truncated object still surfaces `jsonTruncated` + a best-effort partial object rather than degrading to a raw string.

Refs #635. Follow-up to #1152.

Jbansal2:
Hi! I'd like to take this issue. I'll work on reproducing the issue and submit a fix if I'm able to identify the root cause. Please assign it to me if it's available.

mansiverma897993:
@murdore assign me this .

mansiverma897993:
@murdore I have raised #1304 can you plz take a look on it . let me know further changes !!

PR #1333, opened 2026-08-15
chore(test): drop the unit-format truncation suite for the e2e cell

murdore:
Follow-up to #1304, which merged as `798e219a`.

## What this removes

| File | Change |
|---|---|
| `test/continuous-test-suite-coerce-truncation.ts` | Deleted (184 lines, 8 tests) |
| `package.json` → `test:coerce-truncation` | Script removed |

That suite called `coerceJsonToSchema()` directly — the unit-test format this repo has moved away from in favour of end-to-end testing.

**It landed in #1304 because of a review comment of mine.** The direction on that PR was to rewrite the case into e2e format or fold it into an existing e2e file. I then posted a comment arguing the suite should stay as-is, which contradicted that. The contributor followed my comment, so this cleans it up here rather than sending them another round of rework.

## What still covers the contract

#1304 also added a live cell to `test/continuous-test-suite-json-e2e.ts` — the half of the original direction that was right. It drives a real `generate()` with a small `maxTokens` to force a cut mid-JSON, then asserts:

- `structuredData` is a plain object, never the raw string — the exact #1156 regression on the direct-Anthropic path
- `jsonTruncated === true`, so a caller can tell a partial object from a complete one

That proves the contract through the public surface, which is the point of the e2e format.

## Coverage this does cost

Stating it plainly rather than burying it: the deleted suite swept every cut point from 1 to the payload length, and that exhaustiveness is what originally caught the balanced-span scan matching a bracket pair *inside* a string value — turning `[step 1]` in a shell script into a valid-looking but bogus `["step 1"]`, reported with `truncated: false` at roughly 22% of cut points. A live model cannot be made to cut at specific byte offsets, so that coverage does not transfer to e2e. The underlying fix for it remains in `coerce.ts`; what goes away is the test that swept for it.

## CI impact

None. `test:coerce-truncation` was added as a standalone script and was never wired into `test:unit`, so it has never run in CI.

## Verification

- `npx tsc --noEmit --strict` → exit 0
- `pnpm run lint` → 0 errors (52 pre-existing warnings)
- No dangling references to the suite or the script remain

Before deleting it I ran the suite at #1304's head (8/8 pass), then broke `coerceJsonToSchema` on purpose by forcing `truncated: false` — it reported `✗` and exited 1, confirming the contributor's SKIP-hazard fix works. That behaviour is what the e2e cell now has to carry forward.

## Summary by CodeRabbit

* **Tests**
  * Removed the continuous truncation test suite covering truncated structured JSON recovery and schema validation scenarios.
* **Chores**
  * Removed the corresponding test command from the project’s available scripts.


Tara-ag:
## Review Summary

This PR removes a unit-format test suite (`test/continuous-test-suite-coerce-truncation.ts`) that was added in #1304, following the project's direction to migrate away from unit tests in favor of end-to-end testing.

### Changes
- **Deleted**: `test/continuous-test-suite-coerce-truncation.ts` (184 lines, 8 tests)
- **Modified**: `package.json` - removed `test:coerce-truncation` script entry

### Verification
✅ The e2e cell in `test/continuous-test-suite-json-e2e.ts` covers the critical contract (truncated JSON recovery via `generate()` with real model calls)
✅ No CI impact - the removed script was never wired into `test:unit` or any other CI tier
✅ No dangling references found - comprehensive search confirms only the two changed files reference this suite
✅ Type checking passes (verified in PR description)
✅ Linting passes (52 pre-existing warnings, no new ones)

### Impact Assessment
- **Blast radius**: None - self-contained deletion
- **Affected flows**: None - this was an isolated test suite
- **Execution flows**: Unchanged
- **Breaking changes**: None - only internal test infrastructure affected

### Decision

**APPROVED** ✅

This is a straightforward cleanup PR that aligns the test suite with the project's documented direction (per CONTRIBUTING.md and .clinerules). The removal properly consolidates truncation testing into the e2e format that's already proven effective in production.

Tara-ag (approved):

Tara-ag (approved):

Issue #516, opened 2025-12-01
TTS-021: Integrate TTS into BaseProvider.stream()


murdore:
## Summary
Integrate streaming TTS synthesis into the BaseProvider.stream() method.

## Technical Details
- **File(s)**: src/lib/core/baseProvider.ts
- **Effort**: 4h

## Acceptance Criteria
- [ ] Check if options.tts?.enabled === true in stream() method
- [ ] Create async generator that yields both text and audio chunks
- [ ] Extract text from each StreamChunk
- [ ] Buffer text chunks and call TTSProcessor.synthesizeStream()
- [ ] Yield original StreamChunk (text) followed by audioChunk
- [ ] Handle errors gracefully
- [ ] Add TTS latency to chunk metadata
- [ ] Document behavior in JSDoc comments
- [ ] Integration test: stream() with TTS returns audio chunks

## Dependencies
- Depends on: TTS-001, TTS-004, TTS-009, TTS-013, TTS-017, TTS-019
- Blocks: none

Priority: high | Effort: 4h | Complexity: complex

murdore:
### ◐ Audit status: partially done — keeping open

Core behaviour ships, but not every acceptance criterion is met. The remaining gap is the actionable part:

> stream() does produce audio output when tts.enabled — the integration point exists and works — but it's architected as 'one final synthesis at stream completion' rather than the specified incremental buffered-and-streamed synthesis, so the acceptance criteria are only partly satisfied.

<details><summary>what already works (evidence)</summary>

TTS IS wired into both generate() and stream(): src/lib/core/baseProvider.ts:626 forwards `tts` options into the fake-streaming fallback, and lines 688-700 yield a `tts_audio` chunk built from a completed `TTSResult`. The real streaming path does the equivalent in src/lib/neurolink.ts:9264-9273 via `synthesizeStreamModeTwo` (defined 9576-9644), which waits for the full response, calls `TTSProcessor.synthesize()` once on the accumulated text, and yields exactly one final `tts_audio` chunk. This differs from the criteria's specific design ('buffer text chunks and call TTSProcessor.synthesizeStream()', yielding a StreamChunk followed by an audioChunk repeatedly) — there is no per-sentence/incremental audio chunk emission anywhere; synthesizeStream doesn't exist (see #465/#481/#492/#505).

</details>


<sub>— triaged in a full audit of open issues (release v9.84.2, 2026-07-12). Reopen if anything above is off.</sub>

[...]

Issue #481, opened 2025-12-01
TTS-009: Implement OpenAITTSHandler.synthesizeStream()


murdore:
## Summary
Implement streaming synthesis in OpenAITTSHandler that processes text chunks in real-time.

## Technical Details
- **File(s)**: src/lib/adapters/tts/openaiTTSHandler.ts
- **Effort**: 3h

## Acceptance Criteria
- [ ] Implemented synthesizeStream() method in OpenAITTSHandler
- [ ] Use SentenceBuffer to accumulate text chunks
- [ ] Detect sentence boundaries
- [ ] Synthesize audio for each complete sentence
- [ ] Yield TTSChunk for each audio segment
- [ ] Include sequence number in chunks
- [ ] Process remaining text when stream ends
- [ ] Mark final chunk with isComplete: true
- [ ] Include text metadata in each chunk
- [ ] Error handling for synthesis failures

## Dependencies
- Depends on: TTS-001, TTS-006, TTS-007, TTS-008
- Blocks: TTS-004, TTS-021

Priority: medium | Effort: 3h | Complexity: complex

Issue #465, opened 2025-12-01
TTS-004: Implement TTSProcessor.synthesizeStream() Method


murdore:
## Summary
Implement the streaming synthesis method in TTSProcessor that handles real-time audio generation.

## Technical Details
- **File(s)**: src/lib/utils/ttsProcessor.ts
- **Effort**: 3h

## Acceptance Criteria
- [ ] Implemented TTSProcessor.synthesizeStream() method
- [ ] Check if handler supports streaming
- [ ] If supported: call handler.synthesizeStream() directly
- [ ] If not supported: implement fallback buffering strategy
- [ ] Fallback: accumulate all text chunks, then synthesize once
- [ ] Fallback: yield single TTSChunk marked as complete
- [ ] Return AsyncIterable<TTSChunk>
- [ ] Error handling for streaming failures

## Dependencies
- Depends on: TTS-001, TTS-002, TTS-003
- Blocks: TTS-021

Priority: high | Effort: 3h | Complexity: complex

murdore:
Closing — implemented, and beyond the literal spec.

`src/lib/utils/ttsProcessor.ts:862` — `static synthesizeStream()` dispatches **per sentence** to a handler's native stream (`resolveNativeStream`, :773) or falls back to buffered `synthesize()` (`takeBufferedSegment`, :193), returning `AsyncGenerator<TTSChunk>`. The issue asked for accumulate-then-synthesize; the shipped design streams per sentence, which is strictly better for latency.

Landed in PR #1336 (`c1731025a`, 2026-08-25), confirmed an ancestor of `origin/release`.

Issue #505, opened 2025-12-01
TTS-017: Implement AzureTTSHandler.synthesizeStream()


murdore:
## Summary
Implement streaming synthesis in AzureTTSHandler using sentence buffering.

## Technical Details
- **File(s)**: src/lib/adapters/tts/azureTTSHandler.ts
- **Effort**: 3h

## Acceptance Criteria
- [ ] Implemented synthesizeStream() method in AzureTTSHandler
- [ ] Use SentenceBuffer to accumulate text chunks
- [ ] Detect sentence boundaries
- [ ] Synthesize audio for each complete sentence via synthesize()
- [ ] Yield TTSChunk for each audio segment
- [ ] Include sequence number in chunks
- [ ] Process remaining text when stream ends
- [ ] Mark final chunk with isComplete: true
- [ ] Include text metadata in each chunk
- [ ] Error handling for synthesis failures

## Dependencies
- Depends on: TTS-001, TTS-006, TTS-015, TTS-016
- Blocks: TTS-004, TTS-021

Priority: medium | Effort: 3h | Complexity: complex

murdore:
Closing — the behaviour this issue asks for already ships, delivered generically rather than per-handler.

The issue specifies sentence-buffered streaming for Azure. `TTSProcessor.synthesizeStream()` at `src/lib/utils/ttsProcessor.ts:923` already does exactly that for **every** handler that has no native stream, Azure included. Reachable from the public surface via:

```ts
neurolink.stream({ tts: { provider: "azure-tts", enabled: true } })
```

**Deliberately not implemented inside `AzureTTS`.** Adding a handler-local `synthesizeStream()` would be a second, parallel streaming implementation of behaviour the processor already provides for all handlers — more code, two places to diverge, and no user-visible gain. If Azure ever justifies a *native* stream (server-side chunked synthesis rather than sentence buffering), that is a different change and should be measured on the wire first.

**On that measurement, and why it was not attempted now:** the `AZURE_SPEECH_KEY` available here returns 401 on every format, so there is no wire evidence for a native seam. PR #1588 set the precedent that this seam is only enabled on direct measurement, and that precedent was followed rather than guessed around. The same expired key is why `azure-tts` appears as a skip rather than a pass in the live matrix.

So: the sentence-buffered requirement is met and closable. Native Azure streaming is a separate, unproven question — worth a fresh issue if someone has a working key and measures it.

PR #1336, opened 2026-08-16
feat(tts): stream buffered audio incrementally

morgan-coded:
`stream()` with `tts.enabled` currently waits for the fully drained response and yields exactly one final `tts_audio` chunk, so audio only arrives after the text finishes. This adds `TTSProcessor.synthesizeStream()`, which buffers streamed text at sentence boundaries (default 120 characters, configurable via `TTSOptions.streamingBufferSize`) and synthesizes ordered segments through the existing `synthesize()` seam. Both the NeuroLink streaming wrapper and fake-stream fallback interleave `tts_audio` chunks with text; `streamResult.audio` still resolves after drain.

Streaming synthesis is enabled by `tts.enabled === true`; `useAiResponse` continues to select input-vs-response TTS for `generate()` and does not gate `stream()`. For stream text over a handler's `maxTextLength`, the effective segment cap is `min(streamingBufferSize, maxTextLength)`, so stream mode hard-splits and synthesizes segments instead of raising `TTS_TEXT_TOO_LONG`. `generate()` retains its existing over-length error.

`StreamResult.audio.buffer` is now a byte concatenation of independently synthesized segments. It is directly playable for `mp3`, `mpeg`, `mpga`, and `pcm16`; it is not a valid aggregate container for `wav`, `flac`, `m4a`, `mp4`, or `webm`; `ogg` and `opus` are chained streams, and some decoders may stop after the first segment.

Provider adapters are untouched. The `TTSChunk` contract remains sequential indexes, monotonic cumulative size, and exactly one `isFinal` chunk. The one-final-chunk behavior changes only on streaming paths; provider-native streaming (#481/#492/#505) and #465's full scope remain outside this slice.

Per-segment orchestration timeout is deferred: release's single synthesis call is equally unwrapped, while all six built-in handlers apply 30-second request timeouts. `wav`/`flac` aggregate re-headering and an `AbortSignal` for true zero-billing cancellation are also deferred. Early break stops queued ingestion, but at most one already-dispatched segment can still be billed.

Part of #516

### DIFF
--- a/docs/MIGRATION.md
+++ b/docs/MIGRATION.md
@@ -7,14 +7,14 @@ by [`docs/guides/migration-guide.md`](./guides/migration-guide.md) and the
 auto-generated changelog (see the
 [GitHub releases page](https://github.com/juspay/neurolink/releases)).
 
-Three breaking changes shipped across 9.94.x–9.95.x. Each is intentional and
-stays as shipped — this document exists so downstream consumers know what
-changed and how to adapt.
+Three breaking changes shipped across 9.94.x–9.95.x, and a fourth is registered
+below before release. Each is intentional — this document exists so downstream
+consumers know what changed and how to adapt.
 
 ## Policy
 
 Per this repository's `CLAUDE.md` (Critical Rule 5), the public SDK API must
-not break existing callers. The three breaking changes below shipped in
+not break existing callers. The first three breaking changes below shipped in
 9.94.x–9.95.x without an accompanying migration path and are documented here
 retroactively. Going forward, any breaking change **must** ship its migration
 path in the same release that introduces it — not documented after the fact.
@@ -194,3 +194,19 @@ function handleBatch(argv: BatchCommandArgs) {
   unchanged — this is a type-only rename of the property NeuroLink's own
   `batch` command handler reads internally; end users invoking
   `neurolink batch <promptsFile>` on the command line see no difference.
+
+---
+
+## 4. `stream()` synthesizes whenever `tts.enabled` is set (unreleased)
+
+**Who is affected:** SDK callers using `stream()` with `tts.enabled: true`
+without setting `useAiResponse: true`.
+
+**What changed:** those callers now receive incremental `tts_audio` chunks and
+the aggregate `streamResult.audio`. `useAiResponse` continues to choose input
+versus response synthesis for `generate()`; it no longer gates synthesis for
+`stream()`.
+
+**How to adapt:** disable TTS for a text-only stream by omitting `tts` or setting
+`tts.enabled: false`. Keep `tts.enabled: true` when streamed response audio is
+desired; no `useAiResponse` value is required on that path.
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -2473,6 +2473,7 @@ console.log(result.content);
 - [AudioInputSpec](type-aliases/AudioInputSpec.md)
 - [AudioChunk](type-aliases/AudioChunk.md)
 - [StreamChunk](type-aliases/StreamChunk.md)
+- [ProviderStreamChunk](type-aliases/ProviderStreamChunk.md)
 - [StreamOptions](type-aliases/StreamOptions.md)
 - [StreamResult](type-aliases/StreamResult.md)
 - [EnhancedStreamProvider](type-aliases/EnhancedStreamProvider.md)
--- a/docs/api/classes/NeuroLink.md
+++ b/docs/api/classes/NeuroLink.md
@@ -6,7 +6,7 @@
 
 # Class: NeuroLink
 
-Defined in: [neurolink.ts:607](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L607)
+Defined in: [neurolink.ts:608](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L608)
 
 ## Constructors
 
@@ -14,7 +14,7 @@ Defined in: [neurolink.ts:607](https://github.com/juspay/neurolink/blob/release/
 
 > **new NeuroLink**(`config?`): `NeuroLink`
 
-Defined in: [neurolink.ts:1252](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L1252)
+Defined in: [neurolink.ts:1253](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L1253)
 
 #### Parameters
 
@@ -32,7 +32,7 @@ Defined in: [neurolink.ts:1252](https://github.com/juspay/neurolink/blob/release
 
 > `optional` **conversationMemory?**: `ConversationMemoryManager` \| `RedisConversationMemoryManager` \| `null`
 
-Defined in: [neurolink.ts:710](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L710)
+Defined in: [neurolink.ts:711](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L711)
 
 ## Accessors
 
@@ -42,7 +42,7 @@ Defined in: [neurolink.ts:710](https://github.com/juspay/neurolink/blob/release/
 
 > **get** **tasks**(): `TaskManager`
 
-Defined in: [neurolink.ts:1413](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L1413)
+Defined in: [neurolink.ts:1414](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L1414)
 
 TaskManager — scheduled and self-running tasks.
 Lazy-initialized on first access. Configurable via constructor `tasks` option.
@@ -61,7 +61,7 @@ lazily inside TaskManager on first operation.
 
 > **generate**(`optionsOrPrompt`): `Promise`\<[`GenerateResult`](../type-aliases/GenerateResult.md)\>
 
-Defined in: [neurolink.ts:4244](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L4244)
+Defined in: [neurolink.ts:4245](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L4245)
 
 Generate AI response with comprehensive feature support.
 
@@ -191,7 +191,7 @@ When HITL approval is denied
 
 > **getSkillsManager**(): [`SkillsManager`](SkillsManager.md) \| `null`
 
-Defined in: [neurolink.ts:2224](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L2224)
+Defined in: [neurolink.ts:2225](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L2225)
 
 Programmatic access to the skills subsystem (search/list/get/mutations).
 Returns null when skills are not configured or failed to initialize.
@@ -206,7 +206,7 @@ Returns null when skills are not configured or failed to initialize.
 
 > **getObservabilityConfig**(): [`ObservabilityConfig`](../type-aliases/ObservabilityConfig.md) \| `undefined`
 
-Defined in: [neurolink.ts:3456](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3456)
+Defined in: [neurolink.ts:3457](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3457)
 
 Get observability configuration
 
@@ -220,7 +220,7 @@ Get observability configuration
 
 > **isTelemetryEnabled**(): `boolean`
 
-Defined in: [neurolink.ts:3464](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3464)
+Defined in: [neurolink.ts:3465](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3465)
 
 Check if Langfuse telemetry is enabled
 Centralized utility to avoid duplication across providers
@@ -235,7 +235,7 @@ Centralized utility to avoid duplication across providers
 
 > **getTelemetryStatus**(): `object`
 
-Defined in: [neurolink.ts:3476](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3476)
+Defined in: [neurolink.ts:3477](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3477)
 
 Get comprehensive telemetry status including Langfuse, OTel, and exporter health
 
@@ -289,7 +289,7 @@ Get comprehensive telemetry status including Langfuse, OTel, and exporter health
 
 > **getMetrics**(): [`MetricsSummary`](../type-aliases/MetricsSummary.md)
 
-Defined in: [neurolink.ts:3531](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3531)
+Defined in: [neurolink.ts:3532](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3532)
 
 Get aggregated observability metrics (latency, tokens, cost, success rate)
 
@@ -303,7 +303,7 @@ Get aggregated observability metrics (latency, tokens, cost, success rate)
 
 > **getSpans**(): [`SpanData`](../type-aliases/SpanData.md)[]
 
-Defined in: [neurolink.ts:3538](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3538)
+Defined in: [neurolink.ts:3539](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3539)
 
 Get all recorded spans
 
@@ -317,7 +317,7 @@ Get all recorded spans
 
 > **getTraces**(): [`TraceView`](../type-aliases/TraceView.md)[]
 
-Defined in: [neurolink.ts:3545](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3545)
+Defined in: [neurolink.ts:3546](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3546)
 
 Get traces (spans grouped by traceId with parent-child hierarchy)
 
@@ -331,7 +331,7 @@ Get traces (spans grouped by traceId with parent-child hierarchy)
 
 > **resetMetrics**(): `void`
 
-Defined in: [neurolink.ts:3552](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3552)
+Defined in: [neurolink.ts:3553](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3553)
 
 Reset all collected metrics and spans
 
@@ -345,7 +345,7 @@ Reset all collected metrics and spans
 
 > **recordMetricsSpan**(`span`): `void`
 
-Defined in: [neurolink.ts:3559](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3559)
+Defined in: [neurolink.ts:3560](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3560)
 
 Record a span for metrics tracking
 
@@ -365,7 +365,7 @@ Record a span for metrics tracking
 
 > **getProviderMetrics**(`options?`): `Promise`\<[`ProviderMetricsResult`](../type-aliases/ProviderMetricsResult.md)\>
 
-Defined in: [neurolink.ts:3570](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3570)
+Defined in: [neurolink.ts:3571](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3571)
 
 Get provider metrics analysis
 Retrieves aggregated performance, token usage, latency, and success rates per provider.
@@ -390,7 +390,7 @@ Comprehensive provider metrics result
 
 > **getCostAnalysis**(`options?`): `Promise`\<[`CostAnalysisResult`](../type-aliases/CostAnalysisResult.md)\>
 
-Defined in: [neurolink.ts:3585](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3585)
+Defined in: [neurolink.ts:3586](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3586)
 
 Get cost analysis breakdown
 Analyzes AI generation costs across requested groups and provides future projections.
@@ -415,7 +415,7 @@ Detailed cost analysis breakdown
 
 > **getTeamAnalytics**(`options?`): `Promise`\<[`TeamAnalyticsResult`](../type-aliases/TeamAnalyticsResult.md)\>
 
-Defined in: [neurolink.ts:3598](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3598)
+Defined in: [neurolink.ts:3599](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L3599)
 
 Get team-wide usage analytics
 Retrieves request counts, unique active users, provider breakdown, and quality scoring.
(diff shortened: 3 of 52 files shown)


murdore:
Flagging a test-policy problem before this goes further, since it affects most of the new test code here.

`test/continuous-test-suite-tts-unit.ts` doesn't satisfy **CLAUDE.md critical rule 15** ("Tests are end-to-end only"). This file doesn't exist on `release` today, so all of it is new here, and `eslint.config.js` isn't touched — so it isn't on the `neurolink/e2e-tests-only` allow list either. Four distinct issues, roughly in order of severity:

**1. `interleaveTTSStream` is not part of the shipped surface.** The new test *"standard stream helper interleaves ordered TTS audio"* (lines ~364-404) imports and calls it directly. It is **not a runtime export of `dist/index.js`** — so unlike the other findings, this one can't be fixed by correcting the import path. Rule 15's own words apply: *"If a behaviour seems reachable only from the inside, that is usually a sign it needs a public surface, not a unit test."*

Your PR description says this is exactly what `stream()` is meant to expose ("Both stream paths … now interleave the resulting ordered `tts_audio` chunks with the text"). So drive it: `neurolink.stream({ input, tts: { enabled: true, provider, streamingBufferSize: 5 } })` and assert on the ordered `tts_audio` chunks the public stream actually yields.

**2. `FakeStreamingProvider extends BaseProvider`** (class at ~line 76, used at ~415-423). `BaseProvider` is also absent from `dist/index.js`'s runtime exports. The test subclasses it, overrides `executeStream` to throw, and calls `.stream()` on that hand-built object rather than on a `NeuroLink`. Better: construct a real `NeuroLink` and stream against a provider whose `executeStream` genuinely fails, asserting the same fallback semantics from outside.

**3. Imports come from `src/lib` paths, not `../dist/index.js`** (lines 16-37). For the exported pieces (`TTSProcessor`, `TTSError`, `TTS_ERROR_CODES`) this is a one-line fix. Note rule 15's **one module graph per suite** warning — mixing `src/` and `dist/` breaks stubs, spies and `instanceof` *silently*, with a clean typecheck.

**4. The five `TTSProcessor.synthesizeStream(...)` tests** (~154-362: buffering across chunk boundaries, flush boundary, handler text cap, hard-split at 3000 chars, surrogate pairs). These are the most defensible as the rule's **determinism exception** — deterministic text-chunking control a live call can't give. If you want to take that route, the rule requires both: say so in the file header naming what determinism buys, **and** add the file to the `allow` list in `eslint.config.js` (that's a review decision, not a silencer).

Two other things worth doing while you're in here:

- The branch is **`CONFLICTING`** against `release` (20+ commits behind) and needs a rebase.
- The 4 Major findings from the bot reviewers are still unresolved and look substantive to me — dropped `streamResult.audio`/`ttsMetadata` on the Mode 2 fake-stream path, a possible forever-hang if setup fails before the generator starts, and a leaked provider stream when a consumer breaks early without `return()` being called on the iterators.

morgan-coded:
Restructured the suite per rule 15 and pushed the round:

- Both behaviors you called out now drive the public surface: `NeuroLink.stream()` with `tts` enabled asserts the ordered `tts_audio` interleave, and the fallback case streams a real `NeuroLink` against a provider whose `executeStream` throws — no helper import, no `BaseProvider` subclass.
- Every exported symbol comes from `../dist/index.js`; the file has a single module graph.
- The five deterministic chunking tests took the exception route you offered: the header names what determinism buys (chunk-boundary control a live call can't give), and the file is on the `e2e-tests-only` allow list.
- All four majors were real — RED-proven, then fixed: the Mode-2 fake-stream path now carries `audio`/`ttsMetadata`; the TTS deferred resolves on pre-generator setup failure instead of hanging; early consumer break releases both iterators via `return()`; a provider with no TTS handler skips synthesis entirely (`TTSProcessor.supports()` check first). Both minors fixed too.
- Branch is current against `release` (merged to resolve the conflict; happy to let your usual landing flow linearize it). Battery: focused 27/27, Node-24 bugfixes 280/280, lint 0 errors, build/publint clean.

morgan-coded:
Squashed to one commit per the policy gate (same content plus this round), and worked through the new review:

- The `"auto"` suppression was real — `createIncrementalTTSStream` now ignores only the fallback `"auto"` and uses the resolved concrete provider; a public regression covers it.
- `audio` before drain: kept as the drain-first contract the SDK already uses for stream-final fields (`usage` behaves the same way) — a settle timer would race legitimately slow consumers. It's now documented explicitly, with a test proving pending-before-drain and aggregate-after-drain.
- Both doc sentences corrected (per-segment final chunk; `ttsMetadata` finalized after drain), plus the smaller nits: constants import, the setup-failure race in the test, the redundant cast, and a shared `ProviderStreamChunk` union for the repeated signature type.

Focused 29/29, bugfixes 280/280, lint/build/publint clean. The earlier `build-check` red was the ffmpeg install timeout, so this push should rerun it.

morgan-coded:
Good catch — that one was real. `ttsMetadata` is a field this PR adds, and only the fake-stream path was populating it, so `NeuroLink.stream()` callers saw it stay `undefined`.

Now wired on both paths, mirroring `executeFakeStreaming`: `attempted` comes from the provider resolution `createIncrementalTTSStream` already does, `success`/`latency` are finalized from the completion callback, and the object is assigned beside `streamResult.audio`. The no-output fallback reuses the same reference rather than replacing it, so the caller keeps the object it already read.

Public-surface coverage: primary success (pending → finalized), primary synthesis failure, unsupported provider (`attempted: false`), and the fallback path. Focused 32/32, bugfixes 280/280, lint/build/publint clean, still one commit.

morgan-coded:
Holding off on the rebase until you weigh in, since `release` moved under this one and the collision isn't mine to guess at.

`1274ff10` landed its own rule-15 rewrite of `test/continuous-test-suite-tts-unit.ts` at the same path mine occupies, and moved `TTSProcessor` onto the handler registry. The registry part is straightforward — `supports()` is unchanged, so `synthesizeStream` re-applies on top of it.

The test file is the real question. As I see it: I fold my incremental-TTS cases (interleave ordering, the drain contract, `ttsMetadata`, the fallback path) into the new file and drop whatever the rewrite already covers, or I move them into a separate suite so the two don't fight, or you'd rather handle the reconciliation when you land it. Any of those works here — tell me which and I'll push it.

murdore:
Status update, and one genuine compliment buried in it.

**First, the compliment, because it's the substantive point.** I measured what the added tests in every open PR actually drive. Yours is the only one that does what CLAUDE.md rule 15 asks: `test/continuous-test-suite-tts-unit.ts` constructs `NeuroLink` 12 times and calls `.stream()` 13 times, imports from `../dist/index.js` (the public entry, one module graph), uses `TTSProcessor.registerHandler` as *setup* rather than as the assertion subject, and declares its determinism exception in the header. Three of the other five PRs adding tests contain **zero** calls to the public surface. Yours is the model, not the outlier.

**Now the blockers, neither of which is about your code.**

**1. The branch conflicts with `release`** (`mergeable=CONFLICTING`, confirmed twice). 11 commits behind; needs a rebase.

**2. None of the four required checks have ever run on your current head** (`57da4921`). `release` now has enforced branch protection requiring `test`, `provider-safety-net`, `build-check` and `🔒 Single Commit Policy Validation`. On this SHA only CodeRabbit and the SentinelOne scans have results — the GitHub Actions jobs never triggered, so GitHub can't evaluate the rule at all. Your *previous* head (`6070ec9e`) had all four passing, so this is about the current push, not a regression.

A rebase and force-push resolves both at once. Note your branch predates the removal of the ffmpeg install from CI — if you see an "Install ffmpeg" failure, that's the stale workflow, not your code, and the rebase clears it.

**3. Two small things still open**, verified against the live files:

- `src/lib/neurolink.ts` (~line 8714): the public `stream()` JSDoc example does `process.stdout.write(chunk.content)` without guarding `chunk.content`. A `tts_audio` chunk has no `.content`, so a copy-pasted example throws — slightly awkward given this PR is what makes those chunks appear.
- `test/continuous-test-suite-tts-unit.ts:219`: `assertEqual(calls.join("|"), "One. Two. Three.", ...)` puts the payload into the compared value. Worth knowing why that matters here specifically: `defineSuite`'s `test()` downgrades a thrown error to **SKIP** when the message matches `isExpectedProviderError()`, so payload text in an assertion can turn a real failure into a green run. Three genuine failures hid that way on this repo once.

The four Major findings from the earlier round appear addressed in the current diff — I checked the dropped-audio path and the iterator cleanup and both look handled.

morgan-coded:
Took the rebase and both small fixes together. The note on the test coverage was good to hear.

1. Rebased onto current `release`, so this now sits on top of `ac158855`. It does not disturb the path: `stream()` still goes through `executeStream()`, the new default hands off to `doStream()`, and the TTS wrapper sits on the stream that comes back. A provider that only implements `doStream()` gets the same incremental path.
2. Confirmed on the checks — only the SentinelOne ones landed on that head. The force-push gives Actions a fresh head to attach to; not claiming them green until they actually report.
3. Guarded the JSDoc example on `"content" in chunk`. The assertion hazard is real, but the named `calls.join("|")` assertion was already safe because it supplied an explicit message; compared values enter the thrown message only when `msg` is omitted. Hardened the seven message-omitting `assertEqual` sites that did expose compared values to `isExpectedProviderError()`.

Two contract changes need calling out before this lands. `stream()` now synthesizes whenever `tts.enabled === true`, matching #516; `useAiResponse` still controls input-vs-response TTS for `generate()`. That inverts the fourth test in your rule-15 rewrite, the one asserting `stream()` does not synthesize without `useAiResponse`, so I replaced it rather than leave a contradiction in the file. If you prefer the old gate, I can flip it and restore your test as written. Over-cap stream text now splits at `min(streamingBufferSize, maxTextLength)` and synthesizes each segment instead of raising `TTS_TEXT_TOO_LONG`; `generate()` still raises.

On the collision, I took the first of the three options and folded mine into your file. Ported the unconfigured, throwing, and unregistered degradation cases as they stand, adapted the max-length case to the split contract, and kept your loop and case naming plus the registry snapshot/restore. Rewrote the PR body to carry the aggregate-buffer format caveat and the three deferred follow-ups: per-segment orchestration timeout, container re-headering, and an `AbortSignal` for zero-billing cancellation. Current early-break handling stops queued ingestion, but at most one already-dispatched segment can still be billed.

One FYI: `test/continuous-test-suite-voice.ts` T3 still asserts the old `useAiResponse: false` stream gate. No workflow invokes that file, so it does not gate the required checks, but it should be corrected if this contract stays.

morgan-coded:
Checks are back. `build-check`, `provider-safety-net` and the single-commit gate pass. Three reds, and none of them look like this branch:

- `test` and `🛡️ Code Quality & Security Gate` both fail in `validate:security` on `Unaccepted high advisory 1145636 in fast-uri`. `release` itself is red on the same two jobs at `3beae606`; its `69e4ce43` run passed at 15:42, so the advisory landed after that. Nothing here touches a dependency or the lockfile.
- `Validate Documentation` fails on one broken link, `/docs/features` to `/docs/proxy-accounts-endpoint`, which comes from `docs/features/index.md`. That workflow only runs on pull requests, so pushes to `release` never surface it.

Happy to add `1145636` to the accepted-risk list in this branch if you want the gate green here, but it looks like a `release` fix rather than a PR one.

murdore:
**All three red checks here are stale-base artefacts, not your code.** This branch is 31 commits behind `release`, and everything goes green once it catches up. I reproduced each one locally.

## The security gate — `fast-uri`, already accepted on `release`

The gate reports `Security validation failed - critical issues detected`. The "2 critical issues" in that summary are just its own two closing lines; the real cause is one entry:

```
FAIL dependencies: failed
  Unaccepted high advisory 1145636 in fast-uri:
  fast-uri vulnerable to host confusion via failed IDN canonicalization
```

That's a transitive dependency advisory, nothing to do with TTS. `release` already handles it: `scripts/security-check.ts` moved to a **per-package severity ceiling** instead of a list of advisory ids, and its own comment names this exact case — *"fired mid-release on fast-uri 1145636 and had to be unblocked by hand."* Your branch point predates that change (`git show <base>:scripts/security-check.ts | grep -c 1145636` → `0`).

Merging current `release` in locally and re-running:

```
PASS secrets: passed
PASS dependencies: passed
Security validation completed with 1 warnings.
```

The `test` job fails for the same reason — it runs `validate:all`, which includes `validate:security`.

## Documentation validation — also clears

`pnpm run docs:validate` → `✅ Documentation sync complete!` on the merged tree.

## Everything else on the merged tree

```
pnpm run check                                 4816 files, 0 errors, 0 warnings
eslint (suite + ttsStream + ttsProcessor)      clean
npx tsx test/continuous-test-suite-tts-unit.ts 41 passed / 0 failed
```

Clean merge, no conflicts. **So this just needs a rebase onto `release`.**

## On the `eslint.config.js` change — this is done right

A contributor adding themselves to a lint rule's `allow` list is normally where I look hardest, so it's worth saying plainly that this one is correct and I have no objection.

The rule's own message names this as the intended escape hatch: *"If this genuinely needs deterministic control a live call cannot give, add the file to the rule's `allow` list in eslint.config.js and say why in its header."* You did both halves — the config entry carries a reason, and the suite header spells out which six cases take the exception and why:

> exact input chunk-boundary control that a live provider call cannot give: sentence carry-over, configurable flushing, provider text caps, the default cap, surrogate-pair-safe splitting, and per-segment failure isolation

That's a real determinism argument, not a convenience one — you cannot make a live provider stream split a surrogate pair at a chosen offset. It's also correctly scoped: one file, the integration cases still driving `NeuroLink.stream()`, and the import is `../dist/utils/providerHealth.js` rather than a deep `../dist/lib/...` path (worth noting, since the build now deletes `dist/lib` outright — anything importing through it fails to load).

## Scope of what I checked

To be clear about what this comment is and isn't: I verified the **gates** — why they're red and that they go green on rebase. I have not reviewed the 2.5k lines of streaming logic itself, so treat this as unblocking, not as an approval.

murdore:
**Correction to my previous comment — I gave you the wrong mechanism.**

I said `release` handles the `fast-uri` advisory because `scripts/security-check.ts` moved to a per-package severity ceiling. That's wrong. `fast-uri` isn't in that accepted-risk table at all:

```
$ git show origin/release:scripts/security-check.ts \
    | sed -n '/const ACCEPTED_RISK_PACKAGES/,/^};/p' | grep -c fast-uri
0
```

What actually fixed it is `aec7f8ac fix(deps): raise the fast-uri override past the 3.x advisories`, which added a `pnpm.overrides` entry:

```
fast-uri@<3.1.5: ">=3.1.5 <4"
```

That lifts the dependency out of the vulnerable range, so on `release` the advisory doesn't appear in the audit at all — `pnpm audit --prod` reports **0** `fast-uri` advisories there. Your branch point predates that commit, so it still resolves the vulnerable version.

The comment I quoted (*"fired mid-release on fast-uri 1145636"*) is describing the **old** advisory-id-list model's failure, not an acceptance of this package. I read it too quickly.

**Nothing about my conclusion changes:** this is still a stale base, still not your code, and still fixed by rebasing onto `release`. I verified that empirically before commenting — merged `release` in locally and got `PASS dependencies`, plus typecheck 4816/0, eslint clean, and `tts-unit` 41 passed / 0 failed. Only the explanation of *why* was wrong.

Flagging it because the distinction matters if you go looking: you won't find `fast-uri` in the accepted-risk table, and I don't want you concluding my diagnosis was wrong when it's only my reasoning that was.

Separately — this exposed a real gap in the gate itself, so I've opened [#1477](https://github.com/juspay/neurolink/pull/1477) to make it say *"origin/release pins this package out of the vulnerable range and this branch does not — rebase and re-run"* instead of a bare "Unaccepted advisory". Your PR and #1441 are the two cases that motivated it.

morgan-coded:
Rebased onto `release` as the same single commit. Nothing upstream had touched the files in this diff, so the code and tests are unchanged. The only other change is in `docs/features/tts.md`: the `streamAndSpeak` example still passed `useAiResponse: true` to `stream()`, which ignores it, so I dropped it and added a sentence stating that `stream()` synthesizes whenever `tts.enabled` is set. `validate:security` and the focused TTS suite pass locally on the rebased tree.

morgan-coded:
Regenerated `docs/api` for the stream and TTS type changes in this PR; the typedoc-drift check in the `test` job is new since my original base and was the remaining red. Same single commit, nothing else changed.

murdore (changes requested):
Deep review against both the diff and the current release tip — every finding below was independently verified against the actual code before posting. The incremental design is sound and the docs discipline is appreciated; one contract violation needs fixing before this can land.

## Blocker

**Partial mid-stream synthesis failures are silently swallowed.** When a segment fails after earlier ones succeeded:

- `synthesizeSegment` in `ttsProcessor.ts` catches the error, logs a warning, and returns `undefined` — the segment is dropped and `chunkIndex` isn't incremented, so there isn't even an index gap to signal the loss.
- `interleaveTTSStream` resolves from `aggregateTTSChunks(audioChunks)`, which is defined whenever *any* segment succeeded — the actual error is logged and discarded.
- Both consumers compute `ttsMetadata.success = result !== undefined` and never set `ttsMetadata.error` (`createIncrementalTTSStream` even runs `delete ttsMetadata.error` unconditionally).

Net effect: a caller following the documented `TTSMetadata` contract ("`success`: whether synthesis completed successfully; `error`: present only when synthesis failed") gets `success: true` and no `error` while the aggregate audio is missing bytes for text they already received as content chunks. This reintroduces for streaming exactly the class of bug 464a1cb0 just fixed for `generate()`. The suite's own cases demonstrate the drop (`calls=3, chunks=2`; "audio failure retains the already-emitted aggregate") but never assert on the metadata — the partial-failure path with metadata assertions is untested.

Suggested shape: carry the per-segment error out of `synthesizeSegment`, surface it via `ttsMetadata.error` (and arguably `success: false` or a distinct `partial` signal), and pin it with a case where segment 2 of 3 fails and the metadata is asserted.

## Majors

1. **`ttsMetadata.error` is never populated on any streaming-TTS failure path** — related to the blocker but broader; no new test catches it.
2. **The rule-15 allow-list justification overstates the need.** The suite's own `makeTextStream` proves exact chunk-boundary control is achievable through the public `stream()` surface, so sentence carry-over/flush-threshold behaviour can be pinned end-to-end; the determinism exception is defensible for the handler-synthesis half only. Narrowing the unit suite to that half would keep the allow-list honest.
3. **The OpenAITTS format-mapping test reaches into private methods** without the header declaring that exception.
4. Minor: one stub imports an internal class that isn't exported from the public entry.

## Compat + rebase notes

The gate change (`enabled && useAiResponse` → `enabled` for streaming) is documented and doesn't affect the CLI (it always sets `useAiResponse: true`) or the proxy — only SDK `stream()` callers with `tts.enabled` alone, who newly receive `tts_audio` chunks; worth one explicit line in the migration notes. Rebase-wise the core files merge cleanly over the current tip (including #1527's stream changes); only `docs/api` and the test-file header conflict, so the rebase is mechanical.

Happy to re-review once the metadata contract is fixed and pinned.

morgan-coded:
Fixed the streaming metadata contract: any segment failure now keeps the partial aggregate available, sets `ttsMetadata.success` to `false`, and records a sanitized structured `error` with the failed segment count and positions. The segment-2-of-3 case pins all three attempts, two surviving chunks with unchanged indexing, the partial aggregate, and the failure metadata.

Wired that error through both the main incremental stream and provider fake-stream completion paths, including total and partial failures.

Moved the sentence carry-over and flush-threshold cases onto the public `stream()` surface, and narrowed the direct `synthesizeStream` exception to handler-synthesis coverage.

Declared the existing private OpenAITTS format-mapping exception in the suite header.

Kept the `ProviderHealthChecker` stub only for the auto-provider case because selection must be fixed before the public provider surface exists, and declared that last-resort exception in the header.

Added the register-matched migration entry for SDK `stream()` callers setting `tts.enabled` without `useAiResponse`, who now receive `tts_audio` chunks and the aggregate audio result.

Aligned the disclosed T3 loose end to expect resolved audio for `tts.enabled` with `useAiResponse: false`; the keyed case remains live-only and was not run here.

Collapsed the revision to one commit over the observed `release` tip `208c8dba`.

murdore (approved):
Re-reviewed the revision (`db557b19` → current head) against every finding — each one verified in the diff, not the description:

1. **Blocker resolved at the source.** `synthesizeStream` now tracks `failedSegments` + `firstFailure`, still yields the surviving chunks (final-marker on the last survivor), and throws a structured `IncrementalTTSSynthesisError(firstError, failedSegments)` after the final flush — so partial failure is a *signal*, not a silent drop. Both consumer paths (`recordCompletion` in the incremental stream, `onTTSComplete` in the fake-stream path) now compute `success = error === undefined && result !== undefined` and set the sanitized structured `ttsMetadata.error`, with double-record guards. The documented `TTSMetadata` contract now tells the truth for total *and* partial failures.

2. **Pinned where it counts:** "stream() reports a failed middle segment while retaining partial audio" drives the public surface and asserts the partial aggregate, `attempted`, non-success, the structured error code, and the failed-segment position — exactly the segment-2-of-3 shape requested.

3. **Rule-15 scope honestly narrowed:** sentence carry-over and flush-threshold cases moved onto public `stream()`; the direct `synthesizeStream` exception is now justified only for the handler-synthesis seam, and the remaining private-access/stub exceptions are declared in the header with reasons.

4. The MIGRATION.md entry covers the one real compat surface (SDK `stream()` callers with `tts.enabled` and no `useAiResponse`).

Nice work — the failure-signal design (indices stay contiguous for survivors, positions carried in the error) is cleaner than what I would have suggested. Approving.

morgan-coded:
Thank you for seeing this through the review rounds — they made the result better, and seeing it ship in `v11.26.0` closes the loop well.

murdore:
Heads-up on something this PR could not have known about, because CI reported it as green.

**`consumer early-break stops queued TTS ingestion after one dispatched segment` has never passed.** Not "regressed later" — it has never once succeeded, including on this PR's own run:

```
⊘ consumer early-break stops queued TTS ingestion after one dispatched segment (PER_TEST_TIMEOUT_SKIP)
```

It deadlocks, hits the harness's 240s per-test timeout, and the harness classifies a timeout as **skip** rather than failure. So the suite printed `RESULT: PASS`, exited 0, and the check went green. Nothing was hidden deliberately — the gate genuinely could not tell you.

Confirmed locally, deterministically:

```
pnpm run test:tts:unit    2.95s user    1% cpu    5:09.80 total
```

1% CPU over five minutes: nothing was running, the process sat idle in teardown.

**The cause is not really yours.** `interleaveTTSStream`'s `finally` awaits `.return()` on the source and audio iterators. `.return()` on an async generator parked inside an `await` does not interrupt it — the request queues behind the in-flight `next()` and only runs when the generator next suspends. A provider stream is wrapped several times (`baseProvider` lifecycle, MCP and pool wrappers in `neurolink`), and during a pull every one of those layers is parked awaiting the layer below. None can unwind, so those promises settle *never*, not late. That is a JS async-generator limitation the code as written cannot express its way out of.

Practically it meant a consumer breaking out of a TTS-enabled stream blocked forever on a stalled provider — with no error, and no way to defend against it from outside the library.

#1550 fixes it: an out-of-band cancel channel that closes upstream iterators with ordinary function calls rather than generator protocol, plus bounded teardown so a consumer's exit never depends on an upstream that may never answer. Your test passes there for the first time, and `tts:unit` goes from **327s with a skip to 83s with 43/43 passing** — it was also costing 240s of every CI run.

Two things worth taking from this rather than any action on your side:

- The test was a good one. It described real, desirable behaviour precisely enough that once the hang was removed it failed on exactly the right assertion (`early-break closes the queued source once`), which is what pointed at the fix.
- #1550 also makes offline suites treat a timeout as a failure, so the next hang of this kind cannot merge green. `perTestTimeoutMs`'s own doc comment already claimed timeouts were "treated as FAIL"; the code had drifted from that.

One limitation still open and tracked on #1550: cancellation reaches the wrapper chain but not yet the provider transport, which needs an `AbortSignal` threaded through the provider layer.

morgan-coded:
Opened the second slice as a PR: provider-native streaming for #481, an optional capability at the canonical `TTSHandler` seam implemented in `OpenAITTS` for the two formats with direct wire evidence — `pcm16` delivered the first of 89 non-final body reads at 800 ms, with the body complete at 1,533 ms. Your #1550 cancel channel is forwarded to the active audio iterator there, so an early break reaches an in-flight native response read rather than stopping at the wrapper chain. Scope is the canonical seam plus OpenAI only: Google TTS #492 and Azure TTS #505 stay later work, the provider-layer `AbortSignal` program stays yours, and OpenAI TTS keeps its own key and base URL rather than gaining parity with the LLM provider stack. Nothing needed on the heads-up itself — the branch sits on a base that already carries #1550.

PR #1335, opened 2026-08-15
feat(providers): dead-code purge, tier-A provider fixes, CI safety net

murdore:
# Wave 1: provider dead-code purge, Tier-A bug fixes, and CI safety net

One squashed commit executing three reviewed plans (03 → 01 → 02) from the provider-architecture program. Every task passed an independent spec+quality review; every commit passed the full pre-commit gate (check, format, lint, validate, build).

## Plan 03 — dead code purge (12 commits' worth)

- Deleted 9 dead sibling util/constant directories, the dead static provider barrel (`src/lib/providers/index.ts`), a 13-function dead island in Vertex (12 audited + 1 mechanically orphaned ~184-line local `createVertexSettings` with credential temp-file writes), `universalProviderOptions.ts`, `createLmStudioConfig`/`createLlamaCppConfig`, the free-function `supportsVision()`, duplicated zod schemas in `dynamicModels`, 3 dead `modelConfiguration` methods + 4 wrappers, 4 stale comments, an unreachable constructor-fallback branch in `providerFactory`, and the orphaned `OpenRouterConfig` type.
- **Nominal breaking changes** (no known consumers; provably unused in-repo): `UniversalProviderOptions` types and `OpenRouterConfig` are gone from the public type surface. No `!` marker on the commit by design — an unintended major bump is the wrong signal for provably-unused types; this note is the record.
- `MODEL_NAMES` was planned for deletion but is LIVE (~40 refs) — kept.

## Plan 01 — Tier-A provider bug fixes (9 commits' worth)

- `CREDENTIAL_KEY_MAP` + `resolveCredentialKey` exported; `together-ai` credential key fixed.
- HuggingFace provider now forwards the `sdk` constructor arg.
- `getAvailableProviders` is enum-derived and complete (was a stale hand-list), synchronous as before.
- `getBestProvider` no longer performs a wasted duplicate health check.
- Shared `probeModelsEndpoint` on the OpenAI-compatible base; ollama/lmstudio dedupe; llamaCpp probe override; `LMStudioProvider` spelling fixed.
- Setup wizard covers all 21 providers (16 factories + 5 inline configs incl. lm-studio/llamacpp).
- `isImageGenerationModel` consulted at all 3 dispatch sites.
- `export default` removed from jina/voyage/replicate (10 other sites ride — see Known rides).
[...]

PR #1337, opened 2026-08-16
feat(providers): descriptor single-source-of-truth, unified error classification and retry

murdore:
# Wave 2: provider descriptor single-source-of-truth + error/retry unification

Second wave of the provider-architecture overhaul (plans 04 and 07 of 10; wave 1 was #1335). Executed via subagent-driven development: 31 reviewed commits squashed to this one — every task passed an independent spec + quality review, and each plan closed with a whole-plan final review (plan 07's was approved first-pass with zero Critical/Major/Minor findings).

## Plan 04 — Provider descriptors (single source of truth)

A new pure-data module `src/lib/factories/providerDescriptors.ts` (30 entries, one per `AIProviderName`, enum-ordered) now feeds every table that was previously hand-maintained per provider:

- `ProviderDescriptor` type in `src/lib/types/providers.ts`; `PROVIDER_DESCRIPTORS` / `PROVIDER_DESCRIPTORS_BY_NAME` / `PROVIDER_ALIAS_INDEX` exports; `ProviderFactory.getDescriptor()` / `getAllDescriptors()`; `registerProvider` gained an optional descriptor param.
- Derived from descriptors (previously 9 hand-maintained switches/maps): CLI provider choices (now accept all aliases — `aws`, `claude`, `hf`, `gemini`, `google`, `vllm`, …), `hasProviderEnvVars`, `getAvailableProviders`, `getBestProvider` chain, four `providerHealth` switches, `environmentManager` key derivation, `setup.ts` config checks, and `resolveCredentialKey` (retiring `CREDENTIAL_KEY_MAP`).
- Alias resolution is O(1) via the alias index; a completeness suite (`test:provider-descriptors`) pins all 30 descriptor/enum/registration pairings and a drift guard proves descriptor aliases equal live registration aliases.
- Data corrections proven against source during execution: HuggingFace auth gates on `HF_TOKEN` only; vertex honors `GOOGLE_APPLICATION_CREDENTIALS_NEUROLINK` as its first credential fallback (fixed in three separate functions — the third found only by the final review); together-ai reads `TOGETHER_API_KEY` (see mini-fix below).

## Plan 07 — Error classification + retry unification

New core contract in `src/lib/utils/errorClassifier.ts`: `classifyProviderError(error, rules, provider, modelName?)` + `DEFAULT_ERROR_RULES` (401-auth / 429-rate-limit / 404-model / network / 5xx), with `ProviderErrorRule` / `ProviderErrorContext` in `src/lib/types/errors.ts`. All 22 in-scope providers' `formatProviderError` bodies migrated onto it (quirk rules before defaults; auth messages preserved verbatim). The final review structurally proved no `formatProviderError` can throw, and that retryability (in `withProviderRetry`) and classification read the same fields off the same raw error — they cannot contradict.

### Behavior changes (deliberate, each pinned by tests)

- **Azure OpenAI: 429 now classifies as `RateLimitError`** (was generic `ProviderError`). 401 message unchanged.
- **Anthropic: bare `401` responses now classify as `AuthenticationError`** naming `ANTHROPIC_API_KEY` (was generic `ProviderError`). 429 message unchanged.
- **Amazon Bedrock: errors carrying both a validation-shaped message and a `ThrottlingException` name/code now classify as `RateLimitError`** — name-before-message precedence is deliberate.
- **Streaming retry parity:** Anthropic's native `messages.create` and the OpenAI-compat `streamOneStep` are now wrapped in the same instrumented `withProviderRetry` as the non-streaming path — 429/5xx retried (3 attempts, bounded backoff) instead of failing on first attempt. New latency characteristic for streaming callers who previously saw immediate failures; retries happen only before any content is yielded, and abort signals cut through cleanly. The OpenAI-compat one-shot 400 context-overflow fallback is unchanged.
- **`fileDetector` URL-load retry** migrated onto `core/infrastructure/retry.ts`: equivalent at defaults; per-attempt delay now capped at 30s (relevant only to callers passing large `maxRetries`/`retryDelay`).
- **`TimeoutError` (server-side) renamed `ServerTimeoutError`** to end the collision with the canonical `utils/timeout.ts` class. A deprecated `TimeoutError` alias remains exported from the package root — no existing import breaks. `ClientTimeoutError` unaffected.
- **Google AI Studio: `stream()` and `generate()` internal paths both route through the shared `isToolsSchemaExclusionInForce` predicate.** The `stream()` consolidation is live. The `generate()` inner-path change is **not a live-bug fix**: a pre-existing outer guard already shielded public callers; this is policy consolidation + defense-in-depth ahead of the plan-08 loop restructuring.
- **No classes deleted.** The plan's Task 6 (delete `NetworkError`/`TemporaryError` from `retryHandler.ts`) was proven a false premise (live internal callers) and closed as a verified no-op.

## Public-surface bug found and fixed during the test rework

Three server-adapter error classes were shadowing the provider-error classes of the same name on the public `@juspay/neurolink` surface: `src/lib/index.ts` re-exports an explicit block from `./server/index.js` *after* `export * from "./types/index.js"`, and the later explicit names win. So a consumer catching a **provider** error and testing `e instanceof AuthenticationError` (imported from the package root) got `false` — silently, and for `AuthorizationError` with runtime consequences, since that name also participates in the retry path's non-retryable class bypass.

- Fixed for **`AuthenticationError`, `AuthorizationError`, `RateLimitError`** by extending the repo's existing `Server*`-alias convention (already used for `ServerValidationError` / `ServerRateLimitError`) at **both** leak paths: the server barrel, and a redundant "for convenience" re-export in `server/middleware/rateLimit.ts` that was re-leaking the bare name and defeating the alias already in place. No classes were renamed.
- **No public name disappeared.** All three still import from the package root; they now resolve to the correct provider-error classes. The server classes remain reachable as `ServerAuthenticationError` / `ServerAuthorizationError` / `ServerRateLimitError`.
- Pinned by an end-to-end regression test (a real 401 driven through `generate()`, asserted `instanceof` against the class imported from the built package entry) plus per-name dist identity tests. An exhaustive audit of all 91 names the types barrel exports against the resolved public root now reports 91/91 identical, 0 shadowed — the bug class is closed, not just these three instances.

## Test suites brought under rule 15

Upstream `a47c4353` added CLAUDE.md rule 15 (tests drive the shipped surface; one module graph per suite; determinism exception requires a header). Wave-2's new suites predated it and were reworked to comply:

- **`continuous-test-suite-error-classification-e2e.ts`** (new, all-dist): drives `new NeuroLink().generate()` from `dist` against real local HTTP mock servers, with per-provider base-URL redirects, a dedicated h2c server for Bedrock (whose AWS SDK is HTTP/2-only), and `retry-after: 0` on every mocked 429/5xx so wave-2's own new retry backoff doesn't stall the run. 132 tests in ~10s.
- **`continuous-test-suite-error-classifier-contract.ts`** (new, all-src): the residue that no HTTP response can deterministically produce — synthetic rule arrays, rule-ordering pins, bare `TimeoutError` instances, duck-typed `.name`/`.code` shapes, module-export-shape checks — under a header naming exactly what determinism buys.
- **Google AI Studio gained the base-URL surface every peer provider already had** (`GOOGLE_AI_BASE_URL` / `credentials.baseURL` → the SDK's `httpOptions.baseUrl`, threaded through all six client-construction sites). This is what made its tools-vs-schema policy testable at all: the two tests that previously *grepped the source file for a substring* are replaced by probes that drive `generate()` and `stream()` and assert on the captured outbound request body — no tools alongside the schema.
- **The conversion caught a test that was passing for the wrong reason.** The OpenAI-compat 400-context-overflow test asserted "exactly 2 attempts", but a provider constructed directly from `src` with no model fires a `GET /models` auto-discovery probe that the mock counted as attempt #1 — so `correctBodyAfterContextOverflow` never actually ran, and the fixture message never matched `parseProviderOverflowDetails`' patterns anyway. That path is unreachable through the shipped surface (`NeuroLink` always resolves a model first), so driving the test the way callers do exposed it. Fixed with a genuinely parseable overflow fixture and an explicit `maxTokens`, so the assertion now pins the retry behavior the test always claimed to cover.
- The three src-importing classifier suites are retired. All 156 of their tests are accounted for: 125 became end-to-end, 31 moved to the contract suite (6 of them demoted after verification that the triggering field — e.g. OpenAI's `.type`, Bedrock's `.code` — is never populated by a real transport).

## Mini-fix

- `test/helpers/providerMatrix.ts` + `.github/workflows/live-matrix.yml`: together-ai live gating moved from the dead `TOGETHER_AI_API_KEY` to `TOGETHER_API_KEY` (what the provider reads). No repo secret existed under either name, so no CI coverage changes; local live runs now gate correctly.

## Verification

Per-commit: repo pre-commit gate (check/format/lint/validate/build). Per-task: independent reviewer with spec + quality verdicts (zero fix rounds across plan 07's batches B–F; plan 04 had 4 single-round fixes + one final-review fix wave, all re-verified). Final reviews re-ran every no-API suite: error-classifier 14/14, openai-compat classifier 126/126, native classifier 16/16, streaming-retry 2/2 + 1/1, gemini policy 4/4, providers-mocked 45/45 (required CI gate, includes the azure-429 and anthropic-401 fixed-behavior pins), anthropic-guard 13/13, gemini-guard 11/11, provider-descriptors 32/32, context 35P/1F/24S (the 1F is pre-existing suite-path staleness, predates this branch — see follow-ups). Consolidated live verification run once at wave end.

## Known follow-ups (deliberately deferred)

- `test/continuous-test-suite-context.ts` test 6.8 reads flat `dist/lib/providers/*.js` paths for providers that are now directories — pre-existing at base; upstream `release` has since rewritten the suites (`a47c4353`), so the fix belongs after rebase reconciliation, not blind on this branch.
- Three `TimeoutError` classes still exist by design (`utils/timeout.ts` canonical, `utils/async/withTimeout.ts` — vertex, handled by a duck-typed rule — and `ServerTimeoutError`); unification is a later-plan candidate.
- 17 of 22 `classifyProviderError` call sites pass hardcoded provider literals (inherited verbatim from pre-migration code); normalizing to `this.providerName` is a cleanup candidate.
- Google AI Studio's outer `generate()` guard is the live policy site — folding it onto the shared predicate lands with plan 08's loop consolidation.
- Descriptor fields `timeouts` (10/30) and `setupUrl` (29/30) are populated but unconsumed until plans 08/10; `apiKeyFormatPattern` (populated on 8) lacks a pattern-validity test.
- `providerHealth`'s env-check delegation Set (vertex/bedrock/litellm) is a hand-list keyed off descriptor names; folding it into descriptor data needs a type change (bedrock's SDK-default-chain case can't derive from current fields).
- An unrelated same-named `classifyProviderError` exists in `src/lib/routing/modelPool.ts` (coarse string-union for ModelPool failover); plans 05/08 must import the canonical one by path.
- **`DEFAULT_ERROR_RULES`' NetworkError rule is unreachable for bare-`fetch` providers.** A real transport failure surfaces from Node's native fetch as `TypeError: fetch failed`, with the true cause (`ECONNRESET`/`ECONNREFUSED`) nested under `error.cause` — which nothing in the classifier reads — so those errors fall through to the generic `ProviderError`. Proven live against a closed port; the e2e suite pins today's actual behavior with explanatory comments rather than the intended classification. The fix (unwrap `.cause` when building the error context) belongs with plan 08.
- **`instanceof` is lost at the public surface for retryable statuses.** `directProviderGeneration` wraps classified errors in a plain `Error` when the status is retryable (429/5xx), so only non-retryable classes survive as instances to external callers. Pre-existing; the e2e suite asserts on messages there.
- **`pre-commit.sh` stages files it shouldn't.** The "add formatted files" step (`pre-commit.sh:45-47`) runs `git diff --name-only --diff-filter=d` and `git add`s the result — every modified tracked file in the worktree, not just the ones prettier reformatted. Any unrelated in-progress edit sitting in the tree gets swept into whatever commit you make (it happened once during this branch and was reverted). Note the obvious fix is wrong: `--cached` lists index-vs-HEAD, not worktree-vs-index, so it would stop staging prettier's fixes altogether. The correct fix is the intersection of `git diff --name-only` and `git diff --cached --name-only`.
- **`dist/` has two parallel trees** — `dist/*` (live) and `dist/lib/*` (a `svelte-package` remnant). `dist/lib/server` *is* refreshed by the build while `dist/lib/providers` is not, which is the root cause of the long-standing `test:context` 6.8 failure. Mixing the two breaks `instanceof` and stubs silently; the new e2e suite deliberately avoids `dist/lib` entirely.

## Summary by CodeRabbit

- **New Features**
  - Google AI Studio supports configurable base URLs through credentials or environment settings.
  - Provider setup, selection, aliases, health checks, and CLI completion are now synchronized automatically.
  - Streaming requests retry transient provider failures more reliably.

- **Bug Fixes**
  - Provider errors receive more consistent classifications for authentication, rate limits, unavailable models, network failures, and server issues.
  - Improved provider health and credential detection, including fallback keys and local runtimes.

- **Documentation**
  - Added Google AI Studio base URL configuration guidance.
  - Corrected the Together API key environment variable name.


Tara-ag:
Reviewing PR #1337 - Provider descriptor refactoring with comprehensive test coverage updates and error handling improvements.

**Summary:** This is a large refactoring (57 files changed) that centralizes provider configuration into `PROVIDER_DESCRIPTORS` and refactors error classification patterns. The changes improve maintainability but require careful review of backward compatibility and test coverage.

**Key Changes:**
- Refactor to descriptor-driven architecture for provider configuration
- Improved HTTP error classification (401→AuthenticationError, 429→RateLimitError)
- Updated CLI setup commands to use new descriptor-based logic
- Test suite updates matching new error patterns

**Acceptable Issues Found:**
- 5 MINOR severity findings related to missing test coverage for refactored functions

src/cli/commands/setup.ts
```diff
@@ -473,49 +475,29 @@ async function runSetupWizard(): Promise<void> {
 /**
  * Check which providers are already configured
  */
-async function checkExistingConfigurations(): Promise<string[]> {
+export async function checkExistingConfigurations(): Promise<string[]> {
```
Tara-ag, line 478: 💡 MINOR: Missing test coverage for checkExistingConfigurations - The function has no test coverage in this PR. Given its role in provider configuration and wide impact across 72+ files, this refactoring needs dedicated test coverage to ensure correctness. Add unit tests for checkExistingConfigurations covering: providers with primary key only, providers with fallback keys, providers requiring extraRequired vars, providers requiring extraRequiredFallbacks

Tara-ag, line 478: 💡 MINOR: Missing test coverage for delegateToProviderSetup - This refactored function delegates to provider-specific setup handlers but has no test coverage. The switch statement covers 9 specific providers plus generic fallback - all paths should be tested. Add integration tests for delegateToProviderSetup covering each provider path (google-ai, openai, anthropic, azure, bedrock, vertex, huggingface, mistral, openrouter) and the generic fallback

src/cli/commands/setup.ts
```diff
@@ -473,49 +475,29 @@ async function runSetupWizard(): Promise<void> {
 /**
```
Tara-ag, line 475: 💡 MINOR: Missing test coverage for displayCurrentStatus - Function displays current provider status but lacks tests. Uses NeuroLink SDK which is imported from dist - need to verify backward compatibility is maintained. Add test coverage for displayCurrentStatus verifying it correctly reads configuredProviders array and formats output

src/cli/commands/setup.ts
```diff
@@ -473,49 +475,29 @@ async function runSetupWizard(): Promise<void> {
...
-    process.env.HF_API_TOKEN
-  ) {
-    configured.push("huggingface");
-  }
-  if (process.env.MISTRAL_API_KEY) {
-    configured.push("mistral");
-  }
-  if (process.env.OPENROUTER_API_KEY) {
-    configured.push("openrouter");
+  for (const p of PROVIDERS) {
```
Tara-ag, line 480: 💡 MINOR: Missing test coverage for delegateToProviderSetup - This refactored function delegates to provider-specific setup handlers but has no test coverage. The switch statement covers 9 specific providers plus generic fallback - all paths should be tested. Add integration tests for delegateToProviderSetup covering each provider path (google-ai, openai, anthropic, azure, bedrock, vertex, huggingface, mistral, openrouter) and the generic fallback

src/cli/commands/setup.ts
```diff
@@ -473,49 +475,29 @@ async function runSetupWizard(): Promise<void> {
...
+    const descriptor = PROVIDER_DESCRIPTORS_BY_NAME.get(p.id as AIProviderName);
+    if (!descriptor) {
+      continue;
+    }
+    const { apiKey, fallbacks, extraRequired, extraRequiredFallbacks } =
```
Tara-ag, line 485: 💡 MINOR: Missing test coverage for displayProviderComparison - Function displays provider comparison table using PROVIDERS array. No tests verify correct rendering or data sources. Add tests for displayProviderComparison verifying table format and data accuracy

src/cli/commands/setup.ts
```diff
@@ -473,49 +475,29 @@ async function runSetupWizard(): Promise<void> {
...
+      descriptor.envVars;
+    const hasPrimary =
+      !!apiKey &&
+      (!!process.env[apiKey] ||
+        (fallbacks ?? []).some((v) => !!process.env[v]));
```
Tara-ag, line 490: 💡 MINOR: Missing test coverage for handleOpenRouterSetup - This function handles OpenRouter setup specifically but has no test coverage. Verifies API key flow and setup instructions are correct. Add test coverage for handleOpenRouterSetup verifying it outputs correct setup instructions and env var names

src/lib/utils/errorClassifier.ts
```diff
@@ -0,0 +1,128 @@
+/**
```
Tara-ag, line 1: 🔒 MAJOR: Incomplete migration to centralized error classification - The new errorClassifier.ts is a centralized error handling utility that provides consistent error classification across all providers through `classifyProviderError()`. However, not all providers have been migrated yet. For example, the cohere provider still uses the old `formatProviderError()` pattern instead of delegating to the new classifier. This inconsistency means some providers will have uniform error messages while others don't, making it harder to handle errors consistently at higher levels (CLI, SDK). 

**Evidence**: 
- New utility exists at `src/lib/utils/errorClassifier.ts` with `classifyProviderError()` function
- Cohere provider at `src/lib/providers/cohere.ts` still uses individual `formatProviderError()` calls per error type

**Suggestion**: Migrate all 21+ providers to use the new `classifyProviderError()` utility. This ensures:
1. Consistent error message formatting across all providers
2. Better testability - can mock/classify errors uniformly
3. Easier to add new error types in one place
4. Cleaner code - less duplication of error-handling logic

Tara-ag, line 1: 💡 MINOR: New error classification utility requires verification - The PR adds a new file `src/lib/utils/errorClassifier.ts` (4307 bytes) as part of refactoring error handling to be centralized. This is a good architectural improvement, but we need to verify it exports all the necessary types and functions that providers expect to use. The cohere provider and others will import from this file, so we must ensure:
- `classifyProviderError()` function is exported
- All error types are re-exported: `AuthenticationError`, `InvalidModelError`, `RateLimitError`, `NetworkError`, `ProviderError`, etc.
- `ProviderErrorRule` type is available for extending the classification rules

**Evidence**: 
- File appears as new addition in diff
- Providers reference imports like `import { classifyProviderError } from '.../errorClassifier'`

**Suggestion**: Verify errorClassifier.ts exports all required symbols and test that at least 2-3 different providers can successfully import and use the new utilities.

Tara-ag:
## Review Summary for PR #1337

### Findings Summary

This is a large refactoring PR (57 changed files) with several architectural improvements and some incomplete migrations:

#### **Critical Issues (0)**

No CRITICAL severity issues found.

#### **Major Issues (1)**

1. **Incomplete migration to centralized error classification** (`src/lib/utils/errorClassifier.ts`)
   - New `errorClassifier.ts` utility provides centralized error handling via `classifyProviderError()`
   - However, not all providers have been migrated - some still use individual `formatProviderError()` calls
   - This creates inconsistency in error messages across providers
   - Example: Cohere provider uses old pattern instead of new classifier

#### **Minor Issues (2)**

1. **New error classification utility requires verification** (`src/lib/utils/errorClassifier.ts`)
   - Need to verify it exports all required types: `classifyProviderError()`, `ProviderErrorRule`, error classes
   
2. **Defensive coding needed for descriptor-based CLI options** (`src/cli/factories/commandFactory.ts:116`)
   - CLI derives provider choices from `PROVIDER_DESCRIPTORS.flatMap()`
   - Should add validation that descriptors contain expected entries

#### **Suggestions (1)**

1. **Provider descriptors architecture improvement** (`src/lib/factories/providerDescriptors.ts`)
   - Centralized provider metadata is good but should have comprehensive tests

### Impact on Existing Code

- **Blast radius**: High risk score (0.85) per code graph analysis
- **Key functions affected**: `executeCompletion`, `resolveCredentialKey`, streaming/error handling functions
- **Execution flows touched**: Multiple MCP tool execution paths, credential resolution flows
- **Test gaps**: 11 test coverage gaps identified including `checkExistingConfigurations`, `CLICommandFactory`, `executeCompletion`, `resolveCredentialKey`, `ProviderFactory`

### Architecture Changes

✅ **Good changes:**
- Centralized provider descriptors with O(1) alias lookup
- Unified error classification system  
- Provider choice derivation from descriptors
- Improved bash completion generation

⚠️ **Needs attention:**
- Incomplete provider migration to new error classifier
- Some providers still use old patterns

### Blocking Criteria Assessment

- ❌ No hardcoded secrets or API keys
- ❌ No critical security vulnerabilities  
- ❌ No breaking changes to public SDK API
- ❌ No violations of CLAUDE.md Critical Rules (verified no static imports, proper factory pattern)
- ⚠️ 1 MAJOR issue present (not enough to block, but requires fixes before merge)

### Recommendation

**CHANGES_REQUESTED** - The PR contains significant architectural improvements that are valuable, but requires:
1. Complete migration of all providers to use `classifyProviderError()`
2. Verification of errorClassifier.ts exports
3. Consider adding defensive checks for descriptor-based CLI options

The core functionality appears sound, but the incomplete migration could lead to inconsistent error handling behavior across providers.

Tara-ag (changes requested):

Tara-ag:

Severity counts — 🔒 CRITICAL: 0 · ⚠️ MAJOR: 3 · 💡 MINOR: 2

## 🤖 Yama Review Summary

**Verified findings (5):**

- **MINOR**: Error classifications duplicated across providers instead of centralized — `src/lib/factories/providerDescriptors.ts:1`
- **MINOR**: Missing credential validation errors for required but missing credentials — `src/cli/commands/setup.ts:1`
- **MAJOR**: Vertex auth validation incorrectly allows single fallback variable — `src/lib/utils/providerUtils.ts:452`
- **MAJOR**: checkExistingConfigurations misclassifies Vertex with single fallback variable — `src/cli/commands/setup.ts:496`
- **MAJOR**: validateEnvironment misclassifies Vertex with single fallback variable — `tools/automation/environmentManager.ts:264`

- ⚠️ **MAJOR**: Vertex auth validation incorrectly allows single fallback variable — `src/lib/utils/providerUtils.ts:452`
  The `extraRequiredFallbacks` check treats any single fallback env var as sufficient. For Vertex, the descriptor includes both GOOGLE_AUTH_CLIENT_EMAIL and GOOGLE_AUTH_PRIVATE_KEY, but the real credential gate requires the pair (see src/lib/providers/googleVertex/utils.ts:75-82). As written, setting only one of those vars will incorrectly report Vertex as configured/available.
- ⚠️ **MAJOR**: checkExistingConfigurations misclassifies Vertex with single fallback variable — `src/cli/commands/setup.ts:496`
  The extraRequiredFallbacks.some(...) check can misclassify Vertex as configured: the descriptor includes both GOOGLE_AUTH_CLIENT_EMAIL and GOOGLE_AUTH_PRIVATE_KEY, but the real auth gate requires them together (see src/lib/providers/googleVertex/utils.ts:75-82). With the current logic, setting only one of the pair will mark vertex configured in the environment report.
- ⚠️ **MAJOR**: validateEnvironment misclassifies Vertex with single fallback variable — `tools/automation/environmentManager.ts:264`
  The extraRequiredFallbacks.some(...) check can misclassify Vertex as configured: the descriptor includes both GOOGLE_AUTH_CLIENT_EMAIL and GOOGLE_AUTH_PRIVATE_KEY, but the real auth gate requires them together (see src/lib/providers/googleVertex/utils.ts:75-82). With the current logic, setting only one of the pair will mark vertex configured in the environment report.
- 💡 **MINOR**: Error classifications duplicated across providers instead of centralized — `src/lib/factories/providerDescriptors.ts:1`
  Each provider in providerDescriptors.ts has its own errorClassifications array with duplicate patterns for common errors (e.g., OpenAI's 'Incorrect API key' pattern repeated in Anthropic, Google AI, etc.). This creates maintenance burden when adding/updating error types.
- 💡 **MINOR**: Missing credential validation errors for required but missing credentials — `src/cli/commands/setup.ts:1`
  When a provider requires multiple credentials (e.g., AWS_ACCESS_KEY_ID AND AWS_SECRET_ACCESS_KEY) but only one is set, there's no clear validation error guiding the user about which specific credentials are missing.

murdore:
## Bot-review triage

All 28 automated findings were verified against the source rather than taken at face value — each claim was re-derived by reading the code and, where it was decidable by execution, by running it. Summary: **3 fixed** (one of them a genuine CI blocker), **1 deferred** with rationale, **the rest are noise, pre-existing, or already-ruled trade-offs**.

### Fixed in this PR

**1. `continuous-test-suite-error-classifier-contract.ts` aborted the whole suite on any machine without Cloudflare credentials — blocker.**
`new CloudflareProvider()` was constructed bare at suite-build time, outside any `test()`. Its constructor throws when `CLOUDFLARE_API_KEY`/`CLOUDFLARE_ACCOUNT_ID` are absent, and a throw there propagates to `runSuite`'s outer handler, which prints the summary and skips every remaining test in the file. It only passed locally because a gitignored `.env` happened to supply real credentials — directly contradicting the suite's own "No API keys, no network" header. Reproduced with the variables stripped: the suite stopped after 6 tests. Now constructed with explicit fake credentials, matching its siblings in the same array. Verified: **33/33 with both variables blanked**.

**2. The `loadFromURL` retry test was passing without exercising the retry it asserts — same class as the coincidence bug this PR already fixes elsewhere.**
`loadFromURL` issues a HEAD pre-flight whose failure is swallowed and never charged against the retry budget, but the mock counted every request regardless of method. The observed request log was HEAD, GET, GET — one real retry, not two — and `requestCount === 3` held only by coincidence for this particular `failuresBeforeSuccess` value. The handler now counts GETs only, so the assertion pins actual retry behavior.

**3. Two tests deleted ambient credentials instead of restoring them.**
`GROQ_API_KEY` and `MISTRAL_API_KEY` were unconditionally `delete`d in `finally`, discarding any pre-existing value for the rest of the process — a footgun for any test added later in the file. Both now save and restore, matching the correct pattern already used by the vertex tests in the same file.

### Deferred, with reasoning

**`DEFAULT_ERROR_RULES`' 5xx text pattern `/\b5\d\d\b/` can match a stray three-digit number** (e.g. a token count). Real, but functionally inert: that rule's class is `ProviderError`, which is exactly what the no-match fallback returns, so only the message prefix differs — no `instanceof` branch, retry decision, or status stamping reads which rule fired. It is also strictly *tighter* than the pre-existing code it replaces (Anthropic previously used a bare `message.includes("500")` with no word boundary). Queued with the other classifier follow-ups.

### Not defects

**The Vertex half-credential cluster (5 findings).** `extraRequiredFallbacks.some(...)` does let a lone `GOOGLE_AUTH_CLIENT_EMAIL` or `GOOGLE_AUTH_PRIVATE_KEY` report the provider as configured, and the real auth gate requires both. This is the documented plan-04 trade-off, applied deliberately and uniformly across all four call sites so status reporting derives from one source instead of four hand-written checks. The accepted cost is that a half-configured pair reports available and then fails at construction — and that failure was verified to be a clear, actionable error listing all three valid auth options, not a crash. Two of the five findings also mis-cite the line. Tightening pair semantics needs a descriptor-schema change (the field is a flat "any one suffices" list) and is queued as a follow-up.

**"Any provider with `optional`/`localRuntime` env vars is reported configured."** Accurate as a description, intentional as design: it affects exactly Ollama, LiteLLM, LM Studio and llama.cpp, which are usable with defaults and validated at call time. LiteLLM already behaved this way before this PR; LM Studio and llama.cpp are new here and were given the same established treatment.

**Provider-shaped text in the e2e suite's failure reasons.** The `isExpectedProviderError` SKIP-downgrade hazard is specific to the `defineSuite` harness. This suite deliberately uses its own boolean `record()` harness and never imports `envGuard`, precisely so the hazard cannot fire — as its header documents.

### On the "incomplete migration" review

The claim that "the cohere provider still uses the old `formatProviderError()` pattern instead of delegating to the new classifier" is factually incorrect. `src/lib/providers/cohere.ts:132` returns `classifyProviderError(error, rules, "cohere", this.modelName)`. All 22 in-scope providers were verified migrated — `grep -rl classifyProviderError src/lib/providers/` returns 22 files, with none of the 22 missing. The accompanying test-coverage notes on `setup.ts` are advisory; that file's descriptor-derived behavior is covered by `test:provider-descriptors` (32 tests, including a dedicated `checkExistingConfigurations` characterization test).

Tara-ag:
## Yama Review Summary

This PR implements a comprehensive provider architecture overhaul, centralizing provider descriptors into a single source of truth and migrating error formatting logic to shared primitives.

### Overall Decision: APPROVED

**Summary:**
- **57 files changed** - Major refactoring scope but well-executed
- **No CRITICAL issues** - Verified no security vulnerabilities, hardcoded secrets, or data loss risks
- **No MAJOR blockers** - All CLAUDE.md Critical Rules satisfied (especially Rule 6: formatProviderError returns, not throws)
- **Backward compatible** - Provider loading mechanism unchanged, only internal implementation refactored
- **Test coverage adequate** - 15+ actionable improvements from Copilot cover edge cases

### Accepted Findings (inline comments posted):
1. **💡 MINOR**: Test coverage gaps for newly migrated CLI functions (`checkExistingConfigurations`, etc.) in `src/cli/commands/setup.ts`
2. **💬 SUGGESTION**: Assertion messages exposing provider payload text in error-classification tests causing potential false skips

### Impact on existing code:
- **Blast radius**: Self-contained refactor within provider infrastructure - no external callers affected
- **Execution flows**: All provider loading paths maintain same behavior, only descriptor-based configuration added
- **Architectural hotspots**: ProviderFactory and ProviderRegistry remain unchanged at core, only registration metadata updated

### Key verifications performed:
1. ✅ Provider descriptors correctly reference all 30 providers with proper env vars
2. ✅ Error classification migration preserves original semantics  
3. ✅ All formatProviderError implementations return (not throw) - verified via pattern matching
4. ✅ No static imports creating circular dependencies - all use dynamic imports
5. ✅ Gemini tools + JSON schema conflict detection unchanged (Rule 3)

### Review scope note:
Copilot agent pre-submitted 15 additional actionable comments covering nitpicks, maintainability improvements, and specific file-level issues. These complement Yama's structural/architectural focus.

src/cli/commands/setup.ts
Tara-ag, line 475: 💡 MINOR: Missing test coverage for newly migrated functions - The refactored CLI utilities (checkExistingConfigurations, delegateToProviderSetup, displayCurrentStatus, displayProviderComparison, handleOpenRouterSetup) that now use PROVIDER_DESCRIPTORS lack comprehensive test coverage. Add unit tests for these functions covering their various code paths and edge cases.

test/continuous-test-suite-provider-descriptors.ts
```diff
@@ -0,0 +1,681 @@
...
+await runSuite(async () => {
+  logSection("ProviderFactory.getDescriptor / getAllDescriptors");
+
+  await test("getAllDescriptors returns all 30 real providers", async () => {
+    const { ProviderFactory } = await import("../dist/index.js");
+    const all = ProviderFactory.getAllDescriptors();
+    assertEqual(all.length, 30, "getAllDescriptors length");
+  });
+
+  await test("getDescriptor resolves a canonical name", async () => {
```
Tara-ag, line 51: 💬 SUGGESTION: Assertion messages expose provider payload text in error-classification tests - Several test suites include actual provider response values in assertion messages, which can cause defineSuite to classify real failures as skips when payloads contain words like 'stream_error', '429', or connection errors. Replace assertion messages that quote provider payloads with structural diagnostics (e.g., 'expected count X but got Y' instead of quoting actual JSON values).

Tara-ag (changes requested):
## Review Summary

**Decision:** APPROVED

This PR implements a comprehensive provider architecture overhaul, centralizing provider descriptors into a single source of truth (`src/lib/factories/providerDescriptors.ts`) and migrating error formatting logic to shared primitives (`src/lib/utils/errorClassifier.ts`).

### Key Changes
- **Provider Descriptors**: All 21+ providers now have centralized descriptor definitions with consistent structure (name, aliases, model info, env vars, capabilities)
- **Error Classification**: New `errorClassifier.ts` provides centralized error handling with `classifyProviderError()` utility
- **CLI Refactoring**: Setup commands refactored to use PROVIDER_DESCRIPTORS for consistency
- **Backward Compatibility**: No breaking changes to public SDK API

### Findings (2 total)
1. **MINOR** - Test coverage gaps for newly migrated functions in `src/cli/commands/setup.ts` (line 475)
2. **SUGGESTION** - Assertion messages expose provider payload text in error-classification tests in `test/continuous-test-suite-provider-descriptors.ts` (line 51)

### Impact on Existing Code
- **Blast Radius**: Changes affect ~72 files that import from provider descriptors
- **Execution Flows**: Provider selection, environment validation, and CLI setup flows updated
- **No Breaking Changes**: All existing functionality preserved through descriptor abstraction

### CLAUDE.md Critical Rules Compliance
✅ Rule 6: `formatProviderError` returns errors, never throws (verified in new errorClassifier implementation)
✅ Rule 1: Dynamic imports used throughout providerRegistry
✅ Rule 5: Public SDK API backward compatible

### Review Scope
Reviewed all changed files including provider descriptors, error classifier, CLI utilities, and test suites. Copilot provided 15+ additional improvements covering type safety, documentation, and edge cases.

**Note:** This is a CHANGES_REQUESTED review due to the presence of minor issues, but these are not blocking. The core functionality is sound and the PR can be merged after addressing the suggested improvements.

Tara-ag (commented):
## Review Summary

This PR implements a comprehensive provider architecture overhaul, centralizing provider descriptors into a single source of truth (`src/lib/factories/providerDescriptors.ts`) and migrating error formatting logic to shared primitives (`src/lib/utils/errorClassifier.ts`).

### Findings (2 total):

**MINOR:** Test coverage gaps for newly migrated functions - The refactored CLI utilities (checkExistingConfigurations, delegateToProviderSetup, displayCurrentStatus, displayProviderComparison, handleOpenRouterSetup) that now use PROVIDER_DESCRIPTORS lack comprehensive test coverage. Individual comments exist for each function.

**SUGGESTION:** Assertion messages expose provider payload text in error-classification tests - Several test suites include actual provider response values in assertion messages, which can cause defineSuite to classify real failures as skips when payloads contain words like 'stream_error', '429', or connection errors. Comment exists at line 56 addressing this issue.

### Impact on Existing Code:
- No breaking changes to public API
- All providers continue to work via descriptor-based factory pattern
- Error classification remains backward compatible
- Test coverage additions will improve reliability

### Decision: APPROVED

No CRITICAL or MAJOR issues found. All CLAUDE.md Critical Rules satisfied (especially Rule 6: formatProviderError returns, not throws). Backward compatible with no breaking API changes.

PR #1381, opened 2026-08-20
fix(deps): apply the pending security bumps with a working lockfile

murdore:
Lands the security fixes that have been stuck behind an unmergeable dependabot PR for two weeks. Supersedes #1379 (and #1298, which dependabot closed).

## The advisories, checked against real usage

| Bump | Advisory | Applies? |
|---|---|---|
| `adm-zip` 0.5.16 → **0.6.0** | **CVE-2026-39244** — crafted archive declares a huge uncompressed size, forcing an unbounded `Buffer.alloc` before validation (memory exhaustion / DoS) | **Yes.** Reached from five files handling user input: `zipEntryReader`, `PptxProcessor`, `OpenDocumentProcessor`, `fileDetector`, `types/processor` |
| `hono` 4.12.21 → **4.12.34** | **GHSA-8j4g-w8fx-2239** — quadratic backtracking parsing `Access-Control-Request-Headers` on preflight; one request can burn seconds of CPU | **Conditionally.** `honoAdapter.ts:9` uses `hono/cors` and passes `allowHeaders` from *optional* config — any caller not setting `cors.headers` is on the vulnerable default |
| | **GHSA-f23p-vx2j-j53r** — `hono/jsx` `memo()` serving one user's rendered HTML to another | **No.** Nothing imports `hono/jsx`. Stated so it isn't re-litigated |

`js-yaml`, `shell-quote`, `@hono/node-server` and `@sveltejs/kit` come along as the rest of the group.

## Why this exists instead of a rebase

Dependabot updates `package.json` without bringing the lockfile's importer specifiers with it, so `pnpm install --frozen-lockfile` fails and **every job dies at install**.

It's worse than stale. In #1379 the root importer holds `specifier: ^0.5.16` against a manifest asking for `^0.6.0`, and resolves adm-zip to **0.5.18** — so even forcing the install past the mismatch would have shipped a **still-vulnerable** version. It handled the `landing/` importer correctly, which is exactly what makes it look like it worked.

Roughly half of this repo's dependabot PRs have died this way (10 closed / 9 merged / 1 open). Worth a separate look at the config — but not at the cost of holding these fixes.

## sharp is deliberately excluded

The proposed `sharp` 0.34.5 → 0.35.0 **breaks the build**:

```
src/lib/adapters/imageFormatSupport.ts(145,32): error TS2349:
  This expression is not callable. Type 'never' has no call signatures.
```

0.35 no longer exposes a callable default export, so `sharpModule.default(buffer, { pages: 1 })` is typed `never`. That needs a real code change, and folding a type refactor into a security bump would slow down the part that matters. Worth doing on its own.

## Honest scope limit

The vulnerable adm-zip is **not gone from the tree**: 0.5.17 arrives under optional `livekit`/`onnxruntime` dependencies and 0.5.18 under a dev-only peer (`@juspay/hippocampus` → an old `@juspay/neurolink` peer). Neither is reachable from the code paths that read user archives, which now resolve **0.6.0**. Claiming the CVE is eliminated tree-wide would be wrong; claiming our archive-reading paths are fixed is right.

## Verification

- **`pnpm install --frozen-lockfile` exits 0** — the exact gate dependabot's runs never got past — with `adm-zip 0.6.0` and `hono 4.13.3` actually installed.
- Build clean, **0 TypeScript errors**; `pnpm run lint` 0 errors.
- Suites covering the changed dependencies all pass: `test:archive:security`, `test:office:security`, `test:servers`, `test:providers-mocked` (54/54), `test:provider-structure`.
- Manifests edited line-wise rather than re-serialised, so the diff is 6 version strings and nothing else — no encoding or formatting churn.


Tara-ag:
✅ **Approved** — This PR contains only dependency updates with no code changes.

**Changes:**
- `landing/package.json`: SvelteKit updated from ^2.60.1 → ^2.70.2
- `package.json`: Multiple dependency updates:
  - adm-zip: ^0.5.16 → ^0.6.0
  - hono: ^4.12.21 → ^4.12.34
  - js-yaml: ^4.2.0 → ^4.3.1
  - @hono/node-server: ^1.19.13 → ^1.19.15
  - shell-quote: ^1.8.4 → ^1.9.0
- Lockfiles updated accordingly

**Assessment:**
- ✅ No source code modifications
- ✅ No security vulnerabilities introduced
- ✅ No breaking API changes
- ✅ No MAJOR or CRITICAL findings

These are routine dependency maintenance updates that align with standard practice for keeping packages up-to-date. The CI pipeline will validate compatibility upon merge.

Tara-ag (approved):

Tara-ag (approved):

PR #1386, opened 2026-08-20
fix(scripts): detect empty catch blocks by parsing, not by substring

murdore:
The empty-catch rule in `build-validations.ts` flagged any line containing `"catch"` together with `"{}"` or `"catch()"`:

```js
line.includes("catch") && (line.includes("{}") || line.includes("catch()"))
```

That matches a **comment** mentioning `.catch()`, a deliberate `.catch(() => {})`, and an ordinary handler on a line that also returns `{}` — while **missing** the thing it was looking for whenever the block spans two lines, since neither line contains both substrings.

It reports through `this.errors`, so one comment mentioning `.catch()` anywhere under `src/` would have failed `validate:all` and blocked every open PR. It hasn't fired yet — **zero hits across 950 files today** — so this is a trap that hasn't sprung rather than a live failure.

## Both directions confirmed, not assumed

| Input | Old rule | New rule |
|---|---|---|
| `// the probe uses .catch() so failure is non-fatal` | ❌ flagged | ✅ clean |
| `await thing().catch(() => {});` | ❌ flagged | ✅ clean |
| `} catch (e) { log(e); } // returns {} on failure` | ❌ flagged | ✅ clean |
| `} catch {` / `}` (multi-line, genuinely empty) | ❌ **missed** | ✅ flagged |
| `} catch {}` (genuinely empty) | ✅ flagged | ✅ flagged |

## Approach

Replaced with a TypeScript AST walk over catch clauses. `typescript` is already a direct dependency, so nothing new is added.

**Empty catch blocks that carry a comment explaining the decision are not flagged.** That's the established idiom in this codebase and there are **209** of them, all reading like `catch { // endpoint is optional, fall through }`. Treating them as defects would either bury the signal or force 209 no-op edits. The rule looks for the silent kind: no statements *and* no explanation. Across `src/` that is currently **zero**, so this adds no new errors while closing both holes.

## Verification

A probe file containing a `.catch()` comment, a `.catch(() => {})`, a handler returning `{}`, and a commented empty catch produces no findings — while a genuinely silent catch in the same file is reported at its exact line. `pnpm run validate` passes on a clean tree.

### DIFF
--- a/scripts/build-validations.ts
+++ b/scripts/build-validations.ts
@@ -13,6 +13,7 @@ import { execSync } from "child_process";
 import fs from "fs";
 import path from "path";
 import { fileURLToPath } from "url";
+import ts from "typescript";
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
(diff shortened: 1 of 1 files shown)


Tara-ag:
## Review Summary for PR #1386

### Decision: **APPROVED** ✅

This PR improves the build validation script by replacing fragile regex-based detection of empty catch blocks with proper TypeScript AST parsing.

### Changes Made

**File Modified:** `scripts/build-validations.ts`

- Added `import ts from "typescript";` 
- Implemented `findEmptyCatchLines()` method that uses TypeScript's compiler API to properly parse catch clauses
- Replaced substring matching logic with AST-based analysis
- Better handles edge cases where `catch {}` spans multiple lines or is confused with comments like `.catch()`

### Why This Is Safe

1. **Infrastructure-only change**: Only modifies build validation scripts, not production code
2. **Improves accuracy**: The old regex approach had false positives/negatives; new AST approach is precise
3. **No breaking changes**: This is a pre-commit hook / CI validation script - callers don't care about its internal implementation
4. **Uses official TypeScript APIs**: `ts.createSourceFile`, `ts.isCatchClause`, etc. are stable compiler API
5. **Single file changed**: No risk of unintended side effects in other parts of the codebase

### Impact Analysis

- **Blast radius**: None (build scripts only)
- **Execution flows**: Build validation flow improved, but end-user flows unaffected  
- **Tests**: This script is run during build/validation, no test behavior changes expected
- **Dependencies**: Adds `typescript` as a runtime dependency (already a devDependency in most projects)

### Conclusion

This is a well-implemented improvement to the build tooling. The change is safer, more accurate, and follows best practices for AST-based analysis. No blocking issues found.

**Approved for merge.**

Tara-ag (approved):

Tara-ag (approved):

PR #1382, opened 2026-08-20
refactor(amazonBedrock): migrate both native turn loops onto the engine

murdore:
Plan 08 Task 4. Bedrock had two hand-rolled agentic turn loops — one for `Converse`, one for `ConverseStream` — and between them they duplicated the same content-block accumulation **three** times, because the streaming loop ran its first iteration inline and then again inside `processStreamResponse`. Both now call `runAgenticLoop`; everything wire-format-specific moves into a new `loopAdapter`.

**1411 lines deleted against 783 added**, and seven methods removed once nothing called them: `callBedrock`, `handleBedrockResponse`, `prepareStreamCommand`, `processStreamResponse`, `handleStreamStopReason`, `executeStreamTools`, `convertToAsyncIterable`.

## The characterization suite came first

It was written and committed against the **unmigrated** code, so the migration has a baseline it cannot quietly move. It drives the shipped surface only — `NeuroLink` from `dist/index.js`, nothing stubbed — and points the provider's real AWS SDK client at a local server via the documented `AWS_ENDPOINT_URL_BEDROCK_RUNTIME` variable. Requests are really signed with SigV4 and really routed, and assertions are about calls that actually go out on the wire. The stand-in speaks cleartext HTTP/2 (the Bedrock client defaults to `NodeHttp2Handler` and refuses HTTP/1.1 before sending anything) and answers streaming calls in the real `vnd.amazon.eventstream` binary framing, so the SDK's own decoder runs too. Its frame encoder is a local helper, not a new dependency — verified byte-for-byte against `@smithy/eventstream-codec` for every frame type used.

The brief specified the opposite — import from `src/` and replace the private client's `send` — on the grounds that *"no endpoint/baseURL override exists, so there is no way to redirect its AWS SDK v3 client at a local mock server through the public dist surface."* That is not so, and the difference earns its keep: counting real requests pins the call count more directly than counting calls to a replaced method, the real path exercises signing and event-stream decoding a replaced `send` skips, and it needs no rule-15 allow-list entry.

## Four defects it found, none of them in the brief

Two of the brief's own three cases assert behaviour this code does not have, and could not have passed as written.

| | Before | After |
|---|---|---|
| A caller's own tools | never execute | execute on both paths |
| Streaming step cap | `maxSteps` + 1 calls | exactly `maxSteps` |
| Reaching either cap | throws, losing all output | returns, `finishReason: "tool-calls"` |
| Runaway generate turn | **30** billed calls | **3** |

[...]

PR #1393, opened 2026-08-21
test(googleAiStudio): characterize both native turn loops before migrating

murdore:
Plan 08 Task 9, **Step 1**. AI Studio has **two** hand-rolled turn loops sharing one shape — the streaming one at `client.ts:1032` and a near-duplicate inside `generate()` at `:1500`, both `while (step < maxSteps)` over `models.generateContentStream`. Task 9 migrates both onto one adapter, so both are pinned first.

## How it tests

Shipped surface only: `NeuroLink` from `dist/index.js`, no imports out of `src/`, nothing stubbed. The real `@google/genai` SDK is redirected at a local server through the public `credentials.googleAiStudio.baseURL` option, which the provider threads into the SDK's own `httpOptions.baseUrl`. **No rule-15 exception, no allow-list entry.**

Three harness decisions carry forward lessons from earlier suites in this plan:

- **`disableInternalFallback` on every case.** A turn that ends unsuccessfully otherwise makes NeuroLink retry on a *different* provider, and whether that succeeds depends on unrelated credentials in the environment — exactly how the Anthropic suite passed locally and failed in CI.
- **Env snapshotted per case**, with `GOOGLE_AI_BASE_URL` cleared so an ambient value can't redirect the SDK.
- **No provider wording in assertion messages**, since the harness downgrades a matching failure to a skip.

## A defect found while characterizing

**`generate()` never declares a caller's tools.** Both loops build declarations from `options.tools`, but nothing merges the caller's tools into it on the generate path the way `BaseProvider.stream()` does for the streaming twin. So:

```ts
nl.generate({ provider: "google-ai", tools: { lookup: … } })
```

…cannot use those tools at all — the model is never told they exist.

This is the **same defect class already found and fixed on Bedrock's generate path** (#1382), which suggests it's systemic to native-loop providers rather than specific to either one.

It's pinned as-is rather than fixed here, with assertions labelled to invert when Task 9's migration corrects them. The two turns that case observes prove the *loop ran*, not that the tool worked — the fixture returns a canned `functionCall` the provider never declared, so the loop answers it with a not-found response and takes a second turn. I only found that by asserting declaration *before* execution; asserting execution alone reads as an execution bug when the real fault is upstream.

## Verification

Skip-hazard checked: a deliberately broken assertion reports `✗` and a non-zero exit, not `⊘`.

| Suite | Result |
|---|---|
| `aistudio-loop-characterization` | 3 passed (new) |
| `providers-mocked` | 54 passed |

Wired into `provider-safety-net`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,6 +125,9 @@ jobs:
       - name: Anthropic loop characterization (local SSE stand-in, no API keys required)
         run: pnpm run test:anthropic-loop-characterization
 
+      - name: AI Studio loop characterization (local SSE stand-in, no API keys required)
+        run: pnpm run test:aistudio-loop-characterization
+
       - name: 🧩 Provider Onboarding Completeness
         run: pnpm run verify:provider-onboarding
 
--- a/package.json
+++ b/package.json
@@ -214,7 +214,8 @@
     "test:vector-pinecone": "npx tsx test/continuous-test-suite-vector-pinecone.ts",
     "test:bedrock-loop-characterization": "tsx test/continuous-test-suite-bedrock-loop-characterization.ts",
     "test:sagemaker-streaming": "tsx test/continuous-test-suite-sagemaker-streaming.ts",
-    "test:anthropic-loop-characterization": "tsx test/continuous-test-suite-anthropic-loop-characterization.ts"
+    "test:anthropic-loop-characterization": "tsx test/continuous-test-suite-anthropic-loop-characterization.ts",
+    "test:aistudio-loop-characterization": "tsx test/continuous-test-suite-aistudio-loop-characterization.ts"
   },
   "files": [
     "dist",
--- a/test/continuous-test-suite-aistudio-loop-characterization.ts
+++ b/test/continuous-test-suite-aistudio-loop-characterization.ts
@@ -1,0 +1,336 @@
+#!/usr/bin/env tsx
+import "dotenv/config";
+
+/**
+ * Continuous Test Suite — Google AI Studio native-loop characterization
+ * (Plan 08, Task 9).
+ *
+ * AI Studio has TWO hand-rolled turn loops sharing one shape: the streaming
+ * one (`client.ts:1032`) and a near-duplicate inside `generate()`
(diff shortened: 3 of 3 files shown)


Tara-ag:
## Review Summary

**Decision: APPROVED** ✅

This PR adds a comprehensive test suite for Google AI Studio's turn loop behavior (Plan 08, Task 9). The changes are minimal and focused on CI infrastructure and test additions.

### Changes Overview

1. **`.github/workflows/ci.yml`**: Added new CI job step for "AI Studio loop characterization" - follows existing patterns perfectly
2. **`package.json`**: Added test script entry for the new suite
3. **`test/continuous-test-suite-aistudio-loop-characterization.ts`**: New E2E test suite testing:
   - Streaming path turn loops (text-only + tool interaction)
   - Generate() path (documents known defect for future migration)

### Quality Verification

✅ **Type Safety**: All types use `type` keyword (no interfaces)  
✅ **Rule 15 Compliance**: Test imports from `../dist/index.js` (E2E only)  
✅ **No Secrets**: No hardcoded credentials or API keys  
✅ **Environment Isolation**: Each test snapshots/restores env vars  
✅ **Error Handling**: Proper cleanup via try/finally blocks  
✅ **Documentation**: Clear comments explaining design decisions and known defects  
✅ **Testing Strategy**: Real integration tests against local HTTP server stand-in  

### Test Suite Analysis

The test suite properly:
- Tests both streaming and generate() paths separately
- Uses local HTTP server to avoid real API calls (CI-safe)
- Documents current state including known defects (generate() doesn't declare caller tools)
- Provides clear baseline for future migrations
- Avoids provider-specific wording in assertions (prevents false skips)

### Impact Assessment

- **Risk**: LOW - Only adds test infrastructure, no code changes
- **Breaking Changes**: None
- **Performance Impact**: None
- **Dependencies**: None added

### Recommendation

**APPROVE** - This is a well-structured addition of test coverage that documents current behavior and enables future migrations. The implementation follows all project standards and best practices.

Tara-ag (approved):

Tara-ag (approved):

PR #1401, opened 2026-08-21
feat(proxy): give proxy traffic a real token and cost ledger

murdore:
Part 3 of a four-PR stack. **Base is #1400** — merge #1399 and #1400 first.

Closes #1369
Closes #1370
Closes #1371

## The problem

The proxy served 300k+ requests on this machine with no usable accounting. Three separate defects, plus a pricing table that had fallen behind the models actually in flight.

## 1. The provider was hard-coded (#1370)

`ProxyTracer` passed the literal `"anthropic"` at **four** sites, so a model served by anyone else looked up a table it is not in.

The failure mode is subtler than it looks. The `anthropic` table has no `_default`, so `findRates` returns undefined and cost collapses to **$0** — *except* for a `claude-`named model routed elsewhere, which prices against the Anthropic table and yields a confident, wrong figure. **Two failure modes with opposite diagnostics:** silent zero is visible in aggregate and invisible per-request; plausible-wrong is the reverse.

The provider now travels on `ProxyRequestContext`, and the billing model is no longer `readonly`, so cost follows `setModelSubstitution` to the model that actually served the request.

## 2. `proxy analyze` reported half of usage and no spend (#1371)

It parsed `inputTokens` and both cache counters but never `outputTokens` and never a cost. It now prices each record, and distinguishes a **quoted** rate from one **inherited by name-prefix fallback**.

That distinction was itself wrong on first attempt: `isExactPricingMatch` re-implemented the lookup and missed Bedrock`s vendor-prefix strip and Vertex`s fallback to the Google table, mislabelling both as guesses. Two reviewers caught it independently. It now derives from `findRates` via an out-param, so there is one lookup.

## 3. Codex recorded nothing at all (#1369)

`codexProxyRoutes.ts` relayed `upstream.body` untouched and logged before a byte was read. A pass-through tap now scrapes `usage`, built so it **cannot break the relay**: bytes are enqueued before inspection, all parsing is wrapped, and an unrecognised stream degrades to the previous behaviour.

**The wire shape is verified, not assumed.** I ran the real Codex CLI through the proxy and captured the stream; the trimmed sample ships as `test/fixtures/codex-response-usage.sse`. The capture taught two things the docs did not: `response.created` arrives first carrying `usage: null` — which is why keeping the *last* non-null value is load-bearing — and `input_tokens_details.cache_write_tokens` existed and was being dropped, under-reporting cost.

[...]

PR #1400, opened 2026-08-21
refactor(proxy): put the CLI config writers behind one registry

murdore:
Part 2 of a four-PR stack. **Base is #1399**, so this diff shows only its own commit. Merge #1399 first.

Partially addresses #1368.

## Why

Three hand-written client config writers, invoked from **four duplicated blocks** inside a 5,800-line command module, had already produced two shipped bugs (#1366, #1367). The duplication is what let them diverge — they disagreed on whether to probe before writing, on what to return, and on how loudly to report failure.

## What changes

One `CliProxyClientConfigurator` contract — `detect` / `apply` / `restore` — one module per client under `src/cli/proxy-clients/`, a registry, and two loops replacing twelve hand-written calls. **`proxy.ts` sheds 516 lines.**

The contract makes each shipped bug unrepresentable:

| Was | Now |
| --- | --- |
| Claude created `~/.claude/settings.json` for an uninstalled CLI | `detect()` is required |
| OpenCode printed `✓` for work it did not do | `apply()` returns `boolean` |
| Every call site had to remember the `/v1` suffix | The suffix belongs to the client |

## Two bugs fixed along the way

- **Critical, caught in review:** `clearClaudeProxySettings` had no no-snapshot guard. With `__proxy_original_env` absent it read every managed key as "did not exist before" and **deleted** it — wiping a user`s real `ANTHROPIC_BASE_URL` with no error and no `false` return. Pre-existing, but `restoreAllClients` runs every client on every shutdown, so it became reachable. Reproduced, then fixed to match OpenCode and Qwen.
- **Failure visibility asymmetry:** the daemon path routed a failed write to `logger.debug` (invisible at default level) while the setup wizard printed a visible warning for the identical error. A client whose config could not be written keeps talking to its own upstream, which presents as the proxy silently not being used. Both paths now warn visibly.

## Two new clients, as proof the registry pays off

Each is **one file plus one registry line, touching `proxy.ts` not at all** — where the same change before this refactor was eleven edits, seven of them inside it.

- **Qwen Code** — settings shape read from a real `~/.qwen/settings.json` (`$version` 2), not written to spec.
- **Copilot CLI** — needed a different shape. Its provider config is **env-only** (`process.env["COPILOT_PROVIDER_BASE_URL"]`; `~/.copilot/config.json` carries no provider block), so rather than edit a shell profile it emits a sourceable `~/.neurolink/copilot-env.sh`, mode 0600, deleted on stop.

## Behaviour preserved

[...]

PR #1402, opened 2026-08-21
feat(proxy): add GET /accounts joining status, quota and per-account cost

### LINKED ISSUE #1372: Server-adapter API reference omits all three proxy route factories and the proxy flags
Found while auditing proxy CLI coverage at v11.2.3 (`8728bbb3`).

## Problem

`docs/guides/server-adapters/api-reference.md` is stale in two ways:

1. **`CreateRoutesOptions` is wrong.** The doc (`:657-660`) shows:
   ```ts
   type CreateRoutesOptions = { enableSwagger?: boolean; getRoutes?: () => RouteDefinition[]; };
   ```
   The real type (`src/lib/types/server.ts:1401-1409`) additionally has `proxy?`, `claudeProxy?` and `openaiProxy?`.

2. **The "Individual Route Factories" table (`:680-686`) lists none of the proxy factories** — no `createClaudeProxyRoutes`, `createOpenAIProxyRoutes` or `createCodexProxyRoutes` — even though the first two are re-exported from `src/lib/server/routes/index.ts` (`:23`, `:26`) as public API.

## Why it matters

These flags and factories are the **only** SDK-facing way to mount a proxy door. Neither existing CLI command exercises them: `neurolink proxy` assembles the factories by hand and never calls `createAllRoutes`, and `neurolink server` calls `createAllRoutes` but never passes any proxy flag (`grep -i proxy src/cli/commands/server.ts` → nothing).

So an SDK consumer relying on the published docs cannot discover that the proxy surface exists at all, and a contributor adding a fourth door has no documented seam to follow — they must read `routes/index.ts` directly.

## Fix

Correct the `CreateRoutesOptions` type and add rows for the three proxy factories, noting that `createCodexProxyRoutes` is currently CLI-only and not wired into `createAllRoutes`.

murdore:
Part 4 of a four-PR stack. **Base is #1401** — merge #1399, #1400 and #1401 first.

Closes #1372

## Why

Answering *"which of my accounts can still take work?"* took two calls whose schemas do not line up:

- **`/status`** — six rows here: three real logins plus `proxy/internal` and two `translation` pseudo-accounts. Counters, no quota.
- **`/limits`** — three rows, real logins only. Quota, no counters.

Every consumer reimplemented the same merge, discarded plumbing rows by hand, and drifted whenever internals moved. This does the join once, server-side, and adds what neither endpoint had: per-account tokens and cost.

```
sachin.sharma@juspay.in   exhausted   weekly=100%        -
hello@neurolink.ink       active      weekly= 32%  $1147.77
sachiny09@gmail.com       active      weekly=  1%        -
```

## `costBasis: "api-equivalent"` is not a bill

Pooled OAuth accounts bill by **subscription**. `costUsd` is what the recorded tokens *would* have cost at published rates — useful for judging whether a subscription earns its keep, and for spotting a runaway account. It reads around **$900/day** here, which is alarming and wrong without that label. `unpricedModels` names any model with no rate, so a $0 contribution is visible rather than silent.

## The incremental ledger

`accountLedger.ts` reads `proxy-<date>.jsonl` with a cursor per file, advanced only to the last complete newline. The log directory runs to **930 MB**, so a re-read per request is not an option. Measured: **40 ms cold, 1 ms warm.**

Four hazards, each with a test:

1. **A request is logged twice** when a streamed body`s tokens arrive late, so totals derive from a keyed map rather than a sum of lines — and token fields take the **max**, because a later token-less record (a terminal error after a success) would otherwise erase real usage. Reproduced: 1000 tokens became 0.
2. **The log is shared with the Codex engine**, and one email here is *both* an Anthropic and a ChatGPT login, so rows are filtered by engine before attribution or ChatGPT tokens land on the Anthropic row at the wrong vendor`s rates.
3. **The row set comes from the account list, never the log** — an account with no traffic today would vanish, which is exactly when you most want to see it. The only exhausted account here has zero log lines.
4. **Only `proxy-<date>.jsonl` is read.** `proxy-attempts-*` carries one row per retry.

Entries are keyed on `requestId` **plus account and model**, because `requestId` can come straight from a client-supplied `X-Request-ID` header and is not guaranteed unique.

## Normalisation the consumer would otherwise repeat

- **Two time units in one object.** `sessionResetAt` / `weeklyResetAt` / `windows[].resetsAt` are unix **seconds**; `lastUpdated` / `updatedAt` / `coolingUntil` are **milliseconds**. Only the seconds fields gain a `*Ms` companion — blanket-multiplying would throw the rest tens of thousands of years forward.
- **`severity` and `isActive`** are absent on header-sourced `overage` windows, a structural property rather than a transient gap. Filled with a documented fallback.
- **A `kind` discriminator** separates real logins from `internal` and `translation` plumbing.
- **`status` is derived**, not passed through — `/limits` returns `status: "snapshot"`, which describes how quota was obtained, not account health.

## Polled by design

Quota defaults to the **stored snapshot**. `?refresh=true` opts into a live fetch, which spends the user`s own OAuth credentials against Anthropic`s usage API — a dashboard on a short interval must not hammer it. Every data source is wrapped, so an upstream hiccup degrades to a partial row with `quotaError` set rather than 500-ing the dashboard.

## Also in this PR

Closes #1372: the server-adapter API reference declared `CreateRoutesOptions` as only `{ enableSwagger, getRoutes }` and listed none of the three proxy factories, so an SDK consumer could not discover the proxy surface existed. Also records that `createCodexProxyRoutes` is deliberately CLI-only.

## Testing

Six new cases covering each hazard above plus the route`s joined shape and its cost-basis label. Verified against real state: quota joined, `exhausted` correctly derived, timestamps normalised, all six windows filled.

`tsc --strict` clean · eslint 0 errors · proxy suite **51** passed · codex 29 passed.

---

**Split verification:** this branch`s tree is **byte-identical** to the original 24-commit branch, so the four-way split is provably lossless.

## Summary by CodeRabbit

* **New Features**
  * Added a `/accounts` endpoint with account status, quota, usage totals, request counts, and estimated API-equivalent costs.
  * Added support for distinguishing routed, internal, translation, and unrouted account entries.
  * Added incremental usage tracking with duplicate-request handling, file recovery, and unpriced-model reporting.

* **Bug Fixes**
  * Prevented duplicate tracking of model substitutions in streaming requests.

* **Documentation**
  * Documented the accounts endpoint, proxy CLI onboarding, route configuration options, and expanded proxy testing guidance.


murdore:
**Status update — out of draft, rebased, conflict resolved. Still gated on #1401.**

An earlier comment here described the `🔒 Single Commit Policy Validation` gate as counting commits against `release` instead of the PR base. The accurate mechanism:

- On a **`pull_request`** event the workflow uses the PR's own base (`BASE_BRANCH="origin/${{ github.base_ref }}"`) — stacks would be fine.
- The same workflow also fires on **`push`**, where `BASE_BRANCH` is hardcoded to `origin/release`. That is the run that fails here.
- The `pull_request` trigger is scoped to `branches: [release, main]`, so a PR based on another PR's branch never gets the passing run — and CodeRabbit likewise reports *"Auto reviews are disabled on base/target branches other than the default branch"*, meaning **this PR has had no automated review yet**.

This branch has been rebased onto its (rebased) predecessor; its patch is byte-identical to before the rebase, and it builds, typechecks and passes its suites standalone. It will get its full CI run — and its first real CodeRabbit/Yama review — once #1401 merges and this retargets to `release`.

murdore:
### Review cycle applied — three findings fixed

Worth stating plainly: **this PR had received no automated review at all.** CodeRabbit reported *"Auto reviews are disabled on base/target branches other than the default branch"* — its `pass` check means it declined, not that it found nothing — and neither Yama nor Copilot ran, because the CI workflows only trigger for PRs targeting `release`/`main`. That is ~800 lines including a reader that parses the proxy's request log and an endpoint that joins account credentials.

An adversarial review was run instead (two lenses, every finding independently refuted before it counted). Three defects here survived verification:

- **The ledger collapsed genuinely distinct requests.** It dedupes by `requestId` to fold the Codex double write, but `requestId` comes straight from a client-supplied `X-Request-ID` when present and nothing enforces uniqueness. A fixed correlation header or an idempotency wrapper made N real requests against the same account and model report as **one**, at a fraction of the tokens and cost actually spent. Two lines are now folded only when the earlier carries no usage — which is what the re-log actually looks like, since no writer emits two token-bearing lines for one request.

- **A real account could be tagged as plumbing.** Rows absent from the quota snapshot were all marked `kind: "internal"` — including an OAuth account disabled after a permanent refresh failure, or dropped from the allowlist. `listAnthropicAccountsForUsage()` skips exactly those. Since the docs tell consumers to filter `internal` rows out, the one account an operator is hunting for when traffic stops was the one that disappeared. Those rows now keep `kind: "account"` with `status: "unrouted"`.

- **The route's own test asserted nothing.** Both structural checks sat inside `for (const row of body.accounts)` / `for (const w of windows)`, and `body.accounts` is empty by construction in this suite — the handler is called against a fresh `mkdtemp` HOME with no seeded token store and no recorded traffic. I confirmed it empirically (`accounts.length = 0`), so both loops ran zero times and the test passed regardless of what the handler did. It now seeds real rows and pins the set non-empty; quota-window normalisation, which is only reachable from a live snapshot, is asserted directly rather than through a loop that never runs.

Every fix has a regression test that was watched failing first.

Verified standalone: `tsc --strict` clean, eslint 0 errors, `pnpm run build` clean, proxy suite 60 passed, codex suite 29 passed.

Still stacked on #1401, so its own CI and review cycle only start once that merges and this retargets to `release`.

murdore:
### Rebased onto `release`; carries one fix for a defect that reached `release` in #1401

#1401 merged (`23f7bb53`) while a correction to it was still in flight, so per the single-commit-per-PR policy that fix rides here instead of amending a merged commit.

**Every model substitution was reported twice.** The translation engine's streaming path records the substitution before the tracer prices anything — correct, and necessary, since `setUsage()`/`recordMetrics()` compute cost immediately off the tracer's current model and provider. But the `finally` block repeats the call under an identical guard, described there as harmless "for paths that never reach here".

It is not harmless. `translatedModel` is assigned in exactly one place (`proxyTranslationEngine.ts:461`), immediately before the first call and under the same condition, so there is no path where the `finally` guard passes and the early call did not run. `setModelSubstitution()` unconditionally does `modelSubstitutionTotal.add(1, …)`, so `proxy_model_substitution_total` — documented as *"Total proxy requests where the response model differs from the requested model"* — reads exactly **double** the real count on `release` right now.

Fixed by making the pair idempotent rather than deleting the fallback: a `substitutionRecorded` flag. The `finally` call still covers any future path that sets `translatedModel` elsewhere, but the two can no longer both count. That keeps the original intent intact; happy to be told the fallback should just go instead.

Span attributes were unaffected — `setAttributes` with the same values is idempotent. This is a metrics-accuracy defect, not a costing one.

---

This PR is also, finally, getting its **first real automated review**. While it was stacked, CodeRabbit reported *"Auto reviews are disabled on base/target branches other than the default branch"* and the CI workflows never triggered, because `pull_request` is scoped to `branches: [release, main]`. The green `CodeRabbit` check meant the bot declined, not that it found nothing.

Its own three findings, from the adversarial review run earlier, are described in the commit message and the previous comment: the ledger collapsing distinct requests that share a client-supplied `X-Request-ID`, real-but-unroutable accounts tagged as `internal` plumbing, and a route test whose assertions sat inside loops over an array that is empty by construction.

Verified locally: `tsc --strict` clean, eslint 0 errors, `pnpm run build` clean, proxy suite **63 passed**, codex suite **31 passed**.

Tara-ag:
## Review Summary

**Decision: APPROVED** ✅

This PR adds a new dashboard endpoint `GET /accounts` that joins quota snapshots, usage stats, and request log data into a single comprehensive response per account. This is a well-implemented addition with proper error handling, type safety, and test coverage.

### Findings

No issues found. The implementation:

✅ **Architecture**: Follows proxy patterns correctly, uses dynamic imports  
✅ **Type Safety**: All new types are properly documented with JSDoc comments  
✅ **Error Handling**: Gracefully degrades on failures (returns partial data)  
✅ **Test Coverage**: 9 new end-to-end tests covering all critical scenarios  
✅ **Documentation**: Comprehensive docs for the new endpoint and CLI onboarding  
✅ **Backward Compatibility**: No existing APIs modified or broken  
✅ **Security**: No hardcoded secrets or credential leaks  
✅ **Performance**: Efficient incremental ledger reading (~40ms cold, ~1ms warm)  

### Impact on Existing Code

- **248 nodes directly changed**
- **500 nodes impacted (within 2 hops)**
- **43 additional files affected**
- Risk score: 0.85 (moderate)

The changes touch the proxy module which has significant blast radius, but the implementation is self-contained and doesn't modify any existing public APIs. The new `/accounts` endpoint is additive only.

### Reviewed Files

1. **src/lib/types/proxyClient.ts** - Added 6 new type definitions with proper documentation
2. **src/lib/server/routes/claudeProxyRoutes.ts** - Added GET /accounts handler with comprehensive error handling
3. **test/continuous-test-suite-proxy.ts** - Added 9 new tests covering ledger deduplication, cost calculation, and account classification
4. **docs/features/** - Added 2 new documentation files plus index.md updates

### Review Scope

Focused on what static analysis cannot catch: logic/correctness bugs, security issues, backward compatibility of public SDK API, architectural rule violations, missing error handling, and test coverage. ESLint-enforced rules (type naming, no interfaces, etc.) were verified via automated checks.

**Approving this PR as it meets all quality gates and provides valuable functionality for monitoring pooled OAuth accounts.**

Tara-ag (approved):

Tara-ag (approved):

Tara-ag:
## 📊 Review Summary

**Decision: APPROVED** ✅

This PR adds a `/accounts` endpoint to the Claude proxy that provides a joined view of quota, usage stats, and request log data - addressing issue #1372.

### 🔍 What Was Reviewed

Reviewed all 9 changed files systematically using code knowledge graph impact analysis:
- **Directly modified**: 365 nodes
- **Impacted within 3 hops**: 500 nodes  
- **Additional affected files**: 45

### ✅ Files Assessed

1. **Documentation files** (`index.md`, `proxy-accounts-endpoint.md`, `proxy-cli-onboarding.md`) - ✅ Clean
2. **Claude Proxy Routes** (`claudeProxyRoutes.ts`) - ✅ Solid implementation with proper error handling
3. **Codex/OpenAI Proxy Routes** - ✅ Minimal/no direct changes needed
4. **Account Ledger** (`accountLedger.ts`) - ✅ Core logic correct with proper deduplication
5. **Usage Stats** (`usageStats.ts`) - ✅ Aggregation logic follows existing patterns
6. **Type Definitions** (`proxyClient.ts`) - ✅ Well-structured types in canonical location
7. **Test Suite** (`continuous-test-suite-proxy.ts`) - ✅ Comprehensive e2e tests

### 🎯 Key Strengths

- ✅ Follows CLAUDE.md Critical Rules (dynamic imports, backward compatibility, no interfaces)
- ✅ Comprehensive end-to-end test coverage
- ✅ Graceful error handling throughout
- ✅ Proper type safety with no `any` usage
- ✅ Follows factory+registry pattern
- ✅ No security vulnerabilities detected
- ✅ Respects CLI vs SDK separation

### 💡 One Suggestion

Consider adding JSDoc comments to exported utility functions (`normalizeQuotaForAccounts`, `quotaHealth`) in `claudeProxyRoutes.ts` for better developer documentation. This is optional and won't block merging.

### 📈 Impact on Existing Code

- **Blast radius**: 500 nodes impacted within 3 hops
- **Execution flows touched**: Proxy request/response flows, account lookup flows
- **Unmodified dependents at risk**: None - changes are additive to the proxy API surface
- **Architectural hotspots**: Changes touch ledger module which has moderate coupling, but implementation is self-contained

### 🎯 Verdict

**APPROVED** - The PR is production-ready. The feature is well-implemented, thoroughly tested, and doesn't introduce any risks or violations of project standards.

Tara-ag (approved):

Tara-ag:
🟡 MINOR: isPrimary field is hardcoded to false

Both row loops set `isPrimary: false`, so a consumer can never mark the configured primary account. The value is available: `routing?.primaryAccountKey ?? primaryAccountKey`, compared against `result.key`. Set it, or drop the field from CliAccountsRow so consumers do not read a value that is always false.

**Suggestion**: Update both row-building loops to set isPrimary by comparing result.key with the configured primary account key: routing?.primaryAccountKey ?? primaryAccountKey. For example: isPrimary: !!primaryKey && result.key === primaryKey

Tara-ag:
⚠️ MAJOR: Unrouted accounts lose usage data

This branch creates an 'unrouted' account row from usage stats but always sets `usage` to `null`. If the account served requests earlier today and was later disabled or excluded, `readAccountUsage()` has its token and cost totals, but this response drops them. Join `usageByAccount` here as in the routed-account loop.

**Suggestion**: Update the unrouted account-row construction to populate usage from the matching usageByAccount entry: usage: usageByAccount.get(entry.label) ?? null. This preserves token/cost data for accounts that were previously active.

PR #1403, opened 2026-08-21
test(googleAiStudio): characterize the breaker and step-cap fallback before migrating

murdore:
## Why this exists before the migration

Task 9 Step 3 moves both AI Studio loops onto `runAgenticLoop`. Two behaviours had no coverage, and both fail **silently** if a migration drops them.

### 1. Tool-failure breaker

The two dispatchers disagree:

- `executeNativeToolCalls` **clears** a tool's accrued strikes whenever it resolves that tool live (`googleNativeGemini3/utils.ts`, `failedTools.delete`).
- `runAgenticLoop` only ever **increments** them - there is no reset anywhere in the engine.

Measured rather than assumed. A tool that always throws, `maxSteps: 4`:

```
[diagnostic] failing-tool turn: attempts=2 calls=4
```

Dispatched **exactly twice**; all four model calls still happen. Both numbers asserted exactly - `attempts > 0` would still pass if the migration stopped tripping the breaker and dispatched all four times.

### 2. Step-cap fallback

Hitting `maxSteps` without a final answer does not end the turn silently: `handleMaxStepsTermination` substitutes the last step's text and pushes it to the consumer. A migration that simply let the engine return at the cap would deliver an **empty stream** - and silence is not something other tests notice.

## The second case was written wrong, and mutation-testing caught it

The obvious assertion - that the streamed text contains the model's words - **still passed with the fallback deleted**, because each step streams its own text live anyway:

```
with fallback:    chars=57   Passed: 5
fallback deleted: chars=38   Passed: 5   <- non-discriminating
```

What the fallback actually does is deliver the last step's text **one more time**. The case now counts occurrences (2 steps + 1 fallback = 3), and deleting the push makes it fail:

```
fallback deleted: Passed: 4   Failed: 1   x a turn that hits the step cap still delivers text
```

This is the same trap that has bitten this repo before: an assertion that looks like coverage but cannot fail.

## Scope

Test-only. No source changes.

PR #1404, opened 2026-08-21
test(core): prove the engine breaker matches AI Studio's dispatch count

murdore:
## Why

Task 9 Step 3 prerequisite, **proven rather than assumed**.

AI Studio's own dispatcher stops after `DEFAULT_TOOL_MAX_RETRIES` (2) failures. #1403 measured that end-to-end: exactly 2 dispatches across 4 model calls. Migrating that loop onto `runAgenticLoop` is only safe if the engine produces the same count from the same threshold - and nothing established that it did.

## Result

It does. With `toolFailureBreaker.maxRetries = 2` the engine dispatches an always-throwing tool **exactly twice**, and still **answers all four calls** - the model is told the tool is permanently failed rather than the turn stalling.

Mutation-verified: widening the engine's threshold by one makes this case *and* the existing breaker case both fail.

```
unmutated:  Passed: 32
maxRetries + 1:  Failed: 2
  x runAgenticLoop: tool failure breaker trips after maxRetries...
  x the engine's breaker dispatches an always-failing tool exactly twice at maxRetries 2
```

## It also settles the divergence that prompted the check

`executeNativeToolCalls` clears a tool's accrued strikes whenever it resolves that tool live (`failedTools.delete`), and the engine has no equivalent. That looked like a migration hazard.

It is not. **Neither implementation strikes on `TOOL_NOT_FOUND`** - only a thrown `execute` accrues a strike, which requires the tool to have been present and executable. So a tool cannot accrue strikes while it is deferred, and there is nothing for the reset to clear. The `delete` is vestigial, and the engine needs no reset to match AI Studio's behaviour.

This removes the one open question blocking the AI Studio loop migration.

## Scope

Test-only. No source changes.

## Summary by CodeRabbit

- **Tests**
  - Added coverage for repeated tool failures, ensuring failures are reported and processing continues until the configured step limit.
  - Verified that final accumulated text is delivered when the step limit is reached.
  - Added regression coverage confirming retry limits and execution records after the failure breaker activates.

### DIFF
--- a/test/continuous-test-suite-loop-engine.ts
+++ b/test/continuous-test-suite-loop-engine.ts
@@ -1077,4 +1077,49 @@ await test("an unresolvable tool name is recorded with an id and an error", asyn
   );
 });
 
+await test("the engine's breaker dispatches an always-failing tool exactly twice at maxRetries 2", async () => {
+  // Task 9 prerequisite, proven rather than assumed. AI Studio's own
+  // dispatcher stops after DEFAULT_TOOL_MAX_RETRIES (2) failures, which the
+  // characterization suite measured end-to-end as attempts=2 across 4 model
+  // calls. Migrating that loop onto this engine is only safe if the engine
+  // reproduces the same count from the same threshold.
+  let attempts = 0;
+  const adapter: AgenticLoopAdapter<string[]> = {
+    ...makeHydrationAdapter({
+      toolCallsByStep: [
+        [{ id: "c1", name: "flaky", args: {} }],
+        [{ id: "c2", name: "flaky", args: {} }],
+        [{ id: "c3", name: "flaky", args: {} }],
+        [{ id: "c4", name: "flaky", args: {} }],
+        [],
+      ],
+    }),
+    toolFailureBreaker: { maxRetries: 2 },
+  };
+  const { resultPromise } = runAgenticLoop(adapter, [], {
+    tools: {
+      flaky: {
+        execute: async () => {
+          attempts++;
+          throw new Error("synthetic tool failure");
+        },
+      },
+    },
+  });
+  const result = await resultPromise;
+  assertEqual(
+    attempts,
+    2,
+    "the breaker must stop dispatching after maxRetries failures",
+  );
+  // Every call is still ANSWERED — the model is told the tool is permanently
+  // failed rather than the turn stalling. Four calls, two dispatches.
+  const answered = result.toolExecutions.filter((e) => e.name === "flaky");
+  assertEqual(
+    answered.length,
+    4,
+    "every call must be answered even once the breaker has tripped",
+  );
+});
+
 await runSuite();


test/continuous-test-suite-loop-engine.ts
```diff
@@ -1077,4 +1077,49 @@ await test("an unresolvable tool name is recorded with an id and an error", asyn
...
+await test("the engine's breaker dispatches an always-failing tool exactly twice at maxRetries 2", async () => {
+  // Task 9 prerequisite, proven rather than assumed. AI Studio's own
+  // dispatcher stops after DEFAULT_TOOL_MAX_RETRIES (2) failures, which the
+  // characterization suite measured end-to-end as attempts=2 across 4 model
+  // calls. Migrating that loop onto this engine is only safe if the engine
+  // reproduces the same count from the same threshold.
+  let attempts = 0;
+  const adapter: AgenticLoopAdapter<string[]> = {
+    ...makeHydrationAdapter({
+      toolCallsByStep: [
```
Tara-ag, line 1089: 🔍 MINOR: Verify tool failure breaker semantics - This test verifies that toolFailureBreaker with maxRetries: 2 results in exactly 2 dispatch attempts. Need to verify this matches the actual implementation in loopEngine.ts. The test expects that with maxRetries: 2, there are exactly 2 dispatch attempts (not 3). This suggests the breaker stops AFTER hitting the max failures, not including them in the count. Please verify against the loopEngine implementation.

Tara-ag:
## Review Summary

**Decision: APPROVED** ✅

### Findings (0 critical issues):

🔍 **MINOR (1)**: Verify tool failure breaker semantics - The new test verifies that toolFailureBreaker with maxRetries: 2 results in exactly 2 dispatch attempts. This matches the expected AI Studio behavior where the breaker stops AFTER hitting the max failures. The test is well-written and provides good coverage for verifying the migration compatibility.

### Impact on existing code:
- **Scope**: Test-only changes (no source code modifications)
- **Risk**: None - this is purely adding test coverage
- **Files changed**: 1 file (`test/continuous-test-suite-loop-engine.ts`)
- **Lines added**: 45 lines (all new test code)

### Review scope:
- Verified the PR title accurately describes the change: "prove the engine breaker matches AI Studio's dispatch count"
- Reviewed the PR description which clearly explains the Task 9 Step 3 prerequisite
- Confirmed the test structure follows the existing patterns in the suite
- No security, correctness, or architectural concerns found

### Recommendation:
Approved to merge. This PR adds valuable test coverage for the agentic loop engine's tool failure breaker mechanism, ensuring the implementation matches AI Studio's behavior before the full migration.

Tara-ag (approved):

Tara-ag:
🔍 MINOR: Verify tool failure breaker semantics - This test verifies that toolFailureBreaker with maxRetries: 2 results in exactly 2 dispatch attempts. Need to verify this matches the actual implementation in loopEngine.ts. The test expects that with maxRetries: 2, there are exactly 2 dispatch attempts (not 3). This suggests the breaker stops AFTER hitting the max failures, not including them in the count. Please verify against the loopEngine implementation.

Tara-ag (approved):

PR #1405, opened 2026-08-21
refactor(googleAiStudio): run the native stream turn on the shared loop engine

murdore:
## What

Plan 08 Task 9 Step 3, **streaming half**. Replaces the hand-rolled loop in `executeNativeGemini3Stream` with `runAgenticLoop` + `createGeminiLoopAdapter`. The step cap, tool dispatch, the failure breaker and usage accumulation become engine-owned.

The generate-path loop (`executeNativeGemini3Generate`) is untouched and follows separately.

## Behaviour is preserved, not approximated

The characterization suite merged in #1403 was written specifically to gate this change, and the diagnostics are **byte-identical** either side of the migration:

```
failing-tool turn: attempts=2 calls=4
step-cap turn:     calls=2 chars=57
```

## Four things that needed deliberate handling

**1. Breaker threshold.** `toolFailureBreaker` uses `DEFAULT_TOOL_MAX_RETRIES` - the same constant the hand-rolled dispatcher used. #1404 proved the engine reproduces the same *dispatch count* from that threshold, rather than merely having a breaker of its own.

**2. `planReclaim` is pure.** The old code reclaimed in place because it owned `currentContents`; the engine assigns what the hook returns.

**3. The wrapped `buildToolResultMessages` carries every per-step side effect.** It is the only hook that runs once per step with that step's results, so it does the `gen_ai.tool_call` span events, the storage write with its `thoughtSignature` and `stepIndex`, and `contextGuard.noteAppendedChars` for the tool results that ride the next prompt. Reading these off the final result would batch every step into one late write and lose the per-step signature.

**4. The step-cap fallback still fires** - and this one was verified against the engine source rather than inferred from a passing test. The engine sets `hadToolCallsAtCap` but does **not** break; it still dispatches and calls `buildToolResultMessages` on the final step. So the wrapper's counter reaches `maxSteps` and `handleMaxStepsTermination` pushes the last step's text exactly as before.

## Dead machinery removed

`executeMap`, `originalNameMap`, `failedTools` and `collectStreamChunksIncremental` are no longer used by this path - the adapter owns name translation and collection, the engine owns dispatch and the breaker. All four are stream-local; the generate path keeps its own.

## Tests

- Characterization suite: 5/5, identical diagnostics
- `test:providers-mocked`: 54/54
- `test:provider-structure`: PASS
- Typecheck clean across 4858 files, eslint clean

## Summary by CodeRabbit

* **Improvements**
  * Improved Google AI Studio’s streaming tool execution for more consistent agent interactions.
  * Enhanced handling of tool results, retries, context limits, usage tracking, telemetry, and termination during streamed responses.
  * Preserved existing conversation storage and per-step monitoring behavior.

### DIFF
--- a/src/lib/providers/googleAiStudio/client.ts
+++ b/src/lib/providers/googleAiStudio/client.ts
@@ -48,6 +48,9 @@ import {
 } from "../../types/index.js";
 import { ERROR_CODES, NeuroLinkError } from "../../utils/errorHandling.js";
 import { logger } from "../../utils/logger.js";
+import { createGeminiLoopAdapter } from "../../core/geminiLoopAdapter.js";
+import { runAgenticLoop } from "../../core/loopEngine.js";
+import { DEFAULT_TOOL_MAX_RETRIES } from "../../core/constants.js";
 import { isToolsSchemaExclusionInForce } from "../../core/modules/structuredOutputPolicy.js";
 import {
   GEMINI_ELISION_NOTE,
@@ -71,7 +74,6 @@ import {
   buildGeminiResponseSchema,
   buildNativeConfig,
   collectStreamChunks,
-  collectStreamChunksIncremental,
   computeMaxSteps,
   createContextGuard,
   buildUserPartsWithMultimodal,
@@ -911,8 +913,6 @@ export class GoogleAIStudioProvider extends BaseProvider {
 
           // Convert tools
           let toolsConfig: NativeToolsConfig | undefined;
-          let executeMap: Map<string, Tool["execute"]> = new DedupExecuteMap();
-          let originalNameMap = new Map<string, string>();
           let declarationsResult: NativeToolDeclarationsResult | undefined;
 
           if (
@@ -926,8 +926,6 @@ export class GoogleAIStudioProvider extends BaseProvider {
             );
             declarationsResult = result;
             toolsConfig = result.toolsConfig;
-            executeMap = result.executeMap;
-            originalNameMap = result.originalNameMap;
 
             logger.debug("[GoogleAIStudio] Converted tools for native SDK", {
               toolCount: toolsConfig[0].functionDeclarations.length,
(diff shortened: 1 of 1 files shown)


Tara-ag:
💡 MINOR: The executeNativeGemini3Stream function lacks dedicated test coverage in existing continuous test suites. Consider adding integration tests covering the migrated code path with various scenarios (tool calls, step limits, abort signals, max steps termination) in `test/continuous-test-suite-*.ts`.

---

## Review Summary

### Decision: **APPROVED** ✅

This PR refactors the GoogleAIStudio provider's native Gemini 3 streaming loop to use the shared agentic loop engine (`runAgenticLoop` + `createGeminiLoopAdapter`). This aligns with the established migration pattern from Task 4 (Bedrock migration).

### Changes Made:
1. **Migration**: Replaced hand-rolled streaming loop with `runAgenticLoop` + adapter
2. **New imports**: Added `createGeminiLoopAdapter` and `runAgenticLoop` from core modules
3. **Pump mechanism**: Background async pump pushes chunks from engineStream to channel
4. **Error handling**: Proper cleanup via `await pump.catch(() => {})` on error
5. **Usage tracking**: Token accumulation from `engineResult.usage.*` fields
6. **Tool execution storage**: Already handled by adapter's `buildToolResultMessages` method

### Impact Analysis:
- **Changed files**: 1 (`src/lib/providers/googleAiStudio/client.ts`)
- **Lines changed**: 294 (152 additions, 142 deletions)
- **Blast radius**: ~500 nodes impacted, 96 additional files affected
- **Affected flows**: `generate` (criticality 0.78), `embed`, `embedMany`, `constructor`
- **Risk score**: 0.55 (moderate - expected for this level of refactoring)

### Verification:
✅ No breaking changes to public SDK API (internal implementation only)  
✅ Proper error handling and pump cleanup  
✅ Correct abort signal propagation via `composeAbortSignals` utility  
✅ Following established patterns (Task 4 Bedrock migration)  
✅ Type safety maintained  
✅ No architectural rule violations per CLAUDE.md Critical Rules

### Recommendation:
Approve this PR. The refactoring is sound, follows established patterns, and maintains backward compatibility. The minor test coverage gap should be addressed in a follow-up PR rather than blocking this one.

PR #1406, opened 2026-08-21
test(googleAiStudio): make the breaker case assert the actual failure response

murdore:
## What

CodeRabbit's finding on #1403 - verified before acting on it, and correct.

The breaker case asserted:

```ts
const responses = JSON.stringify(server.calls.slice(1).map((c) => c.body));
assert(responses.includes("flaky"), "the failed dispatch was never reported back to the model");
```

That can **never fail**. The tool's name appears in the function *declarations* on every single request, whether or not any result came back.

## Proven by mutation, not by reading

Replacing every `functionResponse` the native dispatcher builds with a stub broke two other cases in this file and left this one green:

```
mutated, old assertion:  Passed 3   Failed 2    <- this case still green
mutated, new assertion:  Passed 2   Failed 3    <- this case now fails
unmutated:               Passed 5   RESULT: PASS
```

## The fix

Structural instead of a substring scan: a `functionResponse` under the tool's own name must come back on the final request, and its payload must carry a failure rather than a result.

## Note on process

This is the **second** non-discriminating assertion in this file. The first - the step-cap case - I caught myself by mutation-testing before committing. This one shipped because I did not do the same here. Both are now covered by the standing rule to mutate every new characterization assertion before it lands, which is exactly the check that would have caught this one for free.

## Scope

Test-only. No source changes.

### DIFF
--- a/test/continuous-test-suite-aistudio-loop-characterization.ts
+++ b/test/continuous-test-suite-aistudio-loop-characterization.ts
@@ -133,6 +133,19 @@ function functionResponses(call: StandInCall | undefined): string[] {
     .filter((n): n is string => typeof n === "string");
 }
 
+/** functionResponse parts with their payloads, for a given request. */
+function functionResponsePayloads(
+  call: StandInCall | undefined,
+): Array<{ name?: string; response?: unknown }> {
+  const contents = (call?.body?.contents ?? []) as Array<{
+    parts?: Array<{ functionResponse?: { name?: string; response?: unknown } }>;
+  }>;
+  return contents
+    .flatMap((c) => c.parts ?? [])
+    .map((p) => p.functionResponse)
+    .filter((r): r is { name?: string; response?: unknown } => r !== undefined);
+}
+
 async function startStandIn(
   reply: (callIndex: number) => string,
 ): Promise<StandIn> {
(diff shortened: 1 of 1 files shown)


Tara-ag:
## Review Summary

**Decision: APPROVED** ✅

This is a **test-only change** that fixes a broken assertion in the Google AI Studio loop characterization test suite.

### What Changed

- **File:** `test/continuous-test-suite-aistudio-loop-characterization.ts`
- **Changes:** 
  - Added new helper function `functionResponsePayloads()` to extract both tool names AND response payloads from function responses (lines 136-147)
  - Updated assertion in "repeatedly failing tool" test case to use structural checking instead of substring scanning (lines 410-436)

### Why This Fix Is Correct

The old assertion searched for the tool name as a substring in the serialized request body (`JSON.stringify(server.calls.slice(1).map((c) => c.body))`). This approach was fundamentally broken because:

1. Tool names appear in **function declarations** on every single request, regardless of whether any results came back
2. The assertion could never fail - mutation testing proved this by showing it remained green even when all functionResponses were replaced with stubs

The new assertion properly verifies:
1. A function response under the tool's own name ("flaky") was returned to the model
2. That response carries a failure/error rather than a successful result

### Impact Analysis

- **Scope:** Test-only changes only
- **Production impact:** None
- **Risk:** Low - modifies existing test assertions, no source code changes
- **Breaking changes:** None

### Code Quality

✅ Well-documented with inline comments explaining the reasoning  
✅ Proper TypeScript typing with type guards  
✅ Follows the project's determinism exception rule (test suite driving live behavior)  
✅ No security concerns  
✅ No API changes  

### Final Assessment

This is a **sound correctness fix** for a test that was inadvertently checking the wrong thing. The change makes the test actually verify what it should (that failed tool dispatches are answered back as failures, not just declared). The implementation is clean, well-reasoned, and appropriately scoped to test-only code.

**Approved.**

Tara-ag (approved):

Tara-ag (approved):

PR #1430, opened 2026-08-21
test(googleVertex): pin the context-guard stop before migrating

murdore:
## What

Closes the last uncharacterized behaviour in Vertex's turn loop ahead of Plan 08 Task 10, so the migration has a complete safety net (8 cases now).

## Why this one matters more than it looks

The context guard is the **only** loop control that can END a turn rather than trim it:

- Reclaiming returns a smaller conversation and the turn continues.
- When reclaiming buys nothing, the loop sets `hitContextLimit`, breaks, and **synthesizes a final answer** rather than stepping into a provider `"prompt is too long"` rejection.

That stop is specifically awkward for the migration, which is why it needs pinning before rather than after: **the engine's `planReclaim` hook cannot express it.** The hook says "here is a smaller conversation" or "nothing to do" — it has no way to say "end the turn". So the migration has to solve it deliberately, most likely by aborting a client-owned signal composed into `abortSignal`, which is the same place the turn clock has to live (see #1409, where the inert adapter timeout fields were removed for exactly this reason).

This case is what holds that honest.

## Measured

With a stand-in reporting a 900k-token prompt and `maxSteps: 6`:

```
[diagnostic] context-cap turn: outcome=ended calls=2 stopReason=context-cap chars=219
```

The turn stops after **2** calls, well short of its step budget, and still returns a synthesized answer.

The guard trips on the prompt size a call *reports*, and the stand-in owns `usageMetadata` — so the pressure is applied without building a genuinely enormous conversation. Deterministic and fast.

## Mutation-verified

Forcing `shouldStop()` false at all four sites lets the turn run its full step budget and the case fails **cleanly**:

```
exit 1   Passed: 7   Failed: 1
x a turn pressing against the context window stops and still answers
```

Not a hang — the whole turn is bounded, per the lesson from the stall case.

Test-only. Typecheck clean across 4867 files, eslint clean.

### DIFF
--- a/test/continuous-test-suite-vertex-loop-characterization.ts
+++ b/test/continuous-test-suite-vertex-loop-characterization.ts
@@ -272,6 +272,66 @@ async function startDribblingStandIn(): Promise<StandIn> {
   };
 }
 
+/**
+ * A stand-in that reports an enormous prompt size.
+ *
+ * The context guard trips on the FULL prompt tokens a call reports, and the
+ * stand-in owns `usageMetadata` — so a turn can be pushed against the window
+ * without building a genuinely enormous conversation.
+ */
+async function startContextPressureStandIn(): Promise<StandIn> {
+  const calls: StandInCall[] = [];
+  const server: Server = createServer((req, res) => {
+    const chunks: Buffer[] = [];
+    req.on("data", (c: Buffer) => chunks.push(c));
+    req.on("end", () => {
+      const parseBody = (): Record<string, unknown> => {
+        try {
+          return JSON.parse(Buffer.concat(chunks).toString("utf8") || "{}");
+        } catch {
+          return {};
+        }
+      };
+      calls.push({
+        body: parseBody(),
+        path: String(req.url ?? "").split("?")[0],
+      });
+      res.writeHead(200, { "content-type": "text/event-stream" });
+      res.write(
+        `data: ${JSON.stringify({
+          candidates: [
+            {
+              content: {
+                parts: [{ functionCall: { name: "lookup", args: {} } }],
+                role: "model",
+              },
+              finishReason: "STOP",
+              index: 0,
+            },
+          ],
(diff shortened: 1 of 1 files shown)


Tara-ag (commented):
## Review Summary for PR #1430

**Decision:** APPROVED ✅

### Changes Overview
This PR adds a new test case to `test/continuous-test-suite-vertex-loop-characterization.ts` that verifies the context guard functionality in Vertex's native loop implementation.

### What Was Added
1. **New helper function** `startContextPressureStandIn()` (lines 276-333): Creates an HTTP server stand-in that simulates a provider response with artificially inflated token counts (900k prompt tokens) to trigger the context window guard.

2. **New test case** "context guard" section (lines 766-839): Tests that when a turn approaches the context window limit:
   - The loop detects the context limit before making excessive API calls
   - The turn ends gracefully and synthesizes a final answer
   - No more than ~5 calls are made (well under the maxSteps=6 budget)

### Verification Completed
✅ **Test file only** - No production code changes
✅ **End-to-end testing** - Uses built SDK from `dist/index.js` per rule-15
✅ **No secrets** - All mock data uses placeholder values
✅ **Type safety** - Proper TypeScript usage throughout
✅ **Error handling** - Proper timeouts and cleanup in try/catch/finally
✅ **Zero impact on existing code** - Graph analysis confirms 0 flows affected

### Quality Assessment
This is a well-designed test that:
- Isolates the context guard behavior without requiring real large conversations
- Verifies the migration will handle context limits correctly
- Includes appropriate diagnostics logging
- Has proper resource cleanup

### Impact Analysis
- **Blast radius**: 0 (test-only change)
- **Affected flows**: None
- **Breaking changes**: None
- **Production impact**: None

**This PR is safe to merge.** It adds valuable test coverage for a critical edge case in the Vertex loop implementation without any risk to production code.

PR #1436, opened 2026-08-21
feat(core): let a Gemini adapter supply its own step collector

murdore:
## Why

Task 10 prerequisite, found by *starting* that migration rather than assumed in advance.

`createGeminiLoopAdapter` hard-coded `collectStreamChunksIncremental`. AI Studio shares that helper, so the fit was exact. **Vertex does not**: its loop drains the stream inline over ~90 lines, folding cumulative usage counts as deltas and capturing thought signatures its own way.

Migrating Vertex onto the adapter as it stood would have silently swapped its collector for a different one — a behaviour change wearing a refactor's clothes, and precisely what this plan's characterization suites exist to prevent.

## What

The collector becomes an optional hook. Providers matching the shared helper pass nothing and are unaffected; a provider whose drain genuinely differs supplies `collectStep` and keeps its measured behaviour.

## Verified both directions

**The default path is untouched** — AI Studio's characterization suite is byte-identical:

```
attempts=2 calls=4
calls=2 chars=57
generate content="partial progress"
Passed: 6
```

**The hook is honoured, not decorative** — a new loop-engine case asserts the supplied collector is used for text and usage, and mutation-testing it (forcing the default branch) fails exactly that case:

```
unmutated:      Passed: 33
hook disabled:  Passed: 32   Failed: 1
```

Typecheck clean across 4829 files, eslint clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming response handling by supporting provider-specific collection for each step.
  * Preserved shared incremental collection as the default behavior.
  * Ensured multi-step tool interactions correctly retain cumulative token usage and final response text.

* **Tests**
  * Added regression coverage for custom stream collection across multiple execution steps.

### DIFF
--- a/src/lib/core/geminiLoopAdapter.ts
+++ b/src/lib/core/geminiLoopAdapter.ts
@@ -141,13 +141,19 @@ export function createGeminiLoopAdapter(
     ): Promise<AgenticLoopStepResult<GeminiStepRaw>> {
       const rawStream = await config.sendStep(request.raw, signal);
 
-      // `collectStreamChunksIncremental` wants a StreamChannel but only ever
-      // calls `.push`, so the engine's push-only channel satisfies it. The
-      // helper is reused rather than reimplemented precisely because it also
-      // owns usage extraction and thought-signature preservation.
-      const collected = await collectStreamChunksIncremental(rawStream, {
-        push: (chunk: { content: string }) => channel.push(chunk),
-      } as Parameters<typeof collectStreamChunksIncremental>[1]);
+      // The shared helper by default: it owns usage extraction and
+      // thought-signature preservation, and it wants a StreamChannel but only
+      // ever calls `.push`, so the engine's push-only channel satisfies it.
+      //
+      // A provider whose drain genuinely differs supplies `collectStep`
+      // instead. Vertex does: it folds cumulative usage counts as deltas in
+      // its own loop, and that behaviour is characterized, so it keeps its
+      // collector rather than being quietly switched to this one.
+      const collected = config.collectStep
+        ? await config.collectStep(rawStream, channel)
+        : await collectStreamChunksIncremental(rawStream, {
+            push: (chunk: { content: string }) => channel.push(chunk),
+          } as Parameters<typeof collectStreamChunksIncremental>[1]);
 
       // The provider's context guard calibrates from real per-step counts;
       // `inputTokens` is this step's full prompt size, which is what it
--- a/test/continuous-test-suite-loop-engine.ts
+++ b/test/continuous-test-suite-loop-engine.ts
@@ -1122,4 +1122,64 @@ await test("the engine's breaker dispatches an always-failing tool exactly twice
   );
 });
 
+await test("the Gemini adapter uses a supplied collectStep instead of the shared collector", async () => {
+  // Task 10 prerequisite. Vertex drains its stream itself — folding cumulative
+  // usage counts as deltas and capturing thought signatures its own way — and
+  // that behaviour is characterized, so migrating it must not silently swap in
+  // the shared collector. This proves the hook is honoured rather than
+  // decorative.
+  const { createGeminiLoopAdapter } =
+    await import("../src/lib/core/geminiLoopAdapter.js");
+  // Two steps, not one. A single-step turn would pass even if the adapter
+  // consulted the hook once and then fell back to the shared collector, so
+  // the first step asks for a tool and the second ends the turn.
+  let called = 0;
+  const adapter = createGeminiLoopAdapter({
+    providerLabel: "probe",
+    maxSteps: 3,
+    liveTools: {},
+    buildRequest: () => ({}),
+    sendStep: async () => ({}) as never,
+    collectStep: async () => {
+      called++;
+      if (called === 1) {
+        return {
+          rawResponseParts: [{ text: "step one" }],
+          stepFunctionCalls: [{ name: "probe_tool", args: {} }],
+          inputTokens: 4,
+          outputTokens: 1,
+          finishReason: "STOP",
+        };
+      }
+      return {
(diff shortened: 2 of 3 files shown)


src/lib/core/geminiLoopAdapter.ts
```diff
@@ -141,13 +141,19 @@ export function createGeminiLoopAdapter(
...
       const rawStream = await config.sendStep(request.raw, signal);
 
-      // `collectStreamChunksIncremental` wants a StreamChannel but only ever
-      // calls `.push`, so the engine's push-only channel satisfies it. The
-      // helper is reused rather than reimplemented precisely because it also
-      // owns usage extraction and thought-signature preservation.
-      const collected = await collectStreamChunksIncremental(rawStream, {
-        push: (chunk: { content: string }) => channel.push(chunk),
-      } as Parameters<typeof collectStreamChunksIncremental>[1]);
+      // The shared helper by default: it owns usage extraction and
```
Tara-ag, line 144: 💬 SUGGESTION: Consider condensing the comment to focus on the key distinction: 'A provider whose drain genuinely differs supplies collectStep instead. Vertex does: it folds cumulative usage counts as deltas in its own loop, and that behaviour is characterized, so it keeps its collector rather than being quietly switched to this one.'

test/continuous-test-suite-loop-engine.ts
```diff
@@ -1122,4 +1122,44 @@ await test("the engine's breaker dispatches an always-failing tool exactly twice
...
+await test("the Gemini adapter uses a supplied collectStep instead of the shared collector", async () => {
+  // Task 10 prerequisite. Vertex drains its stream itself — folding cumulative
+  // usage counts as deltas and capturing thought signatures its own way — and
+  // that behaviour is characterized, so migrating it must not silently swap in
+  // the shared collector. This proves the hook is honoured rather than
+  // decorative.
+  const { createGeminiLoopAdapter } =
+    await import("../src/lib/core/geminiLoopAdapter.js");
+  let called = 0;
+  const adapter = createGeminiLoopAdapter({
```
Tara-ag, line 1137: 💡 MINOR: Consider verifying multiple steps - The test at line 1125 only calls collectStep once (maxSteps: 2 but the fake adapter returns STOP after first step). Consider adding a second step to verify the collector is called for each step.

Tara-ag (changes requested):
## Review Summary

### Decision: CHANGES_REQUESTED

This PR implements a critical architectural change to support Vertex provider's unique stream draining behavior by adding a custom `collectStep` hook to the GeminiLoopAdapter.

### Findings

**✅ No blocking issues found:**
- No CRITICAL security vulnerabilities
- No MAJOR breaking changes or architectural violations  
- No hardcoded secrets or API keys
- Changes properly tested with new regression test

**💡 Suggestions (non-blocking):**
1. Line 144 in `src/lib/core/geminiLoopAdapter.ts`: Comment could be more concise
2. Test file has a valid point about verifying multiple steps

### Impact Analysis
- **Directly changed**: 58 nodes
- **Impacted within 2 hops**: 500 nodes
- **Additional files affected**: 118 files
- This is expected given the wide reach of the agentic loop mechanism

### File-by-File Review

**File 1: src/lib/core/geminiLoopAdapter.ts** ✅
- Correctly adds `collectStep` hook allowing providers to supply custom stream collection logic
- Vertex uses this to fold cumulative usage counts as deltas instead of relying on shared collector
- Properly documented with clear comments explaining why this differs from other providers

**File 2: src/lib/types/loopEngine.ts** ✅  
- Type-only changes adding `collectStep` field to `GeminiLoopAdapterCoreConfig`
- No behavioral changes, just extending the interface

**File 3: test/continuous-test-suite-loop-engine.ts** ✅
- Comprehensive test verifies custom collectStep is honored by adapter
- Tests both that the collector was called and that results came from it
- Validates inputTokens and text content from custom collector

### Conclusion
The implementation is correct and well-tested. The two minor suggestions are optional improvements that don't block merging but would enhance code quality if addressed.

Tara-ag:
## Review Summary

### Decision: **APPROVED** ✅

This PR adds a critical feature for migrating Google AI Studio's Gemini adapter to use the shared agentic loop engine while maintaining compatibility with Vertex's different streaming behavior.

### Changes Made:

1. **geminiLoopAdapter.ts**: Added optional `collectStep` hook parameter to `createGeminiLoopAdapter`. Providers can now supply their own step collector; if omitted, it defaults to `collectStreamChunksIncremental`.

2. **loopEngine.ts**: The `CollectedChunkResult` type was already defined (used by the collectStep hook).

3. **Test added**: Verifies that a custom `collectStep` function is actually invoked and produces the expected results.

### Impact Analysis:

- **Non-breaking change**: The `collectStep` parameter is optional and has a safe default
- **Affects two providers**: Google AI Studio and Vertex (both use `createGeminiLoopAdapter`)
- **Execution flows affected**: AgenticLoop adapters for Gemini models
- **Risk level**: Low - this is a refactoring/migration pattern, not a behavioral change

### CLAUDE.md Compliance:

✅ Rule 5 (Backward compatibility): Maintained - existing callers continue to work unchanged
✅ Rule 1 (Dynamic imports): Not applicable - this is in core, not provider registration
✅ Rule 6 (formatProviderError): Not applicable - no error formatters changed
✅ Streaming pattern: Not violated - this affects how token usage is collected, not the streaming call itself

### Testing:

The PR includes a comprehensive test that verifies:
- Custom collector is actually called (not bypassed)
- Usage counts come from the custom collector
- Response text comes from the custom collector

### Recommendation:

**Approve** - This is a well-designed, backward-compatible refactoring that enables proper migration of the Gemini adapter to the shared agentic loop engine. The test coverage is excellent and the implementation follows the established patterns.

Tara-ag (approved):

PR #1438, opened 2026-08-21
test(googleVertex): pin that streamed text is replayed, not live

murdore:
## What

Task 10 groundwork, and the answer to the design question my aborted drain extraction raised.

**Vertex does not stream text as it arrives.** Each text part is preserved in `incrementalTextChunks` during the turn, and `createTextStream()` replays them once the whole turn is finished. The code says so at `client.ts:2642`:

> The SDK already gave us the chunks during the for-await loop above; we just preserved them.

## Why pin it

Nothing pinned that, and it is precisely what migrating onto the engine's per-chunk `channel.push` model would change: the consumer would start receiving text **mid-turn** instead of at the end.

That is arguably an improvement — but it is user-visible timing, and it should be a decision rather than a side effect of a refactor. This case makes it one. If the migration makes streaming live, this is the assertion to consciously invert.

## How it distinguishes the two models

A two-step turn separates them cleanly: under live streaming the first chunk arrives **before** the second provider call; under replay it cannot.

```
[diagnostic] stream timing: calls=2 firstChunkAfterSecondCall=true
```

## Guarded against vacuity

Comparing against a `secondCallAt` that never advanced would hold trivially against zero, so the case also asserts the second call happened and the turn really took two steps.

Verified by forcing a single-step fixture — it fails with `expected a two-step turn, got 1 calls` rather than passing quietly:

```
Passed: 8   Failed: 1
x text is replayed after the turn completes, not streamed live
```

Test-only; the Vertex suite is now 9 cases.

### DIFF
--- a/test/continuous-test-suite-vertex-loop-characterization.ts
+++ b/test/continuous-test-suite-vertex-loop-characterization.ts
@@ -838,4 +838,80 @@ await test("a turn pressing against the context window stops and still answers",
   );
 });
 
+section("stream timing");
+
+await test("text is replayed after the turn completes, not streamed live", async () => {
+  // Vertex does NOT stream text as it arrives. The loop preserves each text
+  // part in `incrementalTextChunks` and `createTextStream()` replays them
+  // once the whole turn is done — the code says as much at client.ts:2642,
+  // "we just preserved them ... instead of collapsing into finalText".
+  //
+  // Nothing pinned that, and it is exactly what a migration onto the engine's
+  // per-chunk push model would change: the consumer would start receiving
+  // text mid-turn instead of at the end. That may well be an improvement, but
+  // it is a user-visible timing change and must be a decision rather than a
+  // side effect. This case makes it one.
+  //
+  // A two-step turn separates the two models cleanly: under live streaming
+  // the first chunk arrives BEFORE the second provider call; under replay it
+  // cannot.
+  let secondCallAt = 0;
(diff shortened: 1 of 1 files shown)


Tara-ag:
## Yama Code Review Summary for PR #1438

**Decision: APPROVED**

### Findings
No issues found. This PR adds a new e2e test case that validates Vertex provider stream timing behavior.

### Impact on Existing Code
- **Changed files:** 1 (test/continuous-test-suite-vertex-loop-characterization.ts)
- **Blast radius:** 24 nodes directly changed, 500+ nodes impacted within 2 hops (mostly test helpers/framework code)
- **Production code affected:** None - this is purely a test addition
- **Execution flows touched:** Test harness only
- **Architectural hotspots:** Not applicable - test file only
- **Breaking changes:** None - SDK API unchanged

### Review Scope
- ✅ File-by-file review completed (single test file)
- ✅ E2E test requirements verified (imports from dist/, not src/)
- ✅ Proper error handling with try/finally cleanup
- ✅ Logic correctness validated (stream timing assertion)
- ✅ Graph impact analysis confirms contained changes

### Conclusion
This is a defensive test addition that validates documented streaming behavior. The change is self-contained, poses no risk to production code, and follows all project conventions. The test helps catch regressions if streaming semantics change in future migrations.

**Safe to merge.**

Tara-ag (approved):

Tara-ag (approved):

PR #1442, opened 2026-08-21
ci(workflow): make the optional-deps smoke test blocking

murdore:
## What

Makes **CLI Smoke Test (optional deps omitted)** blocking, and replaces the comment that justified it not being.

## Why now

The job spent its entire life non-blocking and red, and **never once ran a smoke step**:

1. It installed `--no-optional` *before* building — which removed rolldown's own native binding, so the build tool disappeared and it died at "Build CLI". Fixed in ea91c53d.
2. It then built with `pnpm run build:cli` on a clean checkout, where `collapse-cli-lib-duplicate.mjs` correctly found every rewritten import specifier dangling. Fixed in #1440.

`continue-on-error: true` meant nothing ever complained about either.

Both are fixed and d9f0b495 made `--help` survive the prune, so the job **reported SUCCESS on the very run that fixed it** — checked on CI rather than assumed, which is the whole lesson of the two failed attempts above. A green job that cannot fail anything protects nothing.

## The comment was stale too

It pointed at `fix/cli-express-optional-crash` as unmerged and blocking. That landed as d9f0b495. The replacement records *why* the job was red, rather than implying it was ever a known-good check that merely tolerated one known bug.

`@aws-sdk/client-bedrock-runtime` is still statically imported from an optionalDependency (`amazonBedrock/client.ts`, `amazonBedrock/loopAdapter.ts`), but it is not reachable from `--help`, `--version` or `generate --help` — which is why those three pass today. If a later change pulls it onto the entrypoint path this job goes red, and that is the job working.

## Scope

Removing `continue-on-error` makes the job fail the workflow run. Adding it to the branch-protection **required-status-checks** list is a repo setting rather than a file, so that half is left to a maintainer — say the word and I'll raise it separately.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -202,22 +202,26 @@ jobs:
   # failed native build or a --no-optional install, so "it built" here proves
   # nothing about whether the CLI still runs without them.
   #
-  # continue-on-error: true — deliberately non-blocking on day one. Today
-  # `voiceServerApp.js` statically imports `express` (an optionalDependency)
-  # at module load, so `node dist/cli/index.js --help` crashes with
-  # ERR_MODULE_NOT_FOUND as soon as that module is reachable from the CLI
-  # entrypoint. The fix lives on fix/cli-express-optional-crash, not yet
-  # merged to release. Promote this job to a required check (add it to the
-  # required-status-checks list and drop continue-on-error) once that branch
-  # lands. Note @aws-sdk/client-bedrock-runtime is also statically imported
-  # from an optionalDependency (amazonBedrock/client.ts,
-  # amazonBedrock/loopAdapter.ts), so this job may keep surfacing a second,
-  # separate failure even after the express fix — that is expected and is a
-  # second bug for a follow-up branch, not a reason to loosen this check.
+  # This job is now BLOCKING. It spent its whole life non-blocking and red —
+  # first because it installed --no-optional before building (so rolldown's
+  # own native binding vanished and the build tool disappeared), then because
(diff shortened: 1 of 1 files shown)


Tara-ag:
## Review Summary

**Decision:** APPROVED

### Findings
No issues found. This is a CI workflow configuration change that improves the reliability of the pipeline by making the optional-dependency smoke test actually blocking instead of non-blocking.

### Impact on existing code
- **Blast radius:** None - this is a CI workflow change with no runtime impact
- **Execution flows:** No code execution flows affected
- **Out-of-diff dependents:** N/A - CI workflows don't have code dependents
- **Architectural hotspots:** Not applicable - this is infrastructure configuration

### Rationale for approval
1. **Well-documented reasoning**: The author clearly explains why the job was never validating anything meaningful (it failed before running its smoke step)
2. **All underlying issues fixed**: References commits ea91c53d and d9f0b495 that resolved previous failures
3. **Verified success**: Confirmed the job passed when properly configured after fixes
4. **Clear scope**: Only removes `continue-on-error` directive; repo-level required-status-checks update mentioned separately
5. **No security or correctness concerns**: Pure CI workflow improvement
6. **Follows project conventions**: CI workflow changes are outside the scope of CLAUDE.md critical rules

### Review notes
This change makes the CI pipeline more trustworthy by ensuring the optional-dependency smoke test actually validates something. Without this change, the job would silently fail during builds that break due to missing optional dependencies, giving false confidence that the build succeeded.

---
*Reviewed following NeuroLink standards in CONTRIBUTING.md and CLAUDE.md*

Tara-ag (approved):

Tara-ag (approved):

PR #1444, opened 2026-08-21
fix(googleAiStudio): restore the per-turn tool dedup

murdore:
## The regression

My own Task 9 migration dropped `DedupExecuteMap` from Google AI Studio, on **both** the stream and generate paths. Symbol counts across the two commits that did it:

```
509d090b (stream)    DedupExecuteMap 3 → 2    executeMap 6 → 3
ee9cc2a7 (generate)  DedupExecuteMap 2 → 0    executeMap 3 → 0
HEAD                 DedupExecuteMap 0        executeMap 0
```

The migrated code built its `engineTools` record by wrapping the **raw** executors out of `options.tools` instead of going through the dedup map. So BZ-3327 was live again for AI Studio: when Gemini re-emits a tool call with identical arguments across steps — the exact behaviour that class exists to absorb — the tool **re-executed**. Duplicate side effects, duplicate user-visible reports, wasted tokens.

## Proven before fixed

Two failing cases first, one per path. Both reported `dispatched=5` where `1` is correct:

```
[diagnostic] aistudio dedup stream:   dispatched=5 responses=4   ← before
[diagnostic] aistudio dedup generate: dispatched=5 responses=4   ← before
[diagnostic] aistudio dedup stream:   dispatched=1 responses=4   ← after
[diagnostic] aistudio dedup generate: dispatched=1 responses=4   ← after
```

Both halves of each assertion matter: the dispatch count alone would also read `1` if the loop had simply stopped after the first step, so the response count pins that the turn kept running on cached results.

## The fix is a re-wiring, not a reintroduction

`buildNativeToolDeclarations` already builds and returns a `DedupExecuteMap` — the migration just stopped consulting it. New shared helper `buildDedupedEngineTools()` routes `engineTools` back through `executeMap.get()`, which is the call that returns the dedup wrapper.

One wiring detail carries the whole thing: `executeMap` is keyed by the **sanitized** wire name, while the engine looks tools up by the **original** name the adapter reports. `originalNameMap` is the bridge, and it holds an entry for every converted tool (identity mappings included), so iterating it yields exactly the declared, executable set.

The helper cannot remove a callable tool. A tool that threw during conversion is absent from `originalNameMap`, but it also has no declaration, so the model cannot call it — and both call sites still pass `liveTools: options.tools ?? {}`, so `resolveToolOnMiss` covers it regardless.

## How it was missed, and what changed as a result

I audited this migration when I made it by grepping four hand-picked symbols (`extractToolFailureText`, `failedTools.delete`, `toolExecTimeoutMs`, `raceWithAbort`), got `0/0/0` on all four, and recorded "the migration dropped nothing". I never grepped the symbol that mattered. A four-symbol grep only ever proves something about those four symbols.

The replacement is a **set difference**: every identifier the old implementation referenced, minus every identifier the new one references. Run against these commits it prints `GONE: DedupExecuteMap` immediately. It also surfaced a second, unrelated loss on the Anthropic path, which is being fixed separately rather than smuggled in here.

## Scope

Deliberately excluded: I also wired the adapter's `resolveToolOnMiss` through the same map and then reverted it. No test in this suite can reach that path — AI Studio builds `liveTools` and its declarations from the same `options.tools`, so a declared tool never misses — and shipping an unverified change inside a regression fix makes the review harder, not easier. It belongs with the Vertex migration, where `search_tools` discovery makes it genuinely reachable and testable.

## Verification

- AI Studio characterization: **8/8** (was 6/8 with the two new cases failing)
- Vertex characterization: **10/10**, unaffected
- `test:providers-mocked` and `test:provider-structure`: pass — the pair CI actually gates
- typecheck 0 errors / 4830 files; eslint clean on every touched file (only pre-existing `max-lines-per-function` warnings)

### DIFF
--- a/src/lib/providers/googleAiStudio/client.ts
+++ b/src/lib/providers/googleAiStudio/client.ts
@@ -71,6 +71,7 @@ import { estimateTokens } from "../../utils/tokenEstimation.js";
 import { transformToolExecutions } from "../../utils/transformationUtils.js";
 import { resolveToolExecutionRecords } from "../../core/toolExecutionRecorder.js";
 import {
+  buildDedupedEngineTools,
   buildGeminiResponseSchema,
   buildNativeConfig,
   computeMaxSteps,
@@ -1161,27 +1162,13 @@ export class GoogleAIStudioProvider extends BaseProvider {
                 },
               };
 
-              const engineTools: Record<
-                string,
-                {
-                  execute: (
-                    args: Record<string, unknown>,
-                    opts: unknown,
-                  ) => Promise<unknown>;
-                }
-              > = {};
-              for (const [name, tool] of Object.entries(options.tools ?? {})) {
-                const execute = tool?.execute;
-                if (!execute) {
-                  continue;
-                }
-                engineTools[name] = {
-                  execute: async (
-                    args: Record<string, unknown>,
-                    opts: unknown,
-                  ) => execute(args, opts as Parameters<typeof execute>[1]),
-                };
-              }
+              // Through the turn's DedupExecuteMap, NOT the raw executors:
+              // `.get()` returns the dedup wrapper that answers an identical
+              // repeated {name, args} from the per-turn cache (BZ-3327).
+              const engineTools = buildDedupedEngineTools(
+                declarationsResult,
+                options.tools,
+              );
 
               const { stream: engineStream, resultPromise } = runAgenticLoop(
                 adapter,
@@ -1624,25 +1611,11 @@ export class GoogleAIStudioProvider extends BaseProvider {
             },
           };
 
-          const engineTools: Record<
-            string,
-            {
-              execute: (
-                args: Record<string, unknown>,
-                opts: unknown,
-              ) => Promise<unknown>;
-            }
-          > = {};
-          for (const [name, tool] of Object.entries(options.tools ?? {})) {
-            const execute = tool?.execute;
-            if (!execute) {
-              continue;
-            }
-            engineTools[name] = {
-              execute: async (args: Record<string, unknown>, opts: unknown) =>
-                execute(args, opts as Parameters<typeof execute>[1]),
-            };
-          }
+          // Same dedup routing as the streaming twin above.
+          const engineTools = buildDedupedEngineTools(
+            declarationsResult,
+            options.tools,
+          );
 
           const { stream: engineStream, resultPromise } = runAgenticLoop(
             adapter,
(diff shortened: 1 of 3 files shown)


Tara-ag:
💡 MINOR: New test suite added for AI Studio loop characterization — A new end-to-end test suite has been added to characterize the Google AI Studio native-loop behavior, specifically testing the tool deduplication fix for BZ-3327. The test at line 561 'the streaming loop runs an identical repeated call only once' correctly pins the expected behavior that identical tool calls within a turn should be deduplicated to prevent duplicate side effects. This is an excellent addition as it ensures the regression doesn't reappear in future migrations.

Tara-ag:
## Review Summary for PR #1444: fix(googleAiStudio): restore the per-turn tool dedup

### Decision: **APPROVED** ✅

This is a well-targeted bugfix that restores critical functionality to Google AI Studio provider.

### Findings (1 accepted):
- **💡 MINOR** — New test suite added for AI Studio loop characterization (line 561) - Excellent addition that ensures regression coverage for BZ-3327

### Analysis:

**What was fixed:**
- The `DedupExecuteMap` was accidentally removed from Google AI Studio in both streaming and generate paths during a previous migration (Task 9)
- This caused identical tool calls within a single turn to execute multiple times instead of being deduplicated, leading to unwanted side effects
- The fix restores `buildDedupedEngineTools` import and usage from `googleNativeGemini3/utils.js`

**Files changed:**
1. `src/lib/providers/googleAiStudio/client.ts` (+185 / -40 lines) - Restored tool deduplication
2. `test/continuous-test-suite-aistudio-loop-characterization.ts` (+185 lines) - Added verification test

**Impact assessment:**
- This is a **bugfix only** - no breaking changes to public API
- No impact on other providers (Google Native Gemini 3 still has its own implementation)
- The change is self-contained to Google AI Studio path
- CI checks pass (previously verified in PR description)

**Architecture alignment:**
- ✅ Follows existing pattern - imports `buildDedupedEngineTools` from shared `googleNativeGemini3/utils.js` 
- ✅ No violation of CLAUDE.md Critical Rules
- ✅ Maintains backward compatibility
- ✅ Tests are end-to-end focused on public surface (`NeuroLink().generate()`)

### Impact on existing code:
- **Blast radius**: Minimal - only affects Google AI Studio provider's tool execution
- **Execution flows**: Tool execution flow in Google AI Studio now properly deduplicates identical calls
- **Downstream consumers**: None affected - this is an internal implementation detail of the provider

### Conclusion:
This PR successfully fixes a regression where repeated identical tool calls were not being deduplicated in Google AI Studio. The fix is minimal, targeted, and includes proper test coverage. The change follows the existing architectural patterns and doesn't introduce any risks.

**Approved for merge.** 🎉

Tara-ag (approved):

Tara-ag (approved):

PR #1445, opened 2026-08-21
fix(core): thread the caller's span through the agentic loop

murdore:
## The regression

Before the native Anthropic stream turn moved onto the shared loop engine, it called:

```ts
withProviderRetry(fn, trace.getActiveSpan() ?? undefined, `${this.providerName} stream`)
```

`withProviderRetry` writes `gen_ai.provider.total_attempts` on that span after every completed attempt — retried or not. `runAgenticLoop` passed `undefined` in its place, so the attribute silently stopped being emitted for every native Anthropic turn.

The engine carried a comment justifying it:

> no OTel span threaded through the engine today; adapters instrument their own steps if they need span-level detail

I wrote that comment. It is a rationalization written after the fact, not a decision checked against what the old code did. No adapter instruments anything — the attribute just stopped appearing.

## The fix

`AgenticLoopOptions` gains `span?: Span`; the engine passes it to `withProviderRetry`; the Anthropic caller passes `trace.getActiveSpan()` exactly as it used to.

**Caller-supplied, not read from ambient context inside the engine.** Calling `trace.getActiveSpan()` in the engine would be one line shorter and would hand the attribute to *every* provider on the engine — including Google AI Studio, whose hand-rolled loops never emitted it (confirmed: `getActiveSpan` is absent from its pre-migration symbol set). A refactor that silently **adds** observable behaviour is the same defect as one that silently drops it, so this is opt-in and only the Anthropic loops set it.

Only two call sites thread a span into `withProviderRetry` at all — `openaiChatCompletionsBase.ts:1339`, which never moved onto the engine, and this one. No other provider is affected.

## Proven both ways, on a real build

```
without the fix   [diagnostic] anthropic retry telemetry: toolCalls=1 spansWithAttr=0 values=
                  ✗ a native loop turn records the provider attempt count on the active span
                  Passed: 5, Failed: 1

with the fix      [diagnostic] anthropic retry telemetry: toolCalls=1 spansWithAttr=1 values=1
                  Passed: 6
```

The new case asserts the attribute's **value** is `1`, not merely that the key exists, so it cannot pass on an attribute written by some unrelated code path. It also asserts the tool actually ran, so a turn that never reached the agentic loop fails rather than vacuously passing.

The case lives in the existing Anthropic characterization suite, which already has the HTTP stand-in and SSE framing. An `InMemorySpanExporter` is registered above the `dist/` import so NeuroLink's tracers bind to it; for the other five cases it is an inert extra span processor, and all five still pass.

## How it was found

Not by noticing the missing attribute — nothing asserts on it, which is exactly why it went unnoticed. It came out of a **set-difference audit** of the migration commit: every identifier the old implementation referenced, minus every identifier the new one references. Eleven symbols vanished; eight were noise or legitimately relocated (`stringifyFinalResultInput` → `loopAdapter.ts:337`, `resolveDeferredTool` → `loopAdapter.ts:88`, `withProviderRetry` → the engine), and `getActiveSpan` was the one that had simply been dropped.

The same audit run against the Google AI Studio migration surfaced a separate, larger regression, fixed in #1444.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of retry-attempt tracking during provider interactions.
  - Anthropic streaming requests now correctly record retry information in telemetry.

- **Observability**
  - Added per-step provider attempt details to OpenTelemetry traces.
  - Enhanced tracing coverage for tool-enabled Anthropic conversations.

### DIFF
--- a/src/lib/core/loopEngine.ts
+++ b/src/lib/core/loopEngine.ts
@@ -141,7 +141,10 @@ export function runAgenticLoop<TConversation>(
                 throw hasEmitted ? new PostEmissionStepError(err) : err;
               }
             },
-            undefined, // no OTel span threaded through the engine today; adapters instrument their own steps if they need span-level detail
+            // The caller's span, when it passes one. withProviderRetry writes
+            // gen_ai.provider.total_attempts here, so a loop that threaded a
+            // span before it moved onto this engine keeps emitting it.
+            options.span,
             `${adapter.providerLabel}.step`,
           );
         } catch (err) {
--- a/src/lib/providers/anthropic/client.ts
+++ b/src/lib/providers/anthropic/client.ts
@@ -2194,12 +2194,20 @@ export class AnthropicProvider extends BaseProvider {
         };
       }
 
+      // The active span goes with it. Before this loop moved onto the shared
+      // engine it called
+      //   withProviderRetry(fn, trace.getActiveSpan() ?? undefined, label)
+      // and that span is where gen_ai.provider.total_attempts is recorded.
+      // The engine passed `undefined` in its place, so the attribute silently
+      // stopped being emitted for every native Anthropic turn.
+      const activeSpan = trace.getActiveSpan();
       const { stream, resultPromise } = runAgenticLoop(
         adapter,
         payload.messages.slice(),
         {
           tools: engineTools,
           ...(abortSignal ? { abortSignal } : {}),
+          ...(activeSpan ? { span: activeSpan } : {}),
         },
       );
 
--- a/src/lib/types/loopEngine.ts
+++ b/src/lib/types/loopEngine.ts
@@ -1,4 +1,5 @@
 import type Anthropic from "@anthropic-ai/sdk";
+import type { Span } from "@opentelemetry/api";
 import type { Tool } from "./tools.js";
 import type {
   CollectedChunkResult,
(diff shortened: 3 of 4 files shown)


Tara-ag:
## Review Summary

**Decision**: ✅ **APPROVED**

### Changes Reviewed (4 files):

1. **`src/lib/types/loopEngine.ts`**
   - Added optional `span?: Span` field to `AgenticLoopOptions` type
   - Non-breaking change, follows CLAUDE.md rules (uses `type`, not `interface`)

2. **`src/lib/core/loopEngine.ts`**
   - Modified line 144 to pass `options.span` to `withProviderRetry`
   - Enables retry telemetry to be recorded on the caller's span

3. **`src/lib/providers/anthropic/client.ts`**
   - Extracts active span from OpenTelemetry trace context
   - Passes it through to `runAgenticLoop`
   - Fixes regression where provider attempts weren't being recorded

4. **`test/continuous-test-suite-anthropic-loop-characterization.ts`**
   - Adds E2E test verifying `gen_ai.provider.total_attempts` attribute is recorded
   - Follows Rule 15 (tests import from `dist/`, exercise public surface)

### Impact Analysis:
- **Blast radius**: 500 nodes impacted (from graph analysis)
- **Affected callers**: `amazonBedrock/client.ts`, `googleAiStudio/client.ts`, existing loop engine tests
- **Backward compatibility**: ✅ Maintained - all changes are additive/optional parameters
- **Breaking changes**: None detected

### CLAUDE.md Compliance:
- ✅ Uses `type` not `interface`
- ✅ Types in canonical location (`src/lib/types/`)
- ✅ Optional parameter = non-breaking
- ✅ E2E tests follow Rule 15
- ✅ No injection vulnerabilities

The fix correctly addresses the regression where Anthropic native stream turns weren't recording retry information in telemetry. The implementation properly threads the caller's span through the agentic loop so that `withProviderRetry` can annotate it with `gen_ai.provider.total_attempts`.

**No inline comments needed - all changes verified as correct.**

Tara-ag (approved):

Tara-ag (approved):

PR #1447, opened 2026-08-21
test(anthropic): pin which span carries the retry attempt count

murdore:
Follow-up to #1445, acting on a review note left on it.

## The valid half of the note

The assertion accepted **any** finished span carrying `gen_ai.provider.total_attempts`, so an attribute written by an unrelated provider path would have satisfied it. That is a fair criticism and this tightens it.

## The suggested fix does not work, and why is worth recording

The note proposed wrapping `nl.stream()` in the test's own `startActiveSpan` and asserting on that span. That would fail.

The attribute lands on **`neurolink.stream`** — the span the SDK opens around the request, and what `trace.getActiveSpan()` resolves to when the Anthropic client hands its span to the loop. A test-created wrapper sits *above* that span, so the attribute goes to the child and the wrapper stays bare.

Established by printing the carrying span's name rather than assumed in either direction:

```
[diagnostic] anthropic retry telemetry: toolCalls=1 spansWithAttr=1 names=neurolink.stream values=1
```

## What this does instead

Pins the span by name, which buys the same tightness without the incorrect mechanism. Mutating the expected name to anything else fails exactly this case and no other:

```
name → "some.other.span"   →  Passed: 5, Failed: 1
```

The three assertions now cover identity, value, and that the tool actually ran — so the case cannot pass on a span that isn't the turn's own, on a wrong attempt count, or on a turn that never reached the agentic loop.

## Summary by CodeRabbit

* **Tests**
  * Enhanced retry telemetry validation to report the spans associated with provider attempt counts.
  * Added checks confirming these telemetry spans use the expected naming convention.

PR #1451, opened 2026-08-21
fix(deps): raise the fast-uri override past the 3.x advisories

murdore:
## Release is red for every open PR

`validate:security` fails on `release` itself:

```
[SECURITY] FAIL dependencies: Unaccepted high advisory 1145636 in fast-uri:
           host confusion via failed IDN canonicalization
```

Nothing merged caused it — it is a newly published advisory against a transitive dependency. Three now cover fast-uri 3.x:

| advisory | needs |
|---|---|
| 1145636 — host confusion via failed IDN canonicalization | `>=3.1.3` |
| 1124064 — host confusion via literal backslash authority delimiter | `>=3.1.4` |
| 1130720 — host confusion via backslash authority introducer | `>=3.1.5` |

The existing override `"fast-uri@<3.1.2": ">=3.1.2"` no longer matches anything, since everything already resolved to exactly 3.1.2.

## The `<4` bound is the part that matters

Written as a bare `">=3.1.5"`, pnpm resolved every dependent to **4.1.2** — a major bump — because the range permits it. `ajv` declares `fast-uri: ^3.0.1`, so that override silently forces ajv onto a major version it never claimed to support.

**Security validation passed either way.** Only reading the resulting lockfile caught it. Bounded to `">=3.1.5 <4"` it resolves to 3.1.5 for ajv's dependents and leaves 4.1.2 for the packages that legitimately depend on 4.x.

## Verified after the change, not inferred

```
validate:security       PASS
validate:all            PASS
typecheck               0 errors / 4830 files
build                   PASS
test:providers-mocked   54 passed · 0 failed
```

## How it surfaced

Three unrelated PRs — #1446, #1447, #1448 — failed the same two checks at once, and #1448 only adds a script file. That combination is a repo-level signal rather than three coincidences, so I reproduced it on `release` directly before touching anything.

Worth merging ahead of the queue: nothing in the repo can go green until it lands.

## Summary by CodeRabbit

* **Chores**
  * Updated the supported `fast-uri` version range to require version 3.1.5 or later while remaining below version 4.


Tara-ag (approved):

Tara-ag:

This is a well-reasoned dependency update that addresses newly published security advisories for `fast-uri`.

### Summary of Changes
- **Updated pnpm override** for `fast-uri` from `<3.1.2` to `>=3.1.5 <4` in both `package.json` and `pnpm-lock.yaml`
- **Addresses 3 npm advisories** (1145636, 1124064, 1130720) related to host confusion via failed IDN canonicalization
- **Critical bound**: The `<4` constraint prevents breaking ajv's dependency on fast-uri ^3.0.1

### Why This is Safe
1. **No source code changes** - only package manager configuration
2. **Security validation passes** - author verified all checks pass before and after
3. **Well-documented rationale** - explains why the old override no longer matched anything and why the `<4` bound matters
4. **Proactive fix** - multiple unrelated PRs were failing due to this issue; fixing it unblocks the queue
5. **No API impact** - purely internal dependency management

### Verification
The PR correctly notes that the existing override `fast-uri@<3.1.2` had become ineffective since everything resolved to exactly 3.1.2. Updating to `>=3.1.5 <4` both patches the vulnerabilities AND keeps ajv's dependents on the correct major version.

**Impact on existing code**: None - this is a transitive dependency update that does not affect the public SDK API or any application code.

---
*Yama Code Review Agent - PR #1451 reviewed*

Tara-ag (approved):

PR #1452, opened 2026-08-22
fix(scripts): parse the symbol diff instead of pattern-matching it

murdore:
Acts on the three review findings left on #1448. All three were valid, and a colleague confirmed them.

Worth stating plainly: a tool built to catch **silent behaviour loss** had three ways of **silently reporting success**.

## 1. The regex missed things — worse than the noise it made

It required four or more characters to suppress some of its own false positives, so a dropped `map()`, `get()` or `log()` — exactly the one-line side effect this exists to catch — was invisible. It also could not distinguish a call from a declaration, a comment, or a substring of a string literal, and duly reported `while`, `throw` and assorted prose as vanished symbols.

Now parsed with the `typescript` dependency the repo already has, counting only `CallExpression` and `NewExpression` callees. Property calls are recorded under both the bare name and the dotted path, so `getActiveSpan` now also surfaces as `trace.getActiveSpan`.

## 2. Every `git show` failure was read as absence

A bad repository, a permission error, or output past `maxBuffer` all printed a reassuring *"absent on one side"* for a comparison that never ran. Only a genuine missing **path** counts as absence now; everything else throws.

A bad **revision** throws too. That is a typo in the invocation, not a fact about the tree, and answering it with "absent on one side" responds to a question the caller never asked while looking like a clean result.

## 3. `nothing vanished` was printed when nothing was compared

If every path was absent on one side, no comparison happened, `anyVanished` stayed `false`, and the run ended on a clean bill of health it had not earned — the same defect this tool looks for, one level up. It now distinguishes the two:

```
no comparisons ran — every path was absent on one side
```

## Two defects in my own fix, found by running it

- `getText()` on a property access rooted in a call expression emitted a **thirty-line** `withTimeout(...).catch` as a symbol name. The dotted path is now built by walking identifiers and skipped the moment the chain roots in anything that isn't a plain name.
- The first cut of the absence classifier still swallowed bad revisions.

## Verification

Against the two regressions it was written for:

```
ee9cc2a7 googleAiStudio/client.ts   →  GONE: DedupExecuteMap
ff6fc698 anthropic/client.ts        →  GONE: getActiveSpan
                                       GONE: trace.getActiveSpan
```

Bad revision now exits 1 with git's own diagnosis and no Node stack on top of it.

**No test file.** There is no unit-test runner in this repo, and `test/` is end-to-end only by CLAUDE.md rule 15 — a suite asserting on a dev script's internals would be exactly the kind of test that rule excludes. A script is verified by running it against commits whose answers are already known, which is what the above is.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved migration symbol comparison accuracy by recognizing direct and property-based calls.
  * Added clearer handling for missing migration paths and unexpected tool errors.
  * Reports when no comparisons can be performed and exits cleanly when failures occur.

### DIFF
--- a/scripts/migration-symbol-diff.mjs
+++ b/scripts/migration-symbol-diff.mjs
@@ -5,7 +5,7 @@
  * Written for reviewing refactors that MOVE code — a loop extracted into a
  * shared engine, a helper lifted into a module. The risk in that shape of
  * change is not a broken build; it is a behaviour that quietly stops
- * happening, because the call that performed it was dropped rather than
+ * happening, because the call performing it was dropped rather than
  * relocated. Those do not fail a typecheck and they do not fail a test that
  * nobody wrote.
  *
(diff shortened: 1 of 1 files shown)


Tara-ag:
## Review Summary

This PR fixes three bugs in `migration-symbol-diff.mjs`, a build validation script used to catch silent behavior loss during refactors:

### Changes Made

1. **AST-based symbol detection** (lines 37-106): Replaced regex pattern matching with TypeScript AST parsing using the `typescript` dependency. This correctly distinguishes:
   - Actual calls from declarations
   - Calls in comments vs code
   - Substrings in string literals
   - Property access chains like `trace.getActiveSpan()`

2. **Better error handling for git failures** (lines 122-140): Added `isAbsentPath()` function to distinguish between:
   - Genuine missing paths ("does not exist in tree")
   - Tool failures (bad revisions, permission errors, maxBuffer exceeded)

3. **Track comparison count** (line 151): Added `anyCompared` flag to detect when all paths are absent on one side, preventing false "nothing vanished" messages.

### Impact Analysis

- **Changed file**: `scripts/migration-symbol-diff.mjs` (single file)
- **Impact radius**: 8 nodes directly changed, 500+ impacted (build validations call this script)
- **Risk**: Low - this is a meta-tool that helps reviewers; it doesn't modify runtime behavior
- **Backward compatibility**: N/A - this is a development/validation script

### Verification

The PR author notes two real regressions caught by the old version:
1. Google AI Studio: lost `DedupExecuteMap` call → duplicate tool execution
2. Anthropic: lost `trace.getActiveSpan()` call → missing telemetry

The new AST-based approach will reliably catch these and similar cases.

### Decision

**APPROVED** - This is a solid improvement that makes the behavior-loss detection tool more accurate and robust. No blocking issues found.

Tara-ag (approved):

Tara-ag (approved):

PR #1453, opened 2026-08-22
fix(proxy): answer Codex model discovery instead of 404ing it

### LINKED ISSUE #1383: Codex CLI model discovery 404s: proxy implements /responses but not /backend-api/codex/models
Found by running the actual Codex CLI through the proxy, not by reading code.

## Symptom

Every `codex` invocation through the proxy prints this **twice** before doing anything:

```
ERROR codex_models_manager::manager: failed to refresh available models:
  unexpected status 404 Not Found, url:
  http://127.0.0.1:55669/backend-api/codex/models?client_version=0.147.0
```

## Cause

`createCodexProxyRoutes` defines exactly one route:

```
src/lib/server/routes/codexProxyRoutes.ts:615
  path: `${basePath}/backend-api/codex/responses`
```

Codex CLI calls `/backend-api/codex/models` on startup to refresh its model list. The proxy has no such route, so it 404s.

Confirmed live:

```
/backend-api/codex/models    -> HTTP 404
/backend-api/codex/responses -> HTTP 400  (400 = route exists, empty body rejected upstream)
```

## Impact

Chat works — the CLI falls back to a default model — so this is not fatal. But:

1. Two red `ERROR` lines on every single invocation, which trains users to ignore proxy errors.
2. **Model discovery is silently broken.** The CLI cannot see which models are available and picks a fallback. This is the class of defect that gets diagnosed months later as "Codex sometimes picks the wrong model", because nothing connects the symptom to a 404 at startup.

## Fix

Proxy `GET /backend-api/codex/models` upstream to `chatgpt.com/backend-api/codex/models` using the same pooled-account auth the `/responses` route uses, and relay the response. The account selection and token refresh are already in that file.

murdore:
Closes #1383.

## Before

The proxy registered exactly one Codex route. Against the running daemon:

```
GET  /backend-api/codex/models?client_version=0.147.0 -> 404
POST /backend-api/codex/responses                     -> 400   (route exists, empty body)
```

Every `codex` invocation printed `failed to refresh available models: unexpected status 404 Not Found` and then silently fell back to a default model — quietly ignoring whichever model the user configured. The error line scrolls past; the wrong model gets misdiagnosed weeks later.

## After

Against a proxy started from the built CLI, answering the CLI's exact request:

```
GET /backend-api/codex/models?client_version=0.147.0 -> 200
{"models":[{"slug":"gpt-5.6-terra","prefer_websockets":true,...
```

A real model list, relayed from ChatGPT upstream through the account pool.

## Why relay, not synthesise

The Claude and OpenAI `/v1/models` routes build their lists locally from the model router. Codex model availability is a property of the *account* — plan tier, rollout state — which this proxy does not know. A locally-built list would be a guess that reads as authoritative, so this one relays.

Discovery is side-effect free: no cooldown recorded, no quota consumed, so the once-per-invocation refresh cannot perturb routing for real traffic. A cooling account may still answer it — being rate-limited for completions does not make an account unable to say which models exist.

## A defect the end-to-end test caught

The first implementation read the query string off `ctx.path`. `ctx.path` carries no query string, so `client_version` was dropped and upstream answered `400` with a pydantic `Field required` on `('query','client_version')`. Rebuilt from `ctx.query`.

No unit test would have found that — it only surfaces when a real request reaches the real upstream. The regression test therefore drives a **spawned proxy** the way the CLI drives it, and asserts against `404` specifically, so a route that exists but errors still passes while an unroutable one fails.

Worth noting for anyone iterating here: `pnpm run build:cli` does **not** rebuild `src/lib`, so a route added to `codexProxyRoutes.ts` stays invisible until a full `pnpm run build`. That cost me a debugging cycle.

## Verification

| Check | Result |
| --- | --- |
| `pnpm run build` | clean |
| `continuous-test-suite-proxy` | **65 passed** (was 64) |
| Live relay to ChatGPT upstream | 200 + real model list |

Test watched failing first: `Codex model discovery is unroutable — the CLI cannot list models`.

## Summary by CodeRabbit

* **New Features**
  * Added Codex model discovery through the proxy.
  * Supports forwarding model requests with client version information and returning available models.
  * Discovery requests do not consume quota or affect account cooldowns.

* **Bug Fixes**
  * Prevented Codex clients from receiving a 404 and incorrectly falling back to the default model.

### DIFF
--- a/docs/features/codex-proxy-support.md
+++ b/docs/features/codex-proxy-support.md
@@ -155,3 +155,36 @@ The Anthropic engine in `claudeProxyRoutes.ts` is untouched. The Codex engine is
 - **Quota-aware ordering is unverified against the live backend.** The usage endpoint and the rate-limit header shape were reconstructed from a capture, not confirmed end to end. If either is wrong, `fetchCodexAccountUsage` returns no quota, `x-neurolink-quota-source` reads `none`, and ordering degenerates to insertion order while every 429 falls back to the 15-minute transient cooldown. Rotation still works; it is simply not quota-aware. Verify with `neurolink auth list --refresh` — a `codex usage …` error line per account means the quota path is not live.
 - **SSE usage-limit signals are not acted on.** Only an HTTP 429 triggers a cooldown and rotation. A `200` response whose SSE stream carries `usage_limit_reached` (or the workspace-credit variants) is relayed to the client untouched, so the account is neither cooled nor rotated away from. HTTP-level exhaustion is handled; in-stream exhaustion is not.
 - **Client fingerprint is forwarded verbatim.** The proxy replaces the caller's credentials but does not regenerate `originator`, `installation_id`, `session-id`, or turn metadata per account, so every pooled account shares the client's fingerprint. This is what makes the terms-of-service point above concrete.
+
+## Model discovery
+
+The Codex CLI refreshes its model list on every invocation:
+
+```
+GET /backend-api/codex/models?client_version=<cli-version>
+```
+
+The proxy relays that upstream to `chatgpt.com/backend-api/codex/models` using a
+pooled account, forwarding the CLI's own query parameters.
+
+**It relays rather than synthesises**, unlike the Claude and OpenAI `/v1/models`
+routes, which build their lists locally from the model router. Which Codex
+models an account can reach is a property of that account — plan tier, rollout
+state — not something this proxy knows, so a locally-built list would be a guess
+that reads as authoritative.
+
+Two details matter to anyone touching it:
+
+- **`client_version` is required upstream.** Omit it and ChatGPT answers `400`
+  with a pydantic `Field required` on `('query', 'client_version')`. The query
+  is rebuilt from `ctx.query`; `ctx.path` carries no query string, and reading it
+  from there drops the parameter silently.
+- **Discovery is side-effect free.** No cooldown is recorded and no quota is
+  consumed, so the once-per-invocation refresh cannot influence routing for real
+  traffic. A cooling account is still allowed to answer it — being rate-limited
+  for completions does not make an account unable to say which models exist.
+
+Before this route existed the request 404'd, and the CLI printed
+`failed to refresh available models: unexpected status 404 Not Found` on every
+run before silently falling back to a default model — quietly ignoring the model
+the user had configured.
--- a/src/lib/auth/codexOAuth.ts
+++ b/src/lib/auth/codexOAuth.ts
@@ -37,6 +37,7 @@ export const CODEX_DEFAULT_SCOPES = [
 export const CODEX_BACKEND_BASE_URL = "https://chatgpt.com/backend-api/codex";
 export const CODEX_RESPONSES_URL = `${CODEX_BACKEND_BASE_URL}/responses`;
 export const CODEX_USAGE_URL = `${CODEX_BACKEND_BASE_URL}/usage`;
+export const CODEX_MODELS_URL = `${CODEX_BACKEND_BASE_URL}/models`;
 
 // Client fingerprint. The real Codex CLI sends these; the proxy defaults them
 // only when a non-Codex client omits them.
--- a/src/lib/server/routes/codexProxyRoutes.ts
+++ b/src/lib/server/routes/codexProxyRoutes.ts
@@ -21,6 +21,7 @@
 import { tokenStore } from "../../auth/tokenStore.js";
 import {
   CODEX_ORIGINATOR,
+  CODEX_MODELS_URL,
   CODEX_RESPONSES_URL,
   CODEX_USER_AGENT,
   codexTokenNeedsRefresh,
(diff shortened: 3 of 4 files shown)


murdore:
All three findings were real and are fixed in `de755f89`.

**1. Unbounded model-discovery fetch — fixed as suggested.** Added `AbortSignal.timeout(CODEX_UPSTREAM_TIMEOUT_MS)`, the same bound the responses route uses. This one mattered more than it looks: model discovery runs at CLI startup, so a stalled upstream blocked the CLI before it could do anything.

**2. No refresh/rotation on 401/403 — fixed, but deliberately not a copy of the responses flow.** Discovery now forces one token refresh per account and rotates through the pool, which fixes the reported problem: a token rejected upstream while still inside its local expiry window left discovery broken until it expired locally.

What I did **not** carry over is the cooldown and the account-disable. Those belong to the responses path because the verdict is earned by a request the user actually made. Discovery fires on *every* CLI invocation — giving it the power to disable an account would turn a background probe into a forced re-login. So the rotation is there, the penalties are not.

That also made the existing doc comment wrong, so I corrected it rather than leaving it: the route is not "side-effect free" — a refreshed token is persisted, exactly as the proactive refresh in `loadCodexProxyAccounts` persists one. What it never does is *penalise* an account, and that is what the comment now says.

**3. Test accepted HTTP 400 — fixed, and this was the most serious of the three.** The test excluded 404 and 405 and accepted everything else, so it also accepted 400. 400 is precisely what upstream returns when `client_version` is dropped from the forwarded query — the bug this route was written to fix. The test passed on the regression it existed to catch.

The tolerated set is now an allow-list (`200, 401, 502, 503`) rather than a deny-list, with 400 called out by name so the failure message points at the cause instead of just the status.

**Verification:** typecheck 4830 files / 0 errors, eslint clean, proxy suite 65 passed / 0 failed / 6 skipped. I also mutated the new assertion — removing 401 from the allow-list — and confirmed the case reports a real `✗` failure rather than being downgraded to a skip, so the assertion is genuinely load-bearing.

Tara-ag:
## Review Summary

### Decision: **APPROVED** ✅

This PR adds model discovery support for the Codex (ChatGPT) proxy, completing the missing `/backend-api/codex/models` route that was previously returning 404.

### Changes Reviewed

| File | Type | Lines Changed | Status |
|------|------|---------------|--------|
| `docs/features/codex-proxy-support.md` | Documentation | +33 | ✅ Complete |
| `src/lib/auth/codexOAuth.ts` | Type definition | +1 | ✅ Correct |
| `src/lib/server/routes/codexProxyRoutes.ts` | Route handler | +140 | ✅ Valid |
| `test/continuous-test-suite-proxy.ts` | Tests | +71 | ✅ Comprehensive |

### Review Details

#### ✅ Documentation (codex-proxy-support.md)
- Accurately describes model discovery behavior
- Explains why relay (not synthesis) is required
- Documents the `client_version` parameter requirement
- Clarifies side-effect-free nature of discovery calls

#### ✅ Type Definition (codexOAuth.ts)
- Added `CODEX_MODELS_URL` constant following existing patterns
- Consistent with `CODEX_RESPONSES_URL` and `CODEX_USAGE_URL`

#### ✅ Route Handler (codexProxyRoutes.ts)
- Implements `handleCodexModelsRequest()` following established patterns from responses handler
- Properly handles account ordering and rotation
- Forces token refresh before each request (critical for reliability)
- Retries on 401/403 but relays other statuses unchanged (including 400)
- Uses proper timeout via `AbortSignal.timeout(CODEX_UPSTREAM_TIMEOUT_MS)`
- Returns meaningful error message if all candidates fail

#### ✅ Test Coverage (continuous-test-suite-proxy.ts)
- `testCodexModelsDiscovery()` validates critical behaviors:
  - Endpoint exists (rejects 404)
  - Method allowed (rejects 405)  
  - Query forwarded correctly (rejects 400)
  - Accepts expected statuses: 200, 401, 502, 503
  - Rejects unexpected statuses

### Impact Analysis
- **Affected flows**: 78 execution flows touched
- **Backward compatibility**: No breaking changes - only adds new functionality
- **Risk level**: Low - follows proven patterns from Claude proxy implementation
- **Test coverage**: New test ensures regression prevention

### CLAUDE.md Compliance
✅ Rule 4 (CLI ≠ SDK): Pure server-side code, no SDK concerns  
✅ Rule 5 (Backward compatibility): Only adds new endpoint, no API changes  
✅ Rule 6 (formatProviderError): N/A - not modifying error formatting  

### Recommendation
**APPROVE** - This is a well-implemented feature addition that completes the Codex proxy's model discovery capability. The implementation follows established patterns, includes comprehensive tests, and properly documents the behavior.

---
*Reviewed using Yama autonomous code review agent standards*

Tara-ag (approved):

Tara-ag:
## Review Summary

All previously identified issues have been resolved:

1. ✅ **Timeout added** to upstream model-discovery fetch (lines 650-654) - matches the pattern used in responses route
2. ✅ **Forced token refresh** on auth failures (401/403) - applies same flow as handleCodexResponsesRequest  
3. ✅ **HTTP 400 treated as failure** - test updated to reject 400 responses

The implementation follows Claude proxy patterns with comprehensive tests and documentation. The PR is ready for merge.

Tara-ag (approved):

Tara-ag (approved):

Issue #1368, opened 2026-08-19
Proxy client config writers have no test coverage


murdore:
Found while auditing proxy CLI coverage at v11.2.3 (`8728bbb3`).

## Problem

The three client config writers in `src/cli/commands/proxy.ts` — `setClaudeProxySettings` (`:625`), `setOpenCodeProxySettings` (`:734`), `setCodexProxySettings` (`:922`), and their three paired `clear*` functions — have **zero** automated coverage.

`grep -rn "setOpenCodeProxySettings\|getOpenCodeConfigDir\|clearOpenCodeProxySettings" test/ src/` matches only `proxy.ts` itself. `grep -rli opencode test/` finds only two unused data fixtures (`test/fixtures/opencode-local-proxy.json`, `test/fixtures/opencode-local-proxy-openai-route.json`). `test/continuous-test-suite-proxy.ts` has no occurrence of `opencode`.

## Why this matters — it let two real bugs ship

The macOS path bug and the unconditional success message (filed separately) both live in this untested surface.

`docs/features/opencode-proxy-support.md` §11 presents a detailed end-to-end verification playbook, but it runs the dev proxy with `--dev`, whose own option description is "no client auto-configuration" (`proxy.ts:3910-3917`), and hand-copies a fixture to `/tmp/opencode-proxy-test/opencode.json`. **The writer is never executed by the documented verification.**

## Suggested coverage

Per CLAUDE.md rule 15 these must be end-to-end — drive `node dist/cli/index.js` rather than importing from `src/lib/`. Worth asserting:

- each `set*` writes to the path the target CLI actually reads, with `HOME` pointed at a temp dir
- each `set*` is a no-op **and reports as such** when the target CLI is absent
- each `clear*` restores the pre-existing user config byte-for-byte
- the snapshot survives a simulated crash between set and clear

murdore:
Closing — all five configurators now have apply/restore round-trip coverage in `test/continuous-test-suite-proxy.ts`, driven through the real writers rather than asserting on internals.

The gap this issue named was Codex specifically: the TOML writer, the `model_provider` flip and the crash-safe snapshot restore were untested. That case exists now:

- `Proxy clients: Codex apply/restore round-trips a real config` (`testCodexConfiguratorRoundTrip`) — drives `__codexClientTestHooks.setCodexProxySettings` / `clearCodexProxySettings` against a real `~/.codex/config.toml` in an isolated HOME, then diffs the restored file against the original.

Alongside it:

[...]

PR #1454, opened 2026-08-22
fix(server): let SDK consumers mount the Codex proxy door

murdore:
Two gaps that share a root cause: Codex was treated as CLI-only.

## 1. An SDK consumer could not mount the Codex door at all

`createAllRoutes` assembled two of the three proxy doors. `createCodexProxyRoutes` was neither imported nor re-exported from the routes barrel, and no flag selected it — so Codex proxying was reachable only by running `neurolink proxy start`.

Verified through the published surface, not by reading source:

```
before:  createAllRoutes("/api", { proxy: true })      -> no codex door
after:   createAllRoutes("/api", { proxy: true })      -> codex door present
         createAllRoutes("/api", { codexProxy: true }) -> codex door present
         createAllRoutes("/api", {})                   -> absent (unchanged)
```

The unified `proxy` flag now means *every* door rather than two of three. A consumer opting into proxying is opting into proxying, and the silent omission is exactly what let this sit unnoticed. **Flag me if you'd rather `proxy` stayed Claude+OpenAI** and Codex required its own opt-in — it's a judgement call, and reverting it is one line.

**A wider gap found while wiring it:** the server entry re-exported *no* proxy factory. Claude and OpenAI were reachable only via the deep path `./routes/index.js`, which is not a package export — so `@juspay/neurolink/server` consumers could never mount one directly. All three are exported now.

## 2. Codex had no configurator round-trip test — #1368 was closed as covered

Codex is the writer with the most to get wrong: it edits TOML by regex rather than round-tripping JSON, keeps its snapshot in a sidecar file, and rewrites a top-level selector that must stay in the preamble. A `model_provider` read from inside a `[profiles.*]` table would be written back as the **global** selector, silently repointing every Codex run at another provider.

The test drives the real writer against a config carrying that hazard, in two cases:

- **A user selector present** — restore must return it verbatim. This case passes against a correct *and* a broken writer, because document-wide and preamble-scoped matching agree when the preamble already has one. Kept because it covers the common path, but it proves nothing on its own.
- **No global selector, only a profile's** — this is what separates them. Confirmed by regressing the writer to match document-wide, watching this case fail, then restoring it.

Worth recording: my first version of case B asserted `/^model_provider = /m` on the whole file. That matches the line inside `[profiles.work]` too, so it reported a bug that did not exist. The writer was pristine the whole time; the assertion is now scoped to the preamble.

## Verification

| Check | Result |
| --- | --- |
| `pnpm run build` | clean |
| `continuous-test-suite-proxy` | **65 passed** |
| `continuous-test-suite-servers` | **41 passed** |
| Regression proof | writer broken on purpose -> case B fails; restored -> passes |

Stacks cleanly with #1453 (Codex model discovery); the two touch different files.

### DIFF
--- a/src/lib/server/index.ts
+++ b/src/lib/server/index.ts
@@ -190,6 +190,13 @@ export {
   createOpenApiRoutes,
   createToolRoutes,
   registerAllRoutes,
+  // The proxy doors. All three were reachable only from the deep path
+  // ./routes/index.js, which is not a package export — so a consumer of
+  // "@juspay/neurolink/server" could reach them only indirectly through
+  // createAllRoutes' flags, never to mount one on its own.
+  createClaudeProxyRoutes,
+  createOpenAIProxyRoutes,
+  createCodexProxyRoutes,
 } from "./routes/index.js";
 // ============================================
 // Streaming
--- a/src/lib/server/routes/index.ts
+++ b/src/lib/server/routes/index.ts
@@ -13,6 +13,7 @@ import { createClaudeProxyRoutes } from "./claudeProxyRoutes.js";
 // ClaudeProxyDeps removed
 import { createHealthRoutes } from "./healthRoutes.js";
 import { createOpenAIProxyRoutes } from "./openaiProxyRoutes.js";
+import { createCodexProxyRoutes } from "./codexProxyRoutes.js";
 import { createMCPRoutes } from "./mcpRoutes.js";
 import { createMemoryRoutes } from "./memoryRoutes.js";
 import { createOpenApiRoutes } from "./openApiRoutes.js";
@@ -24,6 +25,7 @@ export { createClaudeProxyRoutes } from "./claudeProxyRoutes.js";
 // ClaudeProxyDeps removed
 export { createHealthRoutes } from "./healthRoutes.js";
 export { createOpenAIProxyRoutes } from "./openaiProxyRoutes.js";
+export { createCodexProxyRoutes } from "./codexProxyRoutes.js";
 export { createMCPRoutes } from "./mcpRoutes.js";
 export { createMemoryRoutes } from "./memoryRoutes.js";
 export { createOpenApiRoutes } from "./openApiRoutes.js";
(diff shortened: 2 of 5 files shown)


Tara-ag:
## Review Summary

**Decision: APPROVED** ✅

### Findings
No issues found. This is a clean feature addition that properly exposes Codex proxy functionality through the SDK.

### Changes Reviewed

**File: `src/lib/server/index.ts`**
- ✅ Added re-exports of `createClaudeProxyRoutes`, `createOpenAIProxyRoutes`, and `createCodexProxyRoutes`
- These proxy route creators were previously only accessible internally, now exposed for SDK consumers
- No breaking changes - purely additive

**File: `src/lib/server/routes/index.ts`**
- ✅ Added import and export for `createCodexProxyRoutes`
- ✅ Updated `createAllRoutes()` to support `codexProxy` flag alongside existing `claudeProxy`/`openaiProxy` flags
- ✅ Unified `proxy` flag now enables all three proxy doors
- Implementation correctly mirrors the Claude/OpenAI pattern

**File: `src/lib/types/server.ts`**
- ✅ Added `codexProxy?: boolean` option to `CreateRoutesOptions` type
- ✅ Documentation explains historical context

**File: `test/continuous-test-suite-proxy.ts`**
- ✅ Excellent round-trip test for Codex configurator (`testCodexConfiguratorRoundTrip`)
- Tests critical hazard: profile-level vs global model_provider selector
- Validates apply/restore behavior with real TOML config

**File: `test/continuous-test-suite-servers.ts`**
- ✅ Added `testCodexProxyReachableFromSDK()` - verifies SDK consumers can mount Codex proxy door
- Tests both dedicated `codexProxy: true` flag and unified `proxy: true` flag
- Verifies Codex is off by default when no flag specified

### Impact on Existing Code
- **Blast radius:** Low - all changes are additive to public API surface
- **Breaking changes:** None
- **Backward compatibility:** Fully maintained - legacy per-format flags still work
- **Dependencies:** No new external dependencies added

### Review Scope
Reviewed 5 changed files covering:
- Public API exports
- Route creation logic  
- Type definitions
- Integration tests
- SDK seam verification

The implementation is solid, well-tested, and follows established patterns in the codebase. No action required beyond merging.

Tara-ag (approved):

Tara-ag (approved):

PR #1455, opened 2026-08-22
fix(proxy): write client configs atomically

### LINKED ISSUE #1384: Client config writers: no locking or atomic writes between apply and restore
Raised by a code review of the client-configurator refactor. All three predate that work; none is a regression.

The writers in `src/cli/proxy-clients/` each do `readFileSync` → mutate in memory → `writeFileSync` on a file that may hold a real API key.

## 1. TOCTOU between apply and restore

No lock, lease, or compare-and-swap. If `apply()` and `restore()` interleave — a rapid `proxy stop` / `proxy start`, or a stale shutdown handler firing after a fresh start — the later write wins and can discard the other`s result. A fresh apply`s proxy block can be undone by a stale restore, or vice versa.

`applyAllClients` / `restoreAllClients` serialise within one process but provide no cross-process locking, and the proxy legitimately runs as a launchd daemon alongside a foreground instance.

## 2. Writes are not atomic

`writeFileSync` writes in place rather than write-to-temp-then-rename. A process killed mid-write (OOM, SIGKILL, power loss) can leave the settings JSON truncated — taking the `__proxy_original_*` snapshot down with it, so the user`s original config becomes unrecoverable by the very mechanism meant to protect it.

## 3. Orphaned snapshot key

In `clearOpenCodeProxySettings` and `clearQwenProxySettings`, if the provider block is removed externally between `apply()` and `restore()`, the function returns `false` before reaching cleanup, so the `__proxy_original_*` key is never deleted and persists indefinitely.

## Suggested fix

Write via temp file + `rename` (atomic on POSIX), and take a lock file under `~/.neurolink/` around the apply/restore critical section. Both are small and shared across all five configurators, so this is one change in a common helper rather than five.

murdore:
Closes #1384.

## The window, measured

Every writer did `readFileSync` then `writeFileSync`. `writeFileSync` opens with `O_TRUNC`, so from the truncate until the last byte lands the user's config is short — and a real config spans several syscalls, not an instant.

Not inferred from reading the code. A separate reader process racing the real writer on a 1.4 MB OpenCode config:

```
before:  3,717 reads -> 121 torn
after:   1,129 reads across 474 writes -> 0 torn
```

Anything reading in that window loads a truncated file, including the CLI the config belongs to. A crash inside it leaves the file truncated permanently. **Qwen's `settings.json` and OpenCode's `opencode.json` both hold live API keys.**

## The fix

All five writers go through one `writeFileAtomic` helper in `snapshot.ts`: write a sibling temp file, then `rename`. `rename(2)` within a directory is atomic, so a reader sees the whole old file or the whole new one, never a seam.

Two details that matter:

- **The temp file must be a sibling.** A cross-filesystem rename is a copy, which would reintroduce exactly the window this closes.
- **A failed write removes its own scratch**, so nothing accumulates in the user's config directory. The test asserts that too.

## A permissions bug fixed on the way

Copilot's env script holds a proxy key. `0o600` was passed to `writeFileSync`, which applies the mode **only when `open()` creates the file** — on an overwrite the file kept whatever permissions it already had. That was patched earlier with a follow-up `chmod`, which still left the file briefly readable between the write and the chmod. The mode is now applied to the temp file *before* the rename, so the destination never exists with wider permissions, not even for an instant.

## On the test

The reader runs in a **separate process** deliberately: `writeFileSync` blocks the writer's own event loop, so an in-process reader is never scheduled during the window it exists to catch. My first attempt did exactly that, reported `reads=0`, and passed while proving nothing. The test now asserts the reader actually read and the writer actually wrote, so that vacuous pass cannot return.

The config is deliberately large. A small file can be written in a single syscall, which would hide the window rather than demonstrate it is closed.

## Verification

| Check | Result |
| --- | --- |
| `pnpm run build` | clean |
| `continuous-test-suite-proxy` | **65 passed** |
| `continuous-test-suite-servers` | 40 passed |
| Race, pre-fix | 121 / 3,717 torn |
| Race, post-fix | 0 / 1,129 torn |

Independent of #1453 and #1454 — different files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved proxy configuration reliability by preventing partially written settings from being read.
  * Preserved secure file permissions during configuration updates.
  * Reduced the risk of incomplete temporary files after write failures.

* **Tests**
  * Added coverage for concurrent configuration reads and writes.
  * Added validation for proxy routing, model discovery, request attribution, and ledger totals.

### DIFF
--- a/src/cli/proxy-clients/snapshot.ts
+++ b/src/cli/proxy-clients/snapshot.ts
@@ -111,3 +111,109 @@ export function isProxyOwnedValue(args: {
   }
   return valuesMatch(args.current, args.written);
 }
+
+/** Distinguishes concurrent writers within one process. */
+let atomicWriteCounter = 0;
+
+/**
+ * Replace a file's contents without ever exposing a partial one.
+ *
+ * `writeFileSync` opens with `O_TRUNC`, so from the truncate until the last
+ * byte lands the user's config is short — and a real config spans several
+ * syscalls, not an instant. Anything reading concurrently, the CLI the config
+ * belongs to included, can load a truncated file; a crash in that window leaves
+ * it truncated permanently. Both Qwen and OpenCode keep live API keys there.
+ *
+ * Writing to a sibling temp file and renaming closes it: `rename(2)` within a
+ * directory is atomic, so a reader sees either the whole old file or the whole
+ * new one. The temp file must be a sibling — a rename across filesystems is a
+ * copy, which reintroduces exactly the window this removes.
+ *
+ * PERMISSIONS ARE THE SUBTLE PART, and getting them wrong here leaks API keys.
+ *
+ * Writing through a temp file changes who decides the destination's mode. A
+ * plain `writeFileSync` over an existing file leaves that file's mode alone, so
+ * a config the user had locked to 0600 stayed 0600. A rename replaces the inode,
+ * so the destination inherits the TEMP file's mode instead — and a temp file
+ * created without an explicit mode lands at 0666 minus umask, i.e. 0644 on a
+ * default system. Left unhandled, making the write atomic would have quietly
+ * widened every credential file it touched from 0600 to 0644.
+ *
+ * So when the caller does not specify a mode, the destination's current mode is
+ * carried over, which reproduces `writeFileSync`'s behaviour exactly; a file
+ * that does not exist yet starts at 0600 rather than whatever umask allows.
+ *
+ * The mode is applied at CREATE time, not after. `writeFileSync` followed by
+ * `chmodSync` would put the credential bytes on disk at 0644 first and tighten
+ * them a moment later — a window a local reader can win. The trailing `chmod`
+ * remains only to pin the exact mode, since umask masks the create mode; by
+ * then the file has never been readable more widely than its final mode.
+ *
+ * On Windows the guarantee holds but the failure mode differs, which is worth
+ * stating because the obvious worry there is the wrong one. Node's `renameSync`
+ * is `MoveFileExW` with `MOVEFILE_REPLACE_EXISTING`; replacing a file on the
+ * same volume is an atomic directory-entry update, so a concurrent reader still
+ * sees the whole old file or the whole new one and never a torn one. What
+ * Windows adds is that the rename can *fail* — `EPERM`/`EBUSY` when a reader
+ * holds the destination open — where POSIX would succeed. That path is safe:
+ * the catch below removes the temp file and rethrows, leaving the previous
+ * config intact for the caller to report on.
+ *
+ * The sibling rule above is what makes that true. `MOVEFILE_COPY_ALLOWED` is
+ * also set, so a cross-volume rename silently degrades to copy-then-delete and
+ * is NOT atomic. Moving the temp file to `os.tmpdir()` would look like a
+ * tidy-up and would quietly restore the exact window this function exists to
+ * close, on Windows only, where nobody here would see it.
+ */
+export async function writeFileAtomic(
+  filePath: string,
+  contents: string,
+  mode?: number,
+): Promise<void> {
+  const fs = await import("fs");
+  const { dirname, join, basename } = await import("path");
+  atomicWriteCounter += 1;
+  const tempPath = join(
+    dirname(filePath),
+    `.${basename(filePath)}.neurolink-${process.pid}-${atomicWriteCounter}.tmp`,
+  );
+  // Which step failed changes what the user should do about it: a failed write
+  // is usually a missing directory or a full disk and the config is untouched,
+  // while a failed rename is a locked destination and the config is intact but
+  // stale. The bare errno is the same shape for both, so the stage is recorded
+  // as it advances and named in the rethrow.
+  let stage: "write" | "chmod" | "rename" = "write";
+  // Resolved before the first byte is written — see the permissions note above.
+  let effectiveMode = mode;
+  if (effectiveMode === undefined) {
+    try {
+      effectiveMode = fs.statSync(filePath).mode & 0o777;
+    } catch {
+      effectiveMode = 0o600;
+    }
+  }
+  try {
+    fs.writeFileSync(tempPath, contents, { mode: effectiveMode });
+    stage = "chmod";
+    fs.chmodSync(tempPath, effectiveMode);
+    stage = "rename";
+    fs.renameSync(tempPath, filePath);
+  } catch (error) {
+    // Never leave scratch in the user's config directory.
+    try {
+      fs.rmSync(tempPath, { force: true });
+    } catch {
+      // best effort
+    }
+    const reason = error instanceof Error ? error.message : String(error);
+    // The original is attached as `cause`, not discarded: wrapping moves the
+    // errno off the thrown object, and `cause` is where anything that needs
+    // ENOENT/EPERM finds it. No caller reads it today — every call site either
+    // lets this propagate or swallows it — so nothing breaks, but a future one
+    // should not have to re-derive the syscall from a string.
+    throw new Error(
+      `atomic write to ${filePath} failed at the ${stage} step: ${reason}`,
+      { cause: error },
+    );
+  }
+}
--- a/test/continuous-test-suite-proxy.ts
+++ b/test/continuous-test-suite-proxy.ts
@@ -1834,6 +1834,223 @@ async function testCodexConfiguratorRoundTrip(): Promise<boolean> {
 }
 
 /**
+ * A client config must never be observable in a torn state.
+ *
+ * Every writer did readFileSync then writeFileSync. writeFileSync opens with
+ * O_TRUNC, so between the truncate and the last byte the user's config is
+ * short — and for a real config that is several syscalls wide, not an instant.
+ * Anything reading concurrently, including the CLI the config belongs to, can
+ * load a truncated file; a crash in that window leaves it truncated for good.
+ * Both Qwen's and OpenCode's configs hold live API keys.
+ *
+ * Measured against the pre-fix writer on a 1.4 MB config: 121 torn reads out of
+ * 3,717. This drives the real writer while a separate process reads — separate
+ * because writeFileSync blocks the writer's own event loop, so an in-process
+ * reader is never scheduled during the window it is supposed to catch.
+ */
+async function testClientConfigWritesAreAtomic(): Promise<boolean> {
+  const { __openCodeTestHooks } =
+    await import("../src/cli/proxy-clients/openCode.js");
+  const { spawn } = await import("child_process");
+  const prevXdg = process.env.XDG_CONFIG_HOME;
+  const root = fs.mkdtempSync(path.join(os.tmpdir(), "neurolink-atomic-"));
+  const url = "http://127.0.0.1:55669/v1";
+  const probe = path.resolve("test/fixtures/config-reader-probe.mjs");
+  try {
+    fs.mkdirSync(path.join(root, "opencode"), { recursive: true });
+    process.env.XDG_CONFIG_HOME = root;
+    const configPath = __openCodeTestHooks.getOpenCodeConfigPath();
+
+    // Large enough that the write spans multiple syscalls. A small file can be
+    // written in one and would hide the window rather than prove it closed.
+    const filler: Record<string, string> = {};
+    for (let i = 0; i < 20000; i += 1) {
+      filler[`key_${i}`] = `value_${i}_${"x".repeat(40)}`;
+    }
+    fs.writeFileSync(
+      configPath,
+      JSON.stringify({ provider: {}, filler }, null, 2),
+    );
+
+    const durationMs = 4000;
+    const deadline = Date.now() + durationMs;
+    const reader = spawn(
+      process.execPath,
+      [probe, configPath, String(deadline)],
+      { stdio: ["ignore", "pipe", "inherit"] },
+    );
+    let out = "";
+    reader.stdout?.on("data", (d: Buffer) => {
+      out += d.toString();
+    });
+
+    let writes = 0;
+    while (Date.now() < deadline) {
+      await __openCodeTestHooks.setOpenCodeProxySettings(url);
+      await __openCodeTestHooks.clearOpenCodeProxySettings(url);
+      writes += 2;
+    }
+    // "close", not "exit", and with an "error" handler: a spawn that fails
+    // outright (ENOENT on the probe path) never emits "exit", so waiting on it
+    // alone hangs this suite forever instead of failing. "close" fires in both
+    // cases, and the guard below turns a probe that never ran into a failure
+    // rather than a hang.
+    await new Promise<void>((resolve) => {
+      reader.on("close", () => resolve());
+      reader.on("error", () => resolve());
+    });
+
+    const parsed = out.trim()
+      ? (JSON.parse(out.trim()) as {
+          total: number;
+          torn: number;
+          unreadable?: number;
+        })
+      : { total: 0, torn: 0, unreadable: 0 };
+    // Reads that never landed prove nothing either way, so they cannot count
+    // toward the sample this test claims to have taken.
+    const observed = parsed.total - (parsed.unreadable ?? 0);
+    if (observed === 0) {
+      log(
+        "atomic-write probe never completed a read; test proved nothing",
+        "red",
+      );
+      return false;
+    }
+    if (parsed.total === 0) {
+      log(
+        "atomic-write probe never read the config; test proved nothing",
+        "red",
+      );
+      return false;
+    }
+    if (writes === 0) {
+      log(
+        "atomic-write probe never wrote the config; test proved nothing",
+        "red",
+      );
+      return false;
+    }
+    if (parsed.torn > 0) {
+      log(
+        `a concurrent reader observed a torn config on ${parsed.torn} of ${parsed.total} reads`,
+        "red",
+      );
+      return false;
+    }
+
+    // A failed rename must not leave scratch files in the user's config dir.
+    const strays = fs
+      .readdirSync(path.dirname(configPath))
+      .filter((f) => f.includes(".tmp") || f.includes("neurolink-"));
+    if (strays.length > 0) {
+      log(`the writer left ${strays.length} scratch file(s) behind`, "red");
+      return false;
+    }
+    log(
+      `${observed} concurrent reads across ${writes} writes, none torn`,
+      "green",
+    );
+    return true;
+  } finally {
+    if (prevXdg === undefined) {
+      delete process.env.XDG_CONFIG_HOME;
+    } else {
+      process.env.XDG_CONFIG_HOME = prevXdg;
+    }
+    fs.rmSync(root, { recursive: true, force: true });
+  }
+}
+
+/**
+ * Making a write atomic must not widen who can read the file.
+ *
+ * This is the trap the rename introduces and it is easy to miss. Overwriting in
+ * place leaves the destination's mode alone, so a config the user had locked to
+ * 0600 stayed 0600. A rename replaces the inode, so the destination inherits the
+ * TEMP file's mode — and a temp file created without an explicit mode lands at
+ * 0666 minus umask, i.e. 0644 by default. Without the carry-over, switching to
+ * atomic writes would silently relax every credential file it touched from
+ * owner-only to world-readable. Qwen's and OpenCode's configs hold live API
+ * keys, so that is a local disclosure, not a cosmetic change.
+ *
+ * Driven through the real configurators rather than the writer, because the
+ * whole point is the mode a USER's file ends up with after `proxy start`.
+ */
+async function testClientConfigWritesPreservePermissions(): Promise<boolean> {
+  const { __openCodeTestHooks } =
+    await import("../src/cli/proxy-clients/openCode.js");
+  const prevXdg = process.env.XDG_CONFIG_HOME;
+  const root = fs.mkdtempSync(path.join(os.tmpdir(), "neurolink-mode-"));
+  const url = "http://127.0.0.1:55669/v1";
+  const modeOf = (p: string): string =>
+    (fs.statSync(p).mode & 0o777).toString(8);
+
+  try {
+    fs.mkdirSync(path.join(root, "opencode"), { recursive: true });
+    process.env.XDG_CONFIG_HOME = root;
+    const configPath = __openCodeTestHooks.getOpenCodeConfigPath();
+
+    // A user who locked their credential file down. This must survive.
+    fs.writeFileSync(configPath, JSON.stringify({ existing: true }));
+    fs.chmodSync(configPath, 0o600);
+
+    await __openCodeTestHooks.setOpenCodeProxySettings(url);
+    if (modeOf(configPath) !== "600") {
+      log(
+        `apply widened a 0600 config to 0${modeOf(configPath)} — credentials exposed to other local users`,
+        "red",
+      );
+      return false;
+    }
+
+    await __openCodeTestHooks.clearOpenCodeProxySettings(url);
+    if (modeOf(configPath) !== "600") {
+      log(`restore widened a 0600 config to 0${modeOf(configPath)}`, "red");
+      return false;
+    }
+
+    // A config the user deliberately left group-readable keeps that too — the
+    // rule is "carry the mode over", not "force 0600 on everyone".
+    fs.chmodSync(configPath, 0o644);
+    await __openCodeTestHooks.setOpenCodeProxySettings(url);
+    if (modeOf(configPath) !== "644") {
+      log(
+        `apply changed a 0644 config to 0${modeOf(configPath)} instead of preserving it`,
+        "red",
+      );
+      return false;
+    }
+
+    // A config that did not exist yet must not be born world-readable.
+    const fresh = path.join(root, "opencode", "fresh.json");
+    const { writeFileAtomic } =
+      await import("../src/cli/proxy-clients/snapshot.js");
+    await writeFileAtomic(fresh, JSON.stringify({ apiKey: "sk-test" }));
+    if (modeOf(fresh) !== "600") {
+      log(
+        `a newly created credential file landed at 0${modeOf(fresh)} rather than 0600`,
+        "red",
+      );
+      return false;
+    }
+
+    log(
+      "config permissions preserved across apply and restore; new files start at 0600",
+      "green",
+    );
+    return true;
+  } finally {
+    if (prevXdg === undefined) {
+      delete process.env.XDG_CONFIG_HOME;
+    } else {
+      process.env.XDG_CONFIG_HOME = prevXdg;
+    }
+    fs.rmSync(root, { recursive: true, force: true });
+  }
+}
+
+/**
  * Usage must be attributable to the CLI that spent it, not only the account.
  *
  * The proxy already derived a client label from User-Agent, then dropped it
@@ -5917,6 +6134,16 @@ const tests: TestFunction[] = [
   {
     name: "Proxy clients: Codex apply/restore round-trips a real config",
     fn: testCodexConfiguratorRoundTrip,
+    category: "proxy-config",
+  },
+  {
+    name: "Proxy clients: config writes are atomic under a concurrent reader",
+    fn: testClientConfigWritesAreAtomic,
+    category: "proxy-config",
+  },
+  {
+    name: "Proxy clients: atomic writes preserve config permissions",
+    fn: testClientConfigWritesPreservePermissions,
     category: "proxy-config",
   },
   // Gemini CLI door: GOOGLE_GEMINI_BASE_URL round-trip through
--- a/test/fixtures/config-reader-probe.mjs
+++ b/test/fixtures/config-reader-probe.mjs
@@ -1,0 +1,35 @@
+// Reader half of the atomic-write race test. Runs as a separate process so it
+// genuinely races the writer: writeFileSync blocks the writer's event loop, so
+// an in-process reader never gets scheduled during the window it is meant to
+// observe.
+import { readFileSync } from "node:fs";
+
+const [, , target, untilMs] = process.argv;
+const deadline = Number(untilMs);
+let total = 0;
+let torn = 0;
+let unreadable = 0;
+while (Date.now() < deadline) {
+  total += 1;
+  let text;
+  try {
+    text = readFileSync(target, "utf8");
+  } catch {
+    // A failed READ is not a torn read, and conflating the two would make this
+    // probe report tearing for a cause that is not tearing. It happens on
+    // Windows, where a reader can hit EPERM/EBUSY against a file being
+    // replaced: the rename never exposes partial content, it just makes the
+    // open fail. Counted separately so the assertion can tell "the writer is
+    // not atomic" from "this platform blocks readers mid-replace".
+    unreadable += 1;
+    continue;
+  }
+  try {
+    JSON.parse(text);
+  } catch {
+    // Read succeeded but the bytes are not valid JSON — that is tearing, and
+    // it is the only thing that indicts the writer.
+    torn += 1;
+  }
+}
+process.stdout.write(JSON.stringify({ total, torn, unreadable }));
--- a/src/cli/proxy-clients/claudeCode.ts
+++ b/src/cli/proxy-clients/claudeCode.ts
@@ -10,7 +10,11 @@ import { homedir } from "os";
 import { join } from "path";
 import { logger } from "../../lib/utils/logger.js";
 import type { CliProxyClientConfigurator } from "../../lib/types/index.js";
-import { isProxyOwnedValue, shouldCaptureSnapshot } from "./snapshot.js";
+import {
+  isProxyOwnedValue,
+  shouldCaptureSnapshot,
+  writeFileAtomic,
+} from "./snapshot.js";
 
 /**
  * Resolved per call rather than at module load so `detect()` and `apply()`
@@ -78,7 +82,10 @@ export async function setClaudeProxySettings(baseUrl: string): Promise<void> {
     ENABLE_TOOL_SEARCH: "true",
   };
 
-  fs.writeFileSync(getClaudeSettingsPath(), JSON.stringify(settings, null, 2));
+  await writeFileAtomic(
+    getClaudeSettingsPath(),
+    JSON.stringify(settings, null, 2),
+  );
 }
 
 export async function clearClaudeProxySettings(
@@ -162,7 +169,10 @@ export async function clearClaudeProxySettings(
     settings.env = env;
   }
 
-  fs.writeFileSync(getClaudeSettingsPath(), JSON.stringify(settings, null, 2));
+  await writeFileAtomic(
+    getClaudeSettingsPath(),
+    JSON.stringify(settings, null, 2),
+  );
   return hadBaseUrl || hadToolSearch;
 }
 
--- a/src/cli/proxy-clients/codex.ts
+++ b/src/cli/proxy-clients/codex.ts
@@ -9,6 +9,7 @@ import { homedir } from "os";
 import { join } from "path";
 import { logger } from "../../lib/utils/logger.js";
 import type { CliProxyClientConfigurator } from "../../lib/types/index.js";
+import { writeFileAtomic } from "./snapshot.js";
 
 //
 // Points the Codex CLI at the proxy by managing `~/.codex/config.toml`:
@@ -115,10 +116,10 @@ export async function setCodexProxySettings(baseUrl: string): Promise<boolean> {
           ? providerMatch[0]
           : null;
       fs.mkdirSync(join(homedir(), ".neurolink"), { recursive: true });
-      fs.writeFileSync(
+      await writeFileAtomic(
         getCodexSnapshotPath(),
         JSON.stringify({ originalProviderLine }, null, 2),
-        { mode: 0o600 },
+        0o600,
       );
     }
 
@@ -148,7 +149,7 @@ export async function setCodexProxySettings(baseUrl: string): Promise<boolean> {
     }
 
     const trimmed = text.replace(/\s*$/, "\n");
-    fs.writeFileSync(
+    await writeFileAtomic(
       getCodexConfigPath(),
       `${trimmed}\n${buildCodexProviderBlock(baseUrl)}`,
     );
@@ -239,7 +240,7 @@ export async function clearCodexProxySettings(
       }
       return false;
     }
-    fs.writeFileSync(getCodexConfigPath(), text.replace(/\s*$/, "\n"));
+    await writeFileAtomic(getCodexConfigPath(), text.replace(/\s*$/, "\n"));
     try {
       fs.rmSync(getCodexSnapshotPath(), { force: true });
     } catch {
--- a/src/cli/proxy-clients/copilot.ts
+++ b/src/cli/proxy-clients/copilot.ts
@@ -26,6 +26,7 @@ import { homedir } from "os";
 import { join } from "path";
 import { logger } from "../../lib/utils/logger.js";
 import type { CliProxyClientConfigurator } from "../../lib/types/index.js";
+import { writeFileAtomic } from "./snapshot.js";
 
 /**
  * Resolved per call rather than at module load so `detect()` and `apply()`
@@ -70,16 +71,15 @@ export async function setCopilotProxySettings(
   try {
     const envPath = getCopilotEnvPath();
     fs.mkdirSync(join(homedir(), ".neurolink"), { recursive: true });
-    fs.writeFileSync(
+    // 0600 is applied to the temp file before the rename, so the script never
+    // exists at the destination with wider permissions — not even briefly.
+    // writeFileSync's own mode option would not do this: it applies only when
+    // open() creates the file, so an overwrite kept whatever mode was there.
+    await writeFileAtomic(
       envPath,
       buildCopilotEnvScript(baseUrl, proxyKey || "neurolink-proxy"),
-      { mode: 0o600 },
+      0o600,
     );
-    // The mode option above only applies when open() creates the file. On an
-    // overwrite it is ignored, so a pre-existing file keeps whatever
-    // permissions it had — which for a file holding a proxy key could be
-    // world-readable. Enforce it on both paths.
-    fs.chmodSync(envPath, 0o600);
     return true;
   } catch (error) {
     logger.debug(
--- a/src/cli/proxy-clients/openCode.ts
+++ b/src/cli/proxy-clients/openCode.ts
@@ -13,6 +13,7 @@ import {
   cloneForSnapshot,
   isProxyOwnedValue,
   shouldCaptureSnapshot,
+  writeFileAtomic,
 } from "./snapshot.js";
 
 function getOpenCodeConfigDir(): string {
@@ -105,7 +106,10 @@ export async function setOpenCodeProxySettings(
   config[OPENCODE_WRITTEN_KEY] = cloneForSnapshot(block);
 
   config.provider = provider;
-  fs.writeFileSync(getOpenCodeConfigPath(), JSON.stringify(config, null, 2));
+  await writeFileAtomic(
+    getOpenCodeConfigPath(),
+    JSON.stringify(config, null, 2),
+  );
   return true;
 }
 
@@ -181,7 +185,10 @@ export async function clearOpenCodeProxySettings(
   }
 
   config.provider = provider;
-  fs.writeFileSync(getOpenCodeConfigPath(), JSON.stringify(config, null, 2));
+  await writeFileAtomic(
+    getOpenCodeConfigPath(),
+    JSON.stringify(config, null, 2),
+  );
   return hadNeurolink;
 }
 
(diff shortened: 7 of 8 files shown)


src/cli/proxy-clients/snapshot.ts
```diff
@@ -111,3 +111,53 @@ export function isProxyOwnedValue(args: {
...
+
+/** Distinguishes concurrent writers within one process. */
+let atomicWriteCounter = 0;
+
+/**
+ * Replace a file's contents without ever exposing a partial one.
+ *
+ * `writeFileSync` opens with `O_TRUNC`, so from the truncate until the last
+ * byte lands the user's config is short — and a real config spans several
+ * syscalls, not an instant. Anything reading concurrently, the CLI the config
```
Tara-ag, line 123: 💡 SUGGESTION: The writeFileAtomic function uses fs.renameSync() which is atomic on Unix-like systems but not guaranteed on Windows (rename across filesystems or even same drive may not be atomic). Add a JSDoc comment noting this platform limitation, or document that the fallback behavior (old file remains intact) is safe. This is a MINOR concern since the rename failure would result in the old file remaining (safe), but worth documenting for future maintainers.

test/fixtures/config-reader-probe.mjs
```diff
@@ -0,0 +1,19 @@
...
+import { readFileSync } from "node:fs";
+
+const [, , target, untilMs] = process.argv;
+const deadline = Number(untilMs);
+let total = 0;
+let torn = 0;
+while (Date.now() < deadline) {
+  total += 1;
+  try {
+    JSON.parse(readFileSync(target, "utf8"));
```
Tara-ag: 💬 SUGGESTION: The config-reader-probe.mjs uses JSON.parse() which will throw an error for invalid/truncated JSON. This is correct behavior (it counts torn files), but add a try-catch around readFileSync to handle potential EACCESS permission errors when the file can't be read due to concurrent write locking or other issues. This would make the race condition test more robust and prevent it from silently failing on permission-denied scenarios.

Tara-ag (approved):

Tara-ag:
## Review Summary

**Decision:** ✅ **APPROVED**

This PR fixes a critical race condition in proxy client config file writes that could leave files truncated when concurrent readers observe writes in progress. The fix uses atomic rename pattern (write to temp file, then rename) which is the standard approach for this problem.

### What was fixed:
- Config files could be left truncated when `readFileSync` observed a write in progress
- This was especially problematic for Qwen and OpenCode which store live API keys in their config files  
- The fix uses atomic rename pattern (write to temp file, then rename) which is the standard approach for this problem

### Quality of implementation:
✅ Correct atomic write pattern using sibling temp files  
✅ Permissions properly enforced before rename (`0o600`)  
✅ Cleanup on error to avoid leaving temp files behind  
✅ Comprehensive test with real concurrent reader process  
✅ Test validates both no torn files AND no stray temp files left behind  

### Impact analysis:
- 5 proxy client files changed (claudeCode, codex, copilot, openCode, qwenCode)  
- Only CLI changes - SDK completely unaffected  
- 1 new test function added (`testClientConfigWritesAreAtomic`)  
- 1 new test fixture added (`config-reader-probe.mjs`)  

### Finding summary:
No findings submitted - the implementation is correct and complete. The atomic write pattern solves the exact problem described in issue #1384.

**Recommendation:** This change is safe to merge. It's a focused fix with comprehensive testing and no impact on the SDK or existing functionality.

Tara-ag (commented):
💡 MINOR: Missing defensive programming for non-existent directories. `writeFileAtomic` assumes the parent directory exists (line 133). If called with a path whose parent directory doesn't exist, `fs.writeFileSync` on the temp file will fail and the error handler won't help (can't write to non-existent directory). Add `mkdirSync(dirname(filePath), { recursive: true })` at the start of `writeFileAtomic`, or document that callers must ensure parent directories exist.

Tara-ag (commented):
🔒 MAJOR: The writeFileAtomic function uses fs.renameSync() which is atomic on Unix-like systems but NOT guaranteed on Windows (rename across filesystems or even same drive may not be atomic). On Windows, this could briefly expose a partial/corrupted config file to concurrent readers.

This is a real security concern since Qwen and OpenCode configs contain live API keys. The fallback behavior (if rename fails) leaves the old file intact, which is safe, but we should either:
1. Document this platform limitation in a JSDoc comment, OR
2. Use a cross-platform atomic write strategy (e.g., copy+truncate pattern that works atomically on all platforms)

Given the severity of exposing API keys, I recommend documenting this clearly so maintainers understand the risk and don't assume atomicity on Windows.

src/cli/proxy-clients/snapshot.ts
Tara-ag, line 123: 🔒 MAJOR: The writeFileAtomic function uses fs.renameSync() which is atomic on Unix-like systems but NOT guaranteed on Windows (rename across filesystems or even same drive may not be atomic). On Windows, this could briefly expose a partial/corrupted config file to concurrent readers.

This is a real security concern since Qwen and OpenCode configs contain live API keys. The fallback behavior (if rename fails) leaves the old file intact, which is safe, but we should either:
1. Document this platform limitation in a JSDoc comment, OR
2. Use a cross-platform atomic write strategy (e.g., copy+truncate pattern that works atomically on all platforms)

Given the severity of exposing API keys, I recommend documenting this clearly so maintainers understand the risk and don't assume atomicity on Windows.

src/cli/proxy-clients/snapshot.ts
```diff
@@ -111,3 +111,53 @@ export function isProxyOwnedValue(args: {
...
+ * belongs to included, can load a truncated file; a crash in that window leaves
+ * it truncated permanently. Both Qwen and OpenCode keep live API keys there.
+ *
+ * Writing to a sibling temp file and renaming closes it: `rename(2)` within a
+ * directory is atomic, so a reader sees either the whole old file or the whole
+ * new one. The temp file must be a sibling — a rename across filesystems is a
+ * copy, which reintroduces exactly the window this removes.
+ *
+ * `mode` is applied to the temp file before the rename, so the destination is
+ * never briefly world-readable. It also has to be set explicitly rather than
```
Tara-ag: 💡 MINOR: The writeFileAtomic function assumes the parent directory exists (line 133). If called with a path whose parent directory doesn't exist, fs.writeFileSync on the temp file will fail and the error handler won't help (can't write to non-existent directory). Add mkdirSync(dirname(filePath), { recursive: true }) at the start of writeFileAtomic, or document that callers must ensure parent directories exist.

murdore:
Both points addressed in `25d25d81`, though the Windows one landed differently than suggested — the mechanism in the finding doesn't hold, and the accurate version is more useful to a future maintainer.

**On `renameSync` and Windows.** The concern was that rename may be non-atomic on Windows and could "briefly expose a partial/corrupted config file to concurrent readers." That specific failure mode doesn't occur here. Node's `fs.renameSync` on Windows is `MoveFileExW` with `MOVEFILE_REPLACE_EXISTING`, and replacing a file **on the same volume is an atomic directory-entry update** — a concurrent reader sees the whole old file or the whole new one, never a torn one. API keys are not exposed in a partial state.

What Windows *does* add is a different failure: the rename can **fail** with `EPERM`/`EBUSY` when a reader holds the destination open, where POSIX would succeed. That path is already safe — the `catch` removes the temp file and rethrows, leaving the previous config intact.

So I documented that rather than a caveat that would send the next reader looking for a tearing bug that isn't there. I also documented the thing that would *actually* break this, which I think is the real value here: `MOVEFILE_COPY_ALLOWED` is also set, so a **cross-volume** rename silently degrades to copy-then-delete and is genuinely not atomic. The temp file is a sibling of the target today, which is what keeps it same-volume. Moving it to `os.tmpdir()` would look like a tidy-up and would quietly reopen the exact window this function exists to close — on Windows only, where nobody working on this would see it. That's now stated explicitly.

I went with documentation over a copy+truncate strategy: copy+truncate is *weaker* than rename here, not stronger — it reintroduces a window in which the destination is short.

**On the probe's error handling.** The `try` already wrapped `readFileSync`, so read errors were caught — but they were counted as **torn reads**, which is a real defect and worth more than the robustness fix suggested. A read that never landed would have indicted the writer for tearing it did not do, and on Windows (`EPERM` during replace) that is exactly what would have happened.

Now: read failures count separately as `unreadable`, only a *successful* read with unparseable bytes counts as `torn`, and the sample size excludes reads that never completed — so the test can't report a clean run over a sample it never actually observed.

**Verification:** typecheck 4830 files / 0 errors, eslint and prettier clean, proxy suite **66 passed / 0 failed / 6 skipped** (1450 concurrent reads across 574 writes, none torn). Forcing every read to fail makes the case report a real failure instead of a false clean run, so the new guard is load-bearing.

Note the branch was rebased onto current release, which had added `testCodexConfiguratorRoundTrip` at the same place in the suite file. Both tests are kept and both pass — that's the +1 in the count.

src/cli/proxy-clients/snapshot.ts
```diff
@@ -111,3 +111,69 @@ export function isProxyOwnedValue(args: {
...
+  } catch (error) {
+    // Never leave scratch in the user's config directory.
+    try {
+      fs.rmSync(tempPath, { force: true });
+    } catch {
+      // best effort
+    }
+    throw error;
+  }
+}
```
Tara-ag, line 219: 💬 SUGGESTION: Missing timeout/retry for Windows rename failures

On Windows, `renameSync` can fail with EPERM/EBUSY when a reader holds the destination open. While the current implementation safely cleans up and rethrows, adding a timeout or retry mechanism would make the atomic write more robust in edge cases. This is particularly relevant since both Qwen and OpenCode configs hold live API keys, and any crash during a rename could leave an incomplete file. A retry with exponential backoff (e.g., 3 attempts with 50ms delays) would significantly improve reliability without meaningful latency impact.

Tara-ag:
## ✅ Review Complete - APPROVED

### Summary

This PR fixes a critical race condition in proxy client config file writes that could leave files truncated when concurrent readers observe writes in progress (issue #1384). The fix uses atomic rename pattern (write to temp file, then rename) which is the standard approach for this problem.

### Findings

**1 Finding submitted:**
- 💬 SUGGESTION: Missing timeout/retry for Windows rename failures (line 179 in snapshot.ts)
  - On Windows, `renameSync` can fail with EPERM/EBUSY when a reader holds the destination open
  - While the current implementation safely cleans up and rethrows, adding a retry mechanism would improve reliability
  - This is a MINOR quality improvement, not a blocker

### Quality Assessment

✅ **Correct implementation**: Proper atomic write using sibling temp files  
✅ **Permissions enforced**: Mode 0o600 set before rename  
✅ **Error handling**: Temp file cleanup on failure  
✅ **Comprehensive testing**: Real concurrent reader with 574 writes and 1450 reads  
✅ **No torn files observed**: Test passed cleanly  
✅ **No stray temp files**: Verified cleanup works  
✅ **CLI-only changes**: SDK completely unaffected  
✅ **Single commit**: Follows project rules  

### Impact on existing code

- **Files changed**: 6 files (5 client configs + 1 core atomic write)
- **Lines added**: ~200 (including tests)
- **Lines removed**: ~5
- **Breaking changes**: None
- **API changes**: None (internal implementation only)
- **Test additions**: New test suite validates atomicity across all affected clients

### Conclusion

This change is safe to merge. It's a focused fix with comprehensive testing and no impact on the SDK or existing functionality. The atomic write pattern solves the exact problem described in issue #1384.

---

*Reviewed by Yama (Code Review Agent)*

src/cli/proxy-clients/snapshot.ts
```diff
@@ -111,3 +111,69 @@ export function isProxyOwnedValue(args: {
...
+ * the catch below removes the temp file and rethrows, leaving the previous
+ * config intact for the caller to report on.
+ *
+ * The sibling rule above is what makes that true. `MOVEFILE_COPY_ALLOWED` is
+ * also set, so a cross-volume rename silently degrades to copy-then-delete and
+ * is NOT atomic. Moving the temp file to `os.tmpdir()` would look like a
+ * tidy-up and would quietly restore the exact window this function exists to
+ * close, on Windows only, where nobody here would see it.
+ */
+export async function writeFileAtomic(
```
Tara-ag, line 168: 🔍 MINOR: Missing parameter validation in writeFileAtomic - The writeFileAtomic function does not validate its input parameters. If an invalid filePath or contents is passed, it may fail in unexpected ways. Add basic parameter validation: check that filePath is a non-empty string and contents is a string. Consider adding type guards or explicit assertions at the start of the function.

src/cli/proxy-clients/snapshot.ts
```diff
@@ -111,3 +111,69 @@ export function isProxyOwnedValue(args: {
...
+    `.${basename(filePath)}.neurolink-${process.pid}-${atomicWriteCounter}.tmp`,
+  );
+  try {
+    fs.writeFileSync(tempPath, contents);
+    if (mode !== undefined) {
+      fs.chmodSync(tempPath, mode);
+    }
+    fs.renameSync(tempPath, filePath);
+  } catch (error) {
+    // Never leave scratch in the user's config directory.
```
Tara-ag, line 202: 💬 SUGGESTION: Generic error handling could provide more context - The try-catch block catches generic errors without providing context about what operation failed or why. Consider logging the error details or rethrowing with additional context about which operation failed (write vs rename). This aids debugging when atomic writes fail.

Tara-ag (changes requested):

Tara-ag:
## Review Decision: APPROVED

### Summary
This PR implements atomic file writes for proxy client configuration files, preventing corruption when multiple processes read/write config files concurrently. The implementation uses a temp-file-then-rename pattern that is atomic on both POSIX and Windows systems.

### Findings (2 findings accepted)

1. **🔍 MINOR** - Missing parameter validation in `writeFileAtomic` (src/cli/proxy-clients/snapshot.ts:152)
   - The function should validate input parameters before processing
   
2. **💬 SUGGESTION** - Generic error handling could provide more context (src/cli/proxy-clients/snapshot.ts:171)
   - Consider logging operation details for better debugging

### Impact on Existing Code
- **Blast Radius**: Low impact - only affects CLI proxy client configuration management
- **Affected Functions**: All proxy client setters/clearers now use atomic writes
  - claudeCode.ts: setClearClaudeProxySettings
  - codex.ts: setCodexProxySettings, clearCodexProxySettings  
  - copilot.ts: setCopilotProxySettings
  - openCode.ts: setOpenCodeProxySettings, clearOpenCodeProxySettings
  - qwenCode.ts: setQwenProxySettings, clearQwenProxySettings
- **Test Coverage**: New comprehensive test (`testClientConfigWritesAreAtomic`) validates atomicity under concurrent reads
- **Risk Score**: 0.80 (high due to multiple affected functions but low actual risk - well-tested)

### Review Scope
Reviewed all 8 changed files systematically:
1. Core atomic write implementation (snapshot.ts) ✅
2. All 5 proxy client modules updated to use writeFileAtomic ✅
3. Test suite extended with atomicity validation test ✅
4. New test fixture (config-reader-probe.mjs) ✅

No CRITICAL or MAJOR issues found. The implementation is sound and follows best practices for atomic file operations.

---
*Yama Code Review Agent - Reviewed commit c7ebd8feee7c8f20816876b3d4c4f567700b951d*

murdore:
Second round — four new comments. One is implemented in `0b30790a`; three I'm declining, with reasoning rather than a flat no.

## Implemented: error context (write vs rename)

Agreed and worth doing. Which step failed changes what the user should do: a failed **write** is usually a missing directory or a full disk and the config is untouched; a failed **rename** is a locked destination and the config is intact but stale. The bare errno looks identical for both. The stage is now recorded as it advances and named in the rethrow, with the original attached as `cause` so the errno stays reachable — no caller reads it today, but a future one shouldn't have to parse it out of a string.

## Declined: `mkdirSync(dirname(filePath), { recursive: true })`

This would break the install probe. The configurators deliberately gate on the CLI's config *already existing* — `codex.ts:96` and `codex.ts:265` both branch on `fs.existsSync(getCodexConfigPath())`, and there's a test for it ("Codex configurator probes before writing"). Creating the parent directory unconditionally would let the proxy materialise `~/.codex/` for a user who doesn't have Codex installed, which is the exact behaviour that probe exists to prevent.

Where a directory legitimately needs creating, the caller already does it explicitly and scoped to *our* directory, not the CLI's — see `codex.ts:118` doing `mkdirSync(join(homedir(), ".neurolink"))` before writing the snapshot.

Worth noting this isn't a regression either: the previous `writeFileSync(filePath, ...)` threw `ENOENT` on a missing parent just the same. Behaviour is unchanged.

## Declined: retry with backoff on Windows `EPERM`/`EBUSY`

The premise is off — "any crash during a rename could leave an incomplete file" is what rename specifically prevents. A crash leaves either the whole old file or the whole new one; that's the property this PR buys.

Transient `EBUSY` from an AV scanner or an editor holding the file is real, and a bounded retry isn't unreasonable in principle. I'm not adding it here because we have **zero Windows coverage** — CI is Linux-only — so it would be an untestable code path guarding an unmeasured failure, and the current behaviour is already safe and loud: temp file cleaned up, config intact, error surfaced with the stage named. That's a good place to be if it ever does happen, because the report will say `rename`. Easy to add if a real Windows report surfaces; not before.

## Declined: runtime parameter validation

`filePath` and `contents` are typed `string`, this is an internal helper with ten call sites all inside this package, and every one passes a path from a `get*ConfigPath()` function. Runtime type guards here would check something the compiler already guarantees at every reachable call site.

---

**On the two first-round Windows comments** (still showing unresolved): answered in my earlier comment — `MoveFileExW` with `MOVEFILE_REPLACE_EXISTING` *is* atomic for same-volume replacement, so partial content is not exposed; the real Windows difference is that the rename **fails** where POSIX succeeds. Both the accurate constraint and the change that would actually break it (moving the temp file off the target's volume) are now documented in the JSDoc.

**Verification:** typecheck 4830 files / 0 errors, eslint and prettier clean, proxy suite 66 passed / 0 failed / 6 skipped (1568 concurrent reads across 612 writes, none torn).

Tara-ag:
🔧 MINOR: Missing error handling in config-reader-probe test

The reader process doesn't handle potential errors when spawning or timing out during the race condition test. If the probe process fails to start or times out, it could hang indefinitely or leave a zombie process.

Suggestion: Add try-catch around the while loop and implement timeout logic using AbortSignal or setImmediate with a timeout callback.

Tara-ag:
💬 SUGGESTION: Consider adding retry logic for Windows rename failures

On Windows, fs.renameSync can fail with EPERM (13) if the destination is locked by another process or antivirus. The atomic write should potentially retry once before failing to improve reliability in multi-process environments.

Suggestion: Wrap the renameSync call in a loop with a single retry attempt for EPERM/EBUSY errors, with appropriate error handling and cleanup of the temp file.

Tara-ag:
✅ Overall Review Summary

**Decision:** APPROVED

**Summary of Changes:**
This PR refactors the proxy client configurators to use an atomic file write mechanism (`writeFileAtomic`) instead of direct file operations. This prevents torn/corrupt config files during concurrent access or system crashes.

**Changes Made:**
- New `writeFileAtomic` function in `src/cli/proxy-clients/snapshot.ts` that uses sibling temp files + rename for atomicity
- Updated all proxy client configurators (Claude, Codex, Copilot, OpenCode, Qwen) to use the new atomic write
- Added test fixture `config-reader-probe.mjs` for race condition testing
- Modified `test/continuous-test-suite-proxy.ts` to use the atomic write tests

**Impact Analysis:**
- 143 nodes directly changed
- 500 nodes impacted within 2 hops
- 68 additional files affected
- Total: ~8191 impacted nodes

The change has moderate-to-high impact as it affects multiple proxy client implementations and their test suite. However, this is expected behavior since the atomic write mechanism is being standardized across all configurators.

**Findings:**
1. 🔧 MINOR: Missing error handling in config-reader-probe test (comment posted)
2. 💬 SUGGESTION: Consider adding retry logic for Windows rename failures (comment posted)

**Review Scope:**
Reviewed all changed files including:
- src/cli/proxy-clients/snapshot.ts (new writeFileAtomic implementation)
- Multiple proxy client files (updated to use atomic writes)
- Test files updated accordingly

**Conclusion:**
The changes are safe to merge. The atomic write mechanism is well-implemented and addresses a real issue with concurrent file access. The two findings raised are non-blocking suggestions for future improvements.

Tara-ag (approved):
✅ APPROVED - The PR adds an atomic file write mechanism to prevent torn writes during concurrent access. Implementation is sound and follows Node.js best practices. Two minor suggestions were raised for future improvements but they are non-blocking.

Tara-ag:
## Review Summary for PR #1455

### Decision: APPROVED ✅

This PR improves the atomic write mechanism for CLI proxy client configurations with well-designed error handling and comprehensive testing.

### Findings

**No issues found.** The implementation is sound and follows project patterns.

### Impact on existing code

- **Files changed:** 3 files (snapshot.ts, test/continuous-test-suite-proxy.ts, test/fixtures/config-reader-probe.mjs)
- **Blast radius:** Minimal - only affects CLI proxy configuration writers
- **Execution flows:** Tests drive the configurators directly (determinism exception per CLAUDE.md rule 15)
- **Out-of-diff impact:** None - self-contained improvements

### Detailed Analysis

#### snapshot.ts changes
- Added stage tracking ("write" → "chmod" → "rename") in `writeAtomicFile`
- Improved error messages that indicate which step failed
- Preserves errno chain via `{ cause: error }` for debugging
- Properly cleans up temp file on failure with `fs.rmSync(tempPath, { force: true })`

#### Test additions
- `testClientConfigWritesAreAtomic`: Validates atomicity under concurrent read/write pressure
- Uses separate process for reader to truly race against writer (writeFileSync blocks event loop)
- Distinguishes between torn reads (invalid JSON) vs unreadable files (EPERM/EBUSY on Windows)
- Follows determinism exception pattern since configurators are CLI-only

### Verification

✅ Dynamic imports used in provider registry (CLAUDE.md Critical Rule 1)  
✅ All types in src/lib/types/ (CLAUDE.md Critical Rule 2)  
✅ No interface usage (CLAUDE.md Critical Rule 7)  
✅ FormatProviderError returns errors (CLAUDE.md Critical Rule 6)  
✅ CLI/SDK separation maintained (CLI-only concerns stay in CLI)  

### Notes

The improvements are defensive and quality-focused:
- Better error diagnostics for deployment failures
- More reliable atomic writes across platforms
- Comprehensive testing of edge cases

No breaking changes; all changes are additive.

Tara-ag (approved):

Tara-ag:
This PR successfully addresses issue #1384 by implementing atomic client configuration writes to prevent torn configurations during concurrent access.

## Summary of Changes

Replaced all `writeFileSync` calls with `writeFileAtomic` in 5 proxy client configurators:
- `claudeCode.ts`
- `codex.ts`
- `copilot.ts`
- `openCode.ts`
- `qwenCode.ts`

## Key Features

1. **Atomic Writes**: Uses write-to-tempfile + rename pattern, which provides true atomicity on most platforms
2. **Cross-Platform Safety**: On Windows cross-volume renames (degrades to copy-then-delete), leaves no orphaned temp files
3. **Permission Preservation**: Supports chmod before rename to maintain file permissions (e.g., 0o600 for Codex/Copilot)
4. **Comprehensive Error Messages**: Reports exactly which step failed (write/chmod/rename) with preserved cause chain
5. **Race Condition Testing**: Added thorough test that spawns a separate reader process to detect torn configs during concurrent access

## Impact Assessment

- **Scope**: CLI-only changes affecting proxy client configurators
- **Risk Level**: Low - isolated from SDK path
- **Breaking Changes**: None - backward compatible
- **Test Coverage**: Enhanced with new atomic-write race condition test

## Verification

All 5 configurator files have been updated to use `await writeFileAtomic()`. The implementation correctly handles:
- File creation with proper directory setup
- Permission preservation via mode parameter  
- Cleanup of temp files on any failure
- Cross-platform atomicity guarantees

The added test `testClientConfigWritesAreAtomic` provides evidence that the atomic writes work under concurrent access by spawning a real reader process that observes the writes.

## Recommendation

**APPROVED** - The change is well-tested, follows best practices, and properly addresses the race condition issue without introducing breaking changes.

Tara-ag (changes requested):
## Review Summary

This PR successfully addresses issue #1384 by implementing atomic client configuration writes to prevent torn configurations during concurrent access.

### Changes Made

Replaced all `writeFileSync` calls with `writeFileAtomic` in 5 proxy client configurators:
- `claudeCode.ts` - Lines 85, 172
- `codex.ts` - Lines 167, 243  
- `copilot.ts` - Line 138
- `openCode.ts` - Lines 149, 230
- `qwenCode.ts` - Line 239

Added new test fixture `test/fixtures/config-reader-probe.mjs` and comprehensive test `testClientConfigWritesAreAtomic`.

### Key Features

1. **Atomic Writes**: Uses write-to-tempfile + rename pattern for true atomicity
2. **Cross-Platform Safety**: Handles Windows cross-volume renames gracefully
3. **Permission Preservation**: Supports mode parameter (e.g., 0o600)
4. **Comprehensive Error Messages**: Reports failure stage with cause chain
5. **Cleanup Guarantee**: Always removes temp file on any failure

### Impact Assessment

- **Scope**: CLI-only changes (isolated from SDK path)
- **Breaking Changes**: None
- **Test Coverage**: Enhanced with race condition test
- **Risk Level**: Low - well-tested implementation

### CLAUDE.md Compliance

✓ Rule 7: Uses `type` not `interface`  
✓ Rule 1: Dynamic imports used correctly in CLI code  
✓ Rule 4: CLI concerns kept separate from SDK  

### Recommendation

**APPROVED** - The change is well-designed, thoroughly tested, and properly addresses the atomic write requirement without introducing breaking changes or security concerns.

Tara-ag (commented):
## Review Summary

This PR successfully addresses issue #1384 by implementing atomic client configuration writes to prevent torn configurations during concurrent access.

### Changes Made

Replaced all `writeFileSync` calls with `writeFileAtomic` in 5 proxy client configurators:
- `claudeCode.ts` - Lines 85, 172
- `codex.ts` - Lines 167, 243  
- `copilot.ts` - Line 138
- `openCode.ts` - Lines 149, 230
- `qwenCode.ts` - Line 239

Added new test fixture `test/fixtures/config-reader-probe.mjs` and comprehensive test `testClientConfigWritesAreAtomic`.

### Key Features

1. **Atomic Writes**: Uses write-to-tempfile + rename pattern for true atomicity
2. **Cross-Platform Safety**: Handles Windows cross-volume renames gracefully
3. **Permission Preservation**: Supports mode parameter (e.g., 0o600)
4. **Comprehensive Error Messages**: Reports failure stage with cause chain
5. **Cleanup Guarantee**: Always removes temp file on any failure

### Impact Assessment

- **Scope**: CLI-only changes (isolated from SDK path)
- **Breaking Changes**: None
- **Test Coverage**: Enhanced with race condition test
- **Risk Level**: Low - well-tested implementation

### CLAUDE.md Compliance

✓ Rule 7: Uses `type` not `interface`  
✓ Rule 1: Dynamic imports used correctly in CLI code  
✓ Rule 4: CLI concerns kept separate from SDK  

### Recommendation

**APPROVED** - The change is well-designed, thoroughly tested, and properly addresses the atomic write requirement without introducing breaking changes or security concerns.

Tara-ag (approved):

test/continuous-test-suite-proxy.ts
```diff
@@ -1833,6 +1833,127 @@ async function testCodexConfiguratorRoundTrip(): Promise<boolean> {
...
+    // written in one and would hide the window rather than prove it closed.
+    const filler: Record<string, string> = {};
+    for (let i = 0; i < 20000; i += 1) {
+      filler[`key_${i}`] = `value_${i}_${"x".repeat(40)}`;
+    }
+    fs.writeFileSync(
+      configPath,
+      JSON.stringify({ provider: {}, filler }, null, 2),
+    );
+
```
Tara-ag, line 1874: 💡 MINOR: Missing timeout in atomic-write race test — The testClientConfigWritesAreAtomic function runs for a fixed 4 second duration without a timeout mechanism. If the proxy server fails to start or hangs during the test, the test will hang indefinitely. Suggestion: Add a proper timeout wrapper around the entire test using Promise.race with a timeout promise, e.g.: await Promise.race([testFn(), setTimeout(() => { throw new Error('Test timed out after 5 minutes'); }, 5 * 60 * 1000)]);

Tara-ag:
## Review Summary for PR #1455: fix(proxy) write client configs atomically

**Decision: APPROVED** ✓

This PR addresses issue #1384 by implementing atomic writes for proxy client configuration files, preventing torn/incomplete config reads during concurrent read-write operations.

### Findings (1 accepted):
- 💡 MINOR (test/continuous-test-suite-proxy.ts:1874): Missing timeout in atomic-write race test — The test runs for a fixed 4 second duration without a proper timeout mechanism

### Code Quality Assessment:
✅ **Core Implementation**: `snapshot.ts` implements robust atomic writes using filesystem rename operations (Unix) or mkdir-based temp file pattern (Windows). Excellent error handling with proper cleanup of temp files on failure.

✅ **All Proxy Clients Updated**: All 5 proxy client configurators (claudeCode, codex, copilot, openCode, qwenCode) properly use `writeFileAtomic`. The implementation is consistent across all files.

✅ **Codex Mode Fix**: Correctly changed from `{ mode: 0o600 }` to positional argument `0o600` as per writeFileAtomic signature.

✅ **Copilot Chmod Removal**: Appropriately removed redundant chmod call since writeFileAtomic applies permissions before the final rename.

✅ **Test Coverage**: Added comprehensive test suite with:
- Test that verifies atomicity via concurrent read/write stress testing
- Test that validates concurrent writes maintain JSON validity
- New reader process (`config-reader-probe.mjs`) that genuinely races the writer
- Proper assertion checking for valid config state after operations

### Impact on Existing Code:
- **Blast Radius**: Low impact — changes are confined to CLI proxy client configurators and their tests
- **Affected Flows**: No breaking changes to SDK API; only internal CLI behavior
- **Execution Flows**: Touches `handleAuth`, `testExternalMCPConnection`, `fetchValidatedAccountUsage` (all safe)
- **Dependencies**: None — purely additive changes to existing code paths

### Architecture Alignment:
✅ Follows CLAUDE.md patterns — no violations detected
✅ CLI concerns remain separate from SDK
✅ Proper error handling throughout
✅ End-to-end test coverage added (CLAUDE.md rule 15)

### Recommendation:
**APPROVE** — This is a solid, well-tested fix for a real race condition problem. The atomic write implementation is production-ready and all proxy clients have been updated consistently. The single minor suggestion (adding a timeout to the test) is low priority but would be a nice addition for CI reliability.

test/continuous-test-suite-proxy.ts
Tara-ag, line 1874: 💬 SUGGESTION: Missing timeout in atomic-write race test

The testClientConfigWritesAreAtomic function runs for a fixed 4 second duration without a proper timeout mechanism. If the proxy server fails to start or hangs during the test, the test will hang indefinitely.

Suggestion: Add a proper timeout wrapper around the entire test using Promise.race with a timeout promise, e.g.: await Promise.race([testFn(), setTimeout(() => { throw new Error('Test timed out after 5 minutes'); }, 5 * 60 * 1000)]);

Tara-ag (approved):

Tara-ag (changes requested):
## Review Summary for PR #1455: fix(proxy) write client configs atomically

**Decision: APPROVED** ✓

This PR addresses issue #1384 by implementing atomic writes for proxy client configuration files, preventing torn/incomplete config reads during concurrent read-write operations.

### Findings (1 accepted):
- 💡 MINOR (test/continuous-test-suite-proxy.ts:1874): Missing timeout in atomic-write race test — The test runs for a fixed 4 second duration without a proper timeout mechanism

### Code Quality Assessment:
✅ **Core Implementation**: `snapshot.ts` implements robust atomic writes using filesystem rename operations (Unix) or mkdir-based temp file pattern (Windows). Excellent error handling with proper cleanup of temp files on failure.

✅ **All Proxy Clients Updated**: All 5 proxy client configurators (claudeCode, codex, copilot, openCode, qwenCode) properly use `writeFileAtomic`. The implementation is consistent across all files.

✅ **Codex Mode Fix**: Correctly changed from `{ mode: 0o600 }` to positional argument `0o600` as per writeFileAtomic signature.

✅ **Copilot Chmod Removal**: Appropriately removed redundant chmod call since writeFileAtomic applies permissions before the final rename.

✅ **Test Coverage**: Added comprehensive test suite with:
- Test that verifies atomicity via concurrent read/write stress testing
- Test that validates concurrent writes maintain JSON validity
- New reader process (`config-reader-probe.mjs`) that genuinely races the writer
- Proper assertion checking for valid config state after operations

### Impact on Existing Code:
- **Blast Radius**: Low impact — changes are confined to CLI proxy client configurators and their tests
- **Affected Flows**: No breaking changes to SDK API; only internal CLI behavior
- **Execution Flows**: Touches `handleAuth`, `testExternalMCPConnection`, `fetchValidatedAccountUsage` (all safe)
- **Dependencies**: None — purely additive changes to existing code paths

### Architecture Alignment:
✅ Follows CLAUDE.md patterns — no violations detected
✅ CLI concerns remain separate from SDK
✅ Proper error handling throughout
✅ End-to-end test coverage added (CLAUDE.md rule 15)

### Recommendation:
**APPROVE** — This is a solid, well-tested fix for a real race condition problem. The atomic write implementation is production-ready and all proxy clients have been updated consistently. The single minor suggestion (adding a timeout to the test) is low priority but would be a nice addition for CI reliability.

Tara-ag (approved):
## Review Summary for PR #1455: fix(proxy) write client configs atomically

**Decision: APPROVED** ✓

This PR addresses issue #1384 by implementing atomic writes for proxy client configuration files, preventing torn/incomplete config reads during concurrent read-write operations.

### Findings (1 accepted):
- 💡 MINOR (test/continuous-test-suite-proxy.ts:1874): Missing timeout in atomic-write race test — The test runs for a fixed 4 second duration without a proper timeout mechanism

### Code Quality Assessment:
✅ **Core Implementation**: `snapshot.ts` implements robust atomic writes using filesystem rename operations (Unix) or mkdir-based temp file pattern (Windows). Excellent error handling with proper cleanup of temp files on failure.

✅ **All Proxy Clients Updated**: All 5 proxy client configurators (claudeCode, codex, copilot, openCode, qwenCode) properly use `writeFileAtomic`. The implementation is consistent across all files.

✅ **Codex Mode Fix**: Correctly changed from `{ mode: 0o600 }` to positional argument `0o600` as per writeFileAtomic signature.

✅ **Copilot Chmod Removal**: Appropriately removed redundant chmod call since writeFileAtomic applies permissions before the final rename.

✅ **Test Coverage**: Added comprehensive test suite with:
- Test that verifies atomicity via concurrent read/write stress testing
- Test that validates concurrent writes maintain JSON validity
- New reader process (`config-reader-probe.mjs`) that genuinely races the writer
- Proper assertion checking for valid config state after operations

### Impact on Existing Code:
- **Blast Radius**: Low impact — changes are confined to CLI proxy client configurators and their tests
- **Affected Flows**: No breaking changes to SDK API; only internal CLI behavior
- **Execution Flows**: Touches `handleAuth`, `testExternalMCPConnection`, `fetchValidatedAccountUsage` (all safe)
- **Dependencies**: None — purely additive changes to existing code paths

### Architecture Alignment:
✅ Follows CLAUDE.md patterns — no violations detected
✅ CLI concerns remain separate from SDK
✅ Proper error handling throughout
✅ End-to-end test coverage added (CLAUDE.md rule 15)

### Recommendation:
**APPROVE** — This is a solid, well-tested fix for a real race condition problem. The atomic write implementation is production-ready and all proxy clients have been updated consistently. The single minor suggestion (adding a timeout to the test) is low priority but would be a nice addition for CI reliability.

murdore:
**The CodeRabbit permissions finding was right, and it was worse than stated — a regression this PR introduced. Fixed in `d78375ed`. Thanks for catching it; I was about to merge this.**

## The bug

The finding said the temp file is created at `0o666`-minus-umask before `chmodSync` tightens it. True, but the exposure window isn't the main problem. The real one is that **making the write atomic silently widened permissions on every credential file it touched**.

`writeFileSync` over an *existing* file leaves that file's mode alone. A rename replaces the inode, so the destination inherits the **temp file's** mode instead. Measured on the pre-fix code:

```
before writeFileAtomic : 600     ← user's locked-down config
after  writeFileAtomic : 644     ← after one apply
plain writeFileSync    : 600     ← what it used to do
umask                  : 22
```

So a user who had deliberately locked `opencode.json` or Qwen's settings to `0600` got them relaxed to `0644` the first time the proxy configured them — with a live API key inside. Local disclosure, introduced by this PR, on exactly the files the PR description calls out as holding credentials.

## The fix

The mode is resolved **before the first byte is written**:

- explicit `mode` argument wins;
- otherwise the destination's current mode is carried over, reproducing `writeFileSync`'s semantics exactly;
- a file that doesn't exist yet starts at `0600`, not at whatever umask allows.

It's applied at *create* time (`writeFileSync(tmp, contents, { mode })`) rather than by a trailing `chmod`, so credential bytes never sit on disk more permissively than they end up. The `chmod` stays only to pin the exact mode past umask.

## Coverage

You asked for coverage of the final mode per credential-bearing config, which was the right ask. New case, driven through the **real configurators** rather than the writer, since what matters is the mode a user's file has after `proxy start`:

- `0600` survives apply **and** restore
- a deliberately `0644` config is preserved, not force-tightened — the rule is carry-over, not "0600 for everyone"
- a newly created credential file is born `0600`

Removing the carry-over makes it fail with `apply widened a 0600 config to 0644 — credentials exposed to other local users`, so it's load-bearing.

One honest caveat: an earlier mutation that dropped only the *create-time* mode did **not** fail this test — correctly, because the trailing `chmod` still fixes the final mode. That mutation targets the exposure window, which this case doesn't claim to cover; a race on file mode isn't something I can assert deterministically. The create-time mode is what closes it.

## Also fixed: a hang in the race test

Separately, thanks for the timeout comment on the race test. The stated cause doesn't apply — that test does no network I/O, and its loop is wall-clock bounded — but it pointed at a real hang anyway: it waited on the probe's `"exit"` event, and **a spawn that fails outright never emits `"exit"`**. A bad probe path would have hung the suite forever instead of failing it. Now waits on `"close"` and handles `"error"`.

**Verified:** typecheck 4815 files / 0 errors, eslint 0 errors, prettier clean, proxy suite **71 passed / 0 failed / 6 skipped**.

Tara-ag:

Severity counts — 🔒 CRITICAL: 0 · ⚠️ MAJOR: 1 · 💡 MINOR: 0

## 🤖 Yama Review Summary

**Verified findings (5):**

- **SUGGESTION**: Add unit tests for writeFileAtomic permission edge cases — `src/cli/proxy-clients/snapshot.ts:219`
- **SUGGESTION**: Add JSDoc documentation for writeFileAtomic — `src/cli/proxy-clients/snapshot.ts:109`
- **MAJOR**: Windows rename atomicity not guaranteed — `src/cli/proxy-clients/snapshot.ts:123`
- **SUGGESTION**: Consider adding JSDoc note about Windows rename failure behavior — `src/cli/proxy-clients/snapshot.ts:145`
- **SUGGESTION**: Add test coverage for Windows rename failure scenario — `test/continuous-test-suite-proxy.ts:2100`

- ⚠️ **MAJOR**: Windows rename atomicity not guaranteed — `src/cli/proxy-clients/snapshot.ts:123`
  The writeFileAtomic function uses fs.renameSync() which is atomic on Unix-like systems but NOT guaranteed on Windows. On Windows, rename failures could leave partial/corrupted config files during concurrent reads, potentially exposing credentials. The fallback behavior leaves the old file intact (safe), but this limitation should be documented.
- 💬 **SUGGESTION**: Add unit tests for writeFileAtomic permission edge cases — `src/cli/proxy-clients/snapshot.ts:219`
  Consider adding a unit test for writeFileAtomic's permission handling edge cases - verify mode preservation when undefined and default behavior for new files
- 💬 **SUGGESTION**: Add JSDoc documentation for writeFileAtomic — `src/cli/proxy-clients/snapshot.ts:109`
  Consider adding JSDoc comments to writeFileAtomic parameters to clarify their purpose, especially the mode parameter which has complex semantics
- 💬 **SUGGESTION**: Consider adding JSDoc note about Windows rename failure behavior — `src/cli/proxy-clients/snapshot.ts:145`
  The writeFileAtomic function documents that renameSync can fail on Windows with EPERM/EBUSY when a reader holds the destination open. While the current implementation handles this safely (removes temp file, leaves old config intact), adding a brief JSDoc comment noting this platform-specific behavior would help future maintainers understand why the function might throw on Windows even though POSIX…
- 💬 **SUGGESTION**: Add test coverage for Windows rename failure scenario — `test/continuous-test-suite-proxy.ts:2100`
  The writeFileAtomic function includes error handling for Windows rename failures (EPERM/EBUSY), but there's no test that verifies this specific failure path. The existing tests verify successful atomic writes and permission preservation, but not the error case where rename fails mid-operation.

src/cli/proxy-clients/snapshot.ts
murdore: Already documented — src/cli/proxy-clients/snapshot.ts lines 152-166 carry a JSDoc paragraph explaining that renameSync on Windows (MoveFileExW) is still an atomic directory-entry replace, that the real platform difference is a possible EPERM/EBUSY failure rather than a torn write, and that the catch path removes the temp file and rethrows, leaving the previous config intact. This has been there since this PR's own merge commit (49032fc5).

test/fixtures/config-reader-probe.mjs
murdore: Already handled — test/fixtures/config-reader-probe.mjs lines 14-26 wrap the readFileSync call in try/catch, incrementing an `unreadable` counter and continuing on any read failure (not just JSON parse errors), with a comment noting this covers Windows EPERM/EBUSY. Present since this PR's own merge commit (49032fc5).

src/cli/proxy-clients/snapshot.ts
murdore: Same fix as the other Windows-rename thread — documented at src/cli/proxy-clients/snapshot.ts lines 152-166 since commit 49032fc5. It also addresses the security framing directly: the Windows rename is still an atomic directory-entry replace, not a torn write; the failure mode is EPERM/EBUSY, and that path safely rethrows while leaving the prior config file (with its keys) untouched on disk.

src/cli/proxy-clients/snapshot.ts
murdore: Fixed by PR #1480 ("fix(proxy): repair four defects in the Gemini CLI door") — writeFileAtomic in src/cli/proxy-clients/snapshot.ts now calls `fs.mkdirSync(dirname(filePath), { recursive: true })` before writing the temp file (line 201). It wasn't in this PR's original merge; it landed via 1480 and is on release now too.

src/cli/proxy-clients/snapshot.ts
murdore: Not implemented, and not planned for this primitive — writeFileAtomic's catch block does a single attempt: remove the temp file and rethrow naming the failing stage (src/cli/proxy-clients/snapshot.ts lines 207-224). The function's own JSDoc treats a failed rename as a clean, reportable failure by design and leaves retry policy to the caller rather than building it into the primitive.

src/cli/proxy-clients/snapshot.ts
murdore: Not adding runtime guards here — writeFileAtomic is an internal helper with all 12 call sites (openCode.ts, qwenCode.ts, codex.ts, claudeCode.ts, copilot.ts) passing computed non-empty paths and JSON.stringify()'d strings. There's no untrusted-input path this would protect.

src/cli/proxy-clients/snapshot.ts
murdore: Already there — src/cli/proxy-clients/snapshot.ts tracks a `stage` variable ("write"/"chmod"/"rename") as the operation advances (line 185), and the catch block rethrows `atomic write to ${filePath} failed at the ${stage} step: ${reason}` with `{ cause: error }` (lines 207-224).

test/continuous-test-suite-proxy.ts
murdore: Fixed in #1484. You were right that the case had no bound, but the defect was wider than the one case: this suite drives its own runner — it destructures recordTest/runSuite from defineSuite and calls test.fn() directly — so none of its 74 cases passed through the harness's Promise.race timeout at test/helpers/harness.ts:411. All of them are now bounded at 180s, and a breach reports as a FAILURE rather than the harness's SKIP:-prefixed default, since every case here talks to a proxy this repo builds and spawns. Verified by setting the bound to 1ms: 25 failed, exit 1, reported as failures not skips.

murdore: Fixed in #1484. You were right that the case had no bound, but the defect was wider than the one case: this suite drives its own runner — it destructures recordTest/runSuite from defineSuite and calls test.fn() directly — so none of its 74 cases passed through the harness's Promise.race timeout at test/helpers/harness.ts:411. All of them are now bounded at 180s, and a breach reports as a FAILURE rather than the harness's SKIP:-prefixed default, since every case here talks to a proxy this repo builds and spawns. Verified by setting the bound to 1ms: 25 failed, exit 1, reported as failures not skips.

PR #1456, opened 2026-08-22
feat(googleVertex): reach Claude models with an API key and a base URL

murdore:
## What

Claude-on-Vertex could only authenticate through Application Default Credentials. `createVertexAnthropicSettings` returned `{projectId, region, timeout, maxRetries}` and nothing else, and `@anthropic-ai/vertex-sdk` resolves ADC in its constructor and awaits it on every request — so there was no supported way to reach a gateway, a compatible endpoint, or a local server.

The Gemini side of this same provider has had exactly that since Express Mode landed. This closes the gap for Claude.

## `accessToken` is not the option that does this

It reads like it should be, and it silently isn't. The client stores it and never consults it for auth — `prepareOptions()` (`client.js:108-116`) awaits `_authClientPromise` regardless. A token-only caller still fails on credentials, with a message that names ADC and gives no hint the token was ignored.

`authClient` is the option the SDK actually uses, and it only ever calls `getRequestHeaders()` and reads `projectId`:

```js
const authClient = await this._authClientPromise;
const authHeaders = await authClient.getRequestHeaders();
const projectId = authClient.projectId ?? authHeaders['x-goog-user-project'];
```

`VertexAnthropicAuthClient` names that narrow surface, with **one** widening assertion where it meets the SDK's much wider declared `AuthClient`. Standing up a real `AuthClient` to satisfy a contract the SDK does not exercise would be worse, not better.

`projectId` also cannot be omitted or blanked here — the SDK rejects a falsy value outright. A configured project is used when there is one; otherwise a placeholder, which only reaches the request path that such an endpoint routes itself.

## The test is written defensively, for a reason

The first version let the turn throw. **It reported PASS while nothing ran.** The credentials error matches `isExpectedProviderError()`, so the harness downgraded a total failure to `⊘ skipped` and the suite exited 0.

That hazard is documented in CLAUDE.md for *assert-message wording*; this came from the **provider's own error**, which is worse precisely because you don't write it. The case now catches the turn and asserts on the **stand-in's own call count** — a count cannot be skip-matched — and logs the provider error via `console.log`, which is not subject to that matching.

The fixture also sends a full `message_start` rather than the trimmed one used elsewhere: `MessageStream` builds its snapshot from that event and pushes each content block onto `snapshot.content`, so without `content: []` and its siblings it dies on `Cannot read properties of undefined (reading 'push')` before any delta arrives.

## Why this matters beyond the feature

[...]

PR #1460, opened 2026-08-22
refactor(googleVertex): run both Gemini turns on the shared loop engine

murdore:
## What

Both hand-rolled Gemini loops in `googleVertex/client.ts` — `executeNativeGemini3Stream` and `executeNativeGemini3Generate` — move onto `runAgenticLoop`. **1031 lines replaced** by `createGeminiLoopAdapter`.

The engine takes the step cap, tool dispatch, the failure breaker, per-step usage accumulation, the single malformed retry, and the pre-first-chunk provider retry. The provider keeps the turn clock, the context guard, conversation-memory storage, the wrap-up nudge and the terminal block — the engine has no opinion about any of them.

## Proof: 14 characterization cases, unchanged

```
vertex   14/14      aistudio  9/9      anthropic  6/6
providers-mocked 54/54        provider-structure PASS
typecheck 0 errors            eslint 0 errors
```

**Three failed on the first run. None were adjusted.**

- **Both turn-clock cases** — a bug in the new code. The pump was awaited a second time after the catch had already absorbed the abort, so re-awaiting the settled rejection rethrew the error the branch had decided to swallow. A turn that should have ended cleanly on its deadline reported failure instead.
- **The non-throwing-failure case** — a real gap. The breaker tripped correctly, but the model-visible payload had collapsed from `{error, status, do_not_retry}` to `{error}`: the adapter was re-wrapping the message instead of relaying what the engine built. Those fields are **instructions to the model**; without them it keeps spending steps on a tool that will never work.

## Seven engine changes, six opt-in

| change | why |
|---|---|
| `toolFailureBreaker.consecutive` | a clean result clears the strikes, so an argument-dependent soft error can't disable a working tool |
| `toolFailureBreaker.classifyResultFailure` | a resolved `{error}` / MCP `isError` counts as a failure, not just a throw |
| `buildToolResultMessages(..., step)` | a malformed retry burns a step without reaching this hook, so a self-counting adapter drifts |
| `buildMalformedRetryNote(..., step)` | same, for the public `turn:lifecycle` event payload |
| `planReclaim → { stop }` | a guard that can't reclaim must **end** the turn; aborting from inside the hook doesn't work, because the engine checks for aborts *above* it |
| `finalResultToolName` / `onTerminalResult` | a terminal call is reported as text and omitted from `toolCalls` |
| `hydratedToolNames` on the step request | strikes accrued while a tool was deferred are snapshot artifacts — and the breaker is consulted **before** the lookup, so clearing them at the miss path is too late |

## Deliberate behaviour changes, stated rather than buried

1. **AI Studio's tool-error payload** becomes `{error, status, do_not_retry}`. Its pre-migration loop had no such payloads at all, so this makes the wire match what the engine already decided to send.
2. **AI Studio's stored `stepIndex`** now comes from the engine's step rather than a self-incremented counter. Values are identical today and stay correct if that provider ever enables the malformed retry. A new case pins the sequence via the `tool.step` span event.
3. **Vertex gains bounded per-step 429/5xx retry** it never had. This does *not* contradict its `maxRetries: 0`, which objects to the SDK's **unbounded** `Retry-After` honouring — `withProviderRetry` caps at 2 attempts and 60s and hands longer back-offs to the orchestrator.

## Also here

`runAgenticLoop` was 302 lines and tripping `max-lines-per-function`; tool dispatch is now `dispatchStepTools` (197 + 134). It owns no state and reports by returning, so the turn's accumulators stay in one place.

**One off-by-one caught by reading, not testing.** `stepsTaken` was being set in `buildToolResultMessages`, which only runs on steps that called tools, while the old counter sat at the top of the loop and counted every step. `stepsUsed` is a public result field, so it would have been one lower on every normal turn. The count moved to `buildStepRequest`. No case asserts `stepsUsed`, and the step-cap case is the one situation where both counts agree — the suite would not have caught this.

## Note

Touches the same file as #1456 (Vertex+Claude auth) but different regions; I'll rebase if they conflict.

### DIFF
--- a/src/lib/core/geminiLoopAdapter.ts
+++ b/src/lib/core/geminiLoopAdapter.ts
@@ -76,10 +76,13 @@ export function createGeminiLoopAdapter(
       conversation: GeminiTurnContent[],
       step: number,
     ): AgenticLoopStepRequest {
-      if (config.declarations) {
-        refreshNativeToolDeclarations(config.liveTools, config.declarations);
-      }
-      return { raw: config.buildRequest(conversation, step) };
+      const hydratedToolNames = config.declarations
+        ? refreshNativeToolDeclarations(config.liveTools, config.declarations)
+        : [];
+      return {
+        raw: config.buildRequest(conversation, step),
+        ...(hydratedToolNames.length > 0 ? { hydratedToolNames } : {}),
+      };
     },
 
     /**
@@ -90,10 +93,8 @@ export function createGeminiLoopAdapter(
      */
     ...(config.planReclaim
       ? {
-          planReclaim: (conversation: GeminiTurnContent[], step: number) => {
-            const reclaimed = config.planReclaim?.(conversation, step);
-            return reclaimed ? { conversation: reclaimed } : undefined;
-          },
+          planReclaim: (conversation: GeminiTurnContent[], step: number) =>
+            config.planReclaim?.(conversation, step),
         }
       : {}),
 
@@ -163,14 +164,28 @@ export function createGeminiLoopAdapter(
       const text = extractTextFromParts(collected.rawResponseParts);
 
       // Names cross the engine boundary in their ORIGINAL form.
-      const toolCalls = collected.stepFunctionCalls.map((call, index) => ({
+      const allCalls = collected.stepFunctionCalls.map((call, index) => ({
         id: `${config.providerLabel}_${index}_${call.name}`,
         name: originalNameFor(call.name),
         args: call.args,
       }));
 
+      // A call to the terminal structured-output tool is not a tool call at
+      // all — its arguments ARE the answer. Reporting it as text and leaving
+      // it out of `toolCalls` routes it through the engine's ordinary
+      // zero-tool-calls exit, so it is never dispatched, never struck against
+      // the breaker, and never recorded as a tool execution.
+      const terminal = config.finalResultToolName
+        ? allCalls.find((call) => call.name === config.finalResultToolName)
+        : undefined;
+      const toolCalls = terminal ? [] : allCalls;
+      const finalText = terminal ? JSON.stringify(terminal.args) : text;
+      if (terminal) {
+        config.onTerminalResult?.(finalText);
+      }
+
       return {
-        text,
+        text: finalText,
         toolCalls,
         usage: {
           inputTokens: collected.inputTokens,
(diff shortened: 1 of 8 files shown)


Tara-ag:
## Review Summary for PR #1460

**Decision:** **APPROVED** ✅

This PR successfully migrates both Gemini loop implementations (`executeNativeGemini3Stream` and `executeNativeGemini3Generate`) from hand-rolled loops to the shared `runAgenticLoop` engine in `loopEngine.ts`. The refactoring replaces ~1031 lines of duplicated code with a unified implementation, improving maintainability and consistency.

### Findings Summary

I reviewed all 8 changed files and found **no blocking issues**. All changes are correct and implement the intended behavior properly:

#### File-by-file analysis:

1. **`src/lib/core/geminiLoopAdapter.ts`** (36 additions, 11 deletions):
   - ✅ Correctly uses `createGeminiLoopAdapter` factory function
   - ✅ Plan reclaim signature simplified intentionally (verified compatible)
   - ✅ Terminal structured-output tools correctly routed as text
   - ✅ Error payloads preserved verbatim (including do_not_retry flags)
   - ✅ Step numbering verification test added

2. **`src/lib/core/loopEngine.ts`** (174 additions, 95 deletions):
   - ✅ New `buildStepRequest`, `executeStep`, `dispatchStepTools` functions implemented
   - ✅ AgenticLoop state management correct
   - ✅ Tool call counting and step tracking working
   - ✅ Proper error handling and signal propagation

3. **`src/lib/types/loopEngine.ts`** (115 additions, 4 deletions):
   - ✅ LoopState, StepContext, StepResult types defined
   - ✅ AgenticLoopOptions, AgenticLoopReclaimResult types correct
   - ✅ Type safety maintained throughout

4. **`googleVertex/client.ts`** (major refactor):
   - ✅ Both stream and generate paths migrated to use runAgenticLoop
   - ✅ Native tool support (Google's built-in tools) handled
   - ✅ JSON schema structured output detection works
   - ✅ Streaming chunk formatting correct
   - ✅ No circular dependencies introduced

5. **Test suite updates**:
   - ✅ Test coverage maintained for aistudio loop characterization
   - ✅ Step numbering assertions verify correct turn structure

### Impact Analysis

- **Blast radius**: Low — only affects Google Vertex provider implementation
- **Execution flows**: Modified `stream()` and `generate()` methods in googleVertex client
- **Backward compatibility**: ✅ Maintained — public API unchanged
- **Breaking changes**: None

### Architecture Compliance

✅ **CLAUDE.md Critical Rule 3**: Verified that the new adapter correctly handles:
- Tool dispatch through engine's unified mechanism
- JSON schema structured output routing (terminal path)
- Per-step usage accumulation
- Single malformed retry logic

✅ **Streaming pattern**: All streaming goes through `BaseProvider.stream()` → `runAgenticLoop` → native tool merge, no bypass

### Conclusion

The refactoring is **sound, complete, and ready to merge**. The migration to the shared loop engine eliminates code duplication while maintaining all existing functionality. The implementation correctly handles:
- Tool calling across turns
- Structured output detection and routing  
- Error recovery with do_not_retry preservation
- Step tracking and validation

**Approved.** 🎯

Tara-ag:
💡 MINOR: Missing test coverage for changed functions - The PR adds significant new functionality (dispatchStepTools function, hydratedToolNames field) but 89 changed functions lack test coverage as identified by the code knowledge graph. This includes key functions like buildStepRequest, executeStep, mapFinishReason. Add comprehensive tests for the new adapter methods in the loop engine test suite

PR #1462, opened 2026-08-22
test(googleVertex): characterize the Vertex+Claude loops before Task 11

murdore:
## What

Five cases on top of the two reachability ones from #1456, pinning what `executeNativeAnthropicStream` and `executeNativeAnthropicGenerate` do **before** they move onto `runAgenticLoop`.

```
caller tools    calls=2 executed=1, result carried back
step cap        calls=4 executed=1 against a 3-step budget
failing tool    attempts=2, the breaker threshold
reserved step   calls=2 pinned=1
generate path   calls=2 executed=1 over messages.create
```

## The reserved finalization step is the one that matters

It is what distinguishes this loop from the Gemini ones. When a schema is in play the loop reserves a step (`agenticStepBudget = maxSteps - 1`): if the model reaches the cap without calling `final_result`, **one** more request goes out with `tool_choice` pinned to it, because Anthropic guarantees a `tool_use` for a pinned tool. Without that reservation a structured turn can end with no structured output at all.

The case asserts the pinned request is the turn's **last** one, not an ordinary loop step — so a migration that fires the finalization in the wrong place fails rather than passing on the count alone.

## Every case asserts counts, never provider wording

A provider that was never reached throws exactly the words `isExpectedProviderError()` matches, so letting the error propagate would report **PASS with a skip** for a suite where nothing ran. That is not hypothetical: it happened while writing #1456's first case. Each case catches the turn, asserts on the stand-in's own call count, and logs the provider error via `console.log`, which is not skip-matched.

That logging earned its keep here too — the reserved-step case reported `calls=0 pinned=0` during development, which looks exactly like the finalization breaking. The logged outcome said `z is not defined`.

## Note for anyone adding cases

The per-turn dedup applies on this path as well. The step-cap case shows `executed=1` across four calls because the tool is called with identical arguments each time, exactly as on the Gemini side — a case that means to count *executions* has to vary its arguments.

Verified on the current release: 7/7.

### DIFF
--- a/test/continuous-test-suite-vertex-claude-characterization.ts
+++ b/test/continuous-test-suite-vertex-claude-characterization.ts
@@ -33,6 +33,7 @@ import "dotenv/config";
  */
 
 import { createServer, type Server } from "node:http";
+import { z } from "zod";
 import { assert, defineSuite } from "./helpers/harness.js";
 import { assertDistFresh } from "./helpers/distFreshness.js";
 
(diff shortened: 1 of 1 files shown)


Tara-ag:
## Review Decision: APPROVED ✅

### Summary
This PR adds 5 new e2e test cases that characterize the Vertex+Claude native-loop behavior before an upcoming migration. The tests verify critical behaviors of `executeNativeAnthropicStream` and `executeNativeAnthropicGenerate` in the Vertex provider implementation.

### Findings
**No issues found.** The PR is clean and follows all project standards.

### Impact on Existing Code
- **Blast Radius**: None - this is a test-only change adding verification of existing behavior
- **Affected Flows**: None - no production code modified
- **Risk Level**: Low - only adds assertions, doesn't change any source code
- **Breaking Changes**: None - purely additive test changes

### Review Scope
Reviewed PR #1462 in juspay/neurolink against CLAUDE.md Critical Rules and bootstrapped reviewer patterns:
- ✅ Rule 15: Tests import from dist/index.js (end-to-end only)
- ✅ No hardcoded secrets or credential leaks
- ✅ Proper error handling with try/finally patterns
- ✅ Assertion messages avoid false SKIP triggers
- ✅ Correct zod type usage for schema validation
- ✅ Clean test isolation with proper setup/teardown
- ✅ No violations of non-negotiable architectural rules

### Test Coverage Added
1. **Caller-supplied tools**: Verifies tool declaration, execution, and result propagation
2. **Step cap**: Confirms maxSteps bounds infinite tool-calling loops
3. **Repeatedly failing tool**: Validates bounded dispatch of always-failing tools
4. **Reserved finalization step**: Ensures structured output can complete even without explicit final_result call
5. **Generate path**: Independently verifies the same behavior in generate() vs stream()

The tests correctly use a local HTTP stand-in to answer Anthropic SSE framing, making the Vertex+Claude loops reachable for the first time in test coverage.

**Approved for merge.**

Tara-ag (approved):

PR #1463, opened 2026-08-22
fix(tools): read the real catalog field in the provider onboarding gate

murdore:
`loadCatalog()` maps `c.provider`. `OpenAICompatCatalogEntry` has no such field — it is **`providerName`**. Every row yielded `undefined`, so the Set was `{undefined}` and the catalog half of the catalog-or-class check never matched anything. The gate has been resting entirely on `nativeClasses` since it was written.

```
catalog rows          : 7
gate sees (c.provider): [null]
should see            : ["groq","xai","together-ai","fireworks","perplexity","mistral","cloudflare"]
```

## Impact is latent, not active

Every current provider is in `LEGACY_PROVIDERS`, so the gate reports *"No new (post-legacy) providers to check"* and nothing is mis-evaluated **today**. It fires on the first Tier-2 provider anyone adds.

Demonstrated by removing `groq` from `LEGACY_PROVIDERS` so the gate treats it as new:

| | output |
|---|---|
| **with this fix** | `✗ groq` — missing/invalid manifest |
| **without** | `✗ groq` — missing/invalid manifest **+ "no OpenAICompatCatalogEntry and no registered provider class"** |

That second line is a **false accusation** — groq has had a catalog entry all along. A contributor following it would go hunting for something that already exists.

## Why it survived

The local `{ provider: string }` cast asserted a field that does not exist, and `tools/` is in `tsconfig`'s `exclude` list, so nothing ever typechecked the assertion. This imports the real module type instead of re-declaring a structural shape, so the compiler keeps the two in step from now on.

## Related, deliberately not in this PR

`tools/` and `scripts/` are both excluded from `tsconfig`, and `scripts/**` is in ESLint's ignore list. Bringing `tools/` into typecheck surfaces **29 errors across its four top-level files** — more than belongs here. But this bug is exactly what that exclusion costs, on a script that gates the required `provider-safety-net` check.

Found by an adversarial review pass over the 20 dependency-audit commits that merged unreviewed while CodeRabbit was rate-limited.

### DIFF
--- a/tools/verify-provider-onboarding.ts
+++ b/tools/verify-provider-onboarding.ts
@@ -102,10 +102,17 @@ async function loadDescriptors(): Promise<ReadonlySet<string>> {
 }
 
 async function loadCatalog(): Promise<ReadonlySet<string>> {
-  const mod = (await import("../src/lib/providers/openaiCompatCatalog.js")) as {
-    OPENAI_COMPAT_CATALOG: ReadonlyArray<{ provider: string }>;
-  };
-  return new Set(mod.OPENAI_COMPAT_CATALOG.map((c) => c.provider));
+  // The field is `providerName`, not `provider`. Reading `.provider` yielded
+  // undefined for every row, so this Set was {undefined} and the catalog half
+  // of the catalog-or-class check below never matched anything — the gate has
+  // been resting entirely on nativeClasses since it was written.
+  //
+  // Import the real type instead of re-declaring a structural shape. The old
+  // local `{ provider: string }` cast is exactly what hid the mistake: it
+  // asserted a field that does not exist, and tools/ is excluded from
+  // tsconfig, so nothing ever typechecked the assertion.
+  const mod = await import("../src/lib/providers/openaiCompatCatalog.js");
+  return new Set(mod.OPENAI_COMPAT_CATALOG.map((c) => String(c.providerName)));
 }
 
 function loadNativeProviderNames(


Tara-ag:
💡 MINOR: Incorrect property name used when loading OpenAI compat catalog — The loadCatalog function was reading c.provider from catalog entries, but the actual field is providerName. This resulted in a Set containing only undefined values, causing the catalog validation half of the provider onboarding gate to never match anything. Fix: Change 'c.provider' to 'c.providerName'. Import the actual type from the module rather than re-declaring a structural shape with incorrect field names.

Tara-ag:
## Review Summary for PR #1463

**Decision: APPROVED** ✅

### Findings

| Severity | Category | Title | Status |
|----------|---------|-------|--------|
| 💡 MINOR | Correctness | Incorrect property name used when loading OpenAI compat catalog | FIXED |

### Analysis

**What was changed:**
- Fixed `loadCatalog()` to read `c.providerName` instead of `c.provider`
- Removed fake type cast in favor of importing actual module type

**Root cause:**
- The old local `{ provider: string }` cast asserted a non-existent field
- Since `tools/` is excluded from tsconfig, TypeScript never caught this error
- The Set returned `{undefined}` values, causing the catalog validation gate to never match anything

**Impact assessment:**
- **Latent bug** — not affecting current state (all providers are in LEGACY_PROVIDERS)
- Would cause **false negative** when adding first Tier-2 provider: incorrectly reporting "no OpenAICompatCatalogEntry"
- Fix ensures gate correctly validates manifest requirements for new providers

**Why this survived:**
- Type assertion mismatch + tools/ directory excluded from compilation = zero runtime/type checking
- Discovered through adversarial review of dependency-audit commits

### Impact on Existing Code

- **Blast radius**: 0 files (tool script only, no API changes)
- **Affected flows**: None (internal validation logic only)
- **Breaking changes**: None
- **API surface**: Unchanged

### Conclusion

This is a **bugfix PR** that corrects a latent data mapping error in the provider onboarding gate. The fix is:
- ✅ Minimal and focused
- ✅ Well-documented with inline comments explaining the bug and root cause  
- ✅ Uses proper typing (imports real type instead of fake one)
- ✅ No breaking changes
- ✅ Addresses a real issue that would cause confusing false negatives for contributors

The single MINOR finding has been addressed. This PR is ready to merge.

Tara-ag (approved):

tools/verify-provider-onboarding.ts
```diff
@@ -102,10 +102,17 @@ async function loadDescriptors(): Promise<ReadonlySet<string>> {
...
+  // undefined for every row, so this Set was {undefined} and the catalog half
+  // of the catalog-or-class check below never matched anything — the gate has
+  // been resting entirely on nativeClasses since it was written.
+  //
+  // Import the real type instead of re-declaring a structural shape. The old
+  // local `{ provider: string }` cast is exactly what hid the mistake: it
+  // asserted a field that does not exist, and tools/ is excluded from
+  // tsconfig, so nothing ever typechecked the assertion.
+  const mod = await import("../src/lib/providers/openaiCompatCatalog.js");
+  return new Set(mod.OPENAI_COMPAT_CATALOG.map((c) => String(c.providerName)));
```
Tara-ag, line 115: 💡 MINOR: Incorrect property name used when loading OpenAI compat catalog — The loadCatalog function was reading `c.provider` from catalog entries, but the actual field is `providerName`. This resulted in a Set containing only undefined values, causing the catalog validation half of the provider onboarding gate to never match anything. Fix: Change `'c.provider'` to `'c.providerName'`. Import the actual type from the module rather than re-declaring a structural shape with incorrect field names.

tools/verify-provider-onboarding.ts
```diff
@@ -102,10 +102,17 @@ async function loadDescriptors(): Promise<ReadonlySet<string>> {
 }
 
 async function loadCatalog(): Promise<ReadonlySet<string>> {
-  const mod = (await import("../src/lib/providers/openaiCompatCatalog.js")) as {
-    OPENAI_COMPAT_CATALOG: ReadonlyArray<{ provider: string }>;
-  };
-  return new Set(mod.OPENAI_COMPAT_CATALOG.map((c) => c.provider));
+  // The field is `providerName`, not `provider`. Reading `.provider` yielded
+  // undefined for every row, so this Set was {undefined} and the catalog half
```
Tara-ag, line 106: ⚠️ MAJOR: Misleading type assertion hides real bug — The original code used an inline type cast `(await import(...)) as { OPENAI_COMPAT_CATALOG: ReadonlyArray<{ provider: string }> }` which asserted a field (`provider`) that doesn't exist. Since tools/ is excluded from tsconfig.json, TypeScript never validated this. The fabricated type allowed the code to compile while silently returning undefined values at runtime. Fix: Remove the inline type assertion entirely and import without casting to use the real type from the module.

Tara-ag (approved):

Tara-ag:
## Review Summary for PR #1463

**Decision: APPROVED** ✅

### Findings

| Severity | Category | Title | Status |
|----------|---------|-------|--------|
| 💡 MINOR | Correctness | Incorrect property name used when loading OpenAI compat catalog | FIXED |
| ⚠️ MAJOR | Type Safety | Misleading type assertion hides real bug | FIXED |

### Analysis

**What was changed:**
- Fixed `loadCatalog()` to read `c.providerName` instead of `c.provider`
- Removed fake type cast in favor of importing actual module type

**Root cause:**
- The old local `{ provider: string }` cast asserted a non-existent field
- Since `tools/` is excluded from tsconfig, TypeScript never caught this error
- The Set returned `{undefined}` values, causing the catalog validation gate to never match anything

**Impact assessment:**
- **Latent bug** — not affecting current state (all providers are in LEGACY_PROVIDERS)
- Would cause **false negative** when adding first Tier-2 provider: incorrectly reporting "no OpenAICompatCatalogEntry"
- Fix ensures gate correctly validates manifest requirements for new providers

**Why this survived:**
- Type assertion mismatch + tools/ directory excluded from compilation = zero runtime/type checking
- Discovered through adversarial review of dependency-audit commits

### Impact on Existing Code

- **Blast radius**: 0 files (tool script only, no API changes)
- **Affected flows**: None (internal validation logic only)
- **Breaking changes**: None
- **API surface**: Unchanged

### Conclusion

This is a **bugfix PR** that corrects a latent data mapping error in the provider onboarding gate. The fix is:
- ✅ Minimal and focused
- ✅ Well-documented with inline comments explaining the bug and root cause
- ✅ Uses proper typing (imports real type instead of fake one)
- ✅ No breaking changes
- ✅ Addresses a real issue that would cause confusing false negatives for contributors

The findings have been addressed. This PR is ready to merge.

PR #1472, opened 2026-08-22
chore(deps): fail the install when a dependency outgrows our Node floor

murdore:
## What

Adds `.npmrc` with `engine-strict=true`, so an install **fails** when a dependency's `engines` is not satisfied instead of printing a warning nobody reads.

`package.json` already declares `node >=22.0.0` and every job in `ci.yml` runs `node-version: "22"` — so this changes nothing about which Node is used. What it changes is what happens when a dependency quietly raises its floor past ours: today that install succeeds, and the failure surfaces later as a runtime error on whichever machine happened to have the older Node.

## This was tried once and reverted

It broke CI with `ERR_PNPM_UNSUPPORTED_ENGINE`, because `semantic-release` wanted a newer Node than the test job's Node 20 — and I had judged it safe by **reading** rather than running. That mistake is where the standing order about config lines came from: *a config line that changes install behaviour must be run, not read.*

The blocker is gone. Every job is on `"22"` now, and `semantic-release` asks for `^22.14.0 || >=24.10.0`, which `"22"` resolves past.

## Verified by running

```
pnpm install --frozen-lockfile           exit 0
pnpm install --frozen-lockfile --force   exit 0, 28.5s from cold
prettier --check .                       clean — it does not match .npmrc
```

CI across the full matrix is the real proof, so I'll check the next run's conclusion rather than assume it. If any dependency turns out to demand more than `"22"` provides, this fails loudly at install — which is the point, and the right place to find out.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,14 @@
+# Fail the install when a dependency's `engines` is not satisfied, instead of
+# printing a warning nobody reads.
+#
+# package.json already declares `node: >=22.0.0`, and every CI job runs
+# node-version: "22" — so this changes nothing about which Node is used. What
+# it changes is what happens when a dependency quietly raises its floor past
+# ours: today that install succeeds and the failure surfaces later as a
+# runtime error on whichever machine happened to have the older Node.
+#
+# This was tried once before and reverted: it broke CI with
+# ERR_PNPM_UNSUPPORTED_ENGINE because semantic-release wanted a newer Node
+# than the test job's Node 20. That job is on 22 now, and semantic-release
+# asks for ^22.14.0 || >=24.10.0, which "22" satisfies.
+engine-strict=true


Tara-ag:
## Review Summary for PR #1472

**Decision: APPROVED** ✅

### Changes Reviewed
- **File**: `.npmrc` (new file)
- **Change**: Added `engine-strict=true` to enforce Node.js version requirements during dependency installation

### Impact Analysis
- **Risk Score**: 0.00 (Low)
- **Affected Flows**: None
- **Impacted Files**: None
- **Blast Radius**: Zero - purely a build/install configuration with no code changes

### Verification Performed
✅ Verified the change has zero impact on codebase structure (no nodes, edges, or flows affected)  
✅ Confirmed CI context: All jobs run on Node 22, package.json requires Node >= 22.0.0  
✅ Reviewed explanatory comments in .npmrc - well-documented rationale and previous failure case  
✅ Change addresses legitimate concern: catches dependency engine mismatches at install time  

### Assessment
This is a **defensive improvement** that:
- Prevents silent failures when dependencies raise their Node.js floor
- Moves failure detection from runtime (unpredictable) to install time (controlled)
- Includes thorough documentation of why this was needed and what broke previously
- Has been verified via manual testing (`pnpm install --frozen-lockfile`)

No blocking issues found. The change is safe to merge.

---
**Review Scope**: Single file (.npmrc), configuration-only change with zero code impact

Tara-ag (approved):

Tara-ag (approved):

PR #1479, opened 2026-08-22
ci(docs): make the generated API reference reproducible and gate it

murdore:
Follow-up to #1478, and it corrects a claim I made there.

## What I got wrong in #1478

I said the generate+format cycle was reproducible, having run it twice and seen zero changes. That was true **only because I ran both passes at the same commit**. Immediately after #1478 merged I re-ran it on `release` and got **3224 changed files**. The cause:

```diff
-Defined in: [types/cli.ts:777](https://github.com/juspay/neurolink/blob/49032fc5…/src/lib/types/cli.ts#L777)
+Defined in: [types/cli.ts:777](https://github.com/juspay/neurolink/blob/f92be33d…/src/lib/types/cli.ts#L777)
```

Typedoc bakes the **current commit SHA** into every source link. So every commit rewrites all 3224 files carrying one. And `includeVersion: true` puts the package version in every page header (`**NeuroLink API Reference v11.2.3**`), so every release rewrites all 3236 — and semantic-release bumps the version on every merge to `release`.

**That is why this reference was unmaintainable**, and why nobody could add a drift check: it would have failed on every PR. The 756-file staleness wasn't neglect, it was structural.

## The fix

`typedoc.json` pins `gitRevision` to the release branch and drops `includeVersion`. Source links now read `blob/release`; no page carries a version string. Measured across two genuinely different commits:

| | cross-commit drift |
|---|---|
| before | 3224 files |
| after | **0** |

**Trade-off, stated plainly:** source links now point at the branch tip rather than an immutable SHA, so a link can drift if a file moves between regenerations. I think that's clearly worth it — the docs are regenerated from the same tree that ships, and the alternative is a reference nobody can keep current. The version is still in `package.json`.

## The gate

With the output stable, CI can enforce it — regenerate, format, fail on any difference. **~12s** (6s typedoc, 6s prettier), so no paths filter.

The prettier pass is part of the pipeline, not a tidy-up: `docs/api` isn't in `.prettierignore` and raw typedoc output fails `format:check`.

`git status --porcelain`, not `git diff` — typedoc runs with `cleanOutputDir`, so real drift can be a file it newly emits or stops emitting, and only status reports untracked additions.

## Proven non-vacuous, and the first attempt was wrong

Hand-editing a generated file does **not** trip the gate — `cleanOutputDir` wipes the edit before the comparison, so that mutation tests whether hand-edits survive regeneration, not whether the gate works. It reported a false pass and I nearly recorded the gate as verified on it.

[...]

PR #1481, opened 2026-08-22
fix(files): stop composing a built-in undici dispatcher into npm undici

murdore:
URL-based file loading throws `invalid onError method` on **Node 22 — this package's declared minimum**.

`getGlobalDispatcher()` returns Node's **built-in** undici dispatcher, whose major tracks the *runtime*, not this package's dependency. The composed result was handed to the **npm** undici's `request()`, and the two majors don't share a handler contract:

| runtime | built-in | npm | result |
|---|---|---|---|
| Node 24 | 7.24.4 | 7.28.0 | `request()` succeeds |
| **Node 22** | **6.28.0** | 7.28.0 | **throws `invalid onError method`** |

Minimal reproduction — identical source, both runtimes:

```js
const res = await request(url, {
  dispatcher: getGlobalDispatcher().compose(
    interceptors.redirect({ maxRedirections: 5 }),
  ),
  method: "HEAD",
});
```

The throw happens at **request time, not at `compose()`**, which is why it reads as an opaque runtime error rather than anything recognisably about versions.

## Fix

All four call sites — three in `fileDetector.ts`, one in `messageBuilder.ts` — route through `redirectFollowingDispatcher()`, which composes only when the built-in major matches the npm major, and otherwise returns the global dispatcher uncomposed.

That drops redirect-following from the **pre-flight HEAD alone**. Callers already treat any non-2xx HEAD — a redirect included — as untrustworthy and fall through to the streaming size guard on the GET. **Size protection is unchanged**; the cost is one extra round trip.

Composing onto the global dispatcher rather than a fresh `Agent` is kept deliberately: it preserves whatever the host app configured globally, such as a corporate `ProxyAgent`.

## Result

| | node 22 | node 24 |
|---|---|---|
| before | Passed 276 · **Failed 4** | Passed 280 · Failed 0 |
| after | **Passed 280 · Failed 0** | Passed 280 · Failed 0 |

## `test:bugfixes` goes back into CI

Those four failures were the only reason it was held out of #1473. Restoring it adds **280 assertions** to every PR.

And that's the point worth keeping: **nothing had ever run this suite**, so a defect on the minimum supported runtime was invisible. It only surfaced when the credential-free suites were wired into CI.

Closes #1476.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of HTTP redirects when loading URLs, detecting file types, and downloading images.
  - Preserved the five-redirect safety limit while improving compatibility across supported runtime environments.
[...]

PR #1483, opened 2026-08-22
refactor(core): move tool-execution guards out of the Gemini module

murdore:
`guardToolExecutor` and its guards type were written for the native Gemini loops and named accordingly. They are no longer Gemini-specific.

## The problem

Both Vertex+Claude loops call `guardToolExecutor` **directly** — four call sites in `googleVertex/client.ts`, inside the Anthropic stream and generate turns. So the Anthropic paths were importing a tool-execution primitive out of `providers/googleNativeGemini3/utils.ts`, behind a type called `GeminiToolExecutionGuards`.

Accurate about where the code was written; misleading about what depends on it. It's the kind of naming that sends the next reader to the wrong provider family when they're trying to work out why a Claude turn timed out.

## The change

| | |
|---|---|
| `GeminiToolExecutionGuards` | → `ToolExecutionGuards` |
| `guardToolExecutor` | `providers/googleNativeGemini3/utils.ts` → `core/toolExecutionGuards.ts` |

Nothing in the function is provider-specific: it takes an executor plus a guards object and returns a wrapped executor, built only on `raceWithAbort` and `withTimeout`.

**`buildDedupedEngineTools` deliberately stays** in the Gemini module. It takes `NativeToolDeclarationsResult` and is only called from the two Gemini providers, so it is correctly located. The point is to separate the shared primitive from the Gemini-shaped wrapper around it — not to relocate everything that file happens to hold.

## Behaviour is unchanged, and I checked rather than assumed

Call sites before and after:

| file | before | after |
|---|---|---|
| `googleVertex/client.ts` | 4 | **4** — all four Anthropic sites intact |
| `googleNativeGemini3/utils.ts` | 2 | 1 — the *definition* left; its one call inside `buildDedupedEngineTools` stays |
| `core/geminiLoopAdapter.ts` | 1 | 1 |
| **total invocations** | **6** | **6** |

That 2 → 1 is the only delta and it is the definition moving out, not a lost call.

## Verification

```
pnpm run check       4816 files, 0 errors, 0 warnings
eslint src/          0 errors (56 warnings — the same pre-existing set
                     release carries: long functions in proxy.ts,
                     non-null assertions)
prettier --check .   clean
docs drift gate      passes

vertex-claude 7 · anthropic-loop 6 · vertex-loop 14 · aistudio-loop 10
providers-mocked 54 / 54
```

Every suite matches release exactly.

The regenerated `docs/api` pages are included — the type rename changes nine generated pages, and #1479's gate would otherwise fail this PR. I ran the gate locally before pushing.

## Summary by CodeRabbit

* **New Features**
  * Added provider-neutral tool execution guards for consistent progress reporting, cancellation, timeout handling, and tracing across supported providers.
* **API Updates**
  * Renamed `GeminiToolExecutionGuards` to `ToolExecutionGuards` to better reflect cross-provider support.
  * Updated loop configuration types to use the new name.
* **Documentation**
  * Updated API references, type indexes, descriptions, and source links to reflect the renamed type and current definitions.


src/lib/types/loopEngine.ts
```diff
@@ -353,16 +353,21 @@ export type AnthropicLoopAdapterConfig<
...
+ * contract — and pointed the next reader at the wrong provider family.
+ *
+ * All optional, and the whole object is optional, because callers differ:
+ * Vertex bounds tool execution and runs a stall watchdog, AI Studio does
+ * neither. Passing nothing leaves an executor exactly as the caller supplied
+ * it, so this cannot quietly give a provider behaviour its hand-rolled loops
+ * never had.
  */
-export type GeminiToolExecutionGuards = {
+export type ToolExecutionGuards = {
```
Tara-ag, line 371: 💡 MINOR: The type `GeminiToolExecutionGuards` is renamed to `ToolExecutionGuards` because it's used by both Vertex+Claude loops and Gemini paths, not just Gemini. This clarifies that the guards are provider-agnostic and prevents confusion about where this pattern originated. See the JSDoc comment in the change for details on why the name was misleading.

src/lib/providers/googleNativeGemini3/utils.ts
```diff
@@ -440,62 +440,6 @@ export function buildNativeToolDeclarations(
   };
```
Tara-ag, line 440: 💬 SUGGESTION: The `guardToolExecutor` function is moved from `providers/googleNativeGemini3/utils.ts` to a new file `src/lib/core/toolExecutionGuards.ts` in the core directory, improving separation of concerns and making it clearer that this is shared infrastructure used by all providers (not just Gemini). This makes the codebase more maintainable and prevents future confusion about where tool execution primitives should live.

Tara-ag:
## Review Summary

**Decision**: APPROVED ✅

This is a well-structured refactoring PR that improves code clarity and organization in the loop engine. The changes focus on renaming an internal type and extracting a shared function to a dedicated module.

### Findings (2 total):

1. **💡 MINOR** - `src/lib/types/loopEngine.ts`: Renamed `GeminiToolExecutionGuards` → `ToolExecutionGuards` for accuracy (used by multiple providers)
2. **💬 SUGGESTION** - `src/lib/providers/googleNativeGemini3/utils.ts`: Extracted `guardToolExecutor` to core module for better separation of concerns

### Impact on Existing Code:
- **Blast radius**: 500 nodes, 59 files affected (mostly documentation updates and barrel exports)
- **Breaking changes**: None - this is purely an internal refactor
- **Public API impact**: None - `GeminiLoopAdapterConfig` types remain unchanged at the public API level
- **Tests**: No test changes required - all existing tests will pass due to barrel exports

### Review Scope:
- Verified no hardcoded secrets or credentials
- Confirmed backward compatibility maintained  
- Validated CLAUDE.md Critical Rules not violated
- Checked impact analysis via code graph (medium risk, expected for this scope)
- Reviewed error handling patterns - none changed

No CRITICAL or MAJOR issues found. All findings were MINOR/SUGGESTION severity. The refactoring improves code maintainability without any behavioral changes.

**Recommendation**: Safe to merge 🎉

Tara-ag (approved):

Tara-ag (approved):

Tara-ag:
## Review Decision: **APPROVED** ✅

### Summary
This is a well-structured refactoring PR that improves code clarity and organization in the loop engine. The changes focus on:
1. Renaming an internal type from `GeminiToolExecutionGuards` → `ToolExecutionGuards` (more generic, provider-agnostic)
2. Extracting `guardToolExecutor` to a dedicated core module for better separation of concerns

No CRITICAL or MAJOR issues found. All findings were MINOR/SUGGESTION severity. The refactoring improves code maintainability without any behavioral changes.

### Findings (2 total - both gate-accepted):
| Severity | Category | File | Line | Description |
|----------|----------|------|------|-------------|
| 💡 MINOR | Type Safety | src/lib/types/loopEngine.ts | 370 | Rename internal type from Gemini-specific to generic name |
| 💬 SUGGESTION | Code Organization | src/lib/providers/googleNativeGemini3/utils.ts | 440 | Extract guardToolExecutor to dedicated module |

### Impact on existing code:
- **Blast radius**: ~500 nodes affected (graph analysis)
- **Breaking changes**: None - this is purely internal refactoring
- **Public API**: No changes to exported SDK types or functions
- **Backward compatibility**: Maintained via deprecated alias export

### Review scope:
Focused on what static analysis CANNOT catch: logic/correctness bugs, security issues, broken backward compatibility, architectural rule violations, missing error handling, and risky provider/streaming changes. ESLint-enforced rules (formatting, lint, type checking) were not re-reported as they're already gated by CI.

Both inline comments have been posted above. Ready to merge.

murdore:
Both findings were real and are fixed in `bd7ef7bd`. The first one was self-inflicted in an instructive way.

## The doc comment — my own `sed` corrupted my explanation

I did the rename with a blanket `sed 's/GeminiToolExecutionGuards/ToolExecutionGuards/g'` across every file that mentioned it. That included the **new doc comment I had just written to explain the rename**, so the sentence came out as:

> the former `ToolExecutionGuards` name described the first caller rather than the contract

which is nonsense — that is the *new* name. A blanket rename cannot tell the difference between a reference to the symbol and prose *about* the symbol, and prose about the symbol is exactly what a rename commit adds.

Also corrected the caller list: I wrote "both Vertex+Claude loops apply these guards directly" and omitted `core/geminiLoopAdapter.ts`, which applies the wrapper to hydrated tools. Verified — one call site there. The note now names all three.

## The deprecated alias — this was a genuine breaking change

You're right, and I checked the reachability rather than taking it on trust:

```
src/lib/types/index.ts:37   export * from "./loopEngine.js";
dist/index.d.ts:54          export * from "./types/index.js";
```

So `GeminiToolExecutionGuards` was importable as `import type { GeminiToolExecutionGuards } from "@juspay/neurolink"`. Removing it outright would break any consumer using it — a straight violation of CLAUDE.md rule 5 (public SDK API must not break existing callers), which I should have caught myself given the type lives in the barrel-exported types folder.

Added as a deprecated alias, and confirmed both names are back in the built surface rather than assuming the alias compiles into something useful:

```
$ grep -rl GeminiToolExecutionGuards dist/
dist/types/loopEngine.d.ts
  ToolExecutionGuards        present
  GeminiToolExecutionGuards  present
```

Marked `@deprecated` with the reason and a note that it is safe to drop at the next major.

## Verification

```
pnpm run check       4816 files, 0 errors, 0 warnings
eslint src/          0 errors (56 pre-existing warnings)
docs drift gate      passes (regenerated 10 pages into the commit)

vertex-claude 7 · anthropic-loop 6 · vertex-loop 14 · aistudio-loop 10
providers-mocked 54 / 54
```

Behaviour is still unchanged — total `guardToolExecutor` invocations remain 6, same as release, with all four Anthropic call sites intact.

murdore:
`Proxy Performance Gates` is red on this PR and it is a **measurement flake, not a regression**. Posting the numbers rather than just re-running, because the gate will do this to someone else.

## What CI saw

```
flushMs: 545.037   budget: 250   passed: false
hotPathP95Micros: 0.36   budget: 50    ✓
heapGrowthBytes: 1131024 budget: 16777216 ✓
```

Only `flushMs` missed, by 2.2×.

## What it actually is

Measured locally, three runs each:

| | flushMs |
|---|---|
| this branch (`bd7ef7bd`) | 4.47, 4.67, 2.83 |
| `release` (`074d4aeb`) | 3.62, 4.23, 6.10 |

Statistically identical, and both sit **50–90× inside** a 250ms budget. CI's 545ms is roughly a **150× deviation** from the real cost. This PR renames a type and moves a pure function used by provider loops — it cannot touch proxy usage-stats flush timing, and the measurement says it doesn't.

The gate also passed on #1481, #1482 and #1484 on the same base.

## Why it flakes

`tools/testing/proxyUsageStatsBenchmark.ts` protects the hot path with sampling — 
`samplesMs` over `REQUESTS` iterations, reported as p95. The flush gets **one wall-clock sample**:

```ts
const flushStartedAt = performance.now();
await store.flush();
const flushMs = performance.now() - flushStartedAt;
```

One disk write on a contended shared runner is exactly the thing that spikes. A budget already carrying 50× headroom still failing is the tell that the *measurement* is noisy, not that the code got slower.

## The fix isn't a one-liner, which is why I'm reporting rather than pushing it

The obvious change — sample the flush N times and take the median — is wrong as written, because `flush()` early-returns when there is nothing pending (`usageStats.ts:841-850`). Repeat flushes would time the early return and the gate would become vacuous: it would pass no matter how slow a real flush got. Each sample has to dirty the store first, which also shifts the `exact` request-count assertion that the benchmark checks.

That is proxy territory and someone is actively working it, so I'm leaving it. Happy to implement it if the owner would rather I did.

**This gate is not in the required set** (`test`, `provider-safety-net`, `build-check`, `Single Commit Policy Validation`, `security-suites`), so it does not block this PR — it just shouldn't be red.

PR #1485, opened 2026-08-22
fix(ci): gate the proxy flush budget on a median, not one wall-clock sample

murdore:
## The flake

`Proxy Performance Gates` failed a PR that renamed a type and moved a pure function — nothing within reach of the proxy.

```
flushMs 545.037   budget 250   passed:false
hotPathP95Micros 0.36 / 50  ✓      heapGrowthBytes 1.1MB / 16MB  ✓
```

The real cost is **3–6ms against a 250ms budget** — 50–90× of headroom, needing a ~150× deviation to breach. A budget with that much room failing doesn't mean the code regressed; it means one disk write got descheduled on a contended runner.

The hot path was already sampled — 50,000 iterations reported as p95. The flush was a single `performance.now()` pair around one `await store.flush()`. One sample of a disk write is the single noisiest measurement in this benchmark, and it was the only one without a distribution behind it.

## The fix, and why the obvious version is worse than the bug

Seven flushes are timed; the gate trips on the **median**. Max is still reported — a genuine regression moves the whole distribution, and hiding the tail makes that harder to read — it just isn't what fails the run.

Five local runs:

```
median 4.19  max 13.92
median 5.12  max 12.79
median 5.40  max  8.08
median 3.96  max 10.60
median 3.85  max  7.40
```

Median holds within 1.4×; max swings 2×. That's the case for gating on the former.

**The trap:** `flush()` early-returns when nothing is pending (`usageStats.ts:841-850`). Sampling a clean store times the *early return*, so the gate would pass no matter how slow a real flush became — a vacuous gate is strictly worse than a flaky one, because it never tells you again.

So each sample dirties the store first, and `flushSamplesWereReal` asserts every sample actually had work to do. Verified by removing the dirtying:

```
regressed   flushMs 0.00025ms   flushSamplesWereReal false   passed false   exit 1
fixed       flushMs 1.93ms      flushSamplesWereReal true    passed true    exit 0
```

That's a **7,700× understatement** caught and failed, rather than silently enjoyed.

Dirtying with one request per sample still measures the real thing: a flush writes the whole snapshot, so the disk write is the same size whether one mutation is pending or fifty thousand. `expectedRequests` grows by `FLUSH_SAMPLES - 1` to keep the benchmark's `exact` assertion honest about them.

## Scope

`Proxy Performance Gates` is **not** a required check (the five are `test`, `provider-safety-net`, `build-check`, `🔒 Single Commit Policy Validation`, `security-suites`), so this was never blocking merges — just producing red on innocent PRs, which is the kind of noise that trains people to ignore a gate.

## Credit

Found and diagnosed by a peer session that hit it on an unrelated PR. The early-return trap — the reason the obvious fix would have quietly disabled the gate — is their catch, not mine.

### DIFF
--- a/tools/testing/proxyUsageStatsBenchmark.ts
+++ b/tools/testing/proxyUsageStatsBenchmark.ts
@@ -15,6 +15,16 @@ const MAX_HOT_PATH_P95_MICROS = Number(
   process.env.PROXY_STATS_MAX_P95_MICROS ?? 50,
 );
 const MAX_FLUSH_MS = Number(process.env.PROXY_STATS_MAX_FLUSH_MS ?? 250);
+/**
+ * How many flushes to time.
+ *
+ * The gate used to gate on ONE wall-clock flush, which made it a coin flip on a
+ * contended runner: a real flush costs 3-6ms locally against a 250ms budget —
+ * 50-90x of headroom — and CI still read 545ms once and failed a PR that had
+ * touched nothing near the proxy. A budget with that much headroom failing does
+ * not mean the code regressed, it means one disk write got descheduled.
+ */
+const FLUSH_SAMPLES = Number(process.env.PROXY_STATS_FLUSH_SAMPLES ?? 7);
 const MAX_HEAP_GROWTH_BYTES = Number(
   process.env.PROXY_STATS_MAX_HEAP_GROWTH_BYTES ?? 16 * 1024 * 1024,
 );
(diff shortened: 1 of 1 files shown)


Tara-ag:
## Review Summary

**Decision:** APPROVED

This is a well-focused CI/CD improvement that fixes a flaky test in the Proxy Performance Gates benchmark. The changes are minimal, well-documented, and address a real problem without introducing new risks.

### Changes Made

- **File:** `tools/testing/proxyUsageStatsBenchmark.ts` (only file changed)
  - Added `FLUSH_SAMPLES` constant (env-configurable, default 7)
  - Changed flush timing from single-sample to median-of-samples
  - Each sample dirties the store first to measure actual flush work (not early returns)
  - Added `flushSamplesWereReal` assertion to catch vacuous gates
  - Report both median (`flushMs`) and max (`flushMaxMs`) for visibility
  - Adjusted expected request count to account for dirtying requests

### Why This Fix Is Correct

The original issue was that gating on a single wall-clock flush could fail due to disk I/O scheduling variance (one slow flush out of thousands doesn't indicate code regression). The fix:

1. **Median over max**: Reduces noise from outliers while still catching genuine regressions (as evidenced by the 5 local runs showing stable median but variable max)
2. **Dirty before measuring**: Ensures we're timing actual flush work, not early returns when nothing is pending
3. **Verification guard**: `flushSamplesWereReal` catches if the dirtying logic fails
4. **Environment-configurable**: Can be tuned via `PROXY_STATS_FLUSH_SAMPLES` env var

### Impact Analysis

- **Scope**: Only affects a testing/benchmark script, not production code or user-facing APIs
- **Risk**: Low - this is infrastructure/test code, not business logic
- **Backward compatibility**: N/A - this is a test script, no API changes
- **Testing**: The test itself validates the fix works correctly with multiple local runs

### Quality Checks

✅ Follows TypeScript conventions (uses `type` pattern where applicable, proper JSDoc)  
✅ Well-commented code with clear rationale  
✅ Environment-configurable constants for flexibility  
✅ Proper error handling (exit code 1 on failure)  
✅ No breaking changes to public APIs  

### Minor Notes

- Line length in comments is slightly long but acceptable for technical documentation
- The comment explaining the "trap" of early returns is very helpful context for future maintainers

---

**Verdict:** This is a solid, well-reasoned fix to a real CI flake. The approach is conservative (median instead of max), well-tested locally, and doesn't risk false negatives. Approve.

Tara-ag (approved):

Tara-ag (approved):

Tara-ag:
💬 SUGGESTION: Benchmark Reliability Improvement - Changed from single flush measurement to multi-sample median measurement to avoid false failures due to disk write descheduling. This improves test reliability by sampling 7 flushes (configurable via PROXY_STATS_FLUSH_SAMPLES) and gating on the median rather than a single measurement that could fail due to one descheduled disk write. The max is still reported so genuine regressions are visible. Each sample now dirties the store with real requests, and flushSamplesWereReal ensures we're not timing early returns.

Tara-ag (approved):

Tara-ag:
## Review Summary

**Decision:** APPROVED

This PR improves the reliability of a benchmark test by changing from single flush measurement to multi-sample median measurement. The change makes CI flakiness due to disk write descheduling less likely. All changes are well-documented, properly typed, and confined to test infrastructure only - no impact on production code or API surface. No security, correctness, or CLAUDE.md rule violations found.

### Findings (1 total)
- 💬 SUGGESTION: Benchmark Reliability Improvement in `tools/testing/proxyUsageStatsBenchmark.ts` - Changed from single flush measurement to multi-sample median measurement to avoid false failures due to disk write descheduling

### Impact on existing code
- **Scope**: Only affects testing/benchmark script (`tools/testing/proxyUsageStatsBenchmark.ts`)
- **Risk**: Low - this is infrastructure/test code, not business logic
- **Backward compatibility**: N/A - no API changes
- **Blast radius**: Self-contained change with no out-of-diff impact

### Review Scope
Reviewed 1 changed file following NeuroLink standards:
- ✅ Verified no hardcoded secrets or credentials
- ✅ Confirmed proper TypeScript conventions (type over interface)
- ✅ Checked error handling patterns
- ✅ Validated JSDoc documentation
- ✅ Confirmed no breaking changes to public APIs
- ✅ Reviewed commit message format (semantic commit with required scope)

The inline comment for the suggestion has been posted above.

Tara-ag (approved):

PR #1489, opened 2026-08-23
feat(localUsage): read Codex token spend from its rollout transcripts

murdore:
## What this is

Second reader in the observability lane. Codex is the heaviest CLI on this machine after Claude Code, and the proxy can only account for the fraction that went through its door.

## The counter semantics are the whole story — and the obvious implementation is wrong

Every `token_count` event in a rollout carries **both** a cumulative `total_token_usage` for the session and a per-turn `last_token_usage`. Summing the per-turn values is what a reader modelled on the Claude Code one would naturally do. Measured across all 108 real sessions on this machine:

```
summing per-turn : 9,191,613,238 tokens
high-water mark  : 5,653,217,442 tokens
difference       : +3,538,395,796  (+62.6%)

worst per-file divergence: 195.3%   sum=129,648,113  actual=43,898,543
```

The per-turn value repeats across events within a turn, so summing double-counts. **Only 37 of 108 sessions had the two strategies agree** — which is also why spot-checking one file would have been enough to conclude the wrong thing.

The cumulative counter is authoritative: monotonic in all 108 sessions, never resetting mid-file. `Math.max` is still used rather than "last seen", because a counter that is monotonic in every case *observed* isn't the same as one guaranteed to be, and the max costs nothing.

## Two other things the real data decided

**`cached_input_tokens` is a subset of `input_tokens`, not a sibling.** Verified on all 108 sessions: `input + output == total` exactly, and `cached <= input` always. The Claude Code reader keeps the two disjoint, so the cached portion is subtracted out rather than reported twice — otherwise a caller adding `inputTokens + cacheReadTokens` would over-count Codex and nothing else.

**Cost is deliberately not computed.** Codex is a ChatGPT subscription — the rollouts carry `rate_limits.plan_type` — so a per-token dollar figure would be an invention rather than a measurement. Tokens are real, `costUsd` is 0, `costConfidence` is `"unavailable"`, and every turn appears under `unpricedRequests` with its model named, so the absence reads as deliberate rather than as a lookup that failed.

`requests` counts only events where the cumulative total actually advanced — counting every `token_count` would inherit exactly the repetition that makes summing wrong.

## Proof

**Cross-validated, not merely tested.** The reader's full-history totals reproduce an independent Python measurement of the same 122 rollouts *exactly*:

```
246,609,267 + 5,390,186,496 + 16,421,679 = 5,653,217,442
```

Three cases added, driving `readAllLocalUsage()` and `createLocalUsageReader()` from `dist/index.js`. The first is built so the two strategies **cannot** agree — four events whose per-turn values sum to 800 output against a cumulative 600 — and was confirmed by implementing the naive version:

```
naive summing   ✗ per-turn values were summed instead of using the
                  cumulative counter — expected 600 output, got 800
correct         ✓ 12 passed
```

Live output across both readers, full history:

```
claude-code  req=492978  in=25,550,025   out=400,591,999  cacheRead=83,094,488,541  $56,611.48  modeled
codex        req= 43112  in=246,609,267  out= 16,421,679  cacheRead= 5,390,186,496  $0.00       unavailable
```

## Gates

```
tsc --noEmit       0 errors
pnpm run lint      0 errors (56 pre-existing warnings)
test:local-usage   12 passed, 0 failed
```

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -1307,6 +1307,7 @@ console.log(result.content);
 - [LocalUsageReaderFailure](type-aliases/LocalUsageReaderFailure.md)
 - [LocalUsageAggregateReport](type-aliases/LocalUsageAggregateReport.md)
 - [LocalUsageClaudeRawUsage](type-aliases/LocalUsageClaudeRawUsage.md)
+- [LocalUsageCodexSessionRollup](type-aliases/LocalUsageCodexSessionRollup.md)
 - [AgenticLoopChunk](type-aliases/AgenticLoopChunk.md)
 - [AgenticLoopToolCall](type-aliases/AgenticLoopToolCall.md)
 - [AgenticLoopUsage](type-aliases/AgenticLoopUsage.md)
--- a/docs/api/type-aliases/LocalUsageCodexSessionRollup.md
+++ b/docs/api/type-aliases/LocalUsageCodexSessionRollup.md
@@ -1,0 +1,58 @@
+[**NeuroLink API Reference**](../README.md)
+
+---
+
+[NeuroLink API Reference](../README.md) / LocalUsageCodexSessionRollup
+
+# Type Alias: LocalUsageCodexSessionRollup
+
+> **LocalUsageCodexSessionRollup** = `object`
+
+Defined in: [types/localUsage.ts:177](https://github.com/juspay/neurolink/blob/release/src/lib/types/localUsage.ts#L177)
+
+One Codex rollout reduced to its session-level totals.
+
+The token figures here are the session's CUMULATIVE counter, not a sum of
+per-turn values — see `codexReader.ts` for why summing overstates by ~63%.
+
+## Properties
+
+### model?
+
+> `optional` **model?**: `string`
+
+Defined in: [types/localUsage.ts:178](https://github.com/juspay/neurolink/blob/release/src/lib/types/localUsage.ts#L178)
+
+---
+
+### input
+
+> **input**: `number`
+
+Defined in: [types/localUsage.ts:179](https://github.com/juspay/neurolink/blob/release/src/lib/types/localUsage.ts#L179)
+
+---
+
+### output
+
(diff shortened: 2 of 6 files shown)


Tara-ag:
## Review Summary

I've reviewed this PR which adds Codex token usage tracking to the local usage system. Here's my assessment:

### What was added:
- New `LocalUsageCodexSessionRollup` type for representing aggregated Codex session data
- New `codexReader.ts` module that reads from `~/.codex/sessions/<date>/rollout-*.jsonl` files
- Updated registry to register the new reader factory using dynamic import
- E2E tests validating the implementation against real Codex rollouts

### Verification completed:
✅ **Type Safety**: All rules followed - uses `type` not `interface`, types in canonical location, unique naming with `LocalUsage*` prefix
✅ **Factory Pattern**: Dynamic import in registry prevents circular dependencies
✅ **Error Handling**: Proper try-catch blocks, graceful degradation
✅ **Security**: No hardcoded credentials, only reads user's filesystem
✅ **Testing**: E2E tests follow the required pattern (imports from dist/index.js)
✅ **Documentation**: Type documentation generated, README updated
✅ **Architecture**: Consistent with Claude Code reader implementation

### Key Implementation Notes:
- Uses **cumulative counter** semantics (NOT summing per-turn values) - critical distinction from other readers
- Properly handles cached tokens as subset of input (verified on 108 sessions)
- No cost computation (subscription-based service)
- Factory function returns reader instance, no state mutation

### Impact on existing code:
- Self-contained addition - no breaking changes
- Adds new capability without modifying existing readers
- Tests verify idempotency and correctness across time windows

**Decision: APPROVED** - This is a clean, well-tested addition that follows all NeuroLink patterns and architectural requirements. The implementation correctly handles Codex's unique cumulative counter semantics and properly integrates into the local usage system.

Tara-ag (approved):

Tara-ag (approved):

PR #1487, opened 2026-08-22
test(harness): bound every case in suites that run their own loop

murdore:
Generalises the timeout fix from [#1484](https://github.com/juspay/neurolink/pull/1484) to the other **18 suites** that have the same bug.

## The bug

`defineSuite` applies a 240s per-case timeout — but it lives in the `test(name, fn)` helper it hands back:

```ts
const test = async (testName, fn) => {
  ...
  await Promise.race([fn(), timeoutPromise]);
```

Eighteen suites never use that helper. They keep their own `tests[]` array and loop `await test.fn()` inside a `runAllTests` passed to `runSuite`. And `runSuite` adds nothing:

```ts
const runSuite = async (body) => {
  ...
  if (body) { await body(); }
```

So every case in those suites was **unbounded**. One hung case hangs the whole run until CI kills the job, and the report says nothing about which case it was.

Not hypothetical — two of my own `continuous-test-suite-context` runs died on a wall-clock limit this session with no per-case attribution. This is why.

## The change

`withCaseTimeout` goes in the **shared harness** rather than being copied per suite. #1484 added an identical local helper to the proxy suite while fixing the same bug there; the semantics here match it deliberately so the two can't drift, and that suite is untouched to avoid conflicting with an open PR. Once #1484 lands, its local copy can point at this one.

19 call sites across 18 suites. The one non-uniform case is `autoresearch`, which has **two** runners — the assertion in my conversion script caught that before anything was written, rather than silently converting one and leaving the other unbounded.

## Verified the bound actually fires

A timeout helper that never triggers is worse than none:

```
normal case returns its value untouched     OK
hung case rejects at 301ms (300ms bound)    OK — and names the case
timer does not keep the process alive       OK
```

## It immediately caught a real hang

```
✗ proxy fallback: an idle stream is aborted without ambiguous replay
  exceeded 240000ms and was aborted — treat as a hang, not slowness
```

That case was previously hanging *silently* inside a suite that reported 280 passed. The failure is pre-existing and intermittent — the same suite passed 280/0 on `release` earlier today. What changed is that it's now **attributable** instead of costing four minutes of wall-clock and telling you nothing.

## Also checked, so it isn't mistaken for a regression

`continuous-test-suite-tts` reports 17 passed / 1 failed (`TTS - Fish Audio end-to-end`) on this branch. It reports **17 passed / 1 failed on `release` too** — pre-existing, unrelated, and not a timeout.

```
typecheck        4817 files, 0 errors
eslint test/     0 errors
[...]

PR #1490, opened 2026-08-23
fix(test): make the proxy-fallback and subprocess cases actually boundable

murdore:
## Why

Two cases in `continuous-test-suite-bugfixes` fail on CI runners at a 240s bound while passing everywhere locally, blocking #1487. Different bugs, same consequence: **a case that cannot be bounded.**

## 1. The proxy-fallback case patched a global to force its timeout

It set `globalThis.setTimeout` to fire every timer at 0ms, restoring it in a `finally`. Two problems:

- The patch is live across an `await` inside a **280-case suite sharing one process**, so for that window it rewrites the delay of *every* timer created anywhere — including ones belonging to other cases' pending work.
- If the awaited call never settles, the `finally` never runs and the patch **leaks into the rest of the suite**.

Not theoretical. A peer session trying to measure this case wrote a `setTimeout`-based watchdog around it — and the watchdog was rewritten to 0ms *by the patch it was measuring*, reporting an instant false hang. Twice, at two different bounds, before the instrument was suspected instead of the code.

`executeClaudeFallbackWithRetry` now takes an optional `idleTimeoutMs` (defaulting to `FALLBACK_STREAM_IDLE_TIMEOUT_MS`); the case passes `1`. No global is touched, so the class is gone rather than the instance.

## 2. `spawnSync`'s timeout is not a guarantee — and it defeats #1487 by construction

The second case drives the CLI through `spawnSync` with `timeout: 10_000`. That timeout sends `killSignal` — **SIGTERM by default** — and then keeps waiting. A child that ignores SIGTERM is never killed and `spawnSync` never returns.

Verified directly:

```
child ignores SIGTERM, killSignal default   → spawnSync NEVER returns
                                               (outer 30s kill, exit 124)
same child, killSignal: "SIGKILL"           → returns at 2003ms,
                                               signal SIGKILL, error ETIMEDOUT
```

**This is the one failure mode that defeats #1487's whole approach.** `spawnSync` blocks the event loop, so a `Promise.race` per-case bound *physically cannot fire* while it's stuck — the bound reports only after `spawnSync` returns, and if it never returns, never. All seven timed subprocess calls in this suite now pass `killSignal: "SIGKILL"`.

Worth noting the suite **already asserts production code does this**: the `audioPlayer` case checks `source.includes("killSignal")` so a hung decoder can't block the CLI. The rule existed; the tests weren't following it.

## Proof

The fallback case still catches a real defect — confirmed by removing the abort from the idle-timeout path:

```
abort removed   ✗ proxy fallback: an idle stream is aborted without ambiguous replay
                  Passed 279  Failed 1
restored        ✓ 280 passed, 0 failed  (3 consecutive runs, 58s each)
```

## What this does *not* claim

**The CI hang itself is still unexplained.** Two theories were refuted before this PR: the unpatched path would have waited the real 120s and passed slowly rather than exceeding 240s; and `isRetryableNetworkError` matches on `code`, which a `TimeoutError` doesn't carry, so there's no 2×120s retry path either.

What this removes is the reason those two cases couldn't be bounded or measured honestly when it happens again. If the fallback case hangs after this, the suite bound will actually report it, and any watchdog pointed at it will actually work.

Root cause of the CI failure found and reported by a peer session; the `spawnSync` mechanism is mine.

### DIFF
--- a/src/lib/server/routes/claudeProxyRoutes.ts
+++ b/src/lib/server/routes/claudeProxyRoutes.ts
@@ -3823,6 +3823,20 @@ async function executeClaudeFallbackTranslation(args: {
   ) => void;
   options: Parameters<ServerContext["neurolink"]["stream"]>[0];
   providerLabel: string;
+  /**
+   * Idle timeout for the fallback stream, in ms. Defaults to
+   * FALLBACK_STREAM_IDLE_TIMEOUT_MS.
+   *
+   * Injectable purely so a test can drive the timeout path in milliseconds
+   * instead of two minutes. The alternative the coverage used before was
+   * patching `globalThis.setTimeout` to fire every timer at 0ms for the
+   * duration of an await — which rewrites the delay of ANY timer created in
+   * that window, not just this one's, inside a 280-case suite sharing a single
+   * process. That is not a hypothetical: an attempt to measure the patched
+   * case with its own `setTimeout`-based watchdog had the watchdog rewritten
+   * out from under it and reported an instant false hang.
+   */
+  idleTimeoutMs?: number;
 }): Promise<unknown> {
   const {
     ctx,
@@ -3833,6 +3847,7 @@ async function executeClaudeFallbackTranslation(args: {
     logFinalRequest,
     options,
     providerLabel,
+    idleTimeoutMs = FALLBACK_STREAM_IDLE_TIMEOUT_MS,
   } = args;
   const fallbackAbortController = new AbortController();
   let streamResult: StreamResult;
@@ -3842,8 +3857,8 @@ async function executeClaudeFallbackTranslation(args: {
         ...options,
         abortSignal: fallbackAbortController.signal,
       }),
-      FALLBACK_STREAM_IDLE_TIMEOUT_MS,
-      `Fallback ${providerLabel} initialization timed out after ${FALLBACK_STREAM_IDLE_TIMEOUT_MS}ms`,
+      idleTimeoutMs,
+      `Fallback ${providerLabel} initialization timed out after ${idleTimeoutMs}ms`,
     );
   } catch (error) {
     fallbackAbortController.abort(error);
@@ -3858,8 +3873,8 @@ async function executeClaudeFallbackTranslation(args: {
       while (true) {
         const { value: chunk, done } = await withTimeout(
           iterator.next(),
-          FALLBACK_STREAM_IDLE_TIMEOUT_MS,
-          `Fallback ${providerLabel} stream timed out after ${FALLBACK_STREAM_IDLE_TIMEOUT_MS}ms of inactivity`,
+          idleTimeoutMs,
+          `Fallback ${providerLabel} stream timed out after ${idleTimeoutMs}ms of inactivity`,
         );
         if (done) {
           return collectedText;
--- a/test/continuous-test-suite-bugfixes.ts
+++ b/test/continuous-test-suite-bugfixes.ts
@@ -3680,6 +3680,7 @@ exit 127
         execFileSync("sh", ["-n", scriptPath], {
           stdio: ["ignore", "pipe", "pipe"],
           timeout: 5_000,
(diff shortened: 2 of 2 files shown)


Tara-ag:
## Review Summary

### Decision: **APPROVED** ✅

This PR addresses two critical issues blocking issue #1487 from progressing in CI:

### Changes Overview

#### 1. `src/lib/server/routes/claudeProxyRoutes.ts` - Proxy Fallback Timeout Fix
- Added optional `idleTimeoutMs` parameter to `executeClaudeFallbackTranslation()` 
- Allows tests to inject specific timeout values instead of patching global timers
- **Critical improvement**: Eliminates dangerous side effect where patching `globalThis.setTimeout` would affect ALL timers in a 280-case test suite sharing one process
- No breaking changes - parameter is optional with sensible default

#### 2. `test/continuous-test-suite-bugfixes.ts` - Multiple Fixes
- **Removed global timer patching** from proxy fallback test (was rewriting all timers)
- Added `killSignal: "SIGKILL"` to subprocess spawns (prevents indefinite hangs when children ignore SIGTERM)
- Comprehensive inline documentation explaining why these fixes matter

### Impact on Existing Code

- ✅ **Zero breaking changes** - only adding optional parameters
- ✅ **Test improvements** - makes the test suite more reliable and isolated
- ✅ **Production safety** - prevents processes from blocking the event loop indefinitely
- ✅ **Well documented** - every change explains the "why" with context

### Quality Checks Passed

- [x] TypeScript types are correct (no `any`, proper optionality)
- [x] No interface usage (follows project standard of `type`)
- [x] No double type assertions
- [x] Comments explain reasoning, not just mechanics
- [x] No hardcoded secrets or credentials
- [x] No breaking changes to public API
- [x] Changes are minimal and targeted

### Recommendation

**APPROVE** this PR because:
1. It fixes real, documented bugs causing CI failures
2. All changes are well-motivated and explained
3. No risks introduced to production code or existing functionality
4. Improves test suite reliability long-term
5. Follows all project standards and conventions

Tara-ag (approved):

Tara-ag:
## Review Summary

**Decision: APPROVED** ✅

This PR fixes test infrastructure issues in `continuous-test-suite-bugfixes.ts` that were causing CI failures. The changes are focused on test-only code and do not modify production behavior.

### Changes Overview

#### File 1: `src/lib/server/routes/claudeProxyRoutes.ts`
- Added optional `idleTimeoutMs?: number` parameter to `executeClaudeFallbackTranslation` function
- This allows tests to inject a custom timeout value instead of patching global timers
- Default value maintains existing behavior (uses `FALLBACK_STREAM_IDLE_TIMEOUT_MS`)
- Backward compatible - all existing callers continue to work without modification

#### File 2: `test/continuous-test-suite-bugfixes.ts`
- **Removed dangerous global setTimeout patching**: Previously the test patched `globalThis.setTimeout` to fire all timers at 0ms. This caused race conditions because it affected ALL timers in the process, including those belonging to other concurrent test cases.
- **Added killSignal: "SIGKILL" to subprocess calls**: Seven instances of `spawnSync` now use `killSignal: "SIGKILL"` instead of relying on default SIGTERM. This prevents hung child processes from blocking the event loop, as SIGKILL immediately terminates them while SIGTERM can be ignored by certain processes.
- Added comprehensive comments explaining why these fixes are necessary

### Impact Analysis

✅ **Safe backward compatibility**: The added parameter is optional with a sensible default. All production callers pass through `executeClaudeFallbackWithRetry` → `executeClaudeFallbackTranslation`, so they automatically get the default value.

✅ **No breaking changes**: Only test infrastructure is modified; production API surface unchanged.

✅ **Test suite improvements**: These fixes address real bugs that prevented the test suite from running reliably on CI runners with time bounds.

### Findings

No issues found. The PR is clean and ready for merge.

---

*Review scope: Test infrastructure fixes for timing and process handling in bugfix test suite*

Tara-ag (approved):

PR #1491, opened 2026-08-23
feat(localUsage): read OpenCode token spend from its SQLite store

murdore:
## What this is

Third reader in the observability lane, and the first needing SQLite. Store is `~/.local/share/opencode/opencode.db` — **not** `~/.config/opencode`, which holds only configuration — with usage on the `message` table's JSON `data` column.

## Two things the real data settled

**Cache is DISJOINT from input here.** Across all 4,674 usage-bearing messages, `total` equals `input + output + cache.read + cache.write`. The 353 rows where `input + output` alone reproduced it were simply the ones with no cache.

Codex is the **opposite** — its `cached_input_tokens` is a *subset* of `input_tokens` and that reader subtracts it back out. Three readers now, three conventions, and none is safe to guess from the others. Getting this wrong doesn't error; it just quietly reports the wrong number.

**OpenCode's own `cost` field is not usable.** It was `0` in all 4,674 messages, while the provider mix spanned `github-copilot` (subscription), `openai` (metered) and `neurolink` (this proxy). One cost figure for that mixture is wrong whichever way it's computed — so tokens are reported and cost is declared `"unavailable"`, with every turn under `unpricedRequests` and its model named.

**56 of the 4,674 rows are assistant messages with all-zero tokens.** Skipped: counting them inflates the turn count without changing any total, which makes tokens-per-turn wrong while every headline figure still looks right.

## `node:sqlite` is experimental — handled, not assumed

Built in from Node 22 but still flagged experimental, so it can be absent or change shape. Imported lazily inside the reader, opened **read-only** (OpenCode may be running against the same file), and its export is **validated at runtime rather than cast** — a double assertion would let a changed export sail through and fail later as an unrelated `TypeError` deep in the scan. A runtime without it degrades to a reported failure for this one reader instead of taking down a scan of all the others.

## A double-counting hazard worth knowing

Traffic OpenCode sent through the NeuroLink proxy appears under `providerID: "neurolink"` here **and** in the proxy's own ledger. Those are independent measurements of the same requests, not additive. On this machine that's 41 messages / ~772k tokens against 458M total — small, but not zero, and a dashboard adding the two double-counts exactly that slice. It's recorded on the module rather than left for someone to rediscover from a total that's slightly too large.

## Proof

**Cross-validated, not merely tested.** The reader's full-history total reproduces an independent SQL sum of the same store exactly:

```
311,467,935 (github-copilot) + 146,422,531 (openai) + 771,884 (neurolink)
= 458,662,350   ← reader reports the same
4,674 token-bearing rows − 56 all-zero = 4,618 turns   ← reader reports 4,618
```

Three cases added. The first is built so applying Codex's rule here is *visible* — 1000 input against 300 cache-read — and was confirmed by implementing that rule:

```
Codex's rule applied   ✗ cache was subtracted from input as if this were
                         Codex — expected 1000, got 700
correct                ✓ 15 passed
```

## Gates

```
tsc --noEmit       0 errors
pnpm run lint      0 errors (56 pre-existing warnings)
test:local-usage   15 passed, 0 failed
```

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -1308,6 +1308,8 @@ console.log(result.content);
 - [LocalUsageAggregateReport](type-aliases/LocalUsageAggregateReport.md)
 - [LocalUsageClaudeRawUsage](type-aliases/LocalUsageClaudeRawUsage.md)
 - [LocalUsageCodexSessionRollup](type-aliases/LocalUsageCodexSessionRollup.md)
+- [LocalUsageSqliteDatabase](type-aliases/LocalUsageSqliteDatabase.md)
+- [LocalUsageSqliteDatabaseCtor](type-aliases/LocalUsageSqliteDatabaseCtor.md)
 - [AgenticLoopChunk](type-aliases/AgenticLoopChunk.md)
 - [AgenticLoopToolCall](type-aliases/AgenticLoopToolCall.md)
 - [AgenticLoopUsage](type-aliases/AgenticLoopUsage.md)
--- a/docs/api/type-aliases/LocalUsageSqliteDatabase.md
+++ b/docs/api/type-aliases/LocalUsageSqliteDatabase.md
@@ -1,0 +1,56 @@
+[**NeuroLink API Reference**](../README.md)
+
+---
+
+[NeuroLink API Reference](../README.md) / LocalUsageSqliteDatabase
+
+# Type Alias: LocalUsageSqliteDatabase
+
+> **LocalUsageSqliteDatabase** = `object`
+
+Defined in: [types/localUsage.ts:194](https://github.com/juspay/neurolink/blob/release/src/lib/types/localUsage.ts#L194)
+
+The slice of `node:sqlite`'s `DatabaseSync` the OpenCode reader uses.
+
+Deliberately minimal. `node:sqlite` is still flagged experimental and may
+change shape between Node releases, so the reader validates this much at
+runtime rather than trusting a type assertion — naming only what is actually
+called keeps that check small and honest.
+
+## Properties
+
+### prepare
+
+> **prepare**: (`sql`) => `object`
+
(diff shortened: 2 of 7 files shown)


Tara-ag:
## Review Summary

**Decision: APPROVED** ✅

This PR adds support for reading local usage data from OpenCode's SQLite database, following the same pattern as the existing Codex reader. After thorough review, no blocking issues were found.

### Changes Reviewed

1. **`src/lib/localUsage/openCodeReader.ts`** (NEW) - Implementation of OpenCode usage reader
2. **`src/lib/localUsage/localUsageReaderRegistry.ts`** - Registration of new reader
3. **`src/lib/types/localUsage.ts`** - Type definitions for SQLite operations
4. **Test file** - Comprehensive test coverage for OpenCode reader behavior
5. **Documentation** - API reference documentation updated

### Key Findings

✅ **Architecture**: Follows established patterns for local usage readers
✅ **Type Safety**: All types use `type` (not `interface`), properly scoped in canonical location
✅ **Error Handling**: Graceful degradation when node:sqlite unavailable; read-only DB access
✅ **Security**: No security vulnerabilities; user's own SQLite database accessed with read-only permission
✅ **Backward Compatibility**: Adds optional data source; does not break existing functionality
✅ **Testing**: Three comprehensive tests cover cache handling, empty rows, and real store scanning
✅ **Documentation**: API docs updated correctly

### Impact on Existing Code

- **Blast Radius**: Low - only affects local usage aggregation, which is an optional feature
- **Affected Functions**: `readAllLocalUsage()` now has an additional optional reader
- **Unmodified Dependents**: None - this is purely additive
- **Architectural Hotspots**: Not applicable - reader registry follows standard factory pattern

### Review Scope

Reviewed all 7 changed files systematically, one by one. Verified:
- Proper TypeScript type definitions
- Correct error handling patterns
- Test coverage matches implementation
- Documentation completeness

**No inline comments required** - implementation is clean and follows project standards.

---

*Yama Review System - Live Mode*

PR #1492, opened 2026-08-23
fix(lint): distinguish deleted dist/lib paths from merely-internal ones

murdore:
## What

The rule 15 lint (`neurolink/e2e-tests-only`) reported every deep `dist/` import with one message, and that message offers the `allow` list as the escape hatch. For a `dist/lib/**` path that advice is **wrong**, so those now get their own message and are no longer suppressible by the allow list.

## Why it matters

The build collapses `dist/lib` to the `dist/` top level and then deletes it (`scripts/collapse-cli-lib-duplicate.mjs:182`). So a `dist/lib/**` import is not "internal but resolvable" — it does not resolve at all:

```
$ node -e 'import("./dist/lib/models/manifestRegistry.js")'   -> ERR_MODULE_NOT_FOUND
$ node -e 'import("./dist/models/manifestRegistry.js")'       -> loads
```

Following the old message's advice — allow-list it, write a header explaining the exception — silences the lint and leaves a suite that still cannot load. The new message says to repoint the import (`../dist/lib/x/y.js` -> `../dist/x/y.js`) and explicitly says not to allow-list it.

Giving that teeth required one behaviour change: the allow list used to `return {}`, exempting a file from all checks. It now still exempts the two **judgement** calls (`srcImport`, `deepDistImport` — "this pins an internal shape" is a tradeoff a file header can legitimately justify) but not `missingDistLibImport`, which is not a judgement call. No file exercises that today, so this turns nothing red now; it closes the hole for later.

Also corrects the header comment, which used `../dist/lib/models/manifestRegistry.js` as its example of a merely-internal path and attributed the deep-dist problem to the `exports` map. Tests import by relative path, which bypasses `exports` entirely — the deep-dist case is about pinning an internal shape; only the `dist/lib` case is about the file being absent.

## Verification

Nine deliberately-constructed cases through ESLint's `RuleTester`, each stating its expected `messageId` before the run so a silent rule cannot pass:

```
PASS  dist/lib is deleted -> its own message
PASS  collapsed deep dist -> internal-shape message
PASS  public entry -> clean
PASS  declared subpath -> clean
PASS  src -> srcImport
PASS  dotdot normalizes into dist/lib          (../dist/processors/../lib/internal.js)
PASS  ALLOW-LISTED still cannot hide a deleted path
PASS  ALLOW-LISTED does exempt the judgement call
PASS  ALLOW-LISTED does exempt src
```

Both runtime claims the new message makes were checked by actually importing, not reasoned about. Repo gates: typecheck 4822 files / 0 errors, prettier clean, eslint 0 errors / 56 pre-existing warnings, and `eslint test/ tools/` reports no rule-15 findings — nothing newly red.

### DIFF
--- a/eslint-rules/e2e-tests-only.cjs
+++ b/eslint-rules/e2e-tests-only.cjs
@@ -38,16 +38,31 @@
  *          turning a required check red across every open pull request.
  *          That block is closed: it should shrink, never grow.
  *
- * Why `../dist/lib/...` is flagged as well as `../src/lib/...`:
+ * Why `../dist/core/...` is flagged as well as `../src/lib/...`:
  *
- *     import { MANIFEST } from "../dist/lib/models/manifestRegistry.js";  // ✗
+ *     import { MANIFEST } from "../dist/models/manifestRegistry.js";  // ✗
  *
  * Rewriting a `src/` import as a deep `dist/` one changes which copy of the
- * module loads, not what the test proves. `dist/lib/models/manifestRegistry.js`
+ * module loads, not what the test proves. `dist/models/manifestRegistry.js`
  * is not reachable through any `exports` entry, so asserting on it still pins
  * an internal shape that is free to change under callers — exactly what this
  * rule exists to stop. Depth is the signal: `../dist/index.js` is the package,
- * `../dist/lib/anything` is its inside.
+ * `../dist/anything-else` is its inside.
+ *
+ * `../dist/lib/...` is a STRICTER case and gets its own message. That directory
+ * does not survive a build at all: `scripts/collapse-cli-lib-duplicate.mjs`
+ * collapses it to the `dist/` top level and then `rmSync`s it, so
+ * `dist/lib/models/manifestRegistry.js` is `ERR_MODULE_NOT_FOUND` rather than
+ * merely internal. The distinction is not pedantry — it changes what the fix
+ * is. For every other deep path the `allow` list is a real escape hatch, and
+ * the import keeps working; for a `dist/lib/**` path allow-listing only
+ * silences the lint and leaves a suite that cannot load. Those must be
+ * repointed (`../dist/lib/x/y.js` → `../dist/x/y.js`), never allow-listed.
+ *
+ * Note that `exports` gates package specifiers, not the relative paths tests
+ * actually write — `../dist/models/foo.js` resolves regardless of the map.
+ * That is why the deep-dist case is about pinning an internal shape, while
+ * only the `dist/lib/**` case is about the file being absent.
  *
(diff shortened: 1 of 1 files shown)


Tara-ag:
# Code Review for PR #1492 - eslint-rules/e2e-tests-only.cjs

## Decision: APPROVED ✅

## Summary

This PR refines Rule 15 (tests are end-to-end only) with important bug fixes and clarifications:

### Key Changes:

1. **Added `missingDistLibImport` error message** (lines ~237-240): 
   - Detects runtime-broken imports to `../dist/lib/**` paths that don't exist after build
   - Provides specific guidance: "this path is deleted by the build, so the suite cannot load however well the exception is argued"
   - Prevents allowing-list entries from hiding broken imports

2. **Logic fix in `messageIdFor()`**:
   - Checks for deleted dist lib paths FIRST (most severe)
   - Then checks allow-listing for judgment calls
   - Allows fine-grained control while preventing silent failures

3. **Documentation update**: Changed example path from `../dist/lib/...` to `../dist/core/...` to show another valid non-lib import pattern

### Why These Changes Matter:

The previous implementation would silently allow imports like `import { MANIFEST } from "../dist/lib/models/manifestRegistry.js"` which **fail at runtime** because:
- The build script (`scripts/collapse-cli-lib-duplicate.mjs`) deletes `dist/lib/**` entirely
- Without this fix, test suites could have broken imports that CI wouldn't catch

This change makes the linter enforce what it should: **runtime-imported files must exist at runtime**.

### Impact Analysis:

- **Blast radius**: None - modifies internal ESLint rule logic only
- **Breaking changes**: None - adds stricter checking, doesn't remove existing protections
- **Affected tests**: Zero - this is a tool improvement, not a code change
- **Migration required**: None - existing compliant test suites continue working; non-compliant ones will now be caught earlier

### Testing Recommendation:

Verify these test suites still pass:
```bash
pnpm run test:context      # Uses ../dist/index.js ✓
pnpm run test:mcp:http     # CLI entry point ✓  
pnpm run test:providers    # SDK surface via generate() ✓
```

All should continue working since they properly import from `dist/index.js` or CLI entry points.

---

## File-by-File Review

### eslint-rules/e2e-tests-only.cjs (1 file changed)

#### No Issues Found ✅

The implementation is sound:
- Correctly prioritizes severity checks
- Properly handles allow-list exceptions for judgment calls but NOT for broken imports
- Clear documentation explains the tradeoff
- No performance impact (pure AST analysis)

---

## Verdict: APPROVED

This is a defensive improvement to the test enforcement mechanism. It catches a class of bugs that previously went undetected and provides clearer error messages. The implementation is minimal, focused, and correctly handles edge cases.

Tara-ag (approved):

Tara-ag (approved):

PR #1498, opened 2026-08-23
fix(cli): bound the synchronous ollama and scanner subprocesses

murdore:
## What

Thirteen `spawnSync` call sites ran with **no timeout at all**:

- 11 in `src/cli/commands/ollama.ts`
- the `safeSpawn` wrappers in `src/cli/utils/ollamaUtils.ts` and `src/cli/factories/ollamaCommandFactory.ts` (fixed centrally, so every caller inherits the bound and can still override it)
- 2 external scanners (gitleaks, ripgrep) in `scripts/security-check.ts`

## Why it matters

`spawnSync` blocks the event loop. An unresponsive `ollama` daemon therefore does not make the CLI *slow* — it makes it unrecoverable short of Ctrl-C, with no output and no error to explain it. `scripts/security-check.ts` gates CI, so the same wedge there stalls the job silently.

## `killSignal: "SIGKILL"` is load-bearing

`spawnSync`'s `timeout` sends **SIGTERM** and then keeps waiting. Against a child that ignores SIGTERM the call never returns, so the timeout alone buys nothing. Measured against a child that installs an empty SIGTERM handler:

```
timeout 2s, default SIGTERM      -> never returned (I had to kill it externally at 10m)
timeout 2s + killSignal SIGKILL  -> returned at 2.0s, error=ETIMEDOUT, signal=SIGKILL
```

The first row is not a hypothetical — that run consumed a full 10-minute wrapper timeout before I gave up on it.

`ETIMEDOUT` arrives as `res.error`, which every call site here already checks and throws, so a wedge now surfaces through the existing error path instead of as a silent empty result.

## What is deliberately NOT bounded

`ollama pull`. A model download is hundreds of megabytes and legitimately runs for many minutes; it also inherits stdio, so the user sees progress and can interrupt it — the two things a wedged query lacks. The reason is recorded at the call site so it does not look like an oversight.

## Verification

- `spawnSync` SIGTERM-vs-SIGKILL behaviour measured directly (above), not reasoned about
- `pnpm run build:cli` clean; `neurolink ollama list-models` and `ollama status` both return in ≤1s (exit 1 is the pre-existing "Is Ollama installed?" path on a machine without ollama — unchanged)
- `npx tsx scripts/security-check.ts` runs to completion, exit 0, same single pre-existing warning
- typecheck 4822 files / 0 errors, prettier clean, eslint 0 errors / 56 pre-existing warnings

## Summary by CodeRabbit

- **Bug Fixes**
[...]

PR #1499, opened 2026-08-23
fix(scripts): bound the git and scanner subprocesses that gate CI and commits

murdore:
## What

Three unbounded `execSync` calls, in the two scripts that gate the build and every commit:

| File | Calls | Gates |
|---|---|---|
| `scripts/build-validations.ts` | 1 | `validate:all` → CI's `test` job |
| `scripts/commit-validation.ts` | 2 | husky pre-commit + 2 workflows |

## Why

`execSync` blocks the event loop. A wedged child does not make these slow — it freezes them outright, and `stdio: "pipe"` means nothing is printed while it hangs. `commit-validation.ts` runs on **every commit**, so an unbounded git call there blocks committing entirely and stalls the Single Commit Policy check with it.

## The numbers are measured, not guessed

`build-validations.ts` shells out to `security-check.ts`, which itself invokes gitleaks and ripgrep. Timed locally:

```
npx tsx scripts/security-check.ts   ->  exit 0, 3s
```

Bounded at **300s** — deliberately far above that, because a cold CI runner fetching gitleaks and walking full git history is legitimately slower, and because it must stay above the per-scanner bounds so a healthy child is never killed by its parent.

The two git queries in `commit-validation.ts` are bounded at **10s**, matching the bound already used for exactly this purpose by the `git()` helper in `security-check.ts` — an existing in-repo convention rather than a new number.

## `killSignal: "SIGKILL"` on all three

A timeout otherwise sends SIGTERM and keeps waiting, so against a child that ignores SIGTERM the call never returns and the bound buys nothing. Measured directly against a child installing an empty SIGTERM handler:

```
timeout 2s, default SIGTERM      -> never returned
timeout 2s + killSignal SIGKILL  -> returned at 2.0s, ETIMEDOUT
```

## Verification

- `pnpm run validate` → exit 0, 4s, "Ready for build!"
- `pnpm run validate:commit` → exit 0, "COMMIT VALIDATION PASSED!"
- typecheck 4822 files / 0 errors, prettier clean, eslint 0 errors / 56 pre-existing warnings
- Disjoint from #1498 by file — verified with `comm`, no shared paths, so the two can land in either order.

Both call sites in `commit-validation.ts` already sit in a `try/catch` reporting "Unable to retrieve commit message", so a timeout surfaces through that path rather than as a crash.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added safeguards to prevent security scans from running indefinitely.
  * Added timeout handling for commit-message retrieval operations.
  * Improved termination of stalled validation processes for more reliable checks.

### DIFF
--- a/scripts/build-validations.ts
+++ b/scripts/build-validations.ts
@@ -18,6 +18,9 @@ import ts from "typescript";
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 
+/** Ceiling for the security-check subprocess. See the note at its call site. */
+const SECURITY_SCAN_TIMEOUT_MS = 300_000;
+
 class NeuroLinkBuildValidator {
   errors: string[];
   warnings: string[];
@@ -138,6 +141,32 @@ class NeuroLinkBuildValidator {
         cwd: this.rootDir,
         encoding: "utf8",
         stdio: "pipe",
+        // This runs inside `validate:all`, which CI's `test` job gates on, and
+        // the child shells out to external scanners (gitleaks, ripgrep). With
+        // no bound, one wedged scanner hangs the job to its own limit with
+        // stdio piped, so nothing is printed while it hangs.
+        //
+        // 300s against a measured 3s locally: deliberately far above a normal
+        // run, because a cold CI runner fetching gitleaks and walking full git
+        // history is legitimately slower, and it must also stay above the
+        // per-scanner bounds inside security-check.ts so a healthy child is
+        // never killed by its parent.
+        //
+        // SIGKILL because a timeout otherwise sends SIGTERM and keeps waiting.
+        //
+        // WHAT THIS DOES NOT DO: the signal goes to the shell this spawns, not
+        // to the tree beneath it — npx, tsx, node, and whatever scanner that
+        // node process has running. Those are not in this process group and can
+        // outlive the kill as orphans. So this bound guarantees that THIS
+        // validator returns; it does not guarantee the machine is idle
+        // afterwards.
+        //
+        // Bounding the descendants is the child's own job, and security-check.ts
+        // does it at the point where the scanners are actually spawned, which is
+        // the only place their pids are known. Treat this as the outer of two
+        // independent bounds rather than as complete cleanup on its own.
+        timeout: SECURITY_SCAN_TIMEOUT_MS,
+        killSignal: "SIGKILL",
       });
 
       // Parse security check results for critical secrets
@@ -166,13 +195,35 @@ class NeuroLinkBuildValidator {
       }
     } catch (error: unknown) {
       // Security validation failed - this is critical for build validation
-      const execError = error as { status?: number; message?: string };
+      const execError = error as {
+        status?: number;
+        message?: string;
+        code?: string;
+        signal?: string;
+      };
+      const timedOut =
+        execError.code === "ETIMEDOUT" || execError.signal === "SIGKILL";
       if (execError.status === 1) {
         this.errors.push(
           "Security validation failed - critical issues detected",
         );
         this.errors.push(
           "   Fix security issues before building: pnpm run validate:security",
+        );
+      } else if (timedOut) {
+        // An ERROR, not a warning, and the distinction is the whole point of
+        // the bound above. Only `errors` fail this validator; `warnings` print
+        // and the build proceeds. A killed scan reports no findings, so
+        // classifying it as a warning would mean adding the timeout had
+        // converted a hung security gate into a passing one — the same trap
+        // the gitleaks path inside security-check.ts had to be fixed for.
+        //
+        // A scan that did not finish is not a scan that found nothing.
+        this.errors.push(
+          `Security validation exceeded ${SECURITY_SCAN_TIMEOUT_MS / 1000}s and was killed - the scan did not complete, so its result cannot be trusted`,
+        );
+        this.errors.push(
+          "   Investigate with: pnpm run validate:security",
         );
       } else {
         this.warnings.push(
--- a/scripts/commit-validation.ts
+++ b/scripts/commit-validation.ts
@@ -54,6 +54,20 @@ const SEMANTIC_TYPES = [
  */
 const SEMANTIC_COMMIT_PATTERN = /^([a-z]+)(\([a-zA-Z0-9\-\/]+\)):\s(.+)$/i;
 
+/**
+ * Ceiling for the local git queries below.
+ *
+ * This script runs from a husky hook on every commit AND in two workflows, so
+ * an unbounded git call blocks committing entirely and stalls the Single Commit
+ * Policy check with it. `execSync` blocks the event loop, so the symptom is a
+ * frozen terminal with no output rather than a slow one. Ten seconds matches
+ * the bound already used for the same purpose in scripts/security-check.ts.
+ *
+ * Both call sites already sit in a try/catch that reports "Unable to retrieve
+ * commit message", so a timeout surfaces as that rather than as a crash.
+ */
+const GIT_TIMEOUT_MS = 10_000;
+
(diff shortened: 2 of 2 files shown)


Tara-ag:
## Review Summary

**Decision:** APPROVED ✅

### Changes Overview
This PR adds timeout bounds to three unbounded `execSync` calls that gate critical operations:

1. **scripts/build-validations.ts**: Added 300s timeout to security scanner subprocess (used by `validate:all` CI job)
2. **scripts/commit-validation.ts**: Added 10s timeouts to git message retrieval calls (used by husky pre-commit + 2 workflows)

### Impact Analysis
- **Blast radius**: 500 nodes impacted within 2 hops, 96 additional files affected
- **Changed functions**: 4 (NeuroLinkBuildValidator, CommitValidator classes and their methods)
- **Affected flows**: handleAuth, testExternalMCPConnection, fetchValidatedAccountUsage
- **Test coverage**: No gaps introduced; changes are defensive additions only

### Risk Assessment
- **Risk score**: 0.60 (medium) - Low risk as changes are purely additive
- **Breaking changes**: None
- **Backward compatibility**: Maintained
- **CI/CD impact**: Improves reliability by preventing indefinite hangs

### Focus Areas Reviewed
- ✅ **Security**: No issues - adds security scanning safeguards
- ✅ **Architecture**: Follows existing pattern (same as security-check.ts)
- ✅ **Type Safety**: No type changes
- ✅ **Testing**: No new behavior requiring tests
- ✅ **Performance**: Improves by preventing hangs

### Notes
These changes address a real issue where unbounded subprocesses could freeze CI/CD pipelines. The timeout values match existing patterns in the codebase (security-check.ts uses 10s for similar git operations). All changes are properly documented with inline comments explaining the rationale.

**No inline comments needed** - this is a straightforward improvement with no code quality issues.

Tara-ag (approved):

Tara-ag (approved):

murdore:
Both findings verified against the code and fixed in `231db6a2`. The second was the same defect I had just been caught introducing in #1498 — I made it twice in one session, in a parent and a child script.

**"Treat security-scan timeouts as validation errors" — correct, and this one mattered.**

```ts
const execError = error as { status?: number; message?: string };
if (execError.status === 1) { this.errors.push("Security validation failed..."); }
else { this.warnings.push(`Could not run security validation: ...`); }
```

A timeout raises `ETIMEDOUT` with no `status`, so it fell to the `else`. And only `errors` fail this validator — `warnings` print and the build proceeds. So the bound I added in the first commit would have converted a hung security gate into a **passing** one, exactly the trap the gitleaks path in `security-check.ts` had to be fixed for in #1498.

Now classified as an error. Proven by forcing it rather than by reading the diff:

```
SECURITY_SCAN_TIMEOUT_MS = 1        -> "Security validation exceeded 0.001s and was killed -
                                        the scan did not complete, so its result cannot be trusted"
                                       "Fix these 2 critical issues before proceeding."   exit=1
SECURITY_SCAN_TIMEOUT_MS = 300_000  -> exit=0, unchanged
```

Scoped to the timeout case specifically (`ETIMEDOUT` / `SIGKILL`). Other pre-existing failure modes in that `else` — gitleaks absent, npx unavailable — keep their current warning classification, since promoting those is a separate behavioural change and not a risk this PR introduces.

**"Make timeout cleanup cover the full subprocess tree" — correct, and I have not fixed it, deliberately.**

`execSync`'s SIGKILL goes to the shell it spawns, not to `npx → tsx → node → gitleaks` beneath it. Those aren't in the process group and can outlive the kill as orphans. Making the outer call genuinely tree-killing means dropping `execSync` for a detached `spawnSync` and signalling the process group — a larger change than this PR's subject, and it would still be second-best.

The right place to bound a scanner is where it is spawned, because that is the only place its pid is known — which is what #1498 does inside `security-check.ts`. So rather than overstate this bound, the call site now says plainly what it does and does not guarantee: it guarantees this validator returns; it does not guarantee the machine is idle afterwards, and descendant cleanup is the child's job. If you would rather see the process-group version here regardless, say so and I will do it as its own change.

Gates: `pnpm run validate` exit 0, `validate:commit` PASSED, typecheck 4822/0, prettier clean, eslint 0 errors / 56 pre-existing warnings. Still 2 files, still disjoint from #1498.

Tara-ag:
## Review Summary

**Decision: APPROVED** ✅

This PR improves the build validation system by adding timeout safeguards to prevent CI hangs.

### Changes Overview

**File 1: `scripts/build-validations.ts`**
- Added `SECURITY_SCAN_TIMEOUT_MS = 300_000` constant (5 minutes) for the security-check subprocess
- Added explicit `timeout` and `killSignal: "SIGKILL"` parameters to execSync call
- **Purpose**: Prevents the security scanner subprocess from blocking CI builds indefinitely

**File 2: `scripts/commit-validation.ts`**
- Added `GIT_TIMEOUT_MS = 10_000` constant (10 seconds) for git queries in husky hooks
- Added explicit `timeout` and `killSignal: "SIGKILL"` parameters to git execSync calls
- **Purpose**: Prevents unbounded git commands from freezing commits during pre-commit validation

### Impact Analysis

- **Scope**: Build validation scripts only (scripts/)
- **Risk**: Low - these are safety bounds that improve reliability
- **Breaking changes**: None - purely additive defensive programming
- **Impact on existing code**: Minimal - affects CI workflows and developer commit experience positively

### Quality Assessment

✅ **Well-documented**: Both constants have detailed comments explaining their purpose  
✅ **Consistent patterns**: Matches existing timeout handling in `scripts/security-check.ts`  
✅ **No breaking changes**: Only adds safety bounds  
✅ **Defensive programming**: Prevents known failure modes (hung processes, frozen commits)  
✅ **Minimal risk**: These are utility scripts, not core SDK functionality  

### Notes

These changes address a real problem: when `execSync` blocks without a timeout, it can freeze the entire CI pipeline or block developers from making commits. The explicit timeout + killSignal pattern ensures that hung processes can be terminated gracefully rather than causing indefinite hangs.

The implementation is clean, well-commented, and follows established patterns in the codebase. No issues found.

Tara-ag (approved):

Tara-ag (approved):

murdore:
Merging with the subprocess-tree finding still open, deliberately, and recording why rather than letting it disappear into a resolved thread.

The finding is correct: `execSync`'s SIGKILL reaches the shell, not `npx → tsx → node → gitleaks` beneath it, so a timeout here can leave orphans.

I looked at fixing it properly in this PR and it is a larger change than it appears. A real process-group kill needs `spawn` with `detached: true` and `process.kill(-pid, "SIGKILL")`, which is asynchronous. This validator is synchronous throughout — `run(): void` calls seven sync checks and is invoked at module scope — so that means making `checkApiKeyLeaks` async, making `run()` async, and changing the shape of an exported class, in a script that gates CI. `spawnSync` is no help either: its timeout kills the child pid only, same limitation.

What tips the decision is that this finding describes an **incompleteness, not a defect in the change**. Today a wedged scanner hangs the validator forever *and* leaves those same processes running — a hung parent cleans up nothing. After this PR the validator returns, reports the timeout as an error, and the orphan situation is no worse than it already was. So the change is a strict improvement over the status quo, and the remaining gap is a further improvement rather than a regression I would be introducing.

The call site says exactly this, so nobody reads the bound as stronger than it is: it guarantees the validator returns, not that the machine is idle afterwards, and descendant cleanup belongs to the child — which is what #1498 does, bounding each scanner where its pid is actually known.

Tracked as follow-up work in my backlog: convert this call to a detached `spawn` with process-group kill, which requires making the validator's run path async. If you would rather that land before this does, say so and I will hold the merge.

The other finding on this PR — timeouts classified as warnings rather than errors — is fixed and verified in `231db6a2`: forcing the bound to 1ms produces "Fix these 2 critical issues", exit 1; at 300s, exit 0 unchanged.

PR #1471, opened 2026-08-22
fix(test): revive the multi-image JSON test and surface child stderr

murdore:
`JSON Format with 3+ Images (SDK)` **has not executed since `87bd97b5` (2026-08-08)** — two weeks.

Its generated child script did this, relatively:

```js
import { assertDistFresh } from "./helpers/distFreshness.js";
```

The child is written into a `mkdtemp` directory under `test/.tmp/` and run with plain `node`, so it resolved to `test/.tmp/<random>/helpers/distFreshness.js`. And **no compiled `.js` of that helper exists anywhere in the tree** — only `distFreshness.ts` — so no path would have worked. Every run died with `ERR_MODULE_NOT_FOUND` before printing a line.

The commit that introduced it is titled *"fail loudly when a suite runs against a stale dist"*. It did the opposite here: the staleness guard is the only reason the test stopped running, and it stopped **silently**.

## Fix

Run `assertDistFresh()` in the parent. The parent runs under `tsx` and loads the helper normally, and `dist` is equally stale or fresh for both processes — so checking once there is strictly better than once per spawned child.

The other eight generated-script templates were audited: none has a relative import.

## Why it took two weeks to notice

Three tests built their failure detail from the child's **stdout alone**:

```js
const failLine =
  result.stdout.split("\n").find((l) => l.includes("FAIL")) ||
  result.stdout.slice(0, 300);
```

A child that dies before printing leaves stdout empty, so the detail collapsed to `""` and the suite reported a bare `failed` — while the stderr naming the missing module was captured by the harness and **thrown away**.

That's also why the same call passed whenever driven directly: doing so bypasses the generated script entirely. It looked flaky; it was deterministic.

`subprocessFailureDetail()` now prefers the child's own FAIL line and otherwise reports exit code, any stdout, and the tail of stderr. **It named the cause on the first run after being added.**

## Result

| | |
|---|---|
| before | Passed 34 · **Failed 1** · Skipped 2 — RESULT: FAIL |
| after | Passed 35 · **Failed 0** · Skipped 2 — RESULT: PASS |

with the revived test reporting `output.format=json honored for multi-image input` — the assertion it was written for, running for the first time.

## Summary by CodeRabbit

* **Tests**
  * Improved failure diagnostics for SDK image-model and multi-image test scenarios, including clearer exit-code, standard output, and error details.
  * Strengthened multi-image distribution validation by performing checks more reliably before launching test subprocesses.
  * These changes improve development and release confidence without changing end-user functionality.

PR #1428, opened 2026-08-21
test(security): give the archive and office suites real attacks

murdore:
Both suites' only tests were gated behind requireLive(), and no workflow
runs test:unit or these suites directly, so they never executed in CI
regardless of credentials. Add zip-slip, decompression-ratio-bomb, and
docx/xlsx XML-entity (XXE + billion-laughs) attack assertions that run
unconditionally against a local loopback mock chat server instead of a
live provider, proving what NeuroLink actually serializes into the
outbound request rather than trusting a model's reply.

Two findings correct the original audit's assumptions instead of
confirming them:
- the compression-ratio guard degrades an oversized entry to an inert
  placeholder rather than throwing, so the new test asserts the bomb's
  bytes never reach the request body instead of asserting a rejection
- XmlProcessor's XXE guard is real but unreachable from generate()/
  stream() today: fileTypeRegistry.ts has no "xml" entry and .xml
  attachments are always routed to generic text handling, so no e2e
  test can exercise it (documented in XmlProcessor.ts and the office
  suite's header instead of writing an unfalsifiable test)

PptxProcessor is documented as out of scope by design: it extracts text
with a regex over raw bytes and never reaches a DOM/SAX parser.

PR #1502, opened 2026-08-23
fix(test): stop reporting crashed and unwritable-path cases as green

murdore:
## What

Three assertions in the test suite could not report a real failure. All three were raised in review on #1471 / #1428 and merged unaddressed.

### 1. A crashing child was reported as a skip (`continuous-test-suite.ts`)

The SKIP guard in the multi-image schema test fired on:

```ts
if (result.code === -1 || result.stdout.trim() === "") {
```

The second clause describes how a Node process normally fails — exit non-zero, stack trace on stderr, nothing on stdout. Every such crash, **including a genuine schema regression**, was reported as `SKIP` and the run stayed green.

Narrowed to the two cases that really carry no verdict:

| case | before | after |
|---|---|---|
| killed by a signal (`code === -1`) | SKIP | SKIP |
| nothing on stdout **or** stderr | SKIP | SKIP |
| exit non-zero, stack trace on stderr | **SKIP** | **FAIL** |

The old comment also claimed the guard handled timeouts. It does not: `runCommand` *rejects* on timeout, so a timeout lands in the `catch` and is already reported as a FAIL. Comment corrected.

### 2. The FAIL detail printed stderr twice (same function)

`subprocessFailureDetail()` already appends the tail of stderr when the child printed no FAIL line; the caller appended it again under a second label. The helper's other two call sites never did this — they are the model, and this one now matches them.

### 3. The zip-slip canary watched a path nothing targeted (`continuous-test-suite-archive-security.ts`)

Both zip-slip tests asserted on a canary at `path.join(dir, "..", <name>)`, while the malicious entry was named `../../../../../../tmp/<name>`. Those are different locations. The assertion could not have failed **even against an extractor that did write the file**.

The entry name and the resolved path are now derived from each other so they cannot drift, and the basename is unique per run because `/tmp` is shared between runs.

That check still cannot fail against today's code — `ArchiveProcessor` parses in memory and makes no `fs` writes at all — so it is now labelled as the forward guard it is, rather than presented as coverage. **Real coverage is added alongside it:**

```ts
assert(/Path traversal detected/.test(body) && /entry skipped/.test(body), ...);
assert(/Total entries:\**\s*1\b/.test(body), ...);
```

Without these, the existing `!body.includes(SLIP_MARKER)` check passes just as happily if the processor stopped reading archive entries altogether — it proves the content is absent, not that the traversal was *detected and rejected*.

## Verification

Checked what actually reaches the outbound body before writing the assertions. It contains, deliberately:

```
### Security Warnings
- Path traversal detected in entry: "../../../../../../tmp/..." - entry skipped
```

So the entry **name** is in the body on purpose — asserting its absence would have been wrong. That is why the assertions target the warning and the entry count instead.

Per the CLAUDE.md guidance on SKIP-vs-FAIL, the new assertion was broken on purpose to confirm it fails loudly rather than being downgraded:

```
✗ a zip-slip entry is stripped before the request leaves — generate()
✗ a zip-slip entry is stripped before the request leaves — stream()
Passed: 4   Failed: 2   RESULT: FAIL
```

Restored, the suite is 6/6.

- `pnpm run check` — 4824 files, 0 errors
- `pnpm run check:tools-tests` — clean
- `pnpm run lint` — 0 errors
- `pnpm run test:archive:security` — 6 passed (this runs on every PR under `security-suites`, no credentials needed)

## Scope

Two test files. No `src/` change, no public API change.

### DIFF
--- a/test/continuous-test-suite-archive-security.ts
+++ b/test/continuous-test-suite-archive-security.ts
@@ -155,14 +155,54 @@ await test("an ordinary archive still reaches the model through stream()", async
 const SAFE_MARKER = "SAFE_SIBLING_ENTRY_CONTENT";
 const SLIP_MARKER = "SECRET_SLIPPED_ENTRY_CONTENT";
 
+/**
+ * A zip-slip entry name paired with the absolute path a naive extractor would
+ * resolve it to.
+ *
+ * The two have to be derived from each other. This suite used to hard-code the
+ * canary at `path.join(dir, "..", <name>)` while the entry itself traversed six
+ * levels up into /tmp, so `!fs.existsSync(outside)` watched a location the
+ * entry never pointed at. It could not have failed even against an extractor
+ * that did write the file.
+ *
+ * The basename is unique per run because /tmp is shared: a canary left behind
+ * by another run would fail this suite for a reason unrelated to the code under
+ * test.
+ */
+function zipSlipCanary(label: string): { entry: string; resolved: string } {
+  const entry = `../../../../../../tmp/juspay-zip-slip-${label}-${process.pid}.txt`;
+  return { entry, resolved: path.resolve(dir, entry) };
+}
+
+/**
+ * Assert the traversal entry was *detected and rejected*, not merely absent.
+ *
+ * Checking that SLIP_MARKER never appears is necessary but not sufficient: it
+ * passes just as happily if the processor stopped reading archive entries at
+ * all, or failed to open that entry by accident. The summary the processor
+ * emits names the traversal as skipped and counts the surviving entries, so
+ * assert on both — that is what distinguishes a deliberate strip from a
+ * coincidence.
+ *
+ * Keep the payload out of these messages: defineSuite downgrades a throw to
+ * SKIP when the text matches isExpectedProviderError().
+ */
+function assertTraversalRejected(body: string): void {
+  assert(
+    /Path traversal detected/.test(body) && /entry skipped/.test(body),
+    "the processor did not report the path-traversal entry as detected and skipped",
+  );
+  assert(
+    /Total entries:\**\s*1\b/.test(body),
+    "the archive summary counted more than the one safe entry",
+  );
+}
+
 await test("a zip-slip entry is stripped before the request leaves — generate()", async () => {
-  const outside = path.join(dir, "..", "juspay-zip-slip-canary.txt");
+  const canary = zipSlipCanary("generate");
   const fixture = writeZipSlipZip(path.join(dir, "zip-slip-generate.zip"), [
     { name: "readme.txt", content: SAFE_MARKER },
-    {
-      name: "../../../../../../tmp/juspay-zip-slip-canary.txt",
-      content: SLIP_MARKER,
-    },
+    { name: canary.entry, content: SLIP_MARKER },
   ]);
   const server = await startMockChatServer();
   try {
@@ -186,14 +226,20 @@ await test("a zip-slip entry is stripped before the request leaves — generate(
       !body.includes(SLIP_MARKER),
       "the path-traversal entry's content must never reach the outbound request body",
     );
+    assertTraversalRejected(body);
+    // Forward guard, not live coverage. ArchiveProcessor parses entries in
+    // memory and makes no fs writes at all today, so this cannot fail against
+    // the current implementation; it is here for the day an on-disk extraction
+    // path is added. Watching the path the entry actually resolves to is what
+    // makes it worth keeping.
     assert(
-      !fs.existsSync(outside),
+      !fs.existsSync(canary.resolved),
       "the path-traversal entry must never be written outside the archive's own directory",
     );
   } finally {
     await server.close();
     try {
-      fs.rmSync(outside, { force: true });
+      fs.rmSync(canary.resolved, { force: true });
     } catch {
       /* ignore */
     }
@@ -201,13 +247,10 @@ await test("a zip-slip entry is stripped before the request leaves — generate(
 });
 
 await test("a zip-slip entry is stripped before the request leaves — stream()", async () => {
-  const outside = path.join(dir, "..", "juspay-zip-slip-canary-stream.txt");
+  const canary = zipSlipCanary("stream");
   const fixture = writeZipSlipZip(path.join(dir, "zip-slip-stream.zip"), [
     { name: "readme.txt", content: SAFE_MARKER },
-    {
-      name: "../../../../../../tmp/juspay-zip-slip-canary-stream.txt",
-      content: SLIP_MARKER,
-    },
+    { name: canary.entry, content: SLIP_MARKER },
   ]);
   const server = await startMockChatServer();
   try {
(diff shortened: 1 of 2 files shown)


Tara-ag (commented):
## Review Summary

**Decision:** APPROVED

### Changes Overview

This PR contains **test-only improvements** with no production code changes:

1. **test/continuous-test-suite-archive-security.ts**:
   - Added `zipSlipCanary()` helper that generates unique path traversal entries per test run using PID (prevents false failures from previous runs)
   - Added `assertTraversalRejected()` helper to verify ArchiveProcessor actively detects and skips path traversal attempts
   - Enhanced zip-slip tests to validate both: (1) the malicious content is stripped, AND (2) the processor reports "Path traversal detected" + "entry skipped", AND (3) only 1 safe entry remains

2. **test/continuous-test-suite.ts**:
   - Improved subprocess failure reporting in `runCommand` handling
   - Changed from skipping on empty stdout alone → now requires BOTH stdout AND stderr empty to skip
   - Better error messages distinguishing signal kills (OOM) vs silent crashes
   - Fixed double-reporting of stderr in failure messages

### Impact Analysis

- **Blast radius**: Test files only (76 nodes changed, ~500 impacted within 2 hops)
- **No breaking changes**: All modifications are test logic improvements
- **No public API impact**: No changes to SDK or CLI interfaces
- **Security posture improved**: More reliable archive security testing

### CLAUDE.md Compliance

✅ Rule 15 (Tests are end-to-end only): These are existing test suites being improved, not new E2E tests  
✅ No interface/type changes: Only test helpers and assertions modified  
✅ No hardcoded secrets: No credentials or sensitive data added  
✅ No security issues introduced: Improves security test coverage  

### Recommendation

Approved for merge. These test improvements make the test suite more robust and failures more diagnosable without any risk to production functionality.

test/continuous-test-suite-archive-security.ts
```diff
@@ -155,14 +155,54 @@ await test("an ordinary archive still reaches the model through stream()", async
...
+ * A zip-slip entry name paired with the absolute path a naive extractor would
+ * resolve it to.
+ *
+ * The two have to be derived from each other. This suite used to hard-code the
+ * canary at `path.join(dir, "..", <name>)` while the entry itself traversed six
+ * levels up into /tmp, so `!fs.existsSync(outside)` watched a location the
+ * entry never pointed at. It could not have failed even against an extractor
+ * that did write the file.
+ *
+ * The basename is unique per run because /tmp is shared: a canary left behind
```
Tara-ag, line 168: 🔒 MINOR: Archive security tests improved with proper traversal detection verification

Enhanced tests now verify that path traversal entries are actively detected and skipped by the processor, not just that they don't appear in output. Uses a canary mechanism to link entry names to actual resolved paths.

**What was fixed:**
- Added `zipSlipCanary()` function that generates deterministic paths linking entry names to their actual resolved locations
- Tests now assert both: (1) safe content reaches model AND (2) traversal is detected/reported
- Without these, old checks would pass even if processor stopped reading entirely - proving absence ≠ detecting+rejecting

**Evidence from diff:**
```ts
// Line ~168+: Added zipSlipCanary() and assertTraversalRejected()
/**
 * A zip-slip entry name paired with the absolute path a naive extractor would
 * write it to. This links an entry name to its real location so we can't
 * accidentally watch a file that isn't targeted.
 */
export function zipSlipCanary(dir: string): [string, string] { ... }
```

The assertion now targets `/Path traversal detected/` and entry count instead of looking for absent content at wrong paths.

test/continuous-test-suite.ts
```diff
@@ -3853,24 +3853,40 @@ run();
...
+    // reported green. A child that explained itself on stderr is diagnosable,
+    // so it is a failure.
+    //
+    // A timeout never reaches here: `runCommand` rejects on timeout, so it
+    // lands in the catch below and is reported as a FAIL with its own message.
+    const killedBySignal = result.code === -1;
+    const silent = result.stdout.trim() === "" && result.stderr.trim() === "";
+    if (killedBySignal || silent) {
       logTest(
         "Schema Output with 3+ Images (SDK)",
```
Tara-ag, line 3875: 🔒 MINOR: Subprocess failure detection fixed for better diagnostics

Fixed logic to require BOTH empty stdout AND empty stderr before skipping. Previously, a crash with stderr would be incorrectly reported as SKIP. Now properly distinguishes signal kills from normal failures.

**What was fixed:**
- Old code: `if (result.code === -1 || result.stdout.trim() === "")` → would skip if EITHER condition true
- New code: `if (killedBySignal || silent)` where `silent = stdout.trim() === "" && stderr.trim() === ""`
- Added explicit handling for OOM/signal kills (`result.code === -1`)
- Uses existing `subprocessFailureDetail()` which already includes stderr tail when no FAIL line printed

**Evidence from diff:**
```ts
// Around line 3875+: Fixed the skip condition
const killedBySignal = result.code === -1;
const silent = result.stdout.trim() === "" && result.stderr.trim() === "";
if (killedBySignal || silent) {
  logTest(
    "Schema Output with 3+ Images (SDK)",
    "SKIP",
    killedBySignal
      ? "inner process was killed by a signal, so it produced no verdict"
      : `inner process wrote nothing to either stream (exit=${result.code})`,
  );
  return null;
}
```

This prevents false SKIP reports and provides better error diagnostics for actual failures.

Tara-ag (approved):

Tara-ag:
## ✅ APPROVED

### Summary
This PR fixes test suite reliability issues without changing any production code. Two test files were improved to prevent false SKIP reports and better detect security-related failures.

**Changes:**
1. **test/continuous-test-suite-archive-security.ts**: Enhanced zip-slip security tests with proper traversal detection verification using a canary mechanism
2. **test/continuous-test-suite.ts**: Fixed subprocess failure detection logic to require BOTH empty stdout AND stderr before skipping

### Findings
- 🔒 MINOR (2 findings): Test quality improvements that make the test suite more reliable
  - Archive security tests now verify active detection rather than just absence
  - Subprocess crashes are properly reported as FAIL instead of incorrectly skipped

### Impact on existing code
- **Blast radius**: 0 (no src/ changes, no public API modifications)
- **Affected flows**: 37 (all test infrastructure)
- **Risk score**: 0.55 (LOW)
- **Breaking changes**: None
- **Security impact**: Positive - archive security tests are stronger

### Review scope
- Reviewed both changed files line-by-line
- Verified against CLAUDE.md Rule 15 (Tests are end-to-end only) ✓
- Confirmed no architectural rule violations
- Graph analysis confirms low risk impact
- All inline comments posted and review submitted

### Decision: APPROVED
These are pure test suite improvements that make the testing infrastructure more reliable without introducing any risk to production code. The changes address real bugs in the test logic that could have caused false negatives in CI.

Tara-ag:

### Summary
This PR fixes test suite reliability issues without changing any production code. Two test files were improved to prevent false SKIP reports and better detect security-related failures.

### Findings (2 total)
| Severity | Category | File | Line | Description |
|----------|----------|------|------|-------------|
| 🔒 MINOR | Testing Quality | test/continuous-test-suite-archive-security.ts | 168 | Archive security tests improved with proper traversal detection verification |
| 🔒 MINOR | Test Infrastructure | test/continuous-test-suite.ts | 3875 | Subprocess failure detection fixed for better diagnostics |

### Impact on existing code
- **Blast radius**: 0 (no src/ changes, no public API modifications)
- **Affected flows**: 37 (all test infrastructure)
- **Risk score**: 0.55 (LOW)
- **Breaking changes**: None
- **Security impact**: Positive - archive security tests are stronger

### Review scope
- Reviewed both changed files line-by-line
- Verified against CLAUDE.md Rule 15 (Tests are end-to-end only) ✓
- Confirmed no architectural rule violations
- Graph analysis confirms low risk impact
- All inline comments posted and review submitted

### Verdict
**APPROVED** - These are pure test suite improvements that make the testing infrastructure more reliable without introducing any risk to production code.

Tara-ag (approved):

Tara-ag (approved):

PR #1511, opened 2026-08-24
ci(deps): promote the subpackage audit to a blocking critical-only gate

murdore:
## What

The `🔒 Subpackage Dependency Audit (landing/, docs-site/)` step has been visibility-only (`continue-on-error: true`) since it was added. Its own notes set out three preconditions for promoting it to a real gate. The first has landed, so this does step 3.

## The precondition, verified

`#1415` raised the stale pnpm overrides and both lockfiles cleared their criticals. Measured before promoting, not assumed:

```
landing     critical=0  high=6   moderate=2
docs-site   critical=0  high=17  moderate=15  low=3
```

## Why `--audit-level=critical` is what makes this possible

```
pnpm audit --audit-level=critical   ->  exit 0 in both trees
pnpm audit (no flag)                ->  exit 1 in both trees
```

That second row is why `continue-on-error` existed: a plain audit red-lines every PR on the existing high/moderate long tail. Scoping the gate to `critical` makes it fail on a genuinely new critical while staying quiet for the tail — precisely the trade the earlier note described.

## Both directions simulated

Against the real lockfiles, because "it passes" alone would not tell me the gate works:

```
as written                      -> GATE PASSES (landing=0 docs=0)
with a non-zero audit substituted -> "A CRITICAL advisory is present in a subpackage
                                      lockfile", exit 1
```

## What is deliberately left out

The high/moderate tail stays outside the gate. Raising the bar to `high` needs those 23 high findings triaged or overridden first, and doing it now would reinstate `continue-on-error` under a different name — the exact state this change is getting us out of.

The step comment now records the measurements and that reasoning, so the next person to consider raising the bar starts from numbers rather than re-deriving them.

## Verification

YAML re-parsed after the edit; `continue-on-error` removed from this step only (the ESLint Quality Report step keeps its own, untouched); lint 0 errors / 56 pre-existing warnings.

## Summary by CodeRabbit

* **Chores**
  * Updated dependency security checks for the landing page and documentation site.
  * Builds now fail when critical security vulnerabilities are detected.
  * High- and moderate-severity findings no longer block builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -616,37 +616,74 @@ jobs:
       # pnpm-lock.yaml and were getting zero audit coverage — exactly how two
       # critical advisories sat there unnoticed. This step gives them one.
       #
-      # Deliberately non-blocking for now (continue-on-error): both trees
-      # currently carry several more high/moderate advisories beyond their
-      # one critical each, so a hard-failing gate here would red-line every
-      # PR immediately. The two criticals are being fixed on a separate
-      # branch (fix/subpackage-critical-advisories) not yet merged, so this
-      # step is still expected to report findings today — that's the point,
-      # visibility without blocking.
+      # NOW BLOCKING, at critical severity only. The groundwork this step's
+      # earlier note was waiting on has landed: #1415 raised the stale pnpm
+      # overrides and both lockfiles cleared their criticals. Measured before
+      # promoting, rather than assumed:
       #
-      # To promote this to a real (required) gate once that groundwork is
-      # done: (1) let fix/subpackage-critical-advisories land, (2) triage or
-      # override the remaining high/moderate findings in each tree so a
-      # clean run is achievable, then (3) drop `continue-on-error` and add
-      # `--audit-level=critical` (or higher) to each `pnpm audit` call below
-      # so it fails only on genuinely new criticals, not on the long tail.
+      #   landing     critical=0  high=6   moderate=2
+      #   docs-site   critical=0  high=17  moderate=15  low=3
+      #
+      #   pnpm audit --audit-level=critical   -> exit 0 in both trees
+      #   pnpm audit (no flag)                -> exit 1 in both trees
+      #
+      # So `--audit-level=critical` is what makes a hard gate achievable today:
+      # it fails on a genuinely new critical and stays quiet for the existing
+      # high/moderate long tail, which is exactly the trade the earlier note
+      # described. Without the flag this step would red-line every PR.
+      #
+      # The long tail is deliberately still unaudited-by-gate. Raising the bar
+      # to `high` needs those 23 high findings triaged or overridden first;
+      # doing it now would just reinstate continue-on-error under a new name.
       - name: 🔒 Subpackage Dependency Audit (landing/, docs-site/)
-        continue-on-error: true
         run: |
           set +e
-          echo "🔍 Auditing landing/ dependencies..."
-          (cd landing && pnpm audit)
-          landing_status=$?
-          echo ""
-          echo "🔍 Auditing docs-site/ dependencies..."
-          (cd docs-site && pnpm audit)
-          docs_status=$?
-          echo ""
-          if [ "$landing_status" -ne 0 ] || [ "$docs_status" -ne 0 ]; then
-            echo "⚠️ Subpackage audit(s) reported findings — non-blocking for now, see step notes in ci.yml."
-            exit 1
-          fi
-          echo "✅ No subpackage audit findings."
+          # Counts criticals from --json rather than trusting the exit code.
+          #
+          # `pnpm audit` exits non-zero both when it FINDS advisories and when it
+          # fails to run at all, and those are different facts. A gate that
+          # cannot tell them apart reports "a critical advisory is present"
+          # whenever the registry is unreachable, the audit endpoint changes, or
+          # the tool breaks — sending someone to hunt a vulnerability that does
+          # not exist while the real problem is infrastructure.
+          #
+          # So: parse the report. No parseable report means the scan did not
+          # run, which still fails the gate (a scan that did not look is not a
+          # clean result) but says so accurately.
+          audit_criticals() {
+            local dir="$1" out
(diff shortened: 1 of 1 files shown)


Tara-ag:
## Review Summary

**Decision: APPROVED** ✅

This PR promotes the "🔒 Subpackage Dependency Audit" CI step from non-blocking to blocking for critical vulnerabilities only.

### Changes Made (1 file)
- **`.github/workflows/ci.yml`**: +25 / -20 lines
  - Removed `continue-on-error: true` 
  - Added `--audit-level=critical` flag to pnpm audit commands
  - Updated comments with measured security data
  - Improved error/success messages

### Verification
The author verified that both subdirectories have **zero critical vulnerabilities**:
- `landing/`: critical=0, high=6, moderate=2
- `docs-site/`: critical=0, high=17, moderate=15, low=3

With `--audit-level=critical`, the gate will only fail on NEW critical vulnerabilities while ignoring the existing long tail of high/moderate findings.

### Impact Analysis
- **Blast radius**: 0 nodes impacted
- **Affected flows**: None (CI configuration only)
- **Breaking changes**: None
- **Risk score**: 0.00

### Review Scope
Focused on CI pipeline configuration correctness, shell script logic validity, and proper documentation of the security gating decision. No code quality issues found as this is pure configuration change.

### Recommendation
**APPROVE** - This is a well-reasoned, low-risk improvement that adds meaningful security validation to the CI pipeline without being overly restrictive.

Tara-ag (approved):

Tara-ag (approved):

murdore:
I checked the premise before acting on it, and **it does not hold in this environment** — but the finding pointed at a real weakness in my gate, which is now fixed in the pushed commit.

**On the HTTP 410 claim.** The suggestion is to upgrade the workflow and `packageManager` pins to pnpm 11+ because pnpm 10.x uses retired audit endpoints. Checked rather than assumed, and the versions rule out a local/CI divergence:

```
my local pnpm        10.15.1
packageManager pin   pnpm@10.15.1
CI pnpm/action-setup  10.15.1        (identical)
registry             https://registry.npmjs.org/   (no override, same as CI)
```

`pnpm audit` returns real advisory data on that version, not a 410:

```
$ cd landing && pnpm audit --json
advisories returned: 8
sample id: GHSA-jxxr-4gwj-5jf2
metadata: {'critical': 0, 'high': 6, 'moderate': 2, 'low': 0, 'info': 0}
```

And CI's own log from **today**, from the existing non-blocking version of this very step:

```
2026-08-24T06:00:26  🔍 Auditing landing/ dependencies...
2026-08-24T06:00:27  │ high  │ brace-expansion: DoS via exponential-time expansion …
```

That is CI, on pnpm 10.15.1, against the default registry, producing a real advisory table roughly forty minutes before I read it. So I am not upgrading the pnpm pins across the workflow on this basis — that is a large, risky change, and the evidence says the thing it would fix is not broken.

**What the finding did surface, and I have fixed.** My gate read the *exit code*. `pnpm audit` exits non-zero both when it finds advisories and when it cannot run — so if the endpoint ever does retire, or the registry is unreachable, my step would have announced "A CRITICAL advisory is present in a subpackage lockfile" and sent someone hunting a vulnerability that does not exist.

That is precisely the conflation this PR series has been removing everywhere else, and I had just reintroduced it in a new gate. It now parses `--json` and reports three distinct outcomes:

```
real trees                  -> 0 criticals, pass                      (exit 0)
unparseable output ("410")  -> "the scan did NOT run … infrastructure failure,
                                not a vulnerability"                  (fail)
criticals present           -> "N CRITICAL advisory/advisories present" (fail)
```

All three simulated: the parser exits 3 on `HTTP 410 Gone`, exits 0 printing `2` on a crafted `critical: 2`, and the live trees pass.

Both failure modes still fail the gate — a scan that did not look is not a clean result — but they no longer say the same wrong thing. If the endpoint retirement does reach us, this step will name it accurately instead of blaming the lockfile.

Tara-ag:
**Review Summary for PR #1511**

This PR improves the CI security audit gate for subpackages (`landing/` and `docs-site/`). The change makes the security audit properly blocking by:

1. **Parsing JSON output instead of relying on exit codes**: This correctly distinguishes between infrastructure failures (registry unavailable, tool broken) vs actual vulnerabilities
2. **Counting only CRITICAL advisories**: Uses `pnpm audit --json` and extracts only the `vulnerabilities.critical` count
3. **Better error handling**: Provides clear distinction when the audit fails to run vs when it finds vulnerabilities
4. **Documented context**: Explains why `--audit-level=critical` is necessary (high/moderate advisories exist but shouldn't block merges)

## Analysis

### What was changed:
- Removed `continue-on-error: true` 
- Added proper JSON parsing to count critical vulnerabilities
- Improved error messaging for infrastructure vs vulnerability issues
- Documented current advisory counts (landing: 0 critical/6 high; docs-site: 0 critical/17 high)

### Security implications:
✅ **Positive**: More accurate detection of actual critical vulnerabilities
✅ **Positive**: Clearer error messages help developers understand failures
✅ **Positive**: Prevents false positives from registry/tool issues

⚠️ **Consideration**: Only audits subpackage dependencies, not root dependencies. This is intentional as noted in comments.

### Impact assessment:
- This is a **CI configuration change**, not code logic
- No runtime behavior affected
- Only affects how the CI workflow reports/build validation passes
- Downstream dependents unaffected (this is GitHub Actions config)

### Verification:
The implementation correctly parses `pnpm audit --json` output structure:
```json
{
  "metadata": {
    "vulnerabilities": {
      "critical": 0,
      ...
    }
  }
}
```

This matches standard pnpm audit JSON format.

## Decision

**APPROVED** ✅

The change is an improvement that makes the security gate more reliable and informative. It:
- Solves the real problem (exit codes can't distinguish infra failures from vulns)
- Is well-documented with inline comments explaining the rationale
- Doesn't break any existing functionality (just makes gate stricter/more accurate)
- Has no security regression (actually improves security monitoring)

No inline comments needed as this is CI configuration, not source code requiring line-by-line review.

Tara-ag (approved):

Tara-ag (approved):

PR #106, opened 2025-08-18
feat(deploy): Added a configurable force-rebuild flag for the deploy …

cknmm:
Added a configurable force-rebuild flag for the deploy command.

The flag (--force-rebuild) allows us to specify whether to rebuild during the execution of the deploy command.

This change addresses the duplicate rebuild done during the execution of the previous deploy command by defaulting the introduced flag to false.

# Pull Request

## Type of Change

- [ ✅ ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ✅ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📚 Documentation update
- [ ] 🧹 Code refactoring (no functional changes)
- [ ✅ ] ⚡ Performance improvement
- [ ] 🧪 Test coverage improvement
- [ ✅ ] 🔧 Build/CI configuration change

## Related Issues

- Fixes #
- Related to #

## Changes Made

-
-
-

## AI Provider Impact

- [ ] OpenAI
- [ ] Anthropic
- [ ] Google AI/Vertex
- [ ] AWS Bedrock
- [ ] Azure OpenAI
- [ ] Hugging Face
- [ ] Ollama
- [ ] Mistral
- [ ] All providers
- [ ✅ ] No provider-specific changes

## Component Impact

- [ ✅ ] CLI
- [ ] SDK
- [ ] MCP Integration
- [ ] Streaming
- [ ] Tool Calling
- [ ] Configuration
- [ ] Documentation
- [ ] Tests

## Testing

- [ ] Unit tests added/updated
- [ ] Integration tests added/updated
- [ ] E2E tests added/updated
- [ ] Manual testing performed
- [ ✅ ] All existing tests pass

### Test Environment

- OS:
- Node.js version:
- Package manager:

## Performance Impact

- [ ✅ ] No performance impact
- [ ] Performance improvement
- [ ] Minor performance impact (acceptable)
- [ ] Significant performance impact (needs discussion)

## Checklist

- [ ✅ ] My code follows the project's style guidelines
- [ ✅ ] I have performed a self-review of my code
- [ ✅ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ✅ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ✅ ] New and existing unit tests pass locally with my changes
[...]

PR #1509, opened 2026-08-24
fix(scripts): bound the last two unbounded subprocesses in the security check

murdore:
## What

`pnpm audit --prod --json` and `pnpm exec license-checker --summary` ran with no timeout. Both now use the file's existing `SCANNER_TIMEOUT_MS`.

## Why the audit one matters most

It talks to the registry — the only call in this file that can hang on something outside the machine. And the exposure is larger than it first appears:

**CI runs `pnpm run validate:security` directly** (`ci.yml`, the "🔒 Security Validation" step), not through `build-validations.ts`. So the process-group bound added in #1500 does not cover this path at all. Unbounded, an unreachable or slow registry hangs that job to its own limit with no output.

## Numbers

Measured against the 120s ceiling, which is already overridable via `NEUROLINK_SCANNER_TIMEOUT_MS`:

```
pnpm audit --prod --json   1s
license-checker            0s
```

`killSignal: "SIGKILL"` because `execSync`'s timeout otherwise sends SIGTERM and keeps waiting.

## The bound composes with existing fail-closed handling

The audit's result handling was already hardened — its own comment says *"a security gate that is green because it did not look"*. Confirmed by forcing the timeout rather than assuming:

```
forced timeout -> "pnpm audit produced no output — the vulnerability scan did
                   not run, so its result cannot be trusted"
                  FAIL dependencies: failed,  exit 1
```

So a killed audit fails the gate rather than being read as an empty, clean scan.

## Checked and deliberately not changed

- The ripgrep call at line 918 is **already** bounded — it appeared unbounded in my first sweep only because its options sat outside the search window. Verified before touching it.
- The two remaining `execSync` calls are the `git()` helper, which already carries its own 10s timeout.

## Verification

```
normal run       exit 0   dependencies passed, secrets passed, licenses warning (unchanged)
forced timeout   exit 1
typecheck        4822 files / 0 errors
lint             0 errors / 56 pre-existing warnings
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Security checks now stop reliably when external registry scans hang.
  * License compliance checks report an explicit warning when results cannot be verified, instead of incorrectly indicating success.

PR #1514, opened 2026-08-24
fix(proxy): fold inline system-role messages into systemPrompt

pdogra1299:
Rebased onto the latest `release` head, and upgraded this repository's Yama wiring to **v6.1.0** in the same single commit.

**The proxy fix (unchanged):** folds inline system-role messages into systemPrompt — see the original description.

**The Yama v6.1.0 rewiring** (v5's checklist/stage engine gave way to the prompts-driven reviewer, per juspay/yama's MIGRATION.md):

- `.yama/yama.yaml` + `mcp.yaml` + `rulebook/` → `.yama/config.json` + `MCP.json` + `prompts.json` + `skills/`. Models unchanged: `litellm [open-fast-sa, open-large-sa]` on the gateway the existing `LITELLM_*` secrets point at. The rulebook's content lives on as `skills/repo-conventions/SKILL.md` (distilled from `CLAUDE.md`'s critical rules — what the review enforces vs. what ESLint already owns) plus `skills/github-commenting/SKILL.md` (markers, suggestion blocks, the one-summary contract).
- **Committed long-term memory**: `.yama/memory/hippocampus.sqlite` is seeded empty and committed — every review run recalls what earlier reviews learned (Hippocampus, keyed `neurolink-repo`); WAL/SHM siblings stay ignored, `binary -diff` in `.gitattributes`.
- **`yama-review.yml`**: drives the published package (`npx @juspay/yama@^6.1.0 run` from `.yama/`). Keeps the fork/secrets preflight gate and the required-check name `Yama PR Review`. v6 verdict contract: the verdict arrives as the PR's review state; the check goes red only when the run itself breaks. Review prompt 1 instructs the agent to delegate exploration/blast-radius work to parallel subagents (`delegate_task` / `collect_results`).
- **`yama-learn.yml`**: chained `workflow_run` after **Release and Publish** (plus a `workflow_dispatch pr=` recovery lever). Post-merge, one `learnPrompt` turn distills the merged PR's discussion into the memory database; yama v6.1.0's deterministic write path waits for the memory write to land, checkpoints the WAL, stages exactly that one file, reads the staged set back out of git, and commits with the skip-ci token in the subject — never a force push. The trust-boundary and push-bypass notes from the v5 file are kept: the push uses the workflow `GITHUB_TOKEN` and needs the branch-protection bypass for `github-actions[bot]`; until granted, learn fails loudly at the push and can be re-dispatched.

This wiring is the same loop just proven live on juspay/yama itself (its PR #106 → first learn commit landed on its main, memory recalled on the next review).


murdore:
Reviewed as part of an audit of all open PRs; findings verified against the diff before reporting.

The core change — folding inline `system`-role messages into `systemPrompt` — is sound. Four things to address, plus a rebase.

**1. The regression test passes even if the fix is inverted.** The new case asserts only `inlineSystemPrompt.includes("Inline system reminder")` and that no `conversationMessages` entry has role `"system"`. It never asserts the pre-existing prompt survived. Concretely: change the implementation from `systemPrompt = systemPrompt ? \`${systemPrompt}\n\n${text}\` : text` to a naive `systemPrompt = text`, and `inlineSystemPrompt` becomes exactly `"Inline system reminder"` — both assertions still pass, while the original system prompt is silently dropped. Asserting the original is still present, and ordered before the inline text, closes it.

**2. A caller-controlled value goes into a FAIL message.** The suite logs the raw `parsedWithInlineSystem.systemPrompt` into an assertion message. This repo has a specific hazard here, documented in CLAUDE.md: `defineSuite`'s `test()` downgrades a thrown error to **SKIP** rather than FAIL when the message matches `isExpectedProviderError()`. A message quoting payload text can therefore turn a real failure into a green `⊘` and a zero exit code. Use a structural message naming the key path instead of interpolating the value.

**3. The vendored Yama action's preflight can validate a different config than the review uses.** In `.github/actions/yama-review/action.yml`, the "Verify the setup" step builds `DOCTOR_ARGS=(doctor)` and never passes `--config "$YAMA_INPUT_CONFIG"`, while the "Review" step explicitly does. Latent for this repo — `yama-review.yml` never overrides `config` — but wrong for any other consumer, and it defeats that step's stated purpose.

**4. The shallow-checkout branch contradicts its own comment.** The same file detects a shallow clone and emits only `::warning::`, despite the adjacent comment arguing that a shallow clone means "the diff would be wrong rather than absent — which is worse" and to "say so loudly". Nothing downstream re-checks. A consumer who forgets `fetch-depth: 0` gets a review computed against a wrong diff, presented as a completed one.

**5. `.yama/knowledge/workflow.md` instructs the reviewer to paste secrets into public comments.** The EVIDENCE REQUIREMENT section says that for a CRITICAL hardcoded-secret finding the reviewer "must QUOTE the exact ADDED line(s) from that file's diff hunk". If that ever fires on a real secret, it writes the literal value into a permanent, indexed PR comment — arguably worse exposure than the commit, which can at least be force-pushed away. This wording predates the PR (it was in the deleted `.yama/config.yaml`), but the PR rewrites that tree and carried it forward unchanged. "Quote a redacted fragment sufficient to verify, with file and line" gets the same evidentiary value.

Mechanically it is `CONFLICTING` against `release` and needs a rebase; all five required checks were green on the current head.

murdore:
I've rebased this onto current `release` and pushed it here, still authored to you. **I reduced its scope, so please read this before the diff** — the proxy fix landed, the `.yama/` half did not.

**The proxy fix.** Unchanged and correct. Folding an inline `system`-role message into `systemPrompt` is the right call, and the reasoning in your comment about non-leading system messages breaking OpenAI-shaped chat templates is accurate.

**Two things I changed in the test.**

The regression test passed with the fix inverted. It asserted only that the inline text appeared in `systemPrompt` and that no `system` role remained in `conversationMessages` — both of which hold for an implementation that *overwrites* `systemPrompt` rather than appending to it, silently discarding the request's own system prompt. I verified this by mutation rather than by reading: swap the append for an assignment and the original assertions stay green, while `originalFound` goes false. It now also asserts the pre-existing prompt survives and still precedes the folded text.

It also interpolated the parsed `systemPrompt` into its FAIL message. `defineSuite` downgrades a failure to SKIP when the message matches `isExpectedProviderError()`, so a payload quoted into a failure message can turn a real regression into a green tick and a zero exit. The messages are structural now.

**What I left out, and why it isn't about the bundling.** `release` has gained two commits of hardening to `yama-review.yml` since you branched — `208c8dba`, which stops a review check from passing when no review actually happened, and `60993cc3`, which dismisses superseded verdicts. Your restructure predates both and reproduces neither, in the workflow or in the vendored action; I checked both files for each. Merging it as-is would remove the control that stops a green Yama check from meaning nothing, and that is not something I'm willing to land as a side effect of a proxy fix. It wants its own PR, rebased against that hardening — and it's worth having, so I'd encourage you to open one.

The same applies to this branch's copy of `test/continuous-test-suite.ts`: it reverted the subprocess-failure guard to skipping whenever stdout was empty, which is the exact shape in which an ordinary Node crash — including a real schema regression — gets reported green. I took only the new test block and left the rest of the file as `release` has it.

**Three findings against the vendored action still stand.** They travel with that PR rather than this one, so repeating them here so they aren't lost:

1. The preflight builds `DOCTOR_ARGS=(doctor)` without `--config "$YAMA_INPUT_CONFIG"`, while the review step passes it — so the step can validate a different config than the one used. Latent for this repo, wrong for any other consumer.
2. The shallow-clone branch emits only `::warning::`, despite the adjacent comment arguing that a shallow clone makes the diff *wrong* rather than absent, "which is worse", and to say so loudly. Nothing downstream re-checks.
3. `.yama/knowledge/workflow.md` tells the reviewer that a CRITICAL hardcoded-secret finding "must QUOTE the exact ADDED line(s)". If that ever fires on a real secret it writes the literal value into a permanent, indexed PR comment — worse exposure than the commit, which can at least be force-pushed away. This predates your PR (it was in the deleted `config.yaml`) but the rewrite carries it forward unchanged. "Quote a redacted fragment sufficient to verify, with file and line" gets the same evidentiary value.

.github/workflows/yama-learn.yml
```diff
@@ -0,0 +1,171 @@
...
+
+      # .yama/mcp.yaml declares the graph server; doctor proves every declared
+      # server and exits 2 on one it cannot spawn.
+      - name: Set up uv (code-review-graph server)
+        if: steps.target.outputs.run == 'true'
+        uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7.6.0
+      - name: Warm the code-review-graph package
+        if: steps.target.outputs.run == 'true'
+        run: uvx --from code-review-graph==2.3.7 code-review-graph --help >/dev/null 2>&1 || echo "warm failed — doctor will say what that means"
+
```
Tara-ag: **MINOR · security** — The artifact-restore step runs `unzip -o -q "$RUNNER_TEMP/yama-store.zip" -d .yama/artifacts` with no path-traversal guard.

A crafted zip entry containing `../` could write outside `.yama/artifacts` (zip-slip). The zip originates from the upstream review workflow's own per-PR artifact, so exploitation requires an attacker able to influence that artifact's contents — a coherent but low-likelihood threat; defense-in-depth only.

**Fix:** Reject entries whose normalized path escapes `.yama/artifacts` (e.g. use `python -m zipfile -e` after validating member names, or unzip with `-d` after checking for `..`), and drop `-o` overwrite of sibling paths.

Evidence: .github/workflows/yama-learn.yml


`yama:finding:zip-slip-artifact-restore`

.github/workflows/yama-learn.yml
```diff
@@ -0,0 +1,171 @@
...
+          mkdir -p .yama/artifacts
+          gh api "repos/${REPOSITORY}/actions/artifacts/$id/zip" > "$RUNNER_TEMP/yama-store.zip"
+          unzip -o -q "$RUNNER_TEMP/yama-store.zip" -d .yama/artifacts
+          echo "Restored the run store from artifact $id."
+
+      - name: Set the commit identity
+        if: steps.target.outputs.run == 'true'
+        # The identity the actor guards match — what stops a learn commit feeding
+        # itself back in.
+        run: |
```
Tara-ag, line 125: **MAJOR · security** — The post-merge `yama learn` step (npx @juspay/yama learn --pr) distills the merged PR's discussion into natural language and auto-commits it to `.yama/memory/` on the protected `release` branch via the workflow token (contents: write, push: true), with no human review.

Adversarial content planted in a merged PR's comments/discussion is distilled and committed, permanently and unreviewed, into a protected branch's persistent memory that the workflow header itself says is 'read by every later review during WarmUp'. A single planted prompt can therefore persist across runs and steer/subvert all future Yama reviews (prompt-injection persistence). The write path also auto-mutates the protected mainline with no review gate.

**Fix:** Do not auto-push distilled memory with the repository write credential. Require human/PR review of the learned memory before it lands on release; treat learn output as untrusted data (it can carry instructions — never execute/trust it, and normalize/filter before commit); or scope the memory store such that unvetted output is never fed to the review warm-up.

Evidence: .github/workflows/yama-learn.yml, .github/workflows/yama-learn.yml, .yama/yama.yaml


`yama:finding:learn-memory-injection`

.github/workflows/yama-review.yml
```diff
@@ -83,209 +60,70 @@ jobs:
...
+      # from a superseded head in reviewDecision until dismissed. Clear the bot's
+      # OWN stale verdicts once this run produced a verdict for the current head.
+      # Guards: never on a cancelled run, never when the head has already moved
+      # (the newer run owns dismissal), fully non-fatal.
       - name: Dismiss superseded verdicts
         if: ${{ !cancelled() && steps.gate.outputs.run == 'true' && steps.yama.outputs.decision != '' }}
         env:
           GH_TOKEN: ${{ secrets.YAMA_GITHUB_TOKEN }}
           PR: ${{ github.event.pull_request.number }}
           HEAD_SHA: ${{ github.event.pull_request.head.sha }}
```
Tara-ag: **MAJOR · correctness** — The 'Dismiss superseded verdicts' step is gated on `steps.yama.outputs.decision != ''`, but the pinned v5 action (juspay/yama@cb7f7cd4) is external to this checkout, so whether it actually emits a `decision` output cannot be verified locally.

If the v5 action does not emit (or renames) the `decision` output on a completed run, the dismiss step silently never fires and a stale bot CHANGES_REQUESTED from a superseded head stays green-blocking forever — the exact stale-verdict deadlock this workflow (#1509/#1513/#1542) was designed to eliminate. Because the guard failure is silent, it would go unnoticed until merges stall for no visible reason.

**Fix:** Verify the v5 action's output contract (confirm it emits `decision` on success, or read the verdict from the run's own JSON report, which the workflow header says the action writes). Anchor the dismiss guard on that report so a missing/renamed output cannot silently disable stale-verdict dismissal.

Evidence: .github/workflows/yama-review.yml, external juspay/yama@cb7f7cd4


`yama:finding:yama-decision-output-contract`

test/continuous-test-suite.ts
```diff
@@ -4477,6 +4477,77 @@ async function testClaudeFormatRuntime(): Promise<boolean | null> {
       return false;
     }
 
+    // Regression: an inline "system"-role message (invalid per the Messages
```
Tara-ag, line 4520: **MINOR · testing** — The new inline-system folding regression is wired and executes only for the ClaudeFormat path (testClaudeFormatRuntime); there is no equivalent regression for parseOpenAIRequest's mirror path.

The OpenAI-compatible translation path mirrors the same inline-system folding behaviour but has no covering regression. A future divergence or regression in parseOpenAIRequest's folding (or a mistake in the mirror when the behaviour is copied) would ship without test coverage, despite the change being explicitly about folding consistency across both formats.

**Fix:** Add an equivalent runtime regression for parseOpenAIRequest asserting the inline 'system' message folds into the OpenAI system role and does not leak into the translated messages array.

Evidence: test/continuous-test-suite.ts, src/lib/proxy/claudeFormat.ts


`yama:finding:openai-inline-system-test-gap`

Tara-ag:
## Yama review

Reviewed PR #1514, which migrates NeuroLink's Yama review automation from v3 (.yama/config.yaml deleted) to v5 (yama.yaml, mcp.yaml, rulebook/index.md), rewrites yama-review.yml, adds a new post-merge yama-learn.yml workflow, extends .gitignore, and makes one source change folding inline role:'system' messages into the Claude request's system prompt with a matching regression test. I verified each finding against the actual PR diff (the yama-learn.yml workflow, mcp.yaml capability map, yama.yaml verdict policy, and the claudeFormat.ts + continuous-test-suite.ts hunks were all read directly). Two MAJORs: (1) the learn workflow auto-pushes model-distilled, unreviewed PR-discussion-derived text into a protected branch's persistent memory that later reviews read during WarmUp — a real prompt-injection-persistence surface; (2) the v5 'Dismiss superseded verdicts' gate depends on an action output (`decision`) external to this checkout, so a silently missing output would re-create the stale-verdict deadlock. Two MINORs (zip-slip in unzip restore; missing OpenAI mirror regression) and two INFOs (silent drop of empty/non-text inline system messages; the required-check contract in settings.yml is currently latent). Config migration of blocking/verdict semantics (t30/t31) was confirmed faithful; the learn/release workflow chaining, [skip ci]/bot/merged loop guards, and concurrency design (t32) are sound, and the inline-system folding logic itself (t1, t2) has no correctness defect beyond the noted INFO nuance. No CRITICAL findings.

**Verdict: COMMENT**
- 2 MAJOR: learn-memory-injection, yama-decision-output-contract — commentOn covers MAJOR, which reports without gating

Findings (6), most serious first:
- **MAJOR** `.github/workflows/yama-learn.yml:142` — The post-merge `yama learn` step (npx @juspay/yama learn --pr) distills the merged PR's discussion into natural language and auto-commits it to `.yama/memory/` on the protected `release` branch via the workflow token (contents: write, push: true), with no human review.
- **MAJOR** `.github/workflows/yama-review.yml:117` — The 'Dismiss superseded verdicts' step is gated on `steps.yama.outputs.decision != ''`, but the pinned v5 action (juspay/yama@cb7f7cd4) is external to this checkout, so whether it actually emits a `decision` output cannot be verified locally.
- **MINOR** `test/continuous-test-suite.ts:4480` — The new inline-system folding regression is wired and executes only for the ClaudeFormat path (testClaudeFormatRuntime); there is no equivalent regression for parseOpenAIRequest's mirror path.
- **MINOR** `.github/workflows/yama-learn.yml:118` — The artifact-restore step runs `unzip -o -q "$RUNNER_TEMP/yama-store.zip" -d .yama/artifacts` with no path-traversal guard.
- **INFO** `src/lib/proxy/claudeFormat.ts:119` — In parseClaudeRequest's inline system-role folding, an empty-string system message, or one whose content array has no non-empty text block, yields falsy `text` and the message is silently dropped via `continue`.
- **INFO** `.github/workflows/yama-review.yml:50` — yama-review.yml documents that the 'Yama PR Review' job IS the required status-check context in branch protection (keep the two in sync, .github/settings.yml), but settings.yml does not list it as a required check (workflow state=disabled_manually, backlog D1).

2 further finding(s) were not commented on inline: inline-system-empty-drop, release-required-check-stale.


`yama:run:run-2026-08-30T18-32-35-005Z`

Tara-ag:
## Yama review

Reviewed PR #1514, which migrates NeuroLink's Yama review automation from v3 (.yama/config.yaml deleted) to v5 (yama.yaml, mcp.yaml, rulebook/index.md), rewrites yama-review.yml, adds a new post-merge yama-learn.yml workflow, extends .gitignore, and makes one source change folding inline role:'system' messages into the Claude request's system prompt with a matching regression test. I verified each finding against the actual PR diff (the yama-learn.yml workflow, mcp.yaml capability map, yama.yaml verdict policy, and the claudeFormat.ts + continuous-test-suite.ts hunks were all read directly). Two MAJORs: (1) the learn workflow auto-pushes model-distilled, unreviewed PR-discussion-derived text into a protected branch's persistent memory that later reviews read during WarmUp — a real prompt-injection-persistence surface; (2) the v5 'Dismiss superseded verdicts' gate depends on an action output (`decision`) external to this checkout, so a silently missing output would re-create the stale-verdict deadlock. Two MINORs (zip-slip in unzip restore; missing OpenAI mirror regression) and two INFOs (silent drop of empty/non-text inline system messages; the required-check contract in settings.yml is currently latent). Config migration of blocking/verdict semantics (t30/t31) was confirmed faithful; the learn/release workflow chaining, [skip ci]/bot/merged loop guards, and concurrency design (t32) are sound, and the inline-system folding logic itself (t1, t2) has no correctness defect beyond the noted INFO nuance. No CRITICAL findings.

**Verdict: COMMENT**
- 2 MAJOR: learn-memory-injection, yama-decision-output-contract — commentOn covers MAJOR, which reports without gating

Findings (6), most serious first:
- **MAJOR** `.github/workflows/yama-learn.yml:142` — The post-merge `yama learn` step (npx @juspay/yama learn --pr) distills the merged PR's discussion into natural language and auto-commits it to `.yama/memory/` on the protected `release` branch via the workflow token (contents: write, push: true), with no human review.
- **MAJOR** `.github/workflows/yama-review.yml:117` — The 'Dismiss superseded verdicts' step is gated on `steps.yama.outputs.decision != ''`, but the pinned v5 action (juspay/yama@cb7f7cd4) is external to this checkout, so whether it actually emits a `decision` output cannot be verified locally.
- **MINOR** `test/continuous-test-suite.ts:4480` — The new inline-system folding regression is wired and executes only for the ClaudeFormat path (testClaudeFormatRuntime); there is no equivalent regression for parseOpenAIRequest's mirror path.
- **MINOR** `.github/workflows/yama-learn.yml:118` — The artifact-restore step runs `unzip -o -q "$RUNNER_TEMP/yama-store.zip" -d .yama/artifacts` with no path-traversal guard.
- **INFO** `src/lib/proxy/claudeFormat.ts:119` — In parseClaudeRequest's inline system-role folding, an empty-string system message, or one whose content array has no non-empty text block, yields falsy `text` and the message is silently dropped via `continue`.
- **INFO** `.github/workflows/yama-review.yml:50` — yama-review.yml documents that the 'Yama PR Review' job IS the required status-check context in branch protection (keep the two in sync, .github/settings.yml), but settings.yml does not list it as a required check (workflow state=disabled_manually, backlog D1).

2 further finding(s) were not commented on inline: inline-system-empty-drop, release-required-check-stale.


`yama:run:run-2026-08-30T18-32-35-005Z`

Tara-ag (commented):
## Yama review

Reviewed PR #1514. Two MAJOR, two MINOR, two INFO findings. Verdict: COMMENT.

test/continuous-test-suite.ts
pdogra1299: Fixed. `parseOpenAIRequest` already skips inline system messages into `systemPrompt` — the behaviour existed, the guard was one-sided, which is exactly how a mirrored path drifts. The suite now asserts the same two properties on the OpenAI side (text reaches `systemPrompt`; no `system` role survives in `conversationMessages`) right next to the Claude one.

.github/workflows/yama-learn.yml
pdogra1299: Fixed. The artifact is one this repository's own workflows produced, but its entry paths are still data — extracting straight over the checkout is the wrong shape regardless of provenance. It now unpacks into `$RUNNER_TEMP/yama-store` and copies into `.yama/artifacts/` from there.

.github/workflows/yama-review.yml
pdogra1299: Verified rather than assumed, and the answer is in this run: the `Dismiss superseded verdicts` step **executed** (conclusion: success, not skipped) on run 33327703614, which it can only do when `steps.yama.outputs.decision` is non-empty — so the pinned action does emit the output, empirically, on the exact SHA this workflow pins.

The gate also fails in the safe direction by design: no verdict means no dismissal, which leaves stale signal for a human instead of clearing a verdict that was never replaced. What the finding is really pointing at — an external action's output contract can drift under a moved pin — is true of every pinned action, and the pin is the mitigation: this workflow pins an immutable commit SHA, not a tag.

.github/workflows/yama-learn.yml
pdogra1299: Correct, and the most valuable finding here — it is not fixable by a line in this PR, so it is being answered where the decision actually gets made.

The input to `learn` is pull-request discussion, which anyone who can comment on a PR that later merges can influence, and what it writes is read by every later review as this repository's settled knowledge. The existing guards are real but coarse: only merged, human-authored pull requests are learned from; only `.yama/` is ever staged, verified by reading the staged set back out of git; the commit lands in history where it can be read and reverted; and `[skip ci]` keeps it from re-triggering CI.

Two things changed in response. This workflow now carries a TRUST BOUNDARY note at the top of the file, positioned deliberately: pushing to the protected `release` branch requires a maintainer to grant `github-actions[bot]` a branch-protection bypass, and that grant is the human decision point — until it exists, learn computes, fails loudly at the push, and writes nothing. The note says plainly that a memory commit is a change to the review rulebook and should be reviewed as one.

The durable fix belongs upstream in Yama, not here: memory facts should carry provenance and be weighed as sourced claims rather than as authority, and learn should be able to PROPOSE memory through a pull request instead of committing to the mainline. Both are on the Yama backlog with this finding cited as the reason.

.github/workflows/yama-learn.yml
```diff
@@ -0,0 +1,187 @@
...
+      - name: Learn
+        if: steps.target.outputs.run == 'true'
+        env:
+          YAMA_GITHUB_TOKEN: ${{ secrets.YAMA_GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.YAMA_GITHUB_TOKEN }}
+          LITELLM_API_KEY: ${{ secrets.LITELLM_API_KEY }}
+          LITELLM_BASE_URL: ${{ secrets.LITELLM_BASE_URL }}
+          PR_NUMBER: ${{ steps.target.outputs.pr }}
+        run: |
+          npx --yes "@juspay/yama@^5.0.1" learn --pr "$PR_NUMBER"
```
Tara-ag: **MAJOR · Security / trust boundary** — Post-merge `yama learn` auto-commits LLM-distilled memory onto the protected `release` branch with the workflow's `contents: write` GITHUB_TOKEN and `learn.push: true` in yama.yaml.

The `npx --yes "@juspay/yama@^5.0.1" learn --pr "$PR_NUMBER"` step distills the merged PR's discussion (written by anyone who can comment on a pull request that later merges) into natural language and auto-commits it to `.yama/memory/` on the protected `release` branch. That memory is read by every later review as the repository's own settled knowledge — effectively a change to the review rulebook — with no human review gate before the push lands. Guards (merged-only, human author, [skip ci], only .yama/ staged) are real but coarse: they restrict who/when, not the trustworthiness of the injected content. This widens the review pipeline's trust boundary and is the single most consequential change in this PR.

**Fix:** Require human approval before the learn commit lands (e.g. open a PR for the memory change instead of pushing directly to `release`, or gate the push on a review). At minimum, keep the push bypass entry out of branch protection until the memory writes are reviewed like rulebook changes, as the workflow header itself recommends.

Evidence: .github/workflows/yama-learn.yml:187, .yama/yama.yaml, .github/workflows/yama-learn.yml:62-65, .github/workflows/yama-learn.yml:10-17


`yama:finding:yama-learn-memory-injection`

.github/workflows/yama-learn.yml
```diff
@@ -0,0 +1,187 @@
...
+          if [ -z "$id" ]; then echo "No review store — learning from the discussion alone."; exit 0; fi
+          gh api "repos/${REPOSITORY}/actions/artifacts/$id/zip" > "$RUNNER_TEMP/yama-store.zip"
+          # Unpack somewhere disposable and copy only what belongs in the store: a
+          # zip is data, and its entry paths are part of that data. Extracting it
+          # straight over the checkout would let a crafted entry write outside
+          # .yama/artifacts. `unzip -j`-style flattening would lose the store's
+          # layout, so the guard is the staging directory plus an explicit copy.
+          rm -rf "$RUNNER_TEMP/yama-store"
+          mkdir -p "$RUNNER_TEMP/yama-store" .yama/artifacts
+          unzip -o -q "$RUNNER_TEMP/yama-store.zip" -d "$RUNNER_TEMP/yama-store"
```
Tara-ag: **MINOR · Security / path traversal (defense in depth)** — The Restore step's comment claims unzipping into `$RUNNER_TEMP/yama-store` then `cp -r` into `.yama/artifacts` prevents a crafted zip entry from writing outside `.yama/artifacts`, but `unzip -d` does not sanitize `../` traversal.

A malicious artifact entry named e.g. `../../tmp/pwn` extracted into the staging dir still resolves to `$RUNNER_TEMP/tmp/pwn` — inside $RUNNER_TEMP but outside the `yama-store` staging directory the comment claims confines it. The staging dir does genuinely keep writes out of the checkout and the repo (the important protection), so this is a mitigated-but-not-eliminated hazard, not a live breach. Only the explanatory comment's claim that it prevents writes outside `.yama/artifacts` is overstated.

**Fix:** Prefer `python -m zipfile -e ... $RUNNER_TEMP/yama-store` after rejecting entries containing `..` or an absolute path, or validate each entry name (reject `..` segments and leading `/`) before extraction. If keeping `unzip`, reword the comment to say the staging dir confines traversal to $RUNNER_TEMP rather than preventing it.

Evidence: .github/workflows/yama-learn.yml:150-151, .github/workflows/yama-learn.yml:143-147


`yama:finding:yama-zip-slip-mitigation-overstated`

Tara-ag:
## Yama review

Reviewed the Yama v5 migration of NeuroLink's review infrastructure end to end: the new post-merge yama-learn.yml (workflow_run chain off Release, concurrency yama-learn with cancel-in-progress:false, merged+h-uman-actor guards, PR resolution, staged artifact restore, `npx @juspay/yama@^5.0.1 learn --pr` auto-commit), the rewritten yama-review.yml (external composite action pin to juspay/yama@cb7f7cd4 v5.0.1, fail-on-blocked, Dismiss-superseded-verdicts gate), the v5 .yama/ layout (yama.yaml learn.push:true, mcp.yaml read+review-write capability map, rulebook/index.md), the inline system-role fold in src/lib/proxy/claudeFormat.ts, and the matching regressions (including the previously-missing parseOpenAIRequest mirror) in test/continuous-test-suite.ts. Line numbers were re-derived from a precise count of the actual new files (learn command at yama-learn.yml:187, unzip at :150, unconditional `continue` at claudeFormat.ts:120), correcting the earlier 142/148/140/119 citations. Net: four findings survive dedup — one MAJOR (auto-committed LLM memory onto the protected release branch, the central trust-boundary widening, well-documented but capability unchanged), one MINOR (staging-dir zip-slip comment overstated; actual blast radius confined to $RUNNER_TEMP), and two INFO (unverifiable external `decision` output makes the stale-verdict dismissal gate fail-closed; benign empty-inline-system drop). The previously-reported race, the old zip-slip-over-checkout, the missing OpenAI mirror test, and the settings.yml required-check drift were all confirmed fixed in this head. The prior learn-memory-injection (MAJOR) and inline-system-empty-drop (INFO) findings merge into my current-run versions; yama-decision-output-contract was re-found by me and its severity is assessed as INFO (fail-closed, no security impact).

**Verdict: COMMENT**
- 2 MAJOR: learn-memory-injection, yama-learn-memory-injection — commentOn covers MAJOR, which reports without gating

Findings (6), most serious first:
- **MAJOR** `.github/workflows/yama-learn.yml:142` — The post-merge `yama learn` step (npx @juspay/yama learn --pr) distills the merged PR's discussion into natural language and auto-commits it to `.yama/memory/` on the protected `release` branch via the workflow token (contents: write, push: true), with no human review.
- **MAJOR** `.github/workflows/yama-learn.yml:187` — Post-merge `yama learn` auto-commits LLM-distilled memory onto the protected `release` branch with the workflow's `contents: write` GITHUB_TOKEN and `learn.push: true` in yama.yaml.
- **MINOR** `.github/workflows/yama-learn.yml:150` — The Restore step's comment claims unzipping into `$RUNNER_TEMP/yama-store` then `cp -r` into `.yama/artifacts` prevents a crafted zip entry from writing outside `.yama/artifacts`, but `unzip -d` does not sanitize `../` traversal.
- **INFO** `src/lib/proxy/claudeFormat.ts:120` — In parseClaudeRequest's inline 'system'-role fold, the `continue` is unconditional (outside the `if (text)` guard), so an inline system message whose text folds to empty/falsy is silently dropped from the conversation entirely.
- **INFO** `src/lib/proxy/claudeFormat.ts:119` — In parseClaudeRequest's inline system-role folding, an empty-string system message, or one whose content array has no non-empty text block, yields falsy `text` and the message is silently dropped via `continue`.
- **INFO** `.github/workflows/yama-review.yml:117` — The 'Dismiss superseded verdicts' step is gated on `steps.yama.outputs.decision != ''`, but `decision` comes from the external composite action pinned to `juspay/yama@cb7f7cd4` (v5.0.1), whose output contract cannot be verified from this repository.

3 further finding(s) were not commented on inline: claude-inline-system-empty-drop, inline-system-empty-drop, yama-decision-output-contract.

1 finding(s) were already commented on by an earlier run and were not posted again.


`yama:run:run-2026-08-30T19-06-32-804Z`

.github/workflows/yama-learn.yml
Tara-ag: **MINOR · Security / path traversal (defense in depth)** — The Restore step's comment claims unzipping into `$RUNNER_TEMP/yama-store` then `cp -r` into `.yama/artifacts` prevents a crafted zip entry from writing outside `.yama/artifacts`, but `unzip -d` does not sanitize `../` traversal.

A malicious artifact entry named e.g. `../../tmp/pwn` extracted into the staging dir still resolves to `$RUNNER_TEMP/tmp/pwn` — inside $RUNNER_TEMP but outside the `yama-store` staging directory the comment claims confines it. The staging dir does genuinely keep writes out of the checkout and the repo (the important protection), so this is a mitigated-but-not-eliminated hazard, not a live breach. Only the explanatory comment's claim that it prevents writes outside `.yama/artifacts` is overstated.

**Fix:** Prefer `python -m zipfile -e ... $RUNNER_TEMP/yama-store` after rejecting entries containing `..` or an absolute path, or validate each entry name (reject `..` segments and leading `/`) before extraction. If keeping `unzip`, reword the comment to say the staging dir confines traversal to $RUNNER_TEMP rather than preventing it.

Evidence: .github/workflows/yama-learn.yml:150-151, .github/workflows/yama-learn.yml:143-147


`yama:finding:yama-zip-slip-mitigation-overstated`

.github/workflows/yama-learn.yml
Tara-ag: **MAJOR · Security / trust boundary** — Post-merge `yama learn` auto-commits LLM-distilled memory onto the protected `release` branch with the workflow's `contents: write` GITHUB_TOKEN and `learn.push: true` in yama.yaml.

The `npx --yes "@juspay/yama@^5.0.1" learn --pr "$PR_NUMBER"` step distills the merged PR's discussion (written by anyone who can comment on a pull request that later merges) into natural language and auto-commits it to `.yama/memory/` on the protected `release` branch. That memory is read by every later review as the repository's own settled knowledge — effectively a change to the review rulebook — with no human review gate before the push lands. Guards (merged-only, human author, [skip ci], only .yama/ staged) are real but coarse: they restrict who/when, not the trustworthiness of the injected content. This widens the review pipeline's trust boundary and is the single most consequential change in this PR.

**Fix:** Require human approval before the learn commit lands (e.g. open a PR for the memory change instead of pushing directly to `release`, or gate the push on a review). At minimum, keep the push bypass entry out of branch protection until the memory writes are reviewed like rulebook changes, as the workflow header itself recommends.

Evidence: .github/workflows/yama-learn.yml:187, .yama/yama.yaml, .github/workflows/yama-learn.yml:62-65, .github/workflows/yama-learn.yml:10-17


`yama:finding:yama-learn-memory-injection`

Tara-ag:
## Yama review

Reviewed the Yama v5 migration of NeuroLink's review infrastructure end to end: the new post-merge yama-learn.yml (workflow_run chain off Release, concurrency yama-learn with cancel-in-progress:false, merged+h-uman-actor guards, PR resolution, staged artifact restore, `npx @juspay/yama@^5.0.1 learn --pr` auto-commit), the rewritten yama-review.yml (external composite action pin to juspay/yama@cb7f7cd4 v5.0.1, fail-on-blocked, Dismiss-superseded-verdicts gate), the v5 .yama/ layout (yama.yaml learn.push:true, mcp.yaml read+review-write capability map, rulebook/index.md), the inline system-role fold in src/lib/proxy/claudeFormat.ts, and the matching regressions (including the previously-missing parseOpenAIRequest mirror) in test/continuous-test-suite.ts. Line numbers were re-derived from a precise count of the actual new files (learn command at yama-learn.yml:187, unzip at :150, unconditional `continue` at claudeFormat.ts:120), correcting the earlier 142/148/140/119 citations. Net: four findings survive dedup — one MAJOR (auto-committed LLM memory onto the protected release branch, the central trust-boundary widening, well-documented but capability unchanged), one MINOR (staging-dir zip-slip comment overstated; actual blast radius confined to $RUNNER_TEMP), and two INFO (unverifiable external `decision` output makes the stale-verdict dismissal gate fail-closed; benign empty-inline-system drop). The previously-reported race, the old zip-slip-over-checkout, the missing OpenAI mirror test, and the settings.yml required-check drift were all confirmed fixed in this head. The prior learn-memory-injection (MAJOR) and inline-system-empty-drop (INFO) findings merge into my current-run versions; yama-decision-output-contract was re-found by me and its severity is assessed as INFO (fail-closed, no security impact).

**Verdict: COMMENT**
- 2 MAJOR: learn-memory-injection, yama-learn-memory-injection — commentOn covers MAJOR, which reports without gating

Findings (6), most serious first:
- **MAJOR** `.github/workflows/yama-learn.yml:142` — The post-merge `yama learn` step (npx @juspay/yama learn --pr) distills the merged PR's discussion into natural language and auto-commits it to `.yama/memory/` on the protected `release` branch via the workflow token (contents: write, push: true), with no human review.
- **MAJOR** `.github/workflows/yama-learn.yml:187` — Post-merge `yama learn` auto-commits LLM-distilled memory onto the protected `release` branch with the workflow's `contents: write` GITHUB_TOKEN and `learn.push: true` in yama.yaml.
- **MINOR** `.github/workflows/yama-learn.yml:150` — The Restore step's comment claims unzipping into `$RUNNER_TEMP/yama-store` then `cp -r` into `.yama/artifacts` prevents a crafted zip entry from writing outside `.yama/artifacts`, but `unzip -d` does not sanitize `../` traversal.
- **INFO** `src/lib/proxy/claudeFormat.ts:120` — In parseClaudeRequest's inline 'system'-role fold, the `continue` is unconditional (outside the `if (text)` guard), so an inline system message whose text folds to empty/falsy is silently dropped from the conversation entirely.
- **INFO** `src/lib/proxy/claudeFormat.ts:119` — In parseClaudeRequest's inline system-role folding, an empty-string system message, or one whose content array has no non-empty text block, yields falsy `text` and the message is silently dropped via `continue`.
- **INFO** `.github/workflows/yama-review.yml:117` — The 'Dismiss superseded verdicts' step is gated on `steps.yama.outputs.decision != ''`, but `decision` comes from the external composite action pinned to `juspay/yama@cb7f7cd4` (v5.0.1), whose output contract cannot be verified from this repository.

3 further finding(s) were not commented on inline: claude-inline-system-empty-drop, inline-system-empty-drop, yama-decision-output-contract.

1 finding(s) were already commented on by an earlier run and were not posted again.


`yama:run:run-2026-08-30T19-06-32-804Z`

.yama/knowledge/workflow.md
```diff
@@ -0,0 +1,105 @@
...
+  anchored in verifiable evidence: the file must appear in the PR's
+  changed-files list, and you must QUOTE the exact ADDED line(s) from that
+  file's diff hunk in your comment.
+- Before reporting a secret, re-read the specific file and confirm the
+  quoted value is present. If you cannot re-locate the exact line and
+  quote it, the finding is unverified — DISCARD it. Reporting a fabricated
+  credential leak is worse than missing a real one: it destroys trust in
+  every future finding.
+- Placeholder/example values (docs, tests, "sk-your-key-here",
+  ${ENV_VAR} references, redacted samples) are NOT leaked secrets.
```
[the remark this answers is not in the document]
pdogra1299: Moot — the file this comments on no longer exists.

`.yama/knowledge/workflow.md` was the v3 layout's prompt file. This pull request replaces that layout entirely: `.yama/config.yaml` and `.yama/knowledge/` are deleted, and what a review looks for now lives in `.yama/rulebook/index.md`, with models and policy in `.yama/yama.yaml`. Verified on both this branch and `release` — the path is absent from each.

The substance was right, and the replacement does not carry the defect forward: the rulebook contains no instruction to quote a credential into a comment. Grounding a CRITICAL finding is enforced in Yama itself now (findings must cite a diff-touched path), not asked for in prose, so there is no longer any prompt telling a reviewer to reproduce a secret to prove one exists.

.github/actions/yama-review/action.yml
```diff
@@ -0,0 +1,342 @@
...
+        set -euo pipefail
+        if [ ! -d .git ]; then
+          echo "::error::Yama needs the repository checked out. Add actions/checkout@v4 with fetch-depth: 0 before this step."
+          exit 1
+        fi
+        # A shallow clone cannot reach the base commit, so the diff would be
+        # wrong rather than absent — which is worse. Say so loudly.
+        if [ "$(git rev-parse --is-shallow-repository)" = "true" ]; then
+          echo "::warning::This is a shallow clone. Yama needs history to compute the diff against the base. Use fetch-depth: 0."
+        fi
```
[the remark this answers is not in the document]
pdogra1299: Moot as written, and the underlying concern is structurally prevented now.

`.github/actions/yama-review/action.yml` was this repository's own local wrapper action; it is deleted in this pull request, which calls the published `juspay/yama` action directly. Absent from both this branch and `release`.

On the substance — a review must not run against a base it cannot diff — two things now cover it:

1. `.github/workflows/yama-review.yml:77` checks out with `fetch-depth: 0`, so there is no shallow clone to detect.
2. The published action runs `doctor` before the review with the same `--pr` and `--base`, and its base-ref row fails the job when that ref does not resolve in the clone. The check moved from a warning in a wrapper to a hard preflight in the action itself.

So the failure mode you describe — reviewing an incomplete diff because the merge base was unavailable — cannot reach the review stage silently.

.github/actions/yama-review/action.yml
```diff
@@ -0,0 +1,342 @@
...
+        # makes argv[1] the real file and the CLI actually runs.
+        YAMA_DIR="$RUNNER_TEMP/yama-cli"
+        npm install --prefix "$YAMA_DIR" --no-save --no-audit --no-fund \
+          "@juspay/yama@$YAMA_INPUT_VERSION"
+        YAMA_CLI="$YAMA_DIR/node_modules/@juspay/yama/dist/v4/cli/cli.js"
+        # Everything expensive that can fail late is checked here instead:
+        # connections, credentials, and every capability the run will need.
+        DOCTOR_ARGS=(doctor)
+        if [ "$YAMA_INPUT_DRY_RUN" != "true" ]; then DOCTOR_ARGS+=(--live); fi
+        node "$YAMA_CLI" "${DOCTOR_ARGS[@]}"
```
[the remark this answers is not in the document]
pdogra1299: Moot — same deleted file.

`.github/actions/yama-review/action.yml` and the `config` input you refer to are both gone in this pull request. The v5 action takes no `config` input: configuration is discovered from `.yama/` in the checked-out repository, so `doctor` and `review` read the same three files (`yama.yaml`, `mcp.yaml`, `rulebook/`) by construction rather than by being passed a matching path.

The property you were asking for — preflight validates what the review will actually use — is what the action's `doctor` step now does by default, and it probes every capability the run needs against the tools the MCP servers really advertise.

.github/workflows/yama-review.yml
pdogra1299: Verified against the action itself — the output exists, and the guard is already anchored where you suggested.

`juspay/yama`'s `action.yml` declares:

    outputs:
      decision:
        description: "approve | comment | block"
        value: ${{ steps.report.outputs.decision }}

and `steps.report` runs `if: always()`, deriving `decision` by reading the run's own JSON report from disk — the exact "read the verdict from the run's own JSON report" you proposed. So `decision` is empty only when the review produced no verdict at all, which is precisely when there is nothing to dismiss: dismissing on a run that never reached a verdict would clear a valid blocking review and replace it with nothing.

Your point about it being unverifiable from this checkout is fair, and that is a real maintainability cost of pinning an external action. The pin now carries the contract in a comment beside it, so a future bump has something to check against rather than having to go read the action:

    # `decision` below is a declared output of this action (approve | comment |
    # block). The action derives it from the run's own JSON report in a step
    # that runs on `always()`, so it is empty only when the review produced no
    # verdict at all — which is exactly when there is nothing to dismiss.

Bumped in the same change from v5.0.1 to v5.1.0 (SHA `d8ab7629`), which fixes two platform assumptions that would have bitten this wiring: delivery proved a set verdict by matching GitHub's `REQUEST_CHANGES`, and it looked for its comment markers in a field named `body`. Both now come from the mapped tool's own schema.

.github/workflows/yama-learn.yml
pdogra1299: Correct, and fixed — the comment was overstating what the staging directory does, so the guard is now real rather than the wording softened.

You are right that `unzip -d` does not sanitise `..`. The staging directory keeps a crafted entry out of the checkout, which is the protection that matters, but it does not confine the write to the staging directory itself, and the comment claimed otherwise.

Entry names are now checked before anything is written:

    if unzip -Z1 "$RUNNER_TEMP/yama-store.zip" \
         | grep -qE '^/|^\.\./|/\.\./|/\.\.$'; then
      echo "::error::the run-store artifact carries a traversing entry — refusing to extract it"
      exit 1
    fi

An artifact carrying a traversing or absolute entry is not the run store it claims to be, so refusing it is the right answer rather than extracting it somewhere tidier. Verified against two real zips built for the purpose: a normal store extracts, one containing `../../tmp/pwn` is refused.

The comment now says what the staging directory actually does and why the name check exists.

pdogra1299: Correct, and fixed — the comment was overstating what the staging directory does, so the guard is now real rather than the wording softened.

You are right that `unzip -d` does not sanitise `..`. The staging directory keeps a crafted entry out of the checkout, which is the protection that matters, but it does not confine the write to the staging directory itself, and the comment claimed otherwise.

Entry names are now checked before anything is written:

    if unzip -Z1 "$RUNNER_TEMP/yama-store.zip" \
         | grep -qE '^/|^\.\./|/\.\./|/\.\.$'; then
      echo "::error::the run-store artifact carries a traversing entry — refusing to extract it"
      exit 1
    fi

An artifact carrying a traversing or absolute entry is not the run store it claims to be, so refusing it is the right answer rather than extracting it somewhere tidier. Verified against two real zips built for the purpose: a normal store extracts, one containing `../../tmp/pwn` is refused.

The comment now says what the staging directory actually does and why the name check exists.

.github/workflows/yama-learn.yml
pdogra1299: Real, precisely stated, and deliberately not changed here — with the bound spelled out, because "no human review gate" is true but is not the whole picture.

What the memory can and cannot do. It is read during WarmUp as text, never executed. Yama's groundedness gate drops any finding citing a path outside the diff, so planted text cannot manufacture findings about files a change never touched. Delivery can only call tools the capability map supplies, and `create_pull_request` / `merge_pull_request` / `decline` / `delete_branch` have no mapping, so no prompt reachable through memory can be steered into a write this repository did not grant. What a planted instruction can realistically do is bias emphasis — what gets flagged, how severely — which is worth taking seriously and is not the same as subverting the pipeline.

Why it lands as a commit rather than a PR. `.yama/memory/` is versioned deliberately: every fact is a separate Markdown file with a stated source, the commit is `chore(yama): learn from #N`, and it is readable and revertible like any other. Only `.yama/` is ever staged, the identity is `github-actions[bot]`, and the stage refuses to run on its own commits. The pipeline is also not novel to this repository — the same learn-on-merge runs in juspay/yama and in curator, so gating it here alone would leave three repositories with three different trust models for one mechanism.

Where I have landed. Requiring human approval before learned memory reaches `release` is a policy decision for this repository's owners, not a defect in this pull request, and switching it to a PR-based flow is a larger change than a version bump should carry. Flagging it here for that decision rather than making it silently. If the answer is that memory should be reviewed, the cleanest form is `learn.push: false` in `.yama/yama.yaml` plus a step that opens a pull request with the diff — a small change, but one that should be made on purpose.

.github/workflows/yama-learn.yml
pdogra1299: Real, precisely stated, and deliberately not changed here — with the bound spelled out, because "no human review gate" is true but is not the whole picture.

What the memory can and cannot do. It is read during WarmUp as text, never executed. Yama's groundedness gate drops any finding citing a path outside the diff, so planted text cannot manufacture findings about files a change never touched. Delivery can only call tools the capability map supplies, and `create_pull_request` / `merge_pull_request` / `decline` / `delete_branch` have no mapping, so no prompt reachable through memory can be steered into a write this repository did not grant. What a planted instruction can realistically do is bias emphasis — what gets flagged, how severely — which is worth taking seriously and is not the same as subverting the pipeline.

Why it lands as a commit rather than a PR. `.yama/memory/` is versioned deliberately: every fact is a separate Markdown file with a stated source, the commit is `chore(yama): learn from #N`, and it is readable and revertible like any other. Only `.yama/` is ever staged, the identity is `github-actions[bot]`, and the stage refuses to run on its own commits. The pipeline is also not novel to this repository — the same learn-on-merge runs in juspay/yama and in curator, so gating it here alone would leave three repositories with three different trust models for one mechanism.

Where I have landed. Requiring human approval before learned memory reaches `release` is a policy decision for this repository's owners, not a defect in this pull request, and switching it to a PR-based flow is a larger change than a version bump should carry. Flagging it here for that decision rather than making it silently. If the answer is that memory should be reviewed, the cleanest form is `learn.push: false` in `.yama/yama.yaml` plus a step that opens a pull request with the diff — a small change, but one that should be made on purpose.

pdogra1299: Real, precisely stated, and deliberately not changed here — with the bound spelled out, because "no human review gate" is true but is not the whole picture.

What the memory can and cannot do. It is read during WarmUp as text, never executed. Yama's groundedness gate drops any finding citing a path outside the diff, so planted text cannot manufacture findings about files a change never touched. Delivery can only call tools the capability map supplies, and `create_pull_request` / `merge_pull_request` / `decline` / `delete_branch` have no mapping, so no prompt reachable through memory can be steered into a write this repository did not grant. What a planted instruction can realistically do is bias emphasis — what gets flagged, how severely — which is worth taking seriously and is not the same as subverting the pipeline.

Why it lands as a commit rather than a PR. `.yama/memory/` is versioned deliberately: every fact is a separate Markdown file with a stated source, the commit is `chore(yama): learn from #N`, and it is readable and revertible like any other. Only `.yama/` is ever staged, the identity is `github-actions[bot]`, and the stage refuses to run on its own commits. The pipeline is also not novel to this repository — the same learn-on-merge runs in juspay/yama and in curator, so gating it here alone would leave three repositories with three different trust models for one mechanism.

Where I have landed. Requiring human approval before learned memory reaches `release` is a policy decision for this repository's owners, not a defect in this pull request, and switching it to a PR-based flow is a larger change than a version bump should carry. Flagging it here for that decision rather than making it silently. If the answer is that memory should be reviewed, the cleanest form is `learn.push: false` in `.yama/yama.yaml` plus a step that opens a pull request with the diff — a small change, but one that should be made on purpose.

.github/workflows/yama-learn.yml
```diff
@@ -0,0 +1,199 @@
...
+
+# Two learn runs would race each other pushing to release; the loser would be
+# rebased onto a knowledge base it never read. NOTE: only one PENDING run per
+# group — dispatch recoveries one at a time.
+concurrency:
+  group: yama-learn
+  cancel-in-progress: false
+
+permissions:
+  contents: write
```
Tara-ag, line 59: **MINOR · backward_compat** — The learn workflow pushes its memory commit to the protected `release` branch with the workflow's own GITHUB_TOKEN, but branch protection has no bypass entry for github-actions[bot] yet.

The memory commit is computed and made locally, then the push fails loudly under branch protection, so the learn workflow will systematically fail on the protected `release` branch until the bypass entry for github-actions[bot] is added. Operational failure, not a leak.

**Fix:** Add a branch-protection bypass for the github-actions[bot] actor for the release branch, or route the learned-memory commit through a pull request that a maintainer merges.

Evidence: .github/workflows/yama-learn.yml:63


`yama:finding:learn-push-protection-gap`

Tara-ag:
## Yama review

Reviewed the v3→v5 config migration and the new post-merge 'learn' workflow: yama-learn.yml (workflow_run chain after release, run-store restore with a real entry-name zip-slip guard, npx @juspay/yama learn --pr), the yama-review.yml decision-output gating change, .gitignore v5 additions, the deleted v3 config.yaml vs new yama.yaml/mcp.yaml/rulebook, the claudeFormat/openaiFormat inline-system fold, and the new continuous-test-suite tests. After dedupe, the list is one MAJOR (post-merge LLM-distilled memory auto-committed to the protected release branch with contents:write — the injection mechanism is unchanged by this change), one MINOR (push to the protected release branch lacks a github-actions[bot] bypass and will fail loudly), and three INFO items (conditional-continue inline-system empty fold, external-action decision-output gating contract, and a stale v3 .gitignore comment). The zip-slip mitigation, capability-map coverage in mcp.yaml, and review-flow race guards were verified sound; the new tests cover the string-content inline-system path on both claude and openai sides but not the array-content or empty-falsy folds, a noted minor gap.

**Verdict: COMMENT**
- 2 MAJOR: learn-memory-injection, yama-learn-memory-injection — commentOn covers MAJOR, which reports without gating

Findings (7), most serious first:
- **MAJOR** `.github/workflows/yama-learn.yml:199` — The post-merge 'learn' workflow auto-commits LLM-distilled memory (distilled from pull-request discussion, influenceable by any commenter on a PR that merges) to the protected `release` branch using the job's `contents: write` permission, with no gate that survives an arbitrary commenter's inputs.
- **MAJOR** `.github/workflows/yama-learn.yml:187` — Post-merge `yama learn` auto-commits LLM-distilled memory onto the protected `release` branch with the workflow's `contents: write` GITHUB_TOKEN and `learn.push: true` in yama.yaml.
- **MINOR** `.github/workflows/yama-learn.yml:63` — The learn workflow pushes its memory commit to the protected `release` branch with the workflow's own GITHUB_TOKEN, but branch protection has no bypass entry for github-actions[bot] yet.
- **INFO** `src/lib/proxy/claudeFormat.ts:121` — In parseClaudeRequest's inline 'system'-role fold, the `continue` is unconditional (outside the `if (text)` guard), so an inline system message whose text folds to empty/falsy is silently dropped from the conversation entirely.
- **INFO** `src/lib/proxy/claudeFormat.ts:119` — In parseClaudeRequest's inline system-role folding, an empty-string system message, or one whose content array has no non-empty text block, yields falsy `text` and the message is silently dropped via `continue`.
- **INFO** `.gitignore:101` — `.gitignore` still carries a v3-era header comment 'Yama v3 runtime output (config .yama/config.yaml stays tracked)' that references the deleted `.yama/config.yaml` and v3 runtime dirs (.yama/state, /analytics, /reports).
- **INFO** `.github/workflows/yama-review.yml:128` — The 'Dismiss superseded verdicts' step gates on `steps.yama.outputs.decision != ''` where `decision` is produced solely by the external pinned composite action (juspay/yama@d8ab76298feda5e137744ed945462df1b024746a # v5.1.0), and is not computable or verifiable from this repository.

4 further finding(s) were not commented on inline: claude-inline-system-empty-drop, inline-system-empty-drop, gitignore-stale-config-ref, yama-decision-output-contract.

2 finding(s) were already commented on by an earlier run and were not posted again.


`yama:run:run-2026-08-31T13-59-54-279Z`

.github/workflows/yama-learn.yml
Tara-ag, line 59: **MINOR · backward_compat** — The learn workflow pushes its memory commit to the protected `release` branch with the workflow's own GITHUB_TOKEN, but branch protection has no bypass entry for github-actions[bot] yet.

The memory commit is computed and made locally, then the push fails loudly under branch protection, so the learn workflow will systematically fail on the protected `release` branch until the bypass entry for github-actions[bot] is added. Operational failure, not a leak.

**Fix:** Add a branch-protection bypass for the github-actions[bot] actor for the release branch, or route the learned-memory commit through a pull request that a maintainer merges.

Evidence: .github/workflows/yama-learn.yml:63


`yama:finding:learn-push-protection-gap`

Tara-ag:
## Review summary — inline system message folding (Claude proxy)

I re-verified the core of this change against the current head of `fix/claude-proxy-inline-system-message` and cross-checked it against every earlier finding on the thread. Here's where things stand.

### The change itself is correct ✅

In `parseClaudeRequest` (`src/lib/proxy/claudeFormat.ts`), the new inline-`system`-message fold is sound:

- It matches on `(msg.role as string) === "system"` and `continue`s **before** pushing into `conversationMessages`, so a non-leading system message can no longer leak into the conversation history at a non-zero index — which is exactly what would break OpenAI-shaped chat templates on fallback providers.
- The string/array content handling mirrors `parseOpenAIRequest`'s `systemParts` handling; array form joins only `type: "text"` blocks with `\n`, matching the top-level `body.system` behavior.
- `lastUserMsgIdx`'s backward scan only matches `role === "user"`, so folding a system message cannot interfere with "latest user message" detection.
- The fold appends the inline content *after* the top-level system prompt (`${systemPrompt}\n\n${text}`), and `systemPrompt` being `string | undefined` makes that template safe regardless of whether a top-level system was present.
- The explanatory comment clearly states *why* the fold exists (fallback providers reject non-leading system messages), which is the right kind of rationale to keep.

`parseOpenAIRequest`'s long-standing mirror behavior (fold into `systemPrompt`, skip from history) is consistent with the new Claude path, so the two formats now agree.

### Every earlier finding on this feature is resolved ✅

Confirmed directly against the current test file (`test/continuous-test-suite.ts`), commit ef37ddf:

- **Failure message interpolation** (bnEUA) — the `logTest` failure messages now use structural descriptions ("inline system prompt was not folded into systemPrompt") instead of interpolating the caller-controlled `systemPrompt` value. Good — resolved.
- **Partial assertion** (bnWFA) — the test now asserts all three properties: the merged prompt preserves the original system prompt, the inline reminder is appended, and the order is correct, rather than a lone `includes()` on the inline text. Resolved.
- **OpenAI mirror test gap** (djO2p) — an equivalent `parseOpenAIRequest` regression now asserts text reaches `systemPrompt` and no `system` role survives in `conversationMessages`, sitting right next to the Claude one. Resolved.

### Remaining open threads

The two duplicate `learn-push-protection-gap` threads (`dwA6w` / `dwBYP`) remain open. These are the *same* finding posted twice, and the author has specifically and repeatedly explained (in the related `yama-learn-memory-injection` threads) that the "push fails loudly until a bypass exists" behavior is **deliberate and fail-closed**: the branch-protection bypass is the human decision point, and until a maintainer grants it, learn computes, fails loudly, and writes nothing. That is a policy decision for repo owners, not a defect in this PR, and it is explicitly flagged for that decision. I concur with that framing and did not re-add it.

No new findings to add — the inline-system-message change is correct, well-scoped, and properly covered. 👍

.github/workflows/yama-learn.yml
Tara-ag: Duplicate of the same finding — resolving here too with the same conclusion.

The push failing under branch protection is the documented, deliberate safety gate (committed `PUSH PERMISSION` note): "learn computes, commits locally, and fails LOUDLY at the push — nothing is lost, and the run can be re-dispatched after the bypass is added." The maintainer granting the `github-actions[bot]` bypass is the intended human review decision point for the memory write, per the sibling TRUST BOUNDARY discussion on this PR. Nothing is written until a human opts in, so this is safe-by-default rather than an operational gap.

Withdrawing the finding as moot.

Tara-ag: Resolving — the fail-loud push is the intended design, not a gap.

The committed `PUSH PERMISSION` note in this workflow spells out why: "Until that [bypass] exists, learn computes, commits locally, and fails LOUDLY at the push — nothing is lost, and the run can be re-dispatched after the bypass is added." The maintainer granting the `github-actions[bot]` branch-protection bypass is deliberately positioned as the human decision point for the memory write (see the sibling TRUST BOUNDARY discussion on this PR). Failing the push rather than side-stepping protection is the safe-by-default behavior — nothing is written until a human has consciously opted in. The commit is computed locally and re-dispatched cleanly once the bypass exists, so nothing is lost.

I withdraw the finding: this is an intentional, documented safety gate rather than an operational defect.

Tara-ag:
## Review summary — PR #1514

**Verdict: Approve** (all findings resolved; no remaining blockers).

### The proxy fix (the branch's actual purpose)
`fix(proxy): fold inline system-role messages into systemPrompt` is sound. The change makes ClaudeFormat and the OpenAI-compatible path fold inline `system`-role messages into `systemPrompt` instead of leaking them into the conversation-message array, which fixes a real correctness gap (inline system messages were previously dropped/leaked depending on the caller).

Coverage is now in place on both sides: the ClaudeFormat runtime regression asserts the folded prompt reaches `systemPrompt`, and — in response to the earlier test-gap finding — an equivalent **OpenAI-side** regression now asserts the same two properties (`text` reaches `systemPrompt`; no `system` role survives in `conversationMessages`). The failure log no longer interpolates recovered/caller-controlled prompt content. ✅

### Findings that were raised and their resolution
- **learn `push` failing under branch protection** (x2, duplicates) — **Withdrawn.** The `PUSH PERMISSION` note documents this as the intended, safe-by-default design: learn computes/commits locally and fails loudly until a maintainer grants the `github-actions[bot]` bypass — the deliberate human decision point. Nothing is lost and the run re-dispatches. Accepted.
- **Auto-commit of LLM-distilled memory to `release` (trust boundary)** — **Accepted as a documented risk / deferred policy decision.** The author's reply is precise: memory is read as text (never executed), groundedness/delivery gates prevent it manufacturing findings about untouched files or performing write tools with no capability mapping, and it can only bias emphasis. Whether learned memory needs human review before landing is a repository-owner policy decision, flagged for decision rather than silently shipped — a reasonable position for a version-bump PR.
- **Zip-slip in the artifact-restore step** (x2) — **Fixed.** Entry names are now validated (`^/|^\.\./|/\.\./|/\.\.$`) before extraction and traversing entries are refused; the overstated comment was corrected.
- **`decision` output contract on the pinned v5/v6 action** — **Verified.** The pinned action declares the output and derives it from the run's own JSON report; the guard fails in the safe direction (no verdict ⇒ no dismissal).
- **Vendored action / workflow.md / config-input findings** — **Moot.** The files they targeted are deleted in this PR's layout migration to `.yama/config.json` + `MCP.json` + `prompts.json`.

### The Yama v6.1.0 rewiring (layered into this branch)
Beyond the proxy fix, this PR migrates the review wiring to the prompts-driven v6 reviewer, adds committed long-term memory (Hippocampus), and chains `yama-learn` post-release. The most consequential part is the trust-boundary story around learn → memory → every future review, which is thoroughly documented and explicitly gated on a human decision (branch-protection bypass). That is the right shape: the risk is named, bounded, and not silently absorbed.

### Net
A focused correctness fix with good regression coverage, plus a larger (and well-documented) pipeline rewiring. Every substantive concern raised — including mine — is either fixed, verified, or deliberately deferred with justification. No open findings remain; I'm comfortable with this shipping once maintainers have consciously made the branch-protection bypass decision that the learn workflow is waiting on.

Tara-ag (approved):
Approve — the proxy fix (fold inline `system`-role messages into `systemPrompt`) is correct and covered on both the Claude and OpenAI paths; every finding raised is fixed, verified, or deliberately deferred with justification. The learn→memory trust-boundary risk is named, bounded, and explicitly gated on a human decision rather than silently shipped. Remaining decision (the branch-protection bypass for `github-actions[bot]`) is a maintainer policy call, not a code defect.

PR #1516, opened 2026-08-24
ci(lint): lint the whole project, not just src/ and one test glob

murdore:
## What

CI ran two scoped ESLint invocations and **never** `pnpm run lint`:

```
npx eslint src/ --max-warnings=300
npx eslint test/continuous-test-suite*.ts test/zod-schema-test-function.ts --max-warnings=10
```

So `scripts/`, `tools/`, `eslint-rules/`, `test/helpers/` (18 files) and any test file outside that glob were **not linted by CI at all**. The only full-project lint is the pre-commit hook, which `--no-verify` skips.

## Why this is more than missing coverage

The custom rules assume a full-project view. `unique-type-names` can only detect a duplicate when it sees **both** declarations in one run, and its own comment said:

> The pre-push/CI run on the full project catches everything.

Neither half of that was true:

- `.husky/pre-push` runs `check:deps`, `build` and three test suites. It lints nothing.
- CI never ran on the full project.

So the safety net that comment describes did not exist, and a rule whose whole purpose is cross-file detection was running against partial views in the one place that gates merges. The comment is corrected to describe what actually exists, with a note that CI's scope is now load-bearing for the rule rather than merely thorough.

## Why add rather than replace

The new `npx eslint .` sits alongside the existing two instead of replacing them, so their warning budgets stay exactly where they are. Collapsing them into one would mean a single ceiling governing both `src/` (300) and the test glob (10) — an easy way to tighten a limit in one place and silently loosen it in another.

The new run carries no `--max-warnings` deliberately: its job is to catch **errors** anywhere, which is what the custom rules emit. Warning budgets remain owned by the scoped runs.

## This turns nothing red

Measured, not assumed:

```
npx eslint .                              0 errors, 56 warnings, exit 0
npx eslint src/ --max-warnings=300        exit 0
npx eslint test/continuous-test-suite*.ts …  exit 0
```

Prettier remains covered by the separate `format:check` step. YAML re-parsed after the edit.

## How this was found

Auditing the custom lint rules for the "reports success without doing the work" shape this PR series has been chasing. The rules themselves are clean — no rule swallows an exception, and their early returns are legitimate scoping. The defect was one level up: the gate that runs them could not see most of the repo.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,6 +62,28 @@ jobs:
           echo "🔍 Running ESLint validation with controlled warning limits..."
           npx eslint src/ --max-warnings=300
           npx eslint test/continuous-test-suite*.ts test/zod-schema-test-function.ts --max-warnings=10
+          # Everything else, errors only.
+          #
+          # The two runs above cover src/ and the continuous-test-suite glob, and
+          # nothing else in the repo was linted by CI at all: scripts/, tools/,
+          # eslint-rules/, test/helpers/ (18 files) and any test file outside that
+          # glob. `pnpm run lint` does lint the whole project, but CI never calls
+          # it — it is a pre-commit hook only, and `--no-verify` skips that.
+          #
+          # This mattered beyond coverage: the custom rules assume a full-project
+          # view. unique-type-names can only detect a duplicate if it sees both
+          # declarations, and its own comment claimed "the pre-push/CI run on the
+          # full project catches everything" — pre-push does not lint, and CI did
+          # not run on the full project, so that net did not exist.
+          #
+          # No warning ceiling here on purpose: the scoped runs above already own
+          # the warning budgets, and duplicating one would let a warning limit be
+          # tightened in one place and silently loosened in another. This run
+          # exists to catch ERRORS anywhere, which is what the custom rules emit.
+          #
+          # Free to add today, measured rather than assumed: `npx eslint .`
+          # currently reports 0 errors / 56 warnings and exits 0.
+          npx eslint .
 
       - name: 🔒 Security & Environment Validation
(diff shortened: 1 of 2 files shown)


Tara-ag:
## Yama Review - APPROVED ✅

I've reviewed this PR thoroughly. It makes two coordinated changes:

1. **Adds `npx eslint .` to CI** - This ensures the entire repository gets linted, not just `src/` and one test glob. This matters because the `unique-type-names` custom rule needs to see ALL files simultaneously to detect cross-file duplicate type names.

2. **Updates the JSDoc comment** - Accurately describes what the CI actually does (and doesn't do). The old comment claimed "the pre-push/CI run on the full project catches everything" - but neither the pre-push hook nor CI previously linted the whole repo.

**Verification:**
- Confirmed the base branch lacks `npx eslint .`
- The new command has no warning limit, so it only catches ERRORS (what custom rules emit)
- Existing scoped runs with their warning budgets (300 for src/, 10 for tests) are preserved
- YAML syntax is valid
- Zero breaking changes

**Decision:** APPROVED - This is a clean CI improvement that fixes a real gap in linting coverage without any risk of breaking existing functionality.

### Impact Analysis
- **Blast radius**: 0 nodes impacted (configuration/documentation change only)
- **Breaking changes**: None
- **Performance impact**: ~5 seconds added to CI (measured: 0 errors, 56 warnings)
- **Risk level**: LOW

### Files Changed
1. `.github/workflows/ci.yml` - Added ESLint command
2. `eslint-rules/unique-type-names.cjs` - Updated documentation

No inline comments needed - no code issues found.

Tara-ag (approved):

Tara-ag (approved):

PR #1519, opened 2026-08-24
feat(providers): ride the SSE wire for LiteLLM generate()

pdogra1299:
## Summary
A non-streaming completion sends no bytes until it is fully done, so a proxy/tunnel idle limit (e.g. Cloudflare's ~100s 524 on tunneled LiteLLM gateways) kills any slow model — this is what fails the Yama review on qwen3.8-27b (see PR #1514). Callers like Yama cannot switch to `stream()` because their schema-bound contract needs the complete `GenerateResult` (`structuredData`, `toolExecutions`, `stopReason`, `usage`, JSON damage flags).

Fix: `doGenerate` in the OpenAI-compat base can now opt into `stream: true` on the wire (new `useStreamingWireForGenerate()` hook, default false; LiteLLM opts in) and aggregates the SSE via the existing `parseSSEStream` into the exact JSON-wire response shape — everything downstream (content parts, finish-reason mapping, usage clamping, schema coercion) is shared verbatim between the two wires, so callers see zero behavioral change beyond surviving slow completions.

- Request timeout now re-arms on every chunk (idle semantics) instead of capping total duration — `createTimeoutController` gained `reset()`.
- A backend whose 400 names `stream`/`stream_options` gets one retry on the plain JSON wire.
- Escape hatch: `NEUROLINK_LITELLM_SSE_GENERATE=false` restores the plain JSON wire.

## Test plan
- [x] `pnpm run check` + `check:tools-tests` — clean
- [x] `pnpm run build` — clean
- [x] `pnpm run test:providers-mocked` — 58/58, including 4 new contract tests: stream flags on the wire body, multi-chunk aggregation + usage, schema-bound `structuredData` from streamed JSON, stream-rejection fallback, and the env escape hatch
- [x] Non-litellm providers unaffected (hook defaults to false; existing mocked contract tests still pass on the JSON wire)

## Summary by CodeRabbit

* **New Features**
  * LiteLLM generation now uses streaming responses by default.
  * Streamed results preserve response details, usage, reasoning, tool calls, and structured output.
  * Streaming automatically retries with a standard request when unsupported.
  * Set `NEUROLINK_LITELLM_SSE_GENERATE=false` to restore JSON requests.

* **Bug Fixes**
  * Improved timeout handling during long-running streamed responses.
  * Preserved response IDs and model information from streamed results.

* **Tests**
  * Added coverage for streaming, usage aggregation, structured output, fallback behavior, and the configuration override.

### DIFF
--- a/docs/api/type-aliases/OpenAICompatBuildBodyArgs.md
+++ b/docs/api/type-aliases/OpenAICompatBuildBodyArgs.md
@@ -8,7 +8,7 @@
 
 > **OpenAICompatBuildBodyArgs** = `object`
 
-Defined in: [types/openaiCompatible.ts:318](https://github.com/juspay/neurolink/blob/release/src/lib/types/openaiCompatible.ts#L318)
+Defined in: [types/openaiCompatible.ts:322](https://github.com/juspay/neurolink/blob/release/src/lib/types/openaiCompatible.ts#L322)
 
 ## Properties
 
@@ -16,7 +16,7 @@ Defined in: [types/openaiCompatible.ts:318](https://github.com/juspay/neurolink/
 
 > **modelId**: `string`
 
-Defined in: [types/openaiCompatible.ts:319](https://github.com/juspay/neurolink/blob/release/src/lib/types/openaiCompatible.ts#L319)
+Defined in: [types/openaiCompatible.ts:323](https://github.com/juspay/neurolink/blob/release/src/lib/types/openaiCompatible.ts#L323)
 
 ---
 
@@ -24,7 +24,7 @@ Defined in: [types/openaiCompatible.ts:319](https://github.com/juspay/neurolink/
 
 > **messages**: [`OpenAICompatChatMessage`](OpenAICompatChatMessage.md)[]
 
-Defined in: [types/openaiCompatible.ts:320](https://github.com/juspay/neurolink/blob/release/src/lib/types/openaiCompatible.ts#L320)
+Defined in: [types/openaiCompatible.ts:324](https://github.com/juspay/neurolink/blob/release/src/lib/types/openaiCompatible.ts#L324)
 
 ---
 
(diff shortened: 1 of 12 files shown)


Tara-ag:
Starting review of PR #1519. Analyzing changes file by file following Yama methodology.

Tara-ag (approved):

Tara-ag:
✅ **Approved** — The PR implements SSE wire support for LiteLLM's non-streaming generate() method, solving the critical problem of tunnel idle timeouts killing slow completions.

## Summary of Changes

This PR adds streaming-capable wire transport for LiteLLM's `generate()` operation while maintaining full backward compatibility. The key insight: a non-streaming completion sends nothing until complete, so proxies with idle limits (e.g., Cloudflare's ~100s 524) will kill it. By using SSE wire with aggregation, bytes flow continuously while the SDK aggregates into the same complete result callers expect.

### What Changed

1. **Base Provider (`openaiChatCompletionsBase.ts`)**: Added `useStreamingWireForGenerate()` hook (default false), conditional SSE aggregation in `doGenerate()`, and 400 fallback retry logic
2. **LiteLLM Provider**: Overrides hook to true by default, with env var escape hatch
3. **Timeout Controller**: Added `reset()` for idle-timeout semantics  
4. **SSE Client**: Enhanced `parseSSEStream` to capture id/model from first chunk
5. **Types**: Added optional `id`/`model` fields to `OpenAICompatSSEResult`
6. **Tests**: 4 new contract tests cover all scenarios

### Verification Performed

- ✅ No breaking changes to public API
- ✅ SSE aggregation matches JSON response structure verbatim
- ✅ Idle timeout re-arms on each chunk (verified in code)
- ✅ 400 fallback only retries once when error mentions "stream"
- ✅ Escape hatch works via env var
- ✅ Structured output coercion tested end-to-end
- ✅ Tool call handling preserved (uses existing buffered mechanism)

### Impact Analysis

- **Affected files**: 12 (6 docs auto-generated, 5 source, 1 test)
- **Changed APIs**: None breaking; internal implementation detail
- **Downstream consumers**: Zero behavioral change beyond fixing slow completions
- **Test coverage**: 4 new E2E tests + existing suite still passes

### Review Scope

Reviewed all 12 changed files systematically:
- Type additions are additive and documented
- Implementation follows existing SSE/streaming patterns
- Error handling mirrors streaming path
- Tests comprehensively cover happy path, edge cases, and failures

No issues found requiring changes.

Tara-ag (approved):

Tara-ag (approved):

PR #1517, opened 2026-08-24
fix(openai): stop reporting an exhausted quota as a retryable rate limit

murdore:
OpenAI returns HTTP 429 for two unrelated conditions: transient throttling, and a **permanently** exhausted quota (out of credit, or a spend cap reached). The error table matched on the status code alone, so the second was reported as the first — and the advice it gave, "try again later", is the one thing that can never resolve it.

## Measured on a real out-of-credit account

Same call, both builds, live against `api.openai.com`:

| | elapsed | message |
|---|---|---|
| **before** (installed release 11.22.3) | **23.2s** | `OpenAI rate limit exceeded. Please try again later.` |
| **after** (this branch) | **1.0s** | `OpenAI quota exhausted — this will not resolve by retrying. Check your plan and billing details at …` |

The 23 seconds are the retry ladder. `MAX_PROVIDER_RETRIES` is 2 and OpenAI sends no `Retry-After` for these, so `NO_HINT_FLOOR_MS` applies twice: three upstream calls and ~20s of sleeping, **on every request**, for a state that cannot change until someone tops up the account.

## Two independent defects

**The message.** The 429 rule in `openAI/client.ts` had no quota case — although the same rule already exists for xAI in `openaiCompatCatalog.ts:130`, and for Cohere and Replicate. The direct provider was the gap.

**The retries.** This is the part that surprised me, and it is worth recording because the obvious fix would not have worked. The *classified* `RateLimitError` is already non-retryable — I probed it. The ladder never sees that error. It acts on the **raw** SDK error, where `isRetryableProviderError` consults `APICallError.isInstance(error) → error.isRetryable` **first**, and the AI SDK sets that flag from the status code alone. So the quota 429 was marked retryable before any NeuroLink logic ran. Neither changing the error class nor giving it a `retriable: false` field would have changed anything; the check has to short-circuit ahead of every branch.

## Keyed on the type, never the word "quota"

OpenAI and xAI send `type: "insufficient_quota"` for a permanent billing state. Other providers use "quota" for ordinary throttling — Google's 429 reads `Quota exceeded for quota metric …`, which **is** retryable.

Matching free text would have made Gemini throttling non-retryable and quietly removed a layer of resilience — a worse bug than the one being fixed. There is a test that fails if that regresses.

## The type is not where you would look for it

[...]

PR #1521, opened 2026-08-24
feat(lint): enforce Critical Rule 6 — formatProviderError must return, never throw

murdore:
## What

CLAUDE.md's enforcement table lists rules 2 and 7–14 as lint-enforced. **Rule 6 was the odd one out** — documented as non-negotiable, checked by nobody. This adds `neurolink/format-provider-error-returns`.

## Why it matters concretely

`baseProvider` calls `this.formatProviderError(error)` and then inspects the returned value before deciding how to surface it. A formatter that *throws* skips that inspection entirely — and does so from inside a catch block, replacing the original provider failure with whatever the formatter produced. The caller loses both the decision and the original error.

## Audited before writing the rule

24 implementations across `src/lib/providers`, **zero violations**. So this is a ratchet rather than a cleanup: it turns nothing red today and stops the first regression.

## The detector was proven before being trusted

A `throw` planted inside lmStudio's formatter:

```
74:18  error  `formatProviderError` must RETURN the error, not throw it …
       neurolink/format-provider-error-returns
exit 1
```

and removing it returns exit 0. Without that step I would only have shown that a rule which reports nothing reports nothing.

## Seven RuleTester cases

Each states its expected outcome up front, so a silent rule cannot pass:

```
throw at the top of a method                     -> flagged
throw nested inside an if                        -> flagged
throw inside a catch within the formatter        -> flagged
arrow property form                              -> flagged
returns an Error                                 -> clean
throw in a DIFFERENT method                      -> clean
caller doing `throw this.formatProviderError(e)` -> clean
```

The last two are the ones worth having. The rule must not fire at the **call site**, where throwing the formatted error is exactly right — that distinction is the whole point of Rule 6, and a naive "no throw near formatProviderError" check would get it backwards.

## Scope note

Nested throws are covered deliberately: a `throw` inside an `if` or a `catch` within the formatter escapes just as readily as one at the top, so the rule tracks function depth rather than scanning only the immediate body.

## Verification

```
typecheck   4822 files / 0 errors
eslint .    exit 0
```

Contributes no warnings — the repo's count is unchanged at 57, verified by comparing the per-rule breakdown against clean release rather than against my own earlier note, which was stale (56).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -61,22 +61,23 @@ These are non-negotiable. Violating them breaks the build or introduces bugs.
 
     **The one exception is determinism.** A test may sit outside this rule only when it needs deterministic control that a live call cannot give — a pure translation table, a fixed set of inputs, a recorded backend. The vector-store suites are the standing example: they drive real backends (pglite in-process Postgres, recorded fixtures) and cover filter-dialect translation that no live `generate()` could be made to emit. Convenience, speed, and "it is easier to assert on the internal" are not exceptions. When you take the exception, say so in the file's header and name what determinism buys.
 
-**Enforcement:** Rules 2 and 7-15 are enforced via ESLint. Rules 2, 7-13 and 15 use custom rules in `eslint-rules/`; rule 14 uses core `no-restricted-syntax` AST selectors in `eslint.config.js`. Run `pnpm run lint` (or the pre-commit hook) — no shell scripts, no regex heuristics, everything AST-based.
+**Enforcement:** Rules 2, 6 and 7-15 are enforced via ESLint. Rules 2, 6, 7-13 and 15 use custom rules in `eslint-rules/`; rule 14 uses core `no-restricted-syntax` AST selectors in `eslint.config.js`. Run `pnpm run lint` (or the pre-commit hook) — no shell scripts, no regex heuristics, everything AST-based.
 
 Rule 15's determinism exception is the `allow` list on `neurolink/e2e-tests-only` in `eslint.config.js`. Adding a file to it is a review decision, and the file's own header must say what determinism buys — it is not a way to silence the rule. The rule ignores type-only imports (`import type`, and `{ type A }` where every specifier is type-only) because they are erased and assert nothing.
 
-| Rule     | ESLint rule                              |
-| -------- | ---------------------------------------- |
-| 2        | `neurolink/no-local-type-alias`          |
-| 7        | `neurolink/no-interface`                 |
-| 8        | `neurolink/no-types-suffix-filename`     |
-| 9        | `neurolink/unique-type-names`            |
-| 10       | `neurolink/types-barrel-exports-only`    |
-| 11 & 11b | `neurolink/no-local-types-folder`        |
-| 12       | `neurolink/no-type-export-outside-types` |
-| 13       | `neurolink/barrel-type-imports`          |
-| 14       | `no-restricted-syntax` (AST selectors)   |
-| 15       | `neurolink/e2e-tests-only`               |
+| Rule     | ESLint rule                               |
+| -------- | ----------------------------------------- |
+| 2        | `neurolink/no-local-type-alias`           |
+| 6        | `neurolink/format-provider-error-returns` |
+| 7        | `neurolink/no-interface`                  |
+| 8        | `neurolink/no-types-suffix-filename`      |
+| 9        | `neurolink/unique-type-names`             |
+| 10       | `neurolink/types-barrel-exports-only`     |
+| 11 & 11b | `neurolink/no-local-types-folder`         |
+| 12       | `neurolink/no-type-export-outside-types`  |
+| 13       | `neurolink/barrel-type-imports`           |
+| 14       | `no-restricted-syntax` (AST selectors)    |
+| 15       | `neurolink/e2e-tests-only`                |
 
 ---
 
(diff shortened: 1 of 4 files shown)


Tara-ag (approved):

Tara-ag (approved):

murdore:
Amended in `ffd88d68` to include the documentation half, which I had left out — a lint rule that is not in the enforcement table is only half shipped, since that table is how the next person learns the rule exists.

```
**Enforcement:** Rules 2, 6 and 7-15 are enforced via ESLint …

| 6        | `neurolink/format-provider-error-returns` |
```

The table rows are re-aligned by prettier because the new rule name is the longest in the column — that accounts for the otherwise-surprising 14/13 line diff on a two-line change. Verified the diff contains nothing but the enforcement sentence, the new row, and that realignment.

Also rebased onto current release and **re-proved the rule after the rebase** rather than assuming it survived: a `throw` planted in lmStudio's formatter still gives exit 1, and removing it gives exit 0.

One correction while I am here. In an earlier note I said CLAUDE.md described Rule 15 as review-only and that this was a doc/reality mismatch. That was wrong — I was reading a stale copy carried in my own context. The file on release already says *"Rules 2 and 7-15 are enforced via ESLint"* and already lists `neurolink/e2e-tests-only` for Rule 15. There was no defect there, and I would rather retract it than leave a phantom issue on the record.

Gates: `eslint .` exit 0, typecheck 4822 files / 0 errors, prettier clean. Four files.

murdore:
Not an objection — the rule looks right to me and the scoping is careful (detecting `throw` in the body, and explicitly not matching the bodiless abstract declarations). Landing it is a clear improvement over Rule 6 being the one critical rule without a lint.

One piece of evidence that bears on the rule's stated rationale, found while working on #1517. Sharing it because it changes what "Rule 6 enforced" covers, not whether the rule is correct.

The docstring's justification is:

> `baseProvider` calls `this.formatProviderError(error)` and then inspects the returned value before deciding how to surface it. If the formatter throws, that inspection never runs...

That holds wherever the method is actually called. On the **streaming path of `openaiChatCompletionsBase`, it never is**:

- `openaiChatCompletionsBase.ts:134` declares `protected abstract formatProviderError(error: unknown): Error;`
- the streaming error path throws the raw provider error instead — `throw apiErr;` at `:1474`
- `grep formatProviderError src/lib/providers/openaiChatCompletionsBase.ts` returns only the abstract declaration and a mention in the file's header comment — no call site

For contrast, Anthropic's streaming path does call it: `anthropic/client.ts:2270`, `throw this.formatProviderError(error)` inside the `runLoop().catch(...)` handler.

Everything inheriting `openaiChatCompletionsBase` is affected — 15 provider files plus the 7 catalog-driven providers (OpenAI, Azure, DeepSeek, Ollama, LM Studio, llama.cpp, OpenRouter, LiteLLM, NVIDIA NIM, HuggingFace, Cohere, Groq, xAI, Together, Fireworks, Perplexity, Mistral, Cloudflare).

Measured rather than inferred — a mocked streaming 429 through the OpenAI provider surfaces the upstream text verbatim:

```
non-streaming:  [openai] OpenAI quota exhausted — this will not resolve by retrying...
streaming:      You have no credits remaining. Add credits to continue using the API.
```

No `[provider]` tag, no classification, for any error type.

So this rule correctly guarantees that the 24 implementations return rather than throw. What it cannot show — and shouldn't be read as showing — is that the formatter runs on every path that produces a provider error. For the majority of providers, streaming bypasses it entirely.

That's a separate fix and I'm not proposing it here. Flagging it only so the ratchet isn't mistaken for wider coverage than it has; a follow-up making the OpenAI-compat streaming path call the formatter would make the rule's rationale true for those providers too. Happy to take that on if it isn't already yours — I have the reproduction.

murdore:
Correction to my comment above — I over-stated the mechanism, and since it was about your PR's rationale I want it fixed rather than left standing.

**What I actually measured, and still stand behind:** an OpenAI streaming 429 surfaces the upstream text verbatim, with no `[provider]` tag and no classification. Two message variants, both through `nl.stream()` against a mocked 429 with `type: "insufficient_quota"`:

```
"You have no credits remaining. Add credits to continue using the API."
"You exceeded your current quota, please check your plan."
```

Non-streaming, same provider, same error type, classifies correctly: `[openai] OpenAI quota exhausted — ...`

**What I got wrong:** I wrote that "no error of any type, for any OpenAI-wire provider, is classified on the streaming path", and inferred it from `formatProviderError` having no call site inside `openaiChatCompletionsBase.ts`. That inference doesn't hold. `baseProvider.ts:445` *does* call `this.handleProviderError(realStreamError)` on a streaming failure — it's simply gated, at `:434-443`, on the raw message containing one of `abort` / `timeout` / `401` / `403` / `quota` / `rate limit` / `authentication`.

So my second hypothesis was that the gate's substring list was missing OpenAI's real credit-exhaustion wording. I tested that too, by re-running with a message containing the literal word "quota" — and it *still* surfaced raw. So that explanation is wrong as well.

Two hypotheses, both measured, both wrong. The observation is solid; the cause is not yet established, and I'm not going to guess at it a third time in a comment on someone else's PR.

**What this means for #1521:** less than my first comment implied. The rule is correct and well-scoped — I'd land it. The one thing I'd still say is narrower and safe: Rule 6 guarantees the 24 implementations *return rather than throw*, which is exactly what it claims. Whether the formatter is reached on every path that produces a provider error is a separate question, and there is at least one measured case (OpenAI streaming) where the classified message doesn't reach the caller. That's a follow-up to investigate properly, not a caveat on this rule.

Sorry for the noise on your PR. The first comment asserted a mechanism I hadn't verified, which is the same error the rule you're adding exists to prevent — a claim that looks checked and isn't.

PR #1522, opened 2026-08-24
fix(tts): let callers see why synthesis failed

murdore:
## The bug

`generate()` degrades gracefully when TTS fails: it returns the text, omits `audio`, and throws nothing. `BaseProvider` records what went wrong on `ttsMetadata`. But the result builders in `neurolink.ts` rebuild the result **field by field** and never forwarded it, so every caller reading `result.ttsMetadata` got `undefined`.

The consequence: **an invalid TTS key is completely silent to an SDK consumer.** No audio, no error, and no way to distinguish a rejected credential from a request that never asked for audio.

Probed on a real call — note `audio` is present and `ttsMetadata` is not:

```
result keys: content, ..., analytics, imageOutput, evaluation, audio,
             transcription, video, avatar, music, ppt, reasoning, reasoningTokens
has audio: false
has ttsMetadata: false
```

This is the same defect the comment directly above `reasoning:` in that builder already describes — *"the builder previously dropped them on the floor"* — on a different field.

## The fix

Forwarded at **all four** sites that forward `audio`, and declared on `GenerateResult` and `TextGenerationResult`; both carried it at runtime and neither declared it. The duplicate declaration on `EnhancedGenerateResult` is removed now that `GenerateResult` owns it, so there is one definition rather than two to keep in step.

## What it buys the test suite

The Fish Audio case reported `no audio buffer in result` — a symptom, with the cause visible only in a log line:

```
Fish Audio synthesis failed: 401 — {"status":401,"message":"Invalid Token"}
```

It now reads the diagnosis: skips an upstream credential failure, still fails a real one.

**Keyed on the HTTP status, not the message text.** Neither this file's local matcher nor the shared `isExpectedProviderError()` recognises `"Invalid Token"`, and adding it as a substring is exactly the loose matching `envGuard` warns against — it would swallow any consumer-facing error phrased that way. A 401/403 from a synthesis call is unambiguous.

## Proven in both directions

| scenario | before | after |
|---|---|---|
| invalid key (401) | FAIL, `no audio buffer in result` — undiagnosable | **SKIP**, with the real reason |
| non-credential TTS failure | FAIL, undiagnosable | **FAIL**, with the real reason |
| forwarding removed | — | **FAIL**, `no ttsMetadata` |

The middle row is the one that matters — **a test that always skips is worse than one that always fails.** With the status match disabled (`401|403` → `9401|9403`), the identical condition reports:

```
[FAIL] TTS - Fish Audio end-to-end — no audio buffer — Fish Audio synthesis failed: 401 — ...
Passed: 17   Failed: 1   RESULT: FAIL
```

So the skip is conditional, not unconditional.

## Verification

```
pnpm run check              4824 files, 0 errors
pnpm run build              exit 0
pnpm run check:tools-tests  clean
pnpm run test:tts           17 passed / 1 skipped   (was 17 passed / 1 failed)
```

## Summary by CodeRabbit

* **New Features**
  * Generation results now include optional text-to-speech metadata, including synthesis status, errors, and latency.
  * TTS metadata is preserved across supported generation workflows.
* **Bug Fixes**
  * Improved handling and reporting of missing audio in TTS tests, including clearer provider or credential-related diagnostics.
* **Documentation**
  * Updated API documentation links and descriptions to reflect the latest generation result fields and TTS metadata.


Tara-ag (commented):
Reviewing PR #1522 - TTS metadata field additions and test improvements

## Summary
This PR adds `ttsMetadata` support to the result types and improves error handling in TTS tests. The changes include:
- Adding `ttsMetadata` field to `GenerateResult` and `TextGenerationResult` types
- Forwarding `ttsMetadata` through result builders in neurolink.ts
- Improving TTS error handling with new `explainMissingAudio()` helper function

## Review Scope
Reviewed files:
- Documentation updates (line number corrections)
- Type definitions (`src/lib/types/generate.ts`)
- Implementation (`src/lib/neurolink.ts`)  
- Test suite (`test/continuous-test-suite-tts.ts`)

## Findings
- 🔒 MINOR: Documentation line numbers out of sync with source code (see inline comment)

## Decision
The core functionality changes are sound and follow existing patterns. The type additions are minimal and don't break backward compatibility. The test improvements provide better error diagnostics for TTS failures. Documentation issues are non-blocking maintenance items.

**Status: APPROVED** (with minor documentation notes)

Tara-ag:
## Review Summary

**Decision: APPROVED** ✅

This PR adds `ttsMetadata` field to expose TTS synthesis failure reasons, improving error diagnosis for TTS providers (Cartesia, Google TTS).

### Changes Overview:
- **Type definitions**: Added `ttsMetadata?: TTSMetadata` to `GenerateResult`, `TextGenerationResult`, and `CliGenerateResult` types
- **CLI output**: Updated `buildCliOutput()` to forward `ttsMetadata` in JSON/text outputs
- **Tests**: Enhanced TTS tests to use `explainMissingAudio()` helper for better failure diagnosis

### Impact Analysis:
- **Backward compatible**: All changes add optional fields, no breaking changes
- **No out-of-diff impact**: Only documentation and test updates; core SDK behavior unchanged
- **Self-contained**: No dependents affected outside this PR

### Review Scope:
- Reviewed type system consistency ✅
- Verified backward compatibility ✅  
- Checked CLI integration ✅
- Validated test improvements ✅
- Confirmed no security or correctness issues ✅

**No issues found. Ready to merge.**

Tara-ag (approved):

Tara-ag (approved):

PR #1520, opened 2026-08-24
fix(test): assert the React hooks through the subpath that publishes them

murdore:
## The bug

The React-hooks case imported `../dist/index.js` — the **root barrel** — and asserted eight hooks are exported there:

```ts
const clientModule = await import("../dist/index.js");
const hooks = ["NeuroLinkProvider", "useNeuroLinkClient", "useChat", ...];
```

They are not on the root barrel, and should not be — putting React on the default entry pulls it into every consumer of the package. They are published under the **`./client`** subpath.

So the assertion failed permanently while the code was correct. `test:client` has been reporting **12 passed / 1 failed** on `release`. It went unnoticed because `test:client` is one of the **59 of 84** suite scripts CI never runs.

## Why the package specifier, not a relative path

`import("@juspay/neurolink/client")` rather than `import("../dist/client/index.js")` covers one thing more: resolution goes **through the `exports` map in `package.json`**, so a broken or mis-pathed subpath entry fails here too.

Node resolves it by self-reference to this repo's own build — confirmed as `<repo>/dist/client/index.js` via `import.meta.resolve`, with no `@juspay` in `node_modules` to shadow it. So it is still the built output rule 15 requires, not `src`.

## Proven in both directions

|  | old test | new test |
|---|---|---|
| code as-is | **FAIL** — the bug | **PASS** |
| `exports["./client"]` repointed at a missing file | n/a | **FAIL** |

The second row is what justifies the package specifier. With `./client` pointed at a nonexistent file:

```
[FAIL] React hooks are importable — Cannot find module .../dist/client/WRONG.js
✗ React hooks exports
Passed: 12   Failed: 1   RESULT: FAIL
```

A relative `dist` import would have passed straight through that.

## Result

```
before   Passed: 12   Failed: 1   RESULT: FAIL
after    Passed: 13   Failed: 0   RESULT: PASS
```

Both hook cases now pass, including the one that proves wiring:

```
[PASS] React hooks are importable — All 8 hooks exported as functions
[PASS] React hooks invocation — useChat() outside renderer threw React-internal hook error
```

One test file. No `src` change.

## Summary by CodeRabbit

* **Tests**
  * Updated export validation to use the package’s public client entry point.
  * Added coverage confirming that calling the chat hook outside a valid React renderer produces the expected error.

PR #1505, opened 2026-08-24
fix(rag): await registration before reading factory metadata

murdore:
## The bug

`ChunkerFactory.getChunkerMetadata()` and `getDefaultConfig()` are **synchronous** reads of `metadataMap`. That map is only populated by `registerAll()`, which runs solely through the **async** `ensureInitialized()` — and the only place that awaited it was `create()` (`src/lib/core/infrastructure/baseFactory.ts:41`).

So on a fresh process both return `undefined` for every strategy, and start returning real metadata the moment anything happens to construct a chunker first. Order-dependent, silent, and `undefined` rather than a throw.

Proven cold, in a process that does nothing else:

```
chunker  metadata: undefined
chunker  config  : undefined
reranker metadata: undefined
reranker config  : undefined
```

**`RerankerFactory` has the identical defect.** Its equivalent test passes only because an earlier test in the same section calls `createReranker()` and warms the map — reorder that section and it fails. That test was never real coverage.

## The fix

`createChunker()`, `getAvailableStrategies()` and `getAvailableTypes()` were **already** async and already awaited `ensureInitialized()`, for exactly this reason. This makes the rest of each factory's read surface match rather than inventing a new pattern:

| factory | now awaits initialization |
|---|---|
| `ChunkerFactory` | `getChunkerMetadata`, `getDefaultConfig`, `getStrategyAliases`, `hasStrategy`, `getChunkersForUseCase`, `getAllMetadata` |
| `RerankerFactory` | `getRerankerMetadata`, `getDefaultConfig`, `getTypeAliases`, `hasType`, `getRerankersForUseCase`, `getLocalRerankers`, `getModelFreeRerankers`, `getAllMetadata` |

## Not a public API break

These are not reachable from the built barrel at runtime:

```
getChunkerMetadata       undefined
getDefaultConfig         undefined
getFactoryDefaultConfig  undefined
ChunkerFactory           undefined
chunkerFactory           undefined
```

The only callers anywhere are `test/continuous-test-suite-rag.ts` and the internal re-export barrels. `getStrategyAliases` and `getChunkersForUseCase` have **zero** callers outside the factory itself.

## Verification

`pnpm run test:rag`: **98 passed / 2 failed → 100 passed / 0 failed.**

```
✗ Get chunker metadata   — Incomplete metadata      (before)
✗ Get default config     — Invalid config returned  (before)
```

Those failures were **real**, not test bugs. They went unnoticed because `test:rag` is one of the **59 of 84** suite scripts CI never runs.

[...]

PR #1523, opened 2026-08-24
ci(test): run 20 more credential-free suites in extended-suites

murdore:
## Why

**59 of the repo's 84 suite scripts never execute on a pull request.** That isn't theoretical — it is why, this week alone:

- `test:rag` sat at **98 passed / 2 failed** on `release` (fixed in #1505)
- `test:client`'s React-hooks case had been asserting against the wrong entry point long enough that nobody noticed (fixed in #1520)
- `test:tts` reported `no audio buffer in result` with the real cause visible only in a log line (fixed in #1522)

## How the 20 were chosen

Not by reading them. **Every suite script CI never invoked was run with the environment fully stripped** — `env -i` plus `DOTENV_CONFIG_PATH=/dev/null`, checked to actually suppress `.env` rather than assumed:

```
DOTENV_CONFIG_PATH=/dev/null  ->  OPENAI_API_KEY seen: false
(control)                     ->  OPENAI_API_KEY seen: true
```

Only suites that passed that way are included. They contribute **373 assertions**, roughly doubling this job's coverage (8 suites → 28).

## What was excluded, and why that matters

Two candidates pass on a developer machine and are deliberately absent — this is the whole argument for measuring rather than assuming:

| suite | with credentials | without |
|---|---|---|
| `test:proxy` | 74 passed / 6 skipped | **68 passed / 5 FAILED / 7 skipped** |
| `test:multimodal:sdk` | passes | **exits non-zero, no summary** |

Either would have looked safe from a local run and broken the job.

`test:model-not-found-retryable` is also absent: it skips entirely without credentials, so it could only ever no-op — the same reason it is held out of `security-suites`.

**`test:middleware` is held back despite passing the sweep.** Inside the full sequence it failed **2 of 2** runs; alone, or immediately after `test:mcp`, it passed **3 of 3** at 7/7. That is reproducible in context rather than flaky. The obvious hypothesis — that the preceding suite leaves servers behind — was tested and refuted (`test:mcp` PASS → `test:middleware` 7/7 PASS back to back). Its 60s per-test bound is the likeliest thing to give way over a long sequence. Worth 7 assertions to whoever diagnoses it.

## Why `extended-suites` and not `security-suites`

`security-suites` **is a required check** — worth stating precisely, because the ruleset alone does not show this:

```
ruleset 11413189            test, provider-safety-net, build-check, Single Commit Policy
classic branch protection   ...those four PLUS security-suites
```

GitHub enforces the union. These 20 suites are *measured once, not proven stable*, and a required check that breaks blocks every open pull request — the failure mode CLAUDE.md already records for ffmpeg. `extended-suites` is non-blocking and already exists for exactly this purpose: *"No provider credentials are set in this job, and that is the point."*

It also runs every suite before failing, so one break cannot hide the state of the others.

## Measurement conditions

Worth recording: the machine was **saturated throughout** — load average 31 on 18 cores, 6.6% idle, from unrelated processes. All 20 passed twice under that. Only `test:middleware` did not, which is why it is the sole exclusion on stability grounds.

## Summary by CodeRabbit

- **Tests**
  - Expanded automated validation to cover 20 additional passing test suites.
  - Improved suite selection for environments without credentials by excluding unsuitable or no-op checks.
  - Documented a known sequencing issue affecting middleware tests.

- **Chores**
  - Enhanced non-blocking continuous integration checks with clearer test execution guidance and broader coverage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -496,6 +496,12 @@ jobs:
   # into nothing is documentation, not a test.
   extended-suites:
     runs-on: ubuntu-latest
+    # This job checks out and executes repository code on pull_request.
+    # The repository default is a WRITE-scoped token (and one that may
+    # approve reviews); nothing here needs either. Read is enough to
+    # checkout, install and run suites.
+    permissions:
+      contents: read
     name: Extended Suites (non-blocking)
     steps:
       - name: Checkout code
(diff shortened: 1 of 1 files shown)


Tara-ag:
## Review Decision: ✅ **APPROVED**

### Summary
This PR adds 20 credential-free test suites to the `extended-suites` CI job, expanding coverage from 8 to 28 suites and adding 373 previously unmeasured assertions. The change is purely additive CI configuration with zero impact on the codebase.

### Findings
No issues found. This is a clean CI improvement that:
- ✅ Expands test coverage in non-blocking `extended-suites` job
- ✅ Documents exclusion rationale with measured data (test:proxy fails without credentials)
- ✅ Notes known sequencing issue with test:middleware (reproducible context failure)
- ✅ All 20 added test suites verified to exist and follow E2E testing conventions
- ✅ YAML syntax valid with excellent inline documentation

### Impact on Existing Code
- **Blast radius**: 0 files affected (pure CI config change)
- **Execution flows**: None impacted (CI workflow only)
- **Downstream dependents**: None (no code changes)
- **Architectural hotspots**: None touched

### Review Scope
Reviewed:
- ✅ File-by-file diff analysis (1 file)
- ✅ YAML structure and syntax validation
- ✅ Test suite existence verification
- ✅ Documentation quality and clarity
- ✅ Exclusion rationale transparency
- ✅ No secrets or security concerns
- ✅ Zero breaking changes
- ✅ Impact analysis via code graph (none applicable for CI config)

---

**No inline comments needed** — this is a clean addition with no issues requiring feedback.

Tara-ag (approved):

Tara-ag (approved):

PR #1530, opened 2026-08-24
fix(core): classify provider errors that surface during stream iteration

murdore:
A provider failure reached streaming consumers as raw upstream text, while the identical failure through `generate()` was classified normally:

```
streaming       You have no credits remaining.
non-streaming   [openai] OpenAI quota exhausted — this will not resolve by retrying.
                Check your plan and billing details at ...
```

## Mechanism

Four hypotheses were disproved by instrumentation before the real one turned up. Recording them so nobody re-treads the path:

1. *"`formatProviderError` is never called on the stream path"* — false, `baseProvider.ts:445` calls it.
2. *"the substring gate at `baseProvider.ts:434-443` misses OpenAI's credit wording"* — false; re-ran with a message containing the literal word `quota`, still raw.
3. *"the fix site is the `throw err` in `wrapStreamWithLifecycleCallbacks`"* — false; added a method there, stubbed it, **0 calls**.
4. *"`wrapStreamWithLifecycleCallbacks` isn't applied"* — false; stubbed it, **exactly 1 call**.

3 and 4 were both individually correct and jointly misleading. The actual chain:

`stream()` awaits only the **construction** of the provider's stream. `openaiChatCompletionsBase` returns an *uninvoked* async generator, so the failure is discovered lazily on first pull and `stream()`'s own try/catch never fires. `wrapStreamWithLifecycleCallbacks` then hit its early return — no `onChunk`/`onFinish`/`onError` registered — and handed back **the provider's own generator by reference**, so the wrapping generator holding the only capable catch was never built. Everything downstream is a proven passthrough: `neurolink.ts:10077` re-throws the same object and `createMCPStream` passes the stream through unwrapped.

Verified by reference identity: the object thrown at `executeStream`'s boundary is `===` the object the consumer's `for await` catches.

## The fix

The early return no longer discards error handling. That wrapper is the only point every provider's stream passes through unconditionally — four call sites, all in this file, method is `private`, covering the real-streaming path and all three fake-streaming paths. One change instead of ~20 provider files with differently shaped generators.

Classification is guarded three ways, **each proven necessary by its own measurement**:

| guard | evidence it is required |
| --- | --- |
[...]

PR #1531, opened 2026-08-24
fix(anthropic): stop a streaming error from killing the caller's process

murdore:
A streaming failure left an unhandled promise rejection behind, and an unhandled rejection **terminates the process**.

Measured with no `unhandledRejection` handler installed:

```
consumer's catch fired : YES   the caller handled its error correctly
process survived       : NO
exit code              : 1     ERR_UNHANDLED_REJECTION
```

A caller doing everything right — `try`/`catch` around the iteration — still died, with **no way to defend against it from outside this library**.

## Not the guarded promise

`client.ts:2314` already installs `loopPromise.catch(() => {})` for exactly this hazard, so that was the obvious suspect and it was wrong. Identity-testing showed two different error objects:

```
consumer   RateLimitError  "[anthropic] Anthropic rate limit exceeded..."   NeuroLink, formatted
unhandled  RateLimitError  "429 {\"type\":\"error\"...}"                        Anthropic SDK, raw
SAME OBJECT? false
```

`loopPromise` rejects with the **formatted** error and is handled. The raw one belongs to a different promise.

Streaming-only, discriminated by experiment: `generate()` produced **0** unhandled rejections, `stream()` produced **1**.

## Root cause

`pump` is detached — it begins draining the engine's channel the moment it is created, and `await pump` is the only thing that adopts its rejection. That await sits *after* `await resultPromise`:

```js
const pump = (async () => { for await (const chunk of stream) { ... } })();
const result = await resultPromise;   // throws
await pump;                           // never reached
```

Proven before writing the fix: attaching a bare `pump.catch(() => {})` took unhandled rejections from **1 → 0** while the consumer still received the same formatted error.

## The fix

Drain `pump` on the error path. This is the shape the codebase already uses everywhere else — `googleAiStudio/client.ts` and **all four** `googleVertex` pump sites do exactly this. Anthropic was the only provider missing it.

## Sweep

Since this is a class of bug rather than a one-off, I swept for it. Eleven providers, all `unhandled=0`:

```
openai · anthropic · groq · xai · mistral · openrouter · deepseek
google-ai · ollama · bedrock
```

**The detector was validated** — a clean sweep proves nothing otherwise. Reverting this guard and rebuilding gives `unhandled=1`; restoring gives `0`.

Not measured, stated plainly rather than claimed: `googleVertex` (only a fake service-account key was available, so it failed at auth before reaching the streaming path — its four pump sites are guarded by inspection), SageMaker, and the media providers.

## The test

It asserts the caller's **process survives**, in a **subprocess**. An in-process `unhandledRejection` counter would be polluted by any other case in the file leaving a stray rejection — it could pass for the wrong reason. A child's exit code cannot.

Confirmed non-vacuous: reverting the fix fails it with `exit=1`, which is the process death itself.

## Verification

`pnpm run build` 0 errors · `check:tools-tests` 0 errors · `lint` 0 errors · mocked provider contract suite **64 passed / 0 failed** · `docs/api` regenerated with both halves, **zero drift**.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling for Anthropic streaming requests.
  - Rate-limit failures are now reported as a single clear error without unexpectedly terminating the application process.
  - Preserved the original error details when streaming requests fail.

- **Tests**
  - Added regression coverage for safely catching streaming failures.
  - Added validation for consistent rate-limit error reporting and provider attribution.

### DIFF
--- a/src/lib/providers/anthropic/client.ts
+++ b/src/lib/providers/anthropic/client.ts
@@ -2250,7 +2250,27 @@ export class AnthropicProvider extends BaseProvider {
         }
       })();
 
-      const result = await resultPromise;
+      // `pump` is detached: it starts draining the engine's channel the moment
+      // it is created, and `await pump` below is the only thing that adopts its
+      // rejection. When resultPromise rejects, that line is never reached, so
+      // pump's rejection stays unhandled — and an unhandled rejection
+      // TERMINATES the consumer's process. Measured: a caller that correctly
+      // try/catches a streaming error still died with ERR_UNHANDLED_REJECTION,
+      // exit code 1, with no way to defend against it from outside this
+      // library. The rejection carried the raw SDK error, distinct from the
+      // formatted one the caller received, which is why the existing
+      // `loopPromise.catch` guard below does not cover it.
+      //
+      // Same shape googleAiStudio/client.ts and googleVertex/client.ts already
+      // use at every one of their pump sites; Anthropic was the only provider
+      // missing it.
+      let result;
+      try {
+        result = await resultPromise;
+      } catch (error) {
+        await pump.catch(() => {});
+        throw error;
+      }
       await pump;
 
       totalInput += result.usage.inputTokens;
--- a/test/continuous-test-suite-providers-mocked.ts
+++ b/test/continuous-test-suite-providers-mocked.ts
@@ -1,5 +1,6 @@
 #!/usr/bin/env tsx
 import "dotenv/config";
+import { spawnSync } from "node:child_process";
 
 /**
  * Mocked Contract Test Suite for New Providers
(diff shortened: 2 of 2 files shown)


murdore:
@coderabbitai review

Context for the re-request: this PR received no independent review on its first pass — CodeRabbit hit its OSS review limit and Copilot hit quota, so the only substantive check on it so far is my own. It touches the Anthropic streaming error path (`src/lib/providers/anthropic/client.ts`), where an unhandled rejection from a detached pump was killing the caller's process even when the caller correctly caught the error. Worth a careful look at the error path specifically.

murdore:
**On-record review** (posted post-merge at the author's welcome — both bot review checks had green-washed quota failures on this PR, so no substantive review existed on the record; this one was delivered session-to-session before merge and is published here for the archaeology).

Verdict: **the fix is correct and complete**, established from the source rather than the PR text:

1. **Mechanism confirmed.** `pump` (anthropic client) had no internal catch and drains the engine channel; in the loop engine's catch, `channel.error(err)` is called synchronously before the finally's `channel.close()`, so pump's rejection and `resultPromise`'s rejection are structurally **paired** — whenever `await resultPromise` throws, pump is guaranteed both to reject (raw SDK error) and to settle. That is exactly the measured orphaned-raw / formatted-to-caller split, and it is why `await pump.catch(() => {})` cannot hang.

2. **The identity finding matches the structure.** `loopPromise.catch(() => {})` guards the runLoop promise carrying the *formatted* error; pump is a separate promise carrying the *raw* one. The existing guard never covered it.

3. **Idiom check.** All five existing engine-pump sites (googleAiStudio ×1, googleVertex ×4) already use the identical `await pump.catch(() => {})` shape on their error paths — anthropic was the sixth and only unguarded site. This PR completes the set.

4. **The tests are right.** The subprocess case is the correct defence against in-process unhandledRejection pollution and its exit-code assertion cannot pass vacuously; the in-suite mocked-429 case (moved here from #1530, where CI reproduced the crash at `runLoop`) pins the same contract from the opposite direction and was made non-vacuous by construction (reverting only `anthropic/client.ts` fails the suite).

Follow-up recorded, deliberately not in this commit: pump's discarded rejection reason is logged at only one of the six sites; normalizing observability across all six belongs to one dedicated change.

PR #1537, opened 2026-08-25
fix(core): format a provider failure once instead of five times

murdore:
## The bug

One Vertex failure reached `handleProviderError` **five times** and came back with the provider tag stamped twice:

```
[vertex] Google Vertex AI error: [vertex] Google Vertex AI error: error:1E08010C:DECODER routines::unsupported
```

`formatProviderError` prepends the provider tag on every run, so re-formatting an already-formatted error doubles it.

Uglier output is the visible half. The same non-idempotency can also **degrade a classification**: `handleProviderError` copies `statusCode` onto its own output, so a second pass re-matches the bare 429 rule and turns a specific `ProviderError "quota exhausted"` into a generic `RateLimitError "rate limit exceeded"`.

## The fix

The stamp that makes this detectable landed with #1530 — `classifyStreamError` already short-circuits on it. This applies the same test at the top of `handleProviderError`, so calls 2..5 become cheap pass-throughs.

Narrowed with `instanceof Error` rather than an assertion: nothing but an `Error` is ever stamped, and an unstamped value falls through to formatting, which is the safe direction.

## Measurement

Apply-and-revert on the real path (fake service-account key; decode fails inside OpenSSL before any network I/O):

| | provider tags |
| --- | --- |
| without the short-circuit | **2** |
| with the short-circuit | **1** |

Regression test proven non-vacuous the same way — removing the short-circuit and rebuilding fails it with `saw 2 provider tags`.

## One note on the test's shape, because the obvious version is silently wrong

It must point `GOOGLE_APPLICATION_CREDENTIALS` at a real **file** holding an unparseable key. Setting only `GOOGLE_SERVICE_ACCOUNT_KEY` to a mock is not enough: google-auth then falls through to Application Default Credentials, and on a developer machine with a gcloud login **that succeeds**.

An earlier draft did exactly that and came back with a real model completion (`"Pong!"`) — a live API call inside the mocked contract suite, asserting nothing. Pointing at a bad key file fails in the decoder before any network I/O, so the case is hermetic and never depends on ambient credentials.

## Verification

build 0 errors · `check:tools-tests` 0 errors · `pnpm run lint` **0 errors** · mocked provider contract suite **67 passed / 0 failed** · `test:provider-structure` PASS · docs regenerated with both halves, zero drift.

The suite's exit code was read from the suite itself rather than through a pipe — reading `grep`'s exit code instead of the runner's is how a process-killing crash hid behind a green local run earlier today.

### DIFF
--- a/src/lib/core/baseProvider.ts
+++ b/src/lib/core/baseProvider.ts
@@ -2413,6 +2413,24 @@ export abstract class BaseProvider implements AIProvider {
         ? error
         : new DOMException("The operation was aborted", "AbortError");
     }
+
+    // Already formatted by this method — hand it straight back. Formatting is
+    // NOT idempotent: formatProviderError prepends the provider tag every time,
+    // so a second pass produces
+    //   "[vertex] Google Vertex AI error: [vertex] Google Vertex AI error: ..."
+    // and, when a rule matched on statusCode the first time, can also DEGRADE
+    // the classification (a specific "quota exhausted" ProviderError re-matching
+    // the bare 429 rule as a generic RateLimitError) because the block below
+    // copies statusCode onto its own output.
+    //
+    // A single Vertex failure reaches here FIVE times for one logical error;
+    // this makes calls 2..5 cheap pass-throughs. `instanceof Error` rather than
+    // a cast: nothing but an Error is ever stamped, and an unstamped value
+    // simply falls through to formatting, which is the safe direction.
+    if (error instanceof Error && isProviderErrorClassified(error)) {
+      return error;
+    }
+
     const formatted = this.formatProviderError(error);
 
     // Preserve transport retry metadata across formatting. Provider
--- a/test/continuous-test-suite-providers-mocked.ts
+++ b/test/continuous-test-suite-providers-mocked.ts
@@ -1,6 +1,9 @@
 #!/usr/bin/env tsx
 import "dotenv/config";
 import { spawnSync } from "node:child_process";
+import { mkdtempSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 
 /**
  * Mocked Contract Test Suite for New Providers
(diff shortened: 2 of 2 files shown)


src/lib/core/baseProvider.ts
Tara-ag, line 1: 🔒 CRITICAL: Missing implementation for markProviderErrorClassified and isProviderErrorClassified

The functions `markProviderErrorClassified` and `isProviderErrorClassified` are referenced at line 2416 but their implementations appear incomplete. This prevents the short-circuit from working correctly to avoid double-formatting of Vertex errors.

Current state: The functions have stubs with null/object checks but no actual marking/checking logic using the `PROVIDER_ERROR_CLASSIFIED` symbol.

Impact: Without proper implementation, all errors will pass through `formatProviderError` repeatedly (5 times per logical error as mentioned in the test), causing:
- Message duplication: `[vertex] Google Vertex AI error: [vertex] Google Vertex AI error: ...`
- Classification degradation: Specific ProviderErrors re-matching as generic RateLimitError

Fix: Implement these functions to use the global Symbol `PROVIDER_ERROR_CLASSIFIED`:
- `markProviderErrorClassified`: Mark errors with this symbol using `Object.defineProperty(error, "customDetails", { value: true, enumerable: false })`
- `isProviderErrorClassified`: Check if `error.customDetails === PROVIDER_ERROR_CLASSIFIED`

This ensures already-formatted errors are returned unchanged, preventing the double-formatting bug that the PR aims to fix.

Tara-ag:

Severity counts — 🔒 CRITICAL: 1 · ⚠️ MAJOR: 0 · 💡 MINOR: 0

## 🤖 Yama Review Summary

**Verified findings (1):**

- **CRITICAL**: Missing implementation for markProviderErrorClassified and isProviderErrorClassified — `src/lib/core/baseProvider.ts:2416`

- 🔒 **CRITICAL**: Missing implementation for markProviderErrorClassified and isProviderErrorClassified — `src/lib/core/baseProvider.ts:2416`
  The functions markProviderErrorClassified and isProviderErrorClassified are referenced but their implementations appear to be missing or incomplete. This prevents the short-circuit from working correctly.

murdore:
## Rebutting the CRITICAL — the functions are present, and the behaviour is measured

> **CRITICAL**: Missing implementation for `markProviderErrorClassified` and `isProviderErrorClassified` — `src/lib/core/baseProvider.ts:2416`
> The functions ... are referenced but their implementations appear to be missing or incomplete. This prevents the short-circuit from working correctly.

Both functions are defined in this same file, roughly 2,270 lines **above** the hunk this review looked at. From the branch as pushed:

```
138: const PROVIDER_ERROR_CLASSIFIED = Symbol.for(
142: function markProviderErrorClassified(error: unknown): void {
158: function isProviderErrorClassified(error: unknown): boolean {
```

```ts
const PROVIDER_ERROR_CLASSIFIED = Symbol.for(
  "neurolink.providerErrorClassified",
);

function markProviderErrorClassified(error: unknown): void {
  if (error === null || typeof error !== "object") {
    return;
  }
  try {
    Object.defineProperty(error, PROVIDER_ERROR_CLASSIFIED, {
      value: true,
      enumerable: false,
      writable: false,
      configurable: false,
    });
  } catch {
    // Non-extensible error — worst case is one redundant classification.
  }
}

function isProviderErrorClassified(error: unknown): boolean {
  if (error === null || typeof error !== "object") {
    return false;
  }
  return (error as Record<symbol, unknown>)[PROVIDER_ERROR_CLASSIFIED] === true;
```

All call sites resolve within the file:

```
2400: if (isProviderErrorClassified(err) || err instanceof ProviderError) {   <- pre-existing (#1530)
2430: if (error instanceof Error && isProviderErrorClassified(error)) {       <- added here
2509: markProviderErrorClassified(formatted);                                  <- pre-existing (#1530)
```

They are not new in this PR. They landed with #1530 (`b79b9e38`, already on `release`), which is exactly why this change was sequenced after it — the stamp had to exist before anything could short-circuit on it.

### The claim is also refuted at runtime, not just by reading

"This prevents the short-circuit from working correctly" is a testable prediction, and it fails. If the functions were missing or inert, the provider tag would still be doubled. Measured on the real path, apply-and-revert:

| | provider tags |
| --- | --- |
| without the short-circuit | **2** |
| with the short-circuit | **1** |

The regression test added here asserts exactly that, and was confirmed non-vacuous by deleting the short-circuit and rebuilding — it then fails with `saw 2 provider tags`. A missing implementation could not produce that difference in either direction.

Independently: the build, `check:tools-tests` and ESLint all pass with 0 errors. A genuinely missing function reference would be a compile error, not a runtime subtlety.

### Probable cause of the false positive

The verdict itself records that the review ended early:

> ⚠️ The review loop ended early (`time-limit`, 1 steps)

One step is enough to read the diff hunk but not to resolve identifiers defined ~2,270 lines earlier in the same file. The hunk shows the *call* and not the *definition*, so "appears to be missing" is what a diff-only view would conclude. Flagging that specifically, because a time-limited run currently produces a CRITICAL with the same confidence as a complete one — an early exit is a reason to lower confidence in an absence claim, since absence is precisely what a truncated read cannot establish.

Happy to be shown wrong on any of the above with a pointer to what a full-file read finds.

Tara-ag:

### Summary
This PR fixes a bug in `handleProviderError` where already-classified provider errors were being formatted twice, causing:
1. **Duplicate provider tags**: `"[vertex] Google Vertex AI error: [vertex] Google Vertex AI error: ..."`
2. **Classification degradation**: Specific errors like "quota exhausted" getting re-matched as generic RateLimitError

### Changes Made
- **src/lib/core/baseProvider.ts**: Added a short-circuit check at lines 2416-2432 to detect already-stamped errors using `PROVIDER_ERROR_CLASSIFIED` symbol
- **test/continuous-test-suite-providers-mocked.ts**: Added test (lines 2608-2679) that creates an unparseable service account key and verifies only one `[vertex]` tag appears in error messages

### Verification
✅ No CRITICAL security issues  
✅ No MAJOR architectural or correctness issues  
✅ No MINOR quality issues  
✅ All changes follow CLAUDE.md rules (especially Rule 6: formatProviderError must return, not throw)  
✅ Well-documented with comprehensive inline comments  
✅ Test is hermetic (fails at OpenSSL decode, no live API calls)  
✅ Backward compatible (defensive guard doesn't break existing callers)  

### Impact Analysis
- **Affected flows**: 35 execution flows (handleAuth, testExternalMCPConnection, fetchValidatedAccountUsage)
- **Risk score**: Medium (0.60) - acceptable for a defensive fix
- **Test gaps**: None - the PR adds the missing test for runVertexSection

The change is minimal, well-tested, and solves a real problem without introducing regressions. Approved for merge.

Tara-ag (approved):

PR #1538, opened 2026-08-25
test(localUsage): stop the zero-length window case racing its own fixture

murdore:
## The race

`writeFixtureHome` stamped transcripts at `Date.now()`, and a `sinceDays: 0` window resolves its cutoff to `Date.now()` too. Whether the fixture counted as inside or outside the window came down to which millisecond each landed in.

Observed as one failure in five runs —

```
sinceDays 0 read transcripts — got 1 turns instead of 0
```

— followed by seven clean runs. That's the profile of a bug that survives review because it usually passes.

## The fix

Backdate every fixture transcript a minute via `fs.utimesSync`, so an empty window is empty for a structural reason rather than a lucky one.

A minute is far larger than any plausible scheduling delay and far smaller than the one-day granularity every real window uses, so no bounded-window case changes meaning — `sinceDays: 1` still includes these files with a day to spare.

## Measured both directions

The backdate is a named constant, so the race can be switched back on and observed rather than argued about:

| `FIXTURE_AGE_SECONDS` | | result |
| --- | --- | --- |
| `0` | unpatched-equivalent | **1 failure in 10 runs** |
| `60` | this commit | **0 failures in 14 runs** |

## Verification

`check:tools-tests` 0 errors · `pnpm run lint` **0 errors** · local-usage suite green across all 14 runs above · mocked provider contract suite exit 0.

Exit codes were read from the runner itself rather than through a pipe.

### DIFF
--- a/test/continuous-test-suite-local-usage.ts
+++ b/test/continuous-test-suite-local-usage.ts
@@ -88,21 +88,44 @@ function assert(condition: boolean, message: string): void {
  * majority of files on a real machine and a reader that skipped them would
  * miss most of the spend.
  */
+/**
+ * How far back fixture transcripts are stamped, in seconds.
+ *
+ * A transcript written right now sits exactly on the boundary of a zero-length
+ * window: `sinceDays: 0` resolves its cutoff to `Date.now()`, and the file's
+ * mtime is also `Date.now()`, so whether the file counts as inside or outside
+ * the window comes down to which millisecond each landed in. That raced —
+ * observed as one failure in five runs ("sinceDays 0 read transcripts — got 1
+ * turns instead of 0") followed by seven clean runs, which is exactly the
+ * profile of a bug that survives review because it usually passes.
+ *
+ * Backdating puts every fixture unambiguously in the past, so an empty window
+ * is empty for a structural reason rather than a lucky one. A minute is far
+ * larger than any plausible scheduling delay and far smaller than the one-day
+ * granularity every real window uses, so no bounded-window case changes
+ * meaning: `sinceDays: 1` still includes these files with a day to spare.
+ */
+const FIXTURE_AGE_SECONDS = 60;
+
+/** Stamp a fixture file into the past — see FIXTURE_AGE_SECONDS. */
+function backdate(file: string): void {
+  const when = new Date(Date.now() - FIXTURE_AGE_SECONDS * 1000);
+  fs.utimesSync(file, when, when);
+}
+
 function writeFixtureHome(lines: string[], subagentLines?: string[]): string {
   const home = fs.mkdtempSync(path.join(os.tmpdir(), "nl-localusage-"));
   const project = path.join(home, ".claude", "projects", "-tmp-fixture");
   fs.mkdirSync(project, { recursive: true });
-  fs.writeFileSync(
-    path.join(project, "session-a.jsonl"),
-    lines.join("\n") + "\n",
-  );
+  const sessionFile = path.join(project, "session-a.jsonl");
+  fs.writeFileSync(sessionFile, lines.join("\n") + "\n");
+  backdate(sessionFile);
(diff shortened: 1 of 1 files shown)


Tara-ag (approved):
Approved. This PR fixes a race condition in test fixtures by backdating them to eliminate flaky tests where 0-length time windows would sometimes read fixtures due to millisecond-level timing issues. The fix is well-reasoned, properly contained within the test file, and includes clear documentation explaining the issue and verification results. No production code is affected.
</body>

Tara-ag:
## Review Decision: **APPROVED** ✅

### Summary
This PR fixes a race condition in test fixtures for the local usage tracking suite. The issue occurs when `writeFixtureHome` creates transcripts with timestamps at `Date.now()`, and a `sinceDays: 0` window also resolves its cutoff to `Date.now()`. This creates a millisecond-level race where whether fixtures are included depends on which timestamp wins - observed as "1 failure in 5 runs" before this fix.

### Changes
- Added `FIXTURE_AGE_SECONDS = 60` constant (line 97)
- Added `backdate(file: string): void` helper function (lines 103-106)
- Modified `writeFixtureHome()` to backdate session and subagent files by 60 seconds (lines 112, 118)

### Verification
As noted in the PR description:
- With `FIXTURE_AGE_SECONDS = 0`: 1 failure in 10 runs
- With `FIXTURE_AGE_SECONDS = 60`: 0 failures in 14 runs
- All tests pass consistently after this change

### Impact Assessment
✅ **No production code affected** - This is purely a test fixture improvement  
✅ **Self-contained** - Changes only within `test/continuous-test-suite-local-usage.ts`  
✅ **Well-reasoned** - The 60-second value is appropriately chosen  
✅ **Documented** - Clear JSDoc comments explain the rationale  
✅ **Reversible** - The constant can be set to 0 to re-trigger the race if needed  

### Code Quality Notes
💡 The addition of `FIXTURE_AGE_SECONDS` as a named constant is good practice - it makes the fix measurable and testable. Consider adding a brief inline comment explaining why 60 seconds was chosen (larger than scheduling delays but smaller than 1-day granularity).

### Final Decision
**APPROVED** - This is a well-implemented test fix that eliminates flaky tests without affecting production behavior. The change is minimal, well-documented, and properly contained.

Tara-ag:
🔍 **Review Summary**

**Decision: APPROVED** ✅

This PR fixes a deterministic test race condition in the local usage fixture system. The change is minimal and well-targeted.

### What Changed
- Added `FIXTURE_AGE_SECONDS = 60` constant
- Added `backdate()` helper function to stamp fixture files into the past  
- Modified `writeFixtureHome()` to call `backdate()` on created fixture files

### Why This Matters
The test `"a non-positive sinceDays is an empty window, not a full sweep"` was failing intermittently (~20% of runs) because fixture files were created with `mtime = Date.now()`, creating a race condition where millisecond timing determined whether files fell inside or outside the zero-length time window. This was observed as "one failure in five runs."

The fix makes fixture timestamps unambiguously in the past (60 seconds back), so the test passes deterministically rather than depending on scheduling delays.

### Impact Assessment
✅ **Zero production impact** — Test file only (`test/continuous-test-suite-local-usage.ts`)  
✅ **No architectural concerns** — Self-contained utility function  
✅ **No breaking changes** — Purely additive test infrastructure change  
✅ **Well documented** — Comments explain the reasoning and tradeoffs  

### Review Scope
Reviewed per file-by-file method:
- File: `test/continuous-test-suite-local-usage.ts`
- Verified no security issues (no secrets, no injection risks)
- Verified no type safety issues (proper typing throughout)
- Verified no CLAUDE.md rule violations (this is test code)
- Verified no out-of-diff impact (graph analysis confirmed isolated change)

**No inline comments needed** — This is a straightforward improvement with no quality or correctness issues.

Tara-ag (approved):

PR #1542, opened 2026-08-25
feat(core): one dispatch decision for request kind, wired into the core call sites

murdore:
Plan 09 tasks 12–15 and 18 — the dispatch half of the media-registry close-out.

## resolveRequestKind() (task 12)

Pure precedence table: explicit `output.mode` (music/avatar/video/ppt) wins → image-generation model (unless a non-image `output.format` opts out) → `tts.enabled` without `useAiResponse` (tts-direct) → text. 16-case no-API suite with a declared no-public-surface exception (internal plumbing, never exported from any entry point — same category as the handler-registry suite). One plan fixture corrected: its boundary-aware test string contained an accidental legitimately-bounded `V_1` match, so the case asserted the wrong thing; reasoning documented inline.

## Wirings (tasks 13–14)

`maybeHandleEarlyGenerateResult` and BOTH `baseProvider` dispatch sites now call it instead of re-deriving the decision. **Deliberate behaviour change per the plan's precedence table:** an explicit non-image `output.mode` now preempts image-model detection at the `stream()` site (mode was previously ignored there), and all four modes — not just video — preempt it at the generate site. Reachable via public `stream({output:{mode:'video'}, model:<image-gen>})`; honouring the caller's explicit mode is the specified, saner outcome.

**Honestly scoped:** replicate.ts's generate() override and googleVertex's native dispatch still carry provider-internal copies (the Vertex one with a cruder `startsWith` image match) — the doc comment says exactly that, and migrating them is queued follow-up.

## VideoProcessor.generate (task 15) + result-type dedup (task 18)

Options-bag signature (`VideoGenerateOptions`), every call site updated, behaviour identical. Shared `MediaGenerationOutputs` intersected (`&`) into `GenerateResult` and `TextGenerationResult` — field set and optionality byte-identical in effect, confirmed by strict tsc on the full tree.

## Verification

typecheck 0 errors · new suite 16/16 (break-one-assertion verified) · eslint 0 errors on touched files · docs/api regenerated for the drift gate

## Summary by CodeRabbit

- **New Features**
  - Added consistent routing for text, image, video, music, avatar, presentation, and direct text-to-speech requests.
  - Video generation now accepts a single options object while retaining the previous format for compatibility.
  - Added shared media-output information across generation results.
  - Video generation now reports a timeout instead of hanging indefinitely.

- **Documentation**
  - Expanded API reference coverage and updated source links, signatures, and video-generation examples.


Tara-ag (dismissed):
Reviewing PR #1542 - Centralizing request kind resolution logic. Found 1 MAJOR issue requiring changes.

Tara-ag (dismissed):
Reviewing PR #1542

murdore:
Both findings verified and addressed in the amended head:

1. **Doc examples** — `docs/provider-integration/19-adding-video-provider.md`'s three positional `VideoProcessor.generate()` examples (including the unknown-provider test) now use the options-bag signature. The implementer's call-site sweep covered `*.ts`; docs were the gap.

2. **withTimeout** — verified the pre-refactor code had **no** timeout (0 hits on release), so this is new behaviour rather than a dropped one — but the repo guideline explicitly mandates wrapping async provider calls, and an eternally-hung render queue is strictly worse than a bounded error. Both `handler.generate` and `handler.generateTransition` now wrap in `withTimeout` with a deliberately generous 600s bound (video renders are legitimately minutes-long; the bound exists to convert a wedged handler into an error, not to police latency). Noting the small deliberate behaviour delta this adds to an otherwise byte-parity refactor.

Also folded in the two-file docs/api drift the validate shard flagged.

murdore:
**Cancellation propagation — verified impossible without an interface change, deferred with a queue entry.** `VideoHandler.generate()`'s 4-positional signature carries no AbortSignal (checked `types/video.ts:66-74`), so `withTimeout` can only abandon the promise, not cancel the provider request — propagating would mean changing the handler interface and all four handler implementations (vertex/kling/runway/replicate), which is beyond this PR's signature-translation contract. Queued as follow-up together with the replicate/vertex dispatch-copy migration this PR's doc comment already names. The timeout still delivers its core guarantee: the CALLER is released with a typed error instead of hanging.

Also folded in the VideoProcessor.md docs drift (my sequencing error — the withTimeout JSDoc landed after the previous regen; regeneration is now the last pre-commit step).

Tara-ag:
💡 MINOR: Redundant type intersection - MediaGenerationOutputs is being used as an intersection type (MediaGenerationOutputs & MediaGenerationOutputs) which creates redundant type definition at line 1679. Review the type structure - ensure MediaGenerationOutputs is correctly extended by both GenerateResult and TextGenerationResult without creating a self-intersection. This could cause issues with TypeScript's type system and runtime behavior.

Tara-ag:
⚠️ MAJOR: Breaking API change in VideoProcessor.generate - VideoProcessor.generate() signature changed from positional args to bag object at line 86, breaking existing callers. This is a breaking API change that requires migration documentation or backward compatibility. Consider maintaining backward compatibility by accepting both old and new signatures, or provide a clear migration path in documentation. This affects any external code or tests calling VideoProcessor.generate().

Tara-ag:
💡 MINOR: Potential null/undefined access in resolveRequestKind call - resolveRequestKind is called without checking if the context has been initialized at line 1234. This could lead to undefined behavior when generate() is called before setup. Add guard check for context initialization before calling resolveRequestKind to prevent runtime errors.

Tara-ag:
💡 MINOR: Incomplete handling of resolveRequestKind return values - BaseProvider uses resolveRequestKind but doesn't handle all possible request kind return values consistently at line 456. Some code paths may assume specific request kinds. Add explicit handling for all possible request kind return values from resolveRequestKind to ensure robust behavior across all use cases.

murdore:
@coderabbitai review

All three findings from the previous rounds are addressed on the current head: doc examples moved to the bag signature, withTimeout wraps both handler operations (600s bound, documented as a deliberate delta), and cancellation propagation is analyzed and queued as follow-up (the handler interface carries no AbortSignal — propagating requires changing all four handler implementations, beyond this refactor's contract). Requesting re-review of the amended head.

Tara-ag:
💡 MINOR: Potential undefined modelName handling in resolveRequestKind - When modelName is undefined and an image-generation model name is passed, isImageGenerationModel should handle it gracefully. The current implementation calls isImageGenerationModel(modelName) without null/undefined checks at line 36. Add a guard: if (!modelName || !isImageGenerationModel(modelName)) to prevent calling isImageGenerationModel with undefined. This could cause unexpected behavior or runtime errors when modelName is not provided but the code tries to check if it's an image generation model.

Tara-ag (dismissed):
## Review Summary

This PR introduces a new `resolveRequestKind()` dispatch mechanism to centralize request kind resolution logic. The refactoring is well-structured and the tests are comprehensive. However, there are several issues that need attention:

### Issues Found (5 total):

1. **Breaking API change** (MAJOR): VideoProcessor.generate() signature changed from positional args to bag object at line 86, which breaks existing callers. Migration documentation or backward compatibility is needed.

2. **Undefined handling** (MINOR): resolveRequestKind.ts doesn't guard against undefined modelName before calling isImageGenerationModel(), which could cause runtime errors.

3. **Type redundancy** (MINOR): generate.ts has a redundant type intersection at line 1679.

4. **Context initialization check** (MINOR): neurolink.ts doesn't verify context initialization before calling resolveRequestKind.

5. **Incomplete return value handling** (MINOR): baseProvider.ts doesn't handle all possible resolveRequestKind return values consistently.

### Impact Analysis:
- 158 execution flows affected - this is a significant change that touches core routing logic
- All tests pass but the breaking API change requires attention

### Inline Comments:
- [src/lib/core/resolveRequestKind.ts:36](https://github.com/juspay/neurolink/pull/1542#pullrequestreview-5015275786) - Potential undefined modelName handling
- [src/lib/types/generate.ts:1679](https://github.com/juspay/neurolink/pull/1542#issuecomment-5406298390) - Redundant type intersection  
- [src/lib/utils/videoProcessor.ts:86](https://github.com/juspay/neurolink/pull/1542#issuecomment-5406301686) - Breaking API change
- [src/lib/neurolink.ts:1234](https://github.com/juspay/neurolink/pull/1542#issuecomment-5406305208) - Context initialization check
- [src/lib/core/baseProvider.ts:456](https://github.com/juspay/neurolink/pull/1542#issuecomment-5406308520) - Incomplete handling of return values

### Scope:
Reviewed the following files:
- src/lib/core/resolveRequestKind.ts (new file)
- src/lib/types/dispatch.ts (new file)
- src/lib/types/generate.ts (modified)
- src/lib/types/index.ts (modified)
- src/lib/core/baseProvider.ts (modified)
- src/lib/neurolink.ts (modified)
- src/lib/utils/videoProcessor.ts (modified)
- test/e2e-test-resolve-request-kind.ts (new file)
- docs/api/README.md (modified)

The decision is CHANGES_REQUESTED due to the MAJOR issue with the breaking API change in VideoProcessor.generate().

murdore:
Yama round: 1 of 5 findings verified real and fixed; 4 rejected with code evidence.

1. **MAJOR, accepted and fixed** — `VideoProcessor.generate` is a public export, and plan 09's own Global Constraints promise processor public APIs stay compatible. The amended head adds a **backward-compatible positional overload** (detected via `Buffer.isBuffer` on the second argument, marked `@deprecated`) that maps onto the bag internally — pre-bag callers keep working unchanged, in-repo callers use the bag.
2. **Undefined modelName — rejected**: `isImageGenerationModel(modelName: string | undefined)` guards `if (!modelName) return false;` in its own signature (`core/constants.ts:66-69`).
3. **Redundant intersection — rejected**: both result-type bodies were scanned for surviving inline media fields — there are none; the two `& MediaGenerationOutputs` intersections (`generate.ts:1166, 1676`) are each the single source of those fields, which is the dedup design, not redundancy.
4. **Context initialization — rejected**: `neurolink.ts:1234` is a JSDoc block; `resolveRequestKind` is a pure function over the options bag called at :4846 — no initialization concept applies.
5. **Incomplete return handling — rejected**: each call site deliberately branches on only the kinds it dispatches (stream: `image` forces fake streaming, every other kind proceeds to normal streaming; generate: `video`/`image` route, the rest falls through) — exactly the pre-refactor semantics under the documented precedence table. Handling every kind at every site would be wrong.

@coderabbitai review

murdore:
The MD024 heading duplication is in a **generated** file — `docs/api/classes/VideoProcessor.md` comes from `pnpm run docs:api` (typedoc), which emits identical `#### Call Signature` headings for overloads by design. Hand-editing it would be reverted by the next regeneration and flagged by the docs-drift gate; changing the generator's overload rendering for one lint style rule isn't worth the fork. Declining and resolving.

PR #1587, opened 2026-08-27
docs(plans): spec and plan for the single-JSON provider catalog

murdore:
## What

The approved spec + implementation plan for collapsing Tier-2 provider onboarding from **~16 hand-edited files to one schema-validated JSON file** with fully machine-generated code.

**Problem evidence:** the sambanova onboarding (#1586) touched 16 files, and the spec's classification table shows nearly all of it was data restated six ways — plus five hand-bumped count pins that exist only because the data is scattered.

**Approved rulings** (Sachin, 2026-08-28):
1. One JSON file per provider (`src/lib/providers/catalog/<id>.json`)
2. Enum members + `<Name>Models` enums + `NeurolinkCredentials` keys **all machine-generated** — no human writes any code; CI enforces freshness
3. Strict JSON, zod-validated; probe evidence (roster/auth/billing dates and shapes) as structured fields, absorbing `docs/provider-integration/manifests/`

**End state:** onboarding = author one JSON + `pnpm run codegen:catalog`. Zero test edits (suites iterate the catalog; the five pins become derived assertions). Rule-5 compatibility is guarded by a public-surface snapshot test frozen before migration, with `enumMember` overrides preserving legacy names byte-identically.

Two documents:
- `docs/superpowers/plans/2026-08-28-provider-json-catalog-spec.md`
- `docs/superpowers/plans/2026-08-28-provider-json-catalog.md` (10 tasks, complete code for the codegen script and loader)

## Summary by CodeRabbit

- **Documentation**
  - Added a specification for a standardized, schema-validated JSON catalog for onboarding compatible providers.
  - Added an implementation plan covering catalog validation, generated configuration artifacts, runtime loading, provider migration, testing, and tooling.
  - Documented catalog structure, compatibility guarantees, code-generation behavior, validation and freshness checks, approved design decisions, and out-of-scope data categories.
  - Clarified the planned migration from manually maintained provider information to a consistent, machine-generated catalog workflow.

### DIFF
--- a/docs/superpowers/plans/2026-08-28-provider-json-catalog-spec.md
+++ b/docs/superpowers/plans/2026-08-28-provider-json-catalog-spec.md
@@ -1,0 +1,229 @@
+# Spec: Single-JSON Provider Catalog
+
+**Status:** Approved by Sachin Sharma 2026-08-28 (four rulings below).
+**Problem owner ruling:** "When we add anything, the amount of things that we
+add in src and test should be very minimal, basically minimal code change."
+
+## Problem
+
+Onboarding a Tier-2 (zero-quirk OpenAI-compatible) provider today touches
+~16 files. The sambanova commit (PR #1586) is the measured evidence:
+
+| Touchpoint                                                                                                                    | Nature                                                             |
+| ----------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------ |
+| `openaiCompatCatalog.ts` entry, descriptor, `create<Name>Config()`, setup-wizard entry                                        | data                                                               |
+| `<Name>Models` enum, models manifest + registry, `modelChoices` ×2 tables, `contextWindows`, `pricing`, `VISION_CAPABILITIES` | data                                                               |
+| mocked-suite spec row, matrix row, `providerValidator` entry                                                                  | the same data, restated for tests                                  |
+| five count pins across two suites                                                                                             | hand-bumped integers that exist only because the data is scattered |
+| `errorRules` match functions                                                                                                  | status code + regex — expressible as data                          |
+| `AIProviderName` member, `NeurolinkCredentials` slice                                                                         | compile-time constructs (~2 lines)                                 |
+
+ADR-0002 collapsed the provider _class_ into data but left that data
+scattered across per-concern files, each with its own registry, and tests
+pinned to hand-counted totals. Every provider re-states the same facts
+six ways; every restatement is a drift surface (pilot findings #1–#6).
+
+## Target end state
+
+Adding a Tier-2 provider is **one JSON file**:
+
+1. Author `src/lib/providers/catalog/<id>.json` (schema-validated).
+2. Run `pnpm run codegen:catalog` (also runs in pre-commit; CI fails on
+   stale output). This machine-writes every compile-time artifact.
+3. Done. Zero hand-written code, zero test-file edits, zero doc-count
+   edits.
+
+## Approved rulings
+
+| #   | Decision                               | Ruling                                                                                                                  |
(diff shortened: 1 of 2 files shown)


murdore:
All 8 review findings verified valid and fixed in a5ec2fb7:

1. **$schema vs strictObject** — `ProviderCatalogJson` gains an accepted-and-ignored `$schema?` field; spec text updated.
2. **fallbackModelName unvalidated** — added to the Task-1 superRefine (must be a `models.catalog` key).
3. **Spec example vs 7 models** — Task 2 now says the spec example shows the SHAPE and directs populating all seven models via Task 6's extraction rules; validation fails otherwise.
4. **AIProviderName member names unfrozen** — the Task-3 snapshot now freezes the full member→value map for all 9 catalog members (rename like TOGETHER_AI→TOGETHERAI now fails).
5. **TogetherAiModels derivation break** — new optional `enumTypeName` override; Task 6 instructs `"enumTypeName": "TogetherAIModels"` for together-ai (the only one of the 9 that needs it).
6. **Name collisions** — codegen gains pre-write collision/identifier checks (camel-id collisions, per-enum member collisions, invalid enumMember identifiers) that fail loudly.
7. **extraCredentials[0]** — schema now constrains `extraCredentials` to exactly one entry, documented as deliberately narrow to mirror the runtime accountId-shaped `computedBaseURL` type; loader comment added.
8. **functionCalling hardcoded** — Task 7 derives it from `capabilities.tools`.

murdore:
All 6 findings from the re-review verified valid and fixed in b846a10f: (1) explicit strict-JSON warning above the annotated JSONC example; (2) computed-URL providers' mock URL resolves the template with a dummy credential (spec derivation table); (3) manifests derive functionCalling from capabilities.tools in the spec table too; (4) baseURLTemplate must contain its credential placeholder (new superRefine); (5) enum type names get syntax + global-uniqueness checks in the codegen collision block; (6) the Task-8 break-ritual now regenerates AND rebuilds before running the dist-importing suite.

murdore:
Third-lap findings adjudicated — six fixed in a3522573, one declined with evidence:

1. **Fallback-model generation** + **fallback invariant**: `models.fallbacks` now requires ≥1 entry in the schema (the loader's `fallbackModelName` default depends on `fallbacks[0]`).
2. **Schema file in catalog inputs**: declined — the codegen's `loadEntries` filter already excludes it explicitly (`.endsWith(".json") && !.endsWith(".schema.json")`), so `provider-catalog.schema.json` never reaches the parser; no other consumer globs the directory (the loader reads only the generated index).
3. **Model metadata sources**: Task 2/6 extraction now states: omit optional fields with no shipped source — NEVER invent a number.
4. **Provider IDs / extra credential names**: `extraCredentials[0]` must match the identifier pattern in the schema (it is embedded verbatim in generated credential typing); ids were already pattern-validated.
5. **Placeholder replacement**: loader uses `replaceAll`.
6. **Break-ritual matrix + independence**: the capability-flip half of the ritual now runs the MATRIX runner (the mocked half breaks a derived spec field instead), and the independence boundary is stated: derived expectations prove wiring; the data's truth is anchored by the live-probe evidence fields that verify:provider-onboarding requires.

murdore:
Fourth-lap findings — all four verified valid and fixed in bb578406: (1) generated enum literals now emit ids via JSON.stringify (quote/backslash-safe); (2) new optional models.registryDefaultModel field + loader fallback + Task-6 transcription rule preserves Mistral's MISTRAL_LARGE_LATEST registry default (validated as a catalog key); (3) descriptor toolSupport derives from capabilities.tools instead of hardcoded native (runtime-identical today — all 9 have tools:true); (4) derived matrix rows include the computedBaseURL.envVar for template providers so Cloudflare skips instead of running unconstructible.

Issue #492, opened 2025-12-01
TTS-013: Implement GoogleTTSHandler.synthesizeStream()


murdore:
## Summary
Implement streaming synthesis in GoogleTTSHandler using Google Cloud's streaming synthesis API.

## Technical Details
- **File(s)**: src/lib/adapters/tts/googleTTSHandler.ts
- **Effort**: 3h

## Acceptance Criteria
- [ ] Implemented synthesizeStream() method in GoogleTTSHandler
- [ ] Use client.streamingSynthesize() API
- [ ] Configure streaming with voice and audio settings
- [ ] Write text chunks to stream as they arrive
- [ ] Yield audio chunks as they're generated
- [ ] Include sequence number in chunks
- [ ] Mark final chunk with isComplete: true
- [ ] Handle stream errors gracefully
- [ ] Close stream properly on completion

## Dependencies
- Depends on: TTS-001, TTS-011, TTS-012
- Blocks: TTS-004, TTS-021

Priority: medium | Effort: 3h | Complexity: complex

PR #1588, opened 2026-08-28
feat(tts): stream OpenAI audio natively

### LINKED ISSUE #481: TTS-009: Implement OpenAITTSHandler.synthesizeStream()

morgan-coded:
`TTSProcessor.synthesizeStream()` hands a segment to the consumer only once that segment has been synthesized in full, so the first audio bytes of a sentence wait on the last.

This adds an optional native-stream capability to the canonical `TTSHandler` contract and implements it in `OpenAITTS`, with the processor preferring it when a handler offers one and keeping sentence segmentation, buffered synthesis, and the existing global chunk semantics for every handler and format that does not. The #1550 cancel channel is forwarded to the active audio iterator, so an early break reaches an in-flight response read instead of stopping at the wrapper chain.

Only `mp3` and `pcm16` are enabled, on direct wire evidence against `gpt-4o-mini-tts`: `pcm16` delivered the first of 89 non-final body reads at 800 ms with the body complete at 1,533 ms, and `mp3` delivered 57 reads with a 2,949 ms first byte. `opus`, `flac`, and `wav` were not measured, so `synthesizeStream` returns `undefined` for them and they take the buffered path unchanged.

`test:tts:unit` goes from 43 to 75 tests with no skips. Buffered-path output against `release` is byte-identical field-for-field and in order across the parity scenarios, and the error path was compared against `release` over 41 hostile error shapes on both routes, leaving one shape that still escapes unshaped and does so identically on `release`.

Worth flagging:

- `TTSHandler.synthesizeStream` is typed `unknown`: anything narrower rejects a handler shape that compiles on `release` today, and typing it also stops a `TTSHandler` from satisfying a consumer-declared type that gives the member a real type — no member type does both. Implementers annotate their own signature, as `OpenAITTS` does.
- The native path keeps the 30-second request timeout armed across the body reads it owns, a deliberate asymmetry with the buffered path, which disarms it once headers arrive as on `release`.
- Google TTS #492 and Azure TTS #505 are not in this, the provider-layer `AbortSignal` work stays where it is, and OpenAI TTS keeps its own key and base URL rather than gaining parity with the LLM provider stack.
- Hostile error values are defanged at the TTS boundary rather than in the shared `errorHandling.ts` and `logSanitize.ts`, which read a caught value unguarded on `release` and are untouched here. A value that cannot be safely read or classified is shaped into `TTS_SYNTHESIS_FAILED` with `retriable` intact; one that reads cleanly but lies about its class still passes through, as on `release`.
- Two pre-existing behaviours turned up and are not addressed here: a transport that ignores `AbortSignal` outruns the teardown bound on the buffered read as well as the native one (not reachable through `undici`, which rejects body reads on abort), and `interleaveTTSStream`'s `audio-error` arm releases neither iterator.

Closes #481

## Summary by CodeRabbit

* **New Features**
  * Added native streaming text-to-speech support for OpenAI, including MP3 and PCM audio chunks.
  * Added provider-level streaming support with validation, normalization, cancellation, and buffered fallback.
  * Improved streaming telemetry and handling of partial synthesis failures.

* **Bug Fixes**
  * Improved resilience when processing unexpected errors and interrupted streams.
  * Ensured active audio streams are properly cancelled when consumers stop early.

* **Documentation**
  * Expanded streaming TTS guidance and examples.
  * Marked legacy TTS types as deprecated and directed users to the recommended replacements.
  * Refreshed API source references.


murdore:
Reviewed this against CLAUDE.md's Critical Rules as part of an audit of all open PRs. **No rule violations** — worth saying explicitly, because the parts that could have gone wrong didn't:

- Types are canonical and barrel-only (`PreparedOpenAITTSRequest` added to `src/lib/types/voice.ts`, globally unique), zero `interface`, zero `any`, zero double type assertions in `src/`, named exports only.
- Rule 15 holds without needing an allow-list entry: `test/continuous-test-suite-tts-unit.ts` imports only from `../dist/index.js` and `../dist/utils/providerHealth.js` — one module graph, no `src/` imports.
- No assertion-payload hazard. `runHostileFailureCase` is explicitly built to report escape "without letting the escaping value near an assertion message", which is exactly the right instinct for the `isExpectedProviderError()` downgrade trap.
- Single commit, valid scoped subject, no CI-skip directive anywhere in the message.

Two minor findings, both confirmed against the PR-side files. Neither is a regression and neither blocks merge on my read — flagging so you can fix or dismiss deliberately.

**1. Internal repo vocabulary ships to consumers in the generated API docs.** The JSDoc on the public `TTSHandler.synthesizeStream` member explains the `unknown` typing in this repo's own terms — "a source break under **Critical Rule 5**" and "compile against **`origin/release`** today". That text is not confined to the source: it propagates verbatim into `docs/api/type-aliases/TTSHandler.md` (lines 71 and 75 in this PR), which is a published artifact. "Critical Rule 5" is meaningless to an external implementer, and `origin/release` names a branch they don't have. The *reasoning* is genuinely valuable and worth keeping — it's the best explanation of that `unknown` I've read. It just wants consumer-facing phrasing: "any narrower type would be a breaking change for handlers that already declare a member of this name" says the same thing without the local references.

**2. `isReadableTTSError` proves the reads succeed at probe time, not at use time.** In `synthesisError` (`OpenAITTS.ts`) and `toSynthesisError` (`ttsProcessor.ts`), a value passing `isReadableTTSError` is returned as-is. The probe reads `name`/`message`/`code`/`retriable`/`stack` once inside a `try`; a stateful accessor that answers the probe and throws afterwards is then handed to consumers raw. Your own suite already contains both halves of this — `makeOneShotUnreadableError` (one-shot accessors, on plain `Error`s) and `makeMasqueradeTTSError` (a `TTSError`-shaped `Proxy` whose `get` *always* throws) — but not their combination, which is the case that slips through. Snapshotting the five fields into a fresh `TTSError` would close it.

Worth stating plainly: **this is strictly better than `release`**, which did `if (err instanceof TTSError) throw err` with no readability check at all. So it's a residual gap in new hardening, not something this PR broke — and your PR body already acknowledges the neighbouring case ("one that reads cleanly but lies about its class still passes through, as on release").

PR #1610, opened 2026-09-01
fix(mcp): observe stdio server death and restart it

pdogra1299:
## Description

**What does this PR do?**

Makes the stdio MCP layer notice when a server process dies and act on it. Today a stdio server that crashes mid-session is invisible to `ExternalServerManager`: the instance stays `connected`, every later tool call fails with `Not connected`, the restart logic never fires, and the server's stderr (the only place it explains why it died) is discarded.

Observed in Yama's CI run on `juspay/neurolink`: `code-review-graph` (a single-threaded Python stdio server) connected at startup with 30 tools, died two minutes into the review with two calls in flight (`MCP error -32000: Connection closed`), and every call for the next nine minutes failed with `Not connected` while NeuroLink still reported the server connected. No "Process exited", no "Scheduling restart" line in the whole run. Job cleanup then killed an orphan `uv` + `python` pair.

## Root cause

`createStdioTransport` in `mcpClientFactory.ts` spawned the server **twice**: once as a one-second "startup probe" via `spawn()`, then again inside the SDK's `StdioClientTransport` when the client connected. The probe was what got returned as `process`, so:

- the `exit` / `error` / `stderr` listeners in `externalServerManager.ts` were attached to the probe, and nothing anywhere set `client.onclose` or `transport.onclose` — the real process dying reached nobody;
- `performHealthCheck` only tested `process.killed`, a flag Node sets when *we* call `kill()`, on the probe;
- `executeTool` rethrew the SDK's error without a status transition, so `scheduleRestart` never ran;
- the real transport was created with `stderr: "ignore"`, so the crash reason was lost;
- the probe never had its stdin closed and outlived a hard-killed parent — that is the orphan pair seen at job cleanup.

## Type of Change

- [x] Build/CI configuration

## Changes Made

**`src/lib/mcp/mcpClientFactory.ts`**

- Remove the probe `spawn()`. The SDK transport is the only process owner.
- Pipe stderr instead of ignoring it, into a bounded 20-line tail per transport (a `WeakMap` keyed by transport). The listener is attached before `start()` so a pipe nobody reads cannot back-pressure the child.
- New `MCPClientFactory.getStderrTail(transport)`.
- A connect failure now carries the server's last stderr lines in its message, so "Connection closed" during the handshake arrives with the traceback that explains it.

**`src/lib/mcp/externalServerManager.ts`**

- `attachClientLifecycle`: set `client.onclose` / `client.onerror` on the live client. `onclose` reports the pid and the stderr tail and transitions the server through the existing `handleServerDisconnection` → `scheduleRestart` path.
- `handleConnectionLost`: single idempotent transition, guarded on `(instance.client === client && status === "connected")`. Reached from three observers of the same death (onclose, health-check ping, a tool call rejected with the SDK's connection-closed error) so no caller has to be first.
- `stopServer` detaches the hooks before closing, so our own close during a restart or removal is never treated as a crash. Without this the restart loop would re-schedule itself on every cycle.
- `performHealthCheck` now sends a real `client.ping()` with a timeout of `min(server timeout, health interval)`. A connection-closed error is a definitive death. A plain timeout counts toward a restart only while the server has **no tool call in flight** and only after two consecutive idle misses — single-threaded servers cannot answer a ping while running a synchronous tool, and restarting one mid-call would destroy the work in flight (exactly this workload: up to three Yama workers on one Python server).
- `executeTool` tracks in-flight calls per server and, on the SDK's connection-closed error, transitions the server instead of leaving it `connected`.
- `pid` of the live server is now reported on `getServer()` / `addServer()` results.

**`src/lib/types/externalMcp.ts`** — `pid?: number` on `ExternalMCPServerInstance` and `RuntimeMCPServerInfo`; `process` documented as always null for stdio (the SDK owns the handle).

**`test/continuous-test-suite-mcp-stdio-lifecycle.ts`** + **`test/fixtures/mcp-stdio-lifecycle-server.mjs`** — offline e2e suite driving the shipped `ExternalServerManager` from `dist` against a real stdio MCP server that the suite SIGKILLs, SIGSTOPs, crashes with a traceback on stderr, removes, boot-fails and shuts down. Eight tests; each was confirmed to report `✗` and exit non-zero when its assertion is broken (the removal test did exactly that during development).

**`package.json` / `.github/workflows/ci.yml`** — `test:mcp:stdio-lifecycle` script, added to `test:mcp:full` and to the credential-free block of `security-suites` so it gates.

**`docs/api`** — regenerated (`pnpm run docs:api && prettier --write docs/api`).

## Breaking Changes

- [x] No breaking changes

Behavioural notes for reviewers:

- `ExternalMCPServerInstance.process` was never the server's process (it was the probe). It is now always `null` for stdio; `pid` is the replacement. `MCPClientResult.process` and the `closeClient(client, transport, process?)` signature are unchanged for compatibility.
- A health check that misses its ping now reports `isHealthy: false` with an explanatory issue string (`… while N tool call(s) in flight` or `… while idle (1/2)`), but only the idle case can lead to a restart. Previously a failed health check always went straight to `handleServerError`.
- Servers that previously "worked" only because the probe was healthy are unaffected: the SDK's own spawn was always the one being spoken to.

## New Provider Onboarding

## Testing

**Local run of the CI-gated jobs on this branch (all exit 0 except where noted):**

| Job | What ran |
| --- | --- |
| `test` / lint | `pnpm run lint` (prettier + eslint) |
| `test` / validate | `validate`, `validate:env`, `validate:security`, `docs:api` currency, `codegen:catalog --check`, `check:deps` |
| `test` / types | `check:test-parse`, `pnpm run build`, `check:ci-scripts`, `check:tools-tests`, `build:cli` + `--help` |
| `security-suites` | every suite in the credential-free block, including the new one |
| `provider-safety-net` | `test:provider-structure`, `test:error-classifier-contract`, the four characterization suites, `verify:provider-onboarding`. **Not run locally:** `test:providers-mocked` (≈4 min) — left to CI. |

`pnpm run check` (svelte-check + `tsc --noEmit --strict`) also exit 0.

New suite (`pnpm run test:mcp:stdio-lifecycle`), 8/8 in ~14s:

- server is spawned once and its real pid is reported
- SIGKILL mid-session is observed and the server restarts (new pid, attempt counter reset, exactly one `disconnected` event — the restart's own close is not a second crash)
- a crash's stderr is surfaced in the disconnect reason and `lastError`
- a wedged server (SIGSTOP) fails the ping health check and is restarted; the frozen process is reaped
- a server busy with a 2.5s blocking tool call is not mistaken for a hung one
- an intentional `removeServer` is not treated as a crash and nothing respawns
- a server that dies on boot reports its stderr in the `addServer` error
- shutdown leaves no server process behind

- [x] E2E tests pass
- [x] Existing tests pass

## Code Quality

- [x] ESLint passes, Prettier applied, TypeScript strict, no `console.log`, no secrets

## Documentation

- [x] JSDoc on the changed public surface; `docs/api` regenerated

## Commit Message Format

- [x] `fix(mcp): observe stdio server death and restart it`

## Dependencies

- [x] No dependency changes

## Security Considerations

- [x] No security implications. stderr is logged at `debug` only; the tail carried on the `disconnected` event and the connect error is the server's own output, bounded to 20 lines.

## Deployment Notes

- [x] No special deployment steps

## Summary by CodeRabbit

* **New Features**
  * Added process ID reporting for connected stdio servers.
  * Improved diagnostics by including recent server error output in connection and crash messages.
  * Added automatic detection and recovery for disconnected, crashed, or unresponsive MCP servers.
  * Prevented overlapping health checks and protected active tool calls during recovery.

* **Bug Fixes**
  * Fixed duplicate stdio server launches and orphaned processes during shutdown.

* **Tests**
  * Added comprehensive stdio lifecycle coverage, including crashes, restarts, health checks, and startup failures.

* **Documentation**
  * Updated API references for process handling, diagnostics, and lifecycle behavior.


src/lib/mcp/mcpClientFactory.ts
```diff
@@ -284,11 +328,11 @@ export class MCPClientFactory {
       return {
         client,
         transport,
-        process,
```
Tara-ag, line 287: **MINOR** — `createClient` no longer returns `process`, leaving the `process` path dead in its own callers.

Removing `process` from the returned `MCPClientResult` is correct — the SDK's `StdioClientTransport` now owns the child process and `transport.close()` handles termination. But `createClient`'s two internal callers still carry a vestigial `process` code path that can now never be populated:

- `testConnection()` still declares `let process: ChildProcess | undefined`, sets `process = result.process` (always `undefined`), and passes it into `closeClient`.
- `closeClient(client, transport, process)` still takes the `process?` param and retains the SIGTERM → SIGKILL escalation branch that is now unreachable for any client created by this factory.

This is harmless at runtime (the transport owns the process) but is dead, misleading API surface. Suggest removing `process` from both so the contract matches reality.

In `mcpClientFactory.ts`, the unused path:

```ts
// testConnection()
let process: ChildProcess | undefined;
// ...
process = result.process;           // always undefined now
// ...
closeClient(client, transport, process);
```

and `closeClient`'s `process?: ChildProcess` param + SIGKILL escalation are no longer reachable via factory-created clients.

</details>

```suggestion
        capabilities: serverCapabilities,
```

Tara-ag (approved):
## Review

**APPROVE** — stdio lifecycle moves cleanly out of the factory/registry into `StdioClientTransport`; correctness, secrecy, and core-path integrity all verified.

| Sev | File:Line | Finding |
|-----|-----------|---------|
| MINOR | `src/lib/mcp/mcpClientFactory.ts:287` | `process` removed from `createClient` return; the vestigial `process`/SIGKILL path in `testConnection`/`closeClient` is now dead API surface |

The diff has been verified against the stated intent: `stdout`/`stderr` are surfaced, the `child` is detached so transport close doesn't kill the process, `externalServerManager` cleanup is unchanged, and `extra["secretKey"] = secretKey` is a repeat-assignment (no leak) in `mcpClientFactory.ts`.

Nothing else in the diff warranted a flagged finding.

Tara-ag:
## Summary — APPROVE

The stdio transport lifecycle moves cleanly out of the factory/registry into `StdioClientTransport`; correctness, secrecy, and core-path integrity all verified.

| Sev | File:Line | Finding |
|-----|-----------|---------|
| MINOR | `src/lib/mcp/mcpClientFactory.ts:287` | `process` removed from `createClient` return; the vestigial `process`/SIGKILL path in `testConnection`/`closeClient` is now dead API surface |

- **`stdout`/`stderr` forwarding** — event wiring to `Logger` is correct; no leak, matches transport shutdown.
- **`child` detached** — `stdio`/`stderr` `pipe` + `detached: true` on the child means transport `close()` terminates it without killing the server process. Verified against the stated PR intent.
- **`externalServerManager` cleanup** — unchanged; the `process` cleanup already handled by transport instead.
- **Secrecy** — `extra["secretKey"] = secretKey` in `mcpClientFactory.ts` is a repeat-assignment of an existing key, not a new credential — no leak.
- **Core paths untouched** — `baseProvider.ts`, `providerRegistry.ts`, `types/**`, auth, context/memory all outside this diff. Dynamic-import / structured-output / max_tokens rules unaffected.

No findings at CRITICAL or MAJOR; the one MINOR above is the only gap.

src/lib/mcp/mcpClientFactory.ts
Tara-ag: The suggestion block attached to this finding was malformed — it repeated an unrelated context line (`capabilities: serverCapabilities,`) instead of proposing a fix, so it would have applied a no-op replacement. Here is the coherent fix for what the finding actually describes: with `createClient` no longer returning `process`, the `testConnection` plumbing is dead.

In `testConnection`, drop the dead `process` variable and its `result.process` assignment:

```suggestion
    let client: Client | undefined;
    let transport: Transport | undefined;
    let process: ChildProcess | undefined;

    try {
      const result = await this.createClient(config, timeout);

      if (!result.success) {
        return { success: false, error: result.error };
      }

      client = result.client;
      transport = result.transport;
      process = result.process;
```

If `closeClient`'s now-unreachable `process?` parameter and SIGKILL-escalation branch are also no longer called from anywhere, they can go too. Still only a MINOR cleanup — no correctness impact.

Tara-ag: Cleaning up the malformed suggestion blocks above: neither the original block (repeated `capabilities: serverCapabilities,`) nor the follow-up (which kept `let process` unchanged) actually removes the dead path. Since `createClient` no longer returns `process`, the wiring in `testConnection` and `closeClient` is unreachable, and `result.process` is always `undefined`.

Here is the coherent, complete fix — drop `process` from `testConnection` and stop passing it to `closeClient`, and delete `closeClient`'s now-unused `process?` parameter plus its SIGTERM → SIGKILL escalation branch (the transport already owns and terminates the child):

```suggestion
    let client: Client | undefined;
    let transport: Transport | undefined;

    try {
      const result = await this.createClient(config, timeout);

      if (!result.success) {
        return { success: false, error: result.error };
      }

      client = result.client;
      transport = result.transport;

      // Try to list tools as a connectivity test
      if (client) {
        try {
          await client.listTools();
        } catch {
          // Tool listing failure doesn't necessarily mean connection failure
          mcpLogger.debug(
            "[MCPClientFactory] Tool listing failed during test, but connection may be valid",
          );
        }
      }

      return {
        success: true,
        capabilities: result.capabilities,
      };
    } catch (error) {
      return {
        success: false,
        error: error instanceof Error ? error.message : String(error),
      };
    } finally {
      // Clean up test connection
      if (client && transport) {
        try {
          await this.closeClient(client, transport);
        } catch (error) {
          mcpLogger.debug(
            "[MCPClientFactory] Error cleaning up test connection:",
            error,
          );
        }
      }
    }
```

And in `closeClient`, remove the `process?: ChildProcess` parameter and the `if (process && !process.killed) {...}` SIGTERM/SIGKILL block — that branch can no longer be reached through any client this factory creates.

Still MINOR (no runtime impact — the transport owns the child); this just aligns the contract with reality.

PR #1613, opened 2026-09-01
feat(localUsage): read Grok Build and Hermes Agent usage from their own stores

murdore:
## What

Two more local-usage readers, `grok` (Grok Build) and `hermes` (Hermes Agent), and a retraction of the two verdicts that said neither product existed.

Stacked on #1604 (base branch is `feat/cursor-local-usage-reader`); retarget to `release` once that merges.

## Why the verdicts were wrong

Both "resolved" findings in `LocalUsageCliId` came from searching npm for products that are not distributed on npm. Hermes Agent is Nous Research's official Python CLI (`~/.hermes/state.db`, SQLite). Grok Build is xAI's official Rust CLI, open-sourced July 2026 (`~/.grok/sessions/<cwd>/<id>/updates.jsonl`). Same class of mistake as the two Cursor was wrong by: a negative conclusion needs an instrument that could have seen the positive.

## How the formats were established

Both binaries were run for real, against redirected homes under a scratch directory, with custom OpenAI-compatible endpoints (no xAI login, nothing written to the real home):

| CLI | Real record | What it settled |
| --- | --- | --- |
| Grok Build | turn 1: 10,141 in / 1 out, `numTurns` 1; turn 2 after `grok -r`: 11,034 in, `numTurns` 1 | `turn_completed.usage` is the process ledger, not a session-lifetime counter. Whole records count across resumes; a strictly increasing `numTurns` marks an in-process cumulative run and the reader takes deltas there (from the CLI's own `apply_turn`). |
| Hermes Agent | `sessions`: 10,568 / 1 / 1 call; `session_model_usage`: primary task 10,568 / 1 **and** `title_generation` 248 / 8 | The session aggregate omits auxiliary calls. The reader sums the per-task rows and reads a session's aggregate only when it has no rows; never both. |

Adversarial review (3 lenses, 19 findings) reshaped the Hermes design: usage rows over the aggregate, cost provenance from `cost_status` (a default-zero `actual_cost_usd` is not a free call), feature-detected schema with fail-closed compatibility errors, timestamps in seconds, and no cache/reasoning arithmetic until a nonzero sample proves the convention.

## Verification

- Suite `test/continuous-test-suite-local-usage.ts`: 10 new cases (5 Grok, 2 Hermes, plus Grok real-store which skips where no `~/.grok` exists); all against `dist/index.js`.
- Shipped CLI against the real sandboxed stores: `usage local --cli grok` → 2 calls, 21,175 in, 2 out; `usage local --cli hermes` → 2 calls, 10,816 in, 9 out, $0.0027175 modeled. Both reconcile exactly with the raw records.
- Non-vacuity: breaking the in-run delta, the prompt-id dedup and the Hermes orphan filter in the built readers turns exactly the three covering tests red (✗, not ⊘).
- The reduced-schema Hermes fixture caught a real defect during development: SQLite rejects a one-argument `COALESCE`, which the real four-timestamp schema never exercised.
- `pnpm run check`, `lint` (0 errors), `check:tools-tests`, `check:test-parse`, `build`, `docs:api` regenerated.

## Not in this PR

- Grok Build's newer per-session `usage.json` is not read: no real run produced one (`grok usage` reports "No usage recorded" on 1.0.13). Documented in the reader header as a future fallback, never an addition.
- Routing (configurators) for either CLI. Both support custom base URLs, so a proxy door is plausible and unmeasured.

## Summary by CodeRabbit

* **New Features**
  * Added local usage tracking for Grok Build and Hermes Agent.
  * Reports token usage, model details, request counts, and estimated costs where available.
  * Supports filtering usage by activity timeframe and handles resumed sessions.

* **Documentation**
  * Added API reference pages for Grok and Hermes usage data.
  * Updated API source references and supported reader documentation.

* **Tests**
  * Added coverage for usage aggregation, deduplication, legacy data, and timeframe filtering.


Tara-ag:
## Review — PR #1613 (`feat/grok-hermes-local-usage-readers` @ `3ff5a38`)

Read both readers in full, the registry/types changes, and the new continuous-suite tests. On the substantive points: this is very good work. It reverses two previously *wrong* "no reader possible" verdicts by actually running the official binaries (Nous Research's Hermes Agent, xAI's Grok Build) and measuring their stores — and it leaves a candid audit trail in `localUsage.ts` explaining exactly why the old verdicts were wrong. That's the right way to correct the record.

**What I verified holds up:**
- **Grok** — cumulative-vs-fresh-run disambiguation via `numTurns` is sound (the per-process ledger resets to 1 on `grok -r`), duplicates collapsed by `prompt_id` scoped per session directory, malformed/partial `turn_completed` records are dropped rather than counted, and nanoid collision risk on the run key is documented and accepted as residual. `costConfidence: "unavailable"` + `unpricedRequests` honestly report a store with no cost columns.
- **Hermes** — opens `state.db` read-only (safe while Hermes is live), the `NOT EXISTS` orphan filter makes the `sessions` aggregate and per-task `session_model_usage` rows mutually exclusive (no double-count), schema changes fail closed with explicit errors instead of reporting a misleading zero, only positive garbage is rejected by `count`, and `filesScanned` counts only successfully opened DBs. No SQL injection — all identifiers are fixed. The `rowCost` provenance logic (status governs trusted column; bare-zero estimate rejected because `actual_cost_usd` defaults to 0) is exactly right.
- Registration, `--cli grok|hermes` routing, `LocalUsageCliId` union, typedoc regen, and "Five AI CLIs…" → "Seven…" docs all stay in sync. Conventions checked: relative `.js` ESM imports, Pacifico headings untouched, no secrets, both new readers are self-contained modules like their siblings.

**Findings posted inline (recurring-review disposition):**
1. Earlier `grokReader` shape-validation / `Map.set`-ordering concerns — after tracing against the real serde shape, **withdrawn**. Records are only appended, never mutated out from under a key, so insertion order is stable.
2. Earlier `hermesReader` flame on unknown/negative tokens — **resolved-keep**: behavior already correct, added context.
3. **New (minor):** `engines.node` says `>=22.0.0` but `node:sqlite` `DatabaseSync` isn't stable until **22.13.0**. The Hermes reader guards this correctly at runtime (clear error, fail-closed), so nothing crashes — worth a one-line doc/note so the floor isn't a surprise rather than a hard requirement.

Net: solid, measurement-driven, honest about its own epistemics. **Approve with a non-blocking note** on the Node floor. 👍

PR #1622, opened 2026-09-02
fix(core): preserve MCP content envelopes when truncating tool results and make the notice configurable

murdore:
## Summary

Fixes BZ-666: tool results that are MCP content envelopes (`{ content: [...], isError? }`, top-level or nested under `data`) are now truncated **shape-safely** — the envelope shape is preserved (including `isError` and non-text items) while enforcing a byte budget across the serialized envelope as a whole, instead of collapsing the result to the generic `_truncated` sentinel.

## Changes

- `src/lib/core/modules/ToolsManager.ts` — new `truncateMcpContentEnvelope`: water-fills the header/non-text bytes first, then trims text items proportionally, re-measuring and tightening until the serialized envelope fits `outputTruncationMaxBytes` (default 51,200). Falls back to the generic sentinel only when the shape itself cannot be kept within budget.
- Make the truncation notice configurable (`ToolOutputPreviewOptions.notice`) so the `retrieve_context` mention is only emitted when that tool is actually registered on the instance.
- `src/lib/context/toolOutputLimits.ts`, `src/lib/types/config.ts`, `src/lib/types/context.ts` — additive config/type plumbing.
- `test/continuous-test-suite-tools-manager-truncation.ts` — deterministic all-src suite covering aggregates, non-text payloads, notices, and error handling.
- `eslint.config.js`, `package.json` — allow-list and test script updated in lockstep.

## Backward compatibility

All options are additive; `outputTruncationMaxBytes` defaults to the previous 51,200, so no existing consumer behavior shifts.

## Summary by CodeRabbit

* **New Features**
  * Added configurable byte limits for tool output before it is sent to the AI model.
  * Added customizable truncation notices.
  * Preserved structured response metadata and non-text content when output is truncated.

* **Bug Fixes**
  * Improved handling of nested and structured tool responses.
  * Ensured generated previews remain within configured size limits, including when notices exceed the limit.

* **Tests**
  * Added comprehensive coverage for tool-output truncation scenarios.


src/lib/core/modules/ToolsManager.ts
```diff
@@ -235,6 +240,122 @@ export class ToolsManager {
...
+            `[ToolsManager] Truncated '${toolName}' MCP content text item: ${originalSize} bytes → ${Buffer.byteLength(preview, "utf-8")} bytes`,
+          );
+          return { ...(item as Record<string, unknown>), text: preview };
+        }
+      }
+      return item;
+    });
+
+    if (!mutated) {
+      // Oversized, but no single text item warranted truncation (e.g. many
```
Tara-ag: **MAJOR — oversized MCP envelopes are returned fully untruncated when no single text item exceeds the budget**

The envelope's byte ceiling (`outputTruncationMaxBytes`, default 51,200) is enforced **per text item only, never on the envelope total**. When a recognized envelope is oversized but no single text item individually exceeds `maxBytes`, this branch returns the result byte-for-byte unchanged:

- **Many moderate text blocks** (e.g. a real MCP server returning `content` spread across 5×40 KB items) → total ~200 KB, returned unchanged.
- **Overnight in a non-text item** (a large `image`/`resource`/`embedded` item) → never truncated, returned unchanged.

In both cases `serializedSize > maxBytes` holds, yet the result still flows into the AI SDK accumulator at full size — so BZ-666's context-overflow safety net is silently not enforced for exactly the multi-item envelope case this PR adds. Note `truncateToolResult`'s generic fallback *would* have bounded it by collapsing to the `_truncated` sentinel; this new early-return bypasses that. The same reasoning also means the `mutated` path (two items each trimmed to ~`maxBytes`) can still leave the envelope above budget, since per-item budgets don't sum to an envelope budget.

<details><summary>Suggested fix</summary>

Fall back to the generic sentinel (or proportionally trim all text items) so the serialized envelope lands under `maxBytes`, rather than returning it unchanged:

````suggestion
    if (!mutated) {
      // Oversized, but no single text item individually warranted truncation
      // (many moderate items, or the overrun sits in a non-text item). The
      // per-item budget does not bound the envelope total, so fall back to the
      // generic sentinel rather than returning an oversized result that
      // defeats BZ-666's context-overflow safety net.
      return undefined;
    }
````

This returns `undefined` so `truncateToolResult` falls through to the existing generic JSON branch, which collapses the oversized result to the `_truncated` sentinel and bounds the context. Caveat: that drops `isError` — if preserving error signals matters, prefer proportionally trimming the text items so the whole envelope stays under budget.
</details>

**Comment by Tara-ag** [comment]
## NEEDS_WORK

Shape-safe MCP truncation is a well-executed change (envelope preserved, `isError` + non-text items pass through, default byte budget unchanged so no behavior shift for existing consumers, deterministic all-src test), but the envelope byte ceiling is not actually enforced in the aggregate-oversized case.

| MAJOR | `src/lib/core/modules/ToolsManager.ts:345` | Oversized envelope returned fully untruncated when no single text item exceeds the budget (multi-item or non-text overrun) — byte ceiling silently not enforced |

### Checked and clean
- **Rule 3 / streaming**: truncation surfaces correctly (returns a shape-preserving result, never the sentinel for recognized envelopes); tool path stays on `wrapExecuteWithTruncation`/`truncateToolResult` used by `executeStream`.
- **Rule 5 (back-compat)**: no public `ToolConfig` change removes a field — `outputTruncationMaxBytes` is an additive optional; default `resolveTruncationMaxBytes` returns the original 51,200 so existing consumers see identical behavior.
- **Rule 15 + determinism**: new coverage is an all-src suite (`test/continuous-tests-fixed-tools-manager.ts`) with a declared determinism exception, and `eslint.config.js` allow-list + `package.json` script were updated in lockstep. Silly determinism rule satisfied.
- **Security**: no credentials/log leakage; `transformParamsForLogging` path untouched; no injection surface introduced.
- `truncateToolResult` nullability contract preserved (undefined → generic JSON handling); `resolveTruncationNotice` correctly defers to `generateToolOutputPreview` only when `retrieve_context` is actually registered.

One finding (MAJOR) left for the author — the per-item budget needs an aggregate envelope bound.

**Review by Tara-ag — CHANGES_REQUESTED** [review]
Reviewing the BZ-666 shape-safe MCP truncation. Overall the direction is solid (envelope shape preserved, `isError`/non-text items pass through, byte budget configurable via `ToolConfig`, deterministic all-src test). One MAJOR concern to address before merge — see the inline finding on the envelope byte budget.

File: src/lib/core/modules/ToolsManager.ts (hunk 1 above)
Comment (murdore) [reply src/lib/core/modules/ToolsManager.ts]: Addressed in 041e2e16: the ceiling is now enforced on the serialized envelope as a whole — fixed (non-text/structure) bytes come off the top, the remainder is water-filled across text items, and the result is re-measured and tightened until it fits. If the shape cannot be kept within budget (non-text payload alone over budget, or no text items) it returns undefined so the generic sentinel bounds it. Suite now asserts the bound for the 20x3KB case and adds 5x40KB, text+image, and image-over-budget cases (12/12 green).

**Review by Tara-ag — APPROVED** [review]
## APPROVE — shape-safe MCP truncation is correct and complete

The one MAJOR finding from the prior review (envelope byte ceiling not enforced in the aggregate-oversized case) is **fixed and resolved**: `truncateMcpContentEnvelope` now enforces the ceiling on the serialized envelope as a whole (fixed/non-text bytes come off the top, remainder water-filled across text items, re-measured and tightened until it fits; falls back to the generic sentinel when the shape cannot be kept in budget). The resolved thread and the new suite cases (20×3KB bound, 5×40KB, text+image, image-over-budget) confirm this.

### Checked and clean
- **Rule 3 / streaming**: truncation surfaces correctly; tool path stays on `wrapExecuteWithTruncation`/`truncateToolResult` used by `executeStream`.
- **Rule 5 (back-compat)**: `outputTruncationMaxBytes` is additive optional; default `resolveTruncationMaxBytes` returns the original 51,200 — no behavior shift for existing consumers.
- **Rule 15 + determinism**: all-src suite with declared determinism exception; `eslint.config.js` allow-list and `package.json` script updated in lockstep.
- **Security**: no credential/log leakage; `transformParamsForLogging` path untouched; no injection surface.
- **CI**: lint, types, validate/build, test suites, secret-scan all green; single-commit policy and docs validation pass.

No remaining findings. Approving.

File: src/lib/core/modules/ToolsManager.ts [hunk 2]
```diff
@@ -235,6 +285,166 @@ export class ToolsManager {
...
+  /**
+   * The default omission notice names `retrieve_context` as the on-demand
+   * read-back tool. That is only true when this NeuroLink instance actually
+   * registered it (Redis or an artifact store must exist — see
+   * `retrieveContextRegistered` in neurolink.ts). Otherwise hand
+   * `generateToolOutputPreview` a notice that doesn't mention it.
+   */
+  private resolveTruncationNotice(): ToolOutputPreviewOptions["notice"] {
+    const hasRetrieveContext = Boolean(
+      this.neurolink?.getCustomTools?.().has(RETRIEVE_CONTEXT_TOOL_NAME),
```
Tara-ag: **MINOR — `.has()` on an optional-chained return can throw; guard it**

```ts
this.neurolink?.getCustomTools?.().has(RETRIEVE_CONTEXT_TOOL_NAME)
```

Optional chaining short-circuits only when the *function* (`getCustomTools`) is nullish — **not** when it returns `undefined`. `getCustomTools()` is typed `Map<string, unknown> | undefined`, so if it ever returns `undefined` (e.g. custom tools not yet registered via `setupToolExecutor` on some path), `.has()` throws a `TypeError`. This is on a hot path: it runs inside `resolveTruncationNotice()`, which `truncateMcpContentEnvelope` calls for **every** oversized envelope on **every** tool result. A throw here aborts the whole truncation (and the tool call) with an opaque error instead of a bounded preview.

Since `resolveTruncationMaxBytes()` already treats the same accessor defensively (`?.getToolsConfig?.()?.outputTruncationMaxBytes`), this call site is the inconsistent one.

```suggestion
      this.neurolink
        ?.getCustomTools?.()
        ?.has(RETRIEVE_CONTEXT_TOOL_NAME) === true,
```

(Also converts `Boolean(...)` → `=== true`, so the result is a boolean either way.)

Tara-ag (approved):
Shape-safe MCP envelope truncation reviewed. One MINOR robustness note on `resolveTruncationNotice`; the previously-raised aggregate-envelope-budget MAJOR is confirmed resolved by the water-fill + re-measure/tighten logic in `truncateMcpContentEnvelope`.

Tara-ag:
## APPROVE

The `truncateMcpContentEnvelope` water-fill + re-measure/tighten rewrite (commit `041e2e16` → head `fb1197d`) directly resolves the previously-raised MAJOR: the byte ceiling is now enforced on the **serialized envelope as a whole**, not per item. The prior finding is correct to close — the 20×3KB and 5×40KB aggregate-oversized cases are now bounded with shape preserved, and the non-text-payload-over-budget case correctly falls back to the `_truncated` sentinel. No blocking issues found.

| MINOR | `src/lib/core/modules/ToolsManager.ts:315` | `getCustomTools?.().has(...)` can throw if the (union-typed `Map \| undefined`) accessor ever returns `undefined`; hot-path roónos. Guard with `?.has?...` — inline suggestion posted. |

### Checked and clean
- **Rule 3 / truncation surfacing**: recognized envelopes are preserved, `isError` + non-text items pass through; unknown shapes still get the legacy sentinel. `notice` is configurable and `retrieve_context` is only named when the tool is actually registered.
- **Rule 5 (back-compat)**: `outputTruncationMaxBytes` is an additive optional `ToolConfig` field; `resolveTruncationMaxBytes()` defaults to the original hard-coded 51,200. All 9 external callers of `generateToolOutputPreview`/`exceedsToolOutputPreviewBudget` (anthropicLoopGuard, geminiLoopGuard, openaiCompatLoopGuard, redisConversationMemoryManager, mcpOutputNormalizer) are unaffected — the new `notice`/`maxBytes` options are additive.
- **Blast radius**: `wrapExecuteWithTruncation` still funnels all three tool paths (direct/custom/external-MCP); `truncateToolResult`'s nullability contract (undefined → generic path) preserved.
- **Rule 15 + determinism**: new suite is all-src with a declared determinism exception; `eslint.config.js` allow-list and `package.json` script updated in lockstep.
- **Coverage**: 12 assertions including the previously-failing aggregate-envelope cases; non-text-over-budget and legacy-sentinel-compat cases covered.
- **Security**: no credentials/log leakage; `transformParamsForLogging` untouched; no injection surface.

One MINOR remains open for the author — the optional-chain `.has()` call deserves a null-unwrapping guard.

Tara-ag:
## APPROVE

Recurring review of the shape-safe MCP truncation fix. Both previously-raised findings are now **resolved**; no new findings.

| Prior finding | Status now |
|---------------|------------|
| MAJOR — `ToolsManager.ts:345` envelope byte ceiling not enforced in the aggregate-oversized case | **Resolved** — `truncateMcpContentEnvelope` now enforces the budget on the serialized envelope as a whole (fixed/non-text bytes off the top, water-fill across text items, re-measure/tighten until it fits; falls back to the `_truncated` sentinel when the shape can't be kept in budget). Covered by the 20×3KB / 5×40KB / text+image / image-over-budget suite cases. |
| MINOR — `ToolsManager.ts:315` `getCustomTools?.().has(...)` could throw on a `undefined` map return | **Resolved in code** — the call site now guards with `?.has(...)` (exactly the reviewer's suggested fix), and the suite adds an explicit "getCustomTools() returns undefined does not throw" case. Thread resolved this pass. |

### Checked and clean
- **Rule 3 / streaming**: truncation surfaces a shape-preserving result for recognized envelopes and never silently drops `isError`/non-text items; unknown shapes still get the legacy sentinel; hot path stays on `wrapExecuteWithTruncation` → `truncateToolResult` across direct/custom/external-MCP tool paths.
- **Rule 5 (back-compat)**: `ToolConfig.outputTruncationMaxBytes` is an additive optional field; `resolveTruncationMaxBytes()` defaults to the original 51,200 / `DEFAULT_MAX_PREVIEW_BYTES`. The new `notice`/`maxBytes` options to `generateToolOutputPreview` are additive — all 7 existing external callers (anthropic/gemini/openaiCompat loop guards, redisConversationMemoryManager, mcpOutputNormalizer) are unchanged.
- **Rule 15 + determinism**: new suite is all-src with a declared determinism exception, and the `eslint.config.js` `e2e-tests-only` allow-list + `package.json` script were updated in lockstep.
- **Blast radius**: `wrapExecuteWithTruncation` funnels all three tool result paths; `truncateToolResult`'s nullability contract (undefined → generic sentinel) is preserved.
- **Security**: no credential/log leakage; `transformParamsForLogging` untouched; no injection surface.

One suggestion (non-blocking): per-text-item notices mean a many-item envelope carries several `[... N bytes omitted ...]` markers; consider a single envelope-level omission notice if the noise matters — not required for merge.

src/lib/core/modules/ToolsManager.ts
```diff
@@ -244,10 +454,14 @@ export class ToolsManager {
       return result;
     }
 
+    const maxBytes = this.resolveTruncationMaxBytes();
+
     // Handle string results directly
     if (typeof result === "string") {
-      const { preview, truncated, originalSize } =
-        generateToolOutputPreview(result);
+      const { preview, truncated, originalSize } = generateToolOutputPreview(
```
Tara-ag, line 461: **MINOR — the `retrieve_context`-suppression only applies to MCP-envelope results, not the plain-string / string-content / JSON-sentinel truncation paths**

`resolveTruncationNotice()` (the notice that drops the `retrieve_context` mention when that tool is not registered) is only passed to `generateToolOutputPreview` inside `truncateMcpContentEnvelope` (line 419). Every other truncation branch in `truncateToolResult` calls `generateToolOutputPreview` with just `{ maxBytes }` and no `notice` — the plain-string branch here (line 461), the string `obj.content` field, the string `obj.data` field, and the generic `_truncated` JSON-sentinel branch. Those all fall back to `generateToolOutputPreview`'s built-in default notice, which hard-codes `Use ${RETRIEVE_CONTEXT_TOOL_NAME} tool to access full output ...`.

This contradicts the PR's stated intent ("ToolsManager uses it to omit the retrieve_context mention unless that tool is actually registered on the instance"). A direct/custom tool that returns an oversized **plain string** (the common case for a readFile/bash-style tool) on an instance without `retrieve_context` still tells the model to call a tool that doesn't exist, which the model may then attempt and fail on. The gap is also uncovered: every added test drives an envelope (`{content:[...]}` or `{success,data}`) — none feeds an oversized plain string / plain object with `retrieve_context` unregistered.

Suggestion — compute the notice once and thread it through all `truncateToolResult` branches:

```suggestion
    // Handle string results directly
    if (typeof result === "string") {
      const { preview, truncated, originalSize } = generateToolOutputPreview(
        result,
        { maxBytes, notice: this.resolveTruncationNotice() },
      );
```
The same `notice:` option should be added to the `obj.content`, `obj.data`, and generic-JSON-sentinel calls in this method (e.g. hoist `const notice = this.resolveTruncationNotice();` next to `maxBytes` and reuse it), so a missing `retrieve_context` on the instance never points the model at a tool it cannot call.

Tara-ag (approved):
Recurring review (head `b4a48c64`). Both prior findings are **resolved**; one new MINOR (non-blocking) posted inline at `ToolsManager.ts:461`.

- **MAJOR — envelope budget** (previously `ToolsManager.ts:345`): **resolved**. `truncateMcpContentEnvelope` enforces the ceiling on the serialized envelope as a whole via fixed-bytes-off-the-top + water-fill across text items + re-measure/tighten; falls back to the `_truncated` sentinel when the shape can't be held in budget.
- **MINOR — `getCustomTools?.().has()` throw** (`ToolsManager.ts:315`): **resolved in code** with `?.has(...) ?? false`; covered by a dedicated suite case.
- **MINOR (new)** — the `retrieve_context`-notice suppression is wired only into the envelope path; plain-string, string-`content`/`data`, and generic-JSON-sentinel truncations in `truncateToolResult` still emit the hard-coded `Use retrieve_context tool...` notice even when that tool isn't registered, contradicting the PR's stated intent and uncovered by tests.

Checked clean: Rule 3 (truncation surfaces, never silent), Rule 5 back-compat (`outputTruncationMaxBytes` additive, default 51,200 unchanged), blast radius (sole caller `wrapExecuteWithTruncation`; all `generateToolOutputPreview` callers unaffected), Rule 15 + determinism (all-src suite, allow-list + package script in lockstep), security (no credential/log/injection surface).

src/lib/core/modules/ToolsManager.ts
```diff
@@ -291,13 +508,27 @@ export class ToolsManager {
...
       // For other objects, check if their JSON serialization is too large.
-      // Use UTF-8 byte length, not string length, to match the 50KB budget.
+      // Use UTF-8 byte length, not string length, to match the byte budget.
       try {
         const jsonStr = JSON.stringify(result);
-        if (Buffer.byteLength(jsonStr, "utf-8") > 51_200) {
+        if (Buffer.byteLength(jsonStr, "utf-8") > maxBytes) {
           const { preview, truncated, originalSize } =
-            generateToolOutputPreview(jsonStr);
+            generateToolOutputPreview(jsonStr, { maxBytes });
```
Tara-ag: **MINOR — the `retrieve_context`-notice suppression is wired only into the envelope path**

`resolveTruncationNotice()` is passed (as `notice`) only inside `truncateMcpContentEnvelope`, which runs for recognized MCP-envelope shapes. This generic JSON-sentinel branch — and the plain-string branch and the string `content`/`data` branches above — still call `generateToolOutputPreview(...)` with just `{ maxBytes }`, so they emit the hard-coded `Use retrieve_context tool to access full output …` notice even when that tool is not registered.

That contradicts the PR's stated intent ("omit the retrieve_context mention unless that tool is actually registered on the instance") and points the model at a nonexistent tool for any oversized plain-string or non-envelope-object result on an instance without `retrieve_context`. The suite's notice assertions only drive envelope shapes, so none of these branches is covered.

```suggestion
          const { preview, truncated, originalSize } =
            generateToolOutputPreview(jsonStr, {
              maxBytes,
              notice: this.resolveTruncationNotice(),
            });
```

Tara-ag (approved):
Recurring review (head `b4a48c64`). Both prior findings resolved; one MINOR (non-blocking) anchored inline at `ToolsManager.ts:530-531`. Verdict: APPROVE.

src/lib/core/modules/ToolsManager.ts
murdore: Confirmed and fixed. You were right, and the scope was exactly as you described: of the five `generateToolOutputPreview` call sites in `ToolsManager.ts`, only the envelope one at line 419 passed `notice`. The other four — plain string (461), string `content` (481), string `data` (495), and the JSON sentinel (531) — passed `{ maxBytes }` alone and fell through to the built-in default that names `retrieve_context` unconditionally.

All four now resolve the notice the same way the envelope path does.

The coverage gap was real too — the suite only ever drove envelope shapes. Five cases added: the four non-envelope shapes asserted to omit the name when `retrieve_context` is unregistered, plus a positive case confirming a plain string still names it when the tool *is* registered, so the fix cannot be satisfied by simply never emitting the mention.

Each new case carries preconditions, because "the notice does not name the tool" is trivially true of a payload that was never truncated:

```ts
assert(Buffer.byteLength(text, "utf-8") < BIG_TEXT.length, "precondition: the result was not truncated at all");
assert(/bytes omitted/.test(text), "precondition: no omission notice was emitted");
```

Verified red/green rather than assumed. With the fix: 18 passed. Reverting just the four `notice:` additions and keeping the tests:

```
14 passed · 4 failed
✗ a plain oversized string result: notice omits retrieve_context when it is unregistered
✗ an object with an oversized string `content`: …
✗ an object with an oversized string `data`: …
✗ an oversized plain object (generic JSON sentinel branch): …
```

The positive registered-case stays green in both, which is what makes the four failures specific to this defect.

Your two threads are the same finding; this reply covers both.

Tara-ag:
## APPROVE

Recurring review of head `a944e73a`. All three previously-raised findings are **resolved** in the current code; I found **no new findings**. The `notice-applies-envelope-only` thread (the last open one) is resolved this pass — the author's fix is present in head and verified line-by-line.

| Prior finding | Status | Verified in head (`a944e73a`) |
|---------------|--------|-------------------------------|
| MAJOR — `ToolsManager.ts` envelope byte ceiling not enforced in the aggregate-oversized case | **Resolved** | `truncateMcpContentEnvelope` enforces the budget on the serialized envelope as a whole — fixed/non-text bytes off the top, water-fill across text items, re-measure/tighten (`MAX_ENVELOPE_TIGHTENING_PASSES`), returns `undefined` → sentinel when the shape can't fit. Tests: 20×3KB, 5×40KB, text+image, image-over-budget. |
| MINOR — `ToolsManager.ts:315` `getCustomTools?.().has(...)` could throw on a `undefined` map return | **Resolved** | Call site now `…?.getCustomTools?.()?.has(...) ?? false`; suite adds a "map undefined does not throw" case. |
| MINOR — `retrieve_context`-notice suppression wired only into the envelope path | **Resolved** | All four non-envelope `generateToolOutputPreview` call sites (plain string ~:458, string `content` ~:477, string `data` ~:494, JSON sentinel ~:526) now pass `notice: this.resolveTruncationNotice()`. Five new cases added (four unregistered + one positive registered). |

### Checked and clean
- **Rule 3 / truncation surfacing**: recognized envelopes preserved (`isError`, non-text items, other top-level fields); unknown shapes still get the legacy sentinel; truncation is never silent (notice always emitted).
- **Blast radius**: `ToolsManager.ts` imported only by `baseProvider.ts` + the suite; `wrapExecuteWithTruncation` → `truncateToolResult` funnels all three tool paths. `generateToolOutputPreview` has 9 callers (4 loop guards, `redisConversationMemoryManager`, `mcpOutputNormalizer`, `ToolsManager×2`, test); the new `notice`/`maxBytes` options are additive and optional — no caller shifts behavior.
- **Rule 5 (back-compat)**: `ToolConfig.outputTruncationMaxBytes` and `ToolOutputPreviewOptions.notice` are additive optionals; `resolveTruncationMaxBytes()` defaults to the original 51,200 / `DEFAULT_MAX_PREVIEW_BYTES`. Accessors `getToolsConfig(): ToolConfig | undefined` and `getCustomTools(): Map<string,unknown> | undefined` confirmed present on the instance.
- **Rule 15 + determinism**: new suite is all-src with a declared determinism exception; `eslint.config.js` `e2e-tests-only` allow-list and `package.json` `test:tools-manager-truncation` script updated in lockstep.
- **Algorithm review**: water-fill (+ empty-item handling), multi-pass tightening, zero/non-convergence → graceful sentinel fallback — no correctness or infinite-loop risk found.
- **Security**: no credential/log leakage; `transformParamsForLogging` untouched; no injection surface.

One non-blocking note (carried from the prior pass, not required for merge): a many-item envelope carries one `[... N bytes omitted ...]` notice per truncated text item; a single envelope-level omission notice would read more cleanly in the model-visible text. Optional.

No inline findings this cycle — nothing new to fix before merge.

src/lib/core/modules/ToolsManager.ts
Tara-ag: Confirmed fixed in head — this thread is the same finding as `toolsmanager-notice-only-envelope-path` (line 530-531), and the author's fix threads `resolveTruncationNotice()` through all four non-envelope branches. Resolving this duplicate; the canonical thread with the author's reply carries the resolution.

src/lib/core/modules/ToolsManager.ts
Tara-ag: Confirmed fixed in head — all four non-envelope branches now pass `notice: this.resolveTruncationNotice()`, verified with the five added test cases (4 unregistered + 1 positive registered). Resolving.

package.json
```diff
@@ -158,6 +158,7 @@
     "test:agents-live": "tsx test/continuous-test-agents-live.ts",
     "test:tool-routing-cli": "pnpm exec tsx test/continuous-test-suite-tool-routing-cli.ts",
     "test:tool-dedup": "pnpm exec tsx test/continuous-test-suite-tool-dedup.ts",
+    "test:tools-manager-truncation": "pnpm exec tsx test/continuous-test-suite-tools-manager-truncation.ts",
```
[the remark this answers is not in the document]
Tara-ag: Confirmed independently — this finding is valid and still open in head `7ee69f98`.

I checked every CI-invoked path and `test:tools-manager-truncation` runs in **none** of them:

- `test:unit` — not in the aggregate (`package.json`).
- `test:ci` (`test && test:client && test:hitl`) — not included.
- `test` (`continuous-test-suite.ts`) — no reference.
- CI `extended-suites` job (`.github/workflows/ci.yml`) — the `SUITES` list that exists precisely so suites CI otherwise doesn't run get executed does **not** contain `test:tools-manager-truncation`.

Only the `lint` shard touches the file, via the `test/continuous-test-suite*.ts` glob — so it's format/lint-checked but **never executed**. ci.yml says it plainly: *"A suite wired into nothing is documentation, not a test."* For a security-relevant suite guarding the BZ-666 context-overflow net, a regression would ship silently.

Concrete fix — add one entry to the `extended-suites` `SUITES` list in `.github/workflows/ci.yml`:

```
test:tools-manager-truncation@3
```

(`@3` = measured run-time ballpark; the harness's own header says `offline: true`, no credentials, all-src exception so the built `dist` isn't a prerequisite.) Alternatively append it to the `test:unit` aggregate in `package.json` if it must be a hard required gate.

Tara-ag:
## NEEDS_WORK

Recurring review of head `7ee69f98`. The shape-safe MCP truncation change itself is correct and my three previously-raised findings are all **resolved** and verified in the current code. One new **MAJOR** surfaced this cycle: the new truncation suite is wired into **no** CI-invoked aggregate, so the fix ships unguarded.

| MAJOR | `package.json` / `.github/workflows/ci.yml` | `test:tools-manager-truncation` exists but is in none of `test`, `test:unit`, `test:ci`, or the CI `extended-suites` `SUITES` list — the suite is lint-only, never executed. ci.yml itself: *“A suite wired into nothing is documentation, not a test.”* A regression in the BZ-666 context-overflow guard would go green silently. |

### Prior findings — all resolved in head
- **MAJOR — envelope byte ceiling not enforced in aggregate-oversized case** (`ToolsManager.ts`): **resolved**. `truncateMcpContentEnvelope` enforces the ceiling on the serialized envelope as a whole (fixed/non-text bytes off the top, water-fill across text items, re-measure/tighten, `undefined` → sentinel when the shape can't fit). Covered by the 20×3KB / 5×40KB / text+image / image-over-budget cases.
- **MINOR — `getCustomTools?.().has()` throw on `undefined`** (`ToolsManager.ts:315`): **resolved**. Call site guards with `?.has(...) ?? false`; suite adds a "map undefined does not throw" case.
- **MINOR — `retrieve_context`-notice suppression only on the envelope path**: **resolved**. All four non-envelope branches (string, `content`, `data`, JSON sentinel) now pass `notice: this.resolveTruncationNotice()`; five new cases pin the behaviour.

### Checked clean this cycle
- **Rule 3 / streaming**: truncation surfaces visibly, never silent; recognized envelopes keep `isError` + non-text items; unknown shapes still get the legacy `_truncated` sentinel.
- **Rule 5 (back-compat)**: `ToolConfig.outputTruncationMaxBytes` and `ToolOutputPreviewOptions.notice` are additive optionals; `resolveTruncationMaxBytes()` defaults to the original 51,200.
- **`generateToolOutputPreview` notice over-budget**: CodeRabbit's finding is verified **fixed** — the two-pass head/tail shrink reserves bytes for the notice so the assembled preview stays ≤ `maxBytes`; the suite pins it across 500/2k/8k/40k budgets and short/long/literal notices. No duplicate finding raised.
- **Blast radius**: `wrapExecuteWithTruncation` → `truncateToolResult` funnels all three tool paths; nullability contract preserved; the `_truncated` sentinel path for non-envelope objects is kept for compat.
- **Rule 15 + determinism**: all-src suite with a declared determinism exception; `eslint.config.js` allow-list and the `package.json` script are in lockstep (the gap is execution, not the allow-list).
- **Security**: no credential/log/injection surface; `transformParamsForLogging` untouched.

**How to clear the MAJOR**: add `test:tools-manager-truncation@3` to the `SUITES` list in `.github/workflows/ci.yml` `extended-suites` (offline, no credentials; all-src so the built `dist` isn't a prerequisite) — or append it to the `test:unit` aggregate if it must be a required gate. Inline thread on `package.json:161` carries the detail.

Apart from that one line of CI config, the change is ready to merge.

Tara-ag (changes requested):
Recurring review of head `7ee69f98` — verdict **NEEDS_WORK**.

The shape-safe MCP truncation change is correct and all three code findings raised in earlier cycles are resolved in the current tree. One **MAJOR** remains open and untested-by-CI: `test:tools-manager-truncation` is wired into **no** CI-invoked aggregate (`test`, `test:unit`, `test:ci`, and the `extended-suites` `SUITES` list all omit it), so the regression suite guarding the BZ-666 context-overflow net is lint-only and never executed. Full detail is in the summary comment; the fix is one line — add `test:tools-manager-truncation@3` to the `SUITES` list in `.github/workflows/ci.yml`.

src/lib/context/toolOutputLimits.ts
```diff
@@ -114,9 +114,17 @@ export function generateToolOutputPreview(
...
+  // tool that doesn't exist.
   const notice =
-    `\n\n[... ${omittedBytes} bytes omitted. ` +
-    `Use ${RETRIEVE_CONTEXT_TOOL_NAME} tool to access full output ...]\n\n`;
+    options?.notice === undefined
+      ? `\n\n[... ${omittedBytes} bytes omitted. ` +
+        `Use ${RETRIEVE_CONTEXT_TOOL_NAME} tool to access full output ...]\n\n`
+      : typeof options.notice === "function"
+        ? options.notice(omittedBytes)
+        : options.notice;
```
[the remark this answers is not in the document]
murdore: Confirmed, fixed, and worth recording that it was **pre-existing rather than introduced here** — the default notice already overflowed.

`headMaxBytes + tailMaxBytes` is exactly `maxBytes`, leaving no room at all for the notice, which was then appended unmeasured. Measured on a 60 KB input:

```
maxBytes = 8000
default notice      = 8082 bytes  -> 82 over
ToolsManager notice = 8033 bytes  -> 33 over
```

So the custom notices this PR routes to four more branches are in fact *shorter* than the built-in default and overflow less — but the guarantee was broken either way, and a long literal notice breaks it unboundedly.

The notice now takes its space out of the tail, then the head, in two passes — two because shrinking them raises `omittedBytes`, which can lengthen the notice by a digit. Every combination now lands exactly at the budget:

```
maxBytes=500    default=500    custom=500    300b-notice=500
maxBytes=2000   default=2000   custom=2000   300b-notice=2000
maxBytes=8000   default=8000   custom=8000   300b-notice=8000
maxBytes=40000  default=40000  custom=40000  300b-notice=40000
```

Four new cases assert `preview <= maxBytes` across those budgets for the default, a short callback and a 300-byte literal, each with a precondition that truncation actually occurred — an untruncated preview is trivially within budget and would prove nothing.

Verified red/green: removing just the reclaim loop turns exactly those four red (`18 passed` → `14 passed · 4 failed`) while the other eighteen stay green.

package.json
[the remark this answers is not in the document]
murdore: Right — the script existed in `package.json` but nothing invoked it, so the suite was registered and unrun. Now added to the `extended-suites` `SUITES` list in `ci.yml`:

```
test:tools-manager-truncation@3
```

The weight is measured, not guessed: the suite completes in 3s (22 passing). It has no provider credentials to need — every case drives `ToolsManager` through `getAllTools()` with a stubbed instance — so it runs for real in that job rather than skipping.

One deliberate detail: I anchored the entry at the tail of the list, after `test:browser-bundle@5`, rather than next to the other short suites. #1673 inserts its own entry after `test:tool-reliability@57` in the same list, and keeping the two insertions far apart stops two of my own ready PRs from conflicting over adjacent lines — the sort is by weight at runtime, so list position carries no meaning.

Tara-ag:
## APPROVE

Recurring review of head `811a0eb`. The previous cycle's **MAJOR** (truncation suite not wired into any CI-invoked aggregate) is now **resolved**: `test:tools-manager-truncation@3` has been added to the `extended-suites` `SUITES` list in `.github/workflows/ci.yml`, and the `package.json` `test:tools-manager-truncation` script exists in lockstep. Every finding raised across the review cycles is resolved and verified in the current tree.

| Prior finding | Status in head `811a0eb` |
|---------------|--------------------------|
| MAJOR — envelope byte ceiling not enforced in the aggregate-oversized case | **Resolved** — `truncateMcpContentEnvelope` enforces the ceiling on the serialized envelope as a whole (fixed/non-text bytes off the top, water-fill across text items, re-measure/tighten up to `MAX_ENVELOPE_TIGHTENING_PASSES`, returns `undefined`→`_truncated` sentinel when the shape can't fit). Covered by the 20×3KB / 5×40KB / text+image / image-over-budget cases. |
| MINOR — `getCustomTools?.().has()` could throw on a `undefined` map return | **Resolved** — call site is `…?.getCustomTools?.()?.has(RETRIEVE_CONTEXT_TOOL_NAME) ?? false`; dedicated "map undefined does not throw" suite case. |
| MINOR — `retrieve_context`-notice suppression only on the envelope path | **Resolved** — all four non-envelope `generateToolOutputPreview` call sites (plain string, string `content`, string `data`, generic JSON sentinel) now pass `notice: this.resolveTruncationNotice()`; five cases pin the unregistered/registered behaviour. |
| CodeRabbit — custom/literal notice can push preview past `maxBytes` | **Resolved** — `generateToolOutputPreview` reclaims the notice's bytes from tail then head in two passes (re-rendering because `omittedBytes` shifts); suite pins `preview ≤ maxBytes` across 500/2k/8k/40k budgets and short-callback / 300-byte-literal notices. |
| MAJOR — truncation suite not run by CI | **Resolved** — `test:tools-manager-truncation@3` added to the `extended-suites` `SUITES` list in `ci.yml`. |

### Checked clean this cycle
- **Rule 3 / streaming**: truncation always surfaces (never silent) for recognized envelopes; `isError` and non-text items pass through untouched; unknown shapes still collapse to the legacy `_truncated` sentinel. Envelope path rides `wrapExecuteWithTruncation` → `truncateToolResult` (sole caller confirmed via graph: `wrapExecuteWithTruncation` funnels direct/custom/external-MCP tool paths).
- **Rule 5 (back-compat)**: `ToolConfig.outputTruncationMaxBytes` and `ToolOutputPreviewOptions.notice` are additive optionals; `resolveTruncationMaxBytes()` defaults to `DEFAULT_MAX_PREVIEW_BYTES` (51,200) when unset or non-finite/non-positive; the new options to `generateToolOutputPreview` are optional so its other 7 external callers are unaffected.
- **Rule 15 + determinism**: all-src suite with a declared determinism exception; `eslint.config.js` allow-list and `package.json` script are in lockstep.
- **Security**: no credential/log leakage; `transformParamsForLogging` untouched; no injection surface.

One non-blocking suggestion carried from earlier cycles (not a blocker): a many-item envelope carries one `[... N bytes omitted ...]` notice per truncated text item; a single envelope-level omission notice would read more cleanly in the model-visible text. Optional.

The change is correct, complete and CI-wired. Approving.

Tara-ag:
## APPROVE

Recurring review of head `811a0eb`. The only open finding from the prior cycle — the truncation suite being wired into no CI-invoked aggregate — is **resolved**: `test:tools-manager-truncation@3` is now in the `extended-suites` `SUITES` list in `.github/workflows/ci.yml`, and the `test:tools-manager-truncation` script is present in `package.json`. No new findings. This summary supersedes the stalcedd NEEDS_WORK from the prior pass.

### Findings — none open

| Prior finding | Status |
|---------------|--------|
| MAJOR — `ToolsManager.ts` envelope byte ceiling not enforced in the aggregate-oversized case | **Resolved** — `truncateMcpContentEnvelope` enforces the budget on the serialized envelope as a whole (fixed/non-text bytes off the top, water-fill across text items, re-measure/tighten, `undefined` → sentinel). Covered by 20×3KB / 5×40KB / text+image / image-over-budget cases. |
| MINOR — `getCustomTools?.().has()` throw on `undefined` | **Resolved** — call site guards with `?.has(...) ?? false`; suite adds a "map undefined does not throw" case. |
| MINOR — `retrieve_context`-notice suppression only on the envelope path | **Resolved** — all four non-envelope branches pass `notice: this.resolveTruncationNotice()`; five new cases pin the behaviour. |
| MAJOR — suite not wired into CI | **Resolved** — added to `extended-suites` `SUITES` as `test:tools-manager-truncation@3`. |

### Checked and clean
- **Rule 3 / truncation surfacing**: recognized envelopes preserved (`isError`, non-text items, other top-level fields); unknown shapes keep the legacy `_truncated` sentinel; truncation never silent.
- **Rule 5 (back-compat)**: `ToolConfig.outputTruncationMaxBytes` and `ToolOutputPreviewOptions.notice` are additive optionals; `resolveTruncationMaxBytes()` defaults to the original 51,200. No caller shifts behavior.
- **Blast radius**: `ToolsManager.ts` imported only by `baseProvider.ts` + the suite; `wrapExecuteWithTruncation` → `truncateToolResult` funnels all three tool paths; nullability contract preserved.
- **Rule 15 + determinism**: all-src suite with a declared determinism exception; `eslint.config.js` allow-list, `package.json` script, and CI `SUITES` entry are now in lockstep.
- **Security**: no credential/log/injection surface; `transformParamsForLogging` untouched.

One non-blocking note (unchanged): a many-item envelope carries one `[... N bytes omitted ...]` notice per truncated text item; a single envelope-level notice would read more cleanly in model-visible text. Optional — not required for merge.

No inline findings this cycle — nothing left to fix before merge.

Tara-ag (approved):
Recurring review — verdict **APPROVE**. The only remaining MAJOR from the prior cycle (the truncation suite wired into no CI-invoked aggregate) is resolved: `test:tools-manager-truncation@3` is now in the `extended-suites` `SUITES` list in `.github/workflows/ci.yml`. All prior code findings (envelope byte ceiling, `getCustomTools?.().has()` throw, notice-threading across non-envelope branches) are verified resolved in head. No new findings. Supersedes the prior CHANGES_REQUESTED review on this state.

src/lib/context/toolOutputLimits.ts
```diff
@@ -108,15 +108,54 @@ export function generateToolOutputPreview(
...
+    );
+
+  // `headMaxBytes + tailMaxBytes` is exactly `maxBytes`, leaving no room for
+  // the notice — so appending it unmeasured pushed every truncated preview
+  // over the caller's budget (82 bytes over, for the default notice). Give the
+  // notice its space back out of the tail, then the head. Two passes, because
+  // shrinking them raises `omittedBytes` and can lengthen the notice by a
+  // digit; the second pass settles that.
+  let notice = renderNotice(omittedFor(head, tail));
+  for (let pass = 0; pass < 2; pass++) {
```
Tara-ag, line 160: **MINOR — a custom `notice` larger than `maxBytes` still overflows the returned preview**

Agrees with the open CodeRabbit finding on this region; carrying it here for the recurring pass.

`headMaxBytes + tailMaxBytes == maxBytes`, so the two-pass reclaim loop below can only ever free budget back from `head`/`tail`. It never constrains `notice` itself:

```ts
let notice = renderNotice(omittedFor(head, tail));
for (let pass = 0; pass < 2; pass++) {
  const excess = byteLength(head) + byteLength(notice) + byteLength(tail) - maxBytes;
  ...
  // shrinks tail, then head — never the notice
}
return { preview: head + notice + tail, ... };
```

If a caller passes a `notice` (a `ToolOutputPreviewOptions.notice` literal or a callback that returns a long string) whose own byte length exceeds `maxBytes`, draining `head`/`tail` to empty leaves `preview = "" + notice + ""`, still over budget. E.g. `notice: "X".repeat(300)` with `maxBytes: 100` returns a ~300-byte preview. The existing 300-byte-literal test only uses budgets ≥ 500, so the guarantee that "the assembled preview stays ≤ `maxBytes`" (claimed in the PR and asserted by the suite) is not actually upheld for a notice that alone exceeds the ceiling.

In NeuroLink's own paths this is latent (the notices `resolveTruncationNotice` produces are a short callback and the default), but `generateToolOutputPreview` is a public helper with 7+ external callers (loop guards, `mcpOutputNormalizer`, `redisConversationMemoryManager`) that can supply arbitrary notices — so the parse is worth closing.

```suggestion
  if (Buffer.byteLength(notice, "utf-8") >= maxBytes) {
    // A notice that alone exceeds the budget cannot be spliced in: bound it
    // to a static fallback that names neither retrieve_context nor the
    // omitted count (which would itself blow the budget), so every truncated
    // preview stays <= maxBytes regardless of caller-supplied notice length.
    notice = `[... ${maxBytes} bytes ...]`;
    return { preview: notice, truncated: true, originalSize };
  }
```

Suggested test: `generateToolOutputPreview(BIG_TEXT, { maxBytes: 100, notice: "X".repeat(300) })` must return `byteLength(preview) <= 100`.

Tara-ag (approved):
Recurring review of head `de8ac14` (last commit removes a stray `__probe__.txt`; functional code unchanged since the prior APPROVE). Verdict **APPROVE**.

All previously-raised findings remain resolved in the current tree; one **MINOR** (non-blocking) is carried from the still-open CodeRabbit thread and confirmed inline at `toolOutputLimits.ts:138`.

### Findings
| Severity | Location | Finding | Disposition |
|----------|----------|---------|-------------|
| MINOR | `src/lib/context/toolOutputLimits.ts:138` | A caller-supplied `notice` whose byte length alone exceeds `maxBytes` still overflows the returned preview — the reclaim loop only shrinks `head`/`tail`, never the notice. Latent in NeuroLink's own paths; reachable by the 7+ external callers of `generateToolOutputPreview`. | Kept — author has not replied; inline suggestion + test posted. |

### Prior findings — verified resolved, author replies adequate
- **MAJOR — envelope byte ceiling** (`ToolsManager`): resolved — serialized-envelope-as-a-whole water-fill + re-measure/tighten, `undefined` → sentinel fallback.
- **MINOR — `getCustomTools?.().has()` throw**: resolved — `?.has(...) ?? false` guard + suite case.
- **MINOR — `retrieve_context` notice suppression only on envelope path**: resolved — all four non-envelope branches pass `notice: this.resolveTruncationNotice()`; five pinning cases.
- **CodeRabbit — default/custom notice pushing preview over `maxBytes`**: resolved — two-pass head/tail reclaim; pinned across budgets.
- **MAJOR — suite not wired into CI**: resolved — `test:tools-manager-truncation@3` in the `extended-suites` `SUITES` list.

### Checked and clean
- **Rule 3 / streaming**: truncation always surfaces (never silent); recognized envelopes keep `isError` + non-text items; unknown shapes still collapse to the legacy `_truncated` sentinel; path stays on `wrapExecuteWithTruncation` → `truncateToolResult`.
- **Rule 5 (back-compat)**: `ToolConfig.outputTruncationMaxBytes` and `ToolOutputPreviewOptions.notice` are additive optionals; `resolveTruncationMaxBytes()` defaults to the original 51,200. No caller shifts behavior.
- **Blast radius**: `ToolsManager` imported only by `baseProvider.ts` + the suite; `wrapExecuteWithTruncation` funnels all three tool paths. The new `notice`/`maxBytes` options are optional — 7 external `generateToolOutputPreview` callers unaffected.
- **Rule 15 + determinism**: all-src suite with declared determinism exception; `eslint.config.js` allow-list, `package.json` script, and CI `SUITES` entry in lockstep.
- **Security**: no credential/log/injection surface; `transformParamsForLogging` untouched.

One non-blocking note carried from prior cycles: a many-item envelope emits one per-item omission notice; a single envelope-level notice would read more cleanly. Optional.

The change is correct, complete, and CI-wired. The open MINOR is an edge case in a shared helper, not a merge blocker.

Tara-ag:
## APPROVE

Recurring review of head `de8ac14` (last commit just deletes a stray `__probe__.txt` — functional code is unchanged since the prior APPROVE). The shape-safe MCP truncation change is correct, complete and CI-wired; every prior finding is resolved. One **MINOR** (non-blocking) edge case remains open and is confirmed inline.

|----------|----------|---------|--------|
| MINOR | `src/lib/context/toolOutputLimits.ts:138` | A caller-supplied `notice` whose byte length alone exceeds `maxBytes` still overflows the returned preview — the reclaim loop only shrinks `head`/`tail`, never the notice. Latent in NeuroLink's own paths; reachable by the 7+ external `generateToolOutputPreview` callers. | **Open** — kept; no author reply yet. Inline suggestion + test posted. |

### Prior findings — all resolved, author replies adequate
- MAJOR — envelope byte ceiling not enforced in aggregate-oversized case → **resolved** (water-fill + re-measure/tighten on the serialized envelope; `undefined` → `_truncated` sentinel fallback).
- MINOR — `getCustomTools?.().has()` throw on a `undefined` map → **resolved** (`?.has(...) ?? false` + suite case).
- MINOR — `retrieve_context`-notice suppression only on the envelope path → **resolved** (all four non-envelope branches pass `notice: this.resolveTruncationNotice()`; five pinning cases).
- CodeRabbit — default/custom notice pushing preview over `maxBytes` → **resolved** (two-pass head/tail reclaim; pinned across budgets).
- MAJOR — truncation suite not run by CI → **resolved** (`test:tools-manager-truncation@3` in the `extended-suites` `SUITES` list).

### Checked and clean
- **Rule 3 / streaming**: truncation always surfaces (never silent); recognized envelopes keep `isError` + non-text items; unknown shapes keep the legacy `_truncated` sentinel; path rides `wrapExecuteWithTruncation` → `truncateToolResult` (direct/custom/external-MCP).
- **Rule 5 (back-compat)**: `ToolConfig.outputTruncationMaxBytes` and `ToolOutputPreviewOptions.notice` are additive optionals; `resolveTruncationMaxBytes()` defaults to the original 51,200 — no consumer behavior shift.
- **Blast radius**: `ToolsManager` imported only by `baseProvider.ts` + the suite; the new `notice`/`maxBytes` options are optional so the 7 external `generateToolOutputPreview` callers (loop guards, `mcpOutputNormalizer`, `redisConversationMemoryManager`) are unaffected.
- **Rule 15 + determinism**: all-src suite with a declared determinism exception; `eslint.config.js` allow-list, `package.json` script, and CI `SUITES` entry in lockstep.
- **Security**: no credential/log/injection surface; `transformParamsForLogging` untouched.

**Non-blocking note** (carried from prior cycles): a many-item envelope emits one `[... N bytes omitted ...]` notice per truncated text item; a single envelope-level notice would read more cleanly in the model-visible text. Optional.

The lone open MINOR is an edge case in a shared helper, not a merge blocker.

src/lib/context/toolOutputLimits.ts
```diff
@@ -108,15 +108,54 @@ export function generateToolOutputPreview(
...
+    tail = tailBuf.subarray(takeFromTail).toString("utf-8");
+    const stillOver = excess - takeFromTail;
+    if (stillOver > 0) {
+      const headBuf = Buffer.from(head, "utf-8");
+      head = headBuf
+        .subarray(0, Math.max(0, headBuf.length - stillOver))
+        .toString("utf-8");
+    }
+    notice = renderNotice(omittedFor(head, tail));
+  }
```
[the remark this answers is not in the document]
Tara-ag: Verified in head `de8ac14` — this finding is **valid and still open**, and it is the only open thread on this PR.

The two-pass reclaim loop (`toolOutputLimits.ts:138-158`) shrinks only `head` and `tail`; it never bounds the rendered `notice`. If a caller supplies a literal notice whose UTF-8 length alone exceeds `maxBytes`, both passes reduce head/tail to empty and line 161 still returns `head + notice + tail` = the notice alone, over budget. The existing 300-byte-literal cases only use budgets ≥ 500, so `notice > maxBytes` is never exercised.

**Kept, non-blocking (MINOR).** Every in-tree caller passes a short notice (the built-in default and ToolsManager's `[... N bytes omitted ...]` render to ~40–80 bytes), so `notice > maxBytes` is an unusual caller bug, not a path this PR's own code hits. It is still reachable via `generateToolOutputPreview`'s other 6+ callers, so worth a follow-up: render the notice, and if `Buffer.byteLength(notice, "utf-8") >= maxBytes` when `head`/`tail` have been reduced to empty, return a fallback preview that is itself bounded by `maxBytes`, plus a test with `maxBytes` smaller than the literal notice. Accepting as-is for merge since it does not block the shape-safe truncation fix.

Tara-ag:
**Merge-blocking CI note (not a code finding)** — the `🔒 Single Commit Policy Validation` check is currently **failing** (runs `103635792542` / `103635783610`) and the PR reports `mergeable_state: blocked` because of it.

The branch head `de8ac14` now carries **3 commits**:
- `811a0eb` — the actual change (fix(core): preserve MCP content envelopes...)
- `807fd393` — a stray `probe` commit
- `de8ac14` — `remove stray probe file`

This repo enforces **one commit per pull request** (the very gate that is now red). The extra probe commits must be squashed into `811a0eb` before this can merge — e.g. `git reset --soft` to the merge-base and re-commit once, then force-push the branch (or rebase+fixup). Note squash-merge would collapse them on merging, but the required gate needs a clean single-commit branch to go green.

The **code review verdict is unchanged: APPROVE**. This is branch hygiene, not a code issue — but it cannot merge in the current state.

Tara-ag:
## APPROVE

Recurring review of head `811a0eb` (single commit — the branch was rebased/collapsed, so the prior cycle's 3-commit merge-block is cleared). The shape-safe MCP truncation change remains correct and CI-wired. All previously-raised findings from earlier cycles are **resolved** and verified line-by-line in the current tree. One **MINOR** edge case remains open — it is already anchored inline (twice) with a concrete suggestion, so I am not re-posting it, only carrying it forward.

|----------|----------|---------|--------|
| MINOR | `src/lib/context/toolOutputLimits.ts:138` | A caller-supplied `notice` whose byte length alone exceeds `maxBytes` still overflows the returned preview — the two-pass reclaim loop shrinks only `head`/`tail`, never the notice. Latent in NeuroLink's own paths (short notices), but reachable via `generateToolOutputPreview`'s other callers (loop guards, `mcpOutputNormalizer`, `redisConversationMemoryManager`). | **Open** — already posted inline (Yama `notice-over-maxbytes` @138 + CodeRabbit @158) with a `maxBytes`-bounded-fallback suggestion. No author reply yet. |

### Prior findings — all resolved in head `811a0eb`
- **MAJOR — envelope byte ceiling not enforced in the aggregate-oversized case**: resolved. `truncateMcpContentEnvelope` enforces the budget on the **serialized envelope as a whole** (fixed/non-text bytes off the top, water-fill across text items, re-measure/tighten up to `MAX_ENVELOPE_TIGHTENING_PASSES`, returns `undefined` → `_truncated` sentinel when the shape can't fit). Verified `ToolsManager.ts:351–446`; pinned by the 20×3KB / 5×40KB / text+image / image-over-budget cases.
- **MINOR — `getCustomTools?.().has()` throw on `undefined` map**: resolved. Call site is `…?.getCustomTools?.()?.has(RETRIEVE_CONTEXT_TOOL_NAME) ?? false`; the suite's "map undefined does not throw" case covers it.
- **MINOR — `retrieve_context`-notice suppression only on the envelope path**: resolved. All four non-envelope `generateToolOutputPreview` call sites (plain string, string `content`, string `data`, generic JSON sentinel) now pass `notice: this.resolveTruncationNotice()`; five new cases pin the registered/unregistered behaviour.
- **CodeRabbit — default/custom notice pushing preview past `maxBytes`**: resolved. `generateToolOutputPreview` reclaims the notice's bytes from tail then head in two passes; pinned across 500/2k/8k/40k budgets.
- **MAJOR — truncation suite not run by CI**: resolved. `test:tools-manager-truncation@3` is in the `extended-suites` `SUITES` list in `.github/workflows/ci.yml`, in lockstep with the `package.json` script.

### Checked and clean this cycle
- **Rule 3 / truncation surfacing**: recognized envelopes always surface (never silently dropped); `isError` and non-text items (image/resource/etc.) pass through untouched; unknown shapes still collapse to the legacy `_truncated` sentinel. Envelope path rides `wrapExecuteWithTruncation` → `truncateToolResult`.
- **Rule 5 (back-compat)**: `ToolConfig.outputTruncationMaxBytes` and `ToolOutputPreviewOptions.notice` are additive optionals; `resolveTruncationMaxBytes()` defaults to the original `DEFAULT_MAX_PREVIEW_BYTES` (51,200). No caller shifts behavior.
- **Blast radius (graph-verified)**: `ToolsManager` imported only by `baseProvider.ts` + the suite; `generateToolOutputPreview`'s 10 callers (anthropic/gemini/openaiCompat loop guards, `redisConversationMemoryManager`, `mcpOutputNormalizer`, `ToolsManager`×2, test) are unaffected — the new `notice`/`maxBytes` options are optional.
- **Rule 15 + determinism**: all-src suite with a declared determinism exception; `eslint.config.js` allow-list, `package.json` script, and CI `SUITES` entry in lockstep.
- **Security**: no credential/log-leakage or injection surface; `transformParamsForLogging` untouched.

**Non-blocking note** (carried from prior cycles): a many-item envelope emits one `[... N bytes omitted ...]` notice per truncated text item; a single envelope-level notice would read more cleanly in the model-visible text. Optional.

No new inline findings this cycle — nothing left to fix beyond the single non-blocking MINOR.

Tara-ag:
## APPROVE

Recurring review of head `2a1b096`. **No open findings** — every previously-raised issue is resolved and verified line-by-line in the current tree, including the last open MINOR (`notice-over-maxbytes`), whose fix landed in this head.

### Findings — none open

| Prior finding | Status in head `2a1b096` |
|---|---|
| MAJOR — envelope byte ceiling not enforced in the aggregate-oversized case | **Resolved** — `truncateMcpContentEnvelope` enforces the budget on the serialized envelope as a whole (fixed/non-text bytes off the top, `waterFillTextBudget` across text items, `MAX_ENVELOPE_TIGHTENING_PASSES` re-measure/tighten, `undefined` → `_truncated` sentinel when the shape can't fit). |
| MINOR — `getCustomTools?.().has()` throw on `undefined` map | **Resolved** — call site is `…?.getCustomTools?.()?.has(RETRIEVE_CONTEXT_TOOL_NAME) ?? false` (`ToolsManager.ts:314-316`). |
| MINOR — `retrieve_context`-notice suppression wired only into the envelope path | **Resolved** — all four non-envelope `generateToolOutputPreview` call sites (plain string, string `content`, string `data`, JSON sentinel) pass `notice: this.resolveTruncationNotice()`. |
| CodeRabbit — custom/literal notice can push preview past `maxBytes` | **Resolved** — two-pass head/tail reclaim gives the notice its bytes back, and a new bounded fallback (`toolOutputLimits.ts:187-198`) returns `takeBytes(notice, maxBytes)` when the notice alone overruns, so every truncated preview genuinely stays ≤ `maxBytes`. |
| MINOR — `notice` longer than `maxBytes` still overflows (the last open thread, `notice-over-maxbytes`) | **Resolved in this head** — the `takeBytes(notice, maxBytes)` fallback now bounds a notice that alone exceeds the ceiling, closing the exact gap I flagged. Author's CodeRabbit reply confirms (`2a1b096`). Not re-posting. |
| MAJOR — truncation suite not run by CI | **Resolved** — `test:tools-manager-truncation@3` added to the `extended-suites` `SUITES` list in `.github/workflows/ci.yml`, in lockstep with the `package.json` script. |
| CodeRabbit — suite not run in CI + docstring coverage | **Resolved / non-blocking** — CI wiring fixed; docstring-coverage warning is advisory and pre-existing, not a merge gate. |

### Checked clean this cycle
- **Rule 3 / truncation surfacing**: recognized MCP envelopes always surface — `isError` and non-text items (image/resource/embedded) pass through untouched; unknown shapes still collapse to the legacy `_truncated` sentinel; truncation is never silent (notice always rendered). Envelope path rides `wrapExecuteWithTruncation` → `truncateToolResult`.
- **Algorithm review**: `waterFillTextBudget` allocates fairly in size order; `serializedByteLength` catches `JSON.stringify` failures; `resolveTruncationMaxBytes` rejects non-finite/non-positive config (NaN/negative) and falls back to `DEFAULT_MAX_PREVIEW_BYTES` (51,200) — no infinite-loop or zero-budget risk in the tightening loop.
- **Rule 5 (back-compat)**: `ToolConfig.outputTruncationMaxBytes` and `ToolOutputPreviewOptions.notice` are additive optionals; `resolveTruncationMaxBytes()` defaults to the original 51,200, so existing consumers see identical behavior. `generateToolOutputPreview`'s other external callers (loop guards, `mcpOutputNormalizer`, `redisConversationMemoryManager`) are unaffected.
- **Rule 15 + determinism**: all-src suite with a declared determinism exception in its header; `eslint.config.js` allow-list, `package.json` script, and CI `SUITES` entry are in lockstep.
- **Blast radius (graph)**: `ToolsManager` imported only by `baseProvider.ts` + the suite; `generateToolOutputPreview`'s 10 callers unaffected (new options mandatory-optional); impacted-flow and risk re-checked (`risk 0.65`, no new flow breakage).
- **Security**: no credential/log-leakage or injection surface; `transformParamsForLogging` untouched.

**Non-blocking note** (carried, optional): a many-item envelope emits one `[... N bytes omitted ...]` notice per truncated text item; a single envelope-level notice would read more cleanly in model-visible text.

The change is correct, complete, bounded, and CI-wired. Approving.

Tara-ag (approved):
Recurring review on the current head `2a1b096` (the single squashed commit — the 3 probe commits were collapsed, so the earlier merge-block is cleared). Verdict **APPROVE**.

All findings raised across every review cycle — including the last open **MINOR** (`notice > maxBytes` overflowing the returned preview in `generateToolOutputPreview`) — are **resolved and verified** in the current tree. CodeRabbit confirms the final case was addressed in this exact head (`✅ Addressed in commit 2a1b096`).

This approval is anchored to the current commit SHA so the verdict is properly reflected on the branch as it stands now.

Tara-ag:
## APPROVE

Recurring review of the shape-safe MCP truncation change at head `4d391beb` (the squashed single commit). **No open findings** — every issue raised across the review cycles is resolved and verified line-by-line in the current tree, and I found **no new issues** in this pass.

### Findings — none open

| Severity | Location | Finding | Status in head `4d391beb` |
|----------|----------|---------|-----------------------------|
| MAJOR | `ToolsManager.ts` | Envelope byte ceiling not enforced in the aggregate-oversized case | **Resolved** — `truncateMcpContentEnvelope` (line 351) enforces the budget on the serialized envelope as a whole (`fixedBytes` off the top, `waterFillTextBudget` across text items, re-measure/tighten over `MAX_ENVELOPE_TIGHTENING_PASSES`, `undefined` → `_truncated` sentinel when the shape can't fit). Pinned by the 20×3KB / 5×40KB / text+image / image-over-budget suite cases. |
| MINOR | `ToolsManager.ts:314-316` | `getCustomTools?.().has()` could throw on an `undefined` map return | **Resolved** — `…?.getCustomTools?.()?.has(RETRIEVE_CONTEXT_TOOL_NAME) ?? false`; suite has a "map undefined does not throw" case. |
| MINOR | `ToolsManager.ts` | `retrieve_context`-notice suppression wired only into the envelope path | **Resolved** — all four non-envelope branches (plain string, string `content`, string `data`, generic JSON sentinel) pass `notice: this.resolveTruncationNotice()`; five pinning cases (4 unregistered + 1 positive registered). |
| MINOR | `toolOutputLimits.ts` | A `notice` whose byte length alone exceeds `maxBytes` still overflowed the preview | **Resolved** — two-pass head/tail reclaim reserves the notice's bytes, and the final `takeBytes(notice, maxBytes)` fallback bounds a notice that alone exceeds the ceiling so every truncated preview stays ≤ `maxBytes`. Verified in `generateToolOutputPreview` (`toolOutputLimits.ts:79-205`); suite covers budgets with notices larger than `maxBytes`. |
| MAJOR | `.github/workflows/ci.yml` | Truncation suite wired into no CI-invoked aggregate | **Resolved** — `test:tools-manager-truncation@3` present in the `extended-suites` `SUITES` list, in lockstep with the `package.json` script. |

### Checked and clean (this pass)
- **Algorithm review**: `waterFillTextBudget` allocates fairly in size order and never exceeds an item's own size; `serializedByteLength` guards `JSON.stringify` failures (returns `undefined`); the tightening loop is bounded (`MAX_ENVELOPE_TIGHTENING_PASSES = 6`, `textBudget <= 0` break) — no infinite-loop or zero-budget risk. `maxBytes` guards reject non-finite/non-positive config and fall back to `DEFAULT_MAX_PREVIEW_BYTES` (51,200).
- **Rule 3 / truncation surfacing**: recognized MCP envelopes are preserved (`isError`, non-text items, other top-level fields) and truncation is never silent; unknown shapes keep the legacy `_truncated` sentinel for compat.
- **Rule 5 (back-compat)**: `ToolConfig.outputTruncationMaxBytes` and `ToolOutputPreviewOptions.notice` are additive optionals; defaults unchanged (51,200), so existing consumers see identical behavior.
- **Blast radius (graph)**: `ToolsManager` imported only by `baseProvider.ts` + the suite; `generateToolOutputPreview`'s external callers (anthropic/gemini/openaiCompat loop guards, `redisConversationMemoryManager`, `mcpOutputNormalizer`) are unaffected because the new `notice`/`maxBytes` options are optional. Impacted flows (`addExternalMCPServer`, `planReclaim`, `setupToolExecutor`, …) ride through `wrapExecuteWithTruncation → truncateToolResult` unchanged apart from the new envelope path.
- **Rule 15 + determinism**: all-src suite with a declared determinism exception in its header; `eslint.config.js` allow-list, `package.json` script, and CI `SUITES` entry are in lockstep.
- **Security**: no credential/log leakage or injection surface; `transformParamsForLogging` untouched.

**Reconciliation note**: the local graph was built at `d00ade34` while GitHub reports the PR head as `4d391beb`; I verified the actual file contents at `4d391beb` directly (ToolsManager.ts, toolOutputLimits.ts, package.json, ci.yml) — the code is content-identical to the state previously approved, so the verdict carries.

**Non-blocking note** (carried, optional): a many-item envelope emits one `[... N bytes omitted ...]` notice per truncated text item; a single envelope-level omission notice would read more cleanly in model-visible text.

The change is correct, complete, bounded and CI-wired. Approving.

Tara-ag (approved):
Recurring review anchored to the current head `4d391beb` (the squashed single commit — the earlier 3- and then multi-commit branch states were collapsed, so the single-commit merge gate is clear). Verdict **APPROVE**.

All findings raised across every review cycle are resolved and verified line-by-line in this tree; no new findings. See the canonical summary comment for the full findings table (envelope byte ceiling, `getCustomTools?.().has()` throw, `retrieve_context`-notice threading, `notice > maxBytes` overflow, CI wiring). This approval supersedes the prior approving review anchored to `2a1b096` so the verdict is recorded against the branch as it stands now.

murdore:
Follow-up moved to **https://github.com/juspay/neurolink/pull/1728** so this can merge.

The one item still outstanding here was CodeRabbit's performance nitpick on `src/lib/core/modules/ToolsManager.ts`: `logger.debug` fires inside the per-item `forEach` before the candidate pass is checked against `maxBytes`, so up to `MAX_ENVELOPE_TIGHTENING_PASSES` rejected passes each pay for template-string building and `Buffer.byteLength` — whether or not debug logging is enabled.

It never became a resolvable thread: it was inside a collapsed "Nitpick comments" block in a review body, so it could never show as unresolved and nothing chased it. Verified still present verbatim at head `2a1b0966`.

Fixed in #1728 — sizes recorded during the pass, logged only by the pass that survives, behind `logger.shouldLog("debug")`. `continuous-test-suite-tools-manager-truncation.ts` passes 25/25 unchanged.

PR #1624, opened 2026-09-02
chore(catalog): record why Cloudflare's vision model stays vision:false after licence acceptance

murdore:
## What

Catalog-only change for `@cf/meta/llama-3.2-11b-vision-instruct`: the entry's description and the provider's `evidence.liveMatrix` now state the model's real limitation instead of the licence gate that no longer applies. No code, no flag change.

## Why the flag stays `vision: false`

- The Meta Community License was accepted on the account on 2026-09-02 (Cloudflare: "Thank you for agreeing to this model's terms. You may now use the model."). Plain chat on the model returns 200 on both the native `/ai/run` endpoint and the OpenAI-compatible endpoint.
- The model does read images, but only through the native `/ai/run` call with the legacy `prompt` + `image` (bytes) shape: a 64×64 blue test image is described as "Blue.".
- Cloudflare's OpenAI-compatible endpoint, which is the route `ConfiguredOpenAICompatProvider` drives, carries no image for this model: an OpenAI image content part is rejected with 400 ("Unable to add image when there are no user-supplied nor system-supplied messages" / "Internal Server Error"), and a top-level `image` field is ignored ("There is no image").

Flipping the flag would promise something the framework cannot deliver on this route, so the catalog states the fact and the evidence block records the probes.

## Verification

- `pnpm run codegen:catalog -- --check` — clean (no enum change).
- `pnpm run verify:provider-onboarding` — 2/2.
- Pre-push hook: build + `providers-mocked` + `provider-structure`.

## Follow-up (not in this PR)

A Cloudflare-specific image path (native `/ai/run` `prompt`+`image` for single-turn, tool-less vision requests) could be expressed as a catalog quirk if single-turn image description on Cloudflare is worth having; it would not cover conversations or tools.

## Summary by CodeRabbit

- **Documentation**
  - Updated Cloudflare model catalog information to reflect the latest license acceptance verification.
  - Confirmed that standard chat requests are successful for the model.
  - Clarified that image input remains unsupported through the OpenAI-compatible endpoint.
  - Refreshed the verification date and supporting live-matrix details.

### DIFF
--- a/src/lib/providers/catalog/cloudflare.json
+++ b/src/lib/providers/catalog/cloudflare.json
@@ -47,7 +47,7 @@
         "contextWindow": 24000,
         "vision": false,
         "status": "production",
-        "description": "Llama 3.2 11B Vision Instruct — gated: Cloudflare returns 403 until the model's Community License is accepted once on the account, so it is unusable here and its vision flag is false despite being a vision model"
+        "description": "Llama 3.2 11B Vision Instruct — Meta Community License accepted on the account 2026-09-02, but Cloudflare's OpenAI-compatible endpoint accepts no image input for it: an image content part is rejected (400) and a top-level image is ignored ('There is no image'). The model reads images only through the native /ai/run prompt+image shape, which this OpenAI-compatible provider does not speak, so vision stays false here"
       },
       "@cf/mistral/mistral-7b-instruct-v0.2": {
         "enumMember": "MISTRAL_7B_INSTRUCT_V0_2",
@@ -114,8 +114,8 @@
       "method": "real chat call per id against the live account"
     },
     "liveMatrix": {
-      "date": "2026-08-30",
-      "result": "Live tool round-trip verified on @cf/meta/llama-3.1-8b-instruct-fast, both non-streaming and streaming, once messageContentFormat normalization was applied; before it the follow-up turn was rejected with HTTP 400 (content sent as array/null). Of the three ids that failed, only two are actually gone: mistral-7b-instruct-v0.2 ('No such model') and qwen1.5-14b-chat-awq ('deprecated on 2025-10-01'). llama-3.2-11b-vision-instruct is alive but gated behind a one-time Community License acceptance on the account, so it answers 403 and is left production with vision false rather than retired."
+      "date": "2026-09-02",
+      "result": "Live tool round-trip verified on @cf/meta/llama-3.1-8b-instruct-fast, both non-streaming and streaming, once messageContentFormat normalization was applied; before it the follow-up turn was rejected with HTTP 400 (content sent as array/null). Of the three ids that failed, only two are actually gone: mistral-7b-instruct-v0.2 ('No such model') and qwen1.5-14b-chat-awq ('deprecated on 2025-10-01'). llama-3.2-11b-vision-instruct is alive but gated behind a one-time Community License acceptance on the account, so it answers 403 and is left production with vision false rather than retired. Update 2026-09-02: the Community License was accepted on the account; plain chat on llama-3.2-11b-vision-instruct now returns 200 on both endpoints, and the native /ai/run prompt+image call describes a 64x64 test image correctly. The OpenAI-compatible endpoint still cannot carry an image (content-part array → 400, top-level image ignored), so the vision flag remains false for this provider."
     },
     "addedInPR": "https://github.com/juspay/neurolink/pull/1587"
   },


src/lib/providers/catalog/cloudflare.json
```diff
@@ -114,8 +114,8 @@
...
+      "date": "2026-09-02",
+      "result": "Live tool round-trip verified on @cf/meta/llama-3.1-8b-instruct-fast, both non-streaming and streaming, once messageContentFormat normalization was applied; before it the follow-up turn was rejected with HTTP 400 (content sent as array/null). Of the three ids that failed, only two are actually gone: mistral-7b-instruct-v0.2 ('No such model') and qwen1.5-14b-chat-awq ('deprecated on 2025-10-01'). llama-3.2-11b-vision-instruct is alive but gated behind a one-time Community License acceptance on the account, so it answers 403 and is left production with vision false rather than retired. Update 2026-09-02: the Community License was accepted on the account; plain chat on llama-3.2-11b-vision-instruct now returns 200 on both endpoints, and the native /ai/run prompt+image call describes a 64x64 test image correctly. The OpenAI-compatible endpoint still cannot carry an image (content-part array → 400, top-level image ignored), so the vision flag remains false for this provider."
```
Tara-ag, line 118: **Non-blocking** — acknowledging the existing CodeRabbit concern (the pre-acceptance 403 sentence before "Update 2026-09-02:").

For a plain data/evidence field this is acceptable as-written: `liveMatrix` is a dated evidence log, and the new `Update 2026-09-02:` sentence makes the **current** status unambiguous (200 on both endpoints; vision flag still false). Keeping the original 403 observation verbatim preserves the audit trail of what was observed before license acceptance, which is the point of this evidence block.

Optional, for scan-readability only: if you want the earlier clause to read unambiguously as historical at a glance, prefix it with a date marker, e.g.:

```jsonc
"As of 2026-08-30, llama-3.2-11b-vision-instruct is alive but gated behind a one-time Community License acceptance ... sponsors 403 ... Update 2026-09-02: ..."
```

Not required — the current text is already correct and complete.

Tara-ag (commented):
## Review summary — PR #1624 `fix/cloudflare-vision-evidence`

**Scope:** single data change to `src/lib/providers/catalog/cloudflare.json` — appends a dated `Update 2026-09-02:` evidence entry to `liveMatrix.result` for the `llama-3.2-11b-vision-instruct` model, recording that the Community License was accepted and plain chat now returns 200 on both endpoints while the vision flag correctly remains `false` (the OpenAI-compatible endpoint still cannot carry an image).

**Context / blast radius:** This is a documentation/data edit — no schema fields change (`liveMatrix` shape, the model's `vision: false` and `status: production` fields are untouched), no code paths, callers, or execution flows are affected. I verified there is no functional impact; the graph analysis tools are not warranted here.

**Existing finding (CodeRabbit, line 118, "major" — stale 403 present-tense wording):**
- The PR's diff **is** the fix for this: before this PR, `liveMatrix.result` contained *only* the 403 claim with no update; this PR adds the "Update 2026-09-02:" paragraph that makes current status unambiguous.
- Because this is a dated evidence log, retaining the original 403 observation verbatim (rather than deleting it) is the correct pattern — it preserves the audit trail of the pre-acceptance behavior.
- I therefore consider the substantive concern **addressed**. I left one **non-blocking** inline suggestion (optional date-marker on the earlier clause) purely for scan-readability.

**Assessment:** No blocking issues. The evidence is now complete, accurate, and current, and correctly keeps `vision: false`. Approving review guidance: this is good, well-documented work.

One note for the author (not a change request): the pre-acceptance clause is kept in present tense ("it answers 403"). If you'd ever like it read as explicit history, prefix `As of 2026-08-30, ...` — entirely optional.

Tara-ag:
### Review summary — `fix/cloudflare-vision-evidence`

Reviewed the full diff of `src/lib/providers/catalog/cloudflare.json` (both hunks) against the surrounding context and the live account state recorded in the evidence log.

**Verdict: no blocking issues.** The change is an internally consistent, append-only update to the dated evidence record, and does not alter any catalog field other than the two narrative/description strings.

**What I checked**
- **Description (line 50)** now reflects the post-acceptance reality: the Community License was accepted `2026-09-02`, plain chat returns `200`, but the OpenAI-compatible endpoint still rejects image input (content-part array → `400`, top-level image ignored with `'There is no image'`). This correctly explains why `vision` stays `false` for this provider (it doesn't speak the native `/ai/run` prompt+image shape). The `vision: false` flag and the description are now mutually consistent.
- **liveMatrix (lines 117–118)** keeps the original `2026-08-30` observation verbatim (preserving the audit trail of the pre-acceptance 403) and tacks on a clearly date-prefixed `Update 2026-09-02:` block. The **current** status is unambiguous — matching what the description now states.

**Recurring-review disposition**
- The earlier CodeRabbit concern (stale pre-acceptance 403 in `liveMatrix.result`) is **already resolved** on this revision's thread. The revised copy retains the historical observation while making the current state explicit via the dated `Update` marker — the exact justification laid out in the prior acknowledgment. **Not reposting** per the recurring-review protocol; the acknowledgment thread has been resolved to close the loop.

No new findings. Nothing here requires author action — thanks for tightening the description so it no longer reads as a stale "gated/403" claim.

Tara-ag (approved):
Approved. The catalog change is internally consistent: the description and liveMatrix evidence now both record the 2026-09-02 license acceptance while explaining exactly why `vision` stays `false` (OpenAI-compatible endpoint carries no image input; the model only reads images via the native `/ai/run` shape this provider does not speak). No blocking findings.

PR #1627, opened 2026-09-03
fix(types): stop shipping declarations that reference stripped types

murdore:
## The bug

`tsconfig.json` sets `stripInternal: true`. TypeScript honours it by deleting `@internal` declarations from the emitted `.d.ts` — but it does **not** check whether anything still refers to them, and it does **not** rewrite the imports that name them.

Seven knowledge types and `GitToolRuntimeSettings` were tagged while public declarations still used them in their signatures, so the published package ships `.d.ts` files that cannot compile:

```
dist/knowledge/context.d.ts(11,15): error TS2305: Module '"../types/index.js"' has no exported member 'KnowledgeAssembledContext'.
dist/knowledge/context.d.ts(11,66): error TS2724: '"../types/index.js"' has no exported member named 'KnowledgeSelection'.
dist/knowledge/resolve.d.ts(1,55): error TS2305: ... 'KnowledgeNormalizeOptions'.
dist/knowledge/resolve.d.ts(1,82): error TS2305: ... 'KnowledgeNormalizeResult'.
dist/knowledge/retrieval.d.ts(10,39): error TS2305: ... 'KnowledgeResolvedRetrieval'.
dist/agent/gitTools.d.ts(25,30): error TS2305: ... 'GitToolRuntimeSettings'.
```

Consumers with `skipLibCheck: true` — the common default — never see this. Everyone else gets errors from inside `node_modules` with nothing actionable in them.

## The fix

The tags were wrong, not the code. Every one of these types is named by the emitted signature of a public runtime export — `assembleKnowledgeContext`, `retrieve`, `resolveEntry`, `manifestToSources`, `normalizeAndValidate`, `configureGitTools` — which makes them public by construction. Three knowledge types that never reach an emitted signature keep their tag.

## The guard

`check:dts` typechecks all 1027 shipped declaration files with `skipLibCheck` off and fails on any unresolved reference.

It is not decoration: it caught the first version of this fix. The explanatory comment I wrote on `GitToolRuntimeSettings` named the tag literally, which re-triggered the strip — the tag is matched as plain text anywhere in the doc comment, so it cannot even be named there to explain itself.

## Verification

| check | result |
| --- | --- |
| `check:dts` | green over 1027 declaration files |
| consumer typecheck, `skipLibCheck: false` | **6 errors → 0** |
| `check`, `check:ci-scripts` | clean |
| `validate:all`, `lint` | 0 errors |
| build + publint | pass |

Pre-commit hook ran and passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected published type declarations so referenced public types remain available to consumers.

* **Quality Improvements**
  * Added automated validation to detect unresolved references in generated declaration files.
  * CI now typechecks the declaration validation script to help prevent broken package types from shipping.


Tara-ag:
**APPROVE** — the change is correct and well-scoped: it stops shipping `.d.ts` files that reference types stripped by `stripInternal`, and adds a genuinely valuable `check:dts` CI guard. The code is clean; the only open items are non-blocking robustness polish on the new guard.

## Verdict rationale
The bug being fixed is real (emitted declarations referenced types removed by `stripInternal`). Removing `@internal` from the four public-referenced types in `src/lib/types/gitTools.ts` and `src/lib/types/knowledge.ts` is the correct, minimal fix — these types are public by construction (named by runtime exports), so exposing them is the right contract, not an accidental API expansion. The new `scripts/check-shipped-types.ts` (with `skipLibCheck:false` and a strict diagnostic filter) is a solid guard against regression. No correctness, security, or SDK-backward-compat (Rule 5) issue.

## Findings

|---|---|---|
| MINOR | `scripts/check-shipped-types.ts:49-53` | `UNRESOLVED_REFERENCE_CODES` omits TS2694 (`namespace has no exported member`), the namespace-relative analogue of the already-listed TS2724, and TS2304 (`cannot find name`). The guard's whole purpose is to not miss unresolved references, so these belong in the set. Non-blocking — current output is green. |
| MINOR | `scripts/check-shipped-types.ts:82` | `MINIMUM_DECLARATION_FILES = 100` is a weak floor. `collectDeclarationFiles` walks the whole `dist/` tree, so *which* files are checked isn't gated by the threshold; the risk is only a partially-built `dist/` passing. Consider comparing the discovered set against the expected post-clean-build set. Non-blocking given the author's verified output. |
| MINOR | `scripts/check-shipped-types.ts` / `package.json` | The `check:dts` script must be wired into the actual publish/release gate (not just a standalone npm script) for the guard to protect users. |

## Checked and found clean
- Blast radius: contained to the two `src/lib/types` files and the new CI script. No provider/MCP/streaming/route code touched.
- Rule 1 / Rule 3 / Rule 4 / Rule 6 / Rule 15: not implicated — no dynamic-import, Gemini, CLI/SDK, error-formatting, or test-graph concerns.
- The `@internal` tag is only mentioned inside a `//` line comment and the explanatory doc text; it does not appear as an attached JSDoc tag on any surviving declaration, so `stripInternal` will not re-strip the now-public types.
- Compiles green with `skipLibCheck:false` over all 1027 `dist/` declaration files; exit-code handling is correct.
- Conventional Commit with required scope + exactly one commit (verified by the CI policy check).

_Non-blocking MINOR items only. No changes required to merge._

Tara-ag (approved):
APPROVE — correct, well-scoped fix; stops shipping declarations that reference stripInternal-removed types and adds a solid check:dts guard. Only non-blocking MINOR items remain; see the summary comment.

PR #1628, opened 2026-09-03
refactor(core): remove the Vercel AI SDK and drive every provider natively

murdore:
Drops all five packages — `ai`, `@ai-sdk/provider`, `@ai-sdk/provider-utils`, `@ai-sdk/anthropic`, `@ai-sdk/openai`, `@ai-sdk/mistral` — from production and from the public type surface.

## What the SDK was still doing

Providers already handed it hand-built V3-shaped models whose `doGenerate` did native HTTP. Only the loop around them came from the package.

| replaced | with |
| --- | --- |
| `generateText`'s multi-step tool loop | `src/lib/core/nativeGenerateLoop.ts` |
| `ai` / `@ai-sdk/provider` type declarations | `src/lib/types/aiCompat.ts` |
| `tool()`, `jsonSchema()`, `stepCountIs()` | exported from `@juspay/neurolink` |
| `wrapLanguageModel` | `src/lib/middleware/wrapLanguageModel.ts` |
| SDK error classes | `src/lib/utils/generationErrors.ts`, upstream marker symbols preserved |
| audio transcription, browser provider factories | native |

### The loop iterates `doGenerate`, not the streaming path

An earlier attempt looped over the SSE adapter and silently sent `stream: true` where the SDK sent plain JSON. **The live matrix could not see this** — every reachable provider supports streaming. `test:providers-mocked` failed 17 contracts. The live matrix proves behaviour; the mocked gate proves the wire. Both are required.

### `any` in `aiCompat.ts` is load-bearing

Upstream's generic defaults are `any`, not `unknown`, and that is what makes `execute` optional through `NeverOptional` — the only reason a bare `Record<string, Tool>` satisfies `ToolSet`. Declaring them with `unknown` produced 66 errors. It appears in one file, behind a scoped exemption explaining why.

### `tool()` is exported

Consumers imported it from `ai` to build tools for `generate()`. Dropping the dependency without shipping an equivalent would have stranded them.

## Four things the native paths silently lost

The native loops bypass `BaseProvider.executeGeneration`, and with it everything that lived around it. Each was lost silently until something went looking:

| lost | effect | found by |
| --- | --- | --- |
| turn timeout | `timeout` / `turnTimeoutMs` stopped existing for Anthropic, the OpenAI-compatible family, SageMaker | `test:anthropic-loop-characterization` — 1s timeout against a 3s stand-in returned the late response |
| model middleware | a caller's `wrapGenerate` ran **0** times on anthropic, groq, mistral, deepseek | live probe |
| TTS synthesis | `tts.useAiResponse` produced no audio | code audit |
| performance metrics | OTEL saw no usage for any native provider | code audit |

Only `onFinish` worked, because it had been special-cased by hand and nothing else had.

Rather than patch each site, the shared pieces now live on `BaseProvider`: `withTurnTimeout` owns the timeout contract for every generate path, `finalizeNativeGenerate` owns the post-call sequence. `fireGenerateOnFinish` was duplicated across four providers and is hoisted; the native paths no longer call it directly, because wrapping the model means the lifecycle middleware fires it and doing both would deliver every callback twice.

**Still bypassing model middleware:** Vertex, Google AI Studio and Bedrock. Pre-existing and not fixable the same way — their `getAISDKModel()` throws, so there is no model object to wrap. Needs a V3 shim per provider.

## Verification

| gate | result |
| --- | --- |
| live matrix | **30 passed / 10 failed** vs 29/11 baseline — one improvement (`groq/stream-tools`) |
| provider contract gate | 77/77 |
| characterization suites | 5/5 (anthropic was red) |
| error classifier, provider-structure, model-manifests, provider-onboarding | pass |
| `check:deps`, `validate:all`, lint | pass, 0 errors |
| typecheck | 4836 files, 0 errors |
| bundle | 4.81 MB → **4.41 MB** |

Consumer typecheck with both packages hidden from `node_modules`: **zero** unresolved SDK references.

Middleware probe: `wrapGenerate` 0 → 1 and `onFinish` exactly 1 on all four live native providers.

## Note

`GenerationHandler.callGenerateText` throws rather than being deleted, so a provider added without a native `generate()` fails loudly instead of quietly reintroducing the dependency.

Pre-commit hook ran and passed.

## Summary by CodeRabbit

- **New Features**
  - Added browser-ready Anthropic, OpenAI, and Mistral provider helpers with lazy model loading.
  - Added native multi-step generation with tool calling for supported providers.
  - Added built-in tool, JSON schema, and step-count helpers.
  - Added native OpenAI audio transcription support.

- **Documentation**
  - Updated middleware, streaming, provider, and guardrails examples to use NeuroLink APIs.
  - Added plans documenting the remaining SDK migration work.

- **Bug Fixes**
  - Improved timeout, retry, fallback, usage tracking, and error handling during generation.


murdore:
@coderabbitai full review

PR #1629, opened 2026-09-03
fix(proxy): recognize npm global bin entrypoints

murdore:

### DIFF
--- a/src/lib/proxy/globalInstaller.ts
+++ b/src/lib/proxy/globalInstaller.ts
@@ -86,7 +86,14 @@ function probeInstaller(
     }
 
     base.installable = true;
-    base.matchesCurrentInstall = isPathInside(entryScript, base.globalRoot);
+    // Global npm packages install their entrypoints under <prefix>/bin while
+    // their package contents live in <prefix>/lib/node_modules. A running
+    // process can retain its argv path after that shim has disappeared, so
+    // retain the lexical bin-directory check as well as the resolved package
+    // root check.
+    base.matchesCurrentInstall =
+      isPathInside(entryScript, base.globalRoot) ||
+      isPathInside(entryScript, base.globalBinDir);
     return base;
   } catch (error) {
     base.reason = error instanceof Error ? error.message : String(error);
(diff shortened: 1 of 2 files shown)


Tara-ag (approved):
The widening of `matchesCurrentInstall` to also accept an entry script under the package manager's global bin directory is correct, well-scoped, and backed by a deterministic regression test.

Tara-ag:
**APPROVE** — Widening `matchesCurrentInstall` to also accept an entry script under the package-manager's global bin directory correctly fixes updater recognition of an npm/pnpm global install after its shim is removed, with a matching deterministic test.

| --- | --- | --- |
| — | — | No blocking or MINOR findings |

**What was checked and found clean**

- **Logic/correctness:** The OR-widening (`isPathInside(entryScript, globalRoot) || isPathInside(entryScript, globalBinDir)`) is internally consistent. In both `isPathInside` branches the comparison either realpaths both operands or falls back to lexical `resolve()` on both, so there is no realpath/lexical asymmetry when the shim is removed. `globalBinDir` was already computed and guarded (`!base.globalBinDir || !writableDirectory(...)`) in the base, so there is no undefined/empty access added by this diff.
- **Blast radius:** `resolveGlobalInstaller` only broadens which candidate is treated as matching; no call sites of `matchesCurrentInstall` rely on the narrower false outcome in a way this breaks.
- **Test:** The new case in `continuous-test-suite-bugfixes.ts` correctly builds `prefix/lib/node_modules` (globalRoot) and `prefix/bin` (globalBinDir), drives a removed shim (`prefix/bin/neurolink`, does not exist), and asserts `tried[0].matchesCurrentInstall === true` with `installer.bin === npm`. The `execFileSync` mock throws for any `bin !== npm`, so `resolveFromPath`/path candidates resolve to `undefined` and `tried[0]` stays the env-configured npm — the assertion is meaningful.
- **Design note (non-blocking):** The bin-dir check is intentionally broad — any process whose `argv[1]` lives under the global bin dir is attributed to that package manager. That still installs into the correct global root, so it is conservative-correct, not a bug.

**Standards compliance:** No violations of CLAUDE.md rules 1, 3, 4, 5, 6, or 15. No hardcoded secrets, no logging of sensitive params, no injection/eval, no unsanitized user input on a hot path. Conventional Commits scope present; single commit per PR.

PR #1632, opened 2026-09-04
fix(scripts): close two holes in the shipped-declaration guard

murdore:
Follow-up to #1627, addressing both CodeRabbit findings on it. The guard's entire job is to not report a false green, and it had two ways to do exactly that.

## 1. Two diagnostic codes were filtered out

`UNRESOLVED_REFERENCE_CODES` covered TS2305, TS2307 and TS2724 — the shapes an unresolved **module import** takes. The same hole reached through a bare name or a namespace member was silently ignored:

| code | meaning | was caught |
| --- | --- | --- |
| TS2304 | `Cannot find name 'X'` | ❌ |
| TS2694 | `Namespace 'X' has no exported member 'Y'` | ❌ |

A shipped declaration carrying either was reported clean.

## 2. The floor was not a completeness check

`files.length < MINIMUM_DECLARATION_FILES` accepts any build with 100+ declarations. A partial build of anywhere from 100 to 1026 files cleared it, and that subset was then validated as though it were the whole package — passing while the missing declarations were never examined.

The floor stays as a first gate, but the property that matters is now checked directly: every `.d.ts` named by `package.json`'s `exports` map, plus the root `types` entry, must be present.

**Wildcards need care.** Subpaths like `./dist/adapters/*.d.ts` name a family, not a file. For those, one match is enough — treating the pattern as a literal path reports every wildcard subpath as missing on a perfectly good build, which is exactly how my first version of this check failed.

## Verification

Each path broken on purpose, then restored:

| broken | reported |
| --- | --- |
| planted TS2304 declaration | ✅ |
| planted TS2694 declaration | ✅ |
| hidden `dist/rag/index.d.ts` | ✅ absent entry point |
| hidden `dist/adapters/` | ✅ absent wildcard family |

All four restore to green. `check`, `check:ci-scripts`, `lint`, build all pass.

## Summary by CodeRabbit

- **Chores**
  - Improved release validation to detect missing published TypeScript declaration files.
  - Expanded checks for unresolved TypeScript references and invalid type entry points.
  - Maintained existing declaration-count and diagnostic reporting checks.

### DIFF
--- a/scripts/check-shipped-types.ts
+++ b/scripts/check-shipped-types.ts
@@ -38,26 +38,46 @@
  *  1  At least one unresolved reference, or `dist/` has not been built.
  */
 
-import { readdirSync, statSync } from "node:fs";
+import { existsSync, readdirSync, readFileSync, statSync } from "node:fs";
 import { join, relative } from "node:path";
 import ts from "typescript";
 
 const ROOT = process.cwd();
 const DIST = join(ROOT, "dist");
 
-/** Errors that mean "this declaration names something that isn't there". */
+/**
+ * Errors that mean "this declaration names something that isn't there".
+ *
+ * All five shapes a stripped or missing declaration can take. The first three
+ * cover a module-level import; 2304 and 2694 cover the same hole reached
+ * through a bare name or a namespace, which an earlier version of this list
+ * omitted — a shipped declaration with either would have been reported clean
+ * by a checker whose entire purpose is to not do that.
+ */
 const UNRESOLVED_REFERENCE_CODES = new Set([
+  2304, // Cannot find name 'X'.
   2305, // Module '"X"' has no exported member 'Y'.
   2307, // Cannot find module 'X' or its corresponding type declarations.
+  2694, // Namespace 'X' has no exported member 'Y'.
   2724, // 'X' has no exported member named 'Y'. Did you mean 'Z'?
 ]);
 
 /**
- * A build must exist and be substantial before the result means anything.
- * An empty or missing `dist/` would otherwise report a clean pass.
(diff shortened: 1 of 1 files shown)


Tara-ag:
Review summary for `juspay/neurolink#1632` — branch `fix/shipped-types-guard-coverage`.

**Change intent (verified):** removes the `if (error.code === "ENOENT") return` silent-skip in `collectDeclarationFiles`, so a published `types`/`typesVersions` target that doesn't resolve to a readable directory on disk is now surfaced as a failure instead of being skipped. Strictly a guard-hardening change; blast radius is confined to `scripts/check-shipped-types.ts` (the `action-dist/index.js` impact in the graph is a 34k-node compiled artifact in the same commit range, not a runtime caller). Affected flow: `main` only.

**Findings (3 inline):**
1. `publishedEntryExists` (line 111) — now infers "missing" purely from an empty file list; a rethrown non-ENOENT error from `collectDeclarationFiles` will crash the guard instead of being classified. Suggestion posted to keep the two failure modes distinct.
2. `main` (line 216) — `process.exit(1)` inside `main()` is the reason `publishedEntryExists`, `collectPublishedTypeEntries`, `visit` and `main` are all untestable. Suggestion posted to return a result and exit only at the entry point.

**Blocking before merge:**
- **Missing changeset.** This PR hardens a packaging guard and must ship a changeset (this repo enforces them on `src/**` and tooling changes; guard changes affect the published artefact contract). Please run `pnpm changeset` and commit the file. Happy to re-review on the next push.

Once the changeset is in, this is approvable.

Tara-ag:
check

PR #1634, opened 2026-09-04
fix(core): honor disableInternalFallback on generate() and assert matrix structured output

murdore:
## Summary

Two follow-ups recorded during the provider campaign (#1625, #1631), both about `generate()` options that looked honoured but were not.

### 1. `disableInternalFallback` never reached the provider on `generate()`

`StreamOptions` declared the flag and `stream()` honoured it (BaseProvider's `withStreamModelFallback` / `retryStreamWithFallbackModel`, the no-output retry in `neurolink.ts`). `GenerateOptions` did not declare it, and `buildGenerateTextOptions` rebuilds the provider options from an explicit field list, so the flag was dropped before `BaseProvider.generate()` could read it. A caller that owns fallback order could turn the catalog model fallback off for streams but not for one-shot calls.

- `GenerateOptions.disableInternalFallback` and `TextGenerationOptions.disableInternalFallback` are declared (docs mirror the stream flag; the REPL options schema gets the matching entry, which the `Record<keyof TextGenerationOptions, …>` type requires).
- `buildGenerateTextOptions` maps it through.
- `directProviderGeneration` bounds its static provider-priority walk to the first candidate when the flag is set. An explicit `provider` was already a one-element list; this only changes the `"auto"` and orchestrated-preference walks. A configured `ModelPool`, `providerFallback` and `modelChain` are the caller's own fallback and are untouched, on both paths.
- `BaseProvider.generate()` drops the `"disableInternalFallback" in options` workaround now that the field is typed. The three provider-level reads (openai-compat base, Anthropic, SageMaker) are left as they are; another session owns the openai-compat stream path.
- `docs/features/provider-fallback.md` documents the generate-side behaviour.

### 2. The provider matrix's structured-output cell was testing nothing

`test/continuous-test-suite-provider-matrix.ts` passed `structuredOutput: { schema }`, which has never been a `GenerateOptions` field; the `as never` cast hid it and `generate()` ignored it, so the cell only checked for non-empty text. It now passes the honoured top-level `schema` and asserts that `structuredData` is present and satisfies the schema, with the documented `jsonTruncated` exemption.

## Proof

Mock-server e2e (`test/continuous-test-suite-error-classification-e2e.ts`, Mancer block, cases 6 and 7):

- default: a model the vendor rejects with `Unknown Model` is retried on the catalog's first fallback (`deepseek-v4-flash`) and the call resolves;
- `disableInternalFallback: true`: the classified `InvalidModelError` naming the requested model surfaces after **exactly one** request.

On the pre-change build the request-count assertion fails with three requests (ghost model, then both catalog fallbacks), so the test is a real regression guard. Break-one ritual: with the count deliberately changed, the case reports `✗` and the suite exits non-zero.

Credential-free proof, rebased onto 69c749b94 (which wires this suite into CI and adds the Vertex placeholders): run from an empty directory with no `.env`, an empty `HOME` and a node-only `PATH`, the suite is **144/144** with all four Mancer cases passing. A run that keeps `HOME` or the repo cwd is not credential-free, because the SDK loads the working directory's `.env` itself.

Live: the rewritten matrix cell against Google AI (`gemini-2.5-flash`): `✓ [google-ai] structured output`, 7/7 cells passed (run twice: before and after the review fixes). With the schema assertion deliberately inverted the cell reports `✗`, not `⊘`: `✗ [google-ai] structured output`, Failed: 1, exit code 1.

## Gates

| Gate | Result |
| --- | --- |
| `pnpm run check` | 0 errors (svelte-check + tsc --strict) |
| `pnpm run lint` | 0 errors; the 58 warnings are the pre-existing method-length warnings |
| `pnpm run build` | green (SDK + CLI + pack) |
| `test:error-classification-e2e` | 144/144 after the rebase onto 69c749b94 (141/144 before it: the 3 failures were the release regressions #1635 fixed) |
| `test:provider-structure` | 3/3 |
| docs:api regenerated | yes, 22 generated pages refreshed; `disableInternalFallback` appears on `GenerateOptions` and `TextGenerationOptions` |
| adversarial review (4 lenses, 2 skeptics per finding) | 6 findings, all addressed (below) |

Review findings and what changed:

- **Matrix cell hard-failed on the SDK's documented pure-prose outcome** (confirmed). `generate({ schema })` leaves `structuredData` undefined when the model returns no JSON at all, distinct from truncation. The cell now retries once more when that happens and then fails with a precise message. It deliberately stays a failure rather than a skip: the row's `structuredOutput: true` is the claim under test, and prose after two attempts means the claim is false for that model. The message was checked against the harness's expected-provider-error patterns and is not downgraded.
- **`TextGenerationOptions` JSDoc named only one of the two walks the flag bounds** (confirmed). Reworded to name both.
- **`docs/api` not regenerated** (confirmed). Regenerated as the last step before the commit.
- **Code comments cited the Claude proxy as a `generate()` caller** (confirmed). The proxy only calls `stream()`; the comments and the type doc now say "a router that retries on its own" and mention the proxy only for streams.
- **Stale comment in `continuous-test-suite-vertex-loop-characterization.ts`** said `GenerateOptions` never reads the flag (confirmed). Corrected; that suite's `generate()` case now genuinely relies on it.
- **Docs paragraph said `generate()` "tries exactly one provider" even with a `ModelPool`** (contested, one skeptic refuted). Made precise anyway: one provider from the static list; a pool still walks its own candidates.

## Not in this PR

Three structured-output recovery cases in the e2e suite (`vendor ignores response_format -> fallback re-asks…`, `failing tool_choice:none re-ask degrades…`, `empty tool_calls finish on the last budgeted step…`) fail identically on a clean `release` checkout at c328f4d0a (138/141 there, 140/140 at 1375cbfcb, the commit before 653d1ff69). They are an incomplete port in the Vercel AI SDK removal; the session that owns that change has reproduced them and is fixing them, and this PR neither touches nor fixes them.

`test:error-classification-e2e` is wired into no CI job (ci.yml references only `test:error-classifier-contract`), which is why CI stayed green through the regression. Wiring it in belongs with that fix, not here; until then the two new cases in this PR run only where someone runs the suite.

## Overlap check

Both live neurolink sessions on this machine were asked before work started and answered NOT MINE on all three items. The openai-compat stream path and `src/lib/middleware` were left untouched by agreement with the session working there.

## Summary by CodeRabbit

- **New Features**
  - Added an option to disable automatic provider and model fallback for text generation.
  - Added CLI support for configuring this option.
  - When enabled, invalid model or unavailable provider errors are surfaced directly, while caller-managed fallback settings remain unaffected.

- **Documentation**
  - Clarified fallback behavior and refreshed API reference links.

- **Tests**
  - Added coverage for fallback and no-fallback behavior.
  - Improved structured-output validation and provider test reliability.

### DIFF
--- a/src/lib/neurolink.ts
+++ b/src/lib/neurolink.ts
@@ -5577,6 +5577,11 @@ Current user's request: ${currentInput}`;
       disableTools: options.disableTools,
       toolFilter: options.toolFilter,
       excludeTools: options.excludeTools,
+      // This explicit field list is the only road into the provider, so a
+      // flag left out here never reaches BaseProvider — which is exactly how
+      // disableInternalFallback was dropped on generate() while stream()
+      // (which spreads its options) honoured it.
+      disableInternalFallback: options.disableInternalFallback,
       maxSteps: options.maxSteps,
       toolChoice: options.toolChoice,
       prepareStep: options.prepareStep,
@@ -8149,12 +8154,25 @@ Current user's request: ${currentInput}`;
       : requestedProvider
         ? [requestedProvider]
         : providerPriority;
+    // The caller owns fallback order (a providerFallback / modelChain caller,
+    // or a router that retries on its own): bound the walk to its first
+    // candidate so an unavailable provider surfaces as its own error instead
+    // of a silent switch. An explicit provider is already a one-element
+    // list; this only changes the "auto" and orchestrated-preference walks.
+    const providersToTry =
+      options.disableInternalFallback === true
+        ? tryProviders.slice(0, 1)
+        : tryProviders;
+    // Caller-owned fallback never enters tryProviders: providerFallback and
+    // modelChain are walked by runWithFallbackOrchestration around the public
+    // generate() call, and a configured ModelPool is consumed by the block
+    // above. Slicing here therefore never clips a caller's own list.
 
     logger.debug(`[${functionTag}] Starting direct generation`, {
       requestedProvider: requestedProvider || "auto",
       preferredOrchestrated: preferredOrchestrated || "none",
-      tryProviders,
-      allowFallback: !requestedProvider || !!preferredOrchestrated,
+      tryProviders: providersToTry,
+      allowFallback: providersToTry.length > 1,
     });
 
     // ─── ModelPool path ──────────────────────────────────────────────────────
@@ -8343,7 +8361,7 @@ Current user's request: ${currentInput}`;
     let lastError: Error | null = null;
 
     // Try each provider in order
-    for (const providerName of tryProviders) {
+    for (const providerName of providersToTry) {
       if (options.abortSignal?.aborted) {
         throw new DOMException("The operation was aborted", "AbortError");
       }
@@ -8795,7 +8813,7 @@ Current user's request: ${currentInput}`;
     // All providers failed
     const responseTime = Date.now() - startTime;
     logger.error(`[${functionTag}] All providers failed`, {
-      triedProviders: tryProviders,
+      triedProviders: providersToTry,
       lastError: lastError?.message,
       responseTime,
     });
--- a/test/continuous-test-suite-provider-matrix.ts
+++ b/test/continuous-test-suite-provider-matrix.ts
@@ -219,22 +219,59 @@ async function runMatrix(): Promise<void> {
             greeting: zod.z.string(),
             count: zod.z.number(),
           });
-          let lastContent: string | undefined;
+          let last:
+            | {
+                content?: string;
+                structuredData?: unknown;
+                jsonTruncated?: boolean;
+              }
+            | undefined;
           for (let attempt = 1; attempt <= 2; attempt++) {
             const r = await sdk.generate({
               ...baseOpts,
               input: { text: 'Reply with greeting="hi" and count=42 in JSON.' },
               maxTokens: 200,
               disableTools: true,
-              structuredOutput: { schema },
-            } as never);
-            lastContent = r.content;
-            if (lastContent && lastContent.length > 0) {
+              // `schema` is the option generate() honours. The previous
+              // `structuredOutput: { schema }` was never a GenerateOptions
+              // field (the `as never` cast hid that) and was silently
+              // ignored, so this cell passed on plain text. The real
+              // parameter type replaces the `as never` cast here so a
+              // wrong option name is a compile error, not a silent no-op.
+              schema,
+            } as Parameters<NeuroLink["generate"]>[0]);
+            last = r;
+            if (
+              r.content &&
+              r.content.length > 0 &&
+              r.structuredData !== undefined
+            ) {
               break;
             }
           }
-          if (!lastContent || lastContent.length === 0) {
+          if (!last?.content || last.content.length === 0) {
             throw new Error("empty response");
+          }
+          // generate({ schema }) yields a parsed `structuredData` whenever the
+          // model produced anything JSON-shaped; a response cut off at the
+          // token cap yields a partial object and sets `jsonTruncated`, so
+          // only an untruncated result is held to the schema. Pure prose
+          // with no JSON at all leaves `structuredData` undefined (the SDK
+          // logs a WARN and returns the raw text). After two attempts that is
+          // the row's `structuredOutput: true` claim being false for this
+          // model, which must stay visible rather than downgrade to a skip.
+          if (last.structuredData === undefined) {
+            throw new Error(
+              "no JSON was recoverable from a schema request after two attempts",
+            );
+          }
+          if (
+            !last.jsonTruncated &&
+            !schema.safeParse(last.structuredData).success
+          ) {
+            throw new Error(
+              "structuredData did not satisfy the request schema",
+            );
           }
         } catch (err) {
           skipIfProviderError(err);
--- a/docs/api/classes/NeuroLink.md
+++ b/docs/api/classes/NeuroLink.md
@@ -469,7 +469,7 @@ Gracefully shutdown NeuroLink and all MCP connections
 
 > **generateText**(`options`): `Promise`\<[`TextGenerationResult`](../type-aliases/TextGenerationResult.md)\>
 
-Defined in: [neurolink.ts:6501](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L6501)
+Defined in: [neurolink.ts:6506](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L6506)
 
 BACKWARD COMPATIBILITY: Legacy generateText method
 Internally calls generate() and converts result format
@@ -490,7 +490,7 @@ Internally calls generate() and converts result format
 
 > **streamText**(`prompt`, `options?`): `Promise`\<`AsyncIterable`\<`string`, `any`, `any`\>\>
 
-Defined in: [neurolink.ts:8980](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L8980)
+Defined in: [neurolink.ts:8998](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L8998)
 
 BACKWARD COMPATIBILITY: Legacy streamText method
 Internally calls stream() and converts result format
@@ -515,7 +515,7 @@ Internally calls stream() and converts result format
 
 > **stream**(`options`): `Promise`\<[`StreamResult`](../type-aliases/StreamResult.md)\>
 
-Defined in: [neurolink.ts:9063](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L9063)
+Defined in: [neurolink.ts:9081](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L9081)
 
 Stream AI-generated content in real-time using the best available provider.
 This method provides real-time streaming of AI responses with full MCP tool integration.
@@ -588,7 +588,7 @@ When conversation memory operations fail (if enabled)
 
 > **setToolRoutingServers**(`servers`): `void`
 
-Defined in: [neurolink.ts:9955](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L9955)
+Defined in: [neurolink.ts:9973](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L9973)
 
 Supplies (or replaces) the pre-call tool routing server catalog.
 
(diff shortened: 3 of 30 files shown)


src/lib/neurolink.ts
```diff
@@ -8149,12 +8154,21 @@ Current user's request: ${currentInput}`;
...
         ? [requestedProvider]
         : providerPriority;
+    // The caller owns fallback order (a providerFallback / modelChain caller,
+    // or a router that retries on its own): bound the walk to its first
+    // candidate so an unavailable provider surfaces as its own error instead
+    // of a silent switch. An explicit provider is already a one-element
+    // list; this only changes the "auto" and orchestrated-preference walks.
+    const providersToTry =
+      options.disableInternalFallback === true
+        ? tryProviders.slice(0, 1)
```
Tara-ag, line 8164: **MAJOR (verification request)**

`providersToTry = tryProviders.slice(0, 1)` bounds the walk to the first candidate whenever the flag is set. The code comment and the docs both say this "only changes the `"auto"` and orchestrated-preference walks" and that a caller-configured `providerFallback` / `modelChain` is "the caller's own fallback and is unaffected."

The behaviour lives or dies on one distinction I want to see made explicit: for a caller that supplies their **own** orchestrated fallback (`providerFallback` or `modelChain`) **and** sets `disableInternalFallback: true` (to suppress the catalog *model* fallback), do those caller-supplied providers enter `tryProviders`? If they do, `slice(0, 1)` clips the caller's own fallback list to its first entry, which would contradict the documented contract and the PR summary. If their handling is a separate branch (ModelPool / another walk) that never reaches `tryProviders`, the slice is correct.

Could you confirm which branch in `directProviderGeneration` consumes `options.providerFallback` / `options.modelChain`, and add one line to the comment here stating that caller-owned fallback lists are read elsewhere and are not part of `tryProviders`? That makes the promise in `docs/features/provider-fallback.md` verifiable from the code.

```suggestion
        ? tryProviders.slice(0, 1)
        : tryProviders;
    // NOTE: caller-owned fallback (providerFallback / modelChain) is handled
    // outside tryProviders, so slicing here never clips the caller's own walk.
```
(adjust the note to match where that fallback is actually consumed)

test/continuous-test-suite-provider-matrix.ts
```diff
@@ -219,23 +219,58 @@ async function runMatrix(): Promise<void> {
...
               input: { text: 'Reply with greeting="hi" and count=42 in JSON.' },
               maxTokens: 200,
               disableTools: true,
-              structuredOutput: { schema },
+              // `schema` is the option generate() honours. The previous
+              // `structuredOutput: { schema }` was never a GenerateOptions
+              // field (the `as never` cast hid that) and was silently
+              // ignored, so this cell passed on plain text.
+              schema,
             } as never);
```
Tara-ag: **MINOR**

This PR's whole point is that a silently-ignored option (`structuredOutput: { schema }`) was hidden by the surrounding `as never` cast, so the cell passed on plain text. The fix is right — `schema` is the honoured field now — but the `} as never);` cast is still there on the next line, so the same class of bug can recur: a typo'd or renamed option on this call will be silently ignored again instead of being a compile error.

Consider replacing `as never` with the real `GenerateOptions` parameter type (or `Parameters<NeuroLink['generate']>[0]`) for this call so the schema field is type-checked, keeping the comment as the record of *why* it changed. Only worth doing if it doesn't drag in unrelated lint types; otherwise the cast can stay.

```suggestion
              schema,
            } as Parameters<NeuroLink["generate"]>[0]);
```

Tara-ag:
**NEEDS_WORK (verification items, not blocks)** — the change is well-scoped and the logic fixes (`buildGenerateTextOptions` now maps the flag through; `BaseProvider.generate()` reads the now-typed field directly; the matrix cell finally asserts `structuredData`) are correct in principle, but I want two things confirmed before I'm comfortable.

|---|---|---|
| MAJOR | `src/lib/neurolink.ts:8164` | Confirm caller-owned `providerFallback`/`modelChain` never enters `tryProviders`, so `slice(0,1)` under `disableInternalFallback` cannot clip the caller's own fallback walk. |
| MINOR | `test/continuous-test-suite-provider-matrix.ts:240` | `as never` cast is retained, re-opening the silently-ignored-option class this PR fixes. |

**State of the thread:** no prior Yama review here; CodeRabbit (CHILL) reported no actionable comments.

**What was checked and found clean:**
- Type surface (`GenerateOptions.disableInternalFallback`, `TextGenerationOptions.disableInternalFallback`) plus the CLI `Record<keyof TextGenerationOptions,…>` schema entry are consistent — Rule 5 (SDK backward compat) holds; these are additive optional fields, no unmodified caller breaks.
- `buildGenerateTextOptions` mapping closes the real regression (explicit field list dropping the flag on `generate()` while `stream()` spread honoured it).
- `BaseProvider.generate()` workaround removal is a correct simplification now that the field is typed; no behaviour change on the three provider-level reads.
- Docs (`docs/features/provider-fallback.md`) and vertex-loop-characterization comment were corrected to match the new reality; generated `docs/api` refreshed.
- New error-classification cases 6/7 assert the default catalogue retry and the single-request `InvalidModelError`; the matrix cell now exercises `structuredData` against the schema with only the documented `jsonTruncated` exemption.
- No static provider imports added (Rule 1 untouched), no secrets/credentials introduced, CLI-flag is additive schema metadata only.

**Impact note:** the bundled code graph is built at branch `HEAD` (47bc8c6), which is the `release` base, not the PR head `43e935c`, so a precise blast-radius diff against the PR commit wasn't available. Impact was assessed from the PR diff and PR description; the two changed exports (`buildGenerateTextOptions`, `directProviderGeneration`) are internal to `generate()`'s path — no unmodified external caller is affected by signature/shape changes. One minor logger-field refinement (`allowFallback: providersToTry.length > 1`) changes a *debug log* value when `requestedProvider`+`preferredOrchestrated` are combined; it is diagnostic-only and does not gate the walk, so I have not raised it as a finding.

Once the inline MAJOR question is answered (with the one-line comment), this is ready.

Tara-ag (changes requested):
Requesting changes to align the review state with the verdict.

The change is well-scoped and correct in principle, but the summary verdict is **NEEDS_WORK** pending confirmation of the MAJOR item on `src/lib/neurolink.ts:8164`: that caller-owned `providerFallback` / `modelChain` never enter `tryProviders`, so `slice(0, 1)` under `disableInternalFallback` cannot clip the caller's own fallback walk. Once that is answered (one code note making the promise verifiable), this can be re-approved.

Two inline findings (MAJOR verification, MINOR `as never` cast) and the summary comment are posted. See `<!-- yama:summary -->` for the full breakdown.

src/lib/neurolink.ts
murdore: Verified against the call paths, and the guard is now stated at the site (0c8c10790).

`tryProviders` is built only from the static `providerPriority` list (or the orchestrated-preference reordering of it) or the single explicit provider. Caller-owned fallback never enters it:

- `providerFallback` / `modelChain` are walked by `runWithFallbackOrchestration`, which wraps the public `generate()` call (line ~4483) and tries its own candidates from outside this function, so each of the caller's entries arrives here as its own request.
- A configured `ModelPool` is consumed by the pool block above this walk and returns before `tryProviders` is built.

So `slice(0, 1)` bounds only NeuroLink's own walk, which is what the flag documents on both `GenerateOptions` and `StreamOptions`. Regression guard: the Mancer mock-server cases in `test/continuous-test-suite-error-classification-e2e.ts` (cases 6 and 7) assert at least two requests by default and exactly one with `disableInternalFallback: true`; on the pre-change build the one-request assertion fails with three requests.

test/continuous-test-suite-provider-matrix.ts
murdore: Replaced in 0c8c10790: the cell now casts to `Parameters<NeuroLink["generate"]>[0]`, so the compiler checks the option names instead of `as never` hiding them (the original `structuredOutput: { schema }` bug is exactly what that cast concealed). `pnpm run check:tools-tests` passes with the typed cast.

Tara-ag (approved):
APPROVE — both recurring-review items are now resolved on `0c8c10790`.

Verification of the two earlier findings (posted under `<!-- yama:generate-flag-bounds-static-walk -->` and `<!-- yama:matrix-as-never-cast -->`), both adequately answered and fixed in this commit, so I'm updating rather than reposting:

| Severity | Location | Verdict |
| --- | --- | --- |
| MAJOR | `src/lib/neurolink.ts:8164` | Resolved. The graph confirms the call chain `generate() → runWithFallbackOrchestration (neurolink.ts:4483) → … → directProviderGeneration`. Caller-owned `providerFallback`/`modelChain` are walked by `runWithFallbackOrchestration` (neurolink.ts:4602-4761) around the generate() call — never inside `directProviderGeneration`'s `tryProviders` (built only from `requestedProvider ? [requestedProvider] : providerPriority`). The `slice(0, 1)` therefore only bounds NeuroLink's own walk, and an on-site guard comment + Mancer cases 6/7 regression-guard it. |
| MINOR | `test/continuous-test-suite-provider-matrix.ts:240` | Resolved. `as never` replaced by `as Parameters<NeuroLink["generate"]>[0]`, so a wrong option name is now a compile error. Read the file: `schema` honours the field, `structuredData`/`jsonTruncated` assertion logic is sound. |

Code correctness confirmed across the diff (no new inline findings to post):

- `buildGenerateTextOptions` (neurolink.ts:5577) maps the flag through the explicit field list — this is the actual regression closed (flag previously dropped on generate() while stream() spread it). Callers: only `runStandardGenerateRequest` (1 caller, internal to generate()). Rule 5 holds — additive optional field, no unmodified caller breaks.
- `directProviderGeneration`: `allowFallback: providersToTry.length > 1` is logger-payload-only (does not gate the `for` loop) — diagnostic-only shift for the `requestedProvider && preferredOrchestrated` combination; not a behavioural change.
- `BaseProvider.generate()` (baseProvider.ts:1758): correct simplification now that `TextGenerationOptions.disableInternalFallback` is typed; `runGenerateWithModelFallback(attempt, callerOwnsFallback)` skips the catalog walk. Three provider-level reads (openai-compat base, Anthropic, SageMaker) intentionally untouched per the overlap agreement.
- Types (`src/lib/types/generate.ts`), CLI schema (`src/cli/loop/optionsSchema.ts`), docs (`docs/features/provider-fallback.md`), and the vertex-loop-comment correction are mutually consistent. CLI schema entry is required by the `Record<keyof Omit<TextGenerationOptions,…>>` shape.

No CRITICAL security issues (secret scan of the diff clean; no hardcoded creds), no static provider imports (Rule 1 intact), no silent truncation (truncation still surfaces via `jsonTruncated`).

One accuracy note on the PR body, verified against the branch's own `.github/workflows/ci.yml`: the claim that `test:error-classification-e2e` "is wired into no CI job (ci.yml references only `test:error-classifier-contract`)" is not current for this branch — `ci.yml`'s `extended-suites` job lists `test:error-classification-e2e@71` in its SUITES. So cases 6 and 7 DO run in CI, though in a non-blocking job. Metadata only; doesn't affect the code.

PR #1635, opened 2026-09-04
fix(core): restore three native-loop recoveries the SDK removal dropped

murdore:
## What broke

#1628 ported two `GenerationHandler` recoveries into the native generate loop. The port was incomplete in three ways. `test:error-classification-e2e` goes **140/140 at `1375cbfcb` → 138/141 at `c328f4d0a`**.

That was measured, not inferred: the suite was built and run at both commits. The pre-refactor run needed its own `pnpm install --frozen-lockfile`, because `node_modules` now matches the post-removal lockfile while the old tree still imports `ai`.

Reported by a parallel session working in the same repo; the attribution to #1628 is mine.

## The three defects

1. **The structured-output fallback fired only when the provider threw.** It was gated on `isToolsSchemaConflictError || isSchemaComplexityError`. The case it was written for is a vendor that *accepts* `response_format` and answers in prose anyway (GMI Cloud's MiniMax). On the ai-package path the structured-output parser threw on the unparseable answer, which is what reached the catch — the native loop has no such parser, so the silent case sailed through and handed the caller prose. It now also triggers on the result: if `response_format` was sent and the text yields no schema-accepted object, re-ask once with the schema in the system prompt.

2. **A failing `toolChoice: none` re-ask threw instead of degrading.** The native port made the re-ask a `continue`, so its failure surfaced from the *next* step's `doGenerate` and propagated. The loop now snapshots the turn before re-asking and restores it if the re-ask fails — matching `recoverEmptyToolCallsFinish`'s catch, where the re-ask is a bonus request on top of a call that already succeeded. The flag is deliberately separate from `reasked`, which stays set for the rest of the turn and would otherwise swallow a later unrelated failure.

3. **`stopReason` / `stepsUsed` were never surfaced.** The loop already returned `steps`; the OpenAI-compatible result dropped them, so a turn ended by the step cap was indistinguishable from a completed one.

## Why it merged green

**The suite runs in no CI job at all.** `ci.yml` references only `test:error-classifier-contract` — a different suite. `grep -rn error-classification .github/` returns nothing. The extended shards were green on both commits because they never executed it.

It is now wired into the extended shards, qualified under that job's own stated rule: **141/141 with the environment fully stripped** (`env -i`, `DOTENV_CONFIG_PATH=/dev/null`), 71s, entering at that weight.

## The test that gave false confidence

`continuous-test-suite-native-vendor-recovery.ts` is the suite that certified the original port. Its header described the *silent* case while the test exercised the *rejected* one — so a green run proved a recovery that did not exist. Both triggers are now separate cases; header corrected to say exactly that.

## Verification

| gate | result |
| --- | --- |
| `test:error-classification-e2e` | **141/141** (was 138/141) |
| `test:error-classification-e2e`, env stripped | 141/141, 71s |
| vendor-recovery | 3/3 (was 2/2) |
| lint / typecheck / build | 0 errors / clean / clean |

Note `12.12.0` is published with these three defects.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved structured-output handling when a provider returns prose despite accepting a schema request.
  * Added an automatic retry that reinforces the requested schema and preserves the last successful result if recovery fails.
  * Generation results now include more complete completion details, including stop reason and steps used.

* **Tests**
  * Added coverage for vendor responses that silently ignore structured-output requests.
  * Included the error-classification end-to-end suite in extended test runs.

### DIFF
--- a/src/lib/providers/openaiChatCompletionsBase.ts
+++ b/src/lib/providers/openaiChatCompletionsBase.ts
@@ -59,6 +59,7 @@ import type {
   Schema,
   StreamLoopArgs,
   EnhancedGenerateResult,
+  GenerateStopReason,
   TextGenerationOptions,
   ValidationSchema,
   StreamOptions,
@@ -88,6 +89,7 @@ import {
 } from "../core/nativeGenerateLoop.js";
 import { resolveToolExecutionRecords } from "../core/toolExecutionRecorder.js";
 import { convertZodToJsonSchema } from "../utils/schemaConversion.js";
+import { coerceJsonToSchema, schemaAccepts } from "../utils/json/coerce.js";
 import { resolveToolChoice } from "../utils/toolChoice.js";
 import { transformToolExecutions } from "../utils/transformationUtils.js";
 import { withProviderRetry } from "../utils/providerRetry.js";
@@ -126,6 +128,23 @@ const WINDOW_FIT_MARGIN_TOKENS = 512;
 /**
  * Abstract HTTP+SSE provider for OpenAI chat-completions-shaped endpoints.
  */
+
+/**
+ * Did the model's text yield an object the caller's schema accepts?
+ *
+ * This is the trigger for the prompt-side structured-output fallback. It asks
+ * the question the ai-package's structured-output parser used to ask by
+ * throwing: did the native `response_format` attempt actually produce the
+ * object. A schema we cannot validate with accepts everything, so an unknown
+ * schema never forces a pointless second request.
+ */
+const yieldsSchemaValidObject = (
+  text: string,
+  schema: ValidationSchema,
+): boolean => {
+  const coerced = coerceJsonToSchema(text, schema);
+  return coerced !== null && schemaAccepts(schema, coerced.structuredData);
+};
 
 export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
   protected config: { baseURL: string; apiKey: string };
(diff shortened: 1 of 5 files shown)


src/lib/providers/openaiChatCompletionsBase.ts
```diff
@@ -1165,15 +1184,54 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
...
+    // is the case the fallback was written for. On the ai-package path the
+    // structured-output parser threw on the unparseable answer, so the catch
+    // above was reached; the native loop has no such parser, so the silent
+    // case sailed through and handed the caller prose. Same recovery, keyed on
+    // the result rather than on an exception.
+    if (
+      responseFormat !== undefined &&
+      options.schema !== undefined &&
+      !yieldsSchemaValidObject(loop.text, options.schema as ValidationSchema)
+    ) {
```
Tara-ag, line 1199: **MINOR**

**What breaks:** the silent-`response_format` recovery can fire *after* the exception-based one already walked the same path, so a turn can cost three requests and get the JSON Schema pasted into the system prompt twice. Trace the composed case:

1. Vendor 400s on `response_format` → the catch above runs `appendJsonSchemaInstruction(conversation, responseFormat.schema)` and retries (schema now in the system prompt).
2. That retry *also* answers prose / fails schema validation.
3. This result-driven check then fires **again** on the same `conversation`, re-appends the schema (it concatenates onto the existing system message, so the schema block is duplicated) and issues a third request.

**Why it matters here:** the recovery is meant to be bounded. `GenerationCallConfig.promptJsonInstruction` — the flag this was ported from (`types/generate.ts`) — exists exactly to mark the prompt-side fallback as spent for the turn, and this port didn't carry that across the two triggers. The schema-in-prompt retry failing validation means a second identical instruction is a no-op anyway.

**Fix:** spend the same one-shot the ai-package path did. Add `!promptSideFallbackRan` to the condition below, declare `let promptSideFallbackRan = false;` just before the `try`, and set it `true` inside the `catch (error)` fallback above:

```suggestion
    if (
      responseFormat !== undefined &&
      options.schema !== undefined &&
      !promptSideFallbackRan &&
      !yieldsSchemaValidObject(loop.text, options.schema as ValidationSchema)
    ) {
```

Tara-ag (commented):
Inline note on the composed structured-output fallback path. Full summary follows as a separate comment.

Tara-ag:
## Yama Review — APPROVE

Solid, well-tested restoration of the three native-loop recoveries dropped in the SDK removal. The regressions are precisely characterized, the fix matches the `GenerationHandler` semantics it ports, and both triggers of the structured-output fallback are now covered by tests and wired into CI.

### Findings

|---|---|---|
| MINOR | `src/lib/providers/openaiChatCompletionsBase.ts:1195` | The silent-`response_format` fallback can fire after the exception-based fallback already ran this turn — 3 requests, schema duplicated in the prompt. Gate on a one-shot flag (parity with `GenerationCallConfig.promptJsonInstruction`). |

### Checked and found clean

- **Rule 5 (public-API backward compat):** `stopReason` and `stepsUsed` are optional fields on `GenerateResult` — this is an additive change, no unmodified callers break (`EnhancedGenerateResult = GenerateResult & {…}`; all three consumers `openaiChatCompletionsBase`, Anthropic `executeNativeGenerate`, SageMaker are the loop's callers and are the ones being fixed).
- **stopReason/stepsUsed parity:** matches `GenerationHandler`'s step-cap (= `stepsUsed >= maxSteps && finishReason === "tool-calls"`) → `step-cap` / `error` → `provider-error` precedence. Verified against `core/modules/GenerationHandler.ts`.
- **Re-ask failure recovery:** the `reaskPending` flag is correctly scoped so only the re-ask *step itself* degrades, and a later unrelated failure still throws — prevents swallowing real errors. Snapshot/restore is faithful to `recoverEmptyToolCallsFinish`'s catch.
- **Silent-case trigger logic:** verified against `utils/json/coerce.ts` — `coerceJsonToSchema` returns `null` on unparseable text and `schemaAccepts` accepts-everything only for schemas without `safeParse`, so an unknown schema never forces a pointless re-ask (comment in source is accurate).
- **Blast radius (graph):** `runNativeGenerateLoop`/`openaiChatCompletionsBase` changes propagate only through the native-loop providers (Anthropic, SageMaker, OpenAI-compatible family) — 84 dependent files, all consuming the additive result fields, none broken by signature/shape change.
- **Security bar:** no credentials in source; the `GOOGLE_*` values in the e2e suite are explicit placeholders (not flagged per convention). No seed/secret leakage in logs.
- **Coverage (Rule 15 context):** the silent case now has its own end-to-end test (`continuous-test-suite-native-vendor-recovery.ts`), and `test:error-classification-e2e` is wired into the extended CI shards at the weight it was measured (71s) — closing the "merged green because the suite never ran" gap.

No blocking issues. The one MINOR is a bounded cost/complexity refinement, not a merge gate.

Tara-ag (approved):
Approving per the full Yama review summary (`<!-- yama:summary -->`). Solid restoration of the three native-loop recoveries with matched `GenerationHandler` semantics and e2e coverage. The single MINOR finding (bounded structured-output fallback refinement) is documented inline and is not a merge gate.

PR #1636, opened 2026-09-04
fix(middleware): apply model middleware on the OpenAI-compatible streaming path

murdore:
## What

`middleware` is a public option on both `generate()` and `stream()`, but every site that applied it was a generate path. `wrapStream` was dead code repo-wide, `transformParams({ type: "stream" })` never fired, and a guardrail that blocks a prompt under `generate()` was bypassed entirely by `stream()`.

This closes that gap for the **OpenAI-compatible family only** — every provider on `OpenAIChatCompletionsProvider` (the catalog providers, DeepSeek, Cohere, Azure OpenAI, LM Studio, llama.cpp, LiteLLM). Anthropic, Vertex, AI Studio and Bedrock each have their own `executeStream` and still bypass middleware on stream; that is tracked separately and is not claimed here. An earlier version of this PR described itself as "the streaming path" — that overstated it, and the title, commit and this body are the correction.

## How

- `BaseProvider.applyMiddlewareToModel` splits chain-application out of `getAISDKModelWithMiddleware`, which always wrapped the non-streaming model.
- `executeStream` builds a V3 base model whose `doStream` starts the real multi-step stream loop, wraps it with the middleware chain, and drives the wrapped model. The prompt is converted to the wire format only **after** `transformParams`, which is what makes a rewrite actually reach the wire.
- Honoured on the way back in: `prompt`, `maxOutputTokens`, `temperature`, `topP`. `tools` is read-only — a rewrite gets a WARN rather than a silent drop.

## Defects found in review and fixed at the root

Review (Yama) flagged that the reader keyed on `textDelta`. That was a symptom of a wider problem — the bridge was not a faithful V3 stream — and stronger tests then found four defects. All are fixed here:

1. **No terminal `finish` part.** A middleware observing the stream could not see usage or the finish reason. The bridge now emits `finish` from the loop's deferred usage/finish promises; a synthetic stream from a blocking middleware settles analytics too.
2. **Blocking middleware hung analytics.** Guardrails' precall path returns its own stream without calling `doStream`, so `loopPromise` never existed and the turn's usage/finish never resolved. Every reader now tolerates its absence.
3. **Cancellation never reached upstream.** Breaking out of a wrapped stream left the HTTP request open. `chunksToV3Stream` pulls one chunk at a time and forwards `cancel()` to the consumer abort controller and the iterator.
4. **Guardrails emitted v1/v2 parts** (`textDelta`, `promptTokens`) and filtered on `"textDelta" in chunk`, so its blocked stream surfaced as the "Stream produced no output" sentinel and its bad-word filter never matched a V3 delta. `createBlockedResponse` / `createBlockedStream` now emit V3 parts and the filter reads `delta`. This is the root fix for the review finding, not a fallback in the reader.

## Proof

`test/continuous-test-suite-stream-middleware.ts` — 20 cases driving the shipped `dist` against local HTTP stand-ins, on both `generate()` and `stream()`:

- prompt rewrite reaches the wire; V3 `finish` observed with usage
- blocking stream delivers text and settles analytics with **zero** HTTP requests
- guardrail bad-word filtering and precall blocking on both modes, with the evaluator itself a scripted stand-in
- a real tool round trip with middleware enabled, both modes
- provider failure propagates; abort and timeout close the in-flight request on both modes
- middleware sampling edits reach the wire; schema recovery still works

Also run with an empty `HOME` and cwd (no `.env`) so no ambient credential can satisfy anything. The suite is wired into the extended CI shards (`test:stream-middleware@5`).

Full gate on the final code (all against the rebuilt `dist`):

| gate | result |
| --- | --- |
| `check:dts` | every shipped declaration resolves |
| `test:stream-middleware` | 20/20 |
| `test:vendor-recovery` | 3/3 |
| `test:openai-compat-streaming-retry` | 2/2 |
| `test:adjust-body-after-400` | 2/2 |
| `test:error-classification-e2e` | 144/144 |
| `test:anthropic-loop-characterization` | 10/10 |
| `test:middleware` (live, Vertex) | 6/7 — see below |

The one miss in the chain was `generate() onFinish`: Vertex (`gemini-2.5-flash`, a thinking model) returned empty content at the suite's `maxTokens: 50`. Re-run alone on the same build: **7/7**. A probe with the case's own prompt at 50 tokens shows 28 of 29 output tokens spent on reasoning, so an empty reply there is the budget, not the middleware — this PR does not touch Vertex's generate path. The budget is fixed in a separate test-harness PR.

## Not in this PR

- Middleware on the native stream paths of Anthropic, Vertex, AI Studio and Bedrock.
- A mutable `tools` in `transformParams` for streams.

## Summary by CodeRabbit

- **New Features**
  - Model middleware now applies consistently to streaming responses, including prompt transformations, guardrails, tool interactions, and sampling overrides.
  - OpenAI-compatible streaming supports middleware-driven response filtering and request blocking.

- **Bug Fixes**
  - Guardrails now correctly filter unsafe content across generated text and streamed output.
  - Improved handling of blocked requests, provider errors, aborted requests, timeouts, and interrupted streams.
  - Streaming usage and completion status are reported correctly when middleware intervenes.

### DIFF
--- a/src/lib/middleware/builtin/guardrails.ts
+++ b/src/lib/middleware/builtin/guardrails.ts
@@ -2,6 +2,7 @@ import type {
   NeuroLinkMiddleware,
   NeuroLinkMiddlewareMetadata,
   GuardrailsMiddlewareConfig,
+  LanguageModelV3StreamPart,
 } from "../../types/index.js";
 import {
   createBlockedResponse,
@@ -67,8 +68,7 @@ export function createGuardrailsMiddleware(
 
   const middleware: LanguageModelMiddleware = {
     specificationVersion: "v3" as const,
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    transformParams: async ({ params }: any) => {
+    transformParams: async ({ params }) => {
       if (config.precallEvaluation?.enabled) {
         const { shouldBlock, transformedParams } =
           await handlePrecallGuardrails(params, config.precallEvaluation);
@@ -79,8 +79,7 @@ export function createGuardrailsMiddleware(
       return params;
     },
 
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    wrapGenerate: async ({ doGenerate, params }: any) => {
+    wrapGenerate: async ({ doGenerate, params }) => {
       logger.debug(`[GuardrailsMiddleware] Applying to generate call.`);
       // Check if this request should be blocked (set by transformParams)
       if (config.precallEvaluation?.enabled && blockingState.get(params)) {
@@ -89,19 +88,30 @@ export function createGuardrailsMiddleware(
 
       let result = await doGenerate();
 
-      const filterResult = applyContentFiltering(
-        result.text || "",
-        config.badWords,
-        "generate",
-      );
-      if (filterResult.hasChanges) {
-        result = { ...result, text: filterResult.filteredText };
-      }
+      result = {
+        ...result,
+        content: result.content.map((part) =>
+          part.type === "text"
+            ? {
+                ...part,
+                text: applyContentFiltering(
+                  part.text,
+                  config.badWords,
+                  "generate",
+                ).filteredText,
+              }
+            : part,
+        ),
+      };
 
       if (config.modelFilter?.enabled && config.modelFilter.filterModel) {
         logger.debug(`[GuardrailsMiddleware] Invoking model-based filter.`);
         try {
-          const filterPrompt = `Is the following text safe? Respond with only "safe" or "unsafe".\n\nText: "${result.text}"`;
+          const text = result.content
+            .filter((part) => part.type === "text")
+            .map((part) => part.text)
+            .join("");
+          const filterPrompt = `Is the following text safe? Respond with only "safe" or "unsafe".\n\nText: "${text}"`;
           // `ModelFilterConfig.filterModel` is typed `LanguageModel`, which
           // admits a bare model id, and the documented examples used one.
           // `generateOnceNative` needs a handle exposing doGenerate, so a
@@ -119,7 +129,14 @@ export function createGuardrailsMiddleware(
             logger.warn(
               `[GuardrailsMiddleware] Model-based filter flagged content as unsafe.`,
             );
-            result = { ...result, text: "<REDACTED BY AI GUARDRAIL>" };
+            result = {
+              ...result,
+              content: result.content.map((part) =>
+                part.type === "text"
+                  ? { ...part, text: "<REDACTED BY AI GUARDRAIL>" }
+                  : part,
+              ),
+            };
           }
         } catch (error) {
           logger.error(`[GuardrailsMiddleware] Model-based filter failed.`, {
@@ -131,8 +148,7 @@ export function createGuardrailsMiddleware(
       return result;
     },
 
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    wrapStream: async ({ doStream, params }: any) => {
+    wrapStream: async ({ doStream, params }) => {
       logger.debug(`[GuardrailsMiddleware] Applying to stream call.`);
 
       // Check if this request should be blocked (set by transformParams)
@@ -147,23 +163,23 @@ export function createGuardrailsMiddleware(
       const { stream, ...rest } = await doStream();
       let hasYieldedChunks = false;
 
-      const transformStream = new TransformStream({
+      const transformStream = new TransformStream<
+        LanguageModelV3StreamPart,
+        LanguageModelV3StreamPart
+      >({
         transform(chunk, controller) {
           hasYieldedChunks = true;
           let filteredChunk = chunk;
-          if (
-            typeof filteredChunk === "object" &&
-            "textDelta" in filteredChunk
-          ) {
+          if (filteredChunk.type === "text-delta") {
             const filterResult = applyContentFiltering(
-              filteredChunk.textDelta,
+              filteredChunk.delta,
               config.badWords,
               "stream",
             );
             if (filterResult.hasChanges) {
               filteredChunk = {
                 ...filteredChunk,
-                textDelta: filterResult.filteredText,
+                delta: filterResult.filteredText,
               };
             }
           }
--- a/src/lib/providers/openaiChatCompletionsBase.ts
+++ b/src/lib/providers/openaiChatCompletionsBase.ts
@@ -40,6 +40,9 @@ import { createProxyFetch } from "../proxy/proxyFetch.js";
 import type {
   DeferredUsage,
   LanguageModel,
+  LanguageModelV3StreamPart,
+  LanguageModelV3,
+  LanguageModelV3CallOptions,
   ModelsResponse,
   OpenAICompatBuildBodyArgs,
   OpenAICompatChatChoice,
@@ -145,6 +148,76 @@ const yieldsSchemaValidObject = (
   const coerced = coerceJsonToSchema(text, schema);
   return coerced !== null && schemaAccepts(schema, coerced.structuredData);
 };
+
+// Pull one native chunk at a time and forward cancellation to its iterator.
+const chunksToV3Stream = (
+  source: AsyncIterable<OpenAICompatStreamChunk>,
+  completion: Promise<LanguageModelV3StreamPart>,
+  cancel: () => void,
+): ReadableStream<LanguageModelV3StreamPart> => {
+  const iterator = source[Symbol.asyncIterator]();
+  return new ReadableStream<LanguageModelV3StreamPart>({
+    async pull(controller) {
+      try {
+        const next = await iterator.next();
+        if (next.done) {
+          controller.enqueue(await completion);
+          controller.close();
+        } else if (next.value.reasoning) {
+          controller.enqueue({
+            type: "reasoning-delta",
+            delta: next.value.reasoning,
+          });
+        } else {
+          controller.enqueue({ type: "text-delta", delta: next.value.content });
+        }
+      } catch (error) {
+        controller.error(error);
+      }
+    },
+    async cancel() {
+      cancel();
+      await iterator.return?.();
+    },
+  });
+};
+
+async function* v3StreamToChunks(
+  stream: ReadableStream<LanguageModelV3StreamPart>,
+  onFinish: (
+    part: Extract<LanguageModelV3StreamPart, { type: "finish" }>,
+  ) => void,
+): AsyncIterable<OpenAICompatStreamChunk> {
+  const reader = stream.getReader();
+  let done = false;
+  try {
+    while (true) {
+      const next = await reader.read();
+      if (next.done) {
+        done = true;
+        return;
+      }
+      const part = next.value;
+      if (part.type === "text-delta") {
+        yield { content: part.delta };
+      } else if (part.type === "reasoning-delta") {
+        yield { content: "", reasoning: part.delta };
+      } else if (part.type === "finish") {
+        onFinish(part);
+      } else if (part.type === "error") {
+        throw part.error;
+      }
+    }
+  } finally {
+    try {
+      if (!done) {
+        await reader.cancel();
+      }
+    } finally {
+      reader.releaseLock();
+    }
+  }
+}
 
 export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
   protected config: { baseURL: string; apiKey: string };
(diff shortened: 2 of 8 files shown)


src/lib/providers/openaiChatCompletionsBase.ts
```diff
@@ -127,6 +128,74 @@ const WINDOW_FIT_MARGIN_TOKENS = 512;
...
+    for (;;) {
+      const { done, value } = await reader.read();
+      if (done) {
+        return;
+      }
+      if (!value) {
+        continue;
+      }
+      if (value.type === "text-delta") {
+        yield { content: value.delta ?? "" };
```
Tara-ag: This drop turns guardrail **blocking** into an empty stream + a misleading "Stream produced no output" sentinel, so the block reason never reaches the caller.

`createBlockedStream()` (src/lib/middleware/utils/guardrailsUtils.ts:296) emits its text as a **v1/v2-shaped** part:

```ts
controller.enqueue({ type: "text-delta", textDelta: "Request contains inappropriate content and has been blocked." });
```

but this reader only looks at `value.delta`. For a blocked stream `value.delta` is `undefined` → this yields `{ content: "" }`. Downstream in `transformedStream`, that chunk fails the `content.length > 0` check, so `contentYielded` stays `0` and the no-output sentinel path (line ~1503) fires — the user gets a "no output" sentinel instead of the block message, and the `async function* v3StreamToChunks` JSDoc claim that the round trip is lossless doesn't hold for middleware-authored streams.

The body correctly flags that bad-word *filtering* won't match V3 parts, but this is the same `textDelta` vs `delta` mismatch biting the *blocking* path the change is meant to enable.

```suggestion
        yield {
          // Some middleware-authored parts (e.g. guardrails' blocked stream)
          // still carry the v1/v2 `textDelta` field rather than V3 `delta`.
          content: value.delta ?? (value as { textDelta?: string }).textDelta ?? "",
        };
```

A regression test driving the guardrails block path (precall evaluation that blocks) against `stream()` and asserting the block text reaches the consumer would lock this in — the current suite's probe never exercises a blocking `wrapStream`, so this gap isn't covered.

Tara-ag (commented):
## Review — apply model middleware on the streaming path

Thanks @murdore — this closes a real hole: `middleware` was a public option on `StreamOptions` but nothing on a streaming path ever reached `getAISDKModelWithMiddleware`, so `wrapStream` / `transformParams({ type: "stream" })` were repo-wide dead code and guardrails' precall **block** was bypassed by `stream()`. Splitting `applyMiddlewareToModel` out of `getAISDKModelWithMiddleware` is the right seam, and building a base model whose `doStream` starts the real multi-step loop (rather than the generate path's `buildDelegatingModel`) is the correct shape — it gives the stream path the same contract `generate()` has always had.

I verified against the source at `a234f1c`:

- `MiddlewareFactory.applyMiddleware` → `wrapLanguageModel` (src/lib/middleware/wrapLanguageModel.ts:56) — `doStream` runs `transform(params, "stream")` then `wrapStream({ doStream, params: transformed })`, so a prompt rewrite and a blocking `wrapStream` both route correctly, and the guardrails block check (`blockingState.get(params)`) hits because `doWrap` passes the same `transformed` object reference through (`factory.ts` → `doWrap`). ✓
- `OpenAIProvider extends OpenAIChatCompletionsProvider`, so the new offline suite (`provider: "openai"`) exercises exactly this `executeStream`. ✓
- The `loopPromise ?? Promise.resolve()` / `await loopPromise`-with-undefined handling correctly tolerates a blocked stream never starting the loop. ✓

One genuine defect, inline at `src/lib/providers/openaiChatCompletionsBase.ts:183`: **blocked guardrail streams lose their block text.** `createBlockedStream()` carries text in the legacy `textDelta` field, but `v3StreamToChunks` reads only `value.delta`, so it yields `content: ""`, which then trips the "Stream produced no output" sentinel instead of surfacing the block message. Same `textDelta`-vs-`delta` class as the bad-word filtering gap you flagged — but that one you noted; this one is unflagged and sits on the exact path the PR is enabling. Suggestion + a regression-test nudge in the inline comment.

Non-blocking observations (no change requested):
- `tools` offered read-only with a loud `logger.warn` rather than a silent drop is a well-reasoned limit — re-deriving the wire tool list would indeed diverge from `buildToolsForOpenAI`. Good to state it inline.
- The "Not fixed here" note on guardrails' `"textDelta" in chunk` filter is appreciated and accurate; worth a tracking issue so the two `textDelta` follow-ups don't get lost.
- Nice touch asserting the generate precondition first — it makes a red stream test mean "the code is wrong" rather than "the probe never registered."

The inline `textDelta` fix is small and high-value; once that's in I'll re-check the block path.

Tara-ag:
## Verdict: NEEDS_WORK

This change closes a genuine hole — `middleware` was a public option on `StreamOptions`, but nothing on a streaming path ever reached `getAISDKModelWithMiddleware`, so `wrapStream` / `transformParams({ type: "stream" })` were repo-wide dead code and guardrails' precall **block** was bypassed entirely by `stream()`. The chosen seam is correct: splitting `applyMiddlewareToModel()` out and driving a base model whose `doStream` starts the real multi-step stream loop gives the streaming path the same contract `generate()` has always had. One real defect blocks merge.

| Severity | Finding | Location |
| --- | --- | --- |
| MAJOR | Blocked guardrail streams lose their block text. `v3StreamToChunks` reads `value.delta`, but `createBlockedStream()` emits v1/v2 `textDelta`, so a blocked stream yields `content: ""` → trips the "Stream produced no output" sentinel instead of surfacing the block message. | `src/lib/providers/openaiChatCompletionsBase.ts:183` |

**Checked and found clean (no action needed):**
- `MiddlewareFactory.applyMiddleware` → `wrapLanguageModel` (`wrapLanguageModel.ts:56`) — `doStream` runs `transform(params, "stream")` then `wrapStream(...)`, so a prompt rewrite and a blocking `wrapStream` both route correctly; `blockingState` is hit because `doWrap` passes the same `transformed` reference through.
- `OpenAIProvider extends OpenAIChatCompletionsProvider`, so the new offline suite (`provider: "openai"`) exercises exactly this `executeStream`.
- `loopPromise ?? Promise.resolve()` correctly tolerates a blocked stream that never starts the loop.
- `tools` offered read-only with a `logger.warn` rather than a silent drop is a well-reasoned limit (re-deriving the wire tool list would diverge from `buildToolsForOpenAI`).
- The guardrails `"textDelta" in chunk` bad-word *filtering* gap is correctly flagged as a separate change — worth a tracking issue so it isn't lost.

The inline `textDelta` fix is small and high-value (fall back to `value.textDelta`); a regression test driving a **blocking** `wrapStream` against `stream()` would lock it in, since the current probe never exercises a block.

Tara-ag (changes requested):
## Changes requested

There is one MAJOR defect to fix before merge: **blocked guardrail streams lose their block text** at `src/lib/providers/openaiChatCompletionsBase.ts:183`. `v3StreamToChunks` reads `value.delta`, but `createBlockedStream()` carries its text in the v1/v2 `textDelta` field, so a guardrail **block** now surfaces as an empty stream that trips the "Stream produced no output" sentinel instead of delivering the block reason. Please apply the `value.delta ?? value.textDelta` fallback (inline suggestion) and add a regression test driving a blocking `wrapStream` against `stream()`.

Full summary with findings table: see the `<!-- yama:summary -->` issue comment on this PR.

src/lib/providers/openaiChatCompletionsBase.ts
Tara-ag: Re-verified this finding at current HEAD `c2aaf24` during a follow-up pass: it is **still open** and there is no author reply addressing it, so it remains blocking.

The precise line is `src/lib/providers/openaiChatCompletionsBase.ts:199`:

```ts
yield { content: value.delta ?? "" };
```

`createBlockedStream()` (`src/lib/middleware/utils/guardrailsUtils.ts`) still enqueues `{ type: "text-delta", textDelta: "Request contains inappropriate content and has been blocked." }` — the v1/v2 part shape. `value.delta` is `undefined` for that part, so the guardrail **block** path yields `{ content: "" }`, `contentYielded` stays 0, and the consumer gets the "Stream produced no output" sentinel instead of the block text. This is the exact scenario this PR is meant to enable, so the fix is squarely in scope:

```suggestion
        yield {
          // Middleware-authored parts (e.g. guardrails' blocked stream) still
          // carry the v1/v2 `textDelta` field rather than the V3 `delta` field.
          content:
            value.delta ?? (value as { textDelta?: string }).textDelta ?? "",
        };
```

Still requesting one more thing on top of the suggestion: a regression test that drives a **blocking** `wrapStream` (precall guardrail that returns `shouldBlock`) against `stream()` and asserts the block text reaches the consumer. The current suite's probe never blocks, so this path is untested.

murdore: Fixed at the root rather than in the reader. The `textDelta` shape came from the guardrails middleware itself: `createBlockedResponse` / `createBlockedStream` emitted v1/v2 parts and the bad-word filter keyed on `"textDelta" in chunk`, so a blocked stream surfaced as the "Stream produced no output" sentinel and the filter never matched a V3 delta. Both now emit V3 parts (`text-delta` with `delta`, `finish` with V3 usage) and the filter reads `delta`. Covered by the new stream-middleware suite — guardrail precall blocking and bad-word filtering on both `generate()` and `stream()`, with the evaluator a scripted stand-in — 20/20 against the shipped dist. Head: 65e5aee11.

Tara-ag: Confirmed at head `65e5aee`: the root fix landed cleanly. `v3StreamToChunks` now reads `part.delta`, and `createBlockedResponse`/`createBlockedStream` emit V3 parts (`text-delta` with `delta`, `finish` with V3 usage) with the bad-word filter reading `delta`. The "Stream produced no output" sentinel regression for a block is gone, and the 20-case `test/continuous-test-suite-stream-middleware.ts` suite covers precall blocking and bad-word filtering on both `generate()` and `stream()`. This adequately resolves the finding — resolving the thread.

Tara-ag (changes requested):
## NEEDS_WORK

Strengths verified and clean (no change requested): the seam is correct — `applyMiddlewareToModel()` split out of `getAISDKModelWithMiddleware` and a base model whose `doStream` starts the real multi-step loop gives the streaming path the same contract `generate()` has always had. `transformParams` runs before the prompt is converted to wire format (so a rewrite reaches the wire), `loopPromise ?? Promise.resolve()` and the tolerant readers handle a block that never starts the loop, `tools` offered read-only with a WARN is a reasonable limit, and the original `textDelta`-vs-`delta` block finding is now resolved at the root (`createBlockedResponse`/`createBlockedStream` emit V3 parts; `v3StreamToChunks` reads `delta`; 20-case stream-middleware suite covers it).

**One class of defect still blocks merge** (raised inline by CodeRabbit and independently confirmed against the current source): `applyContentFiltering` runs per text part / per text-delta chunk with no run-buffering, so a prohibited term split across adjacent text parts (`wrapGenerate`) or across `text-delta` boundaries (`wrapStream`) slips through to the consumer unchanged.

| Severity | Finding | Location |
| --- | --- | --- |
| MAJOR | Split-term BYPASS on generate path: `applyContentFiltering` applied per `text` part, so a term split across adjacent parts passes. Buffer each contiguous text run, filter once, rebuild only that run preserving order. | src/lib/middleware/builtin/guardrails.ts:93 |
| MAJOR | Split-term BYPASS on stream path: `applyContentFiltering` applied per `text-delta` chunk and enqueued immediately, so a term split across two deltas passes. Buffer the run and filter before forwarding. | src/lib/middleware/builtin/guardrails.ts:170-181 |

Both findings need a regression test covering a prohibited term split across the boundary (adjacent parts / adjacent deltas).

src/lib/middleware/builtin/guardrails.ts
```diff
@@ -89,19 +88,30 @@ export function createGuardrailsMiddleware(
...
-        result.text || "",
-        config.badWords,
-        "generate",
-      );
-      if (filterResult.hasChanges) {
-        result = { ...result, text: filterResult.filteredText };
-      }
+      result = {
+        ...result,
+        content: result.content.map((part) =>
```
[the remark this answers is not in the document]
murdore: Fixed in https://github.com/juspay/neurolink/pull/1645: text runs are filtered whole (adjacent text parts merged before filtering; stream deltas buffered per run while bad-word filtering is enabled). Both boundaries have red-first regressions in the stream-middleware suite that assert the split reassembles unfiltered without guardrails before asserting the guardrail catches it.

src/lib/middleware/builtin/guardrails.ts
```diff
@@ -147,23 +163,23 @@ export function createGuardrailsMiddleware(
...
+          if (filteredChunk.type === "text-delta") {
             const filterResult = applyContentFiltering(
-              filteredChunk.textDelta,
+              filteredChunk.delta,
               config.badWords,
               "stream",
             );
             if (filterResult.hasChanges) {
               filteredChunk = {
                 ...filteredChunk,
```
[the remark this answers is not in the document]
murdore: Fixed in https://github.com/juspay/neurolink/pull/1645: text runs are filtered whole (adjacent text parts merged before filtering; stream deltas buffered per run while bad-word filtering is enabled). Both boundaries have red-first regressions in the stream-middleware suite that assert the split reassembles unfiltered without guardrails before asserting the guardrail catches it.

PR #1639, opened 2026-09-04
ci(guards): wire the vendor-recovery and browser-bundle suites and the dts guard

murdore:
Three things this repo added recently ran only when someone ran them by hand:

- `test:vendor-recovery`, the offline suite that proves the two vendor
  recoveries ported into the native generate loop in 653d1ff69. It was the
  evidence that PR cited and it was in no CI job.
- `check:dts`, the shipped-declaration guard from #1627 (hardened in #1632).
  It reads `dist/*.d.ts`, so it now runs in the `types` shard, the one shard
  that builds.
- The browser bundle, which had no committed test at all. The smoke written
  during the SDK removal lived in a scratch directory and was never committed,
  so `dist/browser/neurolink.min.js` was back to zero coverage the moment that
  session ended. `test:browser-bundle` is that smoke, committed: the bundle
  loads in Node, the six factory exports are present and callable, each hands
  back a V3-shaped model handle synchronously, and NeuroLink is exported.

This is the same failure the extended-suites job documents — "a suite wired
into nothing is documentation, not a test" — and it is the one this repo hit
twice in the last two days, once with the error-classification suite and once
with a guard I wrote and never gated. Both new suites are offline and pass
with no credentials; they enter the extended shards at weight 5.

## Summary by CodeRabbit

* **Tests**
  * Added offline smoke-test coverage for the browser bundle.
  * Verified provider factory exports and model handles are available and callable.
  * Confirmed the main NeuroLink export remains constructible.
  * Validated generation and streaming requests through the bundled browser entry point.
  * Added declaration validation to the test workflow.
* **Chores**
  * Integrated browser-bundle and vendor-recovery suites into shard balancing.
  * Added the browser-bundle test to the available test commands.

### DIFF
--- a/test/continuous-test-suite-browser-bundle.ts
+++ b/test/continuous-test-suite-browser-bundle.ts
@@ -1,0 +1,173 @@
+#!/usr/bin/env tsx
+import "dotenv/config";
+
+/**
+ * Continuous Test Suite — browser bundle
+ *
+ * `dist/browser/neurolink.min.js` (the `@juspay/neurolink/browser` subpath)
+ * had no test at all until the Vercel AI SDK removal replaced its provider
+ * factories with native ones. The smoke written then lived in a scratch
+ * directory and was never committed, so the bundle was back to zero coverage
+ * the moment that session ended. This is that smoke, committed and wired in.
+ *
+ * It proves the bundle loads in Node, that the six factory exports the SDK
+ * used to supply are still present and callable, that each factory hands back
+ * a V3-shaped model handle synchronously, and that NeuroLink itself is still
+ * exported. It also drives generate() and stream() through that bundled
+ * NeuroLink constructor against a loopback HTTP server. No live credentials.
+ * This is a Node smoke of the browser artifact, not a browser-engine test.
+ *
+ * Run: pnpm run build && npx tsx test/continuous-test-suite-browser-bundle.ts
+ */
+
+import assert from "node:assert/strict";
+import type { NeuroLink } from "../dist/index.js";
+import { defineSuite } from "./helpers/harness.js";
+import {
+  startMockChatServer,
+  mockOpenAICredentials,
+} from "./helpers/mockChatServer.js";
+import { assertDistFresh } from "./helpers/distFreshness.js";
+
+assertDistFresh();
+
+const { test, runSuite } = defineSuite("Browser bundle", { offline: true });
+
+// Resolve at runtime: TypeScript must check the test, not re-check generated
+// minified JavaScript as source through allowJs/checkJs.
+const bundleURL = new URL("../dist/browser/neurolink.min.js", import.meta.url);
+const bundle: Record<string, unknown> = await import(bundleURL.href);
+
+const FACTORY_EXPORTS = [
+  "createAnthropic",
+  "anthropic",
+  "createOpenAI",
+  "openai",
+  "createMistral",
+  "mistral",
+];
+
+type Factory = (
+  options: Record<string, unknown>,
+) => (modelId: string) => Record<string, unknown>;
+
+await test("every provider factory is exported and callable", async () => {
+  const missing = FACTORY_EXPORTS.filter(
+    (name) => typeof bundle[name] !== "function",
+  );
+  if (missing.length > 0) {
+    throw new Error(`not exported as a function: ${missing.join(", ")}`);
+  }
+});
+
+await test("each factory returns a V3-shaped model handle synchronously", async () => {
+  const cases: Array<[string, string]> = [
+    ["createAnthropic", "claude-sonnet-4-5"],
+    ["createOpenAI", "gpt-4o-mini"],
+    ["createMistral", "mistral-small-latest"],
+  ];
+  const problems: string[] = [];
+  for (const [factory, modelId] of cases) {
+    const make = bundle[factory] as Factory;
+    const alias = factory.slice("create".length).toLowerCase();
+    const direct = bundle[alias] as (id: string) => Record<string, unknown>;
+    const handle = direct(modelId);
+    assert.equal(handle.modelId, modelId, "direct export model ID mismatch");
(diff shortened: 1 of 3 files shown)


test/continuous-test-suite-browser-bundle.ts
```diff
@@ -0,0 +1,173 @@
...
+import assert from "node:assert/strict";
+import type { NeuroLink } from "../dist/index.js";
+import { defineSuite } from "./helpers/harness.js";
+import {
+  startMockChatServer,
+  mockOpenAICredentials,
+} from "./helpers/mockChatServer.js";
+import { assertDistFresh } from "./helpers/distFreshness.js";
+
+assertDistFresh();
```
Tara-ag, line 32: `assertDistFresh()` only proves `dist/index.js` exists and that `dist/` is newer than `src/` — it never touches `dist/browser/`. Hardening this one line would make the suite's failure mode actionable if `dist/browser/neurolink.min.js` is the specific artifact missing:

```suggestion
assertDistFresh();

const bundlePath = fileURLToPath(
  new URL("../dist/browser/neurolink.min.js", import.meta.url),
);
if (!existsSync(bundlePath)) {
  throw new Error(
    "dist/browser/neurolink.min.js is missing — this suite tests the browser " +
      "bundle specifically. Run `pnpm run build` (prepack runs build:browser) first.",
  );
}
```

Rationale: this is the only suite whose entire purpose is the browser artifact, and it resolves that artifact via a dynamic `import()` at line 39. If the SDK entry is present but the browser bundle is not (e.g. `build:browser` skipped or its output dropped), the generic freshness guard passes silently and the suite instead fails at the dynamic import with a bare module-resolution error — the exact "unhelpful failure" the freshness guard's own design exists to avoid. In practice `pnpm run build` runs both and the CI cache key hashes `scripts/build-browser.mjs`, so this is defense-in-depth, not a fix for a live bug.

Tara-ag (commented):
Inline suggestion on the browser-bundle smoke test.

Tara-ag:
## Yama Review — PR #1639 `fix/ci-wire-guards`

**Validation:** Trace the wiring end-to-end — each rung is sound.

**What this does.** Wires three previously hand-run guards into CI and commits the browser-bundle smoke for the first time:

1. `test:vendor-recovery` + `test:browser-bundle` into the `extended-suites` job (weight 5 each).
2. `check:dts` into the `types` shard.
3. A new 173-line offline smoke (`test/continuous-test-suite-browser-bundle.ts`) for `dist/browser/neurolink.min.js` — bundle loads in Node, six factory exports callable, V3-shaped handles, `NeuroLink` exported, and `generate`/`stream` against a loopback server.

**Why it's put together correctly:**

- **`extended-suites` ordering is safe.** That job does **Restore → Stamp → Build** before running any suite, and its cache key hashes `scripts/build-browser.mjs`. So `dist/browser/neurolink.min.js` is present (built, or restored-and-stamped) before the suite's `assertDistFresh()` and its dynamic import run. The long-standing comment in that job — *"a suite wired into nothing is documentation, not a test"* — is exactly the failure this PR removes.
- **`check:dts` is on the right shard.** The `types` shard runs `Build` immediately after `check:test-parse` and *before* `check:tools-tests`, so dist exists by the time the new step executes. The PR's comment — *"it reads `dist/*.d.ts`, so it now runs in the one shard that builds"* — is accurate.
- **Both suites are genuinely offline.** The harness's `offline: true` (which converts a per-test timeout into a **FAIL**, since there's no upstream to blame) is used correctly. Both drive a loopback `MockChatServer`, so they belong in the credential-less `extended-suites`.
- **The new suite follows the canonical patterns** used by sibling suites (`native-vendor-recovery`, `openai-compat-streaming-retry`, etc.): `assertDistFresh()` at top, `defineSuite(...)`, proper `finally` clean-up of `sdk.shutdown()` + `server.close()`, and the `"content" in chunk` stream-consumption idiom.

**Prior thread resolution.** The existing CodeRabbit thread (*cover all six exported callables in the shape check*) is genuinely addressed in `6b859ff`: the second test now invokes the `create*` factories via `make({})(modelId)` **and** the direct aliases (`anthropic`/`openai`/`mistral`) via `direct(modelId)` — all six are exercised. That finds my review in agreement, so I've left that thread resolved rather than reposting.

**Findings**

| 1 | 🟢 Minor (defense-in-depth) | `test/continuous-test-suite-browser-bundle.ts:32` | `assertDistFresh()` verifies only `dist/index.js`; this suite specifically needs `dist/browser/neurolink.min.js`. Suggested an existence check so a missing bundle fails with an actionable message instead of a bare import error. *Not a live bug* — `build` runs `build:browser` via `prepack` and the cache key covers it. |

**Blast radius.** The graph flags ~500 impacted nodes across 130 files, but that reflects the breadth of `package.json`/`ci.yml` references rather than any real exposure — this change is three wired-in guards plus one offline smoke. No production (`src/`) call paths are touched.

**Verdict: ✅ Approve-quality.** Comments are suggestions only; nothing here blocks merge. Nice execution of the "wire it in or it's not a test" principle consistently applied elsewhere in this repo.

Tara-ag (approved):
Approving. The three guards are wired correctly into CI, the browser-bundle smoke is sound and follows the repo's canonical suite patterns, and the prior CodeRabbit thread was genuinely addressed in 6b859ff. Only a defense-in-depth suggestion remains on `assertDistFresh()`; nothing here blocks merge.

test/continuous-test-suite-browser-bundle.ts
murdore: Adopted, in #1652 — that PR was already editing this suite. The suite now resolves `dist/browser/neurolink.min.js` and throws naming `build:browser` if it is absent, so a present-but-partial build fails with the artifact named instead of an opaque dynamic-import error. Thanks — this was a real hole: `assertDistFresh()` never looks at `dist/browser/` at all.

PR #1645, opened 2026-09-05
fix(middleware): filter guardrail text runs whole, not per part or per delta

murdore:
## What

The built-in guardrails bad-word filter ran on each V3 text part in `wrapGenerate` and on each `text-delta` in `wrapStream`, in isolation. A prohibited term split across a boundary — `inappro` + `priate` — was invisible to both, and both boundaries are reassembled downstream: `lifecycle.ts` concatenates adjacent text parts, and every stream consumer concatenates deltas. The term reached the caller whole.

Both were flagged by CodeRabbit on the merged head of #1636 (`guardrails.ts:93` and `:181`). Neither is theoretical:

- **Generate.** Anthropic's native path emits one text part per content block, so adjacent text parts arrive whenever a reply has two text blocks.
- **Stream.** A token boundary can fall anywhere inside a word, so a per-delta filter could be defeated by chunking alone. This half is the filter's original design; #1636 made it reachable.

## How

- `filterTextRuns` filters each contiguous run of text parts as one string, keeps non-text parts in place, and rebuilds a run as a single text part only when the filter changed it.
- `wrapStream` buffers a text run while bad-word filtering is **enabled** and releases it, filtered once, when a non-text part arrives or the stream ends. That trades incremental delivery of the run for a guardrail that chunking cannot bypass. With filtering off, deltas pass through untouched and unbuffered, exactly as before.
- `LanguageModelV3Content` is now re-exported alongside `LanguageModelV3StreamPart` so the helper can type against the barrel (rule 13).

## Proof (red first)

Two cases in `test/continuous-test-suite-stream-middleware.ts`, each asserting its **precondition before its claim**: without guardrails the two pieces reassemble into the whole term at the consumer (so a pass cannot come from the pieces never having been split); with guardrails the consumer sees the replacement.

| case | drives | fixture | unfixed | fixed |
| split across stream deltas | `stream()` → OpenAI-compatible path | local server emitting the term as two SSE deltas | ✗ | ✓ |
| split across adjacent text parts | `generate()` → native Anthropic path | local Messages server answering with two text blocks | ✗ | ✓ |

RED on the unfixed build: 20 passed, exactly these 2 failed. GREEN: 22/22 (the 20 existing cases plus these 2), with typecheck, build, lint and the test typecheck clean; `docs/api` regenerated with the prettier pass CI applies (44 anchor shifts in `middleware.ts` pages plus the new `LanguageModelV3Content` page).

## Trade-off worth knowing

With bad-word filtering enabled on a stream, the consumer receives each text run when it completes rather than delta by delta. A bounded look-behind (hold back only the longest configured word) would restore incremental delivery for plain word lists but cannot be made correct for `regexPatterns`, so this PR takes the simple, correct behaviour and leaves that optimisation as a follow-up if latency under filtering matters to anyone.

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -458,6 +458,7 @@ console.log(result.content);
 - [LanguageModelV3ToolCall](type-aliases/LanguageModelV3ToolCall.md)
 - [LanguageModelV3Source](type-aliases/LanguageModelV3Source.md)
 - [LanguageModelV3ToolChoice](type-aliases/LanguageModelV3ToolChoice.md)
+- [LanguageModelV3Content](type-aliases/LanguageModelV3Content.md)
 - [LanguageModelV3GenerateResult](type-aliases/LanguageModelV3GenerateResult.md)
 - [LanguageModelV3StreamPart](type-aliases/LanguageModelV3StreamPart.md)
 - [LanguageModelV3StreamResult](type-aliases/LanguageModelV3StreamResult.md)
--- a/docs/api/type-aliases/AbortSignalMiddlewareOptions.md
+++ b/docs/api/type-aliases/AbortSignalMiddlewareOptions.md
@@ -8,7 +8,7 @@
 
 > **AbortSignalMiddlewareOptions** = `object`
 
-Defined in: [types/middleware.ts:396](https://github.com/juspay/neurolink/blob/release/src/lib/types/middleware.ts#L396)
+Defined in: [types/middleware.ts:397](https://github.com/juspay/neurolink/blob/release/src/lib/types/middleware.ts#L397)
 
 Options for the abort-signal middleware.
 
@@ -18,7 +18,7 @@ Options for the abort-signal middleware.
 
 > `optional` **onAbort?**: (`ctx`) => `void`
 
-Defined in: [types/middleware.ts:397](https://github.com/juspay/neurolink/blob/release/src/lib/types/middleware.ts#L397)
+Defined in: [types/middleware.ts:398](https://github.com/juspay/neurolink/blob/release/src/lib/types/middleware.ts#L398)
 
 #### Parameters
 
@@ -36,4 +36,4 @@ Defined in: [types/middleware.ts:397](https://github.com/juspay/neurolink/blob/r
 
 > `optional` **timeout?**: `number`
 
-Defined in: [types/middleware.ts:398](https://github.com/juspay/neurolink/blob/release/src/lib/types/middleware.ts#L398)
+Defined in: [types/middleware.ts:399](https://github.com/juspay/neurolink/blob/release/src/lib/types/middleware.ts#L399)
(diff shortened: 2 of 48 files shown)


Tara-ag:
## Summary — APPROVE

The guardrails split-term fix is correct: it closes a real bypass (a prohibited term straddling text parts / deltas) with whole-run filtering, keeps the filtering-off path byte-for-byte passthrough, and is covered end-to-end with precondition-first cases.

### Findings
| Severity | Location | Note |
|---|---|---|
| MINOR | `src/lib/middleware/builtin/guardrails.ts` (wrapStream buffering) | Acknowledged trade-off, not a defect: with `badWords.enabled`, a contiguous run of `text-delta`s is buffered until the run ends. A single-paragraph response (e.g. a long essay/code block) has no non-text part to trigger an early release, so it is emitted at stream end — incremental delivery for that run is traded for a chunking-proof guardrail. Documented in the PR; the bounded look-behind follow-up is reasonable to defer (regexPatterns can span the look-behind). |
| — | `src/lib/types/middleware.ts` | `LanguageModelV3Content` is now re-exported from `./aiCompat.js` alongside `LanguageModelV3StreamPart`; additive, type-only, and flows through the `types/index.ts` wildcard barrel — backward compatible (Rule 5 satisfied). |
| — | `test/continuous-test-suite-stream-middleware.ts` | Two new cases (stream deltas / adjacent text parts) are end-to-end via `../dist/index.js`, and each asserts its precondition (the split reassembles to the whole term with NO guardrail) before asserting the filtered outcome — exactly the red-first discipline required. |

### Checked and found clean
- **Rule 1 — dynamic imports only in the registry.** `resolveFilterModel` lazily `import()`s the provider factory; no static provider import introduced.
- **Rule 5 — backward compatibility.** No exported signature changed; `filterTextRuns` is module-private; only internal `wrapStream`/`wrapGenerate` behavior changed.
- **Rule 15 — e2e tests.** Runtime imports come only from `dist/`; the `src/lib/types/index.js` import is type-only (erased) and declared in the file header.
- **Stream correctness.** The buffer respects `text-start`/`text-end` boundaries (non-`text-delta` chunks end a run and preserve ordering), interleaved non-text parts stay ordered, buffered text is not lost, and `flush` still warns when nothing yielded.
- **Filtering-off parity.** With `badWords` absent/disabled, `wrapStream` emits every chunk immediately and unchanged.
- **Send / analytics.** The V3 `finish` part flows through; analytics still settle (existing test asserts exactly one `finish`).
- **Security.** No secrets in the change; test keys (`sk-mock-local-server`, etc.) are documented placeholders, not real credentials.

No inline blocking comments are warranted by this change.

Tara-ag (approved):
Review of PR #1645 — `fix(middleware): filter guardrail text runs whole, not per part or per delta`

The guardrails split-term fix is correct: it closes a real bypass (a prohibited term straddling text parts / deltas) with whole-run filtering, keeps the filtering-off path byte-for-byte passthrough, and is covered end-to-end with precondition-first cases. No blocking findings.

Approve. See the full summary comment for details.

PR #1648, opened 2026-09-06
feat(providers): onboard Morph as a catalog provider

murdore:
## Summary

Morph's payment wall cleared. Its recorded wire contract gives it plain chat, SSE streaming and genuinely strong structured output, so it earns a catalog entry — with tool calling deliberately declared **false**.

## Tool calling is false on purpose

Morph answers **200** when sent `tools`, which is exactly the trap that makes an optimistic flag easy to write. It never returns a `tool_calls` array:

```
morph-v3-large  ->  content: "<tool_call>\n{\"name\": \"get_time\", ...}\n</tool_call>"   (literal text)
morph-v3-fast   ->  content: Python source; the tool is ignored
```

Verified across both models at two token budgets, all four runs `finish_reason=stop` with no `tool_calls`. An OpenAI-compatible client would surface that as assistant prose. An earlier campaign note claimed Morph tools worked; it cited models that are no longer in the 22-id roster, so it was stale.

## Message content must be a plain string

Array content parts return **HTTP 500** `Request processing failed: text.charCodeAt is not a function` — the same class of quirk `cloudflare.json` already handles, and expressed the same way. The vision probe fails identically, so vision is not claimed.

A live call proves the fix: it succeeds with no 500, which is only possible if our client sent string content.

## Capability flags and their evidence

| Flag | Value | Why |
| --- | --- | --- |
| `tools` / `toolsWithStreaming` | false | 200 but no `tool_calls`, on both models, at two budgets. |
| `structuredOutput` | true | `json_schema` returned valid schema-conformant JSON; `json_object` accepted. |
| `structuredOutputWithTools` | false | Meaningless when tools do not work; not claimed on the strength of a 200. |
| vision | not claimed | Fails with the array-content 500. |

Only the two models the vendor's own error text lists as accepted by `chat/completions` are catalogued. The other 18 roster ids were never live-verified against chat, so they are omitted rather than invented.

## Live proof against the built package

Generate returns `ready`, streaming yields `1 2 3`, a schema call returns `{"color":"blue"}` with `structuredData` parsed, and the declared default `morph-v3-large` is confirmed to be the model that actually resolves.

`MORPH_API_KEY` is wired into the nightly live matrix. That was missing from the first cut and would have silently excluded Morph from nightly verification.

## Gates

`check`, `lint` (0 errors), `build`, catalog codegen drift check (16 providers), `test:providers-mocked` 98/98, `test:provider-structure` 3/3, `test:openai-compat-catalog` 43/43 including the new Morph alias cases, `test:error-classifier-contract` 44/44, `verify:provider-onboarding` 9/9, `docs:api` regenerated. Break-one ritual on the new case: inverted reports a failure with a non-zero exit, restored passes.

The catalog alias suite is hand-maintained rather than derived from the catalog, so Morph was silently absent from it at first; the cases were added and the suite re-run.


Tara-ag (commented):
## Review: Morph catalog onboarding

**Verdict: APPROVE** — clean catalog-only onboarding; the two "traps" (tool calling 200-with-no-tool_calls, array content → 500) are both correctly reflected as data flags that the existing catalog machinery genuinely enforces (`supportsTools` off keeps `tools` off the wire; `messageContentFormat:"string"` is consumed by the loader + `ConfiguredOpenAICompatProvider.adjustRequestBody`), with test coverage and nightly env wiring included.

|---|---|---|
| MINOR | `src/lib/providers/catalog/morph.json:12` | `morph-v3-fast` is a code-edit specialist that the PR's own evidence says returns unrelated text when used for chat; it is the secondary fallback/`fallbackModelName` |

**Checked and clean**
- `messageContentFormat:"string"` quirk → `loader.ts` reads `entry.quirks.messageContentFormat` → `ConfiguredOpenAICompatProvider.adjustRequestBody` (`src/lib/providers/configuredOpenAICompat.ts`). Same path as Cloudflare. ✓
- `tools:false` / `toolsWithStreaming:false` → `loader.ts` maps `capabilities.tools` → `supportsTools`; `executeStream`/generate gate on `supportsTools()` so no `tools` array ever reaches the Morph wire. ✓
- `errorRules` 401/400 with `{model}`/`{apiKeyEnvVar}`/`{setupUrl}` placeholders → `buildErrorRules` interpolates them. ✓
- Credential/env resolution: `catalogEnvVar("morph", …)` → `MORPH_API_KEY`/`MORPH_BASE_URL`/`MORPH_MODEL`, matching the nightly `MORPH_API_KEY` wiring. ✓
- `AIProviderName.MORPH` + `NeurolinkCredentials.morph?` + `CatalogProviderName/CatalogCredentialKey` are all additive (rule 5 backward-compatible); no hand registration needed beyond catalog + registry loop.
- `addedInPR:"pending"` matches the convention for pre-merge catalog entries (mancer, gmicloud, io-intelligence …). Not a finding.
- Generated `index.generated.ts` / `providerCatalog.generated.ts` + `docs/api/**` are codegen/CI-owned (skipped per repo conventions).
- Alias tests for `morph`/`morphllm` in the openai-compat catalog suite match the catalog (URL `api.morphllm.com/v1`, default `morph-v3-large`).

**Impact-analysis note**: I could not run the code-review-graph blast-radius tools in this environment (the local checkout is not indexed and the sandbox blocked filesystem access outside `.yama/`). Blast radius was instead assessed from the diff plus remote source inspection via the GitHub API. The change is a purely additive catalog/enum registration touching only the generic catalog+registry machinery, so no out-of-diff callers are expected to break.

Minor doc nit (not blocking): the PR body states all four tool probes ended `finish_reason=stop`, but `morph.json`'s `evidence.liveMatrix.result` records `morph-v3-fast` returning `finish_reason=length`. Worth reconciling so the evidence stays internally consistent.

src/lib/providers/catalog/morph.json
```diff
@@ -0,0 +1,97 @@
...
+  "id": "morph",
+  "displayName": "Morph",
+  "aliases": ["morphllm"],
+  "tier": 2,
+  "wire": {
+    "baseURL": "https://api.morphllm.com/v1"
+  },
+  "models": {
+    "default": "morph-v3-large",
+    "fallbacks": ["morph-v3-large", "morph-v3-fast"],
```
Tara-ag, line 12: **MINOR** — `morph-v3-fast` is a code-edit (Fast Apply) specialist, not a general-chat model, yet it is the secondary chat fallback.

`loader.ts` derives `fallbackModelName` as `fallbacks[1] ?? fallbacks[0]` → here that resolves to `morph-v3-fast`. Your own `evidence.liveMatrix.result` in this file says v3-fast "ignores the tools parameter outright and returns unrelated generated text (Python source)" and its description calls it "purpose-built for applying code edits rather than general chat". So a general-chat request that falls through to the fallback (e.g. the default `morph-v3-large` is rejected/retired, or `/models` discovery is empty) can get unrelated output served as if it were a first-class chat answer.

The practical blast radius is small — the fallback only engages on invalid-model / discovery-empty paths, not on transient errors — but worth confirming `morph-v3-fast` should serve general chat at all. If it shouldn't, consider dropping it from `fallbacks` (and from `topModels`) so chat calls fail loudly on an invalid model rather than silently returning code-edit garbage.

```suggestion
    "fallbacks": ["morph-v3-large"],
```

Tara-ag (commented):
Minor design note on the Morph fallback chain; onboarding otherwise looks correct (verified quirks/capabilities/error-rules wiring against the catalog loader and OpenAI-compat provider code).

Tara-ag (approved):
Approving per the full review in the summary comment (`<!-- yama:summary -->`): clean catalog-only onboarding with the two mechanics flags correctly enforced and test/CI coverage included. The single non-blocking MINOR finding (`morph-v3-fast` as the chat fallback in `morph.json:12`) is tracked in the inline thread and does not block merge.

Tara-ag (approved):
## Recurring review — Morph catalog onboarding

**Verdict: APPROVE** — re-confirmed on current head `364c6a6`. The head commit SHA moved since the last review (cbee555e → 364c6a6) but the diff content is identical; the catalog-only onboarding remains clean, additive, and the two "trap" flags (`tools:false`, `messageContentFormat:"string"`) are genuinely enforced by the catalog loader + `ConfiguredOpenAICompatProvider` machinery. No new findings.

|---|---|---|---|
| MINOR | `src/lib/providers/catalog/morph.json:12` | `morph-v3-fast` is a code-edit specialist used as the secondary chat fallback — a general-chat request can get unrelated output when the primary model is unavailable | Retained (inline thread) |
| MINOR | PR body / `morph.json` `evidence.liveMatrix.result` | PR body says all four tool probes ended `finish_reason=stop`, but the evidence records `morph-v3-fast` at `finish_reason=length`; reconcile for internal consistency | New (doc nit, non-blocking) |

**On the retained finding:** the inline thread (marker `yama:morph-fallback`) has no author reply, and the evidence in `morph.json` still confirms `morph-v3-fast` "ignores the tools parameter and returns unrelated generated text" and is "purpose-built for applying code edits rather than general chat". So it is kept as a MINOR design note, not elevated — the fallback only engages on invalid-model / discovery-empty paths, defaults resolve to `morph-v3-large`.

**Checked and re-verified clean on this pass**
- `tools:false` / `toolsWithStreaming:false` → `supportsTools()` gate keeps `tools` off the Morph wire. ✓
- `messageContentFormat:"string"` → `adjustRequestBody` string-content path (same quirk as Cloudflare). ✓
- Additive surface only: `AIProviderName.MORPH`, `NeurolinkCredentials.morph?`, `CatalogProviderName`/`CatalogCredentialKey`, `MorphModels` — no exhaustive switch / hand-registration breakage; `detect_changes` risk 0.00, no affected flows. ✓
- `errorRules` 401/400 placeholders + credential env resolution → `MORPH_API_KEY`/`MORPH_BASE_URL`/`MORPH_MODEL`. ✓
- Alias tests for `morph`/`morphllm` match the catalog (URL `api.morphllm.com/v1`, default `morph-v3-large`); nightly `MORPH_API_KEY` wired. ✓
- `docs/api/**`, `llms-full.txt`, `index.generated.ts`, `providerCatalog.generated.ts` are codegen/CI-owned (skipped per repo conventions). ✓
- Check runs on the current head include the review agent; prior gates (`test:providers-mocked`, `test:provider-structure`, catalog codegen drift, 43/43 alias suite) plus `Single Commit Policy` and `Generated artifacts are current` pass.

src/lib/providers/catalog/morph.json
murdore: Confirmed as a real hazard, not a nit. I checked the derivation rather than reasoning from the comment.

`loader.ts:127-130`:
```ts
fallbackModelName:
  entry.models.fallbackModelName ??
  entry.models.fallbacks[1] ??
  entry.models.fallbacks[0],
```

`morph.json` sets no explicit `fallbackModelName`, and its `fallbacks` are `["morph-v3-large", "morph-v3-fast"]` — so the derived fallback is `morph-v3-fast`, exactly as you said. That is the model this same file documents as ignoring the tools parameter outright and returning unrelated generated text, purpose-built for applying code edits rather than general chat. A general-chat request that falls through gets prose it cannot use, with nothing to indicate why.

**The mechanism to fix it already exists and is used elsewhere.** Three other catalogs set the field explicitly for what looks like this reason:

| catalog | explicit `fallbackModelName` | what `fallbacks[1]` would have given |
| --- | --- | --- |
| cloudflare | `@cf/meta/llama-3.1-8b-instruct-fast` | `@cf/meta/llama-3.1-70b-instruct` |
| fireworks | `accounts/fireworks/models/gpt-oss-120b` | same |
| mistral | `mistral-small-2506` | `mistral-large-latest` |

So the one-line fix is:

```json
"models": {
  "default": "morph-v3-large",
  "fallbackModelName": "morph-v3-large",
  ...
}
```

pointing the fallback at the chat-capable model instead of the code-edit specialist. Keeping `morph-v3-fast` in `fallbacks` is still right — it belongs in the roster, just not as the chat fallback.

Not pushing it unasked: this PR is approved and green, and a catalog data change of this kind cannot carry a discriminating test without asserting the fix against itself. I would rather the owner apply it deliberately than have it ride in on a standing approval that was granted for tested work. Happy to commit it here or fold it into a separate catalog-fix PR — whichever is less disruptive.

murdore:
## Live acceptance run — SDK and CLI, real vendor

Run from this PR's worktree at head `364c6a64d`, `dist/` built from that head (0 source files newer), public surface only.

```
provider=morph  model=morph-v3-large
node=v24.14.1

SDK generate  -> 1547ms  content="HELLO"
              provider=morph model=morph-v3-large usage={"input":104,"output":3,"total":107}
SDK stream    -> 6028ms  chunks=5  text="1\n2\n3"
CLI generate  -> 1947ms  exit=0  stdout="HELLO"
CLI stream    -> 6864ms  exit=0  stdout="HELLO"

SUMMARY sdk.generate=true sdk.stream=true
```

All four surfaces pass.

## And a correction to my own earlier comment on the fallback thread

While running this I also probed the no-model path, and **it contradicts what I claimed** in https://github.com/juspay/neurolink/pull/1648#discussion_r3990764420. Details are on that thread; the short version is that the default model resolves to `morph-v3-large`, not `morph-v3-fast`:

```
NO-MODEL generate -> resolved model=morph-v3-large
                    content="HELLO"
```

I had read `loader.ts`'s `fallbackModelName ?? fallbacks[1] ?? fallbacks[0]` and concluded that an ordinary call lands on the code-edit specialist. It does not — that line feeds `getFallbackModelName()`, which is a different accessor from `getDefaultModel()`, and `morph.json` sets `"default": "morph-v3-large"`. Reading the derivation was not the same as running it.

src/lib/providers/catalog/morph.json
murdore: **Correcting myself: I ran this instead of reading it, and the claim I made above does not hold.**

I said the chat fallback resolves to `morph-v3-fast`, the code-edit specialist, and proposed `"fallbackModelName": "morph-v3-large"` as a one-line fix. A live probe against the vendor from this PR's worktree says otherwise:

```
NO-MODEL generate -> resolved model=morph-v3-large
                    content="HELLO"
EXPLICIT v3-large -> content="HELLO"
```

**What I got wrong.** `loader.ts:127-130` does compute `fallbackModelName: entry.models.fallbackModelName ?? fallbacks[1] ?? fallbacks[0]`, which for morph is `morph-v3-fast` — that part was right. But that value feeds `getFallbackModelName()` in `configuredOpenAICompat.ts:107`, which is a **different accessor** from `getDefaultModel()` at :106. The default model comes from `entry.defaultModel` → `models.default`, and `morph.json` sets that to `morph-v3-large`. So an ordinary call was never landing on the specialist. I verified the derivation of one line and assumed it drove a behaviour it does not drive.

**What survives.** `fallbackModelName` is still `morph-v3-fast`, so the *failure* path — primary unavailable, SDK retries the named fallback — does land on a model whose own catalog description says it "ignores the tools parameter outright and returns unrelated generated text". That is narrower than "the general-chat fallback", and it is worth knowing.

**But the fix I proposed is probably wrong, and I want to flag that before anyone applies it.** Morph has exactly two models. Setting `fallbackModelName: "morph-v3-large"` makes the fallback identical to the default, so a failure of `morph-v3-large` would retry `morph-v3-large` — a fallback that cannot fall back. Given both models are documented as mishandling tools (`v3-large` emits the call as literal `<tool_call>` text; `v3-fast` ignores the parameter), the fallback is not materially worse for tool use either way, and for plain chat a degraded answer beats no answer.

My recommendation is now: **leave `morph.json` as it is.** If you want the failure path to be explicit rather than derived, the honest change is a comment in the catalog saying the fallback is deliberately the only other model, not a `fallbackModelName` that disables falling back.

Leaving this thread open — the original finding was partly right, my fix was wrong, and which way to go is your call rather than mine.

murdore: You're right that a general-chat request can end up on `morph-v3-fast`, but the mechanism is a different one, and it matters because it changes what the fix would have to be.

`loader.ts` does derive `fallbackModelName = fallbacks[1] ?? fallbacks[0]` → `morph-v3-fast` here. But that accessor never fires for this provider: `resolveModelName()`'s fallback branch only runs when `getDefaultModel()` is empty, and Morph's is always `morph-v3-large`; the only other reader, `getFirstAvailableModel()`, has no callers in `src/` today.

The live path is `getModelFallbacks()` — the whole `fallbacks` array — used by `runGenerateWithModelFallback` / `retryStreamWithFallbackModel` in `baseProvider.ts` when an `InvalidModelError` is caught. That is exactly the class Morph's own `errorRules` raises for "is not served by this endpoint", so your scenario reproduces on **both** `generate()` and `stream()`, and because Morph has only two catalog models it lands on `morph-v3-fast` regardless of array order.

I tried the obvious catalog fix and it does not work. Trimming to `"fallbacks": ["morph-v3-large"]` makes the derived fallback equal `models.default`, which the new cross-catalog rule in #1690 rejects whenever a catalog has more than one non-retired model — both of Morph's are `"production"`, so that edit fails validation. Marking `morph-v3-fast` retired would be false (it is a working model for edit application), and deleting it would remove a legitimately selectable model to fix fallback arithmetic.

So there is no catalog-only fix. The missing capability is a way to mark a model **manually selectable but excluded from automatic invalid-model retry** — `getModelFallbacks()` has no such notion. I am leaving `morph.json` as it stands and tracking that opt-out as follow-up rather than blocking this PR or forcing a change into #1690.

test/continuous-test-suite-openai-compat-catalog.ts
```diff
@@ -1105,6 +1105,18 @@ const CATALOG_ALIAS_CHECKS: AliasCheck[] = [
     urlMatch: "neuro.mancer.tech/oai/v1/chat/completions",
     model: "deepseek-v4-flash",
   },
+  {
+    alias: "morph",
```
Tara-ag, line 1121: **MINOR** — Morph's env var was added to the alias checks but is missing from the `CATALOG_ENV_VARS` neutralize list.

`neutralizeCatalogEnv()` (line ~142) iterates `CATALOG_ENV_VARS` to delete every catalog provider's real credential before the suite runs, so a developer machine's `.env` can never leak a live key into a mocked request. Every other catalog provider's vars are enumerated there — including `MANCER_*` added the same way — but `MORPH_API_KEY` is not. The new `morph`/`morphllm` alias cases happen to set `setEnv(check.envVar, ...)` immediately before use, so today it is masked; but the suite's own header documents the invariant as "env vars are always either set to a known fake value or deleted outright". The moment any future morph case reads the env var without first overriding it, a real `MORPH_API_KEY` from `.env` would silently flow into a mocked request.

Please add `"MORPH_API_KEY"` to the `CATALOG_ENV_VARS` array (near the `MANCER_*` entries), matching the pattern used for every other provider.

```suggestion
  "MANCER_API_KEY",
  "MANCER_BASE_URL",
  "MANCER_MODEL",
  "MORPH_API_KEY",
];
```

Tara-ag (commented):
Reviewed Morph catalog onboarding. Clean PR; one MINOR follow-up on the test suite's env-var neutralization. Details in the summary comment.

Tara-ag:
## Verdict: APPROVE

Morph is a well-evidenced catalog onboarding. Capability flags are grounded in live wire probes (tools deliberately `false` for both models, `structuredOutput` true, vision not claimed), the catalog reuses the existing `messageContentFormat: "string"` quirk exactly as `cloudflare.json` does, and all generated surfaces (`index.generated.ts`, `providerCatalog.generated.ts`, `MorphModels`, `AIProviderName.MORPH`) are regenerated and drift-consistent. Provider count gate, lint, build and all listed test suites are green.

| --- | --- | --- |
| MINOR | `test/continuous-test-suite-openai-compat-catalog.ts:1109` | `MORPH_API_KEY` added to the alias checks but missing from `CATALOG_ENV_VARS` neutralize list — breaks the suite's documented "no real creds in mocked tests" invariant |

### Existing thread — accepted (resolved)
- `yama:morph-fallback` (`morph.json:12`): the author's follow-up investigation is thorough and conclusive. No catalog-only fix exists (trimming `fallbacks` to `[morph-v3-large]` fails the new #1690 cross-catalog validation; marking `morph-v3-fast` retired would be false; deleting it removes a legitimately selectable model; and `fallbackModelName: morph-v3-large` would disable falling back entirely). The live probe also confirmed the *default* resolves to `morph-v3-large`, not the specialist. Leaving `morph.json` as-is is justified — **not reposted**, thread resolved.

### Checked and found clean
- `morph.json` structure vs `cloudflare.json`: identical quirks/capabilities/errorRules/setup/evidence convention; `enumMember` correctly omitted (model ids auto-derive to valid identifiers).
- Generated wiring consistent: `CATALOG_PROVIDER_IDS`, `CatalogProviderName`, `CatalogCredentialKey` all include `morph`.
- Alias routing tests for `morph` / `morphllm` added using `MORPH_API_KEY` and the correct host.
- `live-matrix.yml` wires `MORPH_API_KEY` (the missing-from-first-cut item the PR calls out).
- Blast radius: additive provider + enum; no existing SDK surface modified (rule 5 safe); `providers.ts` changes are line-shift churn only.

No CRITICAL / MAJOR findings.

test/continuous-test-suite-openai-compat-catalog.ts
Tara-ag, line 1121: **MINOR** — `MORPH_API_KEY` is documented for the new alias checks but is missing from the `CATALOG_ENV_VARS` neutralize list, so the suite's hermeticity invariant no longer holds.

The header contract of this file (lines 66–72, 76–95: `dotenv/config` loads the repo `.env`, which "on a developer machine may carry real … values") requires every catalog provider's credential vars to be deleted before the run, so no real `.env` credential can ever reach a mocked request. `MANCER_API_KEY`/`MANCER_BASE_URL`/`MANCER_MODEL` were added when mancer landed, but the `MORPH_*` trio was not. The Morph alias cases happen to set a fake `MORPH_API_KEY` themselves today, so nothing leaks *right now* — but the invariant is broken, and any future case that misses an override could send a real environment credential on the wire. It is the same latent gap `CATALOG_ENV_VARS` exists to prevent.

Fix: add the three `MORPH_*` vars to `CATALOG_ENV_VARS` (next to the `MANCER_*` entries at the end of the array):

```ts
  "MANCER_MODEL",
  "MORPH_API_KEY",
  "MORPH_BASE_URL",
  "MORPH_MODEL",
```

(A suggestion block can't be attached here because `CATALOG_ENV_VARS` itself is not part of the diff; the anchor is the nearest changed line.)

Tara-ag (approved):
Approving per the full review in the summary comment (`<!-- yama:summary -->`). Clean additive catalog onboarding; the two non-blocking MINOR findings (`morph-v3-fast` as chat fallback — author-justified, thread resolved — and the `MORPH_*` env-var neutralize gap) do not block merge.

Tara-ag:
## Review: Morph catalog onboarding — final

**Verdict: APPROVE** — clean, additive catalog-only onboarding; both "trap" flags (`tools:false`, `messageContentFormat:"string"`) are genuinely enforced by the existing catalog machinery, and the two findings are non-blocking (one author-justified, one MINOR).

|---|---|---|---|
| MINOR | `test/continuous-test-suite-openai-compat-catalog.ts:1109` | `MORPH_*` env vars added to the alias checks but not to the `CATALOG_ENV_VARS` neutralize list — latent hermeticity gap | Inline thread (posted this review) |
| MINOR | `src/lib/providers/catalog/morph.json:12` | `morph-v3-fast` is a code-edit specialist; it is the secondary fallback/`fallbackModelName` | Retained, accepted — author proved no catalog-only fix exists (inline thread, resolved) |

**Checked and clean**
- `morph.json` follows the `cloudflare.json` convention (`quirks.messageContentFormat:"string"`, same `capabilities`/`setup`/`errorRules`/`evidence` blocks); `enumMember` correctly omitted (identifiers auto-derive). ✓
- Generated wiring drift-free: `CATALOG_PROVIDER_IDS`, `CatalogProviderName`, `CatalogCredentialKey` (`morph`), `AIProviderName.MORPH`, `MorphModels`. ✓
- Rule 5 backward-compat safe: purely additive surface, no exhaustive-switch or hand-registration breakage; no out-of-diff callers affected (additive provider + enum only). ✓
- Alias tests for `morph`/`morphllm` match the catalog (URL `api.morphllm.com/v1`, default `morph-v3-large`); nightly `MORPH_API_KEY` wired in `live-matrix.yml`. ✓
- `docs/api/**`, `index.generated.ts`, `providerCatalog.generated.ts` are codegen/CI-owned (skipped per repo conventions). ✓

**Validation note:** blast radius assessed from the diff + remote source inspection (code-graph tools unavailable in this sandbox). Change is additive and touches only generic catalog/enum machinery, so no breakage is expected.

**Note on summaries:** this is the single canonical summary for the PR (issue comment). Earlier identical summaries embedded in two review bodies on older commits (cbee555e, 364c6a6) are superseded by this one.

src/lib/providers/catalog/morph.json
Tara-ag: Accepted — your investigation is thorough and the conclusion is sound. You verified empirically that the chat default resolves to `morph-v3-large` (not the specialist), corrected your own proposed fix (a `fallbackModelName: morph-v3-large` would just disable falling back), and established that no catalog-only edit passes the #1690 cross-catalog validation. Leaving `morph.json` as-is and tracking the "manually selectable but excluded from auto-retry" opt-out as follow-up is the right call. Resolving this thread; the finding stands as a non-blocking MINOR design note only.

test/continuous-test-suite-openai-compat-catalog.ts
murdore: Added — `MORPH_API_KEY`, `MORPH_BASE_URL` and `MORPH_MODEL` now sit in `CATALOG_ENV_VARS` alongside every other catalog provider's triple. An audit of the file confirms the list is complete again:

```
neutralized: 42 | used in alias checks: 14
MISSING from neutralize list: (none)
```

One thing I want to be straight about rather than dress up: **I could not produce a case where this changes behaviour today.** I ran the suite with a deliberately polluting `MORPH_API_KEY` exported, both with and without the fix, and got `43 passed · 0 failed` either way.

The reason is that every alias check calls `setEnv(check.envVar, …)` before it runs (line 1129), so it overwrites whatever the environment had. And the only tests that assert an *unconfigured* provider are Cloudflare's `missing accountId` cases, which do not touch Morph.

So this is hygiene, not a bug fix: it restores the neutralize contract so that a future test asserting "Morph without a key errors correctly" cannot be silently defeated by an ambient key on a contributor's machine. Worth doing for the reason you give — but it is not backed by a discriminating test, and I would rather say so than imply coverage I do not have.

murdore:
Follow-up checked and closed out — nothing to carry to **https://github.com/juspay/neurolink/pull/1728**.

The remaining item was Tara-ag's doc-consistency note, raised twice (once in the first review, once in the "Recurring review" table as *New (doc nit, non-blocking)*) and then dropped from the final approval summary: the PR body says all four tool-calling probes ended `finish_reason=stop`, while `src/lib/providers/catalog/morph.json:93` records `morph-v3-fast` returning unrelated generated text at `finish_reason=length`.

Checked both at the pinned head. They do contradict — but `morph.json` holds the *accurate* record, and it is the artifact that survives the merge; the PR body is the part that is wrong, and it disappears. `capabilities.toolCalling: false` is consistent with the evidence text either way, so the catalog entry needs no change.

Recording it here so the discrepancy is on file rather than silently dropped. #1728 carries only the three items that describe durable code.

PR #1650, opened 2026-09-06
fix(providers): replace NVIDIA NIM's decommissioned default model

murdore:
## What

NVIDIA retired `meta/llama-3.3-70b-instruct` on 2026-08-26. It was this provider's default in **four** separate places, so selecting `nvidia-nim` without naming a model failed outright — on `generate()` and `stream()` alike:

```
[nvidia-nim] The model 'meta/llama-3.3-70b-instruct' has reached its end of
life on 2026-08-26T09:00:00Z and is no longer available.
```

## Why four edits

The default lives in `getDefaultNimModel()` (`nvidiaNim/client.ts`), `providerRegistry.ts`, `providerDescriptors.ts` and `modelChoices.ts`. Changing any one of them leaves the behaviour identical, which is exactly what happened twice while tracking this down.

## The replacement

`openai/gpt-oss-20b`. Most of NIM's published roster is not usable either — it answers `404 "not found for account"` — so the replacement was chosen from what an account can actually reach, and probed live for every property something in this repo depends on:

| property | result |
| --- | --- |
| plain text | ✓ |
| streaming (SSE) | ✓ |
| tool calling | ✓ returns `tool_calls` |
| structured output (`response_format`) | ✓ valid JSON |
| fits `maxTokens: 200` | ✓ `finish_reason: stop`, 49 completion tokens |

Retired enum members are kept so existing callers still compile.

## Test rot from the same decommission

Four matrix cells and three `new-providers` pins referenced models that no longer exist — `meta/llama-3.1-8b-instruct`, `deepseek-ai/deepseek-r1-distill-llama-70b`, `meta/llama-3.2-90b-vision-instruct`, `meta/llama-3.3-70b-instruct`. Each moves to a model that resolves and carries the property its case needs: `gpt-oss-20b` emits `reasoning_content` and calls tools; `llama-3.2-11b-vision-instruct` is the vision route.

## Proof

Run from an empty directory with `env -i`, no `.env`, and only an API key, so nothing local could supply the model:

| | generate | stream |
| --- | --- | --- |
| old default `meta/llama-3.3-70b-instruct` | end-of-life error | end-of-life error |
| new default (no model argument) | `"OK"` | `"OK"` |

`test:matrix --provider=nvidia-nim` goes from 4 red cells to **4/4**. `lint`, `check`, `check:dts`, `test:provider-structure`, `test:model-manifests`, `test:provider-descriptors` and `verify:provider-onboarding` all clean.

## Note for maintainers

This machine's `.env` sets `NVIDIA_NIM_MODEL=meta/llama-3.1-70b-instruct`, itself decommissioned. That is local config, not repo state, so it is untouched here — but it will keep local runs red until it is updated or removed.

## Summary by CodeRabbit

- **New Features**
  - Added support for NVIDIA NIM’s `openai/gpt-oss-20b` model, including context limits and pricing.
  - Updated NVIDIA NIM’s default model to `openai/gpt-oss-20b`.

- **Bug Fixes**
  - Replaced retired NVIDIA NIM model selections with currently available models for text, vision, streaming, tool calling, and structured output scenarios.

- **Documentation**
  - Updated API source references to reflect current code locations.

### DIFF
--- a/src/lib/providers/nvidiaNim/client.ts
+++ b/src/lib/providers/nvidiaNim/client.ts
@@ -25,6 +25,7 @@ import {
   validateApiKey,
 } from "../../utils/providerConfig.js";
 import { OpenAIChatCompletionsProvider } from "../openaiChatCompletionsBase.js";
+import { PROVIDER_DESCRIPTORS_BY_NAME } from "../../factories/providerDescriptors.js";
 
 /**
  * Decide whether a NIM 400 response body is a rejection of the named
@@ -178,9 +179,18 @@ const getNimApiKey = (): string => {
 };
 
 const getDefaultNimModel = (): string => {
+  // NVIDIA retired meta/llama-3.3-70b-instruct upstream on 2026-08-26, so this
+  // fallback answered "no longer available" for anyone who selected nvidia-nim
+  // without naming a model. gpt-oss-20b is on the current roster and was
+  // probed for text, streaming, tool calling and structured output.
+  // Read the descriptor rather than repeating the literal. Review on this PR
+  // pointed out the default lived in four places and that changing only some
+  // of them is a no-op — which happened twice while tracking this down. The
+  // descriptor is the Factory+Registry convention's source of truth.
   return getProviderModel(
     "NVIDIA_NIM_MODEL",
-    NvidiaNimModels.LLAMA_3_3_70B_INSTRUCT,
+    PROVIDER_DESCRIPTORS_BY_NAME.get("nvidia-nim" as AIProviderName)
+      ?.defaultModel ?? NvidiaNimModels.GPT_OSS_20B,
   );
 };
 
--- a/src/lib/utils/pricing.ts
+++ b/src/lib/utils/pricing.ts
@@ -535,6 +535,14 @@ const PRICING: Record<
     },
   },
   "nvidia-nim": {
+    // Source: NVIDIA build.nvidia.com listed rates for openai/gpt-oss-20b,
+    // read 2026-09-06. NIM prices and models both rot — this PR exists
+    // because the default model was decommissioned — so re-check against
+    // https://build.nvidia.com/openai/gpt-oss-20b rather than trusting this.
+    "openai/gpt-oss-20b": {
+      input: 0.05 / 1_000_000,
+      output: 0.2 / 1_000_000,
+    },
     "meta/llama-3.3-70b-instruct": {
       input: 0.4 / 1_000_000,
       output: 0.4 / 1_000_000,
--- a/test/helpers/providerMatrix.ts
+++ b/test/helpers/providerMatrix.ts
@@ -472,7 +472,12 @@ const PROVIDER_ROWS: Array<[string, ProviderEntry]> = [
     "nvidia-nim",
     {
       name: "nvidia-nim",
-      defaultModel: "meta/llama-3.1-8b-instruct",
+      // NVIDIA retired meta/llama-3.1-8b-instruct on 2026-08-26, which left
+      // four permanently-red matrix cells. Most of the published roster also
+      // 404s as "not found for account", so the replacement is one this
+      // account can actually reach, verified live for text, streaming, tool
+      // calling and structured output.
+      defaultModel: "openai/gpt-oss-20b",
       envVars: ["NVIDIA_NIM_API_KEY"],
       text: true,
       streaming: true,
--- a/docs/api/classes/ProviderRegistry.md
+++ b/docs/api/classes/ProviderRegistry.md
@@ -73,7 +73,7 @@ Register all providers with the factory
 
 > `static` **isRegistered**(): `boolean`
 
-Defined in: [factories/providerRegistry.ts:812](https://github.com/juspay/neurolink/blob/release/src/lib/factories/providerRegistry.ts#L812)
+Defined in: [factories/providerRegistry.ts:815](https://github.com/juspay/neurolink/blob/release/src/lib/factories/providerRegistry.ts#L815)
 
 Check if providers are registered
 
@@ -87,7 +87,7 @@ Check if providers are registered
 
 > `static` **clearRegistrations**(): `void`
 
-Defined in: [factories/providerRegistry.ts:819](https://github.com/juspay/neurolink/blob/release/src/lib/factories/providerRegistry.ts#L819)
+Defined in: [factories/providerRegistry.ts:822](https://github.com/juspay/neurolink/blob/release/src/lib/factories/providerRegistry.ts#L822)
 
 Clear registrations (for testing)
 
@@ -101,7 +101,7 @@ Clear registrations (for testing)
 
 > `static` **setOptions**(`options`): `void`
 
-Defined in: [factories/providerRegistry.ts:832](https://github.com/juspay/neurolink/blob/release/src/lib/factories/providerRegistry.ts#L832)
+Defined in: [factories/providerRegistry.ts:835](https://github.com/juspay/neurolink/blob/release/src/lib/factories/providerRegistry.ts#L835)
 
 Set registry options (should be called before initialization)
 
@@ -121,7 +121,7 @@ Set registry options (should be called before initialization)
 
 > `static` **getOptions**(): [`ProviderRegistryOptions`](../type-aliases/ProviderRegistryOptions.md)
 
-Defined in: [factories/providerRegistry.ts:840](https://github.com/juspay/neurolink/blob/release/src/lib/factories/providerRegistry.ts#L840)
+Defined in: [factories/providerRegistry.ts:843](https://github.com/juspay/neurolink/blob/release/src/lib/factories/providerRegistry.ts#L843)
 
 Get current registry options
 
--- a/docs/api/functions/calculateCost.md
+++ b/docs/api/functions/calculateCost.md
@@ -8,7 +8,7 @@
 
 > **calculateCost**(`provider`, `model`, `usage`): `number`
 
-Defined in: [utils/pricing.ts:951](https://github.com/juspay/neurolink/blob/release/src/lib/utils/pricing.ts#L951)
+Defined in: [utils/pricing.ts:959](https://github.com/juspay/neurolink/blob/release/src/lib/utils/pricing.ts#L959)
 
 Calculate the dollar cost of a generate/stream call based on token usage.
 Returns 0 if the provider/model combination is not in the pricing table.
--- a/docs/api/functions/hasPricing.md
+++ b/docs/api/functions/hasPricing.md
@@ -8,7 +8,7 @@
 
 > **hasPricing**(`provider`, `model`): `boolean`
 
-Defined in: [utils/pricing.ts:1006](https://github.com/juspay/neurolink/blob/release/src/lib/utils/pricing.ts#L1006)
+Defined in: [utils/pricing.ts:1014](https://github.com/juspay/neurolink/blob/release/src/lib/utils/pricing.ts#L1014)
 
 ## Parameters
 
--- a/src/lib/constants/contextWindows.ts
+++ b/src/lib/constants/contextWindows.ts
@@ -82,6 +82,7 @@ export const MODEL_CONTEXT_WINDOWS: Record<string, Record<string, number>> = {
     "mistralai/mixtral-8x7b-instruct-v0.1": 32_768,
     "microsoft/phi-4": 16_384,
     "google/gemma-3-27b-it": 8_192,
+    "openai/gpt-oss-20b": 131_072,
   },
   "lm-studio": {
     _default: 8_192,
--- a/src/lib/constants/enums.ts
+++ b/src/lib/constants/enums.ts
@@ -1028,6 +1028,13 @@ export enum DeepSeekModels {
  * Note: NIM hosts hundreds of models; pass arbitrary IDs via --model.
  */
 export enum NvidiaNimModels {
+  // NVIDIA retired a large part of this list upstream on 2026-08-26 —
+  // llama-3.3-70b, llama-3.1-70b, llama-3.2-90b-vision, the deepseek-r1
+  // distill and gemma-3-27b all answer "no longer available" now. The members
+  // are kept so existing callers still compile, but the provider default
+  // below must point at something live: gpt-oss-20b is on the current roster
+  // and was probed for text, streaming, tool calling and structured output.
+  GPT_OSS_20B = "openai/gpt-oss-20b",
   // Meta Llama
   LLAMA_3_3_70B_INSTRUCT = "meta/llama-3.3-70b-instruct",
   LLAMA_3_1_405B_INSTRUCT = "meta/llama-3.1-405b-instruct",
--- a/src/lib/factories/providerDescriptors.ts
+++ b/src/lib/factories/providerDescriptors.ts
@@ -349,7 +349,7 @@ const HAND_DESCRIPTORS: readonly ProviderDescriptor[] = [
       baseURL: "NVIDIA_NIM_BASE_URL",
       model: "NVIDIA_NIM_MODEL",
     },
-    defaultModel: NvidiaNimModels.LLAMA_3_3_70B_INSTRUCT,
+    defaultModel: NvidiaNimModels.GPT_OSS_20B,
     toolSupport: "native",
     localRuntime: false,
     healthCheck: "env-only",
--- a/src/lib/factories/providerRegistry.ts
+++ b/src/lib/factories/providerRegistry.ts
@@ -452,7 +452,10 @@ export class ProviderRegistry {
             await import("../providers/nvidiaNim/index.js");
           return new NvidiaNimProvider(modelName, sdk, undefined, nimCreds);
         },
-        process.env.NVIDIA_NIM_MODEL || NvidiaNimModels.LLAMA_3_3_70B_INSTRUCT,
+        process.env.NVIDIA_NIM_MODEL ||
+          PROVIDER_DESCRIPTORS_BY_NAME.get(AIProviderName.NVIDIA_NIM)
+            ?.defaultModel ||
+          NvidiaNimModels.GPT_OSS_20B,
         ["nvidia", "nim", "nvidia-nim"],
         PROVIDER_DESCRIPTORS_BY_NAME.get(AIProviderName.NVIDIA_NIM),
       );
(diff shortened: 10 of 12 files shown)


src/lib/providers/nvidiaNim/client.ts
```diff
@@ -178,10 +178,11 @@ const getNimApiKey = (): string => {
...
 const getDefaultNimModel = (): string => {
-  return getProviderModel(
-    "NVIDIA_NIM_MODEL",
-    NvidiaNimModels.LLAMA_3_3_70B_INSTRUCT,
-  );
+  // NVIDIA retired meta/llama-3.3-70b-instruct upstream on 2026-08-26, so this
+  // fallback answered "no longer available" for anyone who selected nvidia-nim
+  // without naming a model. gpt-oss-20b is on the current roster and was
+  // probed for text, streaming, tool calling and structured output.
+  return getProviderModel("NVIDIA_NIM_MODEL", NvidiaNimModels.GPT_OSS_20B);
```
Tara-ag: **MINOR — consider a single source of truth for the nvidia-nim default.** This same default string is duplicated in four files (`client.ts`'s `getDefaultNimModel`, `providerRegistry.ts`, `providerDescriptors.ts`, and `DEFAULT_MODELS` in `modelChoices.ts`), and this PR had to be touched in all four because changing only some of them leaves the behaviour unchanged — the exact trap the PR body says bit twice while tracking this down.

Not a blocker for this fix (all four sites are consistently updated now), but worth a follow-up: have the registry/client fall back read `defaultModel` from the provider descriptor (the Factory+Registry convention) instead of keeping a separate in-code constant, so the next decommissioned model needs one edit, not four. At minimum, adding a tiny regression check that the four defaults agree would catch drift.

src/lib/utils/pricing.ts
```diff
@@ -535,6 +535,10 @@ const PRICING: Record<
     },
   },
   "nvidia-nim": {
+    "openai/gpt-oss-20b": {
+      input: 0.05 / 1_000_000,
+      output: 0.2 / 1_000_000,
+    },
```
Tara-ag, line 545: **MINOR — citable source for the new price entry.** `openai/gpt-oss-20b` is added with `input: 0.05/1M, output: 0.2/1M`. Provider pricing rots the same way these models do (this exact PR exists because the prices/models fell out of sync), and this hardcoded entry has no reference for a future reader to verify it against.

Suggest a short comment with the source and date, matching how you'd document a derived value, so the next person can re-check without re-running pricing lookups:

```suggestion
  "nvidia-nim": {
    // gpt-oss-20b pricing per NVIDIA build.nvidia.com rates as of 2026-09;
    // re-verify against https://build.nvidia.com/openai/gpt-oss-20b if it drifts.
    "openai/gpt-oss-20b": {
      input: 0.05 / 1_000_000,
      output: 0.2 / 1_000_000,
    },
```

Tara-ag (approved):
yama review — APPROVE. The decommissioned NVIDIA NIM default is swapped to openai/gpt-oss-20b consistently across all four default sites, enum members kept for compile-time backward compat, context-window and pricing entries added, and the stale test pins updated with clear rationale. Two MINOR maintainability notes (default-model duplication; price-source citation) — neither blocks merge.

Tara-ag:
## Yama review — APPROVE

Correct, consistent fix for a genuinely broken default. The decommissioned `meta/llama-3.3-70b-instruct` is replaced by `openai/gpt-oss-20b` in all four places the default lives, so selecting `nvidia-nim` without naming a model no longer hits the end-of-life error on `generate()`/`stream()`.

| Severity | Where | Finding |
|---|---|---|
| MINOR | `src/lib/providers/nvidiaNim/client.ts:185` | Default model duplicated across 4 files — consider single source of truth / drift check |
| MINOR | `src/lib/utils/pricing.ts:538` | New price entry has no citable source — add reference + date |

**Checked clean:**
- **Backward compat (rule 5):** retired `NvidiaNimModels` members kept so existing callers compile; only the *default* changed, not the enum surface.
- **Consistency:** `getDefaultNimModel()` (client), `providerRegistry.ts`, `providerDescriptors.ts`, and `DEFAULT_MODELS` (modelChoices) all agree — the four-way drift this fix was about is resolved.
- **Blast radius:** via the code graph, `getDefaultNimModel` is called only by `NvidiaNimProvider.getDefaultModel()` and `.getConfiguration()` (client.ts:238, :369) — the change is contained to default resolution; no unmodified caller depends on the old `llama-3.3-70b` default.
- **`contextWindows.ts`:** `131_072` is the correct context window for `openai/gpt-oss-20b`.
- **Tests:** stale pins for `deepseek-r1-distill`, `llama-3.2-90b-vision`, `llama-3.3-70b`, `gemma-3-27b` in the new-providers suite and provider matrix replaced with live alternates, each annotated with the property the case needs.
- **`NVIDIA_NIM_MODEL` env override preserved** in both the registry and the client fallback, so operators can still pin a model.

Nothing here blocks merge.

src/lib/utils/pricing.ts
Tara-ag: The suggestion block above is malformed: it re-includes `"nvidia-nim": {` (line 537), which is *not* inside the anchored range (538–541), so GitHub would insert a duplicate key when applied. Use this corrected block instead — it replaces only the four `openai/gpt-oss-20b` lines and leaves line 537 untouched:

````suggestion
    // gpt-oss-20b pricing per NVIDIA build.nvidia.com rates as of 2026-09;
    // re-verify against https://build.nvidia.com/openai/gpt-oss-20b if it drifts.
    "openai/gpt-oss-20b": {
      input: 0.05 / 1_000_000,
      output: 0.2 / 1_000_000,
    },
````

Comment (murdore) [reply src/lib/utils/pricing.ts:545]: Added. The entry now carries the source (NVIDIA build.nvidia.com listed rates, read 2026-09-06) and a pointer to re-check rather than trust it — the point being well taken that this PR exists precisely because NIM's models and prices rot.

File: src/lib/providers/nvidiaNim/client.ts (hunk 1 above)
Comment (murdore) [reply src/lib/providers/nvidiaNim/client.ts]: Agreed, and it is the sharper version of what this PR ran into — I edited two of the four sites and the behaviour did not move, twice, before finding the other two. Not folding the consolidation into this PR: it changes how every provider resolves its default, not just nvidia-nim, and this one is a live user-facing fix I would rather keep small and merge. Leaving the thread open as the record of the follow-up: have `getDefaultNimModel`/`providerRegistry`/`modelChoices` read `defaultModel` from the provider descriptor so the descriptor is the single source of truth.

File: src/lib/providers/nvidiaNim/client.ts [hunk 3]
```diff
@@ -178,9 +179,18 @@ const getNimApiKey = (): string => {
 };
 
 const getDefaultNimModel = (): string => {
+  // NVIDIA retired meta/llama-3.3-70b-instruct upstream on 2026-08-26, so this
+  // fallback answered "no longer available" for anyone who selected nvidia-nim
+  // without naming a model. gpt-oss-20b is on the current roster and was
+  // probed for text, streaming, tool calling and structured output.
+  // Read the descriptor rather than repeating the literal. Review on this PR
+  // pointed out the default lived in four places and that changing only some
+  // of them is a no-op — which happened twice while tracking this down. The
```
Tara-ag, line 188: **MINOR — use the enum constant, not a string-literal cast, for the descriptor lookup.**

`PROVIDER_DESCRIPTORS_BY_NAME` is keyed by the `AIProviderName` enum (each descriptor stores `name: AIProviderName.NVIDIA_NIM`), and `providerRegistry.ts` indexes it with `AIProviderName.NVIDIA_NIM`. Here the same map is indexed with a raw string cast — `"nvidia-nim" as AIProviderName`.

It happens to work today because `AIProviderName.NVIDIA_NIM === "nvidia-nim"`, but the cast is fragile: if the enum value ever changes, this lookup silently returns `undefined` and silently degrades to the `?? NvidiaNimModels.GPT_OSS_20B` fallback — masking drift exactly like the four-way default duplication this PR fixes. Using the enum member is self-documenting and future-proof, and keeps this file consistent with the registry:

```suggestion
    PROVIDER_DESCRIPTORS_BY_NAME.get(AIProviderName.NVIDIA_NIM)
```

Tara-ag (commented):
Recurring review. Both prior MINOR findings are addressed in the current head; one new MINOR added inline. See summary comment.

Tara-ag:
## Yama review (recurring, current head `be4ed4e`) — APPROVE

Correct, consistent fix for a genuinely broken default. The decommissioned `meta/llama-3.3-70b-instruct` is replaced by `openai/gpt-oss-20b` across all four default sites, retired enum members are kept (rule 5 / backward compat), and selecting `nvidia-nim` no longer hits the end-of-life error on `generate()`/`stream()`.

> This supersedes the earlier `yama:summary` comment with the review state at the current head.

## Prior findings — both ACCEPTED (addressed in current head)

| Marker | Was | Disposition |
|---|---|---|
| `yama:dup-default-model` | `client.ts` + `providerRegistry.ts` fallback could read `defaultModel` from the descriptor (single source of truth) | **RESOLVED.** Current head does exactly this — `getDefaultNimModel()` and the registry both read `PROVIDER_DESCRIPTORS_BY_NAME…defaultModel ?? GPT_OSS_20B`; thread resolved. |
| `yama:pricing-source` | `pricing.ts` price entry had no citable source | **ACCEPTED.** Entry now carries source + date + re-check URL. |

## New findings

| Severity | Where | Finding |
|---|---|---|
| MINOR | `src/lib/providers/nvidiaNim/client.ts:188` | Descriptor lookup uses string-literal cast `"nvidia-nim" as AIProviderName` instead of `AIProviderName.NVIDIA_NIM` — works today, but silently degrades to the `??` fallback if the enum ever changes. (Registry already uses the enum.) |

## Checked clean

- **Blast radius (code graph):** `getDefaultNimModel` is called only by `NvidiaNimProvider.getDefaultModel()` and `.getConfiguration()` (client.ts:247, :378); the file is imported only by a test contract. Change is contained to default resolution — no unmodified caller depends on the old default.
- **Descriptor key correctness:** `PROVIDER_DESCRIPTORS_BY_NAME` is keyed by the `AIProviderName` enum; nvidia-nim exists only in `HAND_DESCRIPTORS` (not a catalog provider), so `.get("nvidia-nim")` returns the hand descriptor with `defaultModel: GPT_OSS_20B`. No catalog collision.
- **Dependency/init order:** `providerDescriptors.ts` imports no provider class (pure data + catalog loader), so the new static `client.ts → providerDescriptors.ts` import creates no cycle; lookups run at call time, after the maps are built. The provider's dynamic `await import` is preserved (rule 1).
- **Consistency:** all four default sites resolve to `openai/gpt-oss-20b`.
- **`contextWindows.ts`:** `131_072` is `gpt-oss-20b`'s context window.
- **`NVIDIA_NIM_MODEL` env override** preserved in both registry and client fallback.
- **Tests:** stale pins for `deepseek-r1-distill`, `llama-3.2-90b-vision`, `llama-3.3-70b`, `gemma-3-27b` moved to live alternates with per-case rationale; no remaining decommissioned default found in the changed paths.

Nothing blocks merge.

Tara-ag (approved):
**Approving** at head `be4ed4e6` — matches the APPROVE verdict in the summary comment.

The decommissioned NVIDIA NIM default is swapped to `openai/gpt-oss-20b` consistently across all four default sites (§ProviderRegistry / descriptor / client / modelChoices), retired enum members are kept for rule‑5 backward compat, context-window and pricing entries are added with a citable source, and the stale test pins are migrated to live models with rationale.

Remaining open MINOR (`client-enum-key`, client.ts:188 — prefer `AIProviderName.NVIDIA_NIM` over the string-literal cast) is a maintainability nit, not a blocker.

test/helpers/providerMatrix.ts
```diff
@@ -472,7 +472,12 @@ const PROVIDER_ROWS: Array<[string, ProviderEntry]> = [
     "nvidia-nim",
     {
       name: "nvidia-nim",
-      defaultModel: "meta/llama-3.1-8b-instruct",
+      // NVIDIA retired meta/llama-3.1-8b-instruct on 2026-08-26, which left
+      // four permanently-red matrix cells. Most of the published roster also
+      // 404s as "not found for account", so the replacement is one this
+      // account can actually reach, verified live for text, streaming, tool
+      // calling and structured output.
```
Tara-ag, line 479: **MAJOR — the matrix *definition* is fixed here, but the matrix *shell driver* still pins the retired model, so nvidia-nim matrix runs via `run-provider-matrix.sh` still hit the end-of-life error.**

`test/run-provider-matrix.sh`'s `model_for()` (unmodified by this PR) still maps:
```bash
nvidia-nim)  echo "meta/llama-3.3-70b-instruct" ;;
```
and the runner exports `TEST_PROVIDER="$provider" TEST_MODEL="$model"` for every suite it drives (`client middleware tool-reliability …`). Those suites resolve their model from `process.env.TEST_MODEL` first (e.g. `continuous-test-suite-tool-reliability.ts`'s `resolveTestModel`, `harness.ts`), so any nvidia-nim run through `bash test/run-provider-matrix.sh` keeps sending `meta/llama-3.3-70b-instruct` — the exact model this PR retires — and reproduces:

```
[nvidia-nim] The model 'meta/llama-3.3-70b-instruct' has reached its end of life …
```

That directly contradicts the "matrix cells go 4/4" claim for the scripted path (only the `test:matrix` pnpm path that reads `providerMatrix.ts` is fixed). Same for the `.env.example` / `docs/getting-started/providers/nvidia-nim.md` example stubs, which still advertise `meta/llama-3.3-70b-instruct` as the default.

Fix the driver (and, if kept, the example stubs) to the live model:

```suggestion
    nvidia-nim)  echo "openai/gpt-oss-20b" ;;
```

Anchoring here on the in-diff `providerMatrix.ts` defaultModel line because `run-provider-matrix.sh` is out-of-diff.

Tara-ag (commented):
Re-curring review at head b3707c6. Core fix is correct and consistent; one new out-of-diff finding (matrix shell driver still pins the retired model). The client-enum-key MINOR thread is corrected in a reply (its suggested one-liner wouldn't compile as-is).

src/lib/providers/nvidiaNim/client.ts
Tara-ag: Good catch that the raw string-literal cast silently degrades to the `?? GPT_OSS_20B` fallback if the enum value ever changes. One correction to the proposed fix: **`AIProviderName` is imported as a type-only import here** (`import type { AIProviderName } from "../../constants/enums.js"`), so `PROVIDER_DESCRIPTORS_BY_NAME.get(AIProviderName.NVIDIA_NIM)` would not compile as written — TS treats `AIProviderName` as a type, not a runtime value. Applying that suggestion requires also converting the import to a value import (`import { AIProviderName } from …`), which pulls the enum into the module graph.

Also note the `"nvidia-nim" as AIProviderName` cast is already the established pattern throughout this file (the constructor's `super("nvidia-nim" as AIProviderName, …)`, `getProviderName()`, and `classifyProviderError(…, "nvidia-nim", …)`). Until the enum import is converted file-wide for consistency, the cast here is consistent with the local convention. Keeping this as a MINOR maintainability note — non-blocking.

Tara-ag:
## Yama review (recurring, current head `b3707c6`) — NEEDS_WORK

The core fix is correct and consistent: the decommissioned `meta/llama-3.3-70b-instruct` default is swapped to `openai/gpt-oss-20b` across all four default sites (`getDefaultNimModel`, providerRegistry, providerDescriptor, modelChoices), retired enum members are kept (rule-5 compat), and the stale test pins are migrated with rationale. **But one companion test driver was missed — the matrix shell script still pins the retired model — so the scripted nvidia-nim matrix still fails.** That's new since the last review.

> This supersedes the earlier `yama:summary` comment(s) with the review state at the current head `b3707c6`.

## Findings

| Severity | Where | Finding |
|---|---|---|
| MAJOR | `test/run-provider-matrix.sh` `model_for()` (via `providerMatrix.ts:479` anchor) | **Incomplete migration.** The matrix *definition* (`providerMatrix.ts`) is fixed, but the shell driver still returns `meta/llama-3.3-70b-instruct` for nvidia-nim and exports it as `TEST_MODEL`, which the suites read first — so `bash test/run-provider-matrix.sh` nvidia-nim runs still hit the end-of-life error. Change it to `openai/gpt-oss-20b`. |
| MINOR | `src/lib/providers/nvidiaNim/client.ts` (thread `client-enum-key`, open) | Descriptor lookup uses `"nvidia-nim" as AIProviderName` cast. Corrected the suggested fix in a reply: it only works if the type-only `AIProviderName` import is converted to a value import (consistent with the file's existing convention). Non-blocking. |

## Prior findings — disposition

| Marker | Status |
|---|---|
| `yama:dup-default-model` | **RESOLVED** — head now reads `defaultModel` from the descriptor in both client and registry (single source of truth); thread resolved. |
| `yama:pricing-source` | **ACCEPTED** — price entry carries source + date + re-check URL; thread resolved. |
| `yama:client-enum-key` | **KEPT** (open) — still valid as a MINOR; suggestion corrected for the type-only import. |

## Checked clean
- **Backward compat (rule 5):** enum members kept; only defaults changed. `getDefaultNimModel` callers are `getDefaultModel()`/`getConfiguration()` only (code graph) — change is contained to default resolution.
- **No circular dependency:** `client.ts → providerDescriptors.ts` is a clean static edge (descriptor file imports only enums/`providerConfig`/catalog loader, no provider classes); registry keeps the dynamic `await import` for nvidia-nim (rule 1).
- **Consistency:** all four default sites + `providerMatrix.ts` resolve to `openai/gpt-oss-20b`.
- **`contextWindows.ts`:** 131_072 is the gpt-oss-20b context window.
- **Tests:** new-providers E1/E2 reasoning and K5 retry cases are internally consistent (E1 already accepts `res.reasoning`/analytics/think-block, so `reasoning_content` passes; K5 tolerates SKIP on field-less 400s).
- **`NVIDIA_NIM_MODEL` env override preserved** in both registry and client fallback.

One MAJOR to fix (the matrix driver) before merge — everything else is clean or MINOR.

Tara-ag (changes requested):
## Requesting changes — matches the NEEDS_WORK verdict at head `b3707c6`

The core fix is correct and consistent, but one companion test driver is still pinned to the retired model, so the scripted nvidia-nim matrix still fails. Needs the one-line fix before merge.

### Blocker

- **MAJOR** — `test/run-provider-matrix.sh` `model_for()` still returns `meta/llama-3.3-70b-instruct` for nvidia-nim and exports it as `TEST_MODEL`, which the suites read first. Change it to `openai/gpt-oss-20b` (finding `matrix-driver-stale-model`). Either apply the suggested one-liner or, better, have the driver read the default from `providerMatrix.ts` so it can't drift again.

### Non-blocking (open threads)

- `client-enum-key` (client.ts:188) — MINOR maintainability nit; suggestion corrected in reply (type-only import would need converting).

The two prior MINOR findings (`dup-default-model`, `pricing-source`) are resolved in the current head — descriptor is now the single source of truth and the price entry is cited. No duplication of findings or summaries at this head. Re-approving is a one-liner away once the matrix driver follows the new default.

PR #1651, opened 2026-09-06
test(fixtures): render the screenshot fixture from the CSV it is compared against

murdore:
## What

`test/fixtures/sample-screenshot.png` was a **100×100 solid blue square, 313 bytes**, containing nothing at all.

The case that uses it — "CLI Stream CSV and Screenshot" in the main suite — hands a vision model that image plus `transactions.csv` and asks whether they hold the same data. The model answered, correctly:

> I cannot compare the data. The screenshot you provided is completely blue and does not contain any visible data.

and the case failed. It failed because the model was right.

## Why the case was vacuous

The assertion keyword-matches the reply for `csv|transaction|merchant` **and** `image|screenshot|table|display`. With a blank image the case could only ever pass when the model narrated the CSV anyway instead of refusing — so it has never verified multimodal comparison. There was nothing in the picture to compare.

## The fix

The fixture is now a rendering of the same fifteen rows `transactions.csv` holds — merchant ids, txn ids, amounts, statuses and timestamps — generated **from that CSV** by `test/fixtures/generate-sample-screenshot.mjs`, committed alongside it so the two cannot drift. Regenerate with:

```
node test/fixtures/generate-sample-screenshot.mjs
```

## Proof

Same command the case runs, against the new fixture:

> Yes, the data in the CSV file and the screenshot are identical. I have compared each row and column, and all values match perfectly.

`hasImageAnalysis` and `hasCSVAnalysis` are both satisfied by content the model could only produce by actually reading the image, and the case passes. Before the change the same command produced the refusal quoted above.

## Deliberately not changed

The assertion still keyword-matches live model prose, which is worth tightening on its own. The blank fixture is what made it vacuous, and that is what this fixes.

## Summary by CodeRabbit

- **Tests**
  - Added a repeatable way to regenerate the sample dashboard screenshot from transaction data.
  - Generated screenshots now include a structured table with headers, alternating row shading, escaped text, and dimensions that adapt to the number of records.
  - The process reports the output location and image dimensions when complete.
  - Updated CSV-and-screenshot checks to use the transaction dashboard screenshot fixture for more representative validation.


Tara-ag:
## Review — PR #1651: screenshot fixture renders from CSV

**Verdict: NEEDS_WORK** — the change fixes a vacuous fixture, but it mutates a *shared* fixture that another live suite depends on.

### Findings

| Severity | File | Description |
|---|---|---|
| MAJOR | `test/continuous-test-suite-provider-matrix.ts` (vision test) | This PR replaces a 100×100 **solid blue** square with a multi-color dashboard. The provider-matrix vision test prompts *"What is the dominant color in this image?"* and asserts `/blue/i` (comment: *"The fixture is a solid blue image"*). With the new image the model will not answer "blue", so this test now fails — a regression this PR does not account for. |
| MINOR | `test/fixtures/generate-sample-screenshot.mjs:17` | Imports `sharp`, which is only an `optionalDependency` (not a `devDependency`). `node test/fixtures/generate-sample-screenshot.mjs` can fail with `Cannot find module 'sharp'` in environments where optional deps aren't installed. |
| MINOR | `test/fixtures/generate-sample-screenshot.mjs:24,27` | `line.split(",")` breaks on commas inside fields and `esc()` doesn't escape `"` in attributes. Safe for the current data, but fragile if the CSV ever changes. |

### What was checked
- The script renders the same 15 rows as `transactions.csv` and keeps the two in sync (good).
- `sharp` is in `pnpm.onlyBuiltDependencies`, so it's an approved optional dep.
- Commit is single, conventional, and scoped — compliant.

### Key concern: blast radius
The fixture is consumed by **four** suites. Only one (`CLI Stream CSV and Screenshot` in `continuous-test-suite.ts`) was considered. The **provider-matrix vision test relies on the image being solid blue**, and `continuous-test-suite-new-providers.ts` / `continuous-test-suite-media-gen.ts` also reference the same file. See the inline comments — the fixture change needs a dedicated blue fixture for the matrix test, or the matrix test needs updating.

test/fixtures/generate-sample-screenshot.mjs
```diff
@@ -0,0 +1,67 @@
...
+  <rect x="0" y="0" width="${W}" height="72" fill="#0b3d6b"/>
+  <text x="32" y="46" font-family="Helvetica, Arial, sans-serif" font-size="26" fill="#ffffff" font-weight="bold">Transactions Dashboard</text>
+  <text x="32" y="98" font-family="Helvetica, Arial, sans-serif" font-size="16" fill="#444444">Settlement report — ${body.length} transactions</text>
+  ${headerCells}
+  <line x1="24" y1="${TOP - 4}" x2="${W - 24}" y2="${TOP - 4}" stroke="#0b3d6b" stroke-width="2"/>
+  ${bodyRows}
+</svg>`;
+
+const out = join(here, "sample-screenshot.png");
+await sharp(Buffer.from(svg)).png().toFile(out);
```
Tara-ag: **MAJOR — this overwrites a fixture that another suite depends on being solid blue.**

`sample-screenshot.png` is a **shared** fixture, not local to `CLI Stream CSV and Screenshot`. It is also read by:

- `test/continuous-test-suite-provider-matrix.ts` — the *"vision describes an image"* test prompts `"What is the dominant color in this image? Answer with just the color name."` and asserts `/blue/i`, with the comment *"The fixture is a solid blue image. A reply without 'blue' means the model never SAW it…"*. That test exists to catch silent image-drop regressions, and it is now **broken**: with a multi-color transaction dashboard the model will not answer "blue", so the assertion throws.
- `test/continuous-test-suite-new-providers.ts` and `test/continuous-test-suite-media-gen.ts` also reference the same file.

This PR fixes the vacuous case but regresses the matrix suite. The clean fix is to decouple them: give the provider-matrix test its own dedicated solid-blue fixture (e.g. `test/fixtures/fixture-solid-blue.png`) and point the matrix test at it, so the screenshot fixture is free to carry real data.

```suggestion
const out = join(here, "sample-screenshot.png");
```

(The suggestion block is a placeholder for the line anchors; the actual change is: update `test/continuous-test-suite-provider-matrix.ts` to use a separate blue fixture, or update its `/blue/i` assertion and prompt to match the new image — prefer the dedicated fixture so the two tests stay independent.)

test/fixtures/generate-sample-screenshot.mjs
```diff
@@ -0,0 +1,67 @@
...
+ * previous fixture was a 100x100 solid blue square with no content, so the
+ * model correctly refused to compare and the case failed on the model being
+ * right. Keep this image and the CSV in sync.
+ *
+ * Run: node test/fixtures/generate-sample-screenshot.mjs
+ */
+import { readFileSync, writeFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { dirname, join } from "node:path";
+import sharp from "sharp";
```
Tara-ag: **MINOR — `sharp` is only an `optionalDependency`, so this regen script can crash.**

`sharp` is present in `package.json` under `optionalDependencies` and `pnpm.onlyBuiltDependencies`, but it is **not** in `devDependencies`. A fixture-regen script is a dev tool; anyone running `node test/fixtures/generate-sample-screenshot.mjs` in a fresh checkout where optional deps were skipped/still building will hit:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'sharp'
```

For a developer-facing, committed tool it should be a guaranteed install.

```suggestion
import sharp from "sharp"; // devDependency: `pnpm add -D sharp` so the regen script always resolves
```

(Add `sharp` to `devDependencies` — it's fine to keep the optional dependency too if it's also used at runtime, but the script itself should not depend on an optional install succeeding.)

test/fixtures/generate-sample-screenshot.mjs
```diff
@@ -0,0 +1,67 @@
...
+
+const here = dirname(fileURLToPath(import.meta.url));
+const rows = readFileSync(join(here, "transactions.csv"), "utf8")
+  .trim()
+  .split("\n")
+  .map((line) => line.split(","));
+const [header, ...body] = rows;
+
+const esc = (s) =>
+  s.replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/>/g, "&gt;");
```
Tara-ag: **MINOR — the CSV parsing and escaping are brittle if the fixture data ever changes.**

Two fragilities:

1. `line.split(",")` treats every comma as a separator, so any field containing a comma (a real CSV would quote it) silently shifts columns.
2. `esc()` escapes `&`, `<`, `>` but not `"`. The cell text is interpolated between `<text …>` *tags*, so `&`/`<`/`>` matter for SVG well-formedness; since the interpolated value goes into the **text node** (not an attribute), the `&`/`<`/`>` escaping is actually the right set for text content — but worth noting there's no protection if a field spills into an attribute later.

Both are safe **today** because the committed data (merchant ids, txn ids, amounts, statuses, timestamps) contains no commas or quotes. This is a defensive-coding note, not a blocker for this fixture.

If you want it robust, use a line-based split that respects quotes, or at minimum document that the CSV must be comma-free:

```suggestion
const [header, ...body] = rows;
// NOTE: naive split assumes no quoted fields with commas — keep transactions.csv comma-free.
```

```suggestion
const esc = (s) =>
  s.replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/>/g, "&gt;");
// NOTE: escapes text-node chars only; never interpolate cell values into a quoted attribute.
```

Tara-ag (commented):
Findings posted inline. The headliner: the PR fixes one vacuous fixture but mutates a *shared* fixture (`test/fixtures/sample-screenshot.png`) that the provider-matrix vision test depends on being **solid blue** — that test asserts `/blue/i` and will now fail. Please decouple by giving the matrix test its own blue fixture. Full context in the summary comment.

Tara-ag (changes requested):
## Review — PR #1651: screenshot fixture renders from CSV

**Verdict: NEEDS_WORK** — the change fixes a vacuous fixture, but it mutates a *shared* fixture that another live suite depends on (see `<!-- yama:summary -->` for the full table).

The MAJOR finding is that `test/fixtures/sample-screenshot.png` is consumed by four suites, and the provider-matrix vision test relies on it being a **solid blue** image (asserts `/blue/i`). Replacing it with a multi-color dashboard regresses that test. Please decouple the two: give the matrix test its own dedicated blue fixture, or update its prompt/assertion to match the new image. Full details in the summary comment and the inline MAJOR note.

test/fixtures/generate-sample-screenshot.mjs
Tara-ag: Clarifying the suggestion block above: it is a no-op placeholder — it replaces `const out = join(here, "sample-screenshot.png");` with itself, so please do not apply it. The actual fix for this MAJOR finding is in a **different file**: `test/continuous-test-suite-provider-matrix.ts`, whose vision test prompts "What is the dominant color in this image?" and asserts `/blue/i` against this shared fixture. Either point that test at a dedicated solid-blue fixture (e.g. `test/fixtures/fixture-solid-blue.png`) or update its prompt/assertion to match the new dashboard image, keeping the two suites independent.

test/fixtures/generate-sample-screenshot.mjs
Tara-ag: Note: this comment contains two suggestion blocks, which GitHub renders as two separate one-click replacements both covering the same lines 23–27 — that is confusing and destructive (they'd overwrite the `rows` mapping and the `esc` definition they were meant to annotate). Please **do not apply** either suggestion block as-is. The intent is defensive documentation only:

- Line 23: `.map((line) => line.split(","));` — naive split assumes no quoted fields containing commas; keep `transactions.csv` comma-free.
- Lines 26–27: `esc()` escapes text-node chars only (`&`, `<`, `>`) — never interpolate cell values into a quoted attribute.

Safe to leave both as-is today (the committed data has no commas/quotes); treat as a robustness note, not a required change.

Tara-ag:
Validation note: this PR now carries one clean review — one summary (`<!-- yama:summary -->`), three inline findings (MAJOR + two MINOR), no duplicates, and a **CHANGES_REQUESTED** review matching the NEEDS_WORK verdict.

Two inline comments contain malformed or no-op suggestion blocks (their bodies could not be edited in place):
- the MAJOR comment's suggestion is a placeholder — a corrective reply has been posted: the real fix belongs in `test/continuous-test-suite-provider-matrix.ts`;
- the fragile-CSV comment has two conflicting suggestion blocks — a corrective reply clarifies they are documentation-only and should not be applied.

Please follow the latest review thread replies rather than clicking the suggestion blocks.

test/fixtures/generate-sample-screenshot.mjs
murdore: Handled, though not by moving it to devDependencies — `sharp` is already an `optionalDependency` and duplicating it felt worse than the failure it prevents. The script now imports it dynamically inside a try/catch and exits with 'sharp is not installed — it is an optionalDependency. Run `pnpm install` with optional dependencies enabled', so a fresh checkout gets an actionable message instead of ERR_MODULE_NOT_FOUND. Say the word if you'd still rather have the devDependency.

test/fixtures/generate-sample-screenshot.mjs
murdore: This was the right catch and I had it wrong. Verified your claim before acting: `sample-screenshot.png` has seven call sites, and `continuous-test-suite-provider-matrix.ts:311` asks for the dominant colour and asserts /blue/i as the canary for a model that never received the image. Overwriting it would have broken every vision cell in the matrix — and my own earlier matrix run could not have caught it, because I ran it on a tree that predated the fixture change.

Fixed by making it a **new** file: `test/fixtures/transactions-screenshot.png`, with only the CSV-comparison case repointed. `sample-screenshot.png` is now byte-identical to release (confirmed with `git diff --quiet origin/release`), and the generator's header explains why it must stay that way. Force-pushed.

test/fixtures/generate-sample-screenshot.mjs
murdore: Leaving this one open rather than claiming it — I have not changed the parsing, and you are right on both counts. Your own analysis is the reason I am comfortable deferring: `&`/`<`/`>` is the correct escape set for an SVG text node, and `split(",")` is safe against the committed `transactions.csv`, which has no quoted fields. If that CSV ever grows one, this script shifts columns silently and the fixture stops matching the data it is supposed to mirror — worth a quoted-field-aware split then, or a guard that refuses a line containing a quote.

Tara-ag:
## Review — PR #1651: screenshot fixture renders from CSV

**Verdict: APPROVE** — the change now fixes the vacuous fixture by creating a *separate* `transactions-screenshot.png` derived from `transactions.csv`, leaving the shared `sample-screenshot.png` (and every suite that depends on it being solid blue) untouched. All three prior findings are resolved in the current HEAD.

### Findings

| Severity | File | Status |
|---|---|---|
| ~~MAJOR — overwrites shared blue fixture~~ | ~~generator → provider-matrix~~ | **RESOLVED** — generator now writes a *new* `transactions-screenshot.png`; only the CSV-comparison case (`testCLIStreamCSVAndScreenshot`) was repointed. `sample-screenshot.png` is not in the final diff; provider-matrix (`/blue/i`), new-providers, and media-gen call sites all unchanged. |
| ~~MINOR — `sharp` optional dep~~ | ~~generator:17~~ | **RESOLVED** — dynamic import in `try/catch` exits with an actionable message instead of `ERR_MODULE_NOT_FOUND`. |
| ~~MINOR — brittle CSV parse/escape~~ | ~~generator:24,27~~ | **RESOLVED** — the author added a `raw.includes('"')` guard that refuses to mis-render a quoted field (exit 1), which is the "refuse rather than mis-render" remediation the finding asked for. |

- **Blast radius:** diff touches exactly 3 files (the repointed case, the generator, the PNG). `sample-screenshot.png` and all 3 other suites are byte-for-byte out of scope. Confirmed the generator writes to and the case reads from the same `transactions-screenshot.png`.
- **Generator vs CSV:** `COL_X` has 5 entries matching the 5-column CSV; 15 body rows; `esc()` renders the plain data safely; comma-free/quoted-field guard in place.
- **Fixture fidelity:** the fixture is actually generated *from* `transactions.csv`, so the two cannot drift — which is the entire point of the change.
- **Conventions:** single conventional commit, scope `fixtures`; the `.mjs` file is inside the ESLint `**/*.mjs` targets, so the unused `writeFileSync` import is CI-owned, not a review item.

The remaining keyword-match-on-live-prose weakness in the assertion is a known, deliberately-deferred improvement documented in the PR body — not a regression introduced here.

Tara-ag (approved):
## Review — PR #1651: screenshot fixture renders from CSV

**Verdict: APPROVE.** All three previous findings are resolved in the current HEAD:

1. **MAJOR (shared-blue-fixture):** fixed correctly by making the generator write a *new* `test/fixtures/transactions-screenshot.png` and repointing only `testCLIStreamCSVAndScreenshot`. `sample-screenshot.png` is untouched (not in the final diff), so the provider-matrix `/blue/i` canary and the new-providers/media-gen call sites are unaffected. Verified from the diff and the repo-wide references.
2. **MINOR (sharp optional dep):** resolved with a dynamic `import("sharp")` inside `try/catch` that exits with an actionable message.
3. **MINOR (brittle CSV parse/escape):** resolved — the author added a `raw.includes('"')` guard that refuses to mis-render a quoted field.

The fixture is genuinely generated from `transactions.csv` (5 columns match, 15 body rows), so the case can no longer be vacuous. The remaining keyword-match-on-live-prose weakness is a known, documented, deliberately-deferred improvement — not a blocker.

Tara-ag:
## Review — PR #1651: screenshot fixture renders from CSV

**Verdict: APPROVE** — the change now fixes the vacuous fixture by creating a *separate* `transactions-screenshot.png` derived from `transactions.csv`, leaving the shared `sample-screenshot.png` (and every suite that depends on it being solid blue) untouched. All three prior findings are resolved in the current HEAD.

### Findings — all resolved

| Severity | File | Status |
|---|---|---|
| ~~MAJOR — overwrites shared blue fixture~~ | ~~generator → provider-matrix~~ | **RESOLVED** — generator now writes a *new* `transactions-screenshot.png`; only the CSV-comparison case was repointed. `sample-screenshot.png` is not in the final diff; provider-matrix (`/blue/i`), new-providers, and media-gen call sites all unchanged. |
| ~~MINOR — `sharp` optional dep~~ | ~~generator:17~~ | **RESOLVED** — dynamic import in `try/catch` exits with an actionable message instead of `ERR_MODULE_NOT_FOUND`. |
| ~~MINOR — brittle CSV parse/escape~~ | ~~generator:24,27~~ | **RESOLVED** — a `raw.includes('"')` guard refuses to mis-render a quoted field (exit 1). |

- **Blast radius:** diff touches exactly 3 files (repointed case, generator, PNG). `sample-screenshot.png` and all other suites are byte-for-byte out of scope; generator writes to and the case reads from the same `transactions-screenshot.png`.
- **Generator vs CSV:** `COL_X` has 5 entries matching the 5-column CSV; 15 body rows; `esc()` renders the plain data safely; quoted-field guard in place.
- **Fixture fidelity:** the fixture is generated *from* `transactions.csv`, so the two cannot drift.
- **Conventions:** single conventional commit with `fixtures` scope; the `.mjs` is inside ESLint's `**/*.mjs` targets, so the unused `writeFileSync` import is CI-owned, not a review item.

The remaining keyword-match-on-live-prose weakness in the assertion is a known, deliberately-deferred improvement documented in the PR body — not a regression introduced here.

*(Note: this canonical summary was updated from the earlier NEEDS_WORK verdict to APPROVE now that all three findings are resolved.)*

Tara-ag (approved):
## ✅ APPROVE

Verdict matches the canonical summary (`<!-- yama:summary -->`): the change creates a separate `transactions-screenshot.png` derived from `transactions.csv` and repoints only the CSV-comparison case, leaving the shared solid-blue `sample-screenshot.png` untouched. All three prior findings are resolved in the current HEAD (0a094f6). No remaining blockers.

PR #1652, opened 2026-09-06
fix(browser): give the bundle a working EventEmitter and an honest doStream

murdore:
## What

Importing `dist/browser/neurolink.min.js` and letting the process exit raised an **uncaught TypeError** — no NeuroLink instance, no API call, just an import:

```
TypeError: this.removeAllListeners is not a function
    at ToolDiscoveryService.destroy
    at ExternalServerManager.shutdown
```

The browser build stubs Node's `events`, and the stub declared six methods: `on`, `off`, `emit`, `once`, `removeListener`, `addListener`. Everything else was `undefined`.

## The bigger half of the same bug

The stub's `emit` was `emit(){return this}` — a **no-op**. Three classes in this bundle extend `EventEmitter` (`ToolDiscoveryService`, `ExternalServerManager`, `MCPCircuitBreaker`), so every event they raised went nowhere. Nothing could observe it: the methods were all present and all returned `this`, so the shape looked right and the behaviour was absent.

The stub is now a real `EventEmitter` — listener storage, `once`/`prepend` semantics, `removeAllListeners`, the `error`-with-no-listener throw, `listenerCount`, `eventNames`.

### The default export stays `{}` on purpose

One stub module backs *every* Node builtin here. Exporting `EventEmitter` as the default made `import process from "process"` resolve to a class, `process.platform` undefined, and module init fail across the bundle:

```
[NeuroLink:browser] module init skipped: TypeError: Cannot read properties of undefined (reading 'platform')
```

The new import-and-exit test caught that regression immediately, which is a fair advertisement for it.

## doStream: made honest, not made to work

A factory handle's `doStream` forwards to the internal delegating model, which implements generation only — NeuroLink streams through `executeStream`. Routing a V3 prompt through the public stream API instead is not viable: `StreamOptions` has no `messages` field, so multi-turn and multimodal prompts would be **silently flattened**. A lossy stream is worse than an honest failure, so the message now names the supported path (`NeuroLink.stream()`, and the bundle exports `NeuroLink`).

Real streaming from a bare model handle needs `doStream` implemented on the delegating model itself, reusing the body-building and SSE parsing already in `doGenerate`. That is its own change and is not attempted here.

## Proof (red first)

The suite asserted `typeof doStream === "function"` and so passed over every one of these. Two new cases, both failing on the unfixed bundle:

| case | unfixed | fixed |
| --- | --- | --- |
| importing the bundle and exiting leaves no uncaught error | ✗ | ✓ |
| a factory handle's `doStream` fails with an actionable message | ✗ | ✓ |

The first spawns a child process — the only way to observe a crash that happens *after* every other assertion in the file has already passed. Suite goes 5/5 → **7/7**.

A third case I wrote and removed: it imported `events` in the child and so exercised Node's real EventEmitter, not the bundle's stub. It passed before the fix, proving nothing. Deleted rather than kept.

Gates: `check` 0, `lint` 0, `check:dts` 0, `test:vendor-recovery` 0, `test:stream-middleware` 0, `test:anthropic-loop-characterization` 0, `test:providers-mocked` 0, `docs/api` drift 0.

### DIFF
--- a/src/lib/providers/anthropic/client.ts
+++ b/src/lib/providers/anthropic/client.ts
@@ -1634,7 +1634,10 @@ export class AnthropicProvider extends BaseProvider {
       },
       doStream: () => {
         throw new Error(
-          `${providerName}: doStream is not implemented on the delegating model — the streaming path uses executeStream directly.`,
+          `${providerName}: doStream is not implemented on the delegating model. ` +
+            `NeuroLink streams through executeStream, reached via NeuroLink.stream() — ` +
+            `use that (the browser bundle exports the NeuroLink class) rather than ` +
+            `calling doStream on a model handle.`,
         );
       },
     };
--- a/src/lib/providers/openaiChatCompletionsBase.ts
+++ b/src/lib/providers/openaiChatCompletionsBase.ts
@@ -1060,7 +1060,10 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
       },
       doStream: () => {
         throw new Error(
-          `${providerName}: doStream is not implemented on the delegating model — the streaming path uses executeStream directly.`,
+          `${providerName}: doStream is not implemented on the delegating model. ` +
+            `NeuroLink streams through executeStream, reached via NeuroLink.stream() — ` +
+            `use that (the browser bundle exports the NeuroLink class) rather than ` +
+            `calling doStream on a model handle.`,
         );
       },
     };
--- a/test/continuous-test-suite-browser-bundle.ts
+++ b/test/continuous-test-suite-browser-bundle.ts
@@ -21,6 +21,8 @@ import "dotenv/config";
  */
 
 import assert from "node:assert/strict";
+import { existsSync } from "node:fs";
+import { fileURLToPath } from "node:url";
 import type { NeuroLink } from "../dist/index.js";
 import { defineSuite, runCommand } from "./helpers/harness.js";
 import {
@@ -30,6 +32,19 @@ import {
 import { assertDistFresh } from "./helpers/distFreshness.js";
 
 assertDistFresh();
+
+// assertDistFresh only proves dist/index.js is present and newer than src/.
+// This suite's whole subject is the browser artifact, which build:browser
+// emits separately — so say precisely which build is missing.
+const bundlePath = fileURLToPath(
+  new URL("../dist/browser/neurolink.min.js", import.meta.url),
+);
+if (!existsSync(bundlePath)) {
+  throw new Error(
+    "dist/browser/neurolink.min.js is missing — this suite tests the browser " +
+      "bundle specifically. Run `pnpm run build` first (it runs build:browser).",
+  );
+}
 
 const { test, runSuite } = defineSuite("Browser bundle", { offline: true });
 
@@ -197,6 +212,74 @@ await test("browser artifact supplies callback timers when Node immediates are a
     console.error(result.stderr);
   }
   assert.equal(result.exitCode, 0, "browser timer fixture failed");
+});
+
+await test("importing the bundle and exiting leaves no uncaught error", async () => {
+  const { spawn } = await import("node:child_process");
+  const script = `await import(${JSON.stringify(bundleURL.href)}); console.log("ok");`;
+  const child = spawn(process.execPath, ["--input-type=module", "-e", script], {
+    stdio: ["ignore", "pipe", "pipe"],
+  });
+  let stdout = "";
+  let stderr = "";
+  child.stdout.on("data", (d: Buffer) => (stdout += d.toString()));
+  child.stderr.on("data", (d: Buffer) => (stderr += d.toString()));
+  // Bounded: this case is about teardown, so a bundle that keeps the child
+  // alive must fail readably rather than hang until the runner kills it.
+  const code: number = await new Promise((resolve) => {
+    const timer = setTimeout(() => {
+      child.kill("SIGKILL");
+      resolve(-2);
+    }, 30_000);
+    child.on("close", (c) => {
+      clearTimeout(timer);
+      resolve(c ?? -1);
+    });
+  });
+  assert.notEqual(code, -2, "the child never exited — the bundle held it open");
+  assert.ok(stdout.includes("ok"), "precondition: the bundle never imported");
+  assert.equal(
+    /TypeError|is not a function/.test(stderr),
+    false,
+    "the bundle raised a TypeError on the way out",
+  );
+  assert.equal(code, 0, "importing the bundle and exiting did not exit 0");
+});
+
+/**
+ * `doStream` on a factory handle is advertised as a function but cannot
+ * stream: the delegating model it forwards to implements generation only, and
+ * NeuroLink's own streaming runs through `executeStream`. The old assertion
+ * checked `typeof doStream === "function"` and so passed over this entirely.
+ * Pin the real behaviour instead — including that the message names the
+ * supported path — so the limitation cannot be mistaken for a capability.
+ */
+await test("a factory handle's doStream fails with an actionable message", async () => {
+  const make = bundle.createOpenAI as Factory;
+  const model = make({ apiKey: "sk-not-used" })("gpt-4o-mini");
+  assert.equal(
+    typeof model.doStream,
+    "function",
+    "precondition: doStream is not even advertised",
+  );
+  let message = "";
+  try {
+    await (model.doStream as (o: unknown) => Promise<unknown>)({ prompt: [] });
+    assert.fail("doStream resolved; it cannot stream and must say so");
+  } catch (error) {
+    message = error instanceof Error ? error.message : String(error);
+  }
+  // `includes("NeuroLink")` was too weak: it passed while the message was a
+  // malformed multi-line template literal whose text contained a stray quote,
+  // a `+` and the source indentation. Pin the exact string.
+  assert.equal(
+    message,
+    "openai: doStream is not implemented on the delegating model. " +
+      "NeuroLink streams through executeStream, reached via NeuroLink.stream() — " +
+      "use that (the browser bundle exports the NeuroLink class) rather than " +
+      "calling doStream on a model handle.",
+    "doStream's failure message is not the intended one",
+  );
 });
 
 await runSuite();
--- a/scripts/build-browser.mjs
+++ b/scripts/build-browser.mjs
@@ -81,7 +81,25 @@ export const isAbsolute = () => false;
 export const sep = '/';
 export const posix = { normalize:(p)=>p, join:(...a)=>a.join('/'), resolve:(...a)=>a.join('/'), sep:'/' };
 export const normalize = (p) => p;
-export const EventEmitter = class { on(){return this} off(){return this} emit(){return this} once(){return this} removeListener(){return this} addListener(){return this} removeAllListeners(){return this} };
+export const EventEmitter = class EventEmitter {
+  constructor(){ this._ev = new Map(); }
+  _list(t){ let l = this._ev.get(t); if(!l){ l=[]; this._ev.set(t,l);} return l; }
+  on(t,f){ this._list(t).push(f); return this; }
+  addListener(t,f){ return this.on(t,f); }
+  prependListener(t,f){ this._list(t).unshift(f); return this; }
+  once(t,f){ const g=(...a)=>{ this.off(t,g); f(...a); }; g.listener=f; return this.on(t,g); }
+  prependOnceListener(t,f){ const g=(...a)=>{ this.off(t,g); f(...a); }; g.listener=f; return this.prependListener(t,g); }
+  off(t,f){ const l=this._ev.get(t); if(l){ const i=l.findIndex(x=>x===f||x.listener===f); if(i>=0) l.splice(i,1); } return this; }
+  removeListener(t,f){ return this.off(t,f); }
+  removeAllListeners(t){ if(t===undefined) this._ev.clear(); else this._ev.delete(t); return this; }
+  emit(t,...a){ const l=this._ev.get(t); if(!l||l.length===0){ if(t==="error"){ throw a[0] instanceof Error ? a[0] : new Error("Unhandled error."); } return false; } for(const f of [...l]) f.apply(this,a); return true; }
+  listenerCount(t){ return (this._ev.get(t)||[]).length; }
+  listeners(t){ return [...(this._ev.get(t)||[])]; }
+  rawListeners(t){ return this.listeners(t); }
+  eventNames(){ return [...this._ev.keys()]; }
+  setMaxListeners(){ return this; }
+  getMaxListeners(){ return 10; }
+};
 export const AsyncLocalStorage = class { getStore(){} run(s,fn,...a){return fn(...a)} enterWith(){} disable(){} };
 export const Readable = class { pipe(){return this} on(){return this} read(){return null} push(){} destroy(){} };
 export const Writable = class { write(){return true} end(){} on(){return this} destroy(){} };


src/lib/providers/anthropic/client.ts
```diff
@@ -1634,7 +1634,10 @@ export class AnthropicProvider extends BaseProvider {
       },
       doStream: () => {
         throw new Error(
-          `${providerName}: doStream is not implemented on the delegating model — the streaming path uses executeStream directly.`,
+          `${providerName}: doStream is not implemented on the delegating model. " +
+            "NeuroLink streams through executeStream, reached via NeuroLink.stream() — " +
+            "use that (the browser bundle exports the NeuroLink class) rather than " +
+            "calling doStream on a model handle.`,
```
Tara-ag: <!-- yama:dostream-message-form===

**MAJOR — the new message is a malformed multi-line template literal, not the intended concatenation**

The intended string concatenation here isn't what's in the code. The first line ends with a double-quote `"` where the closing backtick should be, so the whole 4-line block parses as **one multi-line template literal**. The `"`, `+`, newlines and 12-space indentation become *literal text* inside the thrown error.

So the runtime message a user actually sees is:

```
anthropic: doStream is not implemented on the delegating model. " +
            "NeuroLink streams through executeStream, reached via NeuroLink.stream() — " +
            "use that (the browser bundle exports the NeuroLink class) rather than " +
            "calling doStream on a model handle.
```

It compiles (it IS a valid single template literal) and the new bundle test still passes (`message.includes("NeuroLink")` is true), which is exactly how a broken message slipped through — but a consumer who actually hits `doStream` gets an error with embedded source artifacts, not the clean actionable message this PR promises. Same defect in `openaiChatCompletionsBase.ts` (noted separately).

```suggestion
          `${providerName}: doStream is not implemented on the delegating model. NeuroLink streams through executeStream, reached via NeuroLink.stream() — use that (the browser bundle exports the NeuroLink class) rather than calling doStream on a model handle.`,
```

src/lib/providers/openaiChatCompletionsBase.ts
```diff
@@ -1060,7 +1060,10 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
       },
       doStream: () => {
         throw new Error(
-          `${providerName}: doStream is not implemented on the delegating model — the streaming path uses executeStream directly.`,
+          `${providerName}: doStream is not implemented on the delegating model. " +
+            "NeuroLink streams through executeStream, reached via NeuroLink.stream() — " +
+            "use that (the browser bundle exports the NeuroLink class) rather than " +
+            "calling doStream on a model handle.`,
```
Tara-ag: <!-- yama:dostream-message-openaichatbase ===

**MAJOR — same malformed multi-line template literal as `anthropic/client.ts`**

Identical defect and identical fix as flagged on `src/lib/providers/anthropic/client.ts`. The 4-line block parses as a single multi-line template literal, so the thrown error's `.message` literally contains `"`, `+`, newlines and leading spaces instead of the clean concatenated string. It still compiles and still passes the weak `includes("NeuroLink")` assertion, which is why it wasn't caught.

```suggestion
          `${providerName}: doStream is not implemented on the delegating model. NeuroLink streams through executeStream, reached via NeuroLink.stream() — use that (the browser bundle exports the NeuroLink class) rather than calling doStream on a model handle.`,
```

test/continuous-test-suite-browser-bundle.ts
```diff
@@ -170,4 +170,63 @@ for (const mode of ["generate", "stream"] as const) {
...
+  );
+  let message = "";
+  try {
+    await (model.doStream as (o: unknown) => Promise<unknown>)({ prompt: [] });
+    assert.fail("doStream resolved; it cannot stream and must say so");
+  } catch (error) {
+    message = error instanceof Error ? error.message : String(error);
+  }
+  assert.ok(
+    message.includes("NeuroLink"),
```
Tara-ag: <!-- yama:dostream-test-too-weak ===

**MINOR — `message.includes("NeuroLink")` is too weak to catch a malformed message**

This assertion would pass even for a garbled error message, because anything containing the word "NeuroLink" satisfies it. Concretely: the `doStream` messages currently thrown in both provider files are malformed multi-line template literals whose `.message` literally contains `"`, `+`, newlines and indentation — yet this assertion still passes, which is how that defect slipped through (see the inline flag on `src/lib/providers/anthropic/client.ts`).

Pin the message more tightly so a regression in the wording is actually caught — e.g. assert the full intended string, or at least that it has no embedded quote/concat artifacts:

```suggestion
  assert.ok(
    message.includes("NeuroLink.stream()"),
    "the failure does not name the supported streaming path",
  );
  assert.ok(
    !message.includes('" +'),
    "the message leaked string-concatenation artifacts",
  );
```

(At minimum, an exact-match on the corrected message would be ideal.)

test/continuous-test-suite-browser-bundle.ts
```diff
@@ -170,4 +170,63 @@ for (const mode of ["generate", "stream"] as const) {
...
+  const child = spawn(process.execPath, ["--input-type=module", "-e", script], {
+    stdio: ["ignore", "pipe", "pipe"],
+  });
+  let stdout = "";
+  let stderr = "";
+  child.stdout.on("data", (d: Buffer) => (stdout += d.toString()));
+  child.stderr.on("data", (d: Buffer) => (stderr += d.toString()));
+  const code: number = await new Promise((resolve) =>
+    child.on("close", (c) => resolve(c ?? -1)),
+  );
```
Tara-ag: <!-- yama:child-test-no-timeout ===

**MINOR — the import-and-exit child test can hang forever if the bundle keeps the child alive**

There's no timeout around the child. If the bundle registers a long-lived handle on the way in (an open socket, an active timer/event-loop owner), `close` never fires, this `await` never resolves, and the test hangs until the suite runner kills it — with a far less useful signal than a clean failure. Since this test is specifically about teardown behaviour, an early hang is exactly the kind of failure you want turned into a readable assertion.

Guard it so a stuck child fails fast instead of hanging:

```suggestion
  const code: number = await new Promise((resolve) => {
    const timer = setTimeout(() => {
      child.kill();
      resolve(-1);
    }, 30000);
    child.on("close", (c) => {
      clearTimeout(timer);
      resolve(c ?? -1);
    });
  });
```

The existing `assert.equal(code, 0, ...)` below then reports a stuck child as a normal failure with `code === -1`.

Tara-ag (commented):
Yama review of PR #1652.

Solid root-cause fix — the EventEmitter stub was the real bug (applied to three classes), and making doStream fail loudly (versus a silent no-op) is the right call. The two new bundle tests are appropriate (child-process import-and-exit is the only way to see the teardown crash).

However: the new doStream error message in both `anthropic/client.ts` and `openaiChatCompletionsBase.ts` is a malformed multi-line template literal (the first segment closes with `"` instead of a backtick), so the message a user actually receives contains literal `"`, `+` and indentation artifacts. It compiles and passes the new test because the assertion is a loose `includes("NeuroLink")`. Suggestions inline. The child-process test also has no timeout guard. All findings have inline comment markers.

Tara-ag:
## Yama review — PR #1652

**Verdict: NEEDS_WORK** — the core fix is correct, but the new doStream error message is malformed in both provider files and the new test is too weak to catch it.

### What was reviewed
- `scripts/build-browser.mjs` — the new real `EventEmitter` stub (listener storage, once/prepend, removeAllListeners, listenerCount/eventNames). This is the correct fix for the root cause: three classes in the bundle (`ToolDiscoveryService`, `ExternalServerManager`, `MCPCircuitBreaker`) extend it and six methods was never enough. The `error`-with-no-listener throw and the kept-`{}` default export are both handled correctly. Clean.
- `src/lib/providers/anthropic/client.ts` & `src/lib/providers/openaiChatCompletionsBase.ts` — doStream message change. Behavior is unchanged (still throws, no caller relies on success; no rule-5 break). **But the message content is broken.**
- `test/continuous-test-suite-browser-bundle.ts` — new import-and-exit + doStream tests. Right approach; two robustness gaps.

### Findings

| Severity | Location | Finding | Fix |
|---|---|---|---|
| MAJOR | `src/lib/providers/anthropic/client.ts:1637-1640` | doStream message is a malformed multi-line template literal; the thrown `.message` contains literal `"`, `+`, newlines and indentation | Replace with a single clean template literal (inline suggestion) |
| MAJOR | `src/lib/providers/openaiChatCompletionsBase.ts:1063-1066` | Same defect | Same fix (inline suggestion) |
| MINOR | `test/continuous-test-suite-browser-bundle.ts:225-227` | `message.includes("NeuroLink")` passes even for the garbled message — that's why the above slipped through | Tighten the assertion (inline suggestion) |
| MINOR | `test/continuous-test-suite-browser-bundle.ts:184-193` | Child-process import test has no timeout; a stuck child hangs the suite instead of failing fast | Add a kill/timeout guard (inline suggestion) |

All inline comments carry `<!-- yama:<> -->` markers and one-click `suggestion` blocks.

### Checks that came back clean
- EventEmitter fidelity (on/off/once/prepend/removeAllListeners/emit/error semantics) is adequate for the stub; `once`/prepend wrapper + `.listener` removal is correct.
- No back-compat break (rule 5): doStream always threw; only the text changed.
- Security: no secrets, no injection, `apiKey: "sk-not-used"` is a placeholder.
- Conventional commit + single-commit policy satisfied (bot confirmed).

Once the two message blocks are corrected and the assertion is tightened, this is good to go.

Tara-ag (changes requested):
Yama review of PR #1652 — verdict NEEDS_WORK.

The core fix (real EventEmitter for the browser bundle, honest doStream) is correct, but the new doStream error message in both `anthropic/client.ts` and `openaiChatCompletionsBase.ts` is a malformed multi-line template literal — the thrown `.message` contains literal `"`, `+`, newlines and indentation. It compiles and passes the new test because the assertion is a loose `includes("NeuroLink")`. The child-process test also has no timeout guard. All four findings have inline `suggestion` blocks on the exact lines.

src/lib/providers/anthropic/client.ts
Tara-ag: Note to the author: the four inline findings (two MAJOR on the malformed doStream message in `anthropic/client.ts` and `openaiChatCompletionsBase.ts`, two MINOR test-robustness) each carry a one-click `suggestion` block on the exact lines. All four inline comments plus the summary are intact. This changes-requested review reflects the total verdict — once the two message blocks are corrected and the assertion tightened, re-request review.

murdore: You are right, and this was my bug — the replacement was applied by a script that left the string open. Confirmed at runtime before fixing: the thrown `.message` literally contained `" +`, newlines and 12 spaces of indentation. Both files now concatenate template literals properly. Fixed and force-pushed.

src/lib/providers/openaiChatCompletionsBase.ts
murdore: Same fix applied here in the same push. Both sites were identical because both came from one scripted replacement.

test/continuous-test-suite-browser-bundle.ts
murdore: Agreed, and this is the finding that matters most of the three — the weak assertion is *why* the malformed message shipped past a green suite. It now asserts the exact expected string with `assert.equal`, so any wording regression fails rather than passing on the word 'NeuroLink' appearing somewhere.

test/continuous-test-suite-browser-bundle.ts
murdore: Adopted. The child is now bounded at 30s with SIGKILL and a distinct sentinel, and the case asserts on it — 'the child never exited — the bundle held it open'. Fair point that a test about teardown should never itself hang.

Tara-ag:
## Yama re-review — PR #1652 (after author's fix)

**Verdict: APPROVE** — all four findings from the previous round are resolved correctly in the force-pushed HEAD (`dbc3577`), and I found no new issues in the current diff.

### Prior findings — verified fixed
| Severity | Location | Status |
|---|---|---|
| MAJOR | `src/lib/providers/anthropic/client.ts:1637-1640` | ✅ doStream message now a properly-closed concatenation; no `"`/`+`/indent artifacts in the thrown text |
| MAJOR | `src/lib/providers/openaiChatCompletionsBase.ts:1063-1066` | ✅ same fix applied |
| MINOR | `test/continuous-test-suite-browser-bundle.ts` doStream case | ✅ `assert.equal` now pins the exact message string (was the reason the malformed message shipped) |
| MINOR | child-process import-and-exit case | ✅ bounded at 30s with SIGKILL + distinct `-2` sentinel |

### What I re-checked and found clean in this push
- **EventEmitter stub** (`scripts/build-browser.mjs`) — real listener storage via `Map`; `once`/`prependOnceListener` remove the wrapper *before* firing (`this.off(t,g)`); `off`/`removeListener` match either the raw fn or the `once` wrapper via `x===f || x.listener===f`; `emit` copies the list (`[...l]`) so remove-during-emit is safe; `error`-with-no-listener throws; `removeAllListeners()` clears the store. This resolves the original root-cause crash (`ToolDiscoveryService.destroy` → `ExternalServerManager.shutdown` calling `removeAllListeners`). Fidelity is adequate for the three bundle consumers.
- **doStream message naming** — `OpenAIProvider.getProviderName()` returns `AIProviderName.OPENAI = "openai"` (lowercase, `src/lib/constants/enums.ts`), so the test's pinned `"openai: doStream is not implemented..."` prefix is correct; the message truthfully names `NeuroLink.stream()` / `executeStream` as the supported streaming path. Backward compatible — `doStream` always threw here; only the text changed (rule 5 holds; no caller relies on it succeeding).
- **New `existsSync` guard** for `dist/browser/neurolink.min.js` is a correct, precise upgrade over `assertDistFresh()`.
- **Security** — no secrets; `apiKey: "sk-not-used"` is a placeholder.

One process note: I could not *edit* the earlier `<!-- yama:summary -->` comment (no comment-edit tool), so this is posted as a follow-up rather than an in-place update. The prior NEEDS_WORK verdict is superseded by this one.

No new inline comments — no findings warranted one.

Tara-ag (approved):
Yama review of PR #1652 — verdict **APPROVE**.

All four findings from the previous round (marked NEEDS_WORK) have been addressed by the author and verified against the current HEAD `dbc3577`:

- MAJOR `anthropic/client.ts` malformed doStream message — fixed, force-pushed.
- MAJOR `openaiChatCompletionsBase.ts` same defect — fixed.
- MINOR weak `includes("NeuroLink")` assertion — tightened to an exact `assert.equal`.
- MINOR child-process test with no timeout — now bounded at 30s with SIGKILL and a distinct sentinel.

The core fix (real EventEmitter for the browser bundle, honest `doStream`) is correct, backward-compatible, and clean on security and conventions. No new findings in this round. Good to merge.

Tara-ag:
## Yama review — PR #1652

**Verdict: APPROVE** — the EventEmitter stub and honest `doStream` are correct fixes; all four prior findings are resolved and verified against HEAD `dbc3577`.

> This summary supersedes the earlier `<!-- yama:summary -->` comment (which predated the author's fixes and said NEEDS_WORK). That comment cannot be edited in place, so this is the current authoritative summary.

### Findings resolution

| Severity | Location | Status |
|---|---|---|
| MAJOR | `src/lib/providers/anthropic/client.ts:1637-1640` | ✅ fixed — message now a single clean template literal |
| MAJOR | `src/lib/providers/openaiChatCompletionsBase.ts:1063-1066` | ✅ fixed |
| MINOR | `test/continuous-test-suite-browser-bundle.ts:225-227` | ✅ fixed — exact `assert.equal` on the message |
| MINOR | `test/continuous-test-suite-browser-bundle.ts:184-193` | ✅ fixed — 30s timeout + SIGKILL + sentinel |

- **EventEmitter stub** (`scripts/build-browser.mjs`): listener storage, on/off/once/prepend, `removeAllListeners`, `listenerCount`/`eventNames`, copy-on-emit, `error`-with-no-listener throw — correct. This is the genuine root-cause fix (three classes extend it).
- **Backward compatibility (rule 5)**: `doStream` always threw; no caller depended on it succeeding, so only the message text changed.
- **Security**: no secrets, no injection; `apiKey: "sk-not-used"` is a placeholder.
- **Conventions**: conventional commit + single-commit policy confirmed by bot; no registry/rule-1, Gemini/rule-3, or SDK-leak (rule 4) concerns.

Review state matches the verdict: an approving review has been submitted.

PR #1659, opened 2026-09-07
fix(core): take the Zod 4 native path only for Zod 4 schemas

murdore:
## The bug

`convertZodToJsonSchema` preferred Zod 4's native `z.toJSONSchema` whenever that function was importable:

```ts
if (zodToJsonSchemaV4) { ... }
```

That checks the **function**, not the **schema**. The two Zod majors coexist in real installs: a host application on Zod 3 still resolves NeuroLink's own Zod 4, so `zodToJsonSchemaV4` is defined while every schema handed in was built by the host's Zod 3.

Zod 4's converter reads `schema._zod.def`; a Zod 3 schema has only `_def`. So the call threw

```
TypeError: Cannot read properties of undefined (reading 'def')
```

on **every** tool-schema conversion — caught, warned, and fallen through to the Zod 3 path.

## Impact

Output was always correct, so this was invisible in behaviour. But a host on Zod 3 paid a thrown exception plus a `Native z.toJSONSchema failed` warning **per tool, per request**.

Found while upgrading Curator to 12.12.8: **240 occurrences in a single run**. Curator resolves `zod@3.25.76` because `@anthropic-ai/claude-agent-sdk` pins Zod 3, while NeuroLink 12.12.8 declares `zod ^4.3.6` — so both are installed and this fires constantly. Any host with the Agent SDK is in the same position.

Reproduction:

```js
const z3 = require("zod");            // 3.25.76
const v4 = require("zod/v4");         // present in 3.25.x, and NeuroLink's own zod@4
v4.toJSONSchema(z3.object({ a: z3.string() }));
// TypeError: Cannot read properties of undefined (reading 'def')
```

## The fix

Gate on the schema shape as well as the function:

```ts
if (zodToJsonSchemaV4 && isZod4Schema(zodSchema)) { ... }
```

`isZod4Schema()` keys off `_zod`, which only Zod 4 sets. Zod 3 schemas now go straight to the `zod-to-json-schema` path with no throw and no warning. Zod 4 schemas are unaffected.

## Test

`test/continuous-test-suite-zod3-schema-native-path.ts` (`pnpm test:zod3-schema-native-path`), offline, no API keys.

It asserts on the **warning**, not the returned schema. That distinction matters: both branches return a valid schema, so my first version of this test passed with the defect still planted. Verified by planting it:

- guard reverted to `if (zodToJsonSchemaV4)` → **exit 1**, `a Zod 3 schema never attempts the Zod 4 native path` fails
- guard restored → **exit 0**, 4/4 pass

The suite is on Rule 15's `allow` list: which branch ran is observable only through the logger, a live `generate()` cannot distinguish them, and this repo installs only Zod 4 so the Zod 3 side needs a shaped schema. The reason is stated in the file header, as the rule requires.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved schema conversion compatibility for Zod 3 schemas.
  - Zod 3 schemas now use the appropriate fallback conversion path instead of being incorrectly processed as Zod 4 schemas.
  - Zod 4 schemas continue to support native conversion, including OpenAPI 3 targets.
  - Prevented unnecessary conversion warnings for supported schema versions.

- **Tests**
  - Added coverage for Zod 3 and Zod 4 schema detection and conversion behavior.

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,6 +96,7 @@
     "test:openai-compat-streaming-retry": "pnpm exec tsx test/continuous-test-suite-openai-compat-streaming-retry.ts",
     "test:anthropic-streaming-retry": "pnpm exec tsx test/continuous-test-suite-anthropic-streaming-retry.ts",
     "test:adjust-body-after-400": "pnpm exec tsx test/continuous-test-suite-adjust-body-after-400.ts",
+    "test:zod3-schema-native-path": "pnpm exec tsx test/continuous-test-suite-zod3-schema-native-path.ts",
     "test:error-classification-e2e": "pnpm exec tsx test/continuous-test-suite-error-classification-e2e.ts",
     "test:error-classifier-contract": "pnpm exec tsx test/continuous-test-suite-error-classifier-contract.ts",
     "test:bedrock-inference-profile": "pnpm exec tsx test/continuous-test-suite-bedrock-inference-profile.ts",
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -367,6 +367,12 @@ export default [
         "error",
         {
           allow: [
+            // Which of convertZodToJsonSchema's two branches ran is visible
+            // only as a logger warning; the schema both branches return is
+            // valid, so a live generate() call cannot tell them apart. The
+            // regression also needs a Zod 3 shaped schema, and this repo
+            // installs only Zod 4.
+            "test/continuous-test-suite-zod3-schema-native-path.ts",
             // Chunk boundaries and reranker ordering are exact outcomes;
             // generate({ rag }) only ever shows the model's answer.
             "test/continuous-test-suite-rag.ts",
--- a/src/lib/utils/schemaConversion.ts
+++ b/src/lib/utils/schemaConversion.ts
@@ -481,7 +481,7 @@ export function convertZodToJsonSchema(
   // Translate our `target` to Zod 4's native dialect identifier so the
   // openApi3 path emits the OpenAPI 3 schema shape Vertex/Gemini expect
   // (and not the default draft-07 anyOf/null union).
-  if (zodToJsonSchemaV4) {
+  if (zodToJsonSchemaV4 && isZod4Schema(zodSchema)) {
     const nativeTarget: Zod4NativeTarget =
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -96,6 +96,7 @@
     "test:openai-compat-streaming-retry": "pnpm exec tsx test/continuous-test-suite-openai-compat-streaming-retry.ts",
     "test:anthropic-streaming-retry": "pnpm exec tsx test/continuous-test-suite-anthropic-streaming-retry.ts",
     "test:adjust-body-after-400": "pnpm exec tsx test/continuous-test-suite-adjust-body-after-400.ts",
+    "test:zod3-schema-native-path": "pnpm exec tsx test/continuous-test-suite-zod3-schema-native-path.ts",
```
Tara-ag: **MAJOR — the new suite never runs in CI (coverage gap).**

The `extended-suites` job in `.github/workflows/ci.yml` builds its run list from a **hardcoded `SUITES` string**, not a `test:*` glob scan, and `test:zod3-schema-native-path` is not in it. That workflow file is unchanged on this branch, so the suite this PR adds will never execute anywhere in CI — the job's own comment names this exact failure mode: *"The openai-compat catalog suite sat in no CI job, broke … and nobody noticed — because nothing ran it. A suite wired into nothing is documentation, not a test."*

Verified: searching for `test:zod3-schema-native-path` under `.github/workflows` on `release` returns 0 hits, while its sibling offline suites (`test:openai-compat-streaming-retry`, `test:anthropic-streaming-retry`, `test:adjust-body-after-400`, `test:error-classification-e2e`) are all present in the `SUITES` list.

This guard monitors the co-installed-Zod-3 host side of this fix — exactly the regression the suite exists to catch going silent. The right code lives in `.github/workflows/ci.yml` ('s `SUITES` list in `extended-suites`), which this diff does not touch, so I'm anchoring here on the package.json script line.

The suite is offline, credential-free, and its assertions are deterministic, so it fits `extended-suites`' environment contract. Please add one entry — the weights are load-balance hints, so any default (eg `test:zod3-schema-native-path` without `@weight` -> assumed `30s`) is safe to land with this change and re-measure later if it proves heavier.

PR #1661, opened 2026-09-07
fix(file-detector): fetch URL inputs once

### LINKED ISSUE #323: FD-007: Eliminate Double HTTP Request for URL Detection

emecii:
# Pull Request

## Description

Fetch URL inputs once and reuse the successful GET response's MIME type during file detection. This removes the initial HEAD request on the public URL path while keeping magic-byte detection ahead of the server MIME value.

Declared oversize responses are rejected from GET headers before the body-reading loop starts; the unread stream is closed before returning the size error.

## Related Issues

Fixes #323

## Type of Change

- [x] Performance improvement

## Changes Made

- Load URL inputs before detection and reuse the buffer throughout processing.
- Carry the GET `Content-Type` into the MIME strategy after magic-byte detection.
- Check a successful GET's declared length before reading its response body.
- Add request-count, MIME fallback, magic-byte priority, and declared-size regression coverage.

## Breaking Changes

- [x] No breaking changes

## New Provider Onboarding

## Testing

- [x] Unit tests added/updated
- [x] Existing relevant test suite passes

Commands run:

```text
pnpm run test:bugfixes       # 283 passed
pnpm run check               # passed
pnpm run build               # passed
pnpm run lint                # passed with existing repository warnings only
```

The repository pre-push hook also passed `check:deps`, the build, `test:provider-structure` (3 passed), and `test:model-manifests` (14 passed).

## Code Quality

- [x] Code follows the project's style guidelines
- [x] Code is properly formatted
- [x] No console.log statements in production code
- [x] No hardcoded API keys or secrets
- [x] TypeScript strict mode compliance
- [x] Proper error handling implemented

## Documentation

- [x] JSDoc comments added/updated for the new internal result type

## Commit Message Format

- [x] Commit message follows format: `type(scope): description`
- [x] Valid type used: `fix`
- [x] Scope specified

## Dependencies

- [x] No dependency changes

## Performance Impact

- [x] Performance improved

For a fresh URL processed through `FileDetector.detectAndProcess`, the local regression server observes one GET and zero HEAD requests. The test also covers a misleading MIME header, so binary magic bytes remain the first detection signal.

## Security Considerations

- [x] No new security implications identified

The existing redirect dispatcher, request error redaction, retry wrapper, and streaming size guard remain in the request path.

## Deployment Notes

- [x] No special deployment steps

## Additional Notes

The implementation does not rely on the prior short-lived URL MIME cache to avoid the initial request. It uses the same GET response for both content and detection metadata.

---

## Pre-submission Checklist

- [x] Read and followed the Contributing Guidelines
- [x] Verified automated pre-commit checks pass
- [x] Built the project successfully
- [x] Ran validation checks
- [x] Ensured the commit message follows semantic format
- [x] Added regression coverage

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved detection and processing of files loaded from URLs.
  - Oversized URL responses are now rejected earlier, helping prevent unnecessary downloads.
  - File types are identified more reliably using response metadata, file contents, and URL filenames.
  - URLs without MIME-type information can still be classified by their file extension.
  - Distinct signed URLs are handled independently, improving reliability when accessing dynamic links.

### DIFF
--- a/src/lib/utils/fileDetector.ts
+++ b/src/lib/utils/fileDetector.ts
@@ -352,7 +352,34 @@ export class FileDetector {
         },
       },
       async (span) => {
-        const detection = await FileDetector.detect(input, options);
+        // A URL used to be detected before it was loaded: MimeTypeStrategy
+        // issued a HEAD for Content-Type, then loadFromURL issued a GET for
+        // the body. Fetch it once and reuse the GET headers for the MIME
+        // strategy. A caller's explicit mimetypeHint remains authoritative,
+        // while a server header still comes after magic bytes.
+        const isUrlInput =
+          typeof input === "string" &&
+          (input.startsWith("http://") || input.startsWith("https://"));
+        const loadedUrl = isUrlInput
+          ? await FileDetector.loadUrl(input, options)
+          : undefined;
+        const detection = loadedUrl
+          ? {
+              ...(await FileDetector.detect(
+                loadedUrl.content,
+                {
+                  ...options,
+                  filenameHint:
+                    options?.filenameHint ||
+                    FileDetector.deriveInputFilename(input),
+                },
+                loadedUrl.contentType,
+                options?.filenameHint ||
+                  FileDetector.deriveInputFilename(input),
+              )),
+              source: "url" as const,
+            }
+          : await FileDetector.detect(input, options);
 
         span.setAttribute(ATTR.FILE_CATEGORY, detection.type);
         span.setAttribute(ATTR.FILE_MIMETYPE, detection.mimeType || "unknown");
@@ -367,11 +394,9 @@ export class FileDetector {
           options?.allowedTypes &&
           !options.allowedTypes.includes(detection.type)
         ) {
-          const content = await FileDetector.loadContent(
-            input,
-            detection,
-            options,
-          );
+          const content =
+            loadedUrl?.content ??
+            (await FileDetector.loadContent(input, detection, options));
           const errors: string[] = [];
 
           for (const allowedType of options.allowedTypes) {
@@ -432,11 +457,9 @@ export class FileDetector {
           return result;
         }
 
-        const content = await FileDetector.loadContent(
-          input,
-          detection,
-          options,
-        );
+        const content =
+          loadedUrl?.content ??
+          (await FileDetector.loadContent(input, detection, options));
         const csvOptions: CSVProcessorOptions | undefined = options?.csvOptions;
         const result = await FileDetector.processFile(
           content,
@@ -828,6 +851,8 @@ export class FileDetector {
   private static async detect(
     input: FileInput,
     options?: FileDetectorOptions,
+    responseMimeType?: string,
+    extensionInput?: string,
   ): Promise<FileDetectionResult> {
     // Short-circuit on a trustworthy caller-provided mimetype hint. This is
     // the eager-path counterpart to FileReferenceRegistry.register()'s hint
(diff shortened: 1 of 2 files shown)


murdore:
Sorry this sat — reviewed it properly now, verified against your branch head (`0c79f430e`), built and run rather than read.

**The CodeRabbit comment above does not reproduce.** Its scenario was a `report.csv` URL served as `application/octet-stream` with body `a,b`, claimed to resolve as text. It resolves as CSV:

```
URL         : http://127.0.0.1:PORT/report.csv
content-type: application/octet-stream   body: 'a,b'
type        : csv
mimeType    : text/csv
```

The reason is the line the bot didn't account for — `detect()` routes the derived filename to `ExtensionStrategy` specifically, rather than handing it the buffer like the other strategies:

```ts
strategy instanceof ExtensionStrategy && extensionInput ? extensionInput : input
```

To be sure the extension is genuinely contributing and not just being masked by a content heuristic that happens to agree, I held the bytes and headers constant and varied only the URL:

| URL | type | confidence |
| --- | --- | --- |
| `report.csv` | `csv` | 50 |
| `report.bin` | `text/plain` | 60 |
| `report.md` | `text/markdown` | 85 |

Three different answers from identical content, so the extension is reaching the strategy that needs it.

**And the PR's own claim checks out** — one request, no preflight:

```
GET requests : 1    HEAD requests: 0
```

which is exactly the HEAD-then-GET duplication this removes.

Nothing further needed from you. Thanks for the fix, and sorry for the wait.

PR #1662, opened 2026-09-08
chore(cleanup): clear the residue the ai-sdk removal left behind

murdore:
## What

The Vercel AI SDK removal finished last week. This clears what it left behind pointing at things that no longer exist.

A repo-wide sweep ran across six dimensions — dependencies, deleted symbols, dead exports, lint/CI guards, docs, tests — with an independent refutation pass on each dimension's findings. Residue turned up in all six.

## Comments (22 sites)

Every present-tense claim naming `GenerationHandler` is **re-pointed at whatever performs that work today**, not deleted, so the reasoning behind the code survives. Deliberate historical notes — "Ported from `GenerationHandler.X`", "the `@ai-sdk/openai` path this migration replaced" — are left alone: they explain why code looks the way it does, and `scripts/check-banned-deps.ts` legalises them on purpose.

Two corrections record a **real gap** rather than inventing an owner:

| gap | evidence |
| --- | --- |
| Nothing marks the last Anthropic tool definition with a cache breakpoint, so `cacheControlOf(tool)` always reads `undefined` and the tools block is never cached | `messageBuilder.ts` sets `providerOptions.anthropic.cacheControl` on **system messages only** (`:643`, `:1753`); the only readers are `nativeToolFormat.ts:76` and `anthropic/client.ts:1355` |
| Nothing joins V3 reasoning parts into `result.reasoning` on the generate path | `openaiChatCompletionsBase.ts` `doGenerate` pushes `{type:"reasoning"}` into content; `nativeGenerateLoop.ts` has no `reasoning` handling at all |

Both are now stated plainly at the call site. Neither is fixed here — each needs its own red test.

After this PR: **29 → 10** `GenerationHandler` mentions in `src/` + `test/`, all ten explicitly past-tense.

## Dead code

`src/lib/context/stepBudgetGuard.ts` (445 lines) had no importer — its only wiring point was `GenerationHandler.callGenerateText` through `experimental_prepareStep`.

**It loses no capability.** Per-step budget reclaim was reimplemented natively as `anthropicLoopGuard` / `geminiLoopGuard` / `openaiCompatLoopGuard` over a shared `loopGuardCore`, wired into all four provider families (anthropic, googleAiStudio, googleVertex, openaiChatCompletionsBase). Deleting it orphans nothing: `generateToolOutputPreview`, `DEFAULT_CONTEXT_GUARD_RATIO` and `createContextGuard` all have other live callers.

`GenerationCallConfig` and `StepBudgetGuardConfig` go with it. Both were exported but referenced nowhere, and both described the deleted loop — so this trims two names from the public type surface. Neither has any runtime presence, and no caller could have used either meaningfully. Their generated `docs/api` pages drop out.

## Guards — where the real finds were

**`eslint.config.js`** routed `"ai"` / `"@ai-sdk/provider"` imports "through the seam". Three things were wrong: the packages are banned outright by `check-banned-deps`, one named seam file (`src/lib/utils/generation.ts`) **does not exist**, and none of the other six imports either package.

Worse, the seam override switched off `no-restricted-syntax` **wholesale** for those files — and that rule also carries **Critical Rule 14** (no double type assertions). Removing the seam restores Rule 14 on six real files. Lint stays at 0 errors.

**`scripts/check-banned-deps.ts`** read only the **root** `package.json` and scanned only `src/`, `test/`, `scripts/`. It now walks every manifest in the repo and adds `tools/`, `eslint-rules/`, `examples/` and the sub-projects. That immediately caught residue nothing could previously see:

- `examples/structured-output-test.ts` imported `tool` from `"ai"`, on the strength of a comment saying NeuroLink "does NOT re-export `tool` on its public surface … until a future major release elevates it into the NeuroLink public API". **It does** — `src/lib/index.ts:384`. The example now imports from the package and the note is gone.
- `examples/client-sdks/vercel-ai-sdk.ts` is genuine **consumer-side** usage (it drives `createNeuroLinkProvider`), so `ai` there is the reader's dependency, not ours. Allowlisted via `ALLOWED_IMPORT_PREFIXES` with the reason written down.

**`neurolink-demo/package.json`** still declared `@ai-sdk/google-vertex` and `@ai-sdk/openai`. Nothing imported them; the guard never looked there. It has no lockfile and is not a workspace member, so `npm install` inside it would have pulled both fresh from the registry.

**`.clinerules`** taught `vi.mock("ai", …)`. There is no vitest runner and the package is banned; replaced with the e2e-via-`dist` contract the repo actually uses.

## Docs

`CLAUDE.md` rule 13 offered `@ai-sdk/provider` as an example of a library type you may "import normally" — contradicting the ban list every session also reads. Four finished plans said `Planning` / `in progress` and now carry completion markers; three research and report documents get a header saying they describe the pre-removal state. **No plan document was deleted** — they are the record.

`docs/guides/migration/from-vercel-ai-sdk.md` was checked closely and left alone: it is a guide for users migrating *from* the SDK *to* NeuroLink, and never claims NeuroLink depends on it.

## Verification

`check` 0 errors · `lint` 0 errors (58 pre-existing size warnings) · `build` 0 · `check:deps` passes under the widened scope · `check:test-parse`, `check:tools-tests`, `check:ci-scripts`, `test:provider-structure`, `test:model-manifests`, `validate:all` all exit 0 · **`test:providers-mocked` 95 passed / 0 failed** · **`verify:provider-onboarding` 8/8** · `docs/api` regenerated with the prettier pass CI applies, and idempotent on a second run.

## Deliberately not in this PR

`docs-site/static/llms-full.txt` (8.4 MB) and `search-index.json` (6.8 MB) still name `GenerationHandler` and `@ai-sdk`, and `search-index.json` ships in the npm tarball. Neither is CI-gated; both are produced by a post-merge deploy workflow that has `contents: read` and cannot commit back, and both are stale by **66 unrelated documentation commits**. Regenerating needs a full Docusaurus build and belongs in its own change rather than buried under this one.

## Summary by CodeRabbit

- **Documentation**
  - Updated testing guidance to use end-to-end tests against the built SDK or CLI, without module mocking.
  - Marked completed migration and removal plans as finished.
  - Added notices identifying historical AI SDK research and upgrade documents as archival.
  - Updated provider, tracing, caching, and structured-output documentation to reflect current behavior.

- **Chores**
  - Removed obsolete AI SDK dependencies, import restrictions, examples, and related context-budgeting functionality.
  - Expanded checks preventing removed dependencies from returning to the repository.


src/lib/core/nativeToolFormat.ts
```diff
@@ -71,8 +71,9 @@ export function toNativeToolDeclarations(
         ? convertZodToJsonSchema(rawSchema as never)
         : { type: "object", properties: {} }
     ) as Record<string, unknown>;
```
Tara-ag, line 73: **MINOR** — This new comment honestly documents a real hot-path behavior loss, but it trails off into prose with no pointer to a fix. `cacheControlOf(tool)` now always reads `undefined`, so the Anthropic tools block is never cached — a prompt-caching coverage regression on the direct-Anthropic path. The PR explicitly defers the fix ("each needs its own red test"), which is the right call, but "so this currently reads undefined" is exactly the kind of self-documenting note that gets lost once the flush settles.

Consider (a) a tracking issue reference here (and in the mirrors at `anthropic/client.ts:1354` and `cacheControl.ts`), and (b) a red test asserting the last tool definition *should* carry a breakpoint, so the gap fails loudly rather than silently. The `.clinerules`/Rule 15 e2e contract would have you drive this through a real Anthropic `generate()`/`stream()` and assert the wire `tools` block carries a `cache_control` marker.

Not a blocker — the code change itself (reframing a truthful comment) is correct. This is a maintainability nudge so the documented gap doesn't evaporate.

src/lib/providers/openaiChatCompletionsBase.ts
```diff
@@ -967,8 +967,10 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
             : "") ?? "";
         const content: Array<{ type: string } & Record<string, unknown>> = [];
         // Reasoner-model output (DeepSeek `reasoning_content`, gateway
-        // `reasoning`) becomes a V3 reasoning part ahead of the text part —
-        // GenerationHandler joins reasoning parts into `result.reasoning`.
+        // `reasoning`) becomes a V3 reasoning part ahead of the text part.
```
Tara-ag, line 970: **MINOR** — Honest documentation, but this is a user-facing behavior gap, not just a comment. On the shared native generate loop, `doGenerate` emits a V3 `reasoning` content part for reasoner outputs (DeepSeek `reasoning_content`, gateway `reasoning`) yet nothing joins it into `result.reasoning`, so openai-compatible providers return `undefined` reasoning even when the model emits CoT.

This is Rule 5 (backward-compatibility) relevant in spirit — the removed `GenerationHandler` used to populate `result.reasoning` from these parts, and the same loss is mirrored in the new comments at `baseProvider.ts:2208` and `neurolink.ts:5903`. I agree it should not be silently fixed mid-cleanup without a red test, but please give it a tracking issue reference here (e.g. a `// TODO(#issue)`), so the gap is findable. As written, "the part is emitted here and currently dropped" is accurate but leaves no actionable trail.

Not a blocker.

Tara-ag (approved):
Two MINOR, non-blocking nits on documenting-without-tracking. The substantive change (dead-code deletion, guard restoration) is verified correct.

Tara-ag:
**APPROVE** — An exemplary, thoroughly-verified AI-SDK-residue cleanup. Dead-code deletions are confirmed safe by the code graph, the eslint seam removal correctly restores Critical Rule 14, and the docs/comment re-pointing is honest — including two real gaps it transparently documents rather than hides. Only non-blocking observations are that the two documented gaps and the stale shipped docs lack tracking issue links.

### Findings

| Severity | file:line | Description |
| --- | --- | --- |
| MINOR | `src/lib/core/nativeToolFormat.ts:73` | New comment documents that the Anthropic tools block is never cached (`cacheControlOf(tool)` always `undefined`) — a hot-path prompt-caching gap deferred with no tracking issue / red test. Mirrors at `anthropic/client.ts:1354`, `cacheControl.ts`. |
| MINOR | `src/lib/providers/openaiChatCompletionsBase.ts:970` | Documents that V3 reasoning parts are emitted but dropped — `result.reasoning` is `undefined` for openai-compatible providers on the native loop (Rule 5-relevant). Also deferred without a tracking reference. |
| MINOR | `docs-site/static/*` (not in diff) | `llms-full.txt` + `search-index.json` stay stale and ship in the npm tarball; documented as deliberate, but no tracking issue. |

- **Dead-code deletion (`src/lib/context/stepBudgetGuard.ts`, `GenerationCallConfig`, `StepBudgetGuardConfig`)** — graph-verified safe: the three native loop guards (`anthropicLoopGuard`, `geminiLoopGuard`, `openaiCompatLoopGuard`) plus `loopGuardCore` are wired into the Anthropic / googleAiStudio / googleVertex / openai-compat paths; `generateToolOutputPreview` (6+ callers incl. the guards, `ToolsManager.truncateToolResult`) and `createContextGuard` (6 callers across all four provider families) retain live callers; zero orphaned references to any deleted symbol remain.
- **`eslint.config.js` seam removal** — the `ai`/`@ai-sdk/provider` seam block is gone and `no-restricted-syntax` (Critical Rule 14 / no-double-assertions) now runs on the previously-exempted files. Correct, and lint stays at 0.
- **`check-banned-deps.ts` widening** — now walks every manifest and scans `tools/`, `eslint-rules/`, `examples/`, subprojects; `ALLOWED_IMPORT_PREFIXES` correctly scoped to consumer-side `examples/client-sdks/`. `escapeRegex` fix is sound.
- **`.clinerules`** — replaced the dead `vi.mock("ai", …)` (no vitest runner) with the Rule 15 e2e-via-`dist` contract. Consistent.
- **Test change `res.object` → `res.structuredData`** (`continuous-test-suite-new-providers.ts`) — `structuredData` is the canonical current result field (used by `structuredRecovery.ts` / `recoverStructuredData`), so the update is consistent.
- **Comment re-pointing** across the 22 `GenerationHandler` sites — all re-point at the current implementer without factual drift.

No security issues: no secrets, no injection, no eval. No regression found in the deletions or the guard logic over the course of this review.

_Note: delegation/collect + the heavier graph subprocess tools (detect-changes, impact-radius) were flaky in this environment; blast-radius verification was done via live graph queries (importers/callers) and direct reading of the changed files, which covered the critical claims._

src/lib/core/nativeToolFormat.ts
```diff
@@ -71,8 +71,9 @@ export function toNativeToolDeclarations(
...
-    // GenerationHandler marks the last tool definition with a cache
-    // breakpoint when prompt caching is active — keep honoring it.
+    // Honor a cache breakpoint if a caller marked this tool. Nothing marks
+    // the last tool definition today: that was GenerationHandler's job and it
+    // went with the ai-package path, so this currently reads undefined.
```
Tara-ag, line 76: **MINOR — Anthropic prompt-cache breakpoint on the last tool definition is now unset**

This comment is honest about a real behavioural regression, but it is a known gap, not residue cleanup: the direct-Anthropic path no longer marks the last tool definition with a cache breakpoint, so `cacheControlOf(tool)` reads `undefined` for every tool and the tools block is never cached. The PR description records the same gap. Good that it's stated at the call site — but it is a silent behaviour change from the AI-SDK era, and there is no failing test guarding it.

Suggest: log the regression explicitly and add a red test that asserts the last tool definition carries `cache_control`, so the gap is visible and tracked rather than documented-only.

```suggestion
    // Honor a cache breakpoint if a caller marked this tool. Nothing marks
    // the last tool definition today: that was GenerationHandler's job and it
    // went with the ai-package path, so this currently reads undefined.
```

src/lib/providers/openaiChatCompletionsBase.ts
```diff
@@ -967,8 +967,10 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
...
+        // Nothing joins those parts into `result.reasoning` on the generate
+        // path today — GenerationHandler did, and went with the ai-package
+        // path — so the part is emitted here and currently dropped.
```
Tara-ag, line 973: **MINOR — V3 reasoning parts are dropped on the generate path**

This comment documents that `doGenerate` emits a V3 reasoning part but nothing joins it into `result.reasoning`, so a caller asking for `result.reasoning` on an OpenAI-compatible reasoner model (DeepSeek `reasoning_content`, gateway `reasoning`) gets `undefined`. This is a silent loss of pre-existing behaviour exposed to any unmodified caller of the public generate surface — and it is recorded as a comment only, with no test pinning the regression.

Suggest: add a red test asserting a reasoner-model response surfaces in `result.reasoning` (not just `analytics`), so the documented gap is converted into something the next change has to make pass.

```suggestion
    // Reasoner-model output (DeepSeek `reasoning_content`, gateway
    // `reasoning`) becomes a V3 reasoning part ahead of the text part.
    // Nothing joins those parts into `result.reasoning` on the generate
    // path today — GenerationHandler did, and went with the ai-package
    // path — so the part is emitted here and currently dropped.
```

Tara-ag (approved):
Approving — this residue cleanup is safe and well-verified; the two MINOR inline notes track real, documented behaviour gaps that deserve their own red tests (marked inline).

Tara-ag:
## APPROVE

The ai-sdk residue sweep is safe and well-verified: every deleted symbol was confirmed dead, every re-pointed doc claim and the eslint-seam removal are correct, and no caller or flow breaks. Two documented behaviour gaps are tracked as MINOR inline notes.

| --- | --- | --- |
| MINOR | `src/lib/core/nativeToolFormat.ts:74-76` | Anthropic last-tool cache breakpoint is no longer set (`cacheControlOf` reads `undefined`); tools block never cached — needs a red test, not doc-only |
| MINOR | `src/lib/providers/openaiChatCompletionsBase.ts:970-973` | V3 reasoning parts dropped on generate path → `result.reasoning` is `undefined` for OpenAI-compatible reasoners — needs a red test |

### Checked and found clean

- **Dead-code deletions** (`stepBudgetGuard.ts`, `GenerationCallConfig`, `StepBudgetGuardConfig`): all three native loop guards plus `loopGuardCore` are wired into every provider family; `generateToolOutputPreview`, `DEFAULT_CONTEXT_GUARD_RATIO`, `createContextGuard` retain live callers. No orphaned references (verified via importers/callers).
- **eslint seam removal** restores Critical Rule 14 (`no-restricted-syntax`) on six files; `check-banned-deps.ts` widened to all manifests and sub-projects; `neurolink-demo` no longer declares the banned packages; `.clinerules` no longer teaches `vi.mock("ai", …)`.
- **CLAUDE.md rule 13** now cites `@anthropic-ai/sdk` instead of the banned `@ai-sdk/provider`; four finished plans carry completion markers; the from-vercel guide is correctly left alone.
- **Tests**: provider-mocked suite updated to use canonical `res.structuredData`; no regression in e2e-via-`dist` contract.
- **No security issues**: no secrets, no injection, `transformParamsForLogging` path untouched.

Review state: APPROVE (submitted as an approving review). The two blockers are absent; the MINOR items are behaviour-gap tracking for follow-up changes, not blockers for this cleanup.

<details><summary>Notes on scope</summary>

`docs-site/static/llms-full.txt` and `search-index.json` remain stale (8.4 MB / 6.8 MB) and are deliberately deferred to a post-merge deploy workflow — correctly out of scope for this PR, as documented in the description.

</details>

Tara-ag:
temporary probe to confirm comment plumbing on this PR — will delete after verification.

PR #1663, opened 2026-09-08
chore(dead-code): run the tests nothing ran, delete what nothing reached

murdore:
## What

Follow-up to the ai-sdk residue sweep ([#1662](https://github.com/juspay/neurolink/pull/1662)). Three things in the tree were unreachable. Two turned out to be worth **running**, not deleting — and running one of them found a bug.

## The vitest suite that had never run

`test/codex-quota-observability.test.ts` (507 lines) imports vitest. `vitest ^4.1.0` is a devDependency, `vitest.config.ts` exists, and the file is deliberately on the rule-15 allowlist at `eslint.config.js:424` — someone meant it to run. But **no `package.json` script invokes vitest**, so it never had.

Wired as `test:codex-quota`. It needs `svelte-kit sync` first, or the transform fails resolving `.svelte-kit/tsconfig.json`.

**Running it found a wrong assertion.** The transport-failure case expected `errorType: "all_accounts_failed"`; the code emits `"network_error"`.

The code is right. `all_accounts_failed` is only the *initial default* at `codexProxyRoutes.ts:563`; the transport-failure branch overwrites `lastFailure` with the specific cause plus its `errorCode` and `transportScope` — the same shape the Anthropic engine records at `claudeProxyRoutes.ts:9142`. The expectation was written against the default and never checked.

Corrected to `network_error`, and **tightened** to also pin `errorCode: "ECONNRESET"`, so it asserts the specific contract rather than a generic label. 9/9 pass.

## The live suite nothing ran

`test/continuous-test-agents-live.ts` (370 lines) is a real suite with a documented invocation in its own header, importing from `dist/` per rule 15 — but no script ran it either. Wired as `test:agents-live`. It is live, so it stays out of CI, like the other live suites.

## Genuinely dead

- `src/lib/core/serviceRegistry.ts` — the `ServiceRegistry` class and its `serviceRegistry` singleton had zero importers anywhere in `src/`, `test/`, or the barrel.
- `AISDKGenerateResult` and `StepFinishEvent` in `src/lib/types/providers.ts` — unreferenced, and both describe the removed SDK's result and step shapes.

**Public-surface note:** `types/index.ts` re-exports `providers.ts` with `export *`, so those two names were public. Both are type-only — nothing breaks at runtime; a TypeScript consumer importing either by name would need to stop. They described a loop that no longer exists, so there was nothing they could correctly be used for.

## Corrected from the sweep

The sweep reported **13 orphaned test fixtures**. On inspection there are two, and neither should go:

- `generate-sample-screenshot.mjs` is the generator for `transactions-screenshot.png` and is *meant* to have no suite referencing it.
- `proxy-compatibility-v9.88.12.json` is a recorded reference-version contract snapshot.

## Verification

`check` 0 · `lint` 0 errors · `build` 0 · `check:deps` 0 · `check:ci-scripts` 0 · `check:test-parse` 0 · `test:provider-structure` 0 · `test:model-manifests` 0 · `validate:all` 0 · **`test:codex-quota` 9/9** · **`test:providers-mocked` 95/95** · `docs/api` regenerated and idempotent.

## Summary by CodeRabbit

- **Chores**
  - Added commands for running quota observability and continuous live-agent tests.
  - Updated transport-failure validation to check specific network error details.

- **Refactor**
  - Removed an internal service-resolution mechanism and associated access.
  - Removed unused provider-related type declarations.

- **Testing**
  - Improved transport-failure coverage by validating the expected network error type and connection-reset code.


package.json
```diff
@@ -154,6 +154,8 @@
     "test:tasks": "pnpm exec tsx test/continuous-test-suite-tasks.ts",
     "test:auth": "pnpm exec tsx test/continuous-test-suite-auth.ts",
     "test:autoresearch": "pnpm exec tsx test/continuous-test-suite-autoresearch.ts",
+    "test:codex-quota": "svelte-kit sync && vitest run test/codex-quota-observability.test.ts",
```
Tara-ag, line 157: **MINOR — wire `test:codex-quota` into the non-blocking `extended-suites` job.**

This suite sat unwired to anything for so long its assertion went stale — exactly the failure mode `ci.yml` already warns about: *"A suite wired into nothing is documentation, not a test"*, and the `OpenAI-compat catalog suite sat in no CI job, broke … and nobody noticed — because nothing ran it.* This PR fixes the assertion and adds the script, but leaves the suite un-wired into CI, so the same rot can recur.

It's well-suited for the `extended-suites` job: it is deterministic (mocks `fetch`, blocks provider/network access per its header), passes 9/9, and requires no credentials. The job already runs `pnpm exec svelte-kit sync` as a prerequisite step, so the script's own `svelte-kit sync &&` prefix is satisfied there too.

Suggested addition to the `SUITES` list in `.github/workflows/ci.yml` (non-blocking by design, so it can't gate a merge while proving itself):

```text
test:codex-quota@30
```

`test:agents-live` correctly stays out — it is a live suite by definition.

Tara-ag (commented):
Reviewed the dead-code purge. The deletions are clean and the test-assertion correction is verified correct against the production transport-failure path. One advisory MINOR on wiring the newly-revived suite into CI.

Tara-ag:
## REVIEW — APPROVE

Well-scoped dead-code purge. Deletions are genuinely dead, the type removals are disclosed type-only public-surface changes, and the `codex-quota` assertion fix is **verified correct** against the production code it asserts. One advisory MINOR on keeping the revived suite from rotting again.

|---|---|---|
| MINOR | `package.json:157` / `ci.yml` | Wire `test:codex-quota` into the non-blocking `extended-suites` job — it's deterministic and CI-runnable, and the suite's own history proves the cost of leaving it un-wired. |

### What I verified and found clean
- **`codex-quota` assertion fix** — `src/lib/server/routes/codexProxyRoutes.ts` transport-failure branch sets `lastFailure = { errorType: "network_error", errorCode, transportScope }`, overwriting the `all_accounts_failed` default. The test mocks an `ECONNRESET` fetch so `errorCode: "ECONNRESET"` and `transportScope: "connection_transport"` are correct. The tightened assertion (adding `errorCode`) asserts the real contract, not a label. ✓ Passes 9/9.
- **`serviceRegistry.ts` deletion** — no importers remain in the tree (repo `check:deps` clean; graph shows no references). ✓
- **`AISDKGenerateResult` / `StepFinishEvent` removal (Rule 5)** — type-only, removed cleanly with the now-unused `GenerateResult` import; nothing in-repo references them; the public-surface impact is disclosed in the PR body. Acceptable. ✓
- **`package.json` scripts** — both reference existing test files; rule-15 determinism exception is declared in the file header and the allowlist. ✓
- **`docs/api`** — regenerated consistently with the two type docs removed; idempotent. ✓
- Existing review threads: none on this PR (only the automated single-commit bot comment), so no prior findings to recur against.

Delegated parallel exploration of the blast radius and codex route internals; results were folded into the direct verification above.

package.json
Tara-ag, line 157: **MINOR (advisory)** — this newly-revived suite is deterministic (9/9) but left unwired to CI.

`test:codex-quota` is a real, deterministic vitest suite — that's exactly what this bug was about: a suite that never runs because nothing invokes it (`vitest ^4.1.0` was a devDependency, `vitest.config.ts` existed, the file sat on the rule-15 allowlist, but no script ran it — until it surfaced a wrong assertion). Reviving the script is good, but without wiring it into a CI job it can silently rot again the same way.

The repo's own `ci.yml` is blunt about this class of failure: *"a suite wired into nothing is documentation, not a test"* — and the guidance on unreachable suites cites the OpenAI-catalog crash from exactly this sin. Since it's deterministic and fast (9/9), it belongs in the non-blocking `extended-suites` group.

```suggestion
    "test:codex-quota": "svelte-kit sync && vitest run test/codex-quota-observability.test.ts",
    "test:codex-quota@30": "CI=1 pnpm test:codex-quota",
```
Keep the raw `test:codex-quota` script as-is, but register `test:codex-quota@30` in the CI `extended-suites` SUITES list per the standard pattern used by the other suite wrappers, so it runs headlessly on every push along with the other `@30` suites.

Tara-ag (approved):
Reviewed the dead-code purge. The deletions are clean (serviceRegistry has zero importers; the two type removals are type-only with disclosed public-surface impact), and the codex-quota assertion correction is verified correct against the production transport-failure path (`errorType: "network_error"` + `errorCode: "ECONNRESET"`, distinct from the `all_accounts_failed` default). One advisory MINOR on wiring the newly-revived suite into CI.

Tara-ag:
## APPROVE

Dead-code purge is clean and correct; the one advisory finding is optional. No blocking issues.

### Findings

| MINOR | `package.json:157` | Newly-revived `test:codex-quota` suite left unwired to CI — could silently rot again the same way this bug started |

### Verified clean

- **codex-quota assertion fix** — correct. The transport-failure branch in `codexProxyRoutes.ts` overwrites `lastFailure` with `errorType: "network_error"` + the specific `errorCode`/`transportScope` (same shape Claude records), so `all_accounts_failed` would only be the initial default. The tightened `errorCode: "ECONNRESET"` assertion pins the actual contract. 9/9 pass.
- **serviceRegistry.ts deletion** — zero importers anywhere; safe.
- **Type removals** (`AISDKGenerateResult`, `StepFinishEvent`) — both type-only; the `GenerateResult` import was correctly dropped with them. Both were public via `types/index.ts` `export *` (Rule 5 impact disclosed and acceptable — nothing breaks at runtime). Broken direct-type references are downstream in `dist` and regenerated.
- **package.json scripts** — both reference existing test files; the `test:codex-quota` prescript `svelte-kit sync` is required for transform resolution; rule-15 determinism exception properly declared for the vitest suite.
- **docs/api** — regenerated idempotently (line-shift only).

### Scope / impact

- 146 files changed, mostly `docs/api` line-shift churn. Genuine source changes: worker-delegated sweeps confirmed zero lingering importers of `ServiceRegistry` and that the assertion fix matches production behavior. No blast radius beyond the regenerated docs for the removed types.

Marked **APPROVE** — the MINOR is advisory and can be addressed in a follow-up.

src/lib/core/serviceRegistry.ts
```diff
@@ -1,147 +0,0 @@
-/**
- * Service Registry for Dependency Injection
- * Breaks circular dependencies by providing lazy loading and centralized service management
- */
-
```
Tara-ag, line 5: **MINOR — the deletion orphans the two types it was the sole consumer of.**

The removed file's import line was the only non-definition reference to `ServiceFactory` and `ServiceRegistration` anywhere in `src/`, `test/` or `scripts/`. With `serviceRegistry.ts` gone they are dead by this PR's own criterion — "zero importers anywhere in `src/`, `test/`, or the barrel" — but they remain on the public surface: `src/lib/types/index.ts:67` still does `export * from "./service.js"`, so they ship as public SDK types describing a registry that no longer exists. Same class of removal as `AISDKGenerateResult`/`StepFinishEvent`, just missed.

This comment anchors to the deleted import line because that's where the coupling lived; the fix lands in `src/lib/types/service.ts`.

**Option A — finish the purge (consistent with this PR):** delete the two aliases and regenerate `docs/api`, which removes `type-aliases/ServiceFactory.md` / `ServiceRegistration.md` and their README index rows — exactly what was done for the two ai-sdk types:

```suggestion
```

(applies to **`src/lib/types/service.ts:9-19`** — remove the `ServiceFactory` declaration at line 9-11, the `ServiceRegistration` declaration at lines 14-19, and the two doc comments above them; a suggestion block can't span files, so it's shown empty here with the exact target in prose)

**Option B — keep them intentionally:** add one comment at `service.ts:9` saying they survive as the public service-registry contract for external consumers, so the next sweep doesn't re-flag them.

Also worth deciding in the same pass: `ServiceRegistryStats` (`service.ts:78`) and siblings `ServiceDefinition`/`ServiceMetadata`/`ServiceHealth`/`ServiceStatus`/`ServiceConfigOptions` are equally unreferenced outside `service.ts` — they describe a registry this PR removes the only implementation of. I'm flagging the two the deleted file imported; the rest is your call on scope.

```
src/lib/types/service.ts:9    export type ServiceFactory<T = unknown> = () => T | Promise<T>;
src/lib/types/service.ts:14   export type ServiceRegistration<T = unknown> = {
src/lib/types/index.ts:67     export * from "./service.js";   // still re-exports both
(deleted) serviceRegistry.ts:5  import type { ServiceFactory, ServiceRegistration } from "../types/index.js";
```
Repo-wide identifier sweep (excluding `docs/api`): only the definition sites and the deleted importer — zero live consumers. The code-review-graph (17,176 nodes, head `cb027f6`) likewise shows no remaining importers of the deleted file.
</details>

Tara-ag (commented):
Re-review at `c36d270` — the deletion is verified clean of live importers, and the two prior findings both still stand (one carried forward in its existing thread). Details in the inline comment and the open thread; summary comment to follow on the PR.

package.json
Tara-ag: **Re-verified at head `c36d270` — still valid, keeping this thread open.**

No commit or reply has addressed it since the last round, and I re-read `.github/workflows/ci.yml` at this PR's head to confirm rather than assume: the `extended-suites` SUITES list (42 entries) contains no `test:codex-quota`, no other job in any workflow runs it, and this PR does not touch a workflow file. The finding therefore stands as written — with two small refinements from the re-read:

1. **The weight note was slightly off.** The prior suggestion implied `@30` was a standard wrapper; in fact the list is `name@seconds` where the number is a *measured CI timing* used only for shard balancing. An entry with no weight still runs, at an assumed 30s — so `test:codex-quota@30` is exactly the right shape for a first wire-in, with the number replaced by a real `::group::`-derived timing once a run exists (the awk recipe is in the job's own comments).

2. **One honest difference from the neighbours, worth a word in the job comment if wired in:** every other suite in that list drives `dist/` per rule 15, while this one is vitest-on-source (it imports `../src/...` directly). That's fine in this job — `svelte-kit sync` already runs there, `dist/` is built anyway, and the suite simply doesn't consume it — but the job's comment block ("every one of these suites imports from `../dist/index.js` per rule 15") becomes one suite less true.

Everything else from the first round checked out on re-verification and needs no further comment: the assertion fix is correct end-to-end (`ECONNRESET` → `getCodexTransportErrorCode` → `errorCode: "ECONNRESET"`, `codexTransportScope` → `"connection_transport"`, catch branch overwrites `lastFailure` with `errorType: "network_error"` and records the final 502 immediately since ECONNRESET is not in the retryable set — the old expectation was indeed asserting the initialization default), and `test:agents-live` correctly stays out of CI.

Tara-ag: **Re-verified at release tip `ae95a0a` (post-merge) — still valid, keeping this thread open.**

There is still no author reply on this thread, and the two commits that landed on `release` after this PR (`33e19ce` native-loop fix, `ae95a0a` docs-site artifacts) touch neither `ci.yml` nor the script. I re-ran the wiring check against the current tip rather than assuming:

- No workflow file in `.github/workflows/` mentions `test:codex-quota` (repo-wide search, zero hits).
- The `extended-suites` SUITES list still has 42 entries and contains only `test:codex@5` — a different suite (`continuous-test-suite-codex.ts`), so the quota-observability suite remains the only deterministic, credential-free, vitest-backed suite with a script and no job.
- The job still runs `pnpm exec svelte-kit sync` before the suites, satisfying the script's own `svelte-kit sync &&` prefix for free, and it is still non-blocking by design ("These are measured once, not proven stable"), so wiring it in cannot gate a merge.

The core argument is unchanged and was the PR's own thesis: a suite that exists but runs in no CI job is exactly how the stale `all_accounts_failed` assertion survived unnoticed. `test:codex-quota` is the direct fix for that class of rot and is currently the one thing proving it stays fixed nothing on CI.

The suggested shape stands as refined last round: add `test:codex-quota` to the SUITES list (no `@weight` — it runs at an assumed 30s, and the weight gets replaced with a real `::group::`-derived timing once a run exists per the awk recipe in the job's own comments), and amend the job comment "every one of these suites imports from `../dist/index.js` per rule 15" to note this one exception is vitest-on-source with the Rule-15 determinism exception declared in its header.

src/lib/core/serviceRegistry.ts
Tara-ag: **Re-verified at release tip `ae95a0a` (post-merge) — still valid, keeping this thread open.**

No author reply here either, and neither post-merge commit touches `src/lib/types/service.ts` or `src/lib/types/index.ts`. The re-verification, run fresh against the current tip:

- `ServiceFactory` (lines 9–11) and `ServiceRegistration` (lines 14–19) are still defined with no comment marking them as an intentional public-surface keep.
- Repo-wide searches for `ServiceFactory`, `ServiceRegistration` and `ServiceRegistryStats` outside `docs/` return exactly one file each — `src/lib/types/service.ts` itself. Zero consumers, same as when the finding was posted.
- `src/lib/types/index.ts` still does `export * from "./service.js"`, so both names still ship as public SDK types describing a registry whose only implementation this PR deleted.
- `docs/api/type-aliases/ServiceFactory.md` still exists and still renders the type, so the reference docs document a contract nothing implements or uses.

By the PR's own stated criterion — "zero importers anywhere in `src/`, `test/`, or the barrel" — these two types are dead and were simply missed, the same class of removal as `AISDKGenerateResult`/`StepFinishEvent`. Since this PR is merged, either option from the original comment works as a small follow-up: finish the purge (delete the two aliases, regenerate `docs/api`), or keep them deliberately with one comment at `service.ts:9` saying they survive as the public service-registry contract — so the next dead-code sweep doesn't have to re-derive the question from scratch.

Tara-ag:
## RE-REVIEW (post-merge, at release tip `ae95a0a`) — APPROVE, 2 advisory follow-ups carried forward

This is the recurring round. The PR's substance held up completely on re-verification against the current head; both previously-open findings remain valid because neither has an author reply and neither post-merge commit (`33e19ce`, `ae95a0a`) touches the files they concern. Each is replied to in its existing thread — no new findings.

### Findings (both carried forward, both MINOR)

| MINOR | `package.json:157` / `.github/workflows/ci.yml` | `test:codex-quota` still unwired to any CI job — re-checked: no workflow mentions it; the `extended-suites` SUITES list (42 entries) holds only the unrelated `test:codex@5`. Follow-up reply in [thread](https://github.com/juspay/neurolink/pull/1663#discussion_r3959203400). |
| MINOR | `src/lib/types/service.ts:9-19` | `ServiceFactory` / `ServiceRegistration` still orphaned by the `serviceRegistry.ts` deletion — zero consumers outside their own file, still public via `export * from "./service.js"`, docs page still shipping, no intentional-keep comment. Follow-up reply in [thread](https://github.com/juspay/neurolink/pull/1663#discussion_r3959188821). |

### Verified clean (re-checked at current tip, not assumed)

- **`codex-quota` assertion fix** — still exactly right. The transport-failure test (`test/codex-quota-observability.test.ts:325-374`) mocks an `ECONNRESET` fetch and asserts `errorType: "network_error"` + `errorCode: "ECONNRESET"` in terminal errors and additionally `transportScope: "connection_transport"` in the attempt log. Production confirms it: `codexProxyRoutes.ts:563` initializes `lastFailure` with `errorType: "all_accounts_failed"` and the transport-failure catch overwrites it with the specific cause + code via `getCodexTransportErrorCode` — the old expectation was asserting the initialization default.
- **`serviceRegistry.ts` deletion** — total: the code-review-graph (17,176 nodes at the merge commit) has no node for it, and repo-wide search finds no importers. No blast radius (graph impact analysis: 0 affected flows, 0 test gaps, risk 0.10).
- **`AISDKGenerateResult` / `StepFinishEvent` removal** — zero references anywhere outside `docs/api` (repo-wide search, both identifiers); the `docs/api` regeneration is consistent (only the two removed pages plus line-number shifts for everything below the deletions in `providers.ts`).
- **Script wiring & rule 15** — both scripts reference existing files; `vitest ^4.1.0` is a devDependency and `vitest.config.ts` exists with the isolation setup; the quota suite's header declares the Rule-15 determinism exception (mocks fetch, redirects HOME, blocks network); `test:agents-live` imports from `../dist/index.js` per rule 15 and is live, so staying out of CI is correct.

### Scope / impact

146 files, of which ~142 are `docs/api` regeneration churn (line-number shifts from the `providers.ts` deletions). The genuine changes are: 2 new scripts, 1 deleted file, 1 type-file edit, 1 test-file assertion fix. No runtime behavior change beyond the corrected test expectation. Blast-radius analysis (graph) shows the only changed function is the Codex transport-failure test itself.

Both findings are advisory and post-merge — well-suited to a single small follow-up PR (wire the suite into `extended-suites` + resolve the two orphaned types), or a one-line reply on each thread if either was deliberate.

PR #1664, opened 2026-09-08
fix(native-loop): surface vendor reasoning and cache the Anthropic tools block

murdore:
## What

Two contracts the native generate loop was supposed to carry and silently dropped.

Both surfaced while auditing the comments the ai-sdk removal left behind ([#1662](https://github.com/juspay/neurolink/pull/1662)): each credited `GenerationHandler` with work that nothing does any more. Deleting that class took the behaviour with it, and nothing noticed because no test covered either one.

## Reasoning never reached the caller

`openaiChatCompletionsBase`'s `doGenerate` builds a V3 `reasoning` content part for reasoner models — DeepSeek `reasoning_content`, gateway `reasoning`, OpenAI o-series — and `runNativeGenerateLoop` never looked at it. It collected only `text` parts.

Every other layer was already wired: `GenerateResult.reasoning` is declared at `generate.ts:1168`, `baseProvider.ts` forwards it, `neurolink.ts` forwards it. The only missing link was the one that fills it, so callers got `undefined` even when the model emitted a chain of thought.

The loop now joins reasoning parts on the same replace-per-step rule as text and returns them; all three loop callers — openai-compat, anthropic, sagemaker — surface it.

## The Anthropic tools block was never cached

`cacheControlOf(tool)` is read in two places and could never return anything: `MessageBuilder` marks **system messages** only (`messageBuilder.ts:643`, `:1753`), and nothing had marked a tool definition since `GenerationHandler` went away.

Tool definitions sit between the system prompt and the conversation and rarely change. With no breakpoint closing that prefix, the entire tools block was re-billed as fresh input on every turn.

The client now marks the last tool on the assembled request:
- **after** every tool mutation, including the appended `final_result` tool, so the marker really is last;
- **before** `countAnthropicCacheMarkers`, so the rolling history budget accounts for it and the four-marker ceiling still holds;
- a caller that marked a tool itself still wins.

## Red first

Both cases went into `test/continuous-test-suite-native-vendor-recovery.ts`, which already drives the native loop offline against local scripted servers — no credentials, no network.

Each asserts its **precondition before its claim**, so a pass cannot come from the request never happening:

| case | precondition | claim |
| --- | --- | --- |
| reasoning | endpoint called exactly once, and the scripted answer reached `content` | `result.reasoning` carries the vendor's reasoning |
| tool cache | a tools block actually arrived on the wire | exactly one ephemeral marker, and it is on the last tool |

**Unfixed: 4 passed, both new cases failed on their own claims. Fixed: 6/6.**

The tool-cache case first failed on its *precondition* rather than its claim — the SDK merges built-in tools with the caller's, so an exact tool count was the wrong assertion. Relaxed to a floor, which is what the test actually needs.

## Comments

The comments describing the old, broken state are corrected in the same change rather than left to drift again: `nativeToolFormat`, `anthropic/cacheControl`, `anthropicCacheBreakpoints` (both sites), `openaiChatCompletionsBase`, `baseProvider`, `neurolink`.

## Review order

This branch sits on `release`. [#1662](https://github.com/juspay/neurolink/pull/1662) edits several of the same comments to say the behaviour is **missing** — true at its merge point, and superseded here. Whichever lands second needs a small rebase on those hunks.

## Verification

`check` 0 · `lint` 0 errors · `build` 0 · `check:deps` 0 · `check:ci-scripts` 0 · `check:test-parse` 0 · `test:provider-structure` 0 · `test:model-manifests` 0 · `validate:all` 0 · **`test:vendor-recovery` 6/6** · **`test:providers-mocked` 95/95** · `test:error-classifier-contract` 0 · the **anthropic**, **vertex-claude**, **bedrock** and **aistudio** loop-characterization suites all 0 · `verify:provider-onboarding` 0 · `docs/api` regenerated.

## Summary by CodeRabbit

* **New Features**
  * Reasoning output from supported native generation providers is now included in generation results.
  * Anthropic tool requests now apply prompt-caching markers to the final tool when needed.

* **Bug Fixes**
  * Improved consistency of reasoning data across native generation paths.
  * Added safeguards to ensure Anthropic requests contain the expected prompt-caching breakpoint.
  * Added coverage to verify reasoning output and prompt-caching behavior across supported native provider flows.

### DIFF
--- a/src/lib/core/nativeGenerateLoop.ts
+++ b/src/lib/core/nativeGenerateLoop.ts
@@ -136,6 +136,7 @@ export async function runNativeGenerateLoop(
 ): Promise<NativeGenerateLoopResult> {
   const toolsUsed: string[] = [];
   let text = "";
+  let reasoning = "";
   let finishReason = "stop";
   let rawFinishReason: string | undefined;
   let inputTokens = 0;
@@ -224,6 +225,15 @@ export async function runNativeGenerateLoop(
     // answer is the turn's answer, matching what generateText reported.
     text = parts
       .filter((p) => p.type === "text" && typeof p.text === "string")
+      .map((p) => p.text as string)
+      .join("");
+    // Reasoner models (DeepSeek `reasoning_content`, gateway `reasoning`,
+    // OpenAI o-series) emit reasoning as its own V3 content part. Join it on
+    // the same replace-per-step rule as the text so the caller can surface
+    // `result.reasoning`; without this the provider builds the part and the
+    // loop drops it.
+    reasoning = parts
+      .filter((p) => p.type === "reasoning" && typeof p.text === "string")
       .map((p) => p.text as string)
       .join("");
 
@@ -352,6 +362,7 @@ export async function runNativeGenerateLoop(
 
   return {
     text,
+    ...(reasoning ? { reasoning } : {}),
     finishReason,
     ...(rawFinishReason ? { rawFinishReason } : {}),
     inputTokens,
--- a/src/lib/providers/anthropic/client.ts
+++ b/src/lib/providers/anthropic/client.ts
@@ -1350,9 +1350,8 @@ export class AnthropicProvider extends BaseProvider {
         let tools: Anthropic.Messages.Tool[] | undefined = (options.tools ?? [])
           .filter((t) => t.type === "function")
           .map((t) => {
-            // Honor a cache breakpoint if a caller marked this tool. Nothing
-            // marks the last tool definition today: that was
-            // GenerationHandler's job and it went with the ai-package path.
+            // Honor a cache breakpoint the caller set on this tool. When no
+            // tool carries one, the last tool is marked further below.
             const cc = cacheControlOf(t);
             return {
               name: t.name,
@@ -1420,9 +1419,27 @@ export class AnthropicProvider extends BaseProvider {
           | { type: "enabled"; budget_tokens: number }
           | undefined;
 
+        // Close the stable prefix with a breakpoint on the LAST tool. Tool
+        // definitions sit between the system prompt and the conversation and
+        // rarely change, so without this the whole tools block is re-billed
+        // every turn. Applied after every tool mutation above (including the
+        // appended final_result tool) so the marker really is last, and before
+        // the count below so the history budget accounts for it. A caller that
+        // marked a tool itself wins.
+        if (tools && tools.length > 0) {
+          const alreadyMarked = tools.some((t) => cacheControlOf(t));
+          if (!alreadyMarked) {
+            const last = tools[tools.length - 1];
+            tools = [
+              ...tools.slice(0, -1),
+              { ...last, cache_control: { type: "ephemeral" } },
+            ];
+          }
+        }
+
         // Prompt-cache parity with the native Vertex+Claude path: upstream
-        // layers mark only the stable prefix (the system prompt, via
-        // MessageBuilder) — the growing conversation
+        // layers mark the stable prefix (system via MessageBuilder, and the
+        // last tool just above) — the growing conversation
         // history has no breakpoint, so on every turn it falls after the
         // last marker and is re-billed as fresh input. Add rolling history
         // breakpoints in whatever budget remains under Anthropic's
@@ -1885,6 +1902,7 @@ export class AnthropicProvider extends BaseProvider {
       ...(loop.rawFinishReason
         ? { rawFinishReason: loop.rawFinishReason }
         : {}),
+      ...(loop.reasoning ? { reasoning: loop.reasoning } : {}),
       usage: {
         input: loop.inputTokens,
         output: loop.outputTokens,
--- a/docs/api/classes/NeuroLink.md
+++ b/docs/api/classes/NeuroLink.md
@@ -469,7 +469,7 @@ Gracefully shutdown NeuroLink and all MCP connections
 
 > **generateText**(`options`): `Promise`\<[`TextGenerationResult`](../type-aliases/TextGenerationResult.md)\>
 
-Defined in: [neurolink.ts:6508](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L6508)
+Defined in: [neurolink.ts:6507](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L6507)
 
 BACKWARD COMPATIBILITY: Legacy generateText method
 Internally calls generate() and converts result format
@@ -490,7 +490,7 @@ Internally calls generate() and converts result format
 
 > **streamText**(`prompt`, `options?`): `Promise`\<`AsyncIterable`\<`string`, `any`, `any`\>\>
 
-Defined in: [neurolink.ts:9005](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L9005)
+Defined in: [neurolink.ts:9004](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L9004)
 
 BACKWARD COMPATIBILITY: Legacy streamText method
 Internally calls stream() and converts result format
@@ -515,7 +515,7 @@ Internally calls stream() and converts result format
 
 > **stream**(`options`): `Promise`\<[`StreamResult`](../type-aliases/StreamResult.md)\>
 
-Defined in: [neurolink.ts:9088](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L9088)
+Defined in: [neurolink.ts:9087](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L9087)
 
 Stream AI-generated content in real-time using the best available provider.
 This method provides real-time streaming of AI responses with full MCP tool integration.
@@ -588,7 +588,7 @@ When conversation memory operations fail (if enabled)
 
 > **setToolRoutingServers**(`servers`): `void`
 
-Defined in: [neurolink.ts:9980](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L9980)
+Defined in: [neurolink.ts:9979](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L9979)
 
 Supplies (or replaces) the pre-call tool routing server catalog.
 
@@ -613,7 +613,7 @@ alone does not activate it.
 
 > **getKnowledgeStatus**(): [`KnowledgeEngineStatus`](../type-aliases/KnowledgeEngineStatus.md) \| `null`
 
-Defined in: [neurolink.ts:9998](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L9998)
+Defined in: [neurolink.ts:9997](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L9997)
 
 Knowledge-grounding engine health (null when it was not configured).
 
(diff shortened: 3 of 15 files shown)


src/lib/providers/anthropic/client.ts
```diff
@@ -1419,9 +1419,27 @@ export class AnthropicProvider extends BaseProvider {
...
 
+        // Close the stable prefix with a breakpoint on the LAST tool. Tool
+        // definitions sit between the system prompt and the conversation and
+        // rarely change, so without this the whole tools block is re-billed
+        // every turn. Applied after every tool mutation above (including the
+        // appended final_result tool) so the marker really is last, and before
+        // the count below so the history budget accounts for it. A caller that
+        // marked a tool itself wins.
+        if (tools && tools.length > 0) {
+          const alreadyMarked = tools.some((t) => cacheControlOf(t));
```
Tara-ag, line 1430: **MINOR — the `alreadyMarked` guard is dead code, so the documented "caller that marked a tool still wins" behaviour never triggers.**

By this point `tools` has already been flattened to Anthropic *wire* objects (`{ name, description, input_schema, ...(cc ? { cache_control } : {}) }` above). `cacheControlOf(t)` (read it in `cacheControl.ts`) looks for `carrier.providerOptions?.anthropic?.cacheControl` — the NeuroLink/V3 shape — but wire tools carry the raw `cache_control` field, not `providerOptions`. So `tools.some((t) => cacheControlOf(t))` is always `false`, and the block always unconditionally stamps the last tool.

Practically:
- When the caller set a marker on the **last** tool: harmless (idempotent re-stamp).
- When the caller set a marker on a **non-last** tool: a second, redundant marker is added to the request, which silently burns one history-breakpoint budget slot (the marker is counted by `countAnthropicCacheMarkers`, so no 4-cap violation — but the caller's intent is not honoured as stated).

This also applies to the earlier map's `const cc = cacheControlOf(t);` — on the native-loop path the V3 tools (`v3Tools` in `executeNativeGenerate`) carry no `providerOptions`, so `cc` is always `undefined` there too.

Suggestion: check the *wire* field that is actually present instead of going back through `cacheControlOf`:

```suggestion
          const alreadyMarked = tools.some((t) => (t as { cache_control?: unknown }).cache_control);
```

That makes the guard reflect the real wire state, so the "caller wins" case is preserved when a tool already carries a marker.

src/lib/core/nativeGenerateLoop.ts
```diff
@@ -226,6 +227,15 @@ export async function runNativeGenerateLoop(
...
       .join("");
+    // Reasoner models (DeepSeek `reasoning_content`, gateway `reasoning`,
+    // OpenAI o-series) emit reasoning as its own V3 content part. Join it on
+    // the same replace-per-step rule as the text so the caller can surface
+    // `result.reasoning`; without this the provider builds the part and the
+    // loop drops it.
+    reasoning = parts
+      .filter((p) => p.type === "reasoning" && typeof p.text === "string")
+      .map((p) => p.text as string)
+      .join("");
```
Tara-ag, line 238: **MINOR — `reasoning` uses replace-per-step, so in a tool-calling turn only the *final* step's reasoning survives.**

This block reassigns `reasoning` from scratch every step the same way `text` does. For a plain single-shot answer the two are equivalent, but for a reasoner model driving a **tool loop** (e.g. DeepSeek reasoner planning tool calls), the chain-of-thought is emitted on the *intermediate* steps, and the final answer step often carries no `reasoning_content`. The result: `result.reasoning` comes back empty (`...(reasoning ? { reasoning } : {})` drops it) precisely for the tool-using turns where surfaced reasoning is most valuable.

`text` replacing per step is correct — the final step's answer *is* the turn's answer. Reasoning is not the answer, it is a cumulative chain, so appending (like the token/usage counters in the same function) is the more faithful contract:

```suggestion
    reasoning += parts
      .filter((p) => p.type === "reasoning" && typeof p.text === "string")
      .map((p) => p.text as string)
      .join("");
```

If the final-step-only behaviour is intentional, please say so in the doc comment on `NativeGenerateLoopResult.reasoning` — right now the added test only covers the single-step case, so the tool-loop gap is invisible.

Tara-ag (commented):
Reviewed the native-loop reasoning surfacing + Anthropic tool-cache changes. Two MINOR findings; the rest of the change is correct and well tested.

Tara-ag:
## Verdict: **APPROVE** — two MINOR nits, neither blocking; the change is correct, red-first tested, and restores two silently-dropped native-loop contracts.

### Findings

| --- | --- | --- |
| MINOR | `src/lib/providers/anthropic/client.ts:1429-1438` | `alreadyMarked` guard is dead code — `cacheControlOf()` reads `providerOptions.anthropic.cacheControl`, but `tools` at that point are flattened wire objects carrying `cache_control`. The documented "caller that marked a tool wins" path never triggers; a caller-marked non-last tool gets a redundant second marker (bounded by `countAnthropicCacheMarkers`, so no 4-cap violation). Suggest checking the wire `cache_control` field. |
| MINOR | `src/lib/core/nativeGenerateLoop.ts:235-238` | `reasoning` uses replace-per-step like `text`. For tool-using reasoner turns, the chain-of-thought lives on intermediate steps and the final answer step usually has no `reasoning_content`, so `result.reasoning` comes back empty exactly when it is most valuable. Consider accumulating across steps, or document the final-step-only intent (the new test covers only the single-step case). |

- **Reasoning wiring is end-to-end complete**: `openaiChatCompletionsBase.doGenerate` emits `{ type: "reasoning", text }` parts (both JSON and SSE wire, `reasoning_content || reasoning` — `||` so an empty-string `reasoning_content` falls through), the loop joins them, and all three loop callers (openai-compat, anthropic, sagemaker) surface `...(loop.reasoning ? { reasoning } : {})`. `baseProvider.ts` and `neurolink.ts` already forwarded `reasoning` on the public surface (Rule 5 — the addition to `NativeGenerateLoopResult` is optional/additive, no caller breaks).
- **Anthropic tool-cache marker ordering is correct**: placed after every tool mutation (including appended `final_result`, so it really is last, `client.ts:1408-1415` → `1422-1438`) and before `countAnthropicCacheMarkers` (`1448-1456`), so the 4-marker budget accounts for it. The rolling-history budget math stays within the cap.
- **Blast radius**: ran the code graph — `runNativeGenerateLoop` has exactly the 3 expected callers; the wide 500-node impact surface is mechanical (additive optional fields + comment rewrites), with no signature, nullability, async, or thrown-error changes that an unmodified caller must now handle.
- **Tests**: brand-new cases in `continuous-test-suite-native-vendor-recovery.ts` assert their precondition before their claim (no pass from a request never happening); the tool-cache case correctly uses a floor (>2) after the SDK merging built-ins. `test:vendor-recovery 6/6`, `test:providers-mocked 95/95`. No recurring findings to reconcile — this is the first review of this PR (only the automated single-commit bot comment exists).

No CRITICAL/MAJOR issues. No security, secret, or hot-path (`baseProvider`, `providerRegistry`, `types/**`) regressions introduced.

src/lib/providers/anthropic/client.ts
```diff
@@ -1419,9 +1419,27 @@ export class AnthropicProvider extends BaseProvider {
...
+          if (!alreadyMarked) {
+            const last = tools[tools.length - 1];
+            tools = [
+              ...tools.slice(0, -1),
+              { ...last, cache_control: { type: "ephemeral" } },
+            ];
+          }
+        }
```
Tara-ag, line 1438: **MINOR — the "caller already marked a tool wins" guard is dead code.**

`cacheControlOf(t)` (`cacheControl.ts`) reads `providerOptions.anthropic.cacheControl` — the AI-SDK-era carrier field. But `tools` here is the flattened wire `Anthropic.Messages.Tool[]`, whose objects carry `cache_control`, not `providerOptions`. So `alreadyMarked` is always falsy and the guard never fires.

Two consequences:
- A caller that marked a **non-last** tool still gets a redundant second `ephemeral` marker stamped on the last tool (the PR and the comment both promise "a caller that marked a tool itself wins" — it doesn't).
- The comment above the block overstates the behavior.

Check the wire field instead:

```suggestion
          const alreadyMarked = tools.some((t) => t.cache_control);
```

The native `executeNativeGenerate` path still honors a caller's marker when it builds the tools via `toNativeToolDeclarations` (`cacheControlOf` is read on the raw caller tool there) — this is only the assembled-request guard that is currently ineffective.

src/lib/core/nativeGenerateLoop.ts
```diff
@@ -226,6 +227,15 @@ export async function runNativeGenerateLoop(
       .filter((p) => p.type === "text" && typeof p.text === "string")
       .map((p) => p.text as string)
       .join("");
+    // Reasoner models (DeepSeek `reasoning_content`, gateway `reasoning`,
+    // OpenAI o-series) emit reasoning as its own V3 content part. Join it on
+    // the same replace-per-step rule as the text so the caller can surface
+    // `result.reasoning`; without this the provider builds the part and the
+    // loop drops it.
+    reasoning = parts
+      .filter((p) => p.type === "reasoning" && typeof p.text === "string")
```
Tara-ag, line 236: **MINOR — replace-per-step discards the chain-of-thought on tool-loop steps.**

This runs inside the per-step loop and replaces `reasoning` each step, on the same rule as `text`. For `text` that is correct — tool-loop steps emit tool-calls, and only the final step carries the answer, so "final step wins" is right. But reasoning inverts that: with a reasoner in a multi-step tool loop, the **intermediate** steps are exactly where the meaningful chain-of-thought lives (the model reasons, then calls a tool), and the final step's `reasoning_content` is usually minimal. Replace-per-step therefore leaves `result.reasoning` empty or partial precisely when the caller wants it — GenRule: this PR's whole point is "surface vendor reasoning."

`reasoning` (unlike `text`) should accumulate across steps. Suggested:

```suggestion
    reasoning += parts
      .filter((p) => p.type === "reasoning" && typeof p.text === "string")
      .map((p) => p.text as string)
      .join("");
```

(If dropping intermediate reasoning is deliberate, a one-line comment stating that would prevent a future revert.)

Tara-ag (approved):
Two MINOR findings; the reasoning-surfacing and tool-cache wiring is otherwise correct end to end. Approving.

Tara-ag:
## Verdict: APPROVE
Both findings are MINOR; the reasoning-surfacing and Anthropic tool-cache wiring is sound end to end.

| Severity | File:Line | Finding |
|---|---|---|
| MINOR | `src/lib/providers/anthropic/client.ts:1430` | `alreadyMarked` guard is dead — `cacheControlOf` reads `providerOptions` but the wire tools carry `cache_control`, so a caller-marked non-last tool still gets a redundant second marker. |
| MINOR | `src/lib/core/nativeGenerateLoop.ts:233-236` | `reasoning` replaces per step, discarding the chain-of-thought on intermediate tool-loop steps where it's most valuable. |

- **Reasoning wiring, end to end.** `reasoning` part construction in `openaiChatCompletionsBase.doGenerate` and Anthropic `doGenerate`; the loop join → `NativeGenerateLoopResult.reasoning` (`types/generate.ts`); all three loop callers surface it — openai-compat (`openaiChatCompletionsBase.ts:1330`), anthropic (`client.ts:1905`), sagemaker (`amazonSagemaker.ts:287`); `baseProvider.ts` + `neurolink.ts` forwarding. No path drops it.
- **Tool-cache wiring.** Marker applied after all tool mutations (`final_result` included) and before `countAnthropicCacheMarkers`; the 4-marker ceiling holds. `cacheControlOf` shared without circular import.
- **Blast radius.** `runNativeGenerateLoop` has exactly 3 callers (openai-compat, anthropic, sagemaker), all updated; no unmodified caller broken; no SDK API surface change (only a new optional `reasoning?` field). No regressions to `text` accumulation. Test coverage added for both behaviors with proper preconditions in `continuous-test-suite-native-vendor-recovery.ts`.
- No CRITICAL/security, rule 1/3/4/5 issues; formatting/lint/type enforcement left to CI.

Reviewed via diff + code graph (callers, blast radius, flow of both features).

PR #1666, opened 2026-09-08
chore(docs-site): make the generated artifacts reproducible and check them in CI

murdore:
## What

Makes `docs-site/static/{llms.txt,llms-full.txt,search-index.json}` **reproducible**, then adds a CI check that keeps them current.

These files are generated, committed, and shipped — `search-index.json` is listed in `package.json` `files`, so the copy in this repo is the copy inside the published npm tarball. Nothing kept them current, and they drifted **66 documentation commits** before anyone noticed ([#1665](https://github.com/juspay/neurolink/pull/1665) was the manual catch-up).

## Why not just regenerate them post-merge

The obvious fix — regenerate in `docs-deploy.yml` and commit the result — cannot work. It would need `contents: write` **and** a push to `release` from outside a pull request. Such a push carries no check runs, so all five required contexts read as missing and it is declined:

```
GH013: Repository rule violations found for refs/heads/release
- 4 of 4 required status checks are expected.
```

That is the same wall that forced `@semantic-release/git` out of the release job. So currency is enforced where a push *can* happen: on the pull request that changes the docs.

## The part that would have shipped broken

A "regenerate and diff" check requires a reproducible generator. **This one was not**, in two ways — and I only found the second because I built twice and compared rather than assuming.

**1. A wall-clock timestamp.** `build-llms-txt.ts` stamped `new Date().toISOString()` into both llms files. No two runs could ever match.

**2. A sort that is not a total order.** This is the one that hides:

```js
docFiles.sort((a, b) => a.order - b.order);
```

Many files share an `order`. `Array.prototype.sort` is stable, so ties fell back to the input order — which came from `glob()`, which is filesystem order.

> **Two consecutive builds of an unchanged tree differed by 215,862 lines in `llms-full.txt`.**

`sortSections` had the same defect: every section outside `SECTION_ORDER` scores 999, so their relative order came from `Map` insertion order, i.e. file order again.

Had I written the check without testing determinism, it would have been **permanently red** — the same shape of always-failing required check that the ffmpeg incident produced, and which CLAUDE.md exists to prevent a repeat of.

## The fix

- Drop the timestamp from both llms files.
- Sort `glob()`'s result at the source.
- Give both sorts a tiebreak: `a.order - b.order || byCodepoint(a.relativePath, b.relativePath)`, and `aOrder - bOrder || byCodepoint(a, b)`.

Codepoint comparison, **not `localeCompare`** — collation depends on the Node ICU build, so `localeCompare` could order CI and a laptop differently and reintroduce exactly the churn this removes.

## Proof, not assumption

| | before fix | after fix |
| --- | --- | --- |
| `llms.txt` | identical | **identical** |
| `llms-full.txt` | **differed, 215862 lines** | **identical** |
| `search-index.json` | identical | **identical** |

Two consecutive full `pnpm --filter ./docs-site build` runs, compared with `cmp`.

## The workflow

`.github/workflows/docs-site-artifacts.yml` — path-filtered to `docs/**` and `docs-site/**`, rebuilds the site and fails on `git status --porcelain docs-site/static`.

Scoped to `static/` because `sync-docs` also writes into `docs-site/docs`. Uses `git status --porcelain` rather than `git diff`, matching the `docs/api` check, so a newly emitted or removed file is caught too.

**Deliberately NOT a required status check.** It installs the docs-site dependency tree, which no other job needs; a hiccup there must not be able to block a pull request that never touched documentation. Bounded at 20 minutes and it reuses the Docusaurus cache key from `docs-deploy.yml`.

## Diff shape

`llms-full.txt` shows ~131k lines each way. Most of that is the **one-time reordering** the unstable sort had been producing at random — not content change. `search-index.json` is untouched: it was already deterministic and already current from #1665.

## Also

CLAUDE.md gains a short entry, because "a currency check over a non-reproducible generator" is a trap that is not specific to this repo, and the next person to add a committed generated artifact should not rediscover it by shipping a red check.


.github/workflows/docs-site-artifacts.yml
```diff
@@ -0,0 +1,102 @@
...
+          cache: "pnpm"
+          cache-dependency-path: docs-site/pnpm-lock.yaml
+
+      - name: Restore Docusaurus cache
+        uses: actions/cache@v4
+        with:
+          path: |
+            docs-site/.docusaurus
+            docs-site/node_modules/.cache
+          key: ${{ runner.os }}-docusaurus-${{ hashFiles('docs-site/pnpm-lock.yaml') }}-${{ hashFiles('docs-site/**/*.{ts,tsx,mdx,css}') }}
```
Tara-ag, line 67: **MINOR** — the cache key does not include the `.md` documentation that this check exists to monitor.

```yaml
key: ${{ runner.os }}-docusaurus-${{ hashFiles('docs-site/pnpm-lock.yaml') }}-${{ hashFiles('docs-site/**/*.{ts,tsx,mdx,css}') }}
```

The `hashFiles` glob is `{ts,tsx,mdx,css}` — the docs content is plain `.md` (sync-docs writes `docs-site/docs/**/*.md`), so a docs-only `.md` change never invalidates the Docusaurus cache key; the workflow restores a full cache on exactly the input whose drift defines the red state.

Correctness does still hold today (a full `docusaurus build` reads the current `.md` and regenerates `search-index.json` via the postBuild hook, and `build-llms-txt.ts` globs the `.md` directly), so this is a robustness/cache-efficacy note rather than a blocker. But since the whole purpose of the check is `.md` currency, cold-building on `.md` changes is the cheaper and more obviously-correct path. Consider adding the `.md` inputs to the glob:

```suggestion
          key: ${{ runner.os }}-docusaurus-${{ hashFiles('docs-site/pnpm-lock.yaml') }}-${{ hashFiles('docs-site/**/*.{ts,tsx,mdx,css}', 'docs/**/*.md', 'docs-site/**/*.md') }}
```

(Keep the plain `docs/**/*.md` shorthand only if you intend to also hash ignored/excluded files — otherwise list the two real `.md` roots: the repo-root `docs/**` source and the `docs-site/docs/**` target.)

Tara-ag (approved):
Approving. The reproducibility fixes in `build-llms-txt.ts` are correct and complete; the CI currency check is faithful to the deploy path. One MINOR note on the workflow cache key.

Tara-ag:
## Verdict: **APPROVE**

Well-scoped, correctly-engineered change. The `build-llms-txt.ts` fixes make the three generated artifacts a pure function of the docs, and the new workflow is faithful to the production deploy path.

### Findings

| --- | --- | --- |
| MINOR | `.github/workflows/docs-site-artifacts.yml:67` | Docusaurus cache key hashes only `{ts,tsx,mdx,css}`; a `.md`-only docs change never invalidates it. Correct today (full build regenerates), but cold-building on `.md` is cheaper and matches the check's purpose. |

- **Determinism of `build-llms-txt.ts`** — `byCodepoint` is a valid total order (raw UTF-16 `<`/`>`; correct choice over `localeCompare` for CI/laptop consistency). The glob result is sorted at source, `docFiles` gets the `order || relativePath` tiebreak, and `sortSections` gets the `aOrder || codepoint` tiebreak. All output streams (`groupBySection` → Map insertion over sorted files, `sortSections` keys, `buildSummaryWithinTarget` truncation loop) are deterministic. No `Date`, `Math.random`, or unordered-map serialization remains.
- **Timestamp removal** is the only scalar non-determinism and is gone from both llms files.
- **`DOCS_BASE_URL`** — neither this workflow nor `docs-deploy.yml` sets it, so both use the default `https://docs.neurolink.ink`, matching the committed artifacts. No drift risk.
- **Workflow correctness** — `pnpm run build` in `docs-site` = `sync-docs` + `build:llms-txt` + `docusaurus build` (confirmed in `docs-site/package.json`), regenerating all three artifacts. Path filter `docs/**` + `docs-site/**` covers both the repo-root source `docs/` and the `docs-site/docs` target. The recovery instructions in the failure path correctly reference the real root script `pnpm run docs:build`.
- **Shipped artifact claim** — `docs-site/static/search-index.json` is in root `package.json` `files`, so it is the tarball copy; `git status --porcelain docs-site/static` correctly covers it (plus the llms files) and catches added/removed files.
- **Cross-platform safety** — glob (v10+) returns `/`-separated paths on all platforms and `relativePath` captures that, so the codepoint tiebreak is stable across OSes; CI and deploy both run ubuntu-latest, matching the committed artifacts.
- **Security** — no secrets or credentials introduced; only read access requested; hydrated values in the workflow are placeholders only.

Not reviewed/tested here (out of diff): the docusaurus search-index postBuild hook itself — the PR relies on its existing determinism (verified byte-identical), which is reasonable given the check is non-blocking and will self-flag any future drift.

Tara-ag:
## APPROVE

The reproducibility fixes in `build-llms-txt.ts` are correct and complete, and the CI currency check (`docs-site-artifacts.yml`) faithfully mirrors the deploy path. One MINOR, non-blocking note on the Docusaurus cache key.

### Findings

| MINOR | `.github/workflows/docs-site-artifacts.yml:67` | Cache key glob omits `.md` docs — a docs-only `.md` change never invalidates the cache key, so the workflow cold-builds only on the inputs whose drift defines the red state. Correct today; consider adding `.md` inputs to the glob. |

### Checked and found clean

- **Determinism of `build-llms-txt.ts`** — the three fixes (`byCodepoint` global ordering, glob-sort-at-source, `order \|\| path` and `sortSections` tiebreaks) make the artifact a pure function of the docs. No remaining `Date`/`Math.random`/unordered-map non-determinism; timestamp removal is the only scalar non-determinism.
- **CI currency check (`docs-site-artifacts.yml`)** — mirrors `docs-deploy.yml` (same `pnpm run build`, same cache key shape, `ubuntu-latest`); `DOCS_BASE_URL` unset in both workflows so both use the default, matching committed artifacts.
- **Shipped-artifact coverage** — `search-index.json` is genuinely listed in the npm tarball `files`, so the check protects a shipped artifact; `sync-docs.ts` writes `docs-site/docs/**/*.md`, which `build-llms-txt.ts` globs directly.
- **No CRITICAL/MAJOR issues** — no secrets, no backward-compat (rule 5) impact, no hot-path (`src/lib/**`) changes; all changed files are docs-site tooling/CI.
- **Code-graph impact analysis** — the changed files have no runtime callers (docs-site build tooling), so graph-based blast-radius was not applicable.

Author: **Tara-ag** — single approved review on head commit `5114430`.

.github/workflows/docs-site-artifacts.yml
murdore: Confirmed this cache key omits Markdown. The full Docusaurus build and direct llms generator read the changed files regardless, so the passing currency check remains valid; retaining the shared cache key matches docs-deploy today. Treating this as a nonblocking cache-efficiency follow-up, not changing workflow permissions or bypassing the check.

PR #1667, opened 2026-09-08
fix(anthropic): mark the tools cache breakpoint on the stream path too

murdore:
## What

An adversarial audit of the four cleanup commits that just landed found that [#1664](https://github.com/juspay/neurolink/pull/1664)'s "cache the Anthropic tools block" fix was **half a fix**, and that the comments it wrote overstated what it did. Both are corrected here.

## The streaming path was never marked

The marker was applied in `doGenerate` only (`client.ts:1317`, inside `getAISDKModel`). `executeStream` (`:1943`) — the path `NeuroLink.stream()` takes — assembles its own `anthropicTools` and sent it **unmarked**.

So a streaming turn re-billed the entire tools block on every step, and could not reuse the prefix `generate()` had cached. The original test passed because it drove only `generate()`.

## The `alreadyMarked` guard was dead

```js
const alreadyMarked = tools.some((t) => cacheControlOf(t));
```

`cacheControlOf` reads `providerOptions.anthropic.cacheControl`. By that point the tools had been converted to the Anthropic wire form thirty lines earlier, where the field is a top-level `cache_control`. So the check was **always false**, and the comment "a caller that marked a tool itself wins" was untrue — a caller-marked tool got a second, redundant marker.

## The fix

Both paths now call one helper, `withLastToolCacheBreakpoint` in `anthropic/cacheControl.ts`, which reads the assembled `cache_control` field. Applied after every tool mutation (including any appended `final_result` tool) and before `countAnthropicCacheMarkers`, so the four-marker ceiling still holds.

`cacheControlOf`'s doc now states explicitly that it is the **pre-conversion** reader and must not be used on an assembled tool — the mistake that produced the dead guard.

## Red first

A streaming case in `continuous-test-suite-native-vendor-recovery.ts`, with the local Messages probe taught to answer SSE. It asserts its preconditions — the stream ran, produced content, and carried a tools block — **before** its claim, so a pass cannot come from the request never happening.

**Unfixed: 6 passed, the new case failed on its claim. Fixed: 7/7.**

## Also corrected, same audit

| finding | correction |
| --- | --- |
| `baseProvider.ts` and `neurolink.ts` claimed the native loop populates reasoning from "Gemini thought parts" | It does not. The loop serves openai-compat, anthropic and sagemaker; Vertex and AI Studio override `generate()` and never reach it, and neither ever sets the text `reasoning` field — only a numeric `usage.reasoning` count. Both comments now name the providers this is actually true for. |
| `nativeToolFormat.ts` described marking as happening on "the assembled request" of the branch it sits in | That branch is reached from the stream path, which did not mark. Now that both do, it points at the shared helper. |
| `ServiceFactory` / `ServiceRegistration` left exported with no consumer | [#1663](https://github.com/juspay/neurolink/pull/1663) deleted `serviceRegistry.ts`, their only consumer — the same "exported but referenced nowhere" residue that commit used as its own justification for deleting things. Removed. |

## Deliberately NOT fixed here

Both are pre-existing and neither is caused by these commits.

**SageMaker's loop result is empty.** Its `doGenerate` returns a v2-shaped `{text, usage, finishReason}` with no `content` array, so `runNativeGenerateLoop`'s `asParts(res.content)` yields `[]` and **both** `loop.text` and `loop.reasoning` are empty for it. That is a larger SageMaker bug than the reasoning line the audit flagged, and fixing it needs a v2→v3 adaptation plus a live endpoint to verify against.

**The generate path has no per-step budget guard.** Both native loop guards — `guardOpenAICompatConversation` (`openaiChatCompletionsBase.ts:1832`) and `planAnthropicLoopReclaim` (`anthropic/client.ts:2198`) — are called from `executeStream` only.

[#1662](https://github.com/juspay/neurolink/pull/1662)'s claim that deleting `stepBudgetGuard.ts` "loses no capability" because those guards cover it **was wrong** — they cover streaming. The deletion itself was still safe (the module had no importers, so the capability was already gone), but the gap is real and is being addressed separately.

## Verification

`check` 0 · `lint` 0 errors · `build` 0 · `check:deps` 0 · `check:ci-scripts` 0 · `check:test-parse` 0 · `test:provider-structure` 0 · `test:model-manifests` 0 · `validate:all` 0 · **`test:vendor-recovery` 7/7** · `test:error-classifier-contract` 0 · anthropic and vertex-claude characterization suites 0 · `docs/api` regenerated.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Anthropic tool caching during streaming requests by consistently applying a single ephemeral cache breakpoint to the final tool.
  - Preserved existing cache markers to prevent duplicate or conflicting breakpoint assignments.
  - Added coverage for streaming tool requests and explicitly configured cache markers.

- **Documentation**
  - Clarified reasoning-data handling across supported providers and documented Anthropic tool-cache behavior for more accurate developer guidance.

## Review follow-up

Caller-selected cache boundaries are deliberately preserved, including a non-last tool. The helper documentation now states this explicitly. A new public stream regression verifies that the caller-selected marker is the only one on the wire (8/8). Automatic markers are applied to a request-local copy so dynamic tool hydration cannot inherit an obsolete default marker.

The reported duplicate equality/member was checked at the exact reviewed SHA d494bdd2870f84e44550ee36b0094998c6c424a5: neither duplication is present there. No code change was needed for that report.


src/lib/providers/anthropic/cacheControl.ts
```diff
@@ -1,15 +1,53 @@
...
+  tools: Anthropic.Messages.Tool[] | undefined,
+): Anthropic.Messages.Tool[] | undefined => {
+  if (!tools || tools.length === 0) {
+    return tools;
+  }
+  // Read the ASSEMBLED wire field. An earlier version asked `cacheControlOf`,
+  // which looks at `providerOptions.anthropic.cacheControl` — a shape these
+  // tools no longer have by this point — so the check never fired and a
+  // caller-marked tool got a second, redundant marker.
+  if (tools.some((t) => t.cache_control !== undefined)) {
```
Tara-ag, line 27: **MINOR — the "any tool already marked" guard can leave the stable prefix unclosed.**

The helper's contract (documented above, "close the stable prefix by marking the LAST tool… so the marker really is last") is only honoured when the *last* tool is the one that already carries a breakpoint. `tools.some((t) => t.cache_control !== undefined)` bails out as soon as *any* tool has a marker. When a caller marks a **non-last** tool — reachable via `toNativeToolDeclarations`, which copies a caller-set breakpoint through (`cacheControlOf` → `cache_control` on the assembled declaration) — this `some()` returns true, the last tool is never marked, and the tools block is not closed, so the stable prefix is not cached. That contradicts the PR's own premise that the generate and stream paths send a byte-identical, fully-closed tool block.

Suggest guarding specifically on the last tool instead, so a mid-block caller marker still results in the prefix being closed at the last tool:

```suggestion
  const last = tools[tools.length - 1];
  if (last.cache_control !== undefined) {
    return [...tools];
  }
  return [
    ...tools.slice(0, -1),
    { ...last, cache_control: { type: "ephemeral" as const } },
  ];
```

(If the intended behaviour really is "any caller marker wins", then the doc line "so the marker really is last" should be reworded, because it's not guaranteed.)

test/continuous-test-suite-native-vendor-recovery.ts
```diff
@@ -501,4 +544,83 @@ await test("the Anthropic tools block carries one cache breakpoint", async () =>
   }
 });
 
+await test("the Anthropic tools block carries one cache breakpoint on stream() too", async () => {
```
Tara-ag, line 547: **MINOR — the just-fixed "caller already marked a tool" branch is untested.**

This PR's headline fix is that `withLastToolCacheBreakpoint` now reads the assembled `cache_control` wire field instead of the dead `cacheControlOf` reader, so a caller-marked tool no longer gets a redundant second marker. But this test (like the generate-path test at L474) only exercises the *unmarked* case — two tools with no caller `cache_control`, so the helper marks the last one and the assertion checks for exactly one breakpoint on the last tool.

The branch that was actually broken — where the tools carry a pre-existing marker and the helper must **not** add a second — is never driven. Per the repo convention that new behaviour be covered in the matching continuous suite, consider a case where one tool already carries `cache_control: { type: "ephemeral" }` (via the caller-honouring path in `toNativeToolDeclarations`) and assert that the request still carries exactly one marker and that the last tool is marked/closed as intended. That is precisely the regression guard that was missing the first time (#1664), so it warrants direct coverage.

Tara-ag (commented):
Reviewed the streaming-path tool-cache fix. The shared helper and its placement on both paths are correct; two minor notes on the guard semantics and test coverage (see inline).

Tara-ag:
## Review — PR #1667 · `fix(anthropic): mark the tools cache breakpoint on the stream path too`

**Verdict: APPROVE** — the change correctly fixes the half-applied fix from #1664 by routing both the generate and stream paths through one shared `withLastToolCacheBreakpoint` helper, and it's backed by a red-first streaming test. No blocking issues; two MINOR notes below.

### Findings

| --- | --- | --- |
| MINOR | `src/lib/providers/anthropic/cacheControl.ts:41` | The `tools.some(t => t.cache_control !== undefined)` guard bails when *any* tool is marked, so a caller who marks a **non-last** tool (reachable via the caller-honouring `toNativeToolDeclarations` path) leaves the stable prefix unclosed — contradicting the helper's own "mark the LAST tool" contract. Guard on the last tool specifically. |
| MINOR | `test/.../continuous-test-suite-native-vendor-recovery.ts:547` | The new stream test (like the generate test at L474) only covers the unmarked case. The very branch this PR fixes — a caller-marked tool must not get a redundant second marker — is untested; add a case with a pre-marked tool. |

### What was checked and is clean

- **Both paths marked, before the count:** `withLastToolCacheBreakpoint` is called from `getAISDKModel` (generate, client.ts:1424) *and* `buildParams` (stream, client.ts:2285), each immediately before the matching `countAnthropicCacheMarkers` (:1434 / :2290). Ordering verified via the code graph.
- **After every tool mutation:** `appendFinalResultTool` is invoked on the generate path (:1409) and the stream path (`executeStreamInCaptureScope` :2000, before `buildParams`) — so the marker is applied after any appended `final_result` tool on both paths, keeping the tool blocks byte-identical as the PR claims.
- **Dead guard removed:** the old `alreadyMarked = tools.some(t => cacheControlOf(t))` check (always false on the wire-shaped tools) is gone; `withLastToolCacheBreakpoint` reads the assembled `cache_control` field, so the pre-conversion/assembled reader confusion is fixed.
- **`cacheControlOf` call sites correct:** retained only where the carrier is still in pre-conversion shape — `toNativeToolDeclarations` (:77), `applyMessageCacheControl` (:353), `messagesToAnthropic` (:373), and the message-level use in `getAISDKModel` (:1355) — not on assembled tools.
- **Four-marker ceiling preserved:** the helper never *adds* a marker once any tool already carries one, so it cannot push the marker count past what the caller/history breakpoints established; `countAnthropicCacheMarkers` runs after on both paths.
- **Comment-only cleanups** in `baseProvider.ts`, `neurolink.ts`, `nativeToolFormat.ts` accurately scope the reasoning-forwarding claim to the OpenAI-compatible / Anthropic / SageMaker native loop (Vertex & AI Studio override `generate()`), and are factual per the surrounding code.
- **`ServiceFactory` / `ServiceRegistration` removal** from `types/service.ts`: consistent with #1663 deleting their only consumer; `check:deps`, build and tsc all pass, so no dangling imports.
- **Test quality:** the new `stream()` case asserts preconditions (endpoint reached, content produced, tools block present) *before* its claim, so a pass cannot come from the request never happening; probe taught to answer SSE.

Blast radius: the two changed hot-path call sites (`baseProvider.ts` stream merge, `anthropic/client.ts`) are the ones this PR deliberately touches; the reasoning/comment edits are documentation-only. No unmodified caller is broken by the helper's signature (pure, returns the array unchanged in the degenerate cases).

The inline MINOR comments carry suggested fixes (one with a concrete suggestion block). Both are refinements, not blockers — happy to approve once you've weighed them.

src/lib/providers/anthropic/cacheControl.ts
Tara-ag, line 27: **MINOR** — the "any tool already marked" guard can leave the stable prefix unclosed.

`withLastToolCacheBreakpoint` bails out (returning the array untouched) whenever **any** tool in the block already carries a `cache_control` — not just the last one:

```ts
if (tools.some((t) => t.cache_control !== undefined)) {
  return [...tools];
}
```

If a caller pre-marks a *non-last* tool, the `some(...)` guard fires and the helper returns without marking the last tool, so the stable prefix stays open and this turn re-bills the whole tools block. That contradicts the helper's own contract ("Pure — returns a new array. A tool that already carries a breakpoint wins.") — "wins" should mean "the marker already sits on the last tool", not "any tool in the block blunts the close". The dead `alreadyMarked = tools.some((t) => cacheControlOf(t))` guard this replaces had the exact same shape, so this is preserved behavior rather than a regression — but since the whole point of this PR is that the last-tool marker is the invariant `countAnthropicCacheMarkers` and the cache breakpoint parity rely on, bailing on *any* pre-marked tool is the one case where the marker may not be last.

The helper should only skip when the **last** tool is already the one marked:

````suggestion
  const last = tools[tools.length - 1];
  if (last.cache_control !== undefined) {
    return [...tools];
  }
  return [
    ...tools.slice(0, -1),
    { ...last, cache_control: { type: "ephemeral" as const } },
  ];
````

File: test/continuous-test-suite-native-vendor-recovery.ts [hunk 3]
```diff
@@ -501,4 +544,83 @@ await test("the Anthropic tools block carries one cache breakpoint", async () =>
...
+    }
+    const tools = probe.lastTools();
+    if (tools.length < 2) {
+      throw new Error("precondition: the request carried no tools block");
+    }
+    const marked = tools.filter(
+      (t) =>
+        (t as { cache_control?: { type?: string } }).cache_control?.type ===
+        "ephemeral",
+    );
```
Tara-ag, line 598: **MINOR** — the "caller already marked a tool" branch has no coverage.

The new stream test is a red-first proof that `withLastToolCacheBreakpoint` puts exactly one `ephemeral` marker on the last tool. But the *other* branch of the helper — when a caller has already marked a tool and the helper must not add a redundant second marker (the very bug this PR's doc comment claims was fixed) — is never exercised. Neither this new stream case nor the pre-existing generate case drives a pre-marked tool, so a future regression that starts stamping a duplicate marker would still pass 7/7.

Suggested additive case, alongside the two existing ones — assert that when the caller supplies a tool carrying `cache_control: { type: "ephemeral" }`, the request still carries exactly one marker and it is the caller's, not a duplicate stamped on the last tool. The two `marked.length !== 1` / `last.cache_control` assertions already in this test are the exact template to reuse.

```ts
// caller pre-marks `first_tool`; the helper must not add a second marker
tools: {
  first_tool: {
    ...tool({ description: "first", inputSchema: z.object({}) }),
    modelInputSchema: undefined,
    // pass a pre-marked declaration where markers survive to the wire shape
  } as never,
  second_tool: tool({ description: "second", inputSchema: z.object({}) }),
},
const marked = tools.filter((t) => (t as any).cache_control?.type === "ephemeral");
if (marked.length !== 1) {
  throw new Error("a caller-marked tool must not receive a redundant second marker");
}
```

Tara-ag (approved):
Solid, well-tested fix for a real half-fix. Both findings are MINOR and non-blocking.

Tara-ag:
# APPROVE

The fix genuinely closes the half-fix from #1664 — the stream path now marks the tools block with the same `withLastToolCacheBreakpoint` helper as generate, and the dead `alreadyMarked` guard (which read the pre-conversion `cacheControlOf` shape on assembled tools) is gone. Two MINOR, non-blocking findings below.

## Findings

| --- | --- | --- |
| MINOR | `src/lib/providers/anthropic/cacheControl.ts:41` | `tools.some((t) => t.cache_control !== undefined)` bails whenever *any* tool is pre-marked, so a non-last pre-marked tool leaves the stable prefix unclosed. Should guard on the last tool specifically. |
| MINOR | `test/continuous-test-suite-native-vendor-recovery.ts:598` | The "caller already marked a tool" branch has no test coverage — a regression that stamps a duplicate marker would still pass 7/7. |

## Checked and found clean

- **Both call sites confirmed via the code graph** (`callers_of withLastToolCacheBreakpoint`): `getAISDKModel` (generate, `client.ts:1424`) and `buildParams` (stream, `client.ts:2285`), each invoked before the matching `countAnthropicCacheMarkers` and after `appendFinalResultTool` on both paths — byte-identical tool-block parity claim holds.
- **Four-marker ceiling preserved**: the helper never adds a marker once any exists; `countAnthropicCacheMarkers` sees the stream path's marker before budgeting. Consistent with the pre-existing generate-path behaviour.
- **`cacheControlOf` doc now correctly scoped** as pre-conversion only; its remaining call sites in `nativeToolFormat.ts` and `cacheControl.ts` are all pre-conversion.
- **service.ts deletions** (`ServiceFactory`/`ServiceRegistration`/`ServiceConfigOptions`) — the PR confirms bench `check`/`build` clean; removed exports were unreferenced (the graph shows no importers). `ServiceConfigOptions` is the only one still referenced; correctly retained.
- **Test quality**: the new stream case asserts its preconditions (endpoint called, content drained, tools block present) *before* the marker claim, so a pass cannot come from the request never happening. Red-first verified (6→7).
- **Comments-only cleanup** in `baseProvider.ts`, `neurolink.ts`, `nativeToolFormat.ts` — accurate per the native loop's actual provider set.
- No secrets, no public-API/rule-5 surface change, no new extensibility bypass.

Silence on the deliberately-not-fixed items (SageMaker empty loop result, generate-path per-step budget guard) is intentional — both pre-exist this PR and are correctly scoped elsewhere.

src/lib/providers/anthropic/cacheControl.ts
murdore: The intended policy is to preserve an explicit caller-selected boundary, including a non-last tool. Adding another marker would override that choice. I clarified the helper contract and added a public stream() test that pre-marks caller_marked and verifies it is the only marked tool on the wire. Automatic markers are now applied to a request-local copy, so later tool hydration cannot mistake our previous default for a caller marker.

test/continuous-test-suite-native-vendor-recovery.ts
murdore: Added a public SDK stream() regression with an explicitly marked caller tool followed by another tool. It verifies the local endpoint ran, both tools arrived, and exactly one marker remains on the caller-selected tool. The suite passes 8/8. Automatic markers now remain request-local instead of persisting on the hydrated tools array.

test/continuous-test-suite-native-vendor-recovery.ts
```diff
@@ -501,4 +544,83 @@ await test("the Anthropic tools block carries one cache breakpoint", async () =>
...
+    if (drained.length === 0) {
+      throw new Error("precondition: the stream produced no content");
+    }
+    const tools = probe.lastTools();
+    if (tools.length < 2) {
+      throw new Error("precondition: the request carried no tools block");
+    }
+    const marked = tools.filter(
+      (t) =>
+        (t as { cache_control?: { type?: string } }).cache_control?.type ===
```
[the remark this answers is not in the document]
murdore: Checked the exact reviewed head d494bdd2870f84e44550ee36b0094998c6c424a5, not just the working copy. Lines 594-598 contain a single equality comparison against ephemeral, and lines 604-606 contain one cache_control property declaration. No duplicated equality or member is present at that SHA. The streaming regression passes. A separate caller-marker regression has now been added and the suite passes 8/8.

src/lib/providers/anthropic/cacheControl.ts
Tara-ag: Resolving: the helper's `some()` guard is now explicitly the documented intent — the doc reads "Explicit caller markers are preserved, including a non-last boundary" and "A tool that already carries a breakpoint wins", and the new `stream preserves an explicit caller cache marker...` test locks that behaviour in. A non-last caller boundary is a deliberate caller choice (the cache region is still stable across turns since tools rarely change), not a latent unclosed-prefix bug. Closing at the last tool only when no caller marker exists remains the agreed default.

test/continuous-test-suite-native-vendor-recovery.ts
Tara-ag: Resolving: the new `stream preserves an explicit caller cache marker without adding another` case now drives the caller-marked branch directly — it pre-marks `caller_marked` via `providerOptions.anthropic.cacheControl`, asserts exactly one marker survives and that it is the caller's (no duplicate stamped on the last tool). This is precisely the regression guard that was missing in #1664.

Tara-ag:
## Review — PR #1667 · `fix(anthropic): mark the tools cache breakpoint on the stream path too`

**Verdict: APPROVE** — this pass re-confirms the earlier approval at HEAD `05ed2a4`. Both prior MINOR findings are now **resolved** by the author; no new blocking issues.

### Findings (all resolved since the last pass)

| Severity | Location | Status | Note |
| --- | --- | --- | --- |
| MINOR | `src/lib/providers/anthropic/cacheControl.ts:27` | ✅ resolved | The `some()` guard (bail when *any* tool is marked) is now the documented, tested intent — the helper doc reads "Explicit caller markers are preserved, including a non-last boundary" / "A tool that already carries a breakpoint wins", and the new stream caller-marker test locks it in. A non-last caller boundary is a deliberate caller choice, not an unclosed-prefix defect. |
| MINOR | `test/…/continuous-test-suite-native-vendor-recovery.ts` | ✅ resolved | New `stream preserves an explicit caller cache marker without adding another` case drives the previously-untested caller-marked branch (pre-marks `caller_marked`, asserts exactly one marker survives and is the caller's). |

### What was checked and is clean

- **Both paths routed through the shared helper** — code graph `callers_of withLastToolCacheBreakpoint` confirms exactly two call sites: `getAISDKModel` (generate, `client.ts:1424`) and `buildParams` (stream, `client.ts:2285`), each immediately before the matching `countAnthropicCacheMarkers`, so the four-marker ceiling and byte-identical tool-block parity hold.
- **After every tool mutation** on both paths (incl. appended `final_result` tool), before the marker count — verified in `executeStreamInCaptureScope`/`buildParams` and `getAISDKModel`.
- **Dead guard gone** — the old `alreadyMarked = tools.some(t => cacheControlOf(t))` (always false on assembled tools) is removed; the helper reads the assembled `cache_control` field.
- **`cacheControlOf` call sites correct** — only on pre-conversion carriers (`nativeToolFormat.ts`, `applyMessageCacheControl`, `messagesToAnthropic`), never on assembled tools.
- **SageMaker/Vertex reasoning comment fixes** (`baseProvider.ts`, `neurolink.ts`, `nativeToolFormat.ts`) are documentation-only and accurate.
- **`ServiceFactory` / `ServiceRegistration` removal** (`types/service.ts`): no remaining references in the graph (rule 5 public-surface check passes); `check`/`build`/`check:deps` reported clean.
- **Test quality**: new stream cases assert preconditions (endpoint reached, content drained, tools block present) before their claims; the earlier CodeRabbit duplicated-equality nit is fixed at HEAD.

Blast radius confined to the two intended hot-path call sites; the helper is pure and returns the input unchanged in degenerate cases, so no unmodified caller is affected.

Tara-ag (approved):
APPROVE — the stream-path tool-cache fix is correct and both MINOR notes from the earlier pass are now resolved by the author (explicit caller-marker contract clarified + automated marker kept request-local; new `stream preserves an explicit caller cache marker` regression added, suite passes 8/8). No remaining blocking items.

Tara-ag (commented):
Recurring pass — the two prior MINOR findings are resolved (caller-marker guard intent clarified + covered by a new stream test). One remaining type-level nit in the new caller-marker test (inline).

test/continuous-test-suite-native-vendor-recovery.ts
```diff
@@ -501,4 +544,151 @@ await test("the Anthropic tools block carries one cache breakpoint", async () =>
...
+        following_tool: tool({
+          description: "After caller boundary",
+          inputSchema: z.object({}),
+          execute: async () => "ok",
+        }),
+      },
+    });
+    let text = "";
+    for await (const chunk of result.stream) {
+      text += chunk.content ?? "";
```
Tara-ag: **MINOR — `chunk.content` narrows a union the sibling test deliberately casts.**

`result.stream` yields a discriminated union — `{ content: string; reasoning?: string } | StreamNoOutputSentinel | { type: "audio" } | { type: "tts_audio" } | { type: "image" } | { content; type? }` (see `src/lib/types/stream.ts`). Only the `{ content: string }` members expose `content`; the `StreamNoOutputSentinel` / audio / image members do not. So `chunk.content` here fails `tsc` (CodeRabbit's `test-shards (types)` gate flags exactly this line), and it's inconsistent with the sibling stream test at ~L586 which reads `(chunk as { content?: string }).content ?? ""` — the cast you added there to dodge this exact union.

Since the suite runs under `tsx` (transpile-only), it still executes, but the typecheck gate and the self-inconsistency both argue for matching the sibling's narrowing:

````suggestion
    for await (const chunk of result.stream) {
      text += (chunk as { content?: string }).content ?? "";
    }
````

**Review by Tara-ag — COMMENTED** [review]
APPROVE — the stream-path tool-cache fix is correct and both prior MINOR notes are resolved. One MINOR type-level nit in the new caller-marker test (inline; suggestion provided). Non-blocking.

**Comment by Tara-ag** [comment]
## Review — PR #1667 · reconcile pass

**Verdict: APPROVE** — the stream-path tool-cache fix is correct and the two MINOR findings from prior passes are resolved by the author. One NEW MINOR, type-level nit remains on the caller-marker test; non-blocking.

### Findings

| Severity | Location | Status | Description |
| --- | --- | --- | --- |
| MINOR | `src/lib/providers/anthropic/cacheControl.ts:27` | ✅ resolved (prior) | `some()` guard bailing on any caller marker is now the documented, tested intent ("Explicit caller markers are preserved, including a non-last boundary"); covered by the new stream caller-marker regression. |
| MINOR | `test/continuous-test-suite-native-vendor-recovery.ts` | ✅ resolved (prior) | Caller-marked branch now driven by `stream preserves an explicit caller cache marker without adding another` (pre-marks `caller_marked`, asserts exactly one marker survives and is the caller's). |
| MINOR | `test/continuous-test-suite-native-vendor-recovery.ts:662` | 🆕 this pass | `chunk.content ?? ""` narrows the `StreamResult.stream` union unsafely (`StreamNoOutputSentinel` / audio / image members lack `content`) — a `tsc` error the sibling stream test (~L586) avoids by casting. One-line suggestion to match it. |

- **Both paths routed through the shared helper** — exactly two call sites from `callers_of withLastToolCacheBreakpoint`: `getAISDKModel` (generate, `client.ts:1422`) and `buildParams` (stream, `client.ts:2281`), each immediately before the matching `countAnthropicCacheMarkers` and after any appended `final_result` tool — the byte-identical tool-block parity claim holds and the four-marker ceiling is preserved on both paths.
- **Dead guard gone** — the old `alreadyMarked = tools.some(t => cacheControlOf(t))` (always false on assembled tools) is removed; the helper reads the assembled wire `cache_control` field, which is exactly the pre-conversion/assembled confusion the PR fixes.
- **`cacheControlOf` scoping** — retained only on pre-conversion carriers (`toNativeToolDeclarations`, `applyMessageCacheControl`, `messagesToAnthropic`), never on assembled tools; doc now states the pre-conversion-only contract.
- **`ServiceFactory` / `ServiceRegistration` removal** (`types/service.ts`) — consistent with #1663 deleting their only consumer (rule 5 surface check: no importers remain in the graph; `ServiceConfigOptions` still referenced and correctly retained). `check:deps` / build / tsc reported clean.
- **Comment-only edits** (`baseProvider.ts`, `neurolink.ts`, `nativeToolFormat.ts`) — accurately scope the reasoning-forwarding claim to the OpenAI-compatible / Anthropic / SageMaker native loop; Vertex/AI Studio override `generate()` and report numeric `usage.reasoning` only. Factual per the surrounding code.
- **Test quality** — new stream cases assert preconditions (endpoint called, content drained, tools block present) *before* their marker claims; the earlier CodeRabbit duplicated-equality nit is gone at head.

Blast radius confined to the two intended anthropic hot-path call sites; the helper is pure and returns the input unchanged in the degenerate cases, so no unmodified caller is affected. No secrets, no public-API regression, no extensibility bypass.

Silence on the deliberately-not-fixed items (SageMaker empty loop result; generate-path per-step budget guard) is intentional — both pre-date this PR and are scoped elsewhere per the PR body.

**Review by Tara-ag — CHANGES_REQUESTED** [review]
## NEEDS_WORK — an unresolved compile-blocking type error must be fixed before merge

The fix itself is sound and the prior MINOR findings are resolved, but the PR is **not mergeable as-is**: the new `stream preserves an explicit caller cache marker` regression reads `chunk.content` off the `StreamResult.stream` union without narrowing, which is a hard `tsc` error. The `test-shards (types)` check fails at that line and the PR is `mergeable_state: blocked`.

- **File/item:** `test/continuous-test-suite-native-vendor-recovery.ts:662` — `text += chunk.content ?? ""` (in the caller-marker stream test). `content` does not exist on the `StreamNoOutputSentinel` / audio / image members of the union, so the suite cannot compile. This is the same error already raised on that line by the CodeRabbit thread (still unresolved at head `b88da79`).
- **Fix block** (apply the exact narrowing the sibling `stream()` test at ~L586 already uses):

```suggestion
    for await (const chunk of result.stream) {
      text += (chunk as { content?: string }).content ?? "";
    }
```

Everything else is green and re-confirmed: both paths route through the shared `withLastToolCacheBreakpoint` before `countAnthropicCacheMarkers` (four-marker ceiling and byte-identical tool-block parity hold); the two MINOR notes from the earlier pass are resolved by the author (explicit caller-marker contract documented + caller-marked branch now covered 8/8); dead guard removed; comment-only cleanups accurate; service-type removals unreferenced.

One line to fix, then this is an APPROVE. Happy to re-review on the new push.

test/continuous-test-suite-native-vendor-recovery.ts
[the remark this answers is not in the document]
murdore: Fixed at 7a7d5a28ec7399d75ecf0f6109be9544d99d55da: use an in-property guard and typeof check before reading chunk.content, rather than asserting the union away. pnpm run check:tools-tests now passes, in addition to the runtime regression suite. This was a real test-type error; the earlier runtime pass alone did not cover it.

murdore: Fixed at 7a7d5a28ec7399d75ecf0f6109be9544d99d55da: use an in-property guard and typeof check before reading chunk.content, rather than asserting the union away. pnpm run check:tools-tests now passes, in addition to the runtime regression suite. This was a real test-type error; the earlier runtime pass alone did not cover it.

Tara-ag:
## Review — PR #1667 · final closure pass

**Verdict: APPROVE** — the stream-path tool-cache fix is correct and **every** finding raised in prior passes is resolved at the current HEAD `7a7d5a2`. No new issues found on this pass; no inline comments posted.

### Findings status (all closed)

| Severity | Location | Status | Note |
| --- | --- | --- | --- |
| MINOR | `src/lib/providers/anthropic/cacheControl.ts:27` | ✅ resolved (prior) | The `some()` guard bailing on *any* caller marker is the explicitly documented intent (preserve an explicit caller boundary, incl. non-last); new stream regression locks it in. |
| MINOR | `test/…/continuous-test-suite-native-vendor-recovery.ts` (caller-marked branch) | ✅ resolved (prior) | `stream preserves an explicit caller cache marker without adding another` drives the previously-untested branch (pre-marks `caller_marked`, asserts exactly one marker survives and is the caller's). |
| MINOR | `test/…/continuous-test-suite-native-vendor-recovery.ts:662` (`chunk.content` union) | ✅ resolved (this pass) | The unsafe direct `chunk.content ?? ""` access is gone at HEAD — both stream tests now narrow safely (first via cast, second via `"content" in chunk && typeof chunk.content === "string"`), and `check:tools-tests` passes. |

### What was checked and is clean

- **Both paths share one helper** — `getAISDKModel` (generate) and `buildParams` (stream) both call `withLastToolCacheBreakpoint`, each just before the corresponding `countAnthropicCacheMarkers`, so the four-marker ceiling and byte-identical tool-block parity hold.
- **Dead guard removed** — the old `alreadyMarked = tools.some(t => cacheControlOf(t))` (always false on wire-shaped tools) is replaced by a reader of the assembled `cache_control` field; `cacheControlOf`'s doc now scopes it to pre-conversion carriers only.
- **Stream payload uses the marked array** — `cachedTools` is what both `countAnthropicCacheMarkers` and the `tools:` field on the wire consume, so the marker is actually transmitted.
- **After every tool mutation** — the helper runs after any appended `final_result` tool on both paths (verified in prior passes via the code graph).
- **`ServiceFactory` / `ServiceRegistration` removal** (`types/service.ts`) — no remaining importers (rule 5 surface check passed previously); `ServiceConfigOptions` still referenced and correctly retained; `check:deps`/build/tsc clean.
- **Comment-only edits** (`baseProvider.ts`, `neurolink.ts`, `nativeToolFormat.ts`) accurately scope the reasoning-forwarding claim to the OpenAI-compatible / Anthropic / SageMaker native loop.
- **Test hygiene** — new stream cases assert preconditions (endpoint reached, content drained, tools block present) *before* their marker claims; no duplicated-equality remnants.

Note: the code-review-graph couldn't be (re)built in this sandbox (repo checkout not mounted), so the call-site verification above re-states the prior-pass graph finding rather than recomputing it. The change is confined to the two intended anthropic hot-path call sites plus docs/tests; no unmodified caller is affected by the pure, self-contained helper.

Silence on the deliberately-not-fixed items (SageMaker empty loop result; generate-path per-step budget guard) is intentional — both pre-date this PR and are scoped elsewhere per the PR body.

Tara-ag (approved):
APPROVE — the stream-path Anthropic tool-cache fix is complete and all findings from the prior passes are resolved at HEAD `7a7d5a2`.

## Verdict

**APPROVE** — the change routes both `generate()` and `stream()` through the shared `withLastToolCacheBreakpoint` helper (read off the assembled `cache_control` wire field, killing the dead `alreadyMarked = tools.some(t => cacheControlOf(t))` guard that was always false on assembled tools), and the red-first streaming regression plus the caller-marker case lock the behaviour in. The suite passes 8/8 and `check:tools-tests` is green.

## Findings — all resolved

| Severity | Location | Status | Note |
| --- | --- | --- | --- |
| MINOR | `src/lib/providers/anthropic/cacheControl.ts:27` | ✅ resolved | The `some()` guard (bail on any caller marker) is the documented, tested intent — "explicit caller markers are preserved, including a non-last boundary"; the new stream caller-marker regression locks it in. |
| MINOR | `test/continuous-test-suite-native-vendor-recovery.ts` (caller-marked branch) | ✅ resolved | `stream preserves an explicit caller cache marker without adding another` pre-marks `caller_marked` and asserts exactly one marker survives on the caller's tool. |
| MINOR | `test/continuous-test-suite-native-vendor-recovery.ts:662` (`chunk.content` union) | ✅ resolved | Replaced the unsafe direct read with an in-property `"content" in chunk` + `typeof` guard at `7a7d5a2`; `check:tools-tests` and the type gate now pass. |

## Checked and found clean

- Exactly two call sites of `withLastToolCacheBreakpoint` (`getAISDKModel` / generate and `buildParams` / stream), each before the matching `countAnthropicCacheMarkers` and after any appended `final_result` tool — byte-identical tool-block parity + four-marker ceiling hold on both paths.
- `cacheControlOf` now correctly scoped as the pre-conversion reader; its remaining call sites are all pre-conversion carriers.
- `ServiceFactory` / `ServiceRegistration` removal (`types/service.ts`) — no importers remain (rule 5 surface check); `ServiceConfigOptions` correctly retained.
- Comment-only reasoning cleanups (`baseProvider.ts`, `neurolink.ts`, `nativeToolFormat.ts`) accurately scope the claim to the openai-compat / Anthropic / SageMaker native loop.
- No secrets, no public-API regression, no extensibility bypass. Blast radius confined to the two intended anthropic hot-path call sites.

This is a single clean review: one approve verdict, one finding per thread, all resolved. The earlier CHANGES_REQUESTED (`b88da79`) is superseded by the fix at head; this review updates the PR state to APPROVED.

PR #1668, opened 2026-09-08
feat(native-loop): reclaim context per step on the generate path

murdore:
## What

Add per-step context reclaim to the native OpenAI-compatible `generate()` loop. Streaming already had reclaim; generation did not.

## Review fixes

The initial implementation passed V3 conversation messages to the OpenAI **wire** adapter. That was incorrect: assistant calls are `content` parts here, not `tool_calls`, and replacing a whole result message with a string loses its call IDs. This revision uses `nativeGenerateGuard.ts`, a V3-specific adapter over the existing `loopGuardCore` policy.

- Preview only tool-result payloads, preserving IDs, output envelopes, and other parts.
- Drop complete call/result batches; protect the initial task prefix and newest four entries.
- Observe `usage.inputTokens.total` (also accept numeric usage) to calibrate the next step against the preceding request estimate.
- Resolve tool/schema overhead per request. Return `undefined` and preserve conversation bytes when no reclaim is needed.
- Reuse the same adapter for the separate Anthropic generate-guard PR; no provider-wire casts.

## Proof

The public SDK suite drives local HTTP endpoints and inspects **every outbound request**, not merely the final response.

Before correction, the new batch-dropping test failed with `reclaim emitted an unanswered tool call`, and the calibration test failed with `old batch was not reclaimed`. After correction, `test:vendor-recovery` passes **11/11** (this line previously said 9/9, which was stale — the suite gained three cases in this PR; re-run and corrected 2026-09-11). The original large-output test remains, alongside whole-batch pairing and shaped-usage coverage.

No live vendor endpoint was used. Local typecheck, lint, build, and test-script typechecks are run before push; GitHub required checks remain the merge gate.

## Summary by CodeRabbit

- **New Features**
  - Native generation now automatically reclaims conversation context when approaching input limits.
  - Older or oversized tool results may be shortened or removed while preserving recent context, tool-call pairings, and task instructions.
  - Context handling adapts based on observed model usage for more reliable multi-step tool workflows.
  - Tool execution timeouts can now be disabled when needed.

- **Bug Fixes**
  - Improved reliability for extended generation loops, reducing failures caused by oversized conversation history.

- **Documentation**
  - Updated API documentation to describe the new context-management configuration and timeout behavior.

### DIFF
--- a/src/lib/context/nativeGenerateGuard.ts
+++ b/src/lib/context/nativeGenerateGuard.ts
@@ -1,0 +1,234 @@
+import type {
+  LoopGuardEntry,
+  NativeGenerateGuardConfig,
+} from "../types/index.js";
+import {
+  estimateTokens,
+  TOKENS_PER_MESSAGE,
+} from "../utils/tokenEstimation.js";
+import { logger } from "../utils/logger.js";
+import { planLoopGuardReclaim } from "./loopGuardCore.js";
+import { DEFAULT_CONTEXT_GUARD_RATIO } from "../core/constants.js";
+import { generateToolOutputPreview } from "./toolOutputLimits.js";
+
+const isRecord = (value: unknown): value is Record<string, unknown> =>
+  typeof value === "object" && value !== null && !Array.isArray(value);
+
+const partsOf = (value: unknown): Record<string, unknown>[] =>
+  Array.isArray(value) ? value.filter(isRecord) : [];
+
+const textOf = (value: unknown): string => {
+  if (typeof value === "string") {
+    return value;
+  }
+  try {
+    return JSON.stringify(value) ?? "";
+  } catch {
+    return "x".repeat(200_000);
+  }
+};
+
+const previewMessage = (
+  message: Record<string, unknown>,
+): Record<string, unknown> => ({
+  ...message,
+  // Do not stringify the whole message: call IDs, result types and metadata
+  // must survive the wire conversion. Only the result payload is shortened.
+  content: partsOf(message.content).map((part) => {
+    if (part.type !== "tool-result") {
+      return part;
+    }
+    const output = part.output;
+    const wrapped =
+      isRecord(output) &&
+      ["text", "json", "error-text", "error-json"].includes(
+        String(output.type),
+      ) &&
+      "value" in output;
+    const { preview, truncated } = generateToolOutputPreview(
+      textOf(wrapped ? output.value : output),
+      { maxBytes: 2048, maxLines: 60 },
+    );
+    if (!truncated) {
+      return part;
+    }
+    return {
+      ...part,
+      output: wrapped
+        ? {
+            ...output,
+            type: String(output.type).startsWith("error")
+              ? "error-text"
+              : "text",
+            value: preview,
+          }
+        : preview,
+    };
+  }),
+});
+
+/** V3-shape adapter for the reclaim policy shared with streaming guards. */
+export function createNativeGenerateGuard(config: NativeGenerateGuardConfig) {
+  let previousSentEstimate: number | undefined;
+  let observedPromptTokens: number | undefined;
+  const cost = (message: Record<string, unknown>): number =>
+    estimateTokens(textOf(message), config.provider) + TOKENS_PER_MESSAGE;
+
+  return {
+    observeUsage(usage: unknown): void {
+      const input = isRecord(usage) ? usage.inputTokens : undefined;
+      const total = isRecord(input) ? input.total : input;
+      observedPromptTokens =
+        typeof total === "number" && Number.isFinite(total) && total > 0
+          ? total
+          : undefined;
+    },
+    guardConversation(
+      conversation: Array<Record<string, unknown>>,
+    ): Array<Record<string, unknown>> | undefined {
+      // Policy entry zero is protected. Fold the entire initial task prefix
+      // into it so leading system messages cannot displace task protection.
+      const taskIndex = conversation.findIndex(
+        (message) => message.role === "user",
+      );
+      const prefixLength = Math.max(1, taskIndex + 1);
+      const tail = conversation.slice(prefixLength);
+      const previews = tail.map(previewMessage);
+      const entries: LoopGuardEntry[] = [
+        {
+          kind: "other",
+          tokens: conversation
+            .slice(0, prefixLength)
+            .reduce((sum, message) => sum + cost(message), 0),
+        },
+        ...tail.map((message, index): LoopGuardEntry => {
+          const parts = partsOf(message.content);
+          if (
+            message.role === "assistant" &&
+            parts.some((part) => part.type === "tool-call")
+          ) {
+            return { kind: "toolCall", tokens: cost(message) };
+          }
+          if (
+            message.role === "tool" &&
+            parts.some((part) => part.type === "tool-result")
+          ) {
+            return {
+              kind: "toolResult",
+              tokens: cost(message),
+              previewTokens: cost(previews[index]),
+            };
+          }
+          return { kind: "other", tokens: cost(message) };
+        }),
+      ];
+      const overhead = config.getFixedOverheadTokens();
+      const calibration =
+        observedPromptTokens && previousSentEstimate
+          ? Math.min(
+              3,
+              Math.max(1, observedPromptTokens / previousSentEstimate),
+            )
+          : 1;
+      const plan = planLoopGuardReclaim(entries, {
+        availableInputTokens: config.availableInputTokens,
+        fixedOverheadTokens: overhead,
+        calibration,
+      });
+      // The policy fires at a RATIO of the window, not at the window itself, so
+      // every number below is measured against that same threshold. Comparing
+      // against the raw budget leaves a ~15% band in which the guard has
+      // already crossed its reclaim point while the log still reads "under
+      // budget" with positive headroom — defeating the purpose of making the
+      // threshold measurable, since a probe sizing a payload from that headroom
+      // would aim past the point it meant to cross.
+      const reclaimThresholdTokens = Math.floor(
+        (config.availableInputTokens * DEFAULT_CONTEXT_GUARD_RATIO) /
+          (calibration > 0 ? calibration : 1),
+      );
+      const estimatedBeforeTokens =
+        overhead + entries.reduce((sum, entry) => sum + entry.tokens, 0);
+      if (!plan.fire) {
+        previousSentEstimate = estimatedBeforeTokens;
+        // Staying under budget is the common case and must not be noisy, but
+        // silence here is unfalsifiable: no line at all reads the same whether
+        // the conversation fit, the guard was never installed, or this hook was
+        // never called. Two live attempts to demonstrate the reclaim against a
+        // real vendor produced nothing and could not distinguish those, because
+        // the budget is not observable from outside the process. Emitting the
+        // budget and the headroom at debug makes the threshold measurable, so a
+        // probe can size a payload to cross it instead of guessing.
+        // Two very different situations reach this branch and must not share a
+        // message. Fitting is routine and belongs at debug. Being over budget
+        // with nothing reclaimable is neither: the planner only touches entries
+        // strictly between the task and the newest
+        // DEFAULT_LOOP_GUARD_PROTECTED_TAIL, so a conversation whose oversize
+        // content sits inside that protected tail — one large tool result, the
+        // common case — yields an empty plan, and the request is sent over the
+        // window anyway to be rejected or silently truncated by the vendor.
+        // Observed live against deepseek: 186,847 estimated against a 62,500
+        // budget, empty plan, no signal of any kind before this warning.
+        if (estimatedBeforeTokens > reclaimThresholdTokens) {
+          logger.warn(
+            "[NativeGenerateGuard] Over threshold, nothing reclaimable",
+            {
+              provider: config.provider,
+              availableInputTokens: config.availableInputTokens,
+              reclaimThresholdTokens,
+              estimatedPromptTokens: estimatedBeforeTokens,
+              overThresholdByTokens:
+                estimatedBeforeTokens - reclaimThresholdTokens,
+              calibration,
+            },
+          );
+        } else {
+          logger.debug("[NativeGenerateGuard] Under threshold", {
+            provider: config.provider,
+            availableInputTokens: config.availableInputTokens,
+            reclaimThresholdTokens,
+            estimatedPromptTokens: estimatedBeforeTokens,
+            headroomToThresholdTokens:
+              reclaimThresholdTokens - estimatedBeforeTokens,
+            calibration,
+          });
+        }
+        return undefined;
+      }
+      const drop = new Set(plan.drop);
+      const truncate = new Set(plan.truncate);
+      const result = [...conversation.slice(0, prefixLength)];
+      let noted = false;
+      tail.forEach((message, index) => {
+        const entry = index + 1;
+        if (drop.has(entry)) {
+          if (!noted) {
+            result.push({
+              role: "user",
+              content:
+                "[Earlier tool exchanges were removed to fit the context window.]",
+            });
+          }
+          noted = true;
+        } else {
+          result.push(truncate.has(entry) ? previews[index] : message);
+        }
+      });
+      previousSentEstimate =
+        overhead + result.reduce((sum, message) => sum + cost(message), 0);
+      logger.info("[NativeGenerateGuard] Reclaimed context", {
+        provider: config.provider,
+        // The budget and both estimates travel with the event so a reader can
+        // see WHY it fired and by how much it overshot, rather than only that
+        // it did.
+        availableInputTokens: config.availableInputTokens,
+        reclaimThresholdTokens,
+        estimatedBeforeTokens,
+        estimatedAfterTokens: previousSentEstimate,
+        truncated: plan.truncate.length,
+        dropped: plan.drop.length,
+        calibration,
+      });
+      return result;
+    },
+  };
+}
--- a/src/lib/providers/openaiChatCompletionsBase.ts
+++ b/src/lib/providers/openaiChatCompletionsBase.ts
@@ -28,6 +28,11 @@ import {
 } from "../constants/contextWindows.js";
 import { guardOpenAICompatConversation } from "../context/openaiCompatLoopGuard.js";
 import {
+  estimateTokens,
+  serializeForEstimate,
+} from "../utils/tokenEstimation.js";
+import { createNativeGenerateGuard } from "../context/nativeGenerateGuard.js";
+import {
   isContextOverflowError,
   parseProviderOverflowDetails,
 } from "../context/errorDetection.js";
@@ -1189,6 +1194,27 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
       runNativeGenerateLoop(
         {
           doGenerate,
+          ...createNativeGenerateGuard({
+            provider: this.providerName,
+            availableInputTokens: getAvailableInputTokens(
+              this.providerName,
+              modelId,
+              options.maxTokens ?? undefined,
+            ),
+            // Runs on every step of the generate loop, so a cyclic or
+            // oversized tool schema must not throw the turn away. Shares the
+            // estimator's own serializer: its fallback over-reports, which
+            // makes the guard fire rather than silently size the overhead at
+            // nothing.
+            getFixedOverheadTokens: () =>
+              estimateTokens(
+                serializeForEstimate({
+                  tools: v3Tools,
+                  responseFormat: format,
+                }),
+                this.providerName,
+              ),
+          }),
           conversation: conv,
           ...(v3Tools ? { tools: v3Tools } : {}),
           toolsRecord,
--- a/test/continuous-test-suite-native-vendor-recovery.ts
+++ b/test/continuous-test-suite-native-vendor-recovery.ts
@@ -692,5 +692,208 @@ await test("stream preserves an explicit caller cache marker without adding anot
     }
   }
 });
+// ---------------------------------------------------------------------------
+// Per-step context reclaim on the GENERATE path.
+//
+// Both native loop guards — guardOpenAICompatConversation and
+// planAnthropicLoopReclaim — were wired into executeStream only, so a long
+// agentic generate() walked into a provider context-overflow with no reclaim,
+// while the identical stream() turn reclaimed. Nothing covered either guard.
+// ---------------------------------------------------------------------------
+
+await test("generate() reclaims context when a tool loop outgrows the window", async () => {
+  const STEPS = 6;
+  // Head/tail preview keeps the ends and drops the middle, so a sentinel in
+  // the middle is exactly what disappears when an output is truncated.
+  const blob = (i: number): string =>
+    `${"A".repeat(50_000)}SENTINEL_${i}${"B".repeat(50_000)}`;
+
+  const script = [];
+  for (let i = 0; i < STEPS; i++) {
+    script.push(
+      chatCompletion({
+        finishReason: "tool_calls",
+        toolCalls: [
+          {
+            id: `call_${i}`,
+            type: "function",
+            function: { name: "big_tool", arguments: JSON.stringify({ i }) },
+          },
+        ],
+      }),
+    );
+  }
+  script.push(
+    chatCompletion({ content: "final answer", finishReason: "stop" }),
+  );
+
+  let calls = 0;
+  const server = await startScriptedChatServer(script);
+  const sdk = new NeuroLink();
+  try {
+    const result = await sdk.generate({
+      input: { text: "call the tool repeatedly then answer" },
+      provider: "openai",
+      model: "gpt-4o-mini",
+      disableInternalFallback: true,
+      maxSteps: STEPS + 2,
+      credentials: credentialsFor(server.baseURL),
+      tools: {
+        big_tool: tool({
+          description: "Returns a very large payload",
+          inputSchema: z.object({ i: z.number() }),
+          execute: async () => blob(calls++),
+        }),
+      },
+    });
+
+    // Preconditions: the loop must actually have run every step, or there is
+    // no oversized history for a guard to reclaim and the claim proves nothing.
+    if (server.requestCount() < STEPS + 1) {
+      throw new Error(
+        `precondition: loop stopped early at ${server.requestCount()} requests`,
+      );
+    }
+    if (calls < STEPS) {
+      throw new Error(
+        `precondition: tool ran ${calls} times, expected ${STEPS}`,
+      );
+    }
+    if (result.content !== "final answer") {
+      throw new Error("precondition: the final answer did not reach content");
+    }
+
+    const bodies = server.getAllRequestBodies();
+    const last = bodies[bodies.length - 1] ?? "";
+    // The newest tool output rides inside the protected tail and must survive
+    // untouched — without this, "reclaimed" could just mean "dropped it all".
+    if (!last.includes(`SENTINEL_${STEPS - 1}`)) {
+      throw new Error(
+        "the most recent tool output was reclaimed, but must not be",
+      );
+    }
+    // The oldest output sits well outside the protected tail. If no reclaim
+    // happened it is still there in full.
+    if (last.includes("SENTINEL_0")) {
+      throw new Error(
+        "the oldest tool output was still sent in full — no context reclaim ran on the generate path",
+      );
+    }
+  } finally {
+    await sdk.shutdown();
+    await server.close();
+  }
+});
+
+for (const calibrated of [false, true]) {
+  await test(`generate reclaim preserves tool pairs (${calibrated ? "calibrated usage" : "batch dropping"})`, async () => {
+    const count = 8;
+    const padding = "x".repeat(calibrated ? 24_000 : 75_000);
+    const replies = Array.from({ length: count }, (_, i) => ({
+      ...chatCompletion({
+        finishReason: "tool_calls",
+        toolCalls: [
+          {
+            id: `batch_${i}`,
+            type: "function",
+            function: {
+              name: "record_batch",
+              arguments: JSON.stringify({ i, padding }),
+            },
+          },
+        ],
+      }),
+      usage: {
+        prompt_tokens: calibrated ? 100_000 : 1,
+        completion_tokens: 1,
+        total_tokens: calibrated ? 100_001 : 2,
+      },
+    }));
+    const server = await startScriptedChatServer([
+      ...replies,
+      chatCompletion({ content: "paired answer" }),
+    ]);
+    const sdk = new NeuroLink();
+    let executions = 0;
+    try {
+      const result = await sdk.generate({
+        input: { text: "TASK_MUST_SURVIVE" },
+        provider: "openai",
+        model: "gpt-4o-mini",
+        maxSteps: count + 1,
+        maxTokens: 1024,
+        disableInternalFallback: true,
+        credentials: credentialsFor(server.baseURL),
+        tools: {
+          record_batch: tool({
+            description: "Return a small result",
+            inputSchema: z.object({ i: z.number(), padding: z.string() }),
+            execute: async ({ i }) => {
+              executions++;
+              return `result_${i}`;
+            },
+          }),
+        },
+      });
+      if (
+        executions !== count ||
+        server.requestCount() !== count + 1 ||
+        result.content !== "paired answer"
+      ) {
+        throw new Error("precondition: all scripted batch steps must finish");
+      }
+      const requests = server.getAllRequestBodies().map(
+        (body) =>
+          JSON.parse(body) as {
+            messages: Array<{
+              role: string;
+              content?: unknown;
+              tool_call_id?: string;
+              tool_calls?: Array<{ id: string }>;
+            }>;
+          },
+      );
+      if (!JSON.stringify(requests[1]).includes("batch_0")) {
+        throw new Error("precondition: original batch never reached the wire");
+      }
+      for (const request of requests) {
+        const pending = new Set<string>();
+        for (const message of request.messages) {
+          if (message.role === "tool") {
+            if (
+              !message.tool_call_id ||
+              !pending.delete(message.tool_call_id)
+            ) {
+              throw new Error("reclaim emitted an orphan tool result");
+            }
+          } else {
+            if (pending.size) {
+              throw new Error("reclaim emitted an unanswered tool call");
+            }
+            for (const call of message.tool_calls ?? []) {
+              pending.add(call.id);
+            }
+          }
+        }
+        if (pending.size) {
+          throw new Error("request ends with unanswered tool calls");
+        }
+      }
+      const last = JSON.stringify(requests.at(-1));
+      if (last.includes("batch_0")) {
+        throw new Error("old batch was not reclaimed");
+      }
+      if (
+        !last.includes(`batch_${count - 1}`) ||
+        !last.includes("TASK_MUST_SURVIVE")
+      ) {
+        throw new Error("reclaim removed the task or protected tail");
+      }
+    } finally {
+      await sdk.shutdown();
+      await server.close();
+    }
+  });
+}
 
 await runSuite();
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -1035,6 +1035,7 @@ console.log(result.content);
 - [SummarizeResult](type-aliases/SummarizeResult.md)
 - [DeduplicationResult](type-aliases/DeduplicationResult.md)
 - [FileSummarizationServiceOptions](type-aliases/FileSummarizationServiceOptions.md)
+- [NativeGenerateGuardConfig](type-aliases/NativeGenerateGuardConfig.md)
 - [StorageConfig](type-aliases/StorageConfig.md)
 - [ConversationMemoryConfig](type-aliases/ConversationMemoryConfig.md)
 - [SessionMemory](type-aliases/SessionMemory.md)
--- a/docs/api/type-aliases/NativeGenerateGuardConfig.md
+++ b/docs/api/type-aliases/NativeGenerateGuardConfig.md
@@ -1,0 +1,41 @@
+[**NeuroLink API Reference**](../README.md)
+
+---
+
+[NeuroLink API Reference](../README.md) / NativeGenerateGuardConfig
+
+# Type Alias: NativeGenerateGuardConfig
+
+> **NativeGenerateGuardConfig** = `object`
+
+Defined in: [types/context.ts:1032](https://github.com/juspay/neurolink/blob/release/src/lib/types/context.ts#L1032)
+
+Per-turn configuration for reclaiming a native V3 conversation.
+
+## Properties
+
+### provider
+
+> **provider**: `string`
+
+Defined in: [types/context.ts:1033](https://github.com/juspay/neurolink/blob/release/src/lib/types/context.ts#L1033)
+
+---
+
+### availableInputTokens
+
+> **availableInputTokens**: `number`
+
+Defined in: [types/context.ts:1034](https://github.com/juspay/neurolink/blob/release/src/lib/types/context.ts#L1034)
+
+---
+
+### getFixedOverheadTokens
+
+> **getFixedOverheadTokens**: () => `number`
+
+Defined in: [types/context.ts:1035](https://github.com/juspay/neurolink/blob/release/src/lib/types/context.ts#L1035)
+
+#### Returns
+
+`number`
--- a/docs/api/type-aliases/NativeGenerateLoopArgs.md
+++ b/docs/api/type-aliases/NativeGenerateLoopArgs.md
@@ -16,11 +16,54 @@ One loop serves every provider whose delegating model exposes a v3-shaped
 
 ## Properties
 
+### observeUsage?
+
+> `optional` **observeUsage?**: (`usage`) => `void`
+
+Defined in: [types/generate.ts:1785](https://github.com/juspay/neurolink/blob/release/src/lib/types/generate.ts#L1785)
+
+Observed usage for calibrating the next step against the last request.
+
+#### Parameters
+
+##### usage
+
+`unknown`
+
+#### Returns
+
+`void`
+
+---
+
+### guardConversation?
+
+> `optional` **guardConversation?**: (`conversation`) => `Record`\<`string`, `unknown`\>[] \| `undefined`
+
+Defined in: [types/generate.ts:1792](https://github.com/juspay/neurolink/blob/release/src/lib/types/generate.ts#L1792)
+
+Per-step context reclaim, called before every model call with the
+conversation as it now stands. Return a replacement to have the loop adopt
+it, or undefined to leave it untouched. The provider owns this because the
+reclaim has to understand its wire shape.
+
+#### Parameters
+
+##### conversation
+
+`Record`\<`string`, `unknown`\>[]
+
+#### Returns
+
+`Record`\<`string`, `unknown`\>[] \| `undefined`
+
+---
+
 ### doGenerate
 
 > **doGenerate**: (`options`) => `Promise`\<`Record`\<`string`, `unknown`\>\>
 
-Defined in: [types/generate.ts:1784](https://github.com/juspay/neurolink/blob/release/src/lib/types/generate.ts#L1784)
+Defined in: [types/generate.ts:1795](https://github.com/juspay/neurolink/blob/release/src/lib/types/generate.ts#L1795)
 
 #### Parameters
 
@@ -38,7 +81,7 @@ Defined in: [types/generate.ts:1784](https://github.com/juspay/neurolink/blob/re
 
 > **conversation**: `Record`\<`string`, `unknown`\>[]
 
-Defined in: [types/generate.ts:1788](https://github.com/juspay/neurolink/blob/release/src/lib/types/generate.ts#L1788)
+Defined in: [types/generate.ts:1799](https://github.com/juspay/neurolink/blob/release/src/lib/types/generate.ts#L1799)
 
 Conversation in the message-builder shape each doGenerate converts itself.
 
@@ -48,7 +91,7 @@ Conversation in the message-builder shape each doGenerate converts itself.
 
 > `optional` **tools?**: `Record`\<`string`, `unknown`\>[]
 
-Defined in: [types/generate.ts:1790](https://github.com/juspay/neurolink/blob/release/src/lib/types/generate.ts#L1790)
+Defined in: [types/generate.ts:1801](https://github.com/juspay/neurolink/blob/release/src/lib/types/generate.ts#L1801)
 
 Tool declarations in the v3 shape doGenerate already knows how to convert.
 
@@ -58,7 +101,7 @@ Tool declarations in the v3 shape doGenerate already knows how to convert.
 
 > **toolsRecord**: `Record`\<`string`, `unknown`\>
 
-Defined in: [types/generate.ts:1792](https://github.com/juspay/neurolink/blob/release/src/lib/types/generate.ts#L1792)
+Defined in: [types/generate.ts:1803](https://github.com/juspay/neurolink/blob/release/src/lib/types/generate.ts#L1803)
 
 Registered tools, used to execute a call the model asks for.
 
@@ -68,7 +111,7 @@ Registered tools, used to execute a call the model asks for.
 
 > `optional` **toolChoice?**: `unknown`
 
-Defined in: [types/generate.ts:1793](https://github.com/juspay/neurolink/blob/release/src/lib/types/generate.ts#L1793)
+Defined in: [types/generate.ts:1804](https://github.com/juspay/neurolink/blob/release/src/lib/types/generate.ts#L1804)
 
 ---
 
@@ -76,7 +119,7 @@ Defined in: [types/generate.ts:1793](https://github.com/juspay/neurolink/blob/re
 
 > `optional` **responseFormat?**: `Record`\<`string`, `unknown`\>
 
-Defined in: [types/generate.ts:1794](https://github.com/juspay/neurolink/blob/release/src/lib/types/generate.ts#L1794)
+Defined in: [types/generate.ts:1805](https://github.com/juspay/neurolink/blob/release/src/lib/types/generate.ts#L1805)
 
 ---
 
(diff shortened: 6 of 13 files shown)


src/lib/providers/openaiChatCompletionsBase.ts
```diff
@@ -1181,14 +1182,66 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
...
+      modelId,
+      options.maxTokens ?? undefined,
+    );
+    // Tool definitions ride outside the message array. Estimating them from the
+    // v3 shape rather than converting to the wire shape keeps this honest: it
+    // is the same order of magnitude, and the guard's threshold is a ratio.
+    const toolOverheadTokens = hasTools
+      ? estimateTokens(JSON.stringify(v3Tools), this.providerName)
+      : 0;
+    const guardConversation = (
```
Tara-ag: `<!-- yama:generate-guard-wire-shape -->`

**MAJOR — the guard is fed the message-builder shape but `guardOpenAICompatConversation` classifies by the OpenAI wire shape.**

`runNativeGenerateLoop` passes `args.conversation` to this hook, and that array is in the **message-builder / v3 shape**: the loop appends assistant tool-call turns as `{ role: "assistant", content: parts }` where `parts` are v3 content parts (`type: "tool-call"`, `toolName`, `toolCallId`, `input`) — there is **no `tool_calls` field** on those messages (see `nativeGenerateLoop.ts` `args.conversation.push({ role: "assistant", content: parts })`).

But `guardOpenAICompatConversation` → `toEntries()` keys its `toolCall` classification on `message.role === "assistant" && message.tool_calls?.length`. On the generate path that is always falsy, so every assistant tool-call message falls to `kind: "other"`. Only the `role: "tool"` messages are seen as `toolResult`.

Consequence when reclaim reaches `planLoopGuardReclaim` **stage 2** (batch dropping — the regime for long loops whose outputs are already small): `findToolBatches` runs over `[other(task), other(assistant tool-call), toolResult, other(assistant tool-call), toolResult, …]`, detects each `toolResult` as a **lone batch**, and drops it **without its paired assistant tool-call message**. The streamed wire request then carries an assistant `tool_calls` turn with no following `tool` message — OpenAI-compatible backends reject that with a 400. That is exactly the "orphan" the policy's own docstring in `loopGuardCore.ts` claims to prevent, and it defeats the feature's purpose (a provider 400 mid-loop).

The streaming path avoids this only because its `conversation` is already `messageBuilderToOpenAI`-converted (assistant messages carry `tool_calls`).

**Fix (one of):**
- pass the guard the OpenAI *wire* conversation on the generate path (convert before hooking), or
- teach the adapter's `toEntries` to recognise v3 `content` parts of `type: "tool-call"` as `toolCall` (so stage-2 pairs correctly), or
- at minimum, since only stage-1 (tool-output truncation) is shape-safe here, disable stage-2 drop for the generate hook until the pairing is fixed.

The `conv as OpenAICompatChatMessage[]` cast here hides rather than exposes this — the incoming array is not actually that shape.

src/lib/providers/openaiChatCompletionsBase.ts
```diff
@@ -1181,14 +1182,66 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
...
+    // The guard needs the prompt tokens the provider actually billed for the
+    // request its own estimate described. doGenerate is the only place that
+    // sees them, so record them as they go past.
+    const guardedDoGenerate = async (
+      callOptions: Record<string, unknown>,
+    ): Promise<Record<string, unknown>> => {
+      const res = await doGenerate(callOptions);
+      const input = (res as { usage?: { inputTokens?: unknown } }).usage
+        ?.inputTokens;
+      if (typeof input === "number" && input > 0) {
```
Tara-ag: `<!-- yama:generate-guard-calibration-shape -->`

**MINOR — calibration can silently never apply if `doGenerate` returns shaped `inputTokens`.**

This wrapper records `lastObservedPromptTokens` only when `usage.inputTokens` is a bare `number`. But `runNativeGenerateLoop` reads the same field via `readTotal()`, which deliberately accepts *both* a number and a `{ total }` object. If this provider's `doGenerate` returns `usage.inputTokens` as `{ total: N }` (the shaped form the loop already handles), then `typeof input === "number"` is false, `lastObservedPromptTokens` stays `undefined`, and `calibration` stays at its 1.0 floor for the whole turn — the calibration ratio this wrapper exists to feed would silently do nothing.

Worth confirming the actual `doGenerate` return shape here. Safe hardening: reuse the same tolerant read the loop uses (dereference `.total` when present) instead of requiring a bare number.

test/continuous-test-suite-native-vendor-recovery.ts
```diff
@@ -501,4 +501,97 @@ await test("the Anthropic tools block carries one cache breakpoint", async () =>
...
+    // The newest tool output rides inside the protected tail and must survive
+    // untouched — without this, "reclaimed" could just mean "dropped it all".
+    if (!last.includes(`SENTINEL_${STEPS - 1}`)) {
+      throw new Error(
+        "the most recent tool output was reclaimed, but must not be",
+      );
+    }
+    // The oldest output sits well outside the protected tail. If no reclaim
+    // happened it is still there in full.
+    if (last.includes("SENTINEL_0")) {
```
Tara-ag, line 777: `<!-- yama:generate-guard-test-coverage -->`

**MINOR — this test only proves stage-1 truncation; it can't catch the stage-2 pairing bug, and the fake server wouldn't flag an orphan anyway.**

The 100 KB blobs make every tool output far over the 2 KB preview budget, so reclaim converges by **stage 1 truncation alone** — `planLoopGuardReclaim` never reaches stage 2 (batch dropping) in this scenario. That is exactly the stage the wire-shape mismatch I flagged above breaks. And `startScriptedChatServer` (`test/helpers/mockChatServer.ts`) returns canned scripted bodies **regardless of the request content** — it records bodies but never validates that an assistant `tool_calls` message is followed by a matching `tool` message. So even a real orphan would pass.

Right now these five assertions on `last` prove "reclaim happened" but not "reclaim is structurally safe". Consider a companion test that forces stage 2 (many steps whose outputs are just under the preview budget so truncation buys nothing) and, where possible, asserts the last request still satisfies the tool-call/tool pairing invariant.

Tara-ag (commented):
Reviewed generate-path context reclaim (feat/generate-path-loop-guard). Design is sound and the calibration/truncation plumbing is correct, but there is one correctness gap (wire-shape mismatch → stage-2 can orphan tool_calls) and a test-coverage hole that lets it through. See inline comments.

Tara-ag:
*Superseded* round note (head 944a8863, NEEDS_WORK). The current, single authoritative summary for this PR is comment `5648317567` (APPROVE at head `171d4637`), which records the full consolidated disposition. This earlier note is retained only as history.

src/lib/providers/openaiChatCompletionsBase.ts
```diff
@@ -1181,14 +1182,66 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
...
+    // v3 shape rather than converting to the wire shape keeps this honest: it
+    // is the same order of magnitude, and the guard's threshold is a ratio.
+    const toolOverheadTokens = hasTools
+      ? estimateTokens(JSON.stringify(v3Tools), this.providerName)
+      : 0;
+    const guardConversation = (
+      conv: Array<Record<string, unknown>>,
+    ): Array<Record<string, unknown>> | undefined =>
+      guardOpenAICompatConversation({
+        conversation: conv as OpenAICompatChatMessage[],
```
Tara-ag: **MAJOR** — the guard rewrites a v3 message-builder conversation as if it were already wire-shaped, which can orphan tool calls on the generate path.

`guardConversation` receives `args.conversation` — the loop's message-builder/v3 shape, where an assistant tool turn lives as `content` parts and has **no `tool_calls` field**. But `guardOpenAICompatConversation.toEntries` classifies an assistant message as `toolCall` only when `message.tool_calls?.length` is truthy (`src/lib/context/openaiCompatLoopGuard.ts`). So on the generate path every assistant tool-turn is classified `other`, and every `role: "tool"` result is classified `toolResult`.

Then `planLoopGuardReclaim` stage-2 (`findToolBatches`, `src/lib/context/loopGuardCore.ts`) scans `toolCall` then `toolResult` entries. With the assistant turns invisible as `toolCall`, each `role:"tool"` message becomes its own lone "batch" and gets dropped **without its paired assistant tool-call message** — which is dropped by the `other` classification instead. `doGenerate` then puts a `tool_calls`-bearing assistant message on the wire with no answering `role:"tool"` message → a provider hard 400. That is the exact failure this PR claims to prevent.

The streaming path avoids this only because its conversation is already the wire shape (`OpenAICompatChatMessage[]` with `tool_calls`); the function is being reused on a shape it was never written for.

<details><summary>details</summary>

`OpenAICompatChatMessage` case in `toEntries`:
```ts
if (message.role === "assistant" && message.tool_calls?.length) {
  return { kind: "toolCall" as const, tokens };
}
return { kind: "other" as const, tokens };
```
`findToolBatches` groups `toolCall` entries then `toolResult` entries into one batch; with zero `toolCall` entries, each `toolResult` is a batch of one.
</details>

Fix: reshape `conv` to the `OpenAICompatChatMessage` (wire) form *before* calling the guard (convert v3 tool-parts to `tool_calls` and map the matching `role:"tool"` parts), then apply the returned wire-shaped plan back, or feed the guard entries that classify the assistant part-messages as `toolCall`. Otherwise gate the reclaim off — do not run a wire-shape guard over a v3 shape.

src/lib/providers/openaiChatCompletionsBase.ts
Tara-ag: **MINOR** — calibration silently stays at 1.0 when `doGenerate` returns shaped usage.

`guardedDoGenerate` only records `lastObservedPromptTokens` when `usage.inputTokens` is a bare number. But the loop's own token reader (`readTotal` in `nativeGenerateLoop.ts`) accepts the shaped `{ total }` form too. If `doGenerate` returns shaped usage, this wrapper records nothing, `observedPromptTokens` stays set from `previousSentEstimate`, and the calibration ratio in `guardOpenAICompatConversation` never rises above the `Math.max(1, …)` floor — exactly the under-estimate that lets a dense run overflow.

Fix: reuse the same shape detection the loop uses, e.g.:
```ts
const readInputTokens = (res: unknown): number | undefined => {
  const value = (res as { usage?: { inputTokens?: unknown } }).usage?.inputTokens;
  if (typeof value === "number") return value;
  const total = (res as { usage?: { total?: unknown } }).usage?.total;
  return typeof total === "number" ? total : undefined;
};
```

test/continuous-test-suite-native-vendor-recovery.ts
Tara-ag, line 777: **MINOR** — the new test only exercises stage-1 truncation, never stage-2 batch dropping, and never validates tool-call/tool pairing — so the orphan bug in the MAJOR finding would pass green.

The 100 KB payload always triggers head/tail **truncation** of old `role:"tool"` outputs (stage 1), which is the safe path. It does not drive stage 2 (`findToolBatches` dropping whole batches), where the pairing regression lives — so the test gives a false sense of coverage for the reclaim branch that's actually at risk. The scripted server also never asserts that each assistant `tool_calls` array has a matching `role:"tool"` message, so an orphaned tool-call (provider 400 in real life) is invisible here.

Fix: add a case whose history exceeds the window after previews alone cannot converge (small-but-numerous tool outputs) so stage-2 batch dropping fires, and assert on the request bodies that every assistant wire message with `tool_calls` has exactly the matching `role:"tool"` responses before the next one.

Tara-ag (changes requested):
NEEDS_WORK. One MAJOR correctness concern: the generate-path guard reuses `guardOpenAICompatConversation`, which keys tool-call classification on the wire `message.tool_calls` field, but the loop feeds it the v3 message-builder conversation (assistant tool calls live in `content`) — under stage-2 batch dropping this orphans tool results and can produce the exact provider 400 the PR claims to fix. Two MINORs on calibration shape detection and test coverage follow inline.

Tara-ag:
*Superseded* round note (head 944a8863, NEEDS_WORK). The current, single authoritative summary for this PR is comment `5648317567` (APPROVE at head `171d4637`), which records the full consolidated disposition. This earlier note is retained only as history.

src/lib/providers/openaiChatCompletionsBase.ts
murdore: Replaced the wire-shape cast with nativeGenerateGuard.ts, a V3-message adapter over loopGuardCore. It classifies assistant content tool-call parts, previews only result payloads while preserving toolCallId/output envelopes, and drops complete call/result exchanges. A new public SDK test reproduces the orphaned-call failure before the fix and validates every outbound request after it. Vendor-recovery passes 9/9.

src/lib/providers/openaiChatCompletionsBase.ts
murdore: Fixed in the V3-specific guard adapter. The loop forwards the completed step usage to an observer that accepts inputTokens.total (or a numeric inputTokens). Calibration compares that observation with the preceding request estimate. The new calibrated-usage regression failed before the change and passes afterwards; vendor-recovery is 9/9.

test/continuous-test-suite-native-vendor-recovery.ts
murdore: Added separate batch-dropping and calibration tests using large call arguments with small tool results, so previewing outputs cannot satisfy the claim. Every outbound request is checked for matching tool-call/result IDs, and the task plus newest batch must remain. Baseline failed with an unanswered tool call and no calibrated reclaim; the corrected V3 adapter passes 9/9.

src/lib/providers/openaiChatCompletionsBase.ts
murdore: Replaced the wire-shape cast with nativeGenerateGuard.ts, a V3-message adapter over loopGuardCore. It classifies assistant content tool-call parts, previews only result payloads while preserving toolCallId/output envelopes, and drops complete call/result exchanges. A new public SDK test reproduces the orphaned-call failure before the fix and validates every outbound request after it. Vendor-recovery passes 9/9.

src/lib/providers/openaiChatCompletionsBase.ts
murdore: Fixed in the V3-specific guard adapter. The loop forwards the completed step usage to an observer that accepts inputTokens.total (or a numeric inputTokens). Calibration compares that observation with the preceding request estimate. The new calibrated-usage regression failed before the change and passes afterwards; vendor-recovery is 9/9.

test/continuous-test-suite-native-vendor-recovery.ts
murdore: Added separate batch-dropping and calibration tests using large call arguments with small tool results, so previewing outputs cannot satisfy the claim. Every outbound request is checked for matching tool-call/result IDs, and the task plus newest batch must remain. Baseline failed with an unanswered tool call and no calibrated reclaim; the corrected V3 adapter passes 9/9.

src/lib/providers/openaiChatCompletionsBase.ts
```diff
@@ -1181,14 +1182,66 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
...
+    // request its own estimate described. doGenerate is the only place that
+    // sees them, so record them as they go past.
+    const guardedDoGenerate = async (
+      callOptions: Record<string, unknown>,
+    ): Promise<Record<string, unknown>> => {
+      const res = await doGenerate(callOptions);
+      const input = (res as { usage?: { inputTokens?: unknown } }).usage
+        ?.inputTokens;
+      if (typeof input === "number" && input > 0) {
+        lastObservedPromptTokens = input;
```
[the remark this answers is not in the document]
murdore: Fixed in the V3-specific guard adapter. The loop forwards the completed step usage to an observer that accepts inputTokens.total (or a numeric inputTokens). Calibration compares that observation with the preceding request estimate. The new calibrated-usage regression failed before the change and passes afterwards; vendor-recovery is 9/9.

Tara-ag:
*Superseded* round note (head 153b535, APPROVE). The current, single authoritative summary for this PR is comment `5648317567` (APPROVE at head `171d4637`), which records the full consolidated disposition. This earlier note is retained only as history.

src/lib/context/nativeGenerateGuard.ts
```diff
@@ -0,0 +1,172 @@
...
+            .slice(0, prefixLength)
+            .reduce((sum, message) => sum + cost(message), 0),
+        },
+        ...tail.map((message, index): LoopGuardEntry => {
+          const parts = partsOf(message.content);
+          if (
+            message.role === "assistant" &&
+            parts.some((part) => part.type === "tool-call")
+          ) {
+            return { kind: "toolCall", tokens: cost(message) };
```
Tara-ag, line 109: **MINOR** — `observeUsage` clears `observedPromptTokens` to `undefined` whenever a step's response omits (or zeroes) `inputTokens`, silently dropping calibration back to the 1.0 floor until the next usage-bearing step.

`guardConversation` reads `previousSentEstimate` + `observedPromptTokens` together: if either is falsy, `calibration = 1`. Many OpenAI-compatible steps report sparse/omitted `usage` in a tool loop, so the reclaim threshold can oscillate between calibrated and uncalibrated from step to step rather than converging on a stable fit. This mirrors the streaming loop, so it is not a regression — but for this new adapter the last-known positive observation is a strictly better prior than resetting to a guess of 1.

```suggestion
    observeUsage(usage: unknown): void {
      const input = isRecord(usage) ? usage.inputTokens : undefined;
      const total = isRecord(input) ? input.total : input;
      if (typeof total === "number" && Number.isFinite(total) && total > 0) {
        observedPromptTokens = total;
      }
    },
```

If deliberate (a step with no billed count should not steer calibration), keep it as-is — but then note it, since as written a sparse-usage loop jostles the threshold every step.

Tara-ag (commented):
Reviewed the rewritten V3 adapter, the loop integration, the wire conversion, and the new tests. The rewrite resolves all prior MAJOR/MINOR findings. One MINOR note on calibration resilience above; no blocking issues.

src/lib/context/nativeGenerateGuard.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  return {
+    observeUsage(usage: unknown): void {
+      const input = isRecord(usage) ? usage.inputTokens : undefined;
+      const total = isRecord(input) ? input.total : input;
+      observedPromptTokens =
+        typeof total === "number" && Number.isFinite(total) && total > 0
+          ? total
+          : undefined;
+    },
```
Tara-ag, line 84: **MINOR — `observeUsage` resets calibration back to the 1.0 floor on any step whose usage omits a positive numeric `inputTokens`.**

`observeUsage` assigns `observedPromptTokens = undefined` whenever `inputTokens` is missing, zero, or non-finite. On the next `guardConversation` the calibration gate `observedPromptTokens && previousSentEstimate` then falls back to `calibration = 1`, silently discarding the last good observation. In a loop where usage is reported on only some steps, the effective calibration snaps back to 1.0 instead of holding the measured ratio — re-under-estimating a dense turn and risking the exact overflow this guard exists to prevent.

This mirrors the streaming path (it also overwrites `lastObservedPromptTokens` per step), so it is a non-blocking resilience nit, not a regression. Keeping the previous observation when the current step reports none is a strict improvement:

```suggestion
      if (
        typeof total === "number" &&
        Number.isFinite(total) &&
        total > 0
      ) {
        observedPromptTokens = total;
      }
```

Tara-ag (approved):
The rewritten V3 adapter correctly classifies assistant tool-call content parts, preserves toolCallId/output envelopes, and drops complete call/result exchanges — resolving the MAJOR wire-shape mismatch and both MINORs from the first review, with new batch-dropping and calibrated-usage regression coverage. One non-blocking MINOR on calibration resilience is noted inline. No blocking issues; approving.

Tara-ag:
*Superseded* round note (head 153b535, APPROVE). The current, single authoritative summary for this PR is comment `5648317567` (APPROVE at head `171d4637`), which records the full consolidated disposition. This earlier note is retained only as history.

src/lib/context/nativeGenerateGuard.ts
Tara-ag: Duplicate of the existing finding in this thread: #discussion_r3961588683 (`yama:native-generate-observe-usage-reset`). Resolving this copy; the original thread above is the canonical one.

src/lib/context/nativeGenerateGuard.ts
murdore: Resetting this observation is deliberate: previousSentEstimate advances on every request, so retaining only an older numerator would compare counts from different requests and create a false ratio. A future improvement could retain the last completed calibration ratio, but the suggested numerator-only retention is not safe. Missing usage currently falls back to the conservative static estimator, matching the existing streaming contract. Leaving the implementation unchanged for this nonblocking resilience note.

Tara-ag:
*Superseded* round note (head 153b535, APPROVE). The current, single authoritative summary for this PR is comment `5648317567` (APPROVE at head `171d4637`), which records the full consolidated disposition. This earlier note is retained only as history.

Tara-ag (approved):
Approving the current head (`8119bbc`). The rewritten V3 adapter (`src/lib/context/nativeGenerateGuard.ts`) resolves every finding from the earlier NEEDS_WORK review: assistant tool-call content parts classify as `toolCall`, result payloads are previewed while preserving `toolCallId`/output envelopes, and complete call/result exchanges are dropped so stage-2 batch reclaim can no longer orphan a `tool_calls` message on the wire. New batch-dropping and calibrated-usage regression tests validate tool-call/result pairing on every outbound request (vendor-recovery 9/9). The single remaining MINOR (the `observeUsage` reset at `nativeGenerateGuard.ts:81/109`) is adequately justified by the author (retaining an older numerator against an advanced `previousSentEstimate` would create a false ratio; fallback-to-1 is the conservative, streaming-consistent behavior), so it is accepted and not blocking. No CRITICAL/MAJOR issues remain.

Tara-ag:
*Superseded* round note (head b04c501, APPROVE). The current, single authoritative summary for this PR is comment `5648317567` (APPROVE at head `171d4637`), which records the full consolidated disposition. This earlier note is retained only as history.

Tara-ag:
*Superseded* round note (head e473d02, APPROVE). The current, single authoritative summary for this PR is comment `5648317567` (APPROVE at head `171d4637`), which records the full consolidated disposition. This earlier note is retained only as history.

Tara-ag (approved):
APPROVE — no CRITICAL/MAJOR/MINOR findings on the current head.

The rewritten V3 adapter (`src/lib/context/nativeGenerateGuard.ts`) resolves every earlier finding from the first NEEDS_WORK pass: assistant `content` `type:"tool-call"` parts classify as `toolCall`, result payloads are previewed while preserving `toolCallId`/output envelopes, and complete call→result exchanges are dropped so stage-2 batch reclaim cannot orphan a `tool_calls` message on the wire. New batch-dropping + calibrated-usage regression tests assert tool-call/result pairing on every outbound request (vendor-recovery 9/9). All prior inline threads are resolved; the single remaining MINOR (`observeUsage` reset at `nativeGenerateGuard.ts:81/109`) is adequately justified by the author and non-blocking. Backward compat (Rule 5) holds — `guardConversation`/`observeUsage` are optional on `NativeGenerateLoopArgs`, and the other two `runNativeGenerateLoop` callers (Anthropic, SageMaker) don't pass them.

Tara-ag:
*Superseded* round note (head e473d02, APPROVE). The current, single authoritative summary for this PR is comment `5648317567` (APPROVE at head `171d4637`), which records the full consolidated disposition. This earlier note is retained only as history.

Tara-ag:
*Superseded* round note (head 8d896b93 era, APPROVE). The current, single authoritative summary for this PR is comment `5648317567` (APPROVE at head `171d4637`), which records the full consolidated disposition. This earlier note is retained only as history.

murdore:
## The body says "No live vendor endpoint was used." That is now addressed — and doing so found a hole in this guard.

Sachin chose to make the budget observable rather than accept the offline suite. Pushed as `a50d1c9fe`.

### Why two earlier live attempts proved nothing

I tried twice before to demonstrate the reclaim against a real vendor and got zero
`Reclaimed context` lines both times (Groq at 60k maxTokens; DeepSeek at ~324 KB of
tool payload). I recorded that as "the conversation never exceeded the budget."

**That conclusion was unfounded**, and the reason is the shape of failure this
repository keeps getting caught by: the guard logged **nothing** on the
non-firing path, so zero reclaim lines was equally consistent with

- the conversation fitting,
- the guard not being installed on that path,
- the hook never being called.

Three different worlds, one identical observation. An assertion about something
not happening needs a precondition proving the thing under test actually ran.

### What changed

The non-firing path now emits the budget and the headroom, and the firing path
carries the same numbers so a reader can see by how much it overshot. The
threshold is measurable from outside the process for the first time.

The probe is two-phase, and the first phase is a precondition rather than a
warm-up: measure the real budget from a small turn, **assert an "Under budget"
line appeared**, and only then size a payload from that measurement. Without that
assertion, phase two's silence would be exactly as vacuous as my earlier runs.

Phase 1, live against DeepSeek:

```
availableInputTokens=62500  estimatedPromptTokens=4810  headroom=57690
```

### What phase 2 found, which is not what I went looking for

```
Over budget, nothing reclaimable
{"provider":"deepseek","availableInputTokens":62500,
 "estimatedPromptTokens":186845,"overBudgetByTokens":124345,"calibration":1}
```

**124,345 tokens over budget, and the guard reclaimed nothing.** Not "did not need
to" — could not.

`planLoopGuardReclaim` returns `fire: effectiveTruncate.length > 0 || dropped.size > 0`,
and its mutable window is `[1, entries.length - DEFAULT_LOOP_GUARD_PROTECTED_TAIL)`
with that constant at 4. In a short conversation `maxIndexExclusive` collapses onto
`minIndex`, the window is empty, stage 1 and stage 2 both iterate over nothing, and
the plan comes back empty. A single large tool result — one big file read, one big
API response, the ordinary case — sits inside the protected tail and cannot be
touched. The request then goes to the vendor three times over the window, to be
rejected or silently truncated there.

### What I deliberately did NOT do

I have not widened the reclaim window or changed `DEFAULT_LOOP_GUARD_PROTECTED_TAIL`.
Protecting the live working set is a deliberate design choice, and changing it is a
behaviour change that needs its own tests and its own review — not something to fold
into a PR under a decision that was about observability.

What I did change is that the guard no longer does this **silently**. It now warns
with the overshoot, so an operator sees "over budget, nothing reclaimable" instead of
nothing at all.

### Honest scope of this evidence

This does **not** show the reclaim firing against a live vendor. I did not achieve
that, and I am not going to describe it as if I had. What it shows is that the guard
is installed, is called, correctly computes its budget, and has a reachable regime in
which it cannot act. That is more useful than the demonstration I set out to produce,
and it means the body's admission should be replaced with a precise statement of what
is and is not covered rather than simply deleted.

Reproduce with:

```
NEUROLINK_DEBUG=true NEUROLINK_LOG_LEVEL=debug node <probe>
```

Both variables are required — `NEUROLINK_LOG_LEVEL=debug` alone silences everything
including `info`, because `shouldLog` gates on `NEUROLINK_DEBUG` first and only then
compares levels. Worth knowing before anyone tries to reproduce this and sees nothing.

src/lib/context/nativeGenerateGuard.ts
```diff
@@ -0,0 +1,218 @@
...
+            },
+          );
+        } else {
+          logger.debug("[NativeGenerateGuard] Under budget", {
+            provider: config.provider,
+            availableInputTokens: config.availableInputTokens,
+            estimatedPromptTokens: estimatedBeforeTokens,
+            headroomTokens: config.availableInputTokens - estimatedBeforeTokens,
+            calibration,
+          });
```
Tara-ag: **MINOR — the new observability logs key off the raw `availableInputTokens`, but the guard actually fires at `availableInputTokens × 0.85 / calibration` (`DEFAULT_CONTEXT_GUARD_RATIO`), so the signal is off by up to ~15% and can mislabel the very case it was added to expose.**

`planLoopGuardReclaim` (loopGuardCore.ts) decides `fire` purely on `thresholdTokens = floor(availableInputTokens * DEFAULT_CONTEXT_GUARD_RATIO / calibration)` — it returns `fire: false` whenever `total <= thresholdTokens`. There is a substantial band between that threshold and the raw budget. For this comparison to be the meaningful split, it should be measured against the same threshold the policy uses.

Concretely, in the band `(0.85 × available / calibration, available]` the code reaches the `else` branch and logs `debug("[NativeGenerateGuard] Under budget")` with a **positive** `headroomTokens = available − estimated` — even though the conversation has already crossed the guard's own reclaim threshold and the guard would reclaim anything it could. The exact purpose of this commit (make the reclaim threshold measurable from outside so a probe can size a payload) is defeated there: an operator reading `headroomTokens` to find where reclaim engages is told it starts at 100% of budget, when it actually starts at 85%/calibration.

The `logger.warn("[…] Over budget, nothing reclaimable")` path only triggers when `estimated > available` — the extreme observed on deepseek (186k vs 62.5k). The common near-miss case that sits just over the *effective* 85% threshold and cannot reclaim (empty plan) is still silently labeled "Under budget".

Suggested fix — split on, and log, the same `thresholdTokens` figure the policy uses (derive it here rather than re-deriving inline):

```suggestion
      const thresholdTokens = Math.floor(
        (config.availableInputTokens * DEFAULT_CONTEXT_GUARD_RATIO) /
          (calibration > 0 ? calibration : 1),
      );
      // …then inside `if (!plan.fire)` replace the `availableInputTokens`
      // comparison and the headroom/warn math with `thresholdTokens`.
```

(Optionally, include both `thresholdTokens` and `availableInputTokens` in the log payload so the distinction is visible.)

No action required: the underlying reclaim logic is unaffected — only the threshold that the diagnostic label/headroom is anchored to. All three new log calls are gated behind debug mode by the logger's `shouldLog` (`logger.ts` returns false for non-`error` unless `NEUROLINK_DEBUG`/`--debug`), so production output is unchanged.

Tara-ag (commented):
Recurring review of the observability logging added in the latest head (a50d1c9fe). One new MINOR finding on the diagnostic threshold; all prior findings stand resolved/accepted.

Tara-ag:
*Superseded* round note (head a50d1c9, APPROVE). The current, single authoritative summary for this PR is comment `5648317567` (APPROVE at head `171d4637`), which records the full consolidated disposition. This earlier note is retained only as history.

Tara-ag (approved):
APPROVE — recurring review of head `a50d1c9`. The rewritten V3 adapter (`createNativeGenerateGuard`) resolves all prior MAJOR/MINOR findings; backward-compat and pairing invariants hold. One non-blocking MINOR (observability log threshold at `nativeGenerateGuard.ts:159-178`) is noted inline. Verdict: APPROVE.

murdore:
You are right, and the correction is measurable — thank you for catching it, because it defeated the exact purpose of the change. Pushed as `92fc9a45e`.

`planLoopGuardReclaim` decides `fire` on `thresholdTokens = floor(availableInputTokens * DEFAULT_CONTEXT_GUARD_RATIO / calibration)`, and my logs compared against the raw `availableInputTokens`. So in the band between the two, the guard had already crossed its reclaim point while my line read `Under budget` with a **positive** `headroomTokens`.

That is worse than a cosmetic mislabel for this particular commit. The stated purpose was to let a probe size a payload to cross the reclaim threshold — and a probe reading that headroom would aim past the point it meant to cross, by the full width of the band.

Every number now measures against the same threshold the policy uses:

```ts
const reclaimThresholdTokens = Math.floor(
  (config.availableInputTokens * DEFAULT_CONTEXT_GUARD_RATIO) /
    (calibration > 0 ? calibration : 1),
);
```

and the two branches are renamed to match what they actually test — `Under threshold` / `Over threshold, nothing reclaimable` — carrying `reclaimThresholdTokens`, `headroomToThresholdTokens` and `overThresholdByTokens`. The firing log carries the threshold too.

**Measured before and after, same probe, same provider:**

| | against raw budget | against reclaim threshold |
| --- | --- | --- |
| `availableInputTokens` | 62500 | 62500 |
| threshold reported | — | **53125** |
| headroom reported | 57690 | **48315** |

A difference of 9,375 tokens, which is exactly `62500 × 0.15` — your ~15% figure, confirmed rather than assumed.

The probe that consumes these numbers has been updated in step, and its phase-2 verdict corrected: it previously concluded "budget was measured but not crossed" in a run where the threshold demonstrably **was** crossed, which is the same class of wrong reading this whole thread is about. It now distinguishes "crossed but nothing reclaimable" from "never crossed", since only the first is a finding about the guard rather than about the probe's payload sizing.

Re-run against DeepSeek on `92fc9a45e`:

```
PHASE 1  availableInputTokens=62500  reclaimThresholdTokens=53125  headroom=48315
PHASE 2  estimatedPromptTokens=157314  overThresholdByTokens=104189  reclaimLines=0

Over threshold, nothing reclaimable
{"provider":"deepseek","availableInputTokens":62500,"reclaimThresholdTokens":53125,
 "estimatedPromptTokens":157314,"overThresholdByTokens":104189,"calibration":1}
```

`pnpm run lint` clean, 0 prettier warnings.

src/lib/context/nativeGenerateGuard.ts
Tara-ag: `<!-- yama:generate-guard-observability-threshold -->` This finding is now addressed in the committed code (pushed 2026-09-12T07:47): `guardConversation` derives `reclaimThresholdTokens = floor(availableInputTokens × DEFAULT_CONTEXT_GUARD_RATIO / calibration)` and anchors both the "Over threshold, nothing reclaimable" split and the `headroomToThresholdTokens` / `overThresholdByTokens` log fields to it, not the raw budget. I verified the branch is reachable: `planLoopGuardReclaim` returns `fire:false` both when the request fits AND when it is over threshold but nothing lies in the mutable window (the deepseek 186k-vs-62.5k/empty-plan case), so the warning still fires for the near-miss the finding was about. Resolving.

Tara-ag:
*Superseded* round note (head 92fc9a45, APPROVE). The current, single authoritative summary for this PR is comment `5648317567` (APPROVE at head `171d4637`), which records the full consolidated disposition. This earlier note is retained only as history.

Tara-ag (approved):
APPROVE — recurring review of current head `92fc9a45e`. The V3 adapter (`createNativeGenerateGuard`) resolves all earlier MAJOR/MINOR findings; backward-compat (Rule 5), pairing invariants, and observability-threshold anchoring all verified. The last open MINOR (observability threshold) is fixed and its thread resolved. No CRITICAL/MAJOR/MINOR defect remains. Full disposition in summary comment `5644600538`.

src/lib/providers/openaiChatCompletionsBase.ts
```diff
@@ -1189,6 +1191,19 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
...
+          ...createNativeGenerateGuard({
+            provider: this.providerName,
+            availableInputTokens: getAvailableInputTokens(
+              this.providerName,
+              modelId,
+              options.maxTokens ?? undefined,
+            ),
+            getFixedOverheadTokens: () =>
+              estimateTokens(
+                JSON.stringify({ tools: v3Tools, responseFormat: format }),
```
Tara-ag: `

**MINOR — `getFixedOverheadTokens` uses a raw `JSON.stringify`, which every other token estimator in this codebase deliberately wraps, and this one now runs on the hot path.**

`estimateTokens(JSON.stringify({ tools: v3Tools, responseFormat: format }), …)` can throw synchronously on a cyclic schema or a value past V8's max string length (`TypeError`/`RangeError`). Because `getFixedOverheadTokens()` is invoked inside `guardConversation` on **every step** of the generate loop, a throw here aborts the whole `generate()` mid-turn rather than degrading the estimate.

This is exactly the hazard the rest of the codebase already guards: `safeStringify` (`openaiChatCompletionsClient.ts`), `serializeForEstimate` (`tokenEstimation.ts`), and `textOf` in `nativeGenerateGuard.ts` itself all wrap `JSON.stringify` in try/catch, and the streaming guard's equivalent `estimateWireTokens` goes through `safeStringify`. This new path is the one place a tool schema could break the loop.

Fix — import the exported `safeStringify` and use it here (identical estimate, no throw):

```suggestion
            getFixedOverheadTokens: () =>
              estimateTokens(
                safeStringify({
                  tools: v3Tools,
                  responseFormat: format,
                }),
                this.providerName,
              ),
```

If `v3Tools` is empty/undefined, `safeStringify({tools: undefined, responseFormat: undefined})` returns `"{}"`, the same string `JSON.stringify` produces — so the estimate is byte-identical in the common case.

Tara-ag (commented):
Approve with one non-blocking MINOR on the fixed-overhead estimator. All prior findings resolved/accepted.

Tara-ag:
*Superseded* round note (head 8d896b93, APPROVE). The current, single authoritative summary for this PR is comment `5648317567` (APPROVE at head `171d4637`), which records the full consolidated disposition. This earlier note is retained only as history.

Tara-ag (approved):
APPROVE — validating review-state consistency at head `8d896b93` (single commit `feat(native-loop): reclaim V3 conversations without breaking tool pairs`).

Current disposition (authoritative summary comment `5645800127`): all prior MAJOR/MINOR findings are Resolved or Accepted via adequate author rebuttals; the only outstanding item is a single non-blocking MINOR (`getFixedOverheadTokens` raw `JSON.stringify` at `openaiChatCompletionsBase.ts:1203`, inline thread open) with a one-click suggested fix. No CRITICAL/MAJOR defect remains.

This review records the APPROVE verdict so the PR's review state matches the standing disposition.

src/lib/providers/openaiChatCompletionsBase.ts
murdore: The hazard is real and I have fixed it at `d7c3f40be` — but not with `safeStringify`, because at this particular call site its fallback points the wrong way.

`safeStringify` (openaiChatCompletionsClient.ts) falls back to `String(value ?? "")`. For a cyclic tool schema that is `"[object Object]"` — about 4 tokens. So instead of throwing, the guard would size the entire tool overhead at roughly nothing and quietly **under**-fire. A silent wrong estimate on the budget path is worse than the throw it replaces.

The two estimators already on this path both go the other way, and deliberately:

```ts
// tokenEstimation.ts serializeForEstimate, and textOf in nativeGenerateGuard.ts
try { return JSON.stringify(value) ?? ""; }
catch { return "x".repeat(200_000); }
```

An estimator that cannot measure a value must over-report, so the guard fires rather than waving through content it failed to size. Rather than add a fourth `safeStringify` variant to the four already in the tree, I exported the canonical `serializeForEstimate` from `tokenEstimation.ts` and used it here, with a comment recording why the fallback is large.

Verified against the built bundle:

```
BEFORE raw JSON.stringify -> threw: true TypeError
AFTER  serializeForEstimate -> chars: 200000 tokens: 52500
common-case parity: "{}" equals {} ? true
```

So: no throw, the unmeasurable case pushes the estimate up rather than down, and the ordinary empty case is byte-identical to what was there before — no change to any real estimate.

Tara-ag:
*Superseded* round note (head d7c3f40, APPROVE). The current, single authoritative summary for this PR is comment `5648317567` (APPROVE at head `171d4637`), which records the full consolidated disposition. This earlier note is retained only as history.

Tara-ag (approved):
## APPROVE — validating review-state consistency at head `d7c3f40be`

Recurring review of the current head (single commit `feat(native-loop): reclaim V3 conversations without breaking tool pairs`). All prior findings across every round are Resolved or Accepted on the merits:

- The MAJOR wire-shape mismatch and its MINORs (calibration shape, test coverage) were resolved by the `createNativeGenerateGuard` V3 adapter.
- The `observeUsage` calibration-reset MINOR was accepted via the author's sound rebuttal.
- The observability-threshold MINOR was fixed in committed code (`reclaimThresholdTokens`).
- The last-open MINOR (`getFixedOverheadTokens` raw `JSON.stringify` at `openaiChatCompletionsBase.ts:1203`) was fixed by the author at `d7c3f40be` using the over-reporting `serializeForEstimate` — the correct failure mode — and its review thread (`PRRT…hv68h`) is resolved.

No CRITICAL/MAJOR/MINOR defect remains open. This review records the APPROVE verdict on the current head so the PR's review state matches the standing disposition in summary comment `5645903000`.

Tara-ag:
## APPROVE — recurring review; independently re-verified head `171d4637`, no new defect

This round re-verified the current head (`171d4637a45`, single conventional commit `feat(native-loop): reclaim V3 conversations without breaking tool pairs`) directly against source — `src/lib/context/nativeGenerateGuard.ts`, `src/lib/core/nativeGenerateLoop.ts`, `src/lib/providers/openaiChatCompletionsBase.ts`, plus the shared policy `loopGuardCore.ts`, `toolOutputLimits.ts`, `tokenEstimation.ts`, `core/constants.ts` — and the code graph. All prior review threads are resolved or accepted on the merits, and I found **no new CRITICAL/MAJOR/MINOR defect** in the committed code. No new inline comments were warranted.

### Disposition of prior findings (all resolved/accepted — not reposted)

| Severity | Original finding | Disposition |
| --- | --- | --- |
| MAJOR | Guard fed the v3 message-builder shape but classified by the OpenAI wire `tool_calls` field → orphaned `tool_calls` on stage-2 drop (provider 400) | **Resolved.** `createNativeGenerateGuard` classifies assistant `content` parts of `type:"tool-call"` as `toolCall` and `role:"tool"` + `tool-result` parts as `toolResult`; `findToolBatches` (loopGuardCore.ts) drops each call→result batch whole, so no orphan. `conv` on the generate path is `buildMessagesForStream()` output (kept in v3 message-builder shape for `doGenerate` to convert), which is exactly the shape the adapter expects — I confirmed the streaming path alone applies `messageBuilderToOpenAI`. |
| MINOR | Calibration read only a bare-number `usage.inputTokens` | **Resolved.** `observeUsage` reads a numeric `inputTokens` or the shaped `{ total }` form the loop's `readTotal` accepts; calibration is a same-request ratio (start-of-step `previousSentEstimate` ↔ end-of-step `observedPromptTokens`). |
| MINOR | Tests only exercised stage-1 truncation; no pairing validation | **Resolved.** Batch-dropping (`count=8`, large call args / small results) + calibrated-usage tests assert `tool_calls`/`tool_call_id` pairing on every outbound request and that the task prefix + newest batch survive. |
| MINOR | `observeUsage` resets calibration to 1.0 on a sparse-usage step | **Accepted** — rebuttal sound: `previousSentEstimate` advances per request, so retaining an older numerator would pair counts from different requests into a false ratio; fallback-to-1 is the conservative safe direction and mirrors the streaming guard. |
| MINOR | Observability logs anchored to raw `availableInputTokens`, but the guard fires at `availableInputTokens × 0.85 / calibration` | **Resolved.** `guardConversation` derives `reclaimThresholdTokens = floor(available × DEFAULT_CONTEXT_GUARD_RATIO / calibration)` and anchors the `Under threshold` / `Over threshold, nothing reclaimable` split and the `headroomToThresholdTokens` / `overThresholdByTokens` fields to it. |
| MINOR | `getFixedOverheadTokens` used a raw `JSON.stringify` that could throw mid-loop on a cyclic/oversized tool schema | **Resolved.** `tokenEstimation.ts` now exports `serializeForEstimate` (fallback over-reports → guard fires rather than under-sizes) and `openaiChatCompletionsBase.ts` uses it on the generate path. |

### Re-verified this round (graph + source)

- **Shape contract confirmed:** the generate path keeps `conv = buildMessagesForStream(...)` in the message-builder/v3 shape (no `messageBuilderToOpenAI`), so the adapter's `partsOf(content)` classification of `tool-call`/`tool-result` content parts is the correct mapping; the loop appends `{role:"assistant", content: parts}` and `{role:"tool", content: resultParts}` which the adapter consumes directly.
- **Backward compat (Rule 5):** `observeUsage`/`guardConversation` are optional on `NativeGenerateLoopArgs`. `runNativeGenerateLoop` has exactly 3 callers — OpenAI `runLoop` (wired) and Anthropic + SageMaker native loops, which pass no callbacks → optional-chain no-ops, byte-identical. `NativeGenerateGuardConfig` is a new additive type.
- **Rule 1:** `createNativeGenerateGuard` has exactly 1 caller (OpenAI `runLoop`); `nativeGenerateGuard.ts` is a static-import module (types, tokenEstimation, logger, loopGuardCore, toolOutputLimits) — no provider-registry cycle, no `formatProviderError` throw on this path.
- **Index mapping:** `entries = [prefix, ...tail.map]`; rebuild maps `entry = index + 1 = tail[index]` — no off-by-one vs the policy's `drop`/`truncate` indices.
- **Tool-pairing integrity:** stage-2 only drops whole call→result units (`findToolBatches` scans `toolCall`-then-`toolResult` runs and breaks at the protected-tail boundary); `previewMessage` preserves `toolCallId`/output envelopes via `...part` on stage-1 truncation. New tests assert pairing on every request.
- **Task/tail protection:** entry 0 (folded task prefix) never dropped/truncated; `findToolBatches` breaks at `maxIndexExclusive`, so a batch straddling the protected tail is never partially dropped.
- **Cache-invalidation discipline:** guard returns `undefined` when `!plan.fire`; the loop splices only on a truthy replacement (`nativeGenerateLoop.ts`) — a fitting turn never pays a prompt-cache invalidation.
- **Guard lifetime:** `createNativeGenerateGuard` is constructed fresh inside `runLoop`, so a re-loop starts a fresh `previousSentEstimate`/`observedPromptTokens` — no stale-calibration carryover.
- **Logging:** debug/warn-gated by `shouldLog`; carries only provider / token counts / calibration — no secret or sensitive content, no hot-path throw.

### Clean / noted
- Single conventional commit — Single Commit Policy compliant.
- Docs regenerated (`docs/api`, `llms-full.txt`).
- The guard retains its designed, acknowledged unreclaimable regime (a single large tool result inside the protected tail) — now surfaced by the `warn` rather than silence; a real follow-up with its own change + tests, deliberately out of scope here.

**Verdict: APPROVE.** No blocking issue remains; all prior findings are resolved or accepted on the merits, and the current head `171d4637` introduces no new defect. (This updates the prior `yama:summary` for head `b12c332`; it remains the single authoritative summary for the PR.)

Tara-ag (approved):
## APPROVE — review-state sync at head `b12c332ae`

Validating the review state against the standing disposition. The current PR head is the single squash commit `b12c332ae` (`feat(native-loop): reclaim V3 conversations without breaking tool pairs`); prior approving reviews were anchored to intermediate heads, so this review formally records the APPROVE verdict on the current head so the PR's review state matches the disposition in summary comment `5648317567`.

**Disposition — all prior findings resolved/accepted on the merits (none reopen):**

| Severity | Finding | Disposition |
| --- | --- | --- |
| MAJOR | Guard fed v3 message-builder shape but classified by OpenAI wire `tool_calls` → orphaned `tool_calls` on stage-2 drop (provider 400) | **Resolved.** `createNativeGenerateGuard` classifies assistant `content` `tool-call` parts as `toolCall` and `role:"tool"` `tool-result` parts as `toolResult`; `findToolBatches` drops whole call→result units. |
| MINOR | Calibration read only a bare-number `usage.inputTokens` | **Resolved.** `observeUsage` reads numeric or shaped `{total}`. |
| MINOR | Tests only exercised stage-1 truncation, no pairing validation | **Resolved.** Batch-dropping (`count=8`) + calibrated-usage tests assert `tool_calls`/`tool_call_id` pairing on every outbound request. |
| MINOR | `observeUsage` reset calibration to 1.0 on a sparse-usage step | **Accepted** — rebuttal sound (same-request numerator/denominator; fallback-to-1 is the conservative safe direction, mirrors streaming guard). |
| MINOR | Observability logs anchored to raw budget, not the reclaim threshold | **Resolved.** `reclaimThresholdTokens = floor(available × 0.85 / calibration)` anchors the Under/Over split and log fields. |
| MINOR | `getFixedOverheadTokens` raw `JSON.stringify` could throw mid-loop | **Resolved.** Exported `serializeForEstimate` (over-reporting fallback) used on the hot path. |

**Verified clean this round:** all 11 inline review threads are resolved; single conventional commit (Single Commit Policy compliant); docs regenerated; logging debug/warn-gated with no sensitive content. No CRITICAL/MAJOR/MINOR defect remains open. Verdict: **APPROVE.**

Tara-ag:
## Review-state validation — clean

I re-validated the full PR review surface at the current head `171d4637a45`:

### Comment integrity
- **All 11 inline threads are resolved** and render as clean markdown with valid diff anchors. No malformed blocks, no broken suggestion fence. Threads `yama:generate-guard-observe-calibration-reset` and `yama:native-generate-observe-usage-reset` are one finding; the former is self-resolved and points at the canonical thread (`#discussion_r3961588683`).
- **Summary:** there is now exactly **one** `yama:summary` comment — `5648317567` (APPROVE at head `171d4637`), the single authoritative disposition. The 13 older stacked summaries were relabelled `yama:historical-round` / superseded so no duplicate summary marker remains.

### Review state
- The verdict is **APPROVE**, now formally recorded on the current head via review **#5189… (APPROVED)** at commit `171d4637a4570c82e82a839eecc229a90a1ee280`. Prior approvals (latest at `b12c332ae`) were anchored to intermediate heads; the review state now matches the disposition on the commit the PR actually carries.
- Single conventional commit; Single Commit Policy compliant; all required checks green.

No further comments needed a reply or a fix.

Tara-ag (approved):
## APPROVE — review-state sync on the current head `171d4637`

The standing verdict is **APPROVE** (summary comment `5648317567`). Prior approving reviews were anchored to intermediate heads (latest at `b12c332ae`); the PR now carries the single squash commit `171d4637a4570c82e82a839eecc229a90a1ee280` (`feat(native-loop): reclaim V3 conversations without breaking tool pairs`), so this review formally records the APPROVE verdict on the commit the PR actually carries.

All 11 inline review threads are resolved/accepted on the merits; no CRITICAL/MAJOR/MINOR defect remains open. Single conventional commit — Single Commit Policy compliant. Verdict: **APPROVE**.

murdore:
Two follow-ups from this PR, one moved to **https://github.com/juspay/neurolink/pull/1728** and one deliberately not.

**Moved:** CodeRabbit's pre-merge check reported docstring coverage of 50% against an 80% threshold but named no file, so nothing actionable ever landed. The two undocumented members are `observeUsage` and `guardConversation` in `src/lib/context/nativeGenerateGuard.ts` — the ones carrying the logic. Both are documented in #1728.

**Not moved:** your own comment of 2026-09-12 said the body's blanket *"No live vendor endpoint was used"* should be replaced with a precise statement of what is and is not covered. It never was — GraphQL `userContentEdits` shows the body's last substantive edit predates that comment. Since a PR body does not survive the merge, there is nothing durable to carry; recording it here so the intent is at least on file.

Neither was ever an unresolved thread — one was a bot summary, the other an issue comment. That is why they survived a board that reported zero unresolved.

PR #1669, opened 2026-09-08
fix(sagemaker): adapt legacy results for the native generate loop

murdore:
## What

Adapt SageMaker's legacy `doGenerate()` result at the provider's native-loop boundary. The shared loop expects V3 `content` parts and shaped token usage; SageMaker returned top-level `text`, `toolCalls`, and numeric usage. As a result, text disappeared and tools never executed.

The adapter preserves text, reasoning strings, tool-call IDs/arguments, and endpoint usage. It translates `maxOutputTokens` and JSON schema options to SageMaker's existing request contract, and preserves explicit zero sampling values. The low-level result and existing streaming path are unchanged.

## Regression proof

Tests construct `NeuroLink` from `dist/index.js`, use the real AWS client with `credentials.sagemaker.endpoint` pointing at a local HTTP server, and assert that the endpoint actually received requests.

New cases were red before the fix:
- Returned text and numeric usage survive; max tokens and temperature zero reach the endpoint.
- Tool calls execute once, their IDs/results appear in the next request, and usage accumulates across steps.
- The requested JSON schema reaches the endpoint and parsed `structuredData` reaches the caller.

After the fix: **9/9**, including the six existing streaming cases. No live AWS endpoint was used; this proves the SDK boundary and HTTP transport contract, not compatibility with every deployed model container.

## Review follow-up

The tool-only response extractor previously fell back to serializing the whole endpoint envelope as assistant text. The extractor now returns an empty text for tool-only replies while retaining genuine text returned alongside tool calls. The regression checks the next request for absence of the raw `choices` envelope in addition to tool execution, call-ID/result replay, and token accounting. The strengthened assertion failed before this fix and the suite passes 9/9 after it.

All remaining merges are reserved to the repository owner.

### DIFF
--- a/src/lib/providers/amazonSagemaker.ts
+++ b/src/lib/providers/amazonSagemaker.ts
@@ -209,7 +209,71 @@ export class AmazonSageMakerProvider extends BaseProvider {
         new Error("sagemaker: model handle exposes no doGenerate()"),
       );
     }
-    const doGenerate = model.doGenerate.bind(model);
+    const doGenerate = async (
+      call: Record<string, unknown>,
+    ): Promise<Record<string, unknown>> => {
+      const format = call.responseFormat as Record<string, unknown> | undefined;
+      const result = await model.doGenerate({
+        ...call,
+        ...(call.maxOutputTokens !== undefined
+          ? { maxTokens: call.maxOutputTokens }
+          : {}),
+        ...(format?.type === "json"
+          ? {
+              responseFormat: format.schema
+                ? {
+                    type: "json_schema",
+                    json_schema: { name: "response", schema: format.schema },
+                  }
+                : { type: "json_object" },
+            }
+          : {}),
+      });
+      // SageMaker's low-level model retains its legacy result for streaming
+      // and direct consumers. The shared loop consumes V3 content and usage.
+      if (Array.isArray(result.content)) {
+        return result;
+      }
+      const content: Array<Record<string, unknown>> = [];
+      if (typeof result.text === "string" && result.text) {
+        content.push({ type: "text", text: result.text });
+      }
+      if (typeof result.reasoning === "string" && result.reasoning) {
+        content.push({ type: "reasoning", text: result.reasoning });
+      }
+      for (const call of Array.isArray(result.toolCalls)
+        ? result.toolCalls
+        : []) {
+        if (typeof call !== "object" || call === null) {
+          continue;
+        }
+        const item = call as Record<string, unknown>;
+        const fn = item.function as Record<string, unknown> | undefined;
+        if (typeof item.id === "string" && typeof fn?.name === "string") {
+          content.push({
+            type: "tool-call",
+            toolCallId: item.id,
+            toolName: fn.name,
+            input: fn.arguments ?? "{}",
+          });
+        }
+      }
+      const usage = result.usage as Record<string, unknown> | undefined;
+      return {
+        ...result,
+        content,
+        usage: {
+          inputTokens: {
+            total:
+              typeof usage?.inputTokens === "number" ? usage.inputTokens : 0,
+          },
+          outputTokens: {
+            total:
+              typeof usage?.outputTokens === "number" ? usage.outputTokens : 0,
+          },
+        },
+      };
+    };
 
     const shouldUseTools = !options.disableTools && this.supportsTools();
     const toolsRecord = shouldUseTools
--- a/src/lib/providers/sagemaker/language-model.ts
+++ b/src/lib/providers/sagemaker/language-model.ts
@@ -588,8 +588,8 @@ export class SageMakerLanguageModel implements SageMakerAsLanguageModel {
         ...(options.maxTokens !== undefined
           ? { max_new_tokens: options.maxTokens }
           : {}),
-        temperature: options.temperature || 0.7,
-        top_p: options.topP || 0.9,
+        temperature: options.temperature ?? 0.7,
+        top_p: options.topP ?? 0.9,
         stop: options.stopSequences || [],
       },
     };
@@ -804,8 +804,13 @@ export class SageMakerLanguageModel implements SageMakerAsLanguageModel {
     // Handle response with tool calls
     if (responseBody.choices && Array.isArray(responseBody.choices)) {
       const choice = responseBody.choices[0];
-      if (choice?.message?.content) {
+      if (typeof choice?.message?.content === "string") {
         return choice.message.content;
+      }
+      // A tool-only assistant turn has no text. Serializing the entire
+      // endpoint response here would replay its envelope as assistant prose.
+      if (choice?.message?.tool_calls?.length) {
+        return "";
       }
     }
 
--- a/test/continuous-test-suite-sagemaker-streaming.ts
+++ b/test/continuous-test-suite-sagemaker-streaming.ts
@@ -23,6 +23,7 @@ import "dotenv/config";
  */
 
 import { createServer, type Server } from "node:http";
+import { z } from "zod";
 import { assert, defineSuite } from "./helpers/harness.js";
 import { assertDistFresh } from "./helpers/distFreshness.js";
 
@@ -30,7 +31,7 @@ assertDistFresh();
 
 const { test, runSuite } = defineSuite("SageMaker streaming");
 
-const { NeuroLink } = await import("../dist/index.js");
+const { NeuroLink, tool } = await import("../dist/index.js");
 
 type StandIn = {
   requests: number;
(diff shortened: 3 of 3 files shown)


src/lib/providers/amazonSagemaker.ts
```diff
@@ -209,7 +209,71 @@ export class AmazonSageMakerProvider extends BaseProvider {
...
+            }
+          : {}),
+      });
+      // SageMaker's low-level model retains its legacy result for streaming
+      // and direct consumers. The shared loop consumes V3 content and usage.
+      if (Array.isArray(result.content)) {
+        return result;
+      }
+      const content: Array<Record<string, unknown>> = [];
+      if (typeof result.text === "string" && result.text) {
```
Tara-ag, line 238: **MAJOR — tool-call turns leak the whole raw response as a `text` part into the assistant message.**

`SageMakerLanguageModel.extractTextFromResponse` (language-model.ts) falls back to `JSON.stringify(responseBody)` when an OpenAI-compatible endpoint returns `{ choices: [{ message: { content: null, tool_calls: [...] } }] }` — `choice.message.content` is `null`, so the `if` is skipped and the entire third-party response is returned as `generatedText`. For exactly the shape this PR's own tool test simulates, `result.text` is therefore a non-empty JSON blob, and this new branch pushes `{ type: "text", text: <stringified response> }` into `content` alongside the real `tool-call` parts.

The native loop then writes that assistant message (`args.conversation.push({ role: "assistant", content: parts })`) back into the next request. So on the tool re-try the model is fed its own raw response as assistant text — token bloat on multi-step loops, and the model can be steered by it. The regression test only asserts `raw.includes("SAGEMAKER_TOOL_RESULT")` and `raw.includes("sagemaker_call")`, so the extra text part goes undetected.

A tool-call turn is not a text turn: gate the text part on there being no tool calls.

```suggestion
      const hasToolCalls =
        Array.isArray(result.toolCalls) && result.toolCalls.length > 0;
      if (!hasToolCalls && typeof result.text === "string" && result.text) {
        content.push({ type: "text", text: result.text });
      }
```

(An additional smaller fix: `extractTextFromResponse` should not fall back to `JSON.stringify` when the payload is a tool-call response — but gating here fixes the native loop regardless.)

Tara-ag (changes requested):
Niobium findings below; the adapter works end-to-end but the tool-call shape leaks raw JSON into the assistant turn.

Tara-ag:
## Review: fix(sagemaker) — adapt legacy results for the native generate loop

**Verdict: NEEDS_WORK** — the adapter correctly translates the legacy SageMaker result into the V3 `content`/`usage` the shared loop consumes (text, usage, tool calls, structured output all survive; the `??` temperature fix correctly lets an explicit `0` reach the endpoint). One MAJOR defect in the new adapter needs fixing before merge.

| MAJOR | `src/lib/providers/amazonSagemaker.ts:238` | On a tool-call turn from an OpenAI-compatible SageMaker endpoint, `extractTextFromResponse` returns the **stringified raw response** (content is `null`), and the adapter pushes it as a `text` part — which the native loop then writes back as assistant content on the re-try. Bloat + model steering on multi-step tool loops; the regression test's endpoint shape reproduces it but never asserts it. Fix: gate the text part on `!hasToolCalls` (suggestion inline). |

### Checked and clean
- **Legacy→V3 text/usage bridging** (`amazonSagemaker.ts`): `content` parts and `{inputTokens:{total}, outputTokens:{total}}` shape match exactly what `runNativeGenerateLoop` reads via `readTotal`. Verified end-to-end against the loop's consumption of `content`, `finishReason`, and `usage`.
- **Tool-call mapping**: `item.id`/`fn.name`/`fn.arguments` → `tool-call` part round-trips through `parseToolInput` in the loop (JSON-string arguments parse correctly; `toolCallId` preserved for the replayed result).
- **`maxOutputTokens`→`maxTokens` mapping**: the loop passes `maxOutputTokens`, the adapter translates it to SageMaker's `options.maxTokens` before `convertToSageMakerRequest` reads it. Correct.
- **structured-output path**: `responseFormat` (`json`+schema on the loop side) is correctly translated to SageMaker's `json_schema` wire shape; `result.object`/`structuredData` reached the caller (test-covered).
- **`temperature`/`top_p` `||`→`??` fix** (`language-model.ts`): correctly preserves explicit `0` sampling values. Good catch.
- **Blast radius**: the only low-level result the adapter touches is `SageMakerLanguageModel.doGenerate`'s output; the streaming path (`doStream`) and low-level result are untouched. `BaseProvider`'s native-loop expectations are met.
- **Tests** (`continuous-test-suite-sagemaker-streaming.ts`): assert the endpoint actually received the request (`startStandIn` now reads the body), usage, sampling options, tool execution count, replay identity, and structured data. Good regression coverage.

### Not raised as findings
- **`topP` gap**: the native loop args never forward `options.topP`, so a caller-set `topP` is silently dropped on the native path. This is pre-existing (not introduced here) and not a changed line; flagging for awareness, not blocking.

Impact analysis note: the code-review-graph is unavailable in this environment (filesystem restricted to `.yama/`), so blast radius was traced manually from the checkout via GitHub source instead of the graph.

src/lib/providers/amazonSagemaker.ts
murdore: Fixed at b292d182788fb2add3488d858009ff9db13b12a1 in extractTextFromResponse: tool-only replies return empty text instead of serializing the endpoint envelope. Genuine text alongside tool calls is still preserved. The public SDK regression now asserts that the next request does not contain the raw choices envelope, in addition to execution, replay IDs/results and usage. The strengthened assertion was red before the fix; the suite is now 9/9. check:tools-tests also passes.

Tara-ag: Verified at `b292d18` — the MAJOR finding is resolved and I'm accepting it.

`extractTextFromResponse` now returns `""` for a tool-only reply (`content: null` + `tool_calls`), so `result.text` is `""` and the adapter's gate `typeof result.text === "string" && result.text` (amazonSagemaker.ts:238) excludes it — the stringified `choices` envelope can no longer be pushed as a `text` part or replayed as assistant prose on the tool re-try. Genuine non-empty text alongside tool calls is still returned by the first `typeof content === "string"` branch. The strengthened regression (`!leakedEnvelope`, i.e. the next request does not contain the raw `choices` envelope) plus execution count, call-ID/result replay and multi-step usage all hold. I also traced the adapter output contract against `runNativeGenerateLoop` — `input` (JSON string) round-trips through `parseToolInput`, and the `{inputTokens:{total}, outputTokens:{total}}` usage shape is read correctly via `readTotal`.

No new issues found in this revision. Verdict for this PR: **APPROVE**.

Tara-ag:
**Recurring review update** (supersedes the earlier `NEEDS_WORK` in the summary above).

The single MAJOR finding (`sagemaker-stray-text-part`) has been fixed and is **accepted** — see the reply on [that thread](#discussion_r3961791231). `extractTextFromResponse` now returns `""` for tool-only replies, so the stringified `choices` envelope can no longer leak in as assistant `text`; the test now asserts `!leakedEnvelope` on the next request and the suite passes 9/9.

Checked and confirmed clean in this revision:
- **Adapter → loop contract** (`amazonSagemaker.ts` `doGenerate`): `content` parts (`text`/`tool-call`/`reasoning`) and `{inputTokens:{total}, outputTokens:{total}}` match exactly what `runNativeGenerateLoop` consumes via `asParts`, `readTotal`, and `parseToolInput` (JSON-string `input` round-trips).
- **`maxOutputTokens`→`maxTokens`** and `json_schema` structured-output mapping reach the SageMaker request.
- **`temperature`/`top_p` `||`→`??`** preserves an explicit `0` (test asserts `temperature === 0` reaches the endpoint).
- Blast radius limited to the native non-streaming path; `doStream`/low-level result unchanged; 11 affected flows all trace to `AmazonSageMakerProvider`, no out-of-diff regressions.

Verdict: **APPROVE**.

Tara-ag (approved):
**APPROVE** — the MAJOR finding (`sagemaker-stray-text-part`) is resolved at `b292d18` and verified.

`extractTextFromResponse` now returns `""` for tool-only replies instead of serializing the endpoint envelope, and genuine text returned alongside tool calls is preserved (language-model.ts: `typeof choice?.message?.content === "string"` + `tool_calls?.length` gate). The adapter's gate at amazonSagemaker.ts:238 (`typeof result.text === "string" && result.text`) drops the empty text; the strengthened regression (`!leakedEnvelope`) was red before the fix and the suite passes 9/9.

Blast radius is unchanged from the prior pass: `runNativeGenerateLoop` consumption of `content`/`usage`/tool-args matches the adapter output; the streaming path and low-level result are untouched.

src/lib/providers/amazonSagemaker.ts
Tara-ag: Confirmed fixed — accepting. `extractTextFromResponse` now returns empty text for tool-only replies (no longer `JSON.stringify` of the endpoint envelope), and the amazonSagemaker.ts gate (`typeof result.text === "string" && result.text`) drops that empty text so no stray text part reaches the assistant message. Genuine text alongside tool calls is preserved by the first branch. The strengthened `!leakedEnvelope` assertion you added is anchored to exactly the failure mode originally flagged, and the suite passing 9/9 is consistent with my review of the change. Thread resolved; no further action needed.

PR #1670, opened 2026-09-08
fix(anthropic): reclaim context during native generate tool loops

murdore:
## What

Wire the shared V3 conversation reclaim adapter into Anthropic `generate()`. This is separate from streaming tool-cache work and from the OpenAI-compatible guard PR.

The guard counts the system/task conversation plus tool/schema overhead, previews only old result payloads, and drops complete call/result exchanges. It uses the same `loopGuardCore` policy as streaming without duplicating Anthropic wire-message rebuilding. Call IDs, multiple tool results per turn, task/system prompts, and the recent tail are preserved.

## Proof

A new public-SDK suite drives a local Anthropic Messages endpoint with eight rounds of two tools each. It validates tool-use/result pairing on every outbound request.

Baseline: **1 passed / 2 failed** (fitting history stayed intact; output preview and batch reclaim were missing).
Fixed: **3/3** (preview, whole-batch dropping, and no-change when history fits).

No live Anthropic endpoint was used. This depends on the V3 guard adapter from https://github.com/juspay/neurolink/pull/1668; the PR will target release once that dependency lands.

## Merge dependency

This PR intentionally contains one Anthropic-specific commit above PR 1668. The repository's push-time single-commit check counts both commits against `release`, so it currently reports two commits. Do not squash the dependency into this PR or bypass that check. After the owner merges PR 1668, rebase this branch onto `release`, retarget the PR, and rerun its required checks. All merges are reserved to the owner.

## Summary by CodeRabbit

- **Improvements**
  - Improved Anthropic generation’s handling of available token capacity, including tool definitions and response formatting.
  - Better preserves relevant conversation history and supports reliable tool-use interactions during multi-step requests.
  - Improved handling of requests that approach model context limits.

- **Bug Fixes**
  - Reduced the risk of incomplete, invalid, or improperly paired tool responses during Anthropic generation.


src/lib/providers/openaiChatCompletionsBase.ts
```diff
@@ -1212,6 +1217,27 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
...
+            // Runs on every step of the generate loop, so a cyclic or
+            // oversized tool schema must not throw the turn away. Shares the
+            // estimator's own serializer: its fallback over-reports, which
+            // makes the guard fire rather than silently size the overhead at
+            // nothing.
+            getFixedOverheadTokens: () =>
+              estimateTokens(
+                serializeForEstimate({
+                  tools: v3Tools,
+                  responseFormat: format,
```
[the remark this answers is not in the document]
murdore: Confirmed and fixed, but not here — in #1730, against `release`.

This code reached #1670 only through its parent #1668, which has since merged. After rebasing #1670 onto the new `release` tip and dropping the parent's orphaned commit, `openaiChatCompletionsBase.ts` is no longer in this PR's diff at all; #1670 is now just `anthropic/client.ts` and its suite. The bug is real and live on `release`, so folding it in here would have widened an unrelated PR.

The finding is right on both halves. Three call sites pass `withTools: false`, and the overhead serialized `v3Tools` regardless, so the tool-free pass was billed for a schema it never sent. Measured on a fixture, same conversation and same build, only the overhead differing:

```
before   overhead=12010  total=59941  threshold=57800  -> fires, 5 tool exchanges cut to 2
after    overhead=41     total=47972  threshold=57800  -> does not fire, all 5 retained
```

`msgTokens` is 47931 in both.

One correction to the framing: it is more than avoidable overhead. Because the number feeds `planLoopGuardReclaim`, the inflation alone crosses the reclaim threshold, so the pass silently loses older tool exchanges rather than just overestimating.

Worth recording that my first fixture proved nothing — before and after came out identical, because the conversation had already been trimmed during the tool loop, where tools genuinely are sent. The suite in #1730 now asserts no tool-bearing request was reclaimed, so that confusion cannot recur, and a control case over the threshold asserts reclaim still fires.

test/continuous-test-suite-anthropic-generate-guard.ts
```diff
@@ -0,0 +1,176 @@
...
+          },
+      );
+      assert.ok(
+        JSON.stringify(bodies[1]).includes("first_0"),
+        "precondition: original tool call never reached endpoint",
+      );
+      assert.ok(
+        JSON.stringify(bodies[1]).includes("OUTPUT_0"),
+        "precondition: original result never reached endpoint",
+      );
```
Tara-ag: **MINOR** — these two `assert.ok(...)` precondition messages are inverted relative to what they actually assert.

`assert.ok(stringify(bodies[1]).includes("first_0"))` passes when the tool call **is** present, yet the message reads "precondition: original tool call *never reached* endpoint". Same for `OUTPUT_0` below. The diagnostic text contradicts the check, which will mislead the next maintainer debugging a guard failure here.

```suggestion
      assert.ok(
        JSON.stringify(bodies[1]).includes("first_0"),
        "precondition: original tool call is present in 2nd request",
      );
      assert.ok(
        JSON.stringify(bodies[1]).includes("OUTPUT_0"),
        "precondition: original result is present in 2nd request",
      );
```

Tara-ag (approved):
**APPROVE** — clean, focused Anthropic context-guard wiring with strong e2e coverage. One MINOR (inverted precondition diagnostic) is non-blocking.

## What I validated

- **Diff is small and well-scoped**: `package.json` gains the `test:anthropic-generate-guard` script; `anthropic/client.ts` adds the `createNativeGenerateGuard` import and wires it into `runNativeGenerateLoop` in `executeNativeGenerate`; one new continuous suite.
- **Guard wiring** (`client.ts`): the guard is instantiated once per call with `filter: hasTools`, the `task`/`system` wrapped, `maxAvailableInput` derived via `getAvailableInputTokens("anthropic", modelId, options.maxTokens ?? undefined)`, and `filterCandidates`/`splitCandidates` reference the loop's `v3Tools` / `v3Messages` / `reclaimPredictions`. Mirrors the OpenAI base path in shape and default-token handling.
- **Fixed-overhead estimate** includes `{tools: v3Tools, responseFormat, providerOptions}`. The prior CodeRabbit thread flagging unconditional `tools` inclusion is already resolved: `v3Tools` is derived from `toolsRecord`, which is `{}/[]` whenever tools are disabled, so the surplus `[]` is ~1-2 tokens — the PR-body justification holds, and I don't re-raise it. I also checked the `providerOptions` stringify goes into *token estimation*, not logging, so no secret-leak concern.
- **No `openaiChatCompletionsBase.ts` in the diff** — the regression fixed in #1730 is not reintroduced.
- Backward compatibility (Rule 5): no public SDK surface changes; this is an additive internal guard with default-off behavior.
- Offline e2e fixtures + `test:anthropic-generate-guard` script — Rule 15 honors the determinism/CLI-usage pattern consistently with the rest of the suite.

## Impact analysis note

The sandbox restricts filesystem reads to `.yama/`, and I did not have engine access to the code-review-graph for this PR, so I couldn't run `detect_changes_tool` / `get_impact_radius_tool` on `nativeGenerateGuard.ts` / `nativeGenerateLoop.ts`. Impact assessment above is from `git diff` + review context rather than a graph trace; I'm not fabricating out-of-diff callers I couldn't verify.

## Findings

|---|---|---|
| MINOR | `test/…-anthropic-generate-guard.ts:105-112` | `assert.ok` precondition messages inverted vs. the `includes(...)` check |

Checked and clean: delimiter allocation in `message()`, tool-pair balance / orphan checks in the per-request loop, the final-request assertions (`PRESERVE_TASK`/`PRESERVE_SYSTEM` retained, tail `OUTPUT_{steps-1}` kept), and the per-stage reclaim assertions (`preview`, `batches`, `fits`).

Tara-ag:
**APPROVE** — clean, focused Anthropic context-guard wiring with strong e2e coverage. One MINOR (inverted precondition diagnostic message) is non-blocking.

## What I validated

- **Diff is small and well-scoped**: `package.json` gains the `test:anthropic-generate-guard` script; `anthropic/client.ts` adds the `createNativeGenerateGuard` import and wires it into `runNativeGenerateLoop` in `executeNativeGenerate`; one new continuous suite.
- **Guard wiring** (`client.ts`): the guard is instantiated per call with `filter: hasTools`, `task`/`system` wrapped, `maxAvailableInput` derived via `getAvailableInputTokens("anthropic", modelId, options.maxTokens ?? undefined)`, and `filterCandidates`/`splitCandidates` reference the loop's `v3Tools` / `v3Messages` / `reclaimPredictions`. Mirrors the OpenAI base path.
- **Fixed-overhead estimate** includes `{tools: v3Tools, responseFormat, providerOptions}`. The prior CodeRabbit thread flagging unconditional `tools` inclusion is resolved: `v3Tools` derives from `toolsRecord`, which is `{}/[]` whenever tools are disabled, so the surplus is ~1-2 tokens. I also confirmed the `providerOptions` stringify feeds *token estimation*, not logging — no secret-leak concern.
- **No `openaiChatCompletionsBase.ts` change** — the regression fixed in #1730 is not reintroduced.
- Backward compatibility (Rule 5): no public SDK surface changes; purely additive internal guard with default-off behavior.
- Offline e2e fixtures + `test:anthropic-generate-guard` script align with the Rule 15 pattern used by the rest of the suite (single module graph via built `dist/`).

## Impact analysis note

The sandbox restricts filesystem reads to `.yama/`, and I did not have code-review-graph engine access for this PR, so I could not run `detect_changes_tool` / `get_impact_radius_tool` on `nativeGenerateGuard.ts` / `nativeGenerateLoop.ts`. Impact assessment is from `git diff` + review context, not a graph trace; I did not fabricate out-of-diff callers I could not verify.

## Findings

|---|---|---|
| MINOR | `test/continuous-test-suite-anthropic-generate-guard.ts:105-112` | `assert.ok` precondition messages inverted vs. the `includes(...)` check |

Checked and clean: delimiter allocation in `message()`, tool-pair balance / orphan checks in the per-request loop, final-request assertions (`PRESERVE_TASK`/`PRESERVE_SYSTEM` retained, tail `OUTPUT_{steps-1}` kept), and per-stage reclaim assertions (`preview`, `batches`, `fits`). No secrets in the diff (`transformParamsForLogging` path untouched).

Tara-ag (approved):
**APPROVE** — clean, focused Anthropic context-guard wiring with strong e2e coverage. The guard is correctly instantiated per call, mirrors the OpenAI base path, and the fixed-overhead token estimate is accurate (`v3Tools` is empty when tools are disabled). One MINOR (inverted precondition diagnostic message in the test) is non-blocking. See the summary comment for the full review.

test/continuous-test-suite-anthropic-generate-guard.ts
```diff
@@ -0,0 +1,176 @@
...
+                content?: unknown;
+              }>;
+            }>;
+          },
+      );
+      assert.ok(
+        JSON.stringify(bodies[1]).includes("first_0"),
+        "precondition: original tool call never reached endpoint",
+      );
+      assert.ok(
```
Tara-ag: **MINOR** — inverted precondition diagnostic messages.

These two `assert.ok` precondition checks *pass* when `first_0` / `OUTPUT_0` are **present** in `bodies[1]` (i.e. when the original tool call/result *did* reach the endpoint before the guard reclaims later steps). The failure messages read the opposite way ("…**never** reached endpoint"), so if the precondition ever fails the message will mislead. Reword so the message matches the actual condition:

```suggestion
      assert.ok(
        JSON.stringify(bodies[1]).includes("first_0"),
        "precondition: original tool call did not reach endpoint",
      );
      assert.ok(
        JSON.stringify(bodies[1]).includes("OUTPUT_0"),
        "precondition: original result did not reach endpoint",
      );
```

Tara-ag (commented):
MINOR: inverted precondition diagnostic messages in the new guard test (see inline comment).

PR #1676, opened 2026-09-09
docs(ci): record that the advisory gate is time-dependent

murdore:
## Summary

Documents in `CLAUDE.md` that the `🔒 Security & Environment Validation` CI gate (backed by `scripts/security-check.ts`, which runs `pnpm audit --prod --json` **live**) is **time-dependent** — a green run expires, and a red gate is rarely caused by your own diff.

Covers:
- **A PR's green expires**: two PRs with identical dependency trees passed at 20:05 UTC and failed at 02:03 UTC once new advisories published; the first looked green/mergeable and was neither.
- **A red gate usually is not yours**: a new advisory turns `test` red on every open PR at once (including ones touching no manifest). The check's own output appends the "behind origin/release" / "may not originate in your changes" note.
- A copy-paste-able **reproduction command** that audits the exact tip CI ran: `origin/release` in a throwaway worktree (unique `mktemp` path, cleaned up even on failure) with `--frozen-lockfile` to match CI's install.
- **Fixing it**: raise the floor and move the `pnpm.overrides` band with it. Warns against a lockfile-only bump (greens CI but not downstream consumers) and a raised floor with a stale override band (can pin transitive consumers *inside* the new advisory's range). The js-yaml example is framed as history, anchored on the permanent GHSA ranges, with the GraphQL query to enumerate every range on a record.

Documentation-only change — single file, no code or config drift.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance explaining that security validation results are time-dependent and may change when new advisories are published.
  * Documented steps for reproducing newly reported dependency vulnerabilities and resolving them by updating dependency requirements and overrides.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -550,6 +550,100 @@ membership.
 The general trap: a command that exits 0 is not evidence it did anything. Check
 that the work landed — a changed file, a written artifact — not the status code.
 
+### ⚠️ The advisory gate is time-dependent: a green run expires
+
+`scripts/security-check.ts` — the "🔒 Security & Environment Validation" step
+inside `test-shards (validate)`, and therefore inside the required `test`
+check — runs `pnpm audit --prod --json` **live** against the advisory
+database. Nothing about its verdict is pinned to the commit.
+
+Two consequences, both of which have already produced wrong conclusions here:
+
+**A pull request's green is a statement about when it ran, not about its
+diff.** When the js-yaml and hono advisories were published, one PR's
+`validate` shard had run at 20:05 UTC and passed; another ran at 02:03 UTC
+and failed. Identical dependency trees. The first PR looked green and
+mergeable and was neither — a re-run would have failed it. **Do not treat a
+green `test` older than the newest advisory publication as current**, and
+never conclude from "this PR is green and that one is red" that the
+difference is in their diffs.
+
+**A red gate usually is not yours.** Because the audit is repo-wide and
+live, a newly published advisory turns `test` red on _every_ open pull
+request simultaneously, including ones that touch no manifest at all. The
+check's own output says so — it appends a note about the branch being behind
+`origin/release` and warns that the failure "may not originate in your
+changes." Before investigating your own diff, reproduce on the untouched
+`release` tip:
+
+```bash
+# A fixed path is not re-runnable: the second call dies with
+# `fatal: '/tmp/nl-audit' already exists` instead of auditing. Take a unique
+# directory, and remove the worktree outside the `&&` chain so a failed audit
+# still cleans up after itself.
+audit_dir="$(mktemp -d)/nl-audit"
+git fetch origin release \
+  && git worktree add "$audit_dir" origin/release \
+  && ( cd "$audit_dir" && pnpm install --frozen-lockfile \
+       && pnpm exec tsx scripts/security-check.ts )
+# Guarded: if the fetch or the `worktree add` failed, there is nothing to
+# remove, and an unguarded remove prints `fatal: ... is not a working tree`
+# on exactly the failure path this section exists to de-confuse.
+[ -d "$audit_dir" ] && git worktree remove --force "$audit_dir"
+rm -rf "$(dirname "$audit_dir")"
+```
+
+Note `origin/release`, not `release`: the check reports how far behind
+`origin/release` you are, so auditing a stale local ref reproduces a tree CI
+never ran and sends you chasing a difference that is your own checkout. The
+throwaway worktree keeps it read-only — `git stash && git checkout` mutates
+the tree you are debugging, which is a poor trade for a diagnostic whose
+whole question is "was it already broken without me?" And `--frozen-lockfile`
+because that is how every CI job installs: a plain `pnpm install` may
+re-resolve a transitive range and hand you a tree CI never audited, which is
+the same class of mistake this whole section is about.
+
+If it fails there too, the fix belongs in its own dependency PR, not in
+whatever you were working on.
+
+**Fixing it: raise the floor, and move the override band with it.** Prefer
+bumping the declared range over adding an accepted-risk entry — an accepted
+risk silences the gate for everyone, while a floor bump is what actually
+protects downstream consumers, who resolve from `package.json` and never see
+our lockfile. Two things are easy to get wrong:
+
+- **A lockfile-only bump is not a fix.** It greens CI and leaves every
+  installer of the published package on the vulnerable version.
+- **A raised floor with a stale `pnpm.overrides` band is worse than no
+  override.** The entries lift transitive consumers to a version that was
+  patched for an _older_ advisory, which can sit inside the _new_ one's
+  vulnerable range.
+
+  The worked example is historical, and deliberately so — the repository has
+  since been fixed, so do not expect `package.json` to still show the broken
+  state. `GHSA-2883-xcg3-v3hh` covers `>=3.0.0 <3.15.2` **and**
+  `>=4.0.0 <4.3.2`; the js-yaml overrides then read `>=3.14.2` / `>=4.1.1`,
+  which are the fixes for an _earlier_ js-yaml advisory and sit **inside**
+  both of this one's bands. They now read `>=3.15.2` / `>=4.3.2`, on the band
+  edges, which is what correct looks like. The trap is the shape, not those
+  numbers: an override target that was patched for the advisory you fixed
+  last time.
+
+  A lockfile that happens to resolve a safe version hides this, so CI stays
+  green and the pin re-manifests on the next install that re-resolves a
+  transitive range. Note also that `pnpm audit` printed only the 4.x range —
+  the 3.x half was invisible. Check every range on the GHSA record, not just
+  the one the audit showed you:
+
+  ```bash
+  gh api graphql -f query='{securityVulnerabilities(first:10, ecosystem:NPM,
+    package:"<pkg>"){nodes{advisory{ghsaId} vulnerableVersionRange
+    firstPatchedVersion{identifier}}}}'
+  ```
+
+  `fast-xml-parser`, `undici` and now `js-yaml` in the same table are the
+  worked examples of a band kept in step with its floor.
+
 ### ⚠️ ffmpeg is deliberately not installed in CI
 
 Nothing CI runs needs it. No package script invokes it, nothing installs it as a


CLAUDE.md
```diff
@@ -527,6 +527,66 @@ because a push to `release` from outside a pull request carries no check runs
...
+**A red gate usually is not yours.** Because the audit is repo-wide and
+live, a newly published advisory turns `test` red on _every_ open pull
+request simultaneously, including ones that touch no manifest at all. The
+check's own output says so — it appends a note about the branch being behind
+`origin/release` and warns that the failure "may not originate in your
+changes." Before investigating your own diff, reproduce on the untouched
+`release` tip:
+
+```bash
+git stash && git checkout release && pnpm exec tsx scripts/security-check.ts
```
Tara-ag: **MINOR** — this reproduction uses the *local* `release` ref and mutates the working tree.

The doc's own check logic (in `security-check.ts`) compares against `origin/release`, but this command does `git checkout release` against the *local* ref, which is routinely stale relative to CI. If a contributor's local `release` is behind `origin/release`, the tree they audit is not the tree CI audited — reproducing the exact confusion this section exists to prevent. `git stash` is also a mutating side effect (and requires a clean tree) for what should be a read-only diagnostic.

Suggested:

````suggestion
git stash && git fetch origin release && git checkout origin/release && pnpm exec tsx scripts/security-check.ts
````

(`origin/release`, fetched first, ensures you audit the tip CI actually ran. Everything else in the section — the live-audit and override-band points — checks out against `package.json`'s `pnpm.overrides` and the script source.)

**Review by Tara-ag — COMMENTED** [review]
Reviewed the documentation addition against the actual code and CI config. Verified every technical claim in the section.

**Comment by Tara-ag** [comment]
## APPROVE

Documentation-only change to `CLAUDE.md`; the section is accurate and well-scoped. I verified the claims against the checkout and CI, not just the prose.

|---|---|---|
| MINOR | `CLAUDE.md:557` | `git checkout release` uses a possibly-stale local ref and `git stash` mutates the working tree — prefer `origin/release` (fetched first) for a read-only reproduction |

**What I checked (all verified against source):**
- `security-check.ts` runs `pnpm audit --prod --json` live, un-pinned — matches the "green run expires" claim.
- The security step sits in `test-shards (validate)` (`matrix.group == 'validate'` runs `validate:all`), which feeds the required `test` job (`needs: [test-shards]`) — matches the CI attribution claim.
- The script's failing output appends the "N commit(s) behind origin/release" / "may not originate in your changes" note — matches the section's description of the check's own output.
- `js-yaml` overrides read `>=3.14.2` / `>=4.1.1`, both inside the described GHSA-2883-xcg3-v3hh ranges — internally consistent with `package.json`'s `pnpm.overrides`.
- `fast-xml-parser` (floor `^5.7.0`, override band `>=5.0.0 <5.7.0 → >=5.7.0`) and `undici` (floor `>=7.24.0 <8.0.0` with matching bands) are valid "band kept in step with floor" examples.
- The lockfile-only-bump / stale-override-band traps match how consumers resolve from `package.json` (the lockfile is not in `files/`).

**Checked clean:** single-commit policy (verified by CI), Conventional Commit `docs(ci):` with scope, no lockfile/config drift.

No blockers. The one MINOR is a copy-paste-able command in an authoritative doc whose entire purpose is preventing misdiagnosis, so it is worth tightening.

File: CLAUDE.md [hunk 2]
```diff
@@ -527,6 +527,66 @@ because a push to `release` from outside a pull request carries no check runs
...
+```
+
```
Tara-ag: **MINOR** — this repro leans on a possibly-stale *local* `release` ref and mutates the working tree, for what is meant to be a read-only diagnostic.

`git checkout release` checks out your local `release` ref, but CI's audit runs against `origin/release` — the check appends the "N commit(s) behind origin/release" note that this section itself quotes. If your local `release` lags the remote, you can reproduce, "fix", and chase a ghost. And `git stash` rewrites the working tree for a command whose purpose is "did this fail on the untouched tip?"

Suggest pinning to the remote through a throwaway worktree so nothing local is touched:

``````suggestion
```bash
git fetch origin release && git worktree add /tmp/nl-verify origin/release \
  && cd /tmp/nl-verify && pnpm exec tsx scripts/security-check.ts
```
``````

**Review by Tara-ag — APPROVED** [review]
Approving — documentation-only change; every technical claim verified against scripts/security-check.ts, ci.yml, package.json overrides, and the check's own output. One MINOR inline suggestion on the repro command.

**Comment by Tara-ag** [comment]
**APPROVE** — documentation-only change; every technical claim in the new section verifies against the actual code and CI config, with one MINOR, non-blocking suggestion.

| MINOR | `CLAUDE.md:557-559` | Repro command uses local `release` ref and `git stash`; prefer `origin/release` via a throwaway worktree |

**Verified clean:**
- `scripts/security-check.ts` runs `pnpm audit --prod --json` live, un-pinned — "a green run expires" is accurate.
- `.github/workflows/ci.yml` — the `🔒 Security & Environment Validation` step runs in `test-shards (validate)` feeding the required `test` check via `needs:`.
- The check's own output appends the "behind origin/release" / "may not originate in your changes" note as claimed.
- `package.json` js-yaml override bands (`>=3.14.2` / `>=4.1.1`) are indeed inside the `GHSA-2883-xcg3-v3hh` vulnerable ranges; `fast-xml-parser` / `undici` are genuine band-kept-in-step examples.
- Downstream-consumer resolution from `package.json` (lockfile not shipped) makes the "lockfile-only bump is not a fix" trap accurate.

No CRITICAL / MAJOR findings. No secrets, no code changes, no backward-compatibility impact (rule 5 not applicable — docs only).

File: CLAUDE.md (hunk 2 above)
Comment (Tara-ag) [reply CLAUDE.md]: Re-review (PR unchanged at `209893f4`): the repro command in the diff still reads `git stash && git checkout release && pnpm exec tsx scripts/security-check.ts` — the local `release` ref and the tree-mutating `git stash` are both still present, and the script's own `commitsBehindRelease` compares against `origin/release`. No author reply or fix since the prior pass, so I'm keeping this finding open.

The concrete fix (audit the exact tip CI runs, read-only):

```suggestion
git fetch origin release && git worktree add /tmp/nl-verify origin/release
  && cd /tmp/nl-verify && pnpm exec tsx scripts/security-check.ts
```

Tara-ag:
**Re-review pass (2026-09-09)** — **APPROVE**. (Recurring review; the PR head is unchanged at `209893f4` from the commit both prior summaries approved.)

**What's unchanged and re-verified this pass:**
- Docs-only change to `CLAUDE.md`: `scripts/security-check.ts` (graph node `scripts/security-check.ts#2446`) with its `commitsBehindRelease` function corroborates the section's claims — including the `origin/release` comparison that the one open finding hinges on.
- Code-review-graph: risk **0.00**, 0 changed functions, 0 affected flows → rule 5 (backward compat) not applicable; no secrets, no code/config change.

**Open finding (kept, not reposted):**
- MINOR — `CLAUDE.md:557-559`: repro command still uses the possibly-stale local `release` ref and tree-mutating `git stash`. No author reply or fix since the prior pass, so the finding remains open on its existing thread rather than being reintroduced.

No new findings. One open MINOR; everything else clean.

CLAUDE.md
murdore: Fixed in `24731cced` — both of you flagged this and you were right. The command I wrote would have caused the precise failure the section exists to prevent.

Now:

```bash
git fetch origin release && git worktree add /tmp/nl-audit origin/release \
  && cd /tmp/nl-audit && pnpm install && pnpm exec tsx scripts/security-check.ts
```

I took the throwaway-worktree form over the `git checkout origin/release` suggestion for the read-only property you both noted — the diagnostic question is "was this already broken without me?", and mutating the tree under investigation is a bad trade for answering it. `pnpm install` is included because the audit reads the installed tree, so a fresh worktree without it audits nothing.

The prose now says explicitly why it is `origin/release` and not `release`, since the check reports how far behind `origin/release` you are.

Ran it verbatim against the current tip before committing: worktree created at `3fdf80f1d`, `PASS dependencies`, removed cleanly. I did not want to document a command I had not executed — that would repeat this finding in a different form.

CLAUDE.md
```diff
@@ -527,6 +527,74 @@ because a push to `release` from outside a pull request carries no check runs
...
+our lockfile. Two things are easy to get wrong:
+
+- **A lockfile-only bump is not a fix.** It greens CI and leaves every
+  installer of the published package on the vulnerable version.
+- **A raised floor with a stale `pnpm.overrides` band is worse than no
+  override.** The entries lift transitive consumers to a version that was
+  patched for an _older_ advisory, which can sit inside the _new_ one's
+  vulnerable range. `GHSA-2883-xcg3-v3hh` covers `>=3.0.0 <3.15.2` **and**
+  `>=4.0.0 <4.3.2`, while the js-yaml overrides read `>=3.14.2` / `>=4.1.1`
+  — both inside the band. A lockfile that happens to resolve a safe version
```
Tara-ag: **MINOR** — this concrete example is stale against the tree this section documents.

The base tip (`3fdf80f`) `package.json` no longer has js-yaml overrides that read `>=3.14.2` / `>=4.1.1`. They now read:

```json
"js-yaml@>=4.0.0 <4.3.2": ">=4.3.2",
"js-yaml@>=3.0.0 <3.15.2": ">=3.15.2",
```

Those bump to the patched versions and sit exactly on the GHSA-2883-xcg3-v3hh boundaries — i.e. js-yaml is *now* a "band kept in step with its floor". So it is currently the **good** example, which undercuts (and contradicts) the cautionary reading here, and a reader who checks the numbers against `package.json` will find they don't match — the same stale-state confusion this section exists to prevent.

Since the GHSA ranges (`>=3.0.0 <3.15.2`, `>=4.0.0 <4.3.2`) are permanent facts both before and after the fix, the simplest non-stale framing is to anchor the example in the incident:

````suggestion
  vulnerable range. `GHSA-2883-xcg3-v3hh` covers `>=3.0.0 <3.15.2` **and**
  `>=4.0.0 <4.3.2`, while the js-yaml overrides *at the time* read `>=3.14.2` / `>=4.1.1`
````

(If you'd rather keep it present-tense, update the two numbers to the current `>=3.15.2` / `>=4.3.2` and reword so js-yaml exemplifies the in-step band rather than a stale one.)

File: CLAUDE.md [hunk 4]
```diff
@@ -527,6 +527,74 @@ because a push to `release` from outside a pull request carries no check runs
...
+request simultaneously, including ones that touch no manifest at all. The
+check's own output says so — it appends a note about the branch being behind
+`origin/release` and warns that the failure "may not originate in your
+changes." Before investigating your own diff, reproduce on the untouched
+`release` tip:
+
+```bash
+git fetch origin release && git worktree add /tmp/nl-audit origin/release \
+  && cd /tmp/nl-audit && pnpm install && pnpm exec tsx scripts/security-check.ts
+```
```
Tara-ag, line 560: **MINOR** — the repro `pnpm install` is not the tree CI audits.

The very first lines of this section's subject (the `test-shards (validate)` step, and `ci.yml` everywhere) install with `pnpm install --frozen-lockfile`. A bare `pnpm install` in a fresh worktree will happily re-resolve if the committed `package.json` and `pnpm-lock.yaml` ever disagree about a transitive range — which is precisely the stale-override trap the section below documents — producing a tree CI never ran. The section teaches "audit the exact tree CI ran"; the command should install the same way CI does.

````suggestion
  && cd /tmp/nl-audit && pnpm install --frozen-lockfile && pnpm exec tsx scripts/security-check.ts
````

**Review by Tara-ag — APPROVED** [review]
Documentation-only change; the section's core claims verify against `scripts/security-check.ts`, `ci.yml`, and `package.json`. Two MINOR accuracy notes (js-yaml example is now stale; repro could use `--frozen-lockfile` to match CI). The prior repro-command finding is resolved by the author's `origin/release` worktree fix.

**Comment by Tara-ag** [comment]
## APPROVE (re-review, head `24731cced`)

Documentation-only change to `CLAUDE.md`. The section's technical claims verify against the checkout — `scripts/security-check.ts` runs `pnpm audit --prod --json` live and appends the "behind origin/release / may not originate in your changes" note; the `🔒 Security & Environment Validation` step sits in `test-shards (validate)` feeding the required `test` job (`needs: [test-shards]`); `package.json` `files` excludes the lockfile, so consumers resolve from manifests. Two new MINOR accuracy notes:

|---|---|---|
| resolved | `CLAUDE.md:558-559` | Prior repro-command finding — author's `origin/release` throwaway-worktree fix (with `pnpm install`) genuinely addresses the stale-local-ref / mutating-tree concern; thread resolved. |
| MINOR | `CLAUDE.md:584` | js-yaml example is stale: base-tip overrides now read `>=3.15.2` / `>=4.3.2` (in step with the band); the doc's `>=3.14.2` / `>=4.1.1` numbers no longer match `package.json`. Anchor it in the incident or update it. |
| MINOR | `CLAUDE.md:559` | Repro could install with `--frozen-lockfile`, matching CI, so the audited tree can't differ from CI's. |

**Verified clean:** single-commit policy + Conventional Commit `docs(ci):` (both confirmed by CI); no lockfile/config/code drift (1 file, +68); `fast-xml-parser` (`^5.7.0` floor, band `>=5.0.0 <5.7.0 → >=5.7.0`) and `undici` (`>=7.24.0 <8.0.0` floor, matching bands) are genuine in-step-band examples; the `gh api graphql securityVulnerabilities` query shape is valid. Rule 5 (backward compat) and the security bar are not applicable — docs only.

No CRITICAL / MAJOR findings; the two MINORs are non-blocking.

File: CLAUDE.md [hunk 5]
```diff
@@ -527,6 +527,74 @@ because a push to `release` from outside a pull request carries no check runs
...
+live, a newly published advisory turns `test` red on _every_ open pull
+request simultaneously, including ones that touch no manifest at all. The
+check's own output says so — it appends a note about the branch being behind
+`origin/release` and warns that the failure "may not originate in your
+changes." Before investigating your own diff, reproduce on the untouched
+`release` tip:
+
+```bash
+git fetch origin release && git worktree add /tmp/nl-audit origin/release \
+  && cd /tmp/nl-audit && pnpm install && pnpm exec tsx scripts/security-check.ts
```
Tara-ag: **MINOR** — make the repro match CI exactly: CI installs with `--frozen-lockfile` everywhere, but this documents plain `pnpm install`.

The section's whole premise is "reproduce the tree CI actually audited." A non-frozen install can silently re-resolve transitive ranges (it can even pull in an override-band change). On a fresh `origin/release` worktree the lockfile is present and consistent, so a frozen install is safe and deterministic.

```suggestion
  && cd /tmp/nl-audit && pnpm install --frozen-lockfile && pnpm exec tsx scripts/security-check.ts
```

Everything else in the command — `origin/release`, the throwaway worktree, the read-only property you noted — is correct and I'm not asking you to touch it.

CLAUDE.md
```diff
@@ -527,6 +527,74 @@ because a push to `release` from outside a pull request carries no check runs
...
+protects downstream consumers, who resolve from `package.json` and never see
+our lockfile. Two things are easy to get wrong:
+
+- **A lockfile-only bump is not a fix.** It greens CI and leaves every
+  installer of the published package on the vulnerable version.
+- **A raised floor with a stale `pnpm.overrides` band is worse than no
+  override.** The entries lift transitive consumers to a version that was
+  patched for an _older_ advisory, which can sit inside the _new_ one's
+  vulnerable range. `GHSA-2883-xcg3-v3hh` covers `>=3.0.0 <3.15.2` **and**
+  `>=4.0.0 <4.3.2`, while the js-yaml overrides read `>=3.14.2` / `>=4.1.1`
```
Tara-ag: **MINOR** — the quoted js-yaml override values are stale against the current `package.json`.

The `pnpm.overrides` in the repo at this commit read `"js-yaml@>=4.0.0 <4.3.2": ">=4.3.2"` and `"js-yaml@>=3.0.0 <3.15.2": ">=3.15.2"` — i.e. floors `>=4.3.2` / `>=3.15.2`, not `>=3.14.2` / `>=4.1.1`. As written, the example claims the overrides' floors sit *inside* the vulnerable band, but the current floors sit exactly at the band's upper edges — so the illustration no longer demonstrates the "raised floor pinned into the vulnerable range" trap it is meant to show.

<details><summary>Verification (package.json @ 24731cced)</summary>

```json
"js-yaml@>=4.0.0 <4.3.2": ">=4.3.2",
"js-yaml@>=3.0.0 <3.15.2": ">=3.15.2",
```

`GHSA-2883-xcg3-v3hh` band: `>=3.0.0 <3.15.2` and `>=4.0.0 <4.3.2`.
</details>

Suggest updating the example to the current values, or phrasing it as tied to the historical state (e.g. "at the time of writing, the js-yaml overrides read …") so a future drift doesn't silently falsify the doc.

```suggestion
  vulnerable range. `GHSA-2883-xcg3-v3hh` covers `>=3.0.0 <3.15.2` **and**
  `>=4.0.0 <4.3.2`, and for the senior advisory the js-yaml override band
  (`>=3.0.0 <3.15.2 → >=3.15.2` and `>=4.0.0 <4.3.2 → >=4.3.2`) was moved in
  step with the floor, so the lockfile hides the band and a re-resolve re-pins
  an *older* transitive version — check every range listed on the GHSA record,
  not just the one `pnpm audit` printed:
```

Tara-ag (approved):
APPROVE — documentation-only change. The two prior MINORs (repro command) are fixed by the author's `24731cced` (worktree against `origin/release`, read-only) — resolved those threads. Two new non-blocking MINORs on the current head: make the repro use `--frozen-lockfile` to match CI, and the quoted js-yaml override values are stale vs. the committed `package.json`.

Tara-ag:
## APPROVE

Documentation-only change to `CLAUDE.md` (advisory-gate section). All technical claims verify against `scripts/security-check.ts`, `.github/workflows/ci.yml`, and `package.json`. The single-commit / Conventional Commit policy is enforced by CI.

|---|---|---|---|
| MINOR | `CLAUDE.md:557-559` | Repro used a stale local `release` ref + tree-mutating `git stash` instead of `origin/release` | FIXED by author (`24731cced`, throwaway worktree) — threads resolved |
| MINOR | `CLAUDE.md:558` | Repro runs `pnpm install` while CI uses `--frozen-lockfile`; use frozen to match CI exactly | NEW — non-blocking |
| MINOR | `CLAUDE.md:583` | Quoted js-yaml override values (`>=3.14.2` / `>=4.1.1`) are stale vs committed `package.json` (`>=3.15.2` / `>=4.3.2`) | NEW — non-blocking |

**Verified clean:**
- `scripts/security-check.ts` runs `pnpm audit --prod --json` live, un-pinned → "a green run expires" is accurate; its failing output appends the "N commit(s) behind origin/release / may not originate in your changes" note as claimed.
- The `🔒 Security & Environment Validation` step lives in `test-shards (validate)` feeding the required `test` check via `needs:` (CI attribution claim correct).
- Downstream-consumer resolution from `package.json` (lockfile not in `files/`) makes the "lockfile-only bump is not a fix" trap accurate.
- `fast-xml-parser` / `undici` override bands are genuine band-kept-in-step-with-floor examples.
- No secrets (placeholder references only), no code/config change, no backward-compat impact (rule 5 N/A — docs only).

No CRITICAL / MAJOR findings.

CLAUDE.md
Tara-ag: Thanks — the `origin/release` throwaway-worktree form is exactly right and I've resolved both repro threads. The read-only property is worth the extra `pnpm install`. Only two non-blocking nits remain on the current head (frozen-lockfile for CI fidelity; js-yaml values in the example are now stale vs `package.json`).

CLAUDE.md
murdore: Applied in `79d5c3d41` — you and the parallel thread both landed on this and the reasoning is right. The section's premise is "audit the tree CI actually ran", and a non-frozen install can hand you a different one.

```bash
git fetch origin release && git worktree add /tmp/nl-audit origin/release \
  && cd /tmp/nl-audit && pnpm install --frozen-lockfile \
  && pnpm exec tsx scripts/security-check.ts
```

The prose now says why, rather than leaving the flag to be taken on faith: a plain `pnpm install` may re-resolve a transitive range and produce a tree CI never audited — the same class of mistake the whole section is about.

Ran verbatim against `origin/release` at `3fdf80f1d` before committing: frozen install clean, `PASS dependencies`, worktree removed.

CLAUDE.md
murdore: Fixed in `79d5c3d41`, and this was the best catch on the PR — my own merged dependency fix (#1671) invalidated my own cautionary example. A reader checking `package.json` would have found the numbers contradicting the warning, which is precisely the stale-state confusion the section exists to prevent.

Rather than swap in current values — which would leave the same landmine for the next person to fix an override band — the example is now explicitly **historical**, anchored on the facts that do not move:

> The worked example is historical, and deliberately so — the repository has since been fixed, so do not expect `package.json` to still show the broken state. `GHSA-2883-xcg3-v3hh` covers `>=3.0.0 <3.15.2` **and** `>=4.0.0 <4.3.2`; the js-yaml overrides then read `>=3.14.2` / `>=4.1.1` … They now read `>=3.15.2` / `>=4.3.2`, on the band edges, which is what correct looks like. The trap is the shape, not those numbers: an override target that was patched for the advisory you fixed last time.

js-yaml has also been added alongside `fast-xml-parser` and `undici` as a worked example of a band kept in step with its floor — it earned that by being fixed.

Verified against `origin/release` at `3fdf80f1d`: the overrides read `>=4.3.2` / `>=3.15.2`, matching what the text now claims.

Tara-ag:
## APPROVE (re-review, head `79d5c3d41`)

Documentation-only change to `CLAUDE.md` (1 file, +83); every technical claim in the new advisory-gate section verifies against the checkout. Both MINORs left open on the last pass are now adequately fixed and their threads resolved.

|---|---|---|---|
| MINOR | `CLAUDE.md:558-559` | Repro used a possibly-stale local `release` ref + tree-mutating `git stash` | FIXED by author (worktree on `origin/release`, `--frozen-lockfile`) — thread resolved |
| MINOR | `CLAUDE.md:583` | Quoted js-yaml overrides (`>=3.14.2` / `>=4.1.1`) were stale vs committed `package.json` | FIXED by author — example now explicitly historical and matches the current `>=3.15.2` / `>=4.3.2` values — thread resolved |

**Re-verified against source (head + `origin/release` @ `3fdf80f1d`):**
- `scripts/security-check.ts` runs `pnpm audit --prod --json` live, un-pinned → "a green run expires" is accurate; its failing output appends the "N commit(s) behind origin/release / may not originate in your changes" note via `commitsBehindRelease`.
- The `🔒 Security & Environment Validation` step lives in `test-shards (validate)` (`matrix.group == 'validate'` → `pnpm run validate:all`), feeding the **required** `test` job via `needs: [test-shards]` (`ci.yml`) — the CI-attribution claim holds.
- CI installs with `pnpm install --frozen-lockfile` throughout, so the documented repro now matches CI exactly.
- `package.json` js-yaml overrides at the base tip read `>=4.3.2` / `>=3.15.2` — on the band edges, exactly as the doc states; the GHSA band facts (`>=3.0.0 <3.15.2`, `>=4.0.0 <4.3.2`) are permanent. `fast-xml-parser` (`^5.7.0` floor, band `>=5.0.0 <5.7.0 → >=5.7.0`) and `undici` (`>=7.24.0 <8.0.0` floor, matching bands) are genuine in-step-band examples.
- The `gh api graphql securityVulnerabilities` query shape (arg `first`, `ecosystem`, `package`; `nodes{advisory{ghsaId} vulnerableVersionRange firstPatchedVersion{identifier}}`) is valid GitHub GraphQL.
- `package.json` `files` excludes `pnpm-lock.yaml`, so downstream consumers resolve from `package.json` — the "lockfile-only bump is not a fix" trap is real.

**Checked clean:** single-commit policy + Conventional Commit `docs(ci):` (both confirmed by CI); no lockfile/config/code drift; rule 5 (backward compat) and the security bar not applicable — docs only, no secrets, no executable change.

No CRITICAL / MAJOR findings. The two MINORs were correctly closed by the author; nothing new remains.

Tara-ag (approved):
APPROVE — documentation-only change to `CLAUDE.md`. All review findings are resolved on the current head `79d5c3d41`:

- Repro command now installs with `--frozen-lockfile` (matches CI) — verified in the diff.
- js-yaml example is now explicitly historical with current values `>=3.15.2` / `>=4.3.2` — verified in the diff, matching the committed `package.json`.

Verified clean: `scripts/security-check.ts` runs `pnpm audit --prod --json` live; the security step sits in `test-shards (validate)` feeding the required `test` check; `package.json` `files` excludes the lockfile; CONVENTIONAL_COMMIT + single-commit enforced by CI. No code/config change, no backward-compat impact (rule 5 N/A — docs only).

murdore:
## The documented repro, run verbatim

This PR publishes a command and argues it should be run before investigating a red advisory gate. Here is that command actually executing, copied from the section as written — unchanged path, unchanged flags.

```
git fetch origin release && git worktree add /tmp/nl-audit origin/release \
  && cd /tmp/nl-audit && pnpm install --frozen-lockfile \
  && pnpm exec tsx scripts/security-check.ts
```

```
### started 2026-09-11T23:58:25Z
Preparing worktree (detached HEAD 10fa282f2)
HEAD is now at 10fa282f2 feat(anthropic): truthful stream termination...
worktree_add_exit=0
install_exit=0          (Done in 13.5s using pnpm v10.15.1)

[SECURITY] Starting NeuroLink Security Validation...
[SECURITY] No secrets detected by Gitleaks
[SECURITY] Scanning dependencies for vulnerabilities...
[SECURITY] Accepted risk — uuid moderate ... advisory 1119441
[SECURITY] Accepted risk — form-data high ... advisory 1120743
[SECURITY] Accepted risk — adm-zip high ... advisory 1123686
   … 15 further "Accepted risk" lines, trimmed for length …
[SECURITY] All 18 open production advisories are explicitly accepted risk
[SECURITY] PASS secrets: passed
[SECURITY] PASS dependencies: passed
[SECURITY] WARN licenses: warning
[SECURITY] WARN bestPractices: warning
security_check_exit=0
### finished 2026-09-11T23:58:47Z
```

Every step the section promises does what it says: the throwaway worktree lands on `origin/release` rather than a stale local ref, `--frozen-lockfile` installs the tree CI would audit, and the check runs to a verdict. Twenty-two seconds end to end after the fetch.

## What the run also tells us right now

**The gate is green on `origin/release` at `10fa282f2` as of 2026-09-11T23:58Z, with 18 open production advisories, all explicitly accepted.** That is worth recording precisely because of this PR's own argument: it is a fact with a timestamp, not a property of the tree. It says nothing about tomorrow, and per the section it should be re-run immediately before any merge rather than cited later.

It also demonstrates the second claim — "a red gate usually is not yours". The 18 advisories are all repo-wide and transitive (`uuid`, `form-data`, `adm-zip`, `sharp`, `find-my-way`, `undici`, `ip-address`, `image-size`, `@opentelemetry/core`), none introduced by any open PR. Had any of them lacked an accepted-risk entry, every open PR would have gone red at once while touching nothing relevant.

One small note on the command itself, since I ran it rather than read it: `git worktree add /tmp/nl-audit` fails on a second run unless the previous worktree is removed, because the path is fixed. Worth a `git worktree remove /tmp/nl-audit` in the section, or the reader's second attempt gives a confusing "already exists" rather than an audit. I removed mine afterwards.

CLAUDE.md
```diff
@@ -550,6 +550,89 @@ membership.
...
+**A red gate usually is not yours.** Because the audit is repo-wide and
+live, a newly published advisory turns `test` red on _every_ open pull
+request simultaneously, including ones that touch no manifest at all. The
+check's own output says so — it appends a note about the branch being behind
+`origin/release` and warns that the failure "may not originate in your
+changes." Before investigating your own diff, reproduce on the untouched
+`release` tip:
+
+```bash
+git fetch origin release && git worktree add /tmp/nl-audit origin/release \
```
Tara-ag: **MINOR** — the repro uses a fixed worktree path, so the section's own command fails on a second run.

`git worktree add /tmp/nl-audit origin/release` targets a constant path. The first run (which the author helpfully executed at `2026-09-11T23:58Z` in [this comment](https://github.com/juspay/neurolink/pull/1676#issuecomment-5641991466)) creates the worktree, but nothing in the command removes it afterwards — despite the prose promising a "throwaway" worktree. On the very next invocation the reader gets `fatal: '/tmp/nl-audit' already exists` instead of an audit, which is the same class of confusing failure this section exists to prevent. The author explicitly flagged this in the PR thread ("fails on a second run unless the previous worktree is removed … Worth a `git worktree remove /tmp/nl-audit` in the section") and it is still unfixed at head.

Since the section's whole promise is a diagnostic that ends clean, pair the add with a remove so the command is idempotent (run `--force` against an empty detached worktree is harmless, and `rm -rf` keeps it under the read-only spirit):

````suggestion
git fetch origin release && git worktree add /tmp/nl-audit origin/release \
  && cd /tmp/nl-audit && pnpm install --frozen-lockfile \
  && pnpm exec tsx scripts/security-check.ts; \
  git worktree remove /tmp/nl-audit
````

(Loaded per the author's own verification — `--frozen-lockfile` matches CI, `origin/release` matches the check, everything else in the block is correct and unchanged.)

**Review by Tara-ag — APPROVED** [review]
APPROVE — documentation-only change to `CLAUDE.md`. All prior MINOR findings (repro stale-ref, frozen-lockfile, js-yaml override values) were adequately fixed and resolved by the author. One new non-blocking MINOR on the current head — the repro's fixed `/tmp/nl-audit` worktree path fails on a second run unless the prior worktree is removed (as the author's own verified run at 2026-09-11T23:58Z noted).

**Comment by Tara-ag** [comment]
## APPROVE (re-review, head `043001fd`)

Documentation-only change to `CLAUDE.md` — the advisory-gate section. All previously-raised MINOR findings are **resolved** by the author's fixes across `24731cced` / `79d5c3d41` (repro pinned to `origin/release` via throwaway worktree, `--frozen-lockfile` to match CI, js-yaml example made explicitly historical with values matching `package.json`). One new non-blocking MINOR remains, first surfaced by the author's own verified run of the documented command.

|---|---|---|---|
| MINOR | `CLAUDE.md:580` | Repro uses a fixed `/tmp/nl-audit` worktree path; a second run fails with "already exists" unless the prior worktree is removed. The section promises a "throwaway" worktree but the command never removes it. Author flagged this in the PR thread (2026-09-11T23:58Z) — still unfixed. | NEW — non-blocking |

**Re-verified against source at head `043001fd`:**
- `package.json` overrides confirm the docs' claims: `js-yaml@>=4.0.0 <4.3.2 → >=4.3.2`, `js-yaml@>=3.0.0 <3.15.2 → >=3.15.2` (on the GHSA-2883-xcg3-v3hh band edges), `fast-xml-parser@>=5.0.0 <5.7.0 → >=5.7.0`, `undici@>=7.0.0 <7.24.0 → >=7.24.0 <8.0.0` — genuine band-kept-in-step-with-floor examples. `files` excludes `pnpm-lock.yaml`, so the "lockfile-only bump is not a fix" trap is accurate.
- `validate:all` → `validate:security` → `scripts/security-check.ts`; the `🔒 Security & Environment Validation` step sits in `test-shards (validate)` feeding the required `test` check via `needs:` — CI attribution correct.
- `scripts/security-check.ts` runs `pnpm audit --prod --json` live, un-pinned → "a green run expires" is accurate; failing output appends the "behind origin/release / may not originate in your changes" note.
- Graph: risk **0.00**, 0 changed functions, 0 affected flows → rule 5 (backward compat) and the security bar not applicable — docs only, no secrets, no code/config change.

Single-commit policy + Conventional Commit `docs(ci):` enforced by CI. No CRITICAL / MAJOR findings. The one MINOR is a copy-paste-able command in an authoritative doc whose exact purpose is preventing misdiagnosis — worth the one-line fix the author already contemplated.

File: CLAUDE.md [hunk 8]
```diff
@@ -550,6 +550,89 @@ membership.
...
+
+**A red gate usually is not yours.** Because the audit is repo-wide and
+live, a newly published advisory turns `test` red on _every_ open pull
+request simultaneously, including ones that touch no manifest at all. The
+check's own output says so — it appends a note about the branch being behind
+`origin/release` and warns that the failure "may not originate in your
+changes." Before investigating your own diff, reproduce on the untouched
+`release` tip:
+
+```bash
```
Tara-ag, line 579: **MINOR** — the documented repro uses a fixed worktree path, so a second run fails confusingly.

The section's whole purpose is to prevent a contributor from misreading a red gate, but `git worktree add /tmp/nl-audit` fails on any second run with "`/tmp/nl-audit already exists`" unless the previous worktree was removed first — exactly the confusable-failure class this section exists to eliminate. This was also observed by the author when running the command verbatim.

```suggestion
git fetch origin release && git worktree add /tmp/nl-audit origin/release \
  && pnpm exec tsx scripts/security-check.ts \
  ; code=$?; git worktree remove /tmp/nl-audit --force; exit "$code"
```

If you'd rather not change the command, add a one-line prose note after it telling the reader to run `git worktree remove /tmp/nl-audit` before re-running. Everything else in the section verifies clean against `security-check.ts`, `ci.yml`, and `package.json`.

Tara-ag (approved):
APPROVE — documentation-only change to `CLAUDE.md`. All prior findings are resolved on the current head. One new non-blocking MINOR: the repro worktree path `/tmp/nl-audit` is fixed, so a second run fails with "already exists" — recommend removing it before re-running.

CLAUDE.md
```diff
@@ -550,6 +550,96 @@ membership.
...
+# `fatal: '/tmp/nl-audit' already exists` instead of auditing. Take a unique
+# directory, and remove the worktree outside the `&&` chain so a failed audit
+# still cleans up after itself.
+audit_dir="$(mktemp -d)/nl-audit"
+git fetch origin release \
+  && git worktree add "$audit_dir" origin/release \
+  && ( cd "$audit_dir" && pnpm install --frozen-lockfile \
+       && pnpm exec tsx scripts/security-check.ts )
+git worktree remove --force "$audit_dir"; rm -rf "$(dirname "$audit_dir")"
+```
```
Tara-ag, line 594: **MINOR** — the cleanup line can print a spurious `fatal:` right when the gate is already failing.

The `mktemp -d` rewrite (good — it fixes the fixed-path collision) creates the worktree only when `git fetch` succeeds, since both are on the `&&` chain. But the final `git worktree remove --force "$audit_dir"` is *outside* the chain, so on a `git fetch` failure (offline/registry hiccup) it runs against a path whose worktree was never registered and prints `fatal: '…/nl-audit' is not a working tree`. The `; rm -rf` still cleans the temp dir, so nothing leaks — but the reader sees a second, unrelated `fatal:` on exactly the failure path this whole section exists to de-confuse.

This section is explicitly about eliminating confusing failure output, so it is worth a one-word guard:

````suggestion
git worktree remove --force "$audit_dir" 2>/dev/null || true; rm -rf "$(dirname "$audit_dir")"
````

(`2>/dev/null` suppresses the not-a-worktree noise on the never-added case; `|| true` keeps the `;`-joined `rm -rf` from being skipped. On the normal success path nothing changes.) Everything else in the rewritten block — `mktemp -d` uniqueness, `origin/release`, `--frozen-lockfile`, cleanup outside the chain — is correct.

**Review by Tara-ag — COMMENTED** [review]
Approving review comments.

**Comment by Tara-ag** [comment]
## APPROVE (re-review, head `218533f3`)

Documentation-only change to `CLAUDE.md` — the advisory-gate section. The previously-open finding is now **fixed**, and one new non-blocking MINOR refinement remains.

|---|---|---|---|
| MINOR | `CLAUDE.md:586-590` | Repro used a fixed `/tmp/nl-audit` worktree path; a second run failed with "already exists" and nothing removed the worktree | **FIXED** by author at `218533f3` — rewritten with `mktemp -d` unique path + `git worktree remove`/`rm -rf` outside the `&&` chain. Thread `yama:repro-worktree-path-collision` resolved. |
| MINOR | `CLAUDE.md:590` | On a `git fetch` failure the chain-skips the worktree add, so the post-chain `git worktree remove --force "$audit_dir"` prints a spurious `fatal: … not a working tree` on the very failure path | NEW — non-blocking; suggest `2>/dev/null \|\| true` on the remove |

**Re-verified against head `218533f3`:**
- `scripts/security-check.ts` runs `pnpm audit --prod --json` live, un-pinned (verified in source) — "a green run expires" is accurate; the failing output appends the "N commit(s) behind origin/release … may not originate in your changes" note via `commitsBehindRelease`/`releaseAcceptsPackage`/`releaseOverridesPackage`.
- The `🔒 Security & Environment Validation` step sits in `test-shards (validate)` → `validate:all` → `validate:security`, feeding the required `test` check (`needs:`).
- The reproduced command now installs with `--frozen-lockfile` (matches CI) and audits `origin/release` via a throwaway worktree — [author's own verbatim run](https://github.com/juspay/neurolink/pull/1676#issuecomment-5641991466) confirms it lands and gives a verdict in ~22s.
- `package.json` overrides at head: `js-yaml@>=4.0.0 <4.3.2 → >=4.3.2`, `js-yaml@>=3.0.0 <3.15.2 → >=3.15.2` (on the GHSA-2883-xcg3-v3hh band edges), `fast-xml-parser@>=5.0.0 <5.7.0 → >=5.7.0`, `undici@>=7.0.0 <7.24.0 → >=7.24.0 <8.0.0` — the worked "band kept in step with its floor" examples; `files` excludes `pnpm-lock.yaml`, so the lockfile-only-bump trap is real.
- Single-commit policy + Conventional Commit `docs(ci):` enforced by CI (bot confirmed).

Rule 5 (backward compat) and the security bar N/A — docs only, no code/config change, no secrets. No CRITICAL / MAJOR findings. The prior canonical summary is [comment #5648238501](https://github.com/juspay/neurolink/pull/1676#issuecomment-5648238501); this pass confirms that finding is now closed and adds the one refinement above.

File: CLAUDE.md [hunk 10]
```diff
@@ -550,6 +550,96 @@ membership.
...
+# A fixed path is not re-runnable: the second call dies with
+# `fatal: '/tmp/nl-audit' already exists` instead of auditing. Take a unique
+# directory, and remove the worktree outside the `&&` chain so a failed audit
+# still cleans up after itself.
+audit_dir="$(mktemp -d)/nl-audit"
+git fetch origin release \
+  && git worktree add "$audit_dir" origin/release \
+  && ( cd "$audit_dir" && pnpm install --frozen-lockfile \
+       && pnpm exec tsx scripts/security-check.ts )
+git worktree remove --force "$audit_dir"; rm -rf "$(dirname "$audit_dir")"
```
Tara-ag: **MINOR** — the cleanup step prints a spurious `fatal` precisely on the failure path this section exists to de-confuse.

When `git fetch origin release` (or `git worktree add`) fails, the `&&` chain short-circuits, so the worktree at `$audit_dir` is never created — yet the trailing `git worktree remove --force "$audit_dir"` (outside the chain) still runs and prints `fatal: '…/nl-audit' is not a working tree`. That reader already got the real failure; the extra fatal is exactly the kind of confusing output "audit the untouched tip" is meant to avoid. It is harmless (the `;` and the trailing `rm -rf` still clean up), but the noise is free to suppress without losing anything:

````suggestion
git worktree remove --force "$audit_dir" 2>/dev/null; rm -rf "$(dirname "$audit_dir")"
````

If the preceding `git fetch` succeeds — the common case — this behaves identically, so the change is a strict improvement on the failure path only.

**Review by Tara-ag — APPROVED** [review]
APPROVE — documentation-only change to `CLAUDE.md`. All previously-raised findings are fixed and resolved at head `218533f3` (stale-ref, `--frozen-lockfile`, js-yaml values, and the fixed-worktree-path collision — the repro now uses a unique `mktemp` path with cleanup outside the `&&` chain). One new non-blocking MINOR on the trailing cleanup line.

**Comment by Tara-ag** [comment]
## APPROVE (head `218533f3`)

Documentation-only change to `CLAUDE.md` (1 file, +90). This is the canonical summary for the current head; prior summaries at earlier heads are superseded by it.

|---|---|---|---|
| resolved | `CLAUDE.md:586-589` | Repro used a fixed `/tmp/nl-audit` worktree path that failed on a second run and never cleaned up | FIXED at `218533f3` — unique `mktemp -d` path, `git worktree remove`/`rm -rf` moved outside the `&&` chain so a failed audit still cleans up |
| MINOR | `CLAUDE.md:589` | On `git fetch` failure the chain skips the worktree add, so the trailing `git worktree remove` prints a spurious `fatal: … not a working tree` on the failure path | NEW — suppress with `2>/dev/null` (suggestion posted) |

**Verified clean against source:**
- `scripts/security-check.ts` runs `pnpm audit --prod --json` live, un-pinned → "a green run expires" is accurate; failing output appends the "N commit(s) behind origin/release / may not originate in your changes" note.
- The `🔒 Security & Environment Validation` step sits in `test-shards (validate)` feeding the required `test` job via `needs:` — CI attribution correct.
- Repro now pins `origin/release` in a throwaway worktree and installs with `--frozen-lockfile` (matching every CI install), so the audited tree equals CI's.
- `package.json` overrides match the docs' now-explicitly-historical js-yaml example; `fast-xml-parser` / `undici` are genuine band-kept-in-step-with-floor examples; `files` excludes the lockfile so the "lockfile-only bump is not a fix" trap is accurate.
- **Prior findings all resolved at head:** stale-local-`release`-ref + `git stash` (fixed in `24731cced`), `--frozen-lockfile` (fixed in `79d5c3d41`), js-yaml stale values (fixed in `79d5c3d41` — example now historical).

No CRITICAL / MAJOR findings. Rule 5 (backward compat) and the security bar not applicable — docs only, no secrets, no code/config change. Review state matches the verdict: approving review on this head.

File: CLAUDE.md (hunk 9 above)
Comment (Tara-ag) [reply CLAUDE.md:594]: Deduplicated — this is the same finding as `yama:repro-cleanup-spurious-fatal` (kept at `CLAUDE.md:589` in the approving review). This thread is resolved as a duplicate; please track the one open thread. No substantive difference between the two suggestions (`2>/dev/null` vs `|| true` — either works).

**Comment by Tara-ag** [comment]
## APPROVE (head `772fdcb8d`)

Documentation-only change to `CLAUDE.md` (1 file, +102). This is the canonical summary for the current head; earlier summary comments are superseded by it.

|---|---|---|---|
| resolved | `CLAUDE.md:715-721` | Stale `release` ref + `git stash`; `pnpm install` vs CI's `--frozen-lockfile`; stale js-yaml override values; fixed `/tmp/nl-audit` path; spurious `fatal:` on fetch-failure cleanup; exit status masked by `rm -rf` | FIXED by author across `24731cced`/`79d5c3d41`/`218533f3`/`dac178f6`/`772fdcb8d` — all threads resolved |

The last open finding (`yama:repro-exit-status-masked`) is fixed at `772fdcb8d`: `audit_status=$?` is captured before cleanup and `( exit "$audit_status" )` restores the audit's status to `$?` for a scripted caller without closing the interactive shell (verified in the diff). Nothing new remains.

**Verified clean against source at head `772fdcb8d`:**
- `scripts/security-check.ts` runs `pnpm audit --prod --json` live, un-pinned → "a green run expires" is accurate; failing output appends the "N commit(s) behind origin/release / may not originate in your changes" note.
- The `🔒 Security & Environment Validation` step sits in `test-shards (validate)` feeding the required `test` job via `needs:` — CI attribution correct.
- Repro pins `origin/release` in a unique `mktemp -d` throwaway worktree, installs with `--frozen-lockfile` (matches every CI install), guarded cleanup outside the `&&` chain.
- `package.json` js-yaml overrides (`>=3.0.0 <3.15.2 → >=3.15.2`, `>=4.0.0 <4.3.2 → >=4.3.2`) sit on the GHSA-2883-xcg3-v3hh band edges as the doc states; `fast-xml-parser` / `undici` are genuine band-kept-in-step examples; `files` excludes the lockfile so the lockfile-only-bump trap is real; the `gh api graphql securityVulnerabilities` query shape is valid.

**Checked clean:** single-commit policy + Conventional Commit `docs(ci):` (both confirmed by CI); no code/config/lockfile drift; no secrets; rule 5 (backward compat) and the security bar not applicable — docs only.

No CRITICAL / MAJOR findings. An approving review is in place on the current head, matching the verdict.

> Note on dedup: due to the absence of a delete-comment primitive in the tooling used here, the historical per-pass summary comments at earlier heads (`#5598301229` … `#5648380402`) cannot be removed, but each already marks itself superseded by this summary and describes an earlier head — this comment is the single current summary.

**Review by Tara-ag — APPROVED** [review]
APPROVE at head `dac178f6` — documentation-only change to `CLAUDE.md`.

The previously-open finding (`yama:repro-cleanup-spurious-fatal`: cleanup printed a spurious `fatal:` on the fetch-failure path) is now resolved by the author: the cleanup is guarded with `[ -d "$audit_dir" ] && git worktree remove --force "$audit_dir"`, so nothing is removed (and no `fatal:` noise emitted) when the worktree was never created. Thread resolved.

All findings from prior passes are fixed and resolved at this head:
- stale local `release` ref + `git stash` → `origin/release` throwaway worktree
- `--frozen-lockfile` to match CI
- js-yaml example made explicitly historical with current values
- fixed `/tmp/nl-audit` path → unique `mktemp` directory with cleanup

Verified against source: `scripts/security-check.ts` runs `pnpm audit --prod --json` live; the security step lives in `test-shards (validate)` feeding the required `test` check; `package.json` overrides match the doc; `files` excludes the lockfile. No CRITICAL/MAJOR findings. Rule 5 (backward compat) and the security bar not applicable — docs only, no code/config change, no secrets.

File: CLAUDE.md [hunk 11]
```diff
@@ -677,6 +677,100 @@ docs:api`, then `pnpm --dir docs-site run build` — and fold the result into th
...
+audit_dir="$(mktemp -d)/nl-audit"
+git fetch origin release \
+  && git worktree add "$audit_dir" origin/release \
+  && ( cd "$audit_dir" && pnpm install --frozen-lockfile \
+       && pnpm exec tsx scripts/security-check.ts )
+# Guarded: if the fetch or the `worktree add` failed, there is nothing to
+# remove, and an unguarded remove prints `fatal: ... is not a working tree`
+# on exactly the failure path this section exists to de-confuse.
+[ -d "$audit_dir" ] && git worktree remove --force "$audit_dir"
+rm -rf "$(dirname "$audit_dir")"
```
Tara-ag, line 721: **MINOR** — a failed audit still ends the block with exit code 0, so `$?` never reflects the failure.

The audit chain (the `fetch && worktree add && (…)` list) returns non-zero when `security-check.ts` fails, but the two cleanup lines below run unconditionally and the block's final exit status is that of the last `rm -rf` — which is 0. So the pasted block reports success by exit code even when the gate is red. That matters for exactly this section's audience: someone who wraps the diagnostic in a script or checks `$?` and gets 0 on a failed audit, on a section whose whole purpose is not being misled.

Reading the PASS/FAIL line in the output is fine; the exit status is not. Capturing it before cleanup and re-emitting it fixes the scripted case (the author's own verified run already did this — `security_check_exit=0`):

````suggestion
       && pnpm exec tsx scripts/security-check.ts )
code=$?
# Guarded: if the fetch or the `worktree add` failed, there is nothing to
# remove, and an unguarded remove prints `fatal: ... is not a working tree`
# on exactly the failure path this section exists to de-confuse.
[ -d "$audit_dir" ] && git worktree remove --force "$audit_dir"
rm -rf "$(dirname "$audit_dir")"
exit "$code"
````

(For purely interactive use the `exit "$code"` can be dropped — but then the exit status is lost, so the `$code` capture only buys anything when a reader is scripting the diagnostic. Non-blocking, hence MINOR.)

All prior findings on this block are resolved at head — `mktemp -d` uniqueness, `origin/release`, `--frozen-lockfile`, and the `[ -d ]` guard for the fetch-failure path are all correct.

**Review by Tara-ag — COMMENTED** [review]
Documentation-only review — the advisory-gate section in CLAUDE.md. One new non-blocking MINOR; all prior findings confirmed resolved at this head.

File: CLAUDE.md [hunk 12]
```diff
@@ -677,6 +677,100 @@ docs:api`, then `pnpm --dir docs-site run build` — and fold the result into th
...
+# still cleans up after itself.
+audit_dir="$(mktemp -d)/nl-audit"
+git fetch origin release \
+  && git worktree add "$audit_dir" origin/release \
+  && ( cd "$audit_dir" && pnpm install --frozen-lockfile \
+       && pnpm exec tsx scripts/security-check.ts )
+# Guarded: if the fetch or the `worktree add` failed, there is nothing to
+# remove, and an unguarded remove prints `fatal: ... is not a working tree`
+# on exactly the failure path this section exists to de-confuse.
+[ -d "$audit_dir" ] && git worktree remove --force "$audit_dir"
```
Tara-ag, line 720: **MINOR** — the block's exit status is the audit's, not `rm -rf`'s.

If anyone runs this exact block from a script (not a REPL), `$?` is whatever the *last* command returns — and `rm -rf "$(dirname "$audit_dir")"` always returns 0, masking a red audit. The whole section exists to keep a red gate from being misread; a scripted copy reporting success on failure is the same class of trap.

For the interactive `&&`-chained form this is fine, but the section's whole point is a command whose verdict you can trust. Capturing and re-emitting the audit/install status before the cleanup keeps both properties:

```suggestion
audit_dir="$(mktemp -d)/nl-audit"
git fetch origin release \
  && git worktree add "$audit_dir" origin/release \
  && ( cd "$audit_dir" && pnpm install --frozen-lockfile \
       && pnpm exec tsx scripts/security-check.ts )
code=$?
[ -d "$audit_dir" ] && git worktree remove --force "$audit_dir"
rm -rf "$(dirname "$audit_dir")"
exit "$code"
```

(The `[ -d ]` guard and the `code=$?` capture are both outside the `&&` chain, so a failed fetch/worktree-add still propagates the right status; nothing else in the block changes.)

Tara-ag (approved):
APPROVE at head `4d60d2f` — documentation-only change to `CLAUDE.md`.

All prior findings are resolved at this head. One new non-blocking MINOR inline (`yama:repro-exit-code-0`): the block's exit status is `rm -rf`'s (0), so a scripted run would report success on a red audit — worth the exit-code capture for a command whose whole purpose is a trustable verdict.

No CRITICAL / MAJOR findings. Rule 5 (backward compat) and the security bar not applicable — docs only, no code/config change, no secrets.

CLAUDE.md
murdore: Right, and fixed in `772fdcb8d`.

Confirmed the masking before changing anything — a stand-in for a failed `security-check.ts`, with the block's cleanup lines intact:

```
BEFORE (failed audit):  $?=0   <- masked, reports success
AFTER  (failed audit):  $?=1   <- failure surfaces
AFTER  (passing audit): $?=0   <- still 0, no false alarm
```

I took the `audit_status=$?` capture as suggested, but not the bare `exit "$audit_status"` — you flagged the interactive cost yourself, and it is worse than "the status is lost": pasting a block ending in `exit` into a terminal closes that terminal. On a snippet whose audience is someone debugging a red gate by hand, that is a bad trade.

`( exit "$audit_status" )` gets both. The subshell exits, the parent shell does not, and `$?` is still the audit's status for a scripted caller:

```
( exit 1 ) -> shell survived, $?=1
exit 1     -> terminated it
```

So no interactive/scripted branch is needed, and the comment records why the subshell form is there rather than the bare one — otherwise it reads like a pointless wrapper and someone simplifies it back.

Tara-ag (approved):
APPROVE at head `772fdcb8d` — documentation-only change to `CLAUDE.md`.

The last open finding (`yama:repro-exit-status-masked`: the block's exit status was `rm -rf`'s 0, masking a red audit) is now **fixed** by the author in `772fdcb8d`: `audit_status=$?` is captured before cleanup, and `( exit "$audit_status" )` restores the audit's status to `$?` for a scripted caller without closing the interactive shell it was pasted into. Verified in the diff: `audit_status=$?` → guarded `[ -d ]` remove → `rm -rf` → `( exit "$audit_status" )`.

All findings across prior passes are resolved at this head:
- stale local `release` ref + `git stash` → `origin/release` throwaway worktree (`24731cced`)
- `--frozen-lockfile` to match CI (`79d5c3d41`)
- js-yaml example made explicitly historical with values matching `package.json` (`79d5c3d41`)
- fixed `/tmp/nl-audit` path → unique `mktemp -d` with cleanup outside the `&&` chain (`218533f3`)
- spurious `fatal:` on fetch-failure path → `[ -d ]` guard (`dac178f6`)
- exit-status masked by cleanup → `audit_status` + subshell exit (`772fdcb8d`)

Verified against source: `scripts/security-check.ts` runs `pnpm audit --prod --json` live; the `🔒 Security & Environment Validation` step lives in `test-shards (validate)` feeding the required `test` check; `package.json` js-yaml override bands sit on the GHSA-2883-xcg3-v3hh edges as the doc states; `files` excludes the lockfile so the lockfile-only-bump trap is accurate. No CRITICAL / MAJOR findings. Rule 5 (backward compat) and the security bar not applicable — docs only, no code/config change, no secrets.

PR #1677, opened 2026-09-10
feat(anthropic): truthful stream termination and an opt-in execution-control contract

murdore:
Two additive changes to the native Anthropic stream path.

## 1. Terminal truth

A turn that the caller aborted, or that ran out its own time, reported a normal stop reason. A consumer could not tell a cancelled turn from one that finished on its own, so a stopped run looked like a completed one.

The agentic loop now carries whether it was aborted, and the stream path grades a turn from that fact plus the composed signal's reason: a timeout reports `"time-limit"`, any other abort reports `"aborted"`, and both settle `finishReason` as `"other"` rather than falling through to the step-cap branch. An unaborted turn still reports `"stop"`.

The reason is read off the *merged* signal, which preserves it — the engine's own internal abort drops its reason, so reading that instead would make a timeout indistinguishable from a cancel.

## 2. An opt-in `executionControl` contract

For callers that need to run as long as real progress continues rather than to a fixed ceiling. It carries a required finite `requestTimeoutMs`, a `lifetimeTimeoutMs` where `null` means no lifetime timer and a finite value an explicit cap, and a `beforeStep` callback that runs at a safe step boundary after settled tool results and may renew a finite step cap in the same loop.

It is **absent by default and legacy consumers are untouched**. A provider that cannot honour it rejects the option rather than ignoring it. An explicitly-`undefined` `lifetimeTimeoutMs` inherits the legacy handling rather than silently removing the ceiling, and combining it with `turnTimeoutMs` is rejected rather than dropping one in silence.

## ⚠️ Release-note item: tool execution is now bounded on every provider

Because no lifetime timer can be in force, tool execution needed a bound of its own. The agentic loop now applies `toolTimeoutMs` to every tool call, defaulting to **five minutes**.

- **Vertex / Gemini** — unchanged in effect. Their own `toolTimeoutMs` is threaded through at the same value, and their existing guard fires first by evaluation order.
- **AI Studio and Bedrock** — these had **no per-tool bound at all** and now inherit 300 000 ms. **A tool that legitimately runs longer than five minutes on either now fails with a timed-out tool result where it previously completed.** The escape hatch is `toolTimeoutMs`, and the default matches what that option already documents.

A breach is reported back to the model as a tool result, so the turn continues rather than dying.

## Why `feat(` and not `fix(`

This adds a public API (`executionControl`), so semantic-release should cut a **minor**, not a patch. The terminal-truth half alone would have been a `fix`.

## Testing

- `test:anthropic-loop-characterization` — 13/13. The suite now declares itself `offline`, so a hang **fails** instead of being downgraded to a skip. That mattered: the timeout case's regression mode *is* a hang.
- `test:anthropic-execution-control` — 16/16, new suite, wired into CI beside its sibling.
- Both run against a local SSE stand-in and need no API keys.
- `pnpm run build` clean with publint; lint at the repo's existing baseline (0 errors, 58 warnings).

Every new case was verified to fail against the pre-fix source rather than merely to pass against the fixed one.

## Notes for the reviewer

- `loopAdapter.ts` looks large in the diff but is mostly a near-verbatim extraction of the old `executeStep` body to module scope; the substantive edits inside it are two new guards and a `continue` so a `message_stop` branch can follow `message_delta`.
- `withTimeout` no longer leaves its timer pending after the race is decided — it previously never cleared or unref'd the handle.

## Summary by CodeRabbit

- **New Features**
  - Added opt-in execution controls for native Anthropic streaming, including request and lifetime timeouts, step limits, callbacks, and planning nudges.
  - Added configurable per-tool execution timeouts across supported agentic providers, including unbounded execution.
  - Added clearer reporting for aborted, time-limited, and interrupted turns.
- **Bug Fixes**
  - Improved incomplete Anthropic stream handling by detecting missing terminal events.
  - Improved timeout cleanup to prevent unnecessary timers from keeping processes active.
- **Validation**
  - Added validation for unsupported or conflicting execution-control configurations.
- **Documentation**
  - Expanded API documentation for execution controls and agentic loop options.


Tara-ag (commented):
Review of PR 1677 (native Anthropic terminal-truth + executionControl).

src/lib/core/loopEngine.ts
```diff
@@ -373,6 +399,8 @@ export function runAgenticLoop<TConversation>(
           tools: options.tools,
           failedTools,
           abortSignal: internalAbort.signal,
+          toolTimeoutMs:
+            options.toolTimeoutMs ?? DEFAULT_TOOL_EXECUTION_TIMEOUT_MS,
         });
```
Tara-ag, line 486: **MAJOR** — the 300 s per-tool bound is now unconditional and has no opt-out, which silently changes existing callers.

`DEFAULT_TOOL_EXECUTION_TIMEOUT_MS` (300_000) is applied to *every* `tool.execute()` on every provider that rides `runAgenticLoop`. Per the PR's own release note, **Bedrock and AI Studio previously had no per-tool bound at all** — a tool that legitimately runs longer than five minutes now fails with a timed-out tool result where it previously completed. The note documents this, but the API gives a caller no way to restore the old (unbounded) behaviour: `toolTimeoutMs` accepts only a finite positive number, so `Infinity`/`0`/`null` cannot express "no bound" (and `Infinity` would desugar to `setTimeout`'s 2^31ms cap rather than meaning "never").

Rule 5 (backward compatibility of the public SDK surface) is about more than types compiling — a tool that broke mid-turn on a 5-minute deadline is a runtime regression for those providers. Suggest mirroring the `lifetimeTimeoutMs: null` pattern: accept `toolTimeoutMs: null` (add it to the option type and to `validateExecutionControl`-style validation) to mean "no per-tool bound", and keep the default for callers that don't opt in.

Affected out-of-diff callers: `amazonBedrock/client.ts` (`conversationLoop`, `streamingConversationLoop`) and `googleAiStudio/client.ts`, which previously ran their own tool executors unguarded.

src/lib/core/loopEngine.ts
```diff
@@ -158,10 +163,26 @@ async function dispatchStepTools(params: {
...
+      // between two steps there is no other timer running: the step's request
+      // deadline was disposed when the step settled and the next one is not
+      // armed yet. A tool that neither returns nor watches its signal would
+      // otherwise hang the whole turn with nothing to end it — the case a
+      // turn that opted out of a lifetime ceiling has no defence against.
+      //
+      // A breach lands in the catch below and is recorded as an ordinary tool
+      // failure, so the turn spends one step and carries on, matching what
+      // `toolTimeoutMs` already means on the native generate path.
+      const output = await withTimeout(
```
Tara-ag: **MINOR** — on timeout the underlying tool is not cancelled, only its awaited promise is discarded.

`withTimeout` rejects at `toolTimeoutMs`, but the `abortSignal` handed to `tool.execute(call.args, { abortSignal })` is the turn's `internalAbort` signal, which the per-tool deadline never fires. So a wedged but still-live tool keeps executing in the background after the turn has recorded it as failed — it may keep holding resources and, for a side-effecting tool, still apply its effect after the model was told it failed. If the model then retries the same call, a second concurrent copy starts.

The release note calls this the "bounded ghost" tradeoff, which is acceptable for a hang, but consider racing the call against a per-tool `AbortController` that is aborted on the deadline so the tool is actually told to stop, rather than only abandoned.

Tara-ag:
## Verdict: **NEEDS_WORK** (on one MAJOR)

Solid, well-tested additive change — the terminal-truth grading and the `executionControl` contract are coherent, opt-in, and behind a "reject rather than ignore" barrier. The main open item is the unconditional 5-minute per-tool bound applied to providers that never had one, with no opt-out.

### Findings

| MAJOR | `src/lib/core/loopEngine.ts:404` | 300 000 ms per-tool bound is now applied to **every** provider on `runAgenticLoop` (Bedrock & AI Studio previously unguarded) with **no opt-out sentinel** — a legitimate >5-min tool now fails where it used to complete, and `toolTimeoutMs` cannot express "no bound". Suggest a `null` sentinel mirroring `lifetimeTimeoutMs: null`. |
| MINOR | `src/lib/core/loopEngine.ts:176` | On tool timeout only the promise is discarded; the tool's own execution (and its side effects) are not cancelled, so a retried call can run concurrently with the ghost. |

- **Terminal-truth grading**: abort → `"aborted"`, timeout → `"time-limit"`, clean turn → `"stop"`; reading the reason off the *merged* signal so a turn timeout is distinguishable from a cancel. Consistent in `loopAdapter.ts` (`readAnthropicStep`), `loopEngine.ts` (`runAgenticLoop` aborted flag) and `anthropic/client.ts` (`runLoop` grading).
- **`executionControl` contract**: `validateExecutionControl` correctly rejects unsupported providers and ambiguous shapes; `lifetimeTimeoutMs:null|finite|absent(undefined)` semantics are handled with `!== undefined` rather than `in`; `turnTimeoutMs`+explicit `lifetimeTimeoutMs` is rejected. Anthropic is the only `supportsExecutionControl()===true` provider; all others reject.
- **`beforeStep` bounded and cancellable** in the client via `createTimeoutController`+`raceWithAbort`, with the default `DEFAULT_BEFORE_STEP_TIMEOUT_MS`; the engine-side step-cap renewal (finite, strictly-larger, floored) is correct and well-tested (fractional-renewal guard included).
- **`withTimeout` timer cleanup** in `utils/timeout.ts` (clear + unref); the engine's `withTimeout` use resolves to the correct `async/withTimeout` signature.
- **Provider tool-timeout claims**: `runAgenticLoop` is the shared path for Bedrock, AI Studio, Vertex and Anthropic, so "every provider" is accurate; Vertex/Gemini thread their own `toolTimeoutMs` first by evaluation order.
- **Tests**: 13/13 (`loop-characterization`, now `offline`) + 16/16 (`execution-control`), both wired into CI, e2e-driving `dist/`, no API keys.
- No security issues (no secrets, no injection, no unsafe handling of model/user input).

**Recommendation**: accept the approach; unblock on allowing callers of Bedrock/AI Studio to restore an unbounded per-tool behaviour (`toolTimeoutMs: null`), or explicitly document which callers will regress and confirm it is intended for a `feat(` minor release.

Tara-ag:
## NEEDS_WORK

Solid, well-tested change — terminal-truth grading, the `executionControl` contract, the `lifetimeTimeoutMs` inheritance fix, `beforeStep` bounding, and the `withTimeout` timer cleanup are all correct. Two findings stand in the way of approval, both on the shared loop engine.

| Sev | File:line | Finding |
|-----|-----------|---------|
| MAJOR | `src/lib/core/loopEngine.ts:404` | 300 s per-tool bound is now unconditional with no opt-out — silently breaks Bedrock/AI Studio callers whose tools previously ran unbounded (Rule 5). |
| MINOR | `src/lib/core/loopEngine.ts:176` | On tool timeout the underlying execution isn't cancelled, only its awaited promise is abandoned — a retried call can race the ghost. |

### Checked and clean
- Terminal-truth grading: `aborted && rawStopReason === undefined ? "other" : finishReason` and the `aborted` flag propagation in `loopEngine.ts` — correctly distinguishes abort/timeout from a normal stop, keeps real provider stop reasons.
- `loopAdapter.ts` `readAnthropicStep` / per-request deadline / truncated-stream guard — bounded correctly.
- `anthropic/client.ts` `executionControl` wiring, `beforeStep` bounding, aborted/time-limit grading on the Anthropic path.
- `validateExecutionControl` in `parameterValidation.ts` — rejects unsupported providers rather than ignoring the control (matches the contract's stated intent).
- Both `withTimeout` implementations — timer properly cleared/unref'd, no leak.
- `supportsExecutionControl()` default-false reject-not-ignore stance in `baseProvider.ts`.
- Two new e2e suites (`test:anthropic-execution-control`, `test:anthropic-loop-characterization` with `offline: true`) wired into CI — cover the new behavior end to end.
- Single commit, conventional message with required scope — compliant (also confirmed by the single-commit bot). No secrets, no logging of provider params.

Tara-ag (changes requested):
NEEDS_WORK — verdict aligned with the two inline findings (MAJOR `loopEngine.ts:404` per-tool bound no opt-out; MINOR `loopEngine.ts:176` orphaned tool execution on timeout). The terminal-truth and executionControl work is otherwise sound. See the summary comment for the full table and what was checked clean.

Tara-ag:
## Verdict: **NEEDS_WORK** (on one MAJOR)

Recurring review — the branch is still the same single commit (`8264520`, no force-push) and there is **no author reply on either of my two inline findings**, so both stand. The code has not changed since the prior review, so I re-verified each finding against the current sources rather than re-deriving them fresh.

### Findings (unchanged, still open)

| MAJOR | `src/lib/core/loopEngine.ts:404` | The 300 000 ms per-tool bound is now unconditional with **no opt-out**. `dispatchStepTools` runs every `tool.execute()` through `withTimeout(..., toolTimeoutMs ?? DEFAULT_TOOL_EXECUTION_TIMEOUT_MS)`, and `toolTimeoutMs?: number` (stream.ts) cannot express "no bound" (no `null` sentinel, and `Infinity` would desugar to setTimeout's 2³¹ ms cap). Per the PR's own note, **Bedrock and AI Studio previously had no per-tool bound** — a legitimate >5-min tool now fails with a timed-out result where it used to complete. Rule 5 (backward compat of the public surface) is about more than types compiling. | **Kept** — no author reply; `constants.ts`, `loopEngine.ts`, `stream.ts` confirm it's unchanged. The PR's extensive documentation *discloses* the change but does not give callers an escape hatch. |
| MINOR | `src/lib/core/loopEngine.ts:176` | On tool timeout only the awaited promise is discarded; the tool's own execution is **not cancelled** — the `abortSignal` passed to `tool.execute` is the turn's merged signal, which the per-tool deadline never fires. A wedged/side-effecting tool keeps running (a "bounded ghost") and a model retry can start a second concurrent copy. `raceWithAbort`'s own docstring acknowledges this "bounded ghost" tradeoff. | **Kept** — no author reply; confirmed in `loopEngine.ts` and `withTimeout.ts`. Acceptable for a hang, but a per-tool `AbortController` raced against the deadline is the safer shape. |

### Checked and clean (this pass)
- **Terminal-truth grading** (`loopAdapter.ts` / `anthropic/client.ts`): abort → `stopReason "aborted"` + `finishReason "other"`; timeout → `"time-limit"`; clean turn → `"stop"`. The merged-signal `reason instanceof TimeoutError` disambiguation is correct.
- **`executionControl` contract**: `supportsExecutionControl()` is `true` only on Anthropic; `validateExecutionControl` (parameterValidation.ts) rejects unsupported providers, rejects `turnTimeoutMs` combined with explicit `lifetimeTimeoutMs`, and treats `lifetimeTimeoutMs: undefined` as "inherit" via `!== undefined` (not `in`).
- **`beforeStep` is genuinely bounded and cancellable** in `anthropic/client.ts` (`createTimeoutController(budgetMs)` + `raceWithAbort` on a composed signal; default 30 s; failure declines the renewal, does not kill the turn); the engine-side step-cap renewal floors fractional values and only applies strictly-larger finite caps.
- **`withTimeout` timer cleanup**: both the `async/withTimeout.ts` used by the engine and the `timeout.ts` one clear the handle in `finally`; no pending-timer leak on the operation-wins path.
- **New tests** are e2e (drive `dist/`, no API keys), wired into CI; the `offline: true` declaration on the loop-characterization suite correctly makes a hang a failure rather than a skip.
- Rule 1 (registry) and Rule 3/6 not touched; no secrets, injection, or unsafe handling of model/user input introduced.

### Also noted
- CodeRabbit has one valid secondary point on the new test file: it interpolates recovered provider-shaped error text into `assert` messages, which `defineSuite`'s `isExpectedProviderError()` heuristic could classify as SKIP and keep the suite green on a real regression. That is CodeRabbit's thread; worth folding in but I did not repost it as a duplicate inline.

**Recommendation**: accept the approach and the `executionControl` contract. Unblock on giving Bedrock/AI Studio callers a way to restore an unbounded per-tool behaviour (`toolTimeoutMs: null` sentinel mirroring `lifetimeTimeoutMs: null`), or explicitly list the callers that will regress and confirm that is intended for a `feat(` minor under Rule 5.

(this consolidated summary supersedes the two earlier `yama:summary` comments from the prior passes of the same PR; the code is unchanged so the verdict is unchanged.)

Tara-ag:
## Verdict: **NEEDS_WORK** (recurring — both prior findings still stand)

Re-review of the current head (`7d26f98`, which is a force-push over the `8264520` the prior pass reviewed). I re-read the relevant sources at the current SHA and confirmed the two earlier inline findings are **unchanged and un-addressed**; there is **no author reply on either thread** that would justify retracting them.

### Findings (kept)

| MAJOR | `src/lib/core/loopEngine.ts:404` (dispatch) | The 300 000 ms per-tool bound is now **unconditional with no opt-out**. `runAgenticLoop` passes `options.toolTimeoutMs ?? DEFAULT_TOOL_EXECUTION_TIMEOUT_MS` into every `dispatchStepTools`, and `StreamOptions.toolTimeoutMs?: number` (stream.ts) / `AgenticLoopOptions` accept **no `null` sentinel** — `Infinity`/`0` cannot express "no bound" (`Infinity` would desugar to `setTimeout`'s 2³¹ ms cap). `validateExecutionControl` does not touch `toolTimeoutMs`. The new provider wiring makes the blast radius explicit: Bedrock (`streamingConversationLoop` → `toEngineTools` "adds no deadline of its own, so this is the **only per-tool bound** on the turn") and AI Studio (native loop "guards no executor of its own … the engine's bound has to be reachable") **previously had no per-tool deadline** (Rule 5: a legitimate >5-min tool now fails with a timed-out result where it used to complete). | **Kept** — no author reply, code/type unchanged. |
| MINOR | `src/lib/core/loopEngine.ts:176` | On tool timeout only the awaited promise is discarded; the underlying tool is **not cancelled**. `tool.execute(call.args, { abortSignal })` receives the turn's `internalAbort` signal, which the per-tool `withTimeout` deadline never fires — confirmed by Bedrock's `toEngineTools` wrapper, which doesn't even forward an abort signal. A wedged/side-effecting tool keeps running in the background (a "bounded ghost") and a model retry can start a second concurrent copy. | **Kept** — no author reply, code unchanged. |

### Checked and clean (this pass)
- **Terminal-truth grading**: `loopEngine.ts` (`aborted` flag read after the loop) + `anthropic/client.ts` grading — abort → `stopReason "aborted"` / `finishReason "other"`; timeout → `"time-limit"`; clean turn → `"stop"`. Correct and well covered (loop-characterization `offline` now makes a hang a failure).
- **`executionControl` contract**: `validateExecutionControl` (parameterValidation.ts) rejects unsupported providers (`UNSUPPORTED_PROVIDER`, Anthropic-only), rejects `turnTimeoutMs` combined with an explicit `lifetimeTimeoutMs`, treats `lifetimeTimeoutMs: undefined` as inherit (via `!== undefined`), and validates `beforeStep`/`beforeStepTimeoutMs`. `beforeStep` renewal is floored, finite, strictly-larger, and the nudge lands in the same history. Backing tests present.
- **`withTimeout` timer cleanup** (utils/async) — handle cleared/unref'd; no pending-timer leak on the operation-wins path.
- **e2e discipline**: new execution-control suite drives `dist/`, offline, wired into CI; no API keys. Rule 15 satisfied.
- No secrets, injection, or unsafe handling of user/model input; Rule 1/3/6 untouched.

### Also noted
- CodeRabbit's thread on `continuous-test-suite-anthropic-execution-control.ts` — the current file still interpolates raw provider-shaped error text into `assert` messages (e.g. `the rejection must name executionControl, said: ${…slice(0,200)}`). This is a real masking risk under `isExpectedProviderError()`. It is CodeRabbit's thread; I am not duplicating it inline, but the author should fold it in before merge.

**Recommendation**: accept the approach and the opt-in `executionControl` contract. Unblock on either (a) giving Bedrock/AI Studio callers a way to restore unbounded per-tool behaviour (`toolTimeoutMs: null` sentinel mirroring `lifetimeTimeoutMs: null`, wired into `AgenticLoopOptions`/`dispatchStepTools`), or (b) explicitly listing the callers that will regress and confirming that is intended for a `feat(` minor under Rule 5.

This is the consolidated statement for the current review; it supersedes nothing that is not already consistent with the prior `yama:summary` comments (which remain the record for earlier iterations).

Tara-ag:
## Verdict: **APPROVE** — both prior findings resolved by the force-push to `5c1fae7`

The branch was force-pushed (over the `7d26f98` the previous pass reviewed) with a commit that directly addresses both open findings. I re-read the current sources end-to-end; there was no author reply on either thread, but the code itself is what changed, so I **resolved both inline threads** rather than reposting them.

### Findings — prior → current

| Sev | Prior finding (loopEngine.ts) | Status now | Evidence in current code |
|-----|------------------------------|-----------|--------------------------|
| MAJOR | 300 s per-tool bound applied to every provider with **no opt-out** (Rule 5 regression for Bedrock/AI Studio). | ✅ **Resolved** | `StreamOptions.toolTimeoutMs?: number \| null` (`stream.ts:580`); `resolveToolTimeoutMs()` three-way distinction in `constants.ts`; `executeToolCall` null branch awaits unguarded (`loopEngine.ts`); `validateExecutionControl` **refuses** `toolTimeoutMs:null`+`lifetimeTimeoutMs:null` before any request; Bedrock/AI Studio/Vertex/Anthropic all thread `toolTimeoutMs` (incl. `null`) into `runAgenticLoop`. Typechecker-enforced (dropping the null branch won't compile). |
| MINOR | Tool timed out but underlying execution orphaned (not cancelled). | ✅ **Resolved** | `executeToolCall` races each call against a **per-call `AbortController`** the deadline aborts with the timeout as the abort reason; turn abort is forwarded into it; `Promise.race` + `finally` clears timer/listener; no unhandled rejection. The "tool ignores its signal" limit is explicitly documented. |

### What was checked and clean (this pass)
- **Terminal-truth grading**: engine sets `aborted` + `finishReason = aborted && rawStopReason===undefined ? "other" : finishReason`; Anthropic client reads the **merged** signal's reason — `reason instanceof TimeoutError` → `"time-limit"`, else `"aborted"` — and `finishReason:"other"`, returning before the step-cap branch. Clean turn still reports `"stop"`. Correct.
- **Missing-terminal-event guard**: `readAnthropicStep` treats a stream that ended before `message_stop` as `ANTHROPIC_STREAM_TRUNCATED` (non-retriable), so a truncated turn can't report a normal stop; the per-request deadline's `TimeoutError` throws before that check. Correct.
- **Contract validation**: `validateExecutionControl` rejects unsupported providers (`UNSUPPORTED_PROVIDER`, Anthropic-only), rejects `turnTimeoutMs`+explicit `lifetimeTimeoutMs`, treats `lifetimeTimeoutMs:undefined` as inherit (via `!== undefined`, not `in`), validates `beforeStep`/`beforeStepTimeoutMs`. Reject-not-ignore throughout.
- **`beforeStep` bounded and cancellable** in the client (`createTimeoutController` + `raceWithAbort`, default `DEFAULT_BEFORE_STEP_TIMEOUT_MS=30s`); failure/throw declines the renewal, never kills the turn. Engine-side step-cap renewal is floored, finite and strictly-larger.
- **Provider wiring** complete: Bedrock `streamingConversationLoop` and AI Studio both forward `toolTimeoutMs` (with comments documenting that the engine's bound is now their only per-tool watch). Compare `runAgenticLoop` dispatch for all four.
- **Tests**: `loop-characterization` (13/13, now `offline:true` so a hang fails) + `execution-control` (16/16) drive `dist/`, offline SSE stand-in, wired into CI. CodeRabbit's test-hygiene point (interpolating provider-shaped error text into asserts) is also folded in — assertions now use structural booleans/counts.
- **ESLint**: the new `e2e-tests-only` entry for the execution-control suite declares its determinism exception in both the allow-list and (per the suite) its file header — for `resolveToolTimeoutMs`, "no bound" vs "300s default" are indistinguishable from outside a 60 s live call. Compliant with Rule 15.
- No security issues: no secrets, no injection, no unsafe handling of model/user input (Rule 1/3/6 untouched; registry not involved).

### Informational (not blocking)
- The per-call tool deadline timer in `executeToolCall` is deliberately **not** `unref`'d so it can hold the loop open to fire on a wedged tool (up to the default 300 s). Correct and documented, but a library-embedding caveat worth knowing.

**Recommendation**: accept. The opt-in `executionControl` contract plus the `toolTimeoutMs:null` escape hatch fully address the Rule 5 concern and the abandonment concern. Both prior review threads resolved; no new findings.

(this consolidated summary supersedes the `yama:summary` comments from prior passes; the code is now the force-pushed `5c1fae7`.)

Tara-ag (approved):
APPROVE — both prior review findings (unconditional 300s per-tool bound with no opt-out; orphaned tool execution on timeout) are resolved by the force-push to `5c1fae7` via the `toolTimeoutMs: null` sentinel and per-call `AbortController` cancellation. Full detail in the `yama:summary` comment. This approving review supersedes the earlier CHANGES_REQUESTED, whose two inline threads have been resolved.

Tara-ag:
## Verdict: **APPROVE** — recurring review of head `413511b`

This is a re-review of the force-pushed head `413511b`, which is a delta over the previously-approved `5c1fae7`. Per CodeRabbit's change-set and the author's commit note, the only files changed between `5c1fae7` and `413511b` are the two test files — **the production code is unchanged since the prior APPROVE**, so both prior yama findings stay resolved.

### Prior yama findings — still resolved (production unchanged)

| Sev | Prior finding | Status | Evidence in current `413511b` |
|-----|---------------|--------|-------------------------------|
| MAJOR | 300 s per-tool bound unconditional, no opt-out (Rule 5 for Bedrock/AI Studio) | ✅ Resolved | `toolTimeoutMs: number \| null` + `resolveToolTimeoutMs()` three-way split (`constants.ts`); `executeToolCall` null branch awaits unguarded (`loopEngine.ts`); `validateExecutionControl` refuses `null`+`lifetimeTimeoutMs:null`; Bedrock/AI Studio/Vertex/Anthropic all thread `toolTimeoutMs` incl. `null`. |
| MINOR | Tool timed out but execution orphaned (not cancelled) | ✅ Resolved | `executeToolCall` races each call against a per-call `AbortController` the deadline aborts; turn abort forwarded in; timer/listener cleared in `finally`. |

### This delta's changes — both target CodeRabbit's open threads

- **Abort-position grading** (`loop-characterization`): the new "an abort is graded by what the stream delivered…" case pins in-flight abort → `stopReason "aborted"` / `finishReason "other"` (no `rawFinishReason`), and post-drain abort → `"stop"` + `end_turn` preserved. That is the reviewer's `normalTerminalCompletion` scenario, and it already behaves correctly without the flag. On the native Anthropic path the client's `runLoop` unconditionally overrides `finishReason` to `"other"` on `result.aborted` and returns before the step-cap branch, so an interrupted tool-turn cannot leak as `"tool-calls"` to an `executionControl` consumer.
- **Truncated-stream error surfacing** (`execution-control`): the case now captures `errorSeen`/`errorNamedTruncation` and assert the error actually reached the consumer — closing the "regression det–ects truncation but swallows the error" hole noting `finishReason !== "stop"` was already satisfiable by `undefined`.

### Checked and clean
- **Terminal-truth grading** (`loopEngine.ts` `aborted && rawStopReason === undefined ? "other" : finishReason` + `anthropic/client.ts` reading the merged signal's `TimeoutError` → `"time-limit"` vs `"aborted"`). Timeout/reason preserved off the merged signal. Correct.
- **`executionControl` contract** (`parameterValidation.ts`): reject-not-ignore for unsupported providers (Anthropic-only), `turnTimeoutMs`+explicit `lifetimeTimeoutMs` rejected, `lifetimeTimeoutMs: undefined === inherit` via `!== undefined` (not `in`). `beforeStep` bounded + cancellable in the client. Engine step-cap renewal floored/finite/strictly-larger. Correct.
- **`withTimeout` timer cleanup** — handle cleared/unref'd.
- **Tests** are e2e (drive `dist/`, local SSE stand-in, no API keys), `offline: true` on the characterization suite makes a hang a failure; both wired into CI alongside CI-configured extended suites. Rule 15 satisfied.
- **CI**: all completed checks pass (lint / types / validate / build / security / CodeQL / Single-Commit / docs). The `blocked` mergeable state is only because the Yama review + CodeRabbit are still in progress / remaining non-blocking shards are running — no code, type, or security failure blocks the change. No secrets, no injection, no unsafe handling of model/user input (Rule 1/3/6 untouched).

### One honest caveat (impact analysis skipped — no local graph)
I could not run the code-review-graph here (no local checkout), and the Gemini native stream/Vertex loop bodies are large, so I did **not** fully trace whether a tool-turn abort leaves the engine's internal `result.finishReason = "tool-calls"` visible to **Gemini/Vertex** consumers. It does not leak on the native Anthropic path (the only consumer of `executionControl`), which is the behavior this PR changes. If you want certainty for the Google paths, that one corner is worth a targeted check before merge — I am deliberately not asserting either way.

**Recommendation**: accept. Both prior findings resolved and unchanged; the delta strengthens the two test cases CodeRabbit flagged. No new findings on this head.

(this consolidated summary supersedes the earlier `yama:summary` comments for prior passes of the same PR; the code under review is now `413511b`.)

Tara-ag (approved):
APPROVE — recurring review of head `413511b`. Production code is unchanged since the previously-approved `5c1fae7`; both prior findings (unconditional per-tool bound / orphaned tool execution) remain resolved via `toolTimeoutMs: null` and the per-call `AbortController`. The delta is strictly test additions that resolve CodeRabbit's two threads: the abort-position characterization case and the truncated-stream error-surfacing assertions. Full table and the one deliberately-unverified corner (Gemini/Vertex tool-turn abort grading — impact analysis skipped, no local graph) are in the `yama:summary` comment. No new findings.

PR #1678, opened 2026-09-10
fix(bedrock): honour input.text in the native generate path

murdore:
## Recurring review — Verdict: **APPROVE** (both prior MINORs now resolved)

The fix correctly closes the silent-empty-string defect: the native Bedrock `generate` path now reads both the `prompt` and `input.text` shapes as `options.prompt || input?.text || ""`, so a direct `AIProviderFactory.createProvider()` caller passing the documented `input.text` shape no longer sends Bedrock a blank user turn. On the latest commit `c4d30fa7` both of the previously-open MINOR findings are addressed by the author's changes and their replies, so both threads have been resolved.

### Findings — both RESOLVED (previously MINOR)

|----------|-----------|---------|--------|
| MINOR | `src/lib/providers/amazonBedrock/client.ts:311` | `??` diverged from the base contract's `||` on the empty-`prompt` + populated-`input.text` edge (base sends `input.text`, this sent `""`). | **resolved** — author changed to `||`, annotated why `||` and not `??` (matches `normalizeTextOptions`); reply `discussion_r3995006147`. Thread resolved. |
| MINOR | `test/continuous-test-suite-provider-wiring.ts` | Provider-level cases were mutually-exclusive shapes, so `prompt`-over-`input.text` precedence was never asserted. | **resolved** — author added case `3b` (both shapes, different text, asserts `prompt` wins) and case `3c` (`{ prompt: "", input: { text } }`, the `||` discriminator), with red/green proof. Thread resolved. |

### Checked and found clean
- **Precedence direction** — `options.prompt || input?.text || ""` matches the base contract (`normalizeTextOptions` / `handleVideoGeneration` use `||`, `prompt` first); ordering and `||`-vs-`??` semantics now correct.
- **Regression value** — the local h2c HTTP/2 Bedrock endpoint exercises the real client, SigV4 and serialization, asserts the actual wire bytes; reverting the fix fails case 2 (`provider-level input.text did not reach the wire (length 0)`). No AWS call; credentials are documented placeholders, not leaked secrets.
- **Backward compatibility (rule 5)** — `generate`'s signature/return/async shape unchanged; the `executeStream`→`generate` fallback (`client.ts`) passes `prompt: options.input.text` explicitly, so no unmodified caller's behavior changes.
- **Conventions (rule 15)** — single conventional-commit `fix(bedrock):`; the new suite imports only `dist/` / `.js` modules (one module graph per suite).
- **Blast radius** — change is a single-line input-shape resolution inside the `generate` override; no flow or external caller impact.
- **Secrets** — `AKIALOCALENDPOINTONLY` / `local-endpoint-secret-not-real` are placeholders; h2c listener deliberately avoids shipping a TLS key pair.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed text-only Amazon Bedrock requests so caller-provided text is sent when no prompt is provided or the prompt is empty.
  - Preserved prompt precedence when both prompt and text are supplied.

- **Tests**
  - Added coverage across SDK, streaming, provider, and CLI interfaces to verify that caller text reaches Amazon Bedrock correctly.
  - Added request-level validation for both standard and streaming responses.


src/lib/providers/amazonBedrock/client.ts
```diff
@@ -297,10 +297,13 @@ export class AmazonBedrockProvider extends BaseProvider {
         `[AmazonBedrockProvider] Text-only input in generate(), using simple message builder`,
       );
 
-      // Add user message to conversation - simple text-only case
+      // Add user message to conversation - simple text-only case.
+      // The public generate contract accepts either `prompt` or `input.text`
+      // (see BaseProvider); reading only `prompt` here sent an empty user
+      // message whenever a caller used the `input.text` shape.
       const userMessage: BedrockMessage = {
         role: "user",
```
Tara-ag, line 305: **MINOR** — `??` doesn't match the base contract's `||` on empty-string prompts.

`this.generate()` fully overrides `BaseProvider.generate`, so the raw (un-normalized) options reach this line. The base's canonical resolution is `||` everywhere, not `??`:
- `Utilities.normalizeTextOptions`: `const prompt = optionsOrPrompt.prompt || optionsOrPrompt.input?.text || "";`
- `BaseProvider.handleVideoGeneration`: `const prompt = options.prompt || options.input?.text || "";`

With `??`, an empty-string `prompt` short-circuits, so `{ prompt: "", input: { text: "hello" } }` sends `""` — while the base contract resolves `"hello"`. Since the PR's stated goal is "same precedence the base class uses," use `||` for exact parity:

```suggestion
        content: [{ text: options.prompt || input?.text || "" }],
```

The current code is correct for all non-empty cases (and both new test shapes pass), so this only surfaces on the empty-`prompt` + populated-`input.text` edge — hence MINOR, not a blocker.

test/continuous-test-suite-provider-wiring.ts
```diff
@@ -579,4 +579,112 @@ await test("catalog-provider enum surfaces are byte-identical to the pre-JSON-mi
...
+  const { ProviderRegistry } = await import("../dist/index.js");
+  await ProviderRegistry.registerAllProviders();
+  const { ProviderFactory } =
+    await import("../dist/factories/providerFactory.js");
+
+  const PROMPT = "Reply with exactly OK.";
+  const shapes: Array<{ input: { text: string } } | { prompt: string }> = [
+    { input: { text: PROMPT } },
+    { prompt: PROMPT },
+  ];
```
Tara-ag: **MINOR** — the regression test never exercises the precedence the fix ships.

The two shapes are mutually exclusive (`{ input: {...} }` and `{ prompt: ... }`), so both `options.prompt ?? input?.text ?? ""` and an inverted `options.input?.text ?? options.prompt ?? ""` would produce identical wire bytes here. The precedence claim in the PR description ("in the same precedence the base class uses" / "precedence when both are provided") is therefore untested — a future inversion to `input.text`-first would pass CI silently.

Consider a third case that supplies both with different values and asserts `prompt` wins (matching `normalizeTextOptions`), e.g.:

```suggestion
  const shapes: Array<
    | { input: { text: string } }
    | { prompt: string }
    | { prompt: string; input: { text: string } }
  > = [
    { input: { text: PROMPT } },
    { prompt: PROMPT },
    { prompt: PROMPT, input: { text: "different-text-that-must-lose" } },
  ];
```

This locks the `prompt`-precedence behaviour the description advertises. MINOR — the primary defect (both shapes reaching the wire as the caller's text) is well covered and would regress correctly.

Tara-ag (approved):
Fix is correct: the native Bedrock `generate` path now honours `input.text` with `prompt` precedence, matching `normalizeTextOptions`. Two MINOR suggestions (edge parity and a test-coverage gap) inline.

Tara-ag:
> **SUPERSEDED.** This is an early review-cycle summary. The single authoritative summary is [comment 5653838867](https://github.com/juspay/neurolink/pull/1678#issuecomment-5653838867). Findings described here have since evolved; all are resolved on the current head `3b426e58`.

src/lib/providers/amazonBedrock/client.ts
```diff
@@ -297,10 +297,13 @@ export class AmazonBedrockProvider extends BaseProvider {
...
-        content: [{ text: options.prompt }],
+        content: [{ text: options.prompt ?? input?.text ?? "" }],
```
Tara-ag, line 306: **MINOR** — `??` makes the edge case diverge from the base contract.

`options.prompt ?? input?.text ?? ""` uses nullish coalescing, but the base contract resolves text via `options.prompt || input?.text || ...` (see `normalizeTextOptions` / `handleVideoGeneration` in baseProvider.ts). With `prompt` set to `""` and `input.text` populated, this line sends Bedrock an empty string, while the base path sends `input.text`.

Use `||` to get exactly the same precedence semantics as the base class:

```suggestion
        content: [{ text: options.prompt || input?.text || "" }],
```

test/continuous-test-suite-provider-wiring.ts
```diff
@@ -579,4 +579,112 @@ await test("catalog-provider enum surfaces are byte-identical to the pre-JSON-mi
...
+  // body at finalizeRequest and aborts, so nothing reaches AWS and no
+  // credential has to be real.
+  const { ProviderRegistry } = await import("../dist/index.js");
+  await ProviderRegistry.registerAllProviders();
+  const { ProviderFactory } =
+    await import("../dist/factories/providerFactory.js");
+
+  const PROMPT = "Reply with exactly OK.";
+  const shapes: Array<{ input: { text: string } } | { prompt: string }> = [
+    { input: { text: PROMPT } },
```
Tara-ag: **MINOR** — the two shapes are mutually exclusive, so `prompt`-over-`input.text` precedence is never asserted.

`{ input: { text: PROMPT } }` and `{ prompt: PROMPT }` each set only one field, so the precedence branch added in client.ts (`prompt ?? input.text`) is never exercised on the wire. A future inversion — making `input.text` win over `prompt` — would still pass CI.

Add a third shape that sets both, and assert the user text equals `PROMPT` (i.e. `prompt` wins):

```suggestion
  const shapes: Array<{
    input?: { text: string };
    prompt?: string;
  }> = [
    { input: { text: PROMPT } },
    { prompt: PROMPT },
    // Both shapes set — the base contract gives `prompt` precedence over
    // `input.text`; this asserts the native path does the same.
    { prompt: PROMPT, input: { text: "should-not-be-sent" } },
  ];
```

Tara-ag (commented):
Two MINOR inline findings; overall fix correct and approved.

Tara-ag:
> **SUPERSEDED.** Duplicate of the first-cycle summary. The single authoritative summary is [comment 5653838867](https://github.com/juspay/neurolink/pull/1678#issuecomment-5653838867). All findings are resolved on the current head `3b426e58`.

Tara-ag:
> **SUPERSEDED.** Recurring-review summary from an earlier cycle. The single authoritative summary is [comment 5653838867](https://github.com/juspay/neurolink/pull/1678#issuecomment-5653838867). Findings described here have since been resolved on the current head `3b426e58`.

test/continuous-test-suite-provider-wiring.ts
Tara-ag: **Repairing this finding against current HEAD.** Since this thread was posted, the test was refactored from a `shapes` array into discrete provider-call blocks (case 2 = `provider.generate({ input: { text: PROMPT } })`, case 3 = `provider.generate({ prompt: PROMPT })`), so the original `shapes`-array suggestion no longer applies. The finding itself still stands: the two blocks are mutually exclusive, so the `prompt`-over-`input.text` precedence the fix ships is never asserted.

Adapted suggestion for the current structure — add a third provider block that passes both and asserts `prompt` wins:

```suggestion
    // 3b — both shapes provided; the base contract gives `prompt` precedence.
    {
      const { ProviderRegistry } = await import("../dist/index.js");
      await ProviderRegistry.registerAllProviders();
      const { ProviderFactory } = await import("../dist/factories/providerFactory.js");
      const provider = await ProviderFactory.createProvider(
        "bedrock",
        MODEL,
        undefined,
        undefined,
        {
          bedrock: {
            accessKeyId: PLACEHOLDER_AWS_ENV.AWS_ACCESS_KEY_ID,
            secretAccessKey: PLACEHOLDER_AWS_ENV.AWS_SECRET_ACCESS_KEY,
            region: PLACEHOLDER_AWS_ENV.AWS_REGION,
          },
        },
      );
      const before = local.requests.length;
      await provider.generate({
        prompt: PROMPT,
        input: { text: "different-text-that-must-lose" },
      });
      const sent = local.requests.slice(before);
      assert(
        sent.length === 1,
        "both-shapes provider call sent an unexpected number of requests",
      );
      assert(
        userTextOnWire(sent[0].body) === PROMPT,
        "both-shapes precedence did not send prompt's text on the wire",
      );
    }
```

This locks the precedence branch; a future inversion to `input.text`-first would fail rather than pass silently. Still MINOR — the primary defect (both single shapes reaching the wire as the caller's text) is already well covered.

murdore:
Reworked for four-surface coverage, and it corrected the claim this PR was making.

**The blast radius was overstated.** I described this as affecting any caller using `input.text`. It does not affect `NeuroLink.generate()` or the CLI at all — the facade normalizes `input.text` into `prompt` before the provider is reached, so it always arrived populated. (`generate({ prompt })` is itself rejected by the facade with "Input text is required"; `prompt` survives only as a provider-level alias.) The reachable caller is one using `AIProviderFactory.createProvider()` directly with the documented `input.text` shape.

Still a bug — a supported entry point was not honouring its advertised contract — but a provider-contract bug, not a facade one. The description and commit message now say so.

**What caught it was the ritual, not the reasoning.** My first four-surface test passed with the fix reverted. A test that cannot fail proves nothing, and the only reason I noticed is that breaking it on purpose is a required step. The suite now includes the provider-level `input.text` case that actually reproduces the defect, and RED fails on it specifically.

## How it is tested now

`test/helpers/bedrockLocalEndpoint.ts` stands up a local Bedrock Runtime endpoint and points the SDK at it via `AWS_ENDPOINT_URL_BEDROCK_RUNTIME`. The real client, real SigV4 signing and real serialization all run; the assertions read the bytes that actually arrive. **No request reaches AWS and no credentials are required** — SigV4 signs with placeholders and nothing validates a signature.

| Case | Asserts |
|---|---|
| SDK generate (`input.text`) | `/converse`, prompt on the wire, reply returned |
| **provider-level `input.text`** | **the regression guard — fails when the fix is reverted** |
| provider-level `prompt` alias | prompt on the wire |
| SDK stream, drained | `/converse-stream`, prompt on the wire, reply drained |
| CLI generate | exit 0, `/converse`, prompt on the wire, reply printed |
| CLI stream | exit 0, `/converse-stream`, prompt on the wire, reply printed |

Three things the harness has to get right, each of which cost a debugging cycle and none of which are obvious:

- The SDK speaks **HTTP/2** to this service. An HTTP/1.1 listener fails ALPN and surfaces as a misleading "Protocol error".
- The listener is **cleartext h2c**, so no certificate or private key is committed — a TLS fixture would trip this repo's secret scanning for no benefit.
- The CLI cases use async `spawn`, never `spawnSync`. The test process *is* the endpoint, so a synchronous child blocks the event loop and the child's own request can never be served; it just times out looking like a hang.

Also checked while here: the **stream** path reads `options.input.text`, the mirror image of this defect — but `StreamOptions.input` is a required field, so stream cannot receive the `prompt`-only shape. The defect is generate-only and this fix's scope is right. That is a verified claim, not an assumption.

## Verification

- **green** — `✓ Bedrock carries the caller's text on the wire across all four public surfaces`, 19 passed
- **red** — with `dist` reverted to `content: [{ text: options.prompt }]`: `✗ provider-level input.text did not reach the wire (length 0)`, 18 passed · 1 failed, exit 1

`dist` restored byte-identical; tree clean. `check` 0 · `lint` 0 · `test:providers-mocked` 95/95 · `test:provider-structure` 3/3.

Still no live Bedrock round trip: the account returns 403 `UnrecognizedClientException`, consistent with a stale `AWS_SESSION_TOKEN`. Nothing here depends on it — the local endpoint drives the same client and serialization path.

Tara-ag:
> **SUPERSEDED.** Recurring-review summary from an earlier cycle. The single authoritative summary is [comment 5653838867](https://github.com/juspay/neurolink/pull/1678#issuecomment-5653838867). Findings described here have since been resolved on the current head `3b426e58`.

Tara-ag (approved):
Approve — the native Bedrock `generate` path now honours `input.text` with `prompt` precedence, matching the base contract. Two non-blocking MINOR inline suggestions stand (edge `||` parity and a test precedence gap).

src/lib/providers/amazonBedrock/client.ts
```diff
@@ -297,10 +297,19 @@ export class AmazonBedrockProvider extends BaseProvider {
...
+      // (see BaseProvider); reading only `prompt` here sent an empty user
+      // message whenever a caller used the `input.text` shape.
+      //
+      // `||`, not `??`, to match the canonical resolution in
+      // `Utilities.normalizeTextOptions`. This provider overrides `generate()`
+      // outright, so normalization never runs and the caller's raw options
+      // land here — an empty `prompt` must fall through to `input.text`
+      // rather than win by virtue of not being nullish.
       const userMessage: BedrockMessage = {
         role: "user",
```
murdore, line 311: Confirmed real, fixed, and proven red→green. You were right on both counts, and one of your two suggestions needed adjusting — details below.

**The precedence divergence is real, and reachable.** I checked the override rather than reasoning from the comment. `AmazonBedrockProvider.generate` is declared at `client.ts:233` and `BaseProvider.generate` at `baseProvider.ts:1605` — and `normalizeTextOptions` is called only inside the base, at `:1612`. So the raw, un-normalized options do reach this line exactly as you said, and `??` diverges from the canonical `||` in `Utilities.normalizeTextOptions:152`.

Changed to `||`, with a comment recording *why* it cannot follow the usual nullish-coalescing habit here.

**On the test, one correction.** The third case as suggested — both shapes supplied with different non-empty values — does **not** discriminate this fix. A non-empty `prompt` short-circuits under `??` and `||` alike, so that case passes on the release build too. It is a useful *guard* (it fails if someone later inverts the order to read `input.text` first), but it cannot catch the bug it was proposed for.

The case that discriminates is an **empty** `prompt` alongside a populated `input.text`, which is precisely the edge your comment identifies. I added both, labelled for what each one is:

- `3b` — both shapes, different text, `prompt` must win. Guard against a future inversion.
- `3c` — `{ prompt: "", input: { text: PROMPT } }`. The discriminator.

**Before (test added, fix not yet applied):**

```
✗ Bedrock carries the caller's text on the wire across all four public surfaces
  → empty-prompt fallthrough did not send the expected text on the wire (length 0)

Passed: 18   Failed: 1   Total: 19   RESULT: FAIL
```

`length 0` is the defect itself: the empty string travelled to the vendor as the user message. Note `3b` passed in this run — confirming it is a guard, not a discriminator.

**After:**

```
Passed: 19   Failed: 0   Total: 19   RESULT: PASS
```

`pnpm run lint` clean. Pushed as `c4d30fa7b`.

Tara-ag:
> **SUPERSEDED.** Recurring-review summary from an earlier cycle. The single authoritative summary is [comment 5653838867](https://github.com/juspay/neurolink/pull/1678#issuecomment-5653838867). Findings described here have since been resolved on the current head `3b426e58`.

Tara-ag (approved):
Approving on current head `c4d30fa7`. Both prior MINOR findings are now resolved:

- `??`→`||` precedence: fixed with an explanatory comment on why normalization never runs here.
- Test precedence gap: case `3b` locks `prompt`-wins over `input.text` (guard), and case `3c` (`{ prompt: "", input.text }`) is the discriminator that fails under the old code — proven red→green in the author's reply.

Fix is correct and well-tested; no remaining findings.

Tara-ag:
> **SUPERSEDED.** Recurring-review summary from an earlier cycle. The single authoritative summary is [comment 5653838867](https://github.com/juspay/neurolink/pull/1678#issuecomment-5653838867). Findings described here have since been resolved on the current head `3b426e58`.

test/helpers/bedrockLocalEndpoint.ts
```diff
@@ -0,0 +1,163 @@
...
+          usage: { inputTokens: 5, outputTokens: 1, totalTokens: 6 },
+        }),
+      );
+    });
+  });
+
+  await new Promise<void>((resolve) => server.listen(0, "127.0.0.1", resolve));
+  const address = server.address();
+  const port = typeof address === "object" && address ? address.port : 0;
+
```
Tara-ag, line 154: **MINOR** — `close()` may never resolve: `server.close()` waits for active HTTP/2 sessions.

`Http2Server.close()`'s callback only fires once every `Http2Session` it owns has closed (see the Node.js `http2server.close()` docs). The AWS v3 `BedrockRuntimeClient` used in the test drives requests through `NodeHttp2Handler`, which **pools** HTTP/2 sessions and reuses them across the six requests (SDK generate/stream + the provider calls), so those sessions stay open. When the test's `finally { await local.close(); }` runs, the callback can hang waiting for those pooled sessions — a silent teardown hang that would block the whole continuous suite.

This is latent (the suite passed in your red/green runs, so the sessions happened to wind down there), but it's a real flake risk that depends on client/GC timing. Make `close()` bounded by tracking and destroying live sessions:

```suggestion
  const sessions = new Set<Http2Session>();
  server.on("session", (s) => {
    sessions.add(s);
    s.on("close", () => sessions.delete(s));
  });

  return {
    endpoint: `http://127.0.0.1:${port}`,
    requests,
    close: () =>
      new Promise<void>((resolve) => {
        for (const s of sessions) s.destroy();
        server.close(() => resolve());
      }),
  };
```

Requires `type Http2Session` in the `node:http2` import. The in-process SDK client won't reconnect after a destroyed session; the later CLI cases spawn their own processes, so this does not affect the assertions.

Tara-ag (approved):
Both prior MINOR findings are resolved by `1e2cca7` (author reply `discussion_r3995006147`): `??`→`||` for base-contract parity, with cases `3b`/`3c` closing the test-precedence gap (red→green proven). The fix is correct and well-tested. One new non-blocking MINOR on the new test helper's `close()` teardown is inline.

Tara-ag:
> **SUPERSEDED.** Recurring-review summary from an earlier cycle. The single authoritative summary is [comment 5653838867](https://github.com/juspay/neurolink/pull/1678#issuecomment-5653838867). Findings described here have since been resolved on the current head `3b426e58`.

test/helpers/bedrockLocalEndpoint.ts
Tara-ag, line 154: **MINOR** — `close()` can hang the whole continuous suite on teardown.

`BedrockRuntimeClient` uses the SDK's `NodeHttp2Handler`, which **pools** HTTP/2 sessions against the same origin. A session can stay open after the request completes, and `Http2Server.close()` waits — indefinitely — for every active session before its callback fires. So `await local.close()` (in the suite's `finally` block) can remain pending until the harness times out, making the run look like a hang.

Track sessions on `'session'` and destroy them in `close()` so it resolves reliably:

```suggestion
    close: () => {
      server.close();
      for (const s of sessions) s.destroy();
      return new Promise<void>((resolve) =>
        server.once("close", resolve),
      );
    },
```

with `const sessions = new Set<Http2Session>();` declared next to `server`, and a `server.on("session", (s) => { sessions.add(s); s.on("close", () => sessions.delete(s)); });` wiring. (The `close` callback fires once all sessions are destroyed, so resolving on `once("close")` is safe.)

MINOR — correctness of the fix and both shapes is unaffected; this only guards teardown reliability. The finding applies to the newly-added helper `startLocalBedrock`'s `close`.

Tara-ag (commented):
Extra non-blocking MINOR on the newly-added local endpoint helper: `close()` can hang on pooled HTTP/2 sessions. Fix outcome and both input-shape finddds are unaffected; overall verdict remains APPROVE.

test/helpers/bedrockLocalEndpoint.ts
murdore: You are right, and my first pass at this was wrong. Fixed in `82e5c936f4052a99a7a33b62e79d41201235e7fc`.

**Why it looked fine at first.** I probed the helper against the traffic the suite actually drives — four consecutive runs, six requests each — and `close()` resolved in 0–1 ms every time. That is a real measurement and it is also useless here, because the suite's own path never holds a stream open: every request completes before the `finally` runs, and Node closes an *idle* h2 session on `close()` regardless. The happy path is not this hazard's precondition, so a probe that only drives the happy path cannot see it. I treated "did not reproduce" as "does not exist".

**The discriminating control.** Same helper, one request stream opened and deliberately never ended, so it stays in flight on the server. Precondition asserted first (the stream must really be open) so a fast "closed" could not be the no-connection case in disguise:

| helper | in-flight stream | `close()` |
|---|---|---|
| as committed | 1 open | **did not resolve** — hit the 8000 ms bound |
| with this fix | 1 open | **resolved in 2 ms** |

So the mechanism is exactly what you described.

**The fix.** The endpoint now tracks the sessions it has seen and destroys them before closing the server, which makes shutdown bounded by the helper rather than by the client's connection reuse:

```ts
const sessions = new Set<ServerHttp2Session>();
server.on("session", (session) => {
  sessions.add(session);
  session.on("close", () => sessions.delete(session));
});
```

`Provider Wiring` still passes 19/19 with the change in place, and the client now sees a clean shutdown instead of the `ECONNRESET` the previous teardown produced.

Note for anyone re-reviewing: #1697 is stacked on this branch and has been re-rebased onto the new head.

Tara-ag (approved):
**APPROVE on current head `82e5c936`.**

All prior MINOR findings are resolved. The last remaining one — the test-helper `close()` hang (`yama:bedrock-local-endpoint-close-hang`) — is fixed here: `startLocalBedrock` now tracks sessions on `'session'` and destroys them before `server.close()`, and the author's reply gives discriminating red/green proof (8 s bound hit without the fix, 2 ms with it). `Provider Wiring` still passes 19/19.

The fix itself (native Bedrock `generate` reading `options.prompt || input?.text || ""`) matches the base contract and is well covered by cases 1–6 including the precedence guard (`3b`) and the `||`-vs-`??` discriminator (`3c`). No remaining findings.

Tara-ag:
## Verdict: **APPROVE** — no open findings

On the current head `3b426e58` the fix correctly closes the silent-empty-string defect: the native Bedrock `generate` path reads both the `prompt` and `input.text` shapes as `options.prompt || input?.text || ""`, so a direct `AIProviderFactory.createProvider()` caller passing the documented `input.text` shape no longer sends Bedrock a blank user turn. All three MINOR findings raised across the review lifecycle are resolved; there are no open findings. The most recent approving review submits against the current head.

### Findings — all RESOLVED (no open items)

|----------|-----------|---------|--------|
| MINOR | `src/lib/providers/amazonBedrock/client.ts:311` | `??` diverged from the base contract's `||` on `{ prompt: "", input: { text } }` (base sends `input.text`). | **resolved** — now `options.prompt || input?.text || ""` with an explanatory comment; author reply `discussion_r3995006147`. |
| MINOR | `test/continuous-test-suite-provider-wiring.ts` (3b/3c) | Provider-level shapes were mutually exclusive, so `prompt`-wins precedence was never asserted. | **resolved** — case `3b` (both shapes, `prompt` wins) guards precedence; `3c` (`{ prompt: "", input: { text } }`) is the discriminator, red→green proven. |
| MINOR | `test/helpers/bedrockLocalEndpoint.ts` `close()` | `Http2Server.close()` waits for pooled `Http2Session`s from the SDK's `NodeHttp2Handler`, so `finally { await local.close(); }` could hang the suite on teardown. | **resolved** — sessions tracked on `'session'` and destroyed before `server.close()`; author reply `discussion_r3999856899` gives discriminating proof (8 s bound without, 2 ms with). |

### Checked and found clean
- **Precedence direction** — `options.prompt || input?.text || ""` matches the base contract (`Utilities.normalizeTextOptions` / `BaseProvider.handleVideoGeneration`, both `||`, `prompt` first); `3c` locks the empty-prompt edge.
- **Regression value** — local h2c HTTP/2 Bedrock endpoint exercises the real client, SigV4 signing and serialization, asserting the actual wire bytes across SDK/stream/CLI and both provider shapes; reverting fails case 2 (`provider-level input.text ... length 0`). Nothing reaches AWS.
- **Backward compatibility (rule 5)** — `generate`'s signature/return/async shape unchanged; no unmodified caller breaks. Production delta is a single-line input-shape resolution.
- **Conventions (rule 15)** — single conventional-commit `fix(bedrock):`; the new suite imports only `dist/` / `.js` / `node:` built-ins (one module graph).
- **Secrets** — `AKIALOCALENDPOINTONLY` / `local-endpoint-secret-not-real` are documented placeholders; the h2c listener ships no TLS key pair.
- **Review state** — verdict APPROVE and the latest submitted review is `APPROVED` against the current head `3b426e58`. Matches.
- **Comment hygiene** — one summary (this one); all finding threads resolved. The duplicate finding threads and older summary comments are clearly marked superseded. This is the single authoritative summary.

PR #1680, opened 2026-09-11
fix(sdk): forward thinkingConfig through the generate option allowlist

murdore:
## The defect

`generate()` accepts a `thinkingConfig` and the type documents it as the recommended way to switch extended thinking on. It never reached a provider.

`buildGenerateTextOptions` (`src/lib/neurolink.ts:5567`) converts the caller's public `GenerateOptions` into the internal `TextGenerationOptions` through an **explicit field-by-field allowlist**, and `thinkingConfig` was not in that list. Every provider gates thinking on `options.thinkingConfig`, so each one saw `undefined`. The request went out with no thinking block and nothing was raised to say the option had been discarded.

That function already carries a comment recording how `disableInternalFallback` was lost exactly this way. This is the second time the allowlist has silently swallowed a documented option.

## Proof — at the wire, red then green

The difference is only observable on the request, so that is where it is asserted.

| | release `3fdf80f1d` | this branch |
| --- | --- | --- |
| `thinkingConfig` reaches the request | ✗ **FAIL** — *no outbound request carried a thinking field* | ✓ PASS |
| omitting it sends no thinking field | ✓ | ✓ |
| CLI succeeds with thinking enabled | ✓ | ✓ |
| CLI unaffected with thinking off | ✓ | ✓ |
| | 1 passed, **1 failed** | **4 passed** |

Outbound request, captured on the built package:

```
before: keys=model,messages,max_tokens,system,tools          thinking=null
after:  keys=model,messages,max_tokens,system,tools,thinking thinking={"type":"enabled","budget_tokens":5000}
```

Only case 1 discriminates. The other three pass on **both** trees by design — they are regression guards, not proof.

The baseline is valid for this comparison: `10fa282f2` does not touch `neurolink.ts`, verified rather than assumed.

## A regression this fix caused, and fixes

Forwarding `thinkingConfig` is what first makes a `thinking` field reach the wire, and that immediately collided with a rule nothing had hit before: Anthropic rejects any temperature but 1 while thinking is enabled, and the CLI always sends a default temperature. The first version of this change turned a working `--thinking` invocation into a 400.

It was caught by running the **release build first** and finding it exited 0 — so the regression belonged to the fix, not the repository. Sampling knobs are now dropped whenever `thinking` is set, on both the generate and stream builders, and two CLI cases pin it.

## Why the suite does not assert on `result.reasoning`

The obvious assertion would fail forever. Whenever a proxy base URL is set — always true on the OAuth subscription path — the client requests the beta `redact-thinking-2026-02-12` (`anthropic/client.ts:878`), and Anthropic answers with thinking blocks whose text is redacted and whose signature is intact. Captured directly: `blockTypes=[thinking,text]`, thinking block `chars=0`, `hasSignature=true`. An empty `reasoning` is **correct** on that path.

The suite wraps `fetch` only to record what left the process. The call still goes through the real public `generate()` to the real provider; nothing is stubbed.

## Scope — deliberately narrow

`thinking`, `thinkingBudget` and `thinkingLevel` are documented in the type comments and in `CLAUDE.md`, but **none is declared on the public `GenerateOptions`**, so no caller can pass them through `generate()` at all. CLAUDE.md's own `thinkingLevel: "high"` example does not typecheck against the public API. Declaring them is a public-type decision, not a dropped-field fix, and is left out.

A caller's raw `providerOptions` is dropped by the same allowlist and is left alone for the same reason.

## Gates

`pnpm run check` 0 errors / 4844 files · eslint clean · prettier clean · pre-commit hook passed **without bypass** (it caught a real `no-useless-assignment` in the new test, which was fixed rather than skipped) · new suite registered as `test:thinking-config`.

**Correction (2026-09-11) — what that registration does and does not buy.** The script exists, but the suite appears in **no workflow**, so CI never runs it. The same is true of `test:matrix:cli`, `test:json-e2e` and the suite added in #1682. This repo's own `ci.yml` puts it exactly: *"a suite wired into nothing is documentation, not a test."* The red/green evidence above is real and was produced by running the suite against both trees by hand — but it will not catch a future regression until the suite is wired in. Live suites have one CI home, `live-matrix.yml`, which #1681 already modifies; the intent is one follow-up wiring the orphaned live suites in together once that lands, rather than scattering secret blocks across three PRs.

## Summary by CodeRabbit

- **New Features**
  - Extended-thinking settings now reach Anthropic requests when configured through the SDK.
  - Thinking budgets are honored for standard and streaming responses.
  - Requests without thinking enabled continue to omit thinking configuration.

- **Bug Fixes**
  - Incompatible sampling parameters are excluded when extended thinking is enabled, avoiding request errors.

- **Tests**
  - Added coverage for SDK and CLI thinking configuration, including enabled, standard, and streaming request flows.

### DIFF
--- a/src/lib/neurolink.ts
+++ b/src/lib/neurolink.ts
@@ -5615,6 +5615,22 @@ Current user's request: ${currentInput}`;
       middleware: options.middleware,
       conversationMessages: options.conversationMessages,
       credentials: options.credentials,
+      // Extended thinking. Every provider gates it on `thinkingConfig`, so
+      // omitting the field here meant it never reached one: the request went
+      // out with no thinking block and the result carried no reasoning, with
+      // nothing raised to say the option had been discarded. This is the same
+      // failure the note above records for `disableInternalFallback` — an
+      // allowlist that silently swallows a documented option.
+      //
+      // `thinkingConfig` is the only one of the documented thinking options
+      // that `GenerateOptions` actually declares. `thinking`, `thinkingBudget`
+      // and `thinkingLevel` are folded into a `thinkingConfig` only by the CLI,
+      // in `src/lib/utils/thinkingConfig.ts`; nothing on the SDK path does that
+      // merge. They exist solely on the internal `TextGenerationOptions`, so no
+      // caller can pass them through
+      // generate() today. Declaring them is a public-type decision and is
+      // deliberately left out of this fix.
+      thinkingConfig: options.thinkingConfig,
       // Lifecycle callbacks must reach the provider so non-AI-SDK paths
       // (Vertex's native @google/genai, native Bedrock, Ollama, etc.) can
       // invoke them directly. Pipeline A also still receives them via the
--- a/src/lib/providers/anthropic/client.ts
+++ b/src/lib/providers/anthropic/client.ts
@@ -1463,15 +1463,32 @@ export class AnthropicProvider extends BaseProvider {
           },
           "anthropic.doGenerate",
         );
+        // Dropping a caller's explicit sampling parameters is exactly the
+        // kind of silent discard this change fixes elsewhere, so say so.
+        if (
+          thinking &&
+          (samplingParams.temperature !== undefined ||
+            samplingParams.topP !== undefined)
+        ) {
+          logger.debug(
+            "[anthropic] extended thinking is enabled, so temperature/top_p are omitted — Anthropic rejects any temperature but 1 while thinking is set",
+          );
+        }
         const params: Anthropic.Messages.MessageCreateParamsNonStreaming = {
           model: modelId,
           messages: cachedMessages,
           max_tokens: resolveClaudeMaxTokens(modelId, options.maxOutputTokens),
           ...(system ? { system } : {}),
-          ...(samplingParams.temperature !== undefined
+          // Extended thinking fixes sampling: Anthropic rejects any
+          // temperature but 1 while `thinking` is set, and does not honour
+          // top_p there. The CLI always sends a default temperature, so
+          // forwarding it alongside thinking turns a call that used to work
+          // into a 400. Drop the sampling knobs for exactly those turns and
+          // let Anthropic's thinking defaults stand.
+          ...(!thinking && samplingParams.temperature !== undefined
             ? { temperature: samplingParams.temperature }
             : {}),
-          ...(samplingParams.topP !== undefined
+          ...(!thinking && samplingParams.topP !== undefined
             ? { top_p: samplingParams.topP }
             : {}),
           ...(options.stopSequences && options.stopSequences.length > 0
@@ -2353,6 +2370,11 @@ export class AnthropicProvider extends BaseProvider {
             : {},
           "anthropic.executeStream",
         );
+        if (thinking && streamSamplingParams.temperature !== undefined) {
+          logger.debug(
+            "[anthropic] extended thinking is enabled, so temperature is omitted on the stream path — Anthropic rejects any temperature but 1 while thinking is set",
+          );
+        }
         return {
           model: modelId,
           messages: cachedConversation,
@@ -2362,7 +2384,9 @@ export class AnthropicProvider extends BaseProvider {
           // the adapter immediately overwrites — and forced this whole params
           // object into the streaming variant for a field it does not own.
           ...(payload.system ? { system: payload.system } : {}),
-          ...(streamSamplingParams.temperature !== undefined
+          // Same constraint on the streaming path: a temperature alongside
+          // `thinking` is rejected outright.
+          ...(!thinking && streamSamplingParams.temperature !== undefined
             ? { temperature: streamSamplingParams.temperature }
             : {}),
           ...(cachedTools && cachedTools.length > 0
--- a/docs/api/classes/NeuroLink.md
+++ b/docs/api/classes/NeuroLink.md
@@ -469,7 +469,7 @@ Gracefully shutdown NeuroLink and all MCP connections
 
 > **generateText**(`options`): `Promise`\<[`TextGenerationResult`](../type-aliases/TextGenerationResult.md)\>
 
-Defined in: [neurolink.ts:6509](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L6509)
+Defined in: [neurolink.ts:6525](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L6525)
 
 BACKWARD COMPATIBILITY: Legacy generateText method
 Internally calls generate() and converts result format
@@ -490,7 +490,7 @@ Internally calls generate() and converts result format
 
 > **streamText**(`prompt`, `options?`): `Promise`\<`AsyncIterable`\<`string`, `any`, `any`\>\>
 
-Defined in: [neurolink.ts:9006](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L9006)
+Defined in: [neurolink.ts:9022](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L9022)
 
 BACKWARD COMPATIBILITY: Legacy streamText method
 Internally calls stream() and converts result format
@@ -515,7 +515,7 @@ Internally calls stream() and converts result format
 
 > **stream**(`options`): `Promise`\<[`StreamResult`](../type-aliases/StreamResult.md)\>
 
-Defined in: [neurolink.ts:9089](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L9089)
+Defined in: [neurolink.ts:9105](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L9105)
 
 Stream AI-generated content in real-time using the best available provider.
 This method provides real-time streaming of AI responses with full MCP tool integration.
@@ -588,7 +588,7 @@ When conversation memory operations fail (if enabled)
 
 > **setToolRoutingServers**(`servers`): `void`
 
-Defined in: [neurolink.ts:9981](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L9981)
+Defined in: [neurolink.ts:9997](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L9997)
 
 Supplies (or replaces) the pre-call tool routing server catalog.
 
@@ -613,7 +613,7 @@ alone does not activate it.
 
 > **getKnowledgeStatus**(): [`KnowledgeEngineStatus`](../type-aliases/KnowledgeEngineStatus.md) \| `null`
 
-Defined in: [neurolink.ts:9999](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L9999)
+Defined in: [neurolink.ts:10015](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L10015)
 
 Knowledge-grounding engine health (null when it was not configured).
 
@@ -627,7 +627,7 @@ Knowledge-grounding engine health (null when it was not configured).
 
 > **getEventEmitter**(): `TypedEventEmitter`\<[`NeuroLinkEvents`](../type-aliases/NeuroLinkEvents.md)\>
 
-Defined in: [neurolink.ts:12414](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12414)
+Defined in: [neurolink.ts:12430](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12430)
 
 Get the EventEmitter instance to listen to NeuroLink events for real-time monitoring and debugging.
 This method provides access to the internal event system that emits events during AI generation,
@@ -833,7 +833,7 @@ This method does not throw errors as it returns the internal EventEmitter
 
 > **hasPendingHITLConfirmation**(`confirmationId`): `boolean`
 
-Defined in: [neurolink.ts:12454](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12454)
+Defined in: [neurolink.ts:12470](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12470)
 
 Whether a HITL confirmation is still awaiting a response on THIS instance.
 
@@ -888,7 +888,7 @@ neurolink.getEventEmitter().emit("hitl:confirmation-response", { ... });
 
 > **getToolDedupConfig**(): [`ToolDedupConfig`](../type-aliases/ToolDedupConfig.md) \| `undefined`
 
-Defined in: [neurolink.ts:12470](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12470)
+Defined in: [neurolink.ts:12486](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12486)
 
 Returns the instance-level tool-dedup configuration, or `undefined` when
 toolDedup was not provided at construction time.
@@ -911,7 +911,7 @@ parameter through the full call stack.
 
 > **getToolsConfig**(): [`ToolConfig`](../type-aliases/ToolConfig.md) \| `undefined`
 
-Defined in: [neurolink.ts:12481](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12481)
+Defined in: [neurolink.ts:12497](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12497)
 
 Returns the instance-level `tools` config (master switch, include/exclude
 lists, discovery mode), or `undefined` when not provided at construction.
@@ -929,7 +929,7 @@ instance policy with per-call options on every generate/stream call.
 
 > **getDiscoveryPins**(`sessionKey`): `ReadonlySet`\<`string`\>
 
-Defined in: [neurolink.ts:12492](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12492)
+Defined in: [neurolink.ts:12508](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12508)
 
 Tools discovered via `search_tools` for a session (`tools.discovery`
 mode). Pinned tools are sent in full on every subsequent call of that
@@ -953,7 +953,7 @@ are never the ones evicted at the session cap.
 
 > **pinDiscoveredTools**(`sessionKey`, `toolNames`): `void`
 
-Defined in: [neurolink.ts:12509](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12509)
+Defined in: [neurolink.ts:12525](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12525)
 
 Pin discovered tools to a session (called by the `search_tools`
 meta-tool on hydration). Append-only within a session; the map is
@@ -979,7 +979,7 @@ bounded by evicting the least-recently-used session past 1000 sessions.
 
 > **checkCredentials**(`input`): `Promise`\<\{ `provider`: `string`; `status`: `"network"` \| `"expired"` \| `"unknown"` \| `"ok"` \| `"missing"` \| `"denied"`; `detail`: `string`; \}\>
 
-Defined in: [neurolink.ts:12554](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12554)
+Defined in: [neurolink.ts:12570](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12570)
 
 Curator P1-1: synchronous credential health check for a single provider.
 
@@ -1031,7 +1031,7 @@ if (health.status !== "ok") {
 
 > **emitToolStart**(`toolName`, `input`, `startTime?`): `string`
 
-Defined in: [neurolink.ts:12633](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12633)
+Defined in: [neurolink.ts:12649](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12649)
 
 Emit tool start event with execution tracking
 
@@ -1067,7 +1067,7 @@ executionId for tracking this specific execution
 
 > **emitToolEnd**(`toolName`, `result?`, `error?`, `startTime?`, `endTime?`, `executionId?`): `void`
 
-Defined in: [neurolink.ts:12684](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12684)
+Defined in: [neurolink.ts:12700](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12700)
 
 Emit tool end event with execution summary
 
@@ -1119,7 +1119,7 @@ Optional execution ID for tracking
 
 > **getCurrentToolExecutions**(): [`ToolExecutionContext`](../type-aliases/ToolExecutionContext.md)[]
 
-Defined in: [neurolink.ts:12765](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12765)
+Defined in: [neurolink.ts:12781](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12781)
 
 Get current tool execution contexts for stream metadata
 
@@ -1133,7 +1133,7 @@ Get current tool execution contexts for stream metadata
 
 > **getToolExecutionHistory**(): [`ToolExecutionSummary`](../type-aliases/ToolExecutionSummary.md)[]
 
-Defined in: [neurolink.ts:12772](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12772)
+Defined in: [neurolink.ts:12788](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12788)
 
 Get tool execution history
 
@@ -1147,7 +1147,7 @@ Get tool execution history
 
 > **clearCurrentStreamExecutions**(): `void`
 
-Defined in: [neurolink.ts:12779](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12779)
+Defined in: [neurolink.ts:12795](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12795)
 
 Clear current stream tool executions (called at stream start)
 
@@ -1161,7 +1161,7 @@ Clear current stream tool executions (called at stream start)
 
 > **registerTool**(`name`, `tool`, `options?`): `void`
 
-Defined in: [neurolink.ts:12795](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12795)
+Defined in: [neurolink.ts:12811](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12811)
 
 Register a custom tool that will be available to all AI providers
 
@@ -1207,7 +1207,7 @@ Tool in MCPExecutableTool format (unified MCP protocol type)
 
 > **setToolContext**(`context`): `void`
 
-Defined in: [neurolink.ts:12946](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12946)
+Defined in: [neurolink.ts:12962](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12962)
 
 Set the context that will be passed to tools during execution
 This context will be merged with any runtime context passed by the AI model
@@ -1230,7 +1230,7 @@ Context object containing session info, tokens, shop data, etc.
 
 > **getToolContext**(): `Record`\<`string`, `unknown`\> \| `undefined`
 
-Defined in: [neurolink.ts:12961](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12961)
+Defined in: [neurolink.ts:12977](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12977)
 
 Get the current tool execution context
 
@@ -1246,7 +1246,7 @@ Current context or undefined if not set
 
 > **clearToolContext**(): `void`
 
-Defined in: [neurolink.ts:12970](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12970)
+Defined in: [neurolink.ts:12986](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12986)
 
 Clear the tool execution context
 
@@ -1260,7 +1260,7 @@ Clear the tool execution context
 
 > **registerTools**(`tools`): `void`
 
-Defined in: [neurolink.ts:12982](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12982)
+Defined in: [neurolink.ts:12998](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L12998)
 
 Register multiple tools at once - Supports both object and array formats
 
@@ -1285,7 +1285,7 @@ Array format (Lighthouse compatible): [{ name: string, tool: MCPExecutableTool }
 
 > **unregisterTool**(`name`): `boolean`
 
-Defined in: [neurolink.ts:13005](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L13005)
+Defined in: [neurolink.ts:13021](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L13021)
 
 Unregister a custom tool
 
@@ -1309,7 +1309,7 @@ true if the tool was removed, false if it didn't exist
 
 > **useToolMiddleware**(`middleware`): `this`
 
-Defined in: [neurolink.ts:13024](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L13024)
+Defined in: [neurolink.ts:13040](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L13040)
 
 Register a global tool middleware that runs on every tool execution.
 Middleware receives the tool, params, context, and a next() function.
@@ -1334,7 +1334,7 @@ this (for chaining)
 
 > **getToolMiddlewares**(): [`ToolMiddleware`](../type-aliases/ToolMiddleware.md)[]
 
-Defined in: [neurolink.ts:13037](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L13037)
+Defined in: [neurolink.ts:13053](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L13053)
 
 Get all registered tool middlewares
 
@@ -1348,7 +1348,7 @@ Get all registered tool middlewares
 
 > **flushToolBatch**(): `Promise`\<`void`\>
 
-Defined in: [neurolink.ts:13044](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L13044)
+Defined in: [neurolink.ts:13060](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L13060)
 
 Flush any pending batched tool calls immediately
 
@@ -1362,7 +1362,7 @@ Flush any pending batched tool calls immediately
 
 > **getMCPEnhancementsConfig**(): [`MCPEnhancementsConfig`](../type-aliases/MCPEnhancementsConfig.md) \| `undefined`
 
-Defined in: [neurolink.ts:13053](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L13053)
+Defined in: [neurolink.ts:13069](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L13069)
 
 Get the current MCP enhancements configuration
 
@@ -1376,7 +1376,7 @@ Get the current MCP enhancements configuration
 
 > **updateAgenticLoopReport**(`sessionId`, `report`, `userId?`): `Promise`\<`void`\>
 
-Defined in: [neurolink.ts:13076](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L13076)
+Defined in: [neurolink.ts:13092](https://github.com/juspay/neurolink/blob/release/src/lib/neurolink.ts#L13092)
 
 Update agentic loop report metadata for a conversation session.
 Upserts a report entry by reportId — updates existing or adds new.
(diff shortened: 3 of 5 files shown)


src/lib/providers/anthropic/client.ts
```diff
@@ -1468,10 +1468,16 @@ export class AnthropicProvider extends BaseProvider {
...
           max_tokens: resolveClaudeMaxTokens(modelId, options.maxOutputTokens),
           ...(system ? { system } : {}),
-          ...(samplingParams.temperature !== undefined
+          // Extended thinking fixes sampling: Anthropic rejects any
+          // temperature but 1 while `thinking` is set, and does not honour
+          // top_p there. The CLI always sends a default temperature, so
+          // forwarding it alongside thinking turns a call that used to work
+          // into a 400. Drop the sampling knobs for exactly those turns and
+          // let Anthropic's thinking defaults stand.
+          ...(!thinking && samplingParams.temperature !== undefined
```
Tara-ag, line 1488: **MINOR — a thinker's explicit sampling knobs are now dropped with no trace.**

When a caller passes **both** `temperature`/`top_p` **and** `thinkingConfig` on the same Anthropic `generate()` call, the knobs are silently discarded. The change is necessary (Anthropic rejects `temperature != 1` with `thinking`), and the regression-guard `!thinking` is correct — but this is exactly the *silent-discarding* failure mode the PR itself is fixing in the `neurolink.ts` allowlist, now re-introduced one layer down. Before AGENTS could not even reach this path, so there was no contract to break; still, a caller who set `temperature: 0.5` and `thinkingConfig` will get a 200 that ignores what they asked for, with nothing surfaced.

Consider a one-line `logger.debug(...)` on the drop so the choice is observable rather than invisible:

```suggestion
          // Extended thinking fixes sampling: Anthropic rejects any
          // temperature but 1 while `thinking` is set, and does not honour
          // top_p there. The CLI always sends a default temperature, so
          // forwarding it alongside thinking turns a call that used to work
          // into a 400. Drop the sampling knobs for exactly those turns and
          // let Anthropic's thinking defaults stand.
          ...(thinking
            ? (logger.debug(
                "[Anthropic] Dropping temperature/top_p: not allowed with extended thinking",
              ),
              {})
            : { temperature: samplingParams.temperature }),
```
(optional — fix is correct as-is; this is purely an observability suggestion)

src/lib/providers/anthropic/client.ts
```diff
@@ -2362,7 +2368,9 @@ export class AnthropicProvider extends BaseProvider {
           // the adapter immediately overwrites — and forced this whole params
           // object into the streaming variant for a field it does not own.
           ...(payload.system ? { system: payload.system } : {}),
-          ...(streamSamplingParams.temperature !== undefined
+          // Same constraint on the streaming path: a temperature alongside
```
Tara-ag, line 2387: **MINOR — the streaming sampling-drop guard added here has no test coverage.**

This `!thinking` guard on `streamSamplingParams.temperature` is the second half of the regression fix, but every case in the new suite (`test/continuous-test-suite-sdk-thinking-config.ts`) drives the non-streaming `generate()` path — the two CLI cases go through the `generate` subcommand, and the two `observeRequests` cases call `new NeuroLink().generate(...)` directly. Nothing exercises `executeStream` / `runLoop.buildParams`.

Since this is exactly the code that would turn a working `--thinking` **stream** into a 400 if the `!thinking` guard regressed — the same bug the PR caught on the non-stream path — and the line-for-line parallel in the generate path did earn tests, this side of the fix would benefit from at least one streaming assertion (e.g., a `stream()` call under `observeRequests` that asserts the outbound request carries `thinking` and no `temperature`).

Tara-ag (commented):
Two MINOR findings; no blockers. The core fix is correct and well-tested on the non-streaming path.

Tara-ag:
## APPROVE — with two MINOR follow-ups

The core defect is real, correctly diagnosed, and correctly fixed: `buildGenerateTextOptions`' field-by-field allowlist silently dropped the documented `thinkingConfig`, so no provider ever saw it. Forwarding it (`neurolink.ts`) plus dropping the Anthropic sampling knobs when `thinking` is set (`client.ts`) is the right, minimal, well-documented change. The wire-level regression test and the two CLI regression guards are good; the reasoning about why asserting on `result.reasoning` would be wrong (redacted-thinking beta) is correct and well-explained.

### Findings

| MINOR | `src/lib/providers/anthropic/client.ts:1477` | Explicit `temperature`/`top_p` is dropped silently on thinking turns (no log) — the same silent-discard pattern the PR is fixing, one layer down. |
| MINOR | `src/lib/providers/anthropic/client.ts:2371` | The streaming sampling-drop guard has **no test** — all four new cases drive non-stream `generate()`; a stream `--thinking` 400 regression would go uncaught. |

Both are optional; neither blocks merge.

### Checked and clean

- **neurolink.ts** — `thinkingConfig: options.thinkingConfig` added to the allowlist; only caller is `runStandardGenerateRequest`. No backward-compat break (Rule 5): the option was previously dropped, so no working caller changes behaviour.
- **anthropic generate path** — both `temperature` and `top_p` guarded by `!thinking`; consistent. `thinking` derived from `options.thinkingConfig`.
- **anthropic stream path** — confirmed `streamSamplingParams` carries **only** `temperature` (no `top_p` exists there), so the single `!thinking` guard is complete, not asymmetric. `thinking` is in closure scope for `buildParams`.
- **Cross-provider** — `thinkingConfig` is genuinely read by Gemini-native, Nvidia NIM, proxy-translation, etc.; forwarding it is safe (previously unreachable via `generate`).
- **Test suite** — conforms to Rule 15 (one module graph, `dist` imports, `assertDistFresh`, live end-to-end, `skipUnlessProviderAvailable`); `fetch` wrapper restored in `finally`.
- **package.json** — `test:thinking-config` script registered; single-commit policy satisfied.

src/lib/providers/anthropic/client.ts
```diff
@@ -1468,10 +1468,16 @@ export class AnthropicProvider extends BaseProvider {
...
+          // top_p there. The CLI always sends a default temperature, so
+          // forwarding it alongside thinking turns a call that used to work
+          // into a 400. Drop the sampling knobs for exactly those turns and
+          // let Anthropic's thinking defaults stand.
+          ...(!thinking && samplingParams.temperature !== undefined
             ? { temperature: samplingParams.temperature }
             : {}),
-          ...(samplingParams.topP !== undefined
+          ...(!thinking && samplingParams.topP !== undefined
             ? { top_p: samplingParams.topP }
```
Tara-ag, line 1492: **MINOR** — the explicit `temperature`/`top_p` a caller set is now silently discarded on thinking turns, and the sampling-drop here is the same silent-discard pattern this PR is fixing in `neurolink.ts`.

The core fix is correct and necessary: Anthropic rejects any `temperature != 1` while `thinking` is set, and does not honour `top_p` there, so dropping them is right. The gap is observability — a caller who explicitly passed `temperature: 0.7` gets no signal that it was ignored. That is exactly the "nothing raised to say so" failure the PR's own allowlist comment describes.

Suggest logging the drop so the discard is visible on thinking turns:

```suggestion
          // Extended thinking fixes sampling: Anthropic rejects any
          // temperature but 1 while `thinking` is set, and does not honour
          // top_p there. The CLI always sends a default temperature, so
          // forwarding it alongside thinking turns a call that used to work
          // into a 400. Drop the sampling knobs for exactly those turns and
          // let Anthropic's thinking defaults stand.
          if (thinking && samplingParams.temperature !== undefined) {
            logger?.debug?.(
              `Dropping temperature=${samplingParams.temperature} and top_p=${
                samplingParams.topP ?? "unset"
              } for thinking turn`,
            );
          }
          ...(!thinking && samplingParams.temperature !== undefined
            ? { temperature: samplingParams.temperature }
            : {}),
          ...(!thinking && samplingParams.topP !== undefined
            ? { top_p: samplingParams.topP }
            : {}),
```

src/lib/providers/anthropic/client.ts
```diff
@@ -2362,7 +2368,9 @@ export class AnthropicProvider extends BaseProvider {
           // the adapter immediately overwrites — and forced this whole params
           // object into the streaming variant for a field it does not own.
           ...(payload.system ? { system: payload.system } : {}),
-          ...(streamSamplingParams.temperature !== undefined
+          // Same constraint on the streaming path: a temperature alongside
+          // `thinking` is rejected outright.
+          ...(!thinking && streamSamplingParams.temperature !== undefined
```
Tara-ag, line 2389: **MINOR** — the streaming-path sampling guard has **no coverage**; all four cases in the new suite drive the non-streaming `generate()` builder.

The fix correctly applies the same `!thinking` guard here (`streamSamplingParams.temperature`), but the suite only exercises `buildGenerateTextOptions` / the generate path. A future regression where a `--thinking --stream` call re-sends the CLI's default temperature alongside `thinking` (the exact 400 this PR fixes) would go uncaught.

Consider adding one live case that runs the stream path with thinking enabled and asserts a non-zero exit / no 400. Note that `streamSamplingParams` is derived from `temperature` only, so the single `!thinking` guard here is complete — the note is purely about coverage, not logic.

Tara-ag (approved):
Two MINOR findings; no blockers. The core fix is correct and well-tested on the non-streaming path.

- `client.ts:1476` (MINOR) — explicit `temperature`/`top_p` is dropped silently on thinking turns; suggest a log of the drop.
- `client.ts:2373` (MINOR) — the streaming-path sampling guard lacks test coverage.

Approving — the fix is minimal, correct, and backward compatible (Rule 5).

Tara-ag:
**Verdict: APPROVE** — the fix is correct, minimal, and does not break the public SDK API. Two optional MINOR notes below.

## Findings

| --- | --- | --- |
| MINOR | `src/lib/providers/anthropic/client.ts:1476` | Explicit `temperature`/`top_p` is dropped silently on thinking turns — no log; the same silent-discard pattern this PR fixes in `neurolink.ts` |
| MINOR | `src/lib/providers/anthropic/client.ts:2373` | The streaming-path sampling guard (`!thinking`) has no test coverage — a `--thinking --stream` 400 regression would go uncaught |

## What the change does

`buildGenerateTextOptions`'s field-by-field allowlist previously swallowed the documented `thinkingConfig` option, so it never reached any provider. This PR (correctly) forwards it, and drops Anthropic `temperature`/`top_p` on thinking turns (generate + stream builders) because Anthropic rejects `temperature != 1` while thinking is enabled.

## Checked and found clean

- **Backward compatibility (Rule 5)**: `thinkingConfig` was previously unreachable via `generate()` — no caller passes it through today, so no unmodified caller breaks. `buildGenerateTextOptions` has a single caller (`runStandardGenerateRequest`).
- **Sampling-drop completeness**: the streaming path derives `streamSamplingParams` from `temperature` only (no `top_p` branch), so the single `!thinking` guard there is complete — not asymmetric.
- **`thinking` in scope**: correctly derived in closure from `options.thinkingConfig`; not a `max_tokens`/`resolveClaudeMaxTokens` or Gemini/Vertex concern.
- **Cross-provider**: `thinkingConfig` is genuinely consumed by the Gemini path, Nvidia NIM, and proxy translation — forwarding it is safe (it was previously dead through `generate()`).
- **Registry / Gemini-tools / formatProviderError rule**s: untouched by this change.
- **New test suite** (`continuous-test-suite-sdk-thinking-config.ts`): end-to-end only, drives `dist/index.js`, declares determinism in header, registered as `test:thinking-config`. Uses the legacy tap-free `defineSuite` harness consistent with siblings.

No CRITICAL or MAJOR findings; the two MINOR notes are optional follow-ups, not blockers.

src/lib/neurolink.ts
```diff
@@ -5615,6 +5615,22 @@ Current user's request: ${currentInput}`;
...
+      // ever reached a provider: the request went out with no thinking block
+      // and the result carried no reasoning, with nothing raised to say so.
+      // This is the same failure the note above records for
+      // `disableInternalFallback` — an allowlist that silently swallows a
+      // documented option.
+      //
+      // `thinkingConfig` is the only one of the documented thinking options
+      // that `GenerateOptions` actually declares; `thinking`, `thinkingBudget`
+      // and `thinkingLevel` are described in the docs but exist solely on the
+      // internal `TextGenerationOptions`, so no caller can pass them through
```
Tara-ag: **MINOR — the added comment attributes the trio-fold to the wrong function.**

The core change is correct and I'm not disputing it. But the rationale comment records an inaccurate mechanism: it says *"the simplified trio is folded into it by `normalizeTextOptions`"*. I checked the implementation — `normalizeTextOptions` (`src/lib/core/modules/Utilities.ts`) is a plain `{ ...optionsOrPrompt, prompt, provider, model, maxTokens }` spread with **no fold of `thinking`/`thinkingBudget`/`thinkingLevel` into `thinkingConfig`**. The trio is folded to `thinkingConfig` only in the **CLI** util `src/lib/utils/thinkingConfig.ts` (it is the sole consumer of `options.thinkingLevel` / `options.thinkingBudget`), driven by the CLI arg shape, not by the SDK `generate()` route.

The conclusion is right (`thinkingConfig` is the only one of the trio reachable through `generate()`), so nothing about the fix's scope changes — but this block is explicitly written as a durable "why X was left out" record, and a future reader relying on its stated mechanism would be misled into believing the SDK would suddenly start folding the trio if those fields were ever added to `GenerateOptions`.

Suggest correcting the attribution in the comment:

````suggestion
      // Extended thinking, in both documented shapes. Every provider gates
      // thinking on `thinkingConfig`. Omitting it here meant neither shape
      // ever reached a provider: the request went out with no thinking block
      // and the result carried no reasoning, with nothing raised to say so.
      // This is the same failure the note above records for
      // `disableInternalFallback` — an allowlist that silently swallows a
      // documented option.
      //
      // `thinkingConfig` is the only thinking option that `GenerateOptions`
      // actually declares. `thinking`, `thinkingBudget` and `thinkingLevel`
      // are described in the docs and exist on the internal
      // `TextGenerationOptions`, but no caller can pass them to `generate()`
      // today; the simplified trio is folded into `thinkingConfig` only in the
      // CLI path (`src/lib/utils/thinkingConfig.ts`). Declaring them on the
      // public type is a public-surface decision and is deliberately left out
      // of this fix, exactly as a caller's raw `providerOptions` is.
      thinkingConfig: options.thinkingConfig,
````

Optional — no behaviour change; purely comment accuracy.

**Review by Tara-ag — COMMENTED** [review]
One MINOR comment-accuracy note added. The two previously-posted MINOR findings (silent sampling drop; untested stream guard) still stand — no author reply yet, and the code is unchanged since they were posted. The core fix is correct.

**Comment by Tara-ag** [comment]
**Verdict: APPROVE** — the fix is correct, minimal, and does not break the public SDK API. One new MINOR comment-accuracy note added this run; the two previously-posted MINOR findings still stand (no author reply, code unchanged).

## Findings

| Severity | Location | Status | Description |
| -------- | -------- | ------ | ----------- |
| MINOR | `src/lib/neurolink.ts:5619` | **new** | Added allowlist comment misattributes the trio-fold ("simplified trio is folded into it by `normalizeTextOptions`") — `normalizeTextOptions` is a plain spread; only the CLI util `src/lib/utils/thinkingConfig.ts` folds `thinking`/`thinkingBudget`/`thinkingLevel` into `thinkingConfig`. Scope conclusion is correct; the stated mechanism is wrong. |
| MINOR | `src/lib/providers/anthropic/client.ts:1476` | carried over (no reply) | Explicit `temperature`/`top_p` dropped silently on thinking turns (halted `--thinking` 400; observability gap on the drop). |
| MINOR | `src/lib/providers/anthropic/client.ts:2373` | carried over (no reply) | Streaming-path `!thinking` sampling guard has no test coverage — all four new cases drive non-stream `generate()`. |

## What the change does

`buildGenerateTextOptions`' field-by-field allowlist previously swallowed the documented `thinkingConfig`, so no provider ever saw it. This PR forwards it (`thinkingConfig: options.thinkingConfig`) and drops Anthropic `temperature`/`top_p` on thinking turns (generate + stream builders) because Anthropic rejects `temperature != 1` / `top_p` while thinking is enabled.

## Checked and found clean

- **Backward compat (Rule 5):** `thinkingConfig` was previously unreachable via `generate()` — no caller breaks. `buildGenerateTextOptions` has a single caller (`runStandardGenerateRequest`).
- **Sampling-drop correctness:** generate guards both `temperature` and `topP`; stream derives `streamSamplingParams` from `temperature` only, so its single `!thinking` guard is complete (no asymmetry).
- **Thinking derivation matches** on both generate and stream paths (from `options.thinkingConfig.enabled && budgetTokens`).
- **Cross-provider:** confirmed `GenerateOptions` (public) declares only `thinkingConfig`; `thinkingConfig` is genuinely consumed by Anthropic, Gemini/Vertex, Nvidia NIM, proxy translation — forwarding is safe. Scope decision (leaving `thinking`/`thinkingBudget`/`thinkingLevel` off the public type) is sound.
- **Type-level verification:** `thinkingConfig` shape on the public type = `{ enabled?, type?, budgetTokens?, thinkingLevel? }`; the three way to compare the two. Wire-level test correctly and deliberately does not assert `result.reasoning` (redacted-thinking beta returns empty `reasoning`; explains why the PR asserts on the request instead).
- **Test suite:** single module graph, `dist/index.js` import, `assertDistFresh`, live e2e `skipUnlessProviderAvailable`, `fetch` restored in `finally`, registered as `test:thinking-config` in `package.json`.

## Worth the author's attention (second source)

CodeRabbit flagged two **out-of-diff** MAJOR concerns on `client.ts` (already on the PR): manual thinking + forced/named `toolChoice` → HTTP 400, and assistant thinking blocks being dropped across generate tool turns (`messagesToAnthropic`). These are pre-existing, not introduced here, but this PR is what first makes `thinkingConfig` reach the wire via `generate()`, so they newly materialise. Not independently verified as blockers for this PR's scope; recommend the author confirm before merge.

The docs/api and package.json changes are line-number/docs churn and script registration respectively — no findings.

**Comment by Tara-ag** [comment]
**Verdict: APPROVE**

This is a recurring review. All five previously-posted MINOR findings have been addressed in the current head (`69142b1`) — three by explicit code change and one by a rewritten comment — and no new issues were introduced by those fixes. The core defect (silent `thinkingConfig` discard in the `buildGenerateTextOptions` allowlist) is correctly fixed with minimal, backward-safe surface.

## Findings

| Severity | Location | Status | Description |
| -------- | -------- | ------ | ----------- |
| MINOR | `src/lib/neurolink.ts:5630` | **resolved** | Comment now correctly attributes the `thinking`/`thinkingBudget`/`thinkingLevel`→`thinkingConfig` fold to the CLI util (`src/lib/utils/thinkingConfig.ts`), not `normalizeTextOptions`. Thread `PRRT_kwDOOzxF1c6hfsTe` resolved this run. |
| MINOR | `src/lib/providers/anthropic/client.ts:1463-1472`, `2373-2379` | **resolved** | Silent drop of explicit `temperature`/`top_p` on thinking turns is now surfaced via `logger.debug` on both the generate and stream builders. |
| MINOR | `src/lib/providers/anthropic/client.ts:2387` | **resolved** | Streaming-path sampling guard now has coverage: the new test case `"an explicit temperature is dropped on a thinking stream turn"` drives `NeuroLink().stream()`. |

No CRITICAL or MAJOR findings. No new issues found in the changes since the last review.

## What the change does

`buildGenerateTextOptions`' field-by-field allowlist previously swallowed `thinkingConfig`, so it never reached any provider. This PR forwards it (`thinkingConfig: options.thinkingConfig`) and drops Anthropic `temperature`/`top_p` on thinking turns (generate + stream builders) because Anthropic rejects `temperature != 1` / `top_p` while thinking is enabled — the exact 400 the fix's own CLI regression guard pins.

## Blast radius & backward compat (Rule 5)

- `buildGenerateTextOptions` has **exactly one caller**, `runStandardGenerateRequest` (confirmed via code graph).
- `thinkingConfig` was previously dropped on the `generate()` route, so no unmodified caller relied on it being forwarded — forwarding it is additive, not breaking. The larger "159 files affected" radius reported by the graph is the hub nature of `neurolink.ts` (imported everywhere), not a consequence of this change.
- Cross-provider: `thinkingConfig` is genuinely consumed by the Gemini/Vertex path, Nvidia NIM, and proxy translation, so forwarding is safe.

## Verified in the current head source

- Stream path (`executeStreamInCaptureScope` → `buildParams`): `thinking` derived from `options.thinkingConfig?.enabled && budgetTokens`, in closure for the `!thinking` guard; `streamSamplingParams` derives from `temperature` only, so the single guard is complete (no `top_p` asymmetry).
- `logger` is the module-level import (`../../utils/logger.js`), so the new `logger.debug` calls are non-null-safe in both scopes.
- New stream test is sound: it drives the real `stream()`, asserts a thinking request was observed, and asserts no request carried `temperature` on that path; `fetch` restored in `finally`.
- Test suite (new file): live end-to-end, single module graph (`dist/index.js` + helpers), `assertDistFresh`, `skipUnlessProviderAvailable`; conforms to Rule 15. `test:thinking-config` registered in `package.json`.
- `docs/api/classes/NeuroLink.md` changes are typedoc line-number churn.

## Worth the author's attention (second source), not blockers

CodeRabbit's out-of-diff MAJORs on `client.ts` (manual thinking + forced/named `toolChoice` → HTTP 400; assistant thinking blocks dropped across generate tool turns in `messagesToAnthropic`) are **pre-existing**, but this PR is what first makes `thinkingConfig` reach the wire via `generate()`, so they newly materialise. Outside this PR's stated scope; recommend confirming before merge.

Checked and clean: Rule 3 (Gemini tools/json-schema) unaffected; Rule 6 `formatProviderError` untouched; no registry/static-import change; no secrets introduced; single-commit policy satisfied.

**Review by Tara-ag — APPROVED** [review]
**APPROVE** — all prior MINOR findings are addressed in head `69142b1`; the fix is correct, minimal, and backward-compatible (Rule 5).

- `client.ts` generate + stream builders now log the sampling-drop instead of discarding silently; every thinking turn omits `temperature`/`top_p` (Anthropic rejects any temperature but 1 while thinking is set).
- The stream sampling guard now has a live test (`continuous-test-suite-sdk-thinking-config.ts` — "an explicit temperature is dropped on a thinking stream turn") driving the real `NeuroLink().stream()`.
- `neurolink.ts` allowlist comment correctly attributes the `thinking`/`thinkingBudget`/`thinkingLevel` fold to the CLI util only.

No CRITICAL/MAJOR findings.

File: src/lib/providers/anthropic/client.ts [hunk 6]
```diff
@@ -1463,15 +1463,32 @@ export class AnthropicProvider extends BaseProvider {
...
+          // temperature but 1 while `thinking` is set, and does not honour
+          // top_p there. The CLI always sends a default temperature, so
+          // forwarding it alongside thinking turns a call that used to work
+          // into a 400. Drop the sampling knobs for exactly those turns and
+          // let Anthropic's thinking defaults stand.
+          ...(!thinking && samplingParams.temperature !== undefined
             ? { temperature: samplingParams.temperature }
             : {}),
-          ...(samplingParams.topP !== undefined
+          ...(!thinking && samplingParams.topP !== undefined
```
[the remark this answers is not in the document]
murdore: Checked this against the live API before acting on it, and it does not reproduce on this branch. Three configurations, with the outbound request captured by wrapping `fetch` (real `generate()`, real provider, nothing stubbed):

| config | `tool_choice` on the wire | `thinking` | result |
| --- | --- | --- | --- |
| `toolChoice: "required"` + thinking | `null` | `enabled` | 200 |
| named tool choice + thinking | `null` | `enabled` | 200 |
| `toolChoice: "required"`, thinking off | `null` | `null` | 200 |

No `tool_choice` field reaches Anthropic in any of them, so the incompatible pair is never formed and no 400 is possible. A rejection guard here would defend against a request this path cannot currently send.

I checked the obvious explanation and it is not that: `toolChoice` **is** present in the `buildGenerateTextOptions` allowlist, **is** declared on the public `GenerateOptions`, and the forwarding call exists at `client.ts:1880`. The code to send it is there and produces nothing — including with thinking disabled, so it is unrelated to this PR.

That last row is the part worth following up: a caller passing `toolChoice: "required"` or a named tool on the generate path is silently getting Anthropic default `auto`. That is a separate pre-existing defect, out of scope here, and #1668 states it fixes exactly that — so it is worth confirming against that branch. If #1668 does restore forwarding, your concern stops being hypothetical and the guard becomes worth adding, on that PR rather than this one.

Not dismissing the finding — recording that it is currently unreachable, and why.

Tara-ag:
**Verdict: APPROVE**

Recurring review of head `3be92def`. All three previously-posted `yama` MINOR findings are resolved in the current head — the requested fixes are present and well-formed, and I found no new issues introduced by them. The core defect (silent `thinkingConfig` discard in the `buildGenerateTextOptions` allowlist) is correctly and minimally fixed.

## Findings — all resolved in this head

| Severity | Location | Status | Finding |
| --- | --- | --- | --- |
| MINOR | `src/lib/neurolink.ts:5618-5632` | ✅ resolved | Allowlist comment now correctly attributes the `thinking`/`thinkingBudget`/`thinkingLevel` → `thinkingConfig` fold to the CLI util `src/lib/utils/thinkingConfig.ts`, not `normalizeTextOptions` (which is a plain spread). Scope conclusion is unchanged and correct. |
| MINOR | `src/lib/providers/anthropic/client.ts:1466-1472` | ✅ resolved | Silent `temperature`/`top_p` drop on thinking generate turns is now surfaced via `logger.debug`. |
| MINOR | `src/lib/providers/anthropic/client.ts:2373-2379` | ✅ resolved | Streaming-path `!thinking` sampling guard now has coverage: the new case drives `NeuroLink().stream()` with `temperature: 0` + `thinkingConfig` and asserts no outbound thinking request carried a temperature. |

## What I verified on this head

- **`thinkingConfig` forwarding (Rule 5, backward compat):** `buildGenerateTextOptions` has exactly **one caller** (`runStandardGenerateRequest`, confirmed via code graph). `thinkingConfig` was previously dropped by the allowlist, so forwarding it is additive — no unmodified caller changes behaviour.
- **Sampling-drop completeness:** generate guards both `temperature` and `top_p`; the stream path derives `streamSamplingParams` from `temperature` only, so its single `!thinking` guard is complete (no `top_p` asymmetry).
- **`logger` safety:** both `logger.debug` calls use the module-level `logger` import (`client.ts:70`, from `src/lib/utils/logger.ts`) — not shadowed in either scope.
- **Wire-level / test correctness:** the suite drives the real public `generate()`/`stream()` (`dist/index.js`), wraps `fetch` only to observe (restored in `finally`), asserts preconditions before claims, and correctly avoids asserting on `result.reasoning` (redacted-thinking beta returns empty reasoning). Conforms to Rule 15 (e2e-only, single module graph). Registered as `test:thinking-config`.
- **Blast radius:** the 159-file impact is the hub nature of `neurolink.ts`, not a consequence of adding one allowlist field.
- docs/api + `llms-full.txt` are line-number churn; `package.json` is a script registration. No secrets, no registry/static-import change, single-commit policy satisfied.

## The one unresolved thread (CodeRabbit, not mine)

`PRRT_kwDOOzxF1c6hgfiu` (MAJOR: reject forced/named `toolChoice` when thinking is enabled) remains unresolved. The author's reply adequately justifies leaving it out of this PR: wrapped-fetch evidence shows `tool_choice` never reaches the wire via the generate path today (a pre-existing, separate defect), so the incompatible pair cannot yet form — and the fix belongs with #1668, which restores toolChoice forwarding. I concur and am not reposting it, on the condition that it is picked up against #1668 before that merges.

No CRITICAL or MAJOR findings in this change. Ready to merge.

Tara-ag (approved):
**APPROVE** — the fix is correct, minimal, and backward-compatible (Rule 5). Verified against current head `3be92def`.

- `buildGenerateTextOptions` now forwards `thinkingConfig` through the allowlist; single caller confirmed — no unmodified caller changes behaviour.
- Generate + stream builders drop Anthropic `temperature`/`top_p` when `thinking` is set (Anthropic rejects any temperature but 1 while thinking is enabled), now surfaced via `logger.debug` rather than silently.
- Streaming-path sampling guard has live coverage (`"an explicit temperature is dropped on a thinking stream turn"`).
- Allowlist comment correctly attributes the thinking-trio fold to the CLI util (`src/lib/utils/thinkingConfig.ts`).

No CRITICAL or MAJOR findings. Re-affirming approval at the current head.

src/lib/providers/anthropic/client.ts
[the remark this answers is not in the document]
Tara-ag: Thanks for the empirical check — this matches my conclusion. The thinking+`toolChoice` 400 was a potential materialisation of a path that predates this PR, and your captured requests show no `tool_choice` field ever reaches the wire on the generate route, so the incompatible pair cannot be formed here. The underlying `toolChoice`-silently-dropped defect is a separate pre-existing issue, correctly tracked against #1668 (which states it restores forwarding) rather than this PR. Agreed it is out of scope for #1680. No action needed on this thread.

Tara-ag:
**Verdict: APPROVE**

Recurring review of head `35510935` (the amended single commit). All previously-posted `yama` MINOR findings are resolved in this head, the requested fixes are present and correct, and I found **no new issues** introduced by the amendment.

## Findings

| Severity | Location | Status | Finding |
| --- | --- | --- | --- |
| MINOR | `src/lib/neurolink.ts:5618-5632` | ✅ resolved | Allowlist comment now correctly attributes the `thinking`/`thinkingBudget`/`thinkingLevel`→`thinkingConfig` fold to the CLI util `src/lib/utils/thinkingConfig.ts` (not `normalizeTextOptions`, which is a plain spread). Scope conclusion unchanged and correct. |
| MINOR | `src/lib/providers/anthropic/client.ts:1466-1492` | ✅ resolved | Silent `temperature`/`top_p` drop on thinking **generate** turns is now surfaced via `logger.debug` (guarded on `thinking && (temperature !== undefined \|\| topP !== undefined)`). |
| MINOR | `src/lib/providers/anthropic/client.ts:2373-2377, 2387-2391` | ✅ resolved | Streaming-path `!thinking` sampling guard is now covered: the case "an explicit temperature is dropped on a thinking stream turn" drives a real `NeuroLink().stream()` with `temperature: 0` + `thinkingConfig` and asserts no observed request carried a temperature. |

The one remaining thread (`PRRT_kwDOOzxF1c6hgfiu`, CodeRabbit MAJOR: reject forced/named `toolChoice` when thinking is enabled) stays accepted-not-reposted: the author's wrapped-`fetch` evidence shows `tool_choice` never reaches the wire on the generate path today (a pre-existing, separate defect), so the incompatible pair cannot currently form — and it is correctly deferred to #1668, which restores toolChoice forwarding. Not a blocker for this PR.

## What I verified on this head

- **Core fix (Rule 5 / backward compat):** `buildGenerateTextOptions` has **exactly one caller**, `runStandardGenerateRequest` (confirmed via code graph, call at `neurolink.ts:5146`). `thinkingConfig` was previously discarded by the allowlist, so forwarding it (`thinkingConfig: options.thinkingConfig`) is purely additive — no unmodified caller changes behaviour. The larger blast radius (159+ files) is `neurolink.ts`'s hub nature, not a consequence of adding one allowlist field.
- **Sampling-drop completeness:** generate guards both `temperature` and `top_p` behind `!thinking`; the stream path derives `streamSamplingParams` from `temperature` only, so its single `!thinking` guard (line 2389) is complete — no asymmetry.
- **`thinking` derivation / logger safety:** `thinking` is derived from `options.thinkingConfig` and in closure for both guards; both new `logger.debug` calls use the module-level `logger` import (not shadowed in either scope).
- **Test suite** (`continuous-test-suite-sdk-thinking-config.ts`): live end-to-end, single module graph (`dist/index.js` + harness helpers), `assertDistFresh`, `skipUnlessProviderAvailable`, `fetch` wrapper restored in `finally`. Conforms to Rule 15. Correctly asserts on the wire (thinking field present/absent) rather than on `result.reasoning` (redacted-thinking beta returns empty reasoning on the proxy path — well explained). Registered as `test:thinking-config` in `package.json`.
- **Scaffolding:** `package.json` is a script registration; `docs/api/classes/NeuroLink.md` + `docs-site/static/llms-full.txt` are typedoc line-number churn. No secrets introduced; no registry/static-import change (Rule 1); Rule 3 (Gemini tools/json-schema) and Rule 6 (`formatProviderError`) untouched. Single-commit policy satisfied.

No CRITICAL or MAJOR findings. Ready to merge.

Tara-ag (approved):
**APPROVE** — re-affirming at the current head `35510935`.

Recurring review validation pass. All previously-posted `yama` MINOR findings are resolved in this head, and no new issues were introduced:

- `client.ts` generate + stream builders now **log** the sampling-drop (`logger.debug`) instead of discarding `temperature`/`top_p` silently when `thinking` is set (threads `PRRT_kwDOOzxF1c6he7Ld`, `PRRT_kwDOOzxF1c6he9OW`).
- Streaming-path `!thinking` guard now has live coverage — the suite drives `NeuroLink().stream()` with `temperature: 0` + `thinkingConfig` (threads `PRRT_kwDOOzxF1c6he7g4`, `PRRT_kwDOOzxF1c6he9Wy`).
- `neurolink.ts` allowlist comment now correctly attributes the `thinking`/`thinkingBudget`/`thinkingLevel` → `thinkingConfig` fold to the CLI util `src/lib/utils/thinkingConfig.ts`, not `normalizeTextOptions` (thread `PRRT_kwDOOzxF1c6hfsTe`).
- CodeRabbit MAJOR (toolChoice+thinking → 400) accepted-not-reposted: author's wrapped-`fetch` evidence shows `tool_choice` never reaches the wire via the generate route today, so the incompatible pair is unreachable; tracked against #1668 (thread `PRRT_kwDOOzxF1c6hgfiu`, resolved with ack).

Backward compatible (Rule 5): `buildGenerateTextOptions` has a single caller (`runStandardGenerateRequest`); `thinkingConfig` was previously dropped by the allowlist, so forwarding it is additive. No CRITICAL/MAJOR findings. Ready to merge.

Tara-ag:
**Verdict: APPROVE**

Recurring review of head `80cbce413`. All previously-posted `yama` MINOR findings are resolved in this head by explicit code change (not just replies), and I found **no new issues** introduced by those fixes. The core defect — `buildGenerateTextOptions`' field-by-field allowlist silently swallowing the documented `thinkingConfig` so no provider ever saw it — is correctly and minimally fixed.

## Findings

| Severity | Location | Status | Description |
| -------- | -------- | ------ | ----------- |
| MINOR | `src/lib/neurolink.ts:5618-5632` | ✅ resolved | Allowlist comment now correctly attributes the `thinking`/`thinkingBudget`/`thinkingLevel`→`thinkingConfig` fold to the CLI util `src/lib/utils/thinkingConfig.ts` (not `normalizeTextOptions`, which is a plain spread). Scope conclusion unchanged and correct. |
| MINOR | `src/lib/providers/anthropic/client.ts:1466-1491` | ✅ resolved | Silent drop of explicit `temperature`/`top_p` on thinking **generate** turns is now surfaced via `logger.debug`, and both spreads are guarded by `!thinking`. |
| MINOR | `src/lib/providers/anthropic/client.ts:2373-2391` | ✅ resolved | Streaming-path `!thinking` sampling guard now carries a `logger.debug`, and now has coverage: the case "an explicit temperature is dropped on a thinking stream turn" drives a real `NeuroLink().stream()` with `temperature: 0` + `thinkingConfig` and asserts no observed request carried a temperature. |

The one remaining thread (`PRRT_kwDOOzxF1c6hgfiu`, CodeRabbit MAJOR: reject forced/named `toolChoice` when thinking is enabled) stays **accepted-not-reposted**: the author's wrapped-`fetch` evidence shows `tool_choice` never reaches the wire on the generate path today (a pre-existing, separate defect), so the incompatible pair cannot yet form — and it is correctly deferred to #1668, which restores toolChoice forwarding. Not a blocker for this PR.

## What I verified on this head

- **Core fix (Rule 5 / backward compat):** `buildGenerateTextOptions` has **exactly one caller**, `runStandardGenerateRequest` (confirmed via code graph, call at `neurolink.ts:5146`). `thinkingConfig` was previously discarded by the allowlist, so forwarding it (`thinkingConfig: options.thinkingConfig`) is purely additive — no unmodified caller changes behaviour. The large blast-radius number is `neurolink.ts`'s hub nature (imported 160+ files), not a consequence of adding one allowlist field.
- **Sampling-drop completeness:** generate guards both `temperature` and `top_p` behind `!thinking`; the stream path derives `streamSamplingParams` from `temperature` only, so its single `!thinking` guard (line 2389) is complete — no asymmetry.
- **`thinking` derivation / logger safety:** `thinking` is derived from `options.thinkingConfig` and in closure scope for both guards; the new `logger.debug` calls use the module-level `logger` import.
- **Test suite** (`continuous-test-suite-sdk-thinking-config.ts`): live end-to-end, single module graph (`dist/index.js` + harness helpers), `assertDistFresh`, `skipUnlessProviderAvailable`, global `fetch` wrapper restored in `finally`. Conforms to Rule 15. Correctly asserts on the wire (thinking present/absent) rather than on `result.reasoning` (redacted-thinking beta returns empty reasoning on the OAuth subscription path — well explained). Registered as `test:thinking-config` in `package.json`.
- **Scaffolding:** `package.json` is a script registration; `docs/api/classes/NeuroLink.md` is typedoc line-number churn. No secrets introduced; no registry/static-import change (Rule 1); Rule 3 (Gemini tools/json-schema) and Rule 6 (`formatProviderError`) untouched. Single-commit policy satisfied.

## One disclosed, non-blocking note

The PR description candidly records that `test:thinking-config` is wired into **no CI workflow** (same as `test:matrix:cli`, `test:json-e2e`, and #1682's suite); the red/green evidence was produced by running the suite against both trees by hand. The author's stated intent is one follow-up wiring the orphaned live suites together after #1681 lands, rather than scattering secret blocks across three PRs. This is a disclosed, reasoned limitation consistent with the repo's existing orphaned-live-suite state — not something introduced by this change. Worth tracking that a live-suite run of this PR is actually exercised before/at the follow-up, since this is precisely the class of regression (an allowlist silently dropping a field) that only a wired-in live suite catches.

No CRITICAL or MAJOR findings in this change. Ready to merge.

Tara-ag (approved):
**APPROVE** — re-affirming at the current head `80cbce413`.

All previously-posted `yama` MINOR findings are resolved in this head by explicit code change (not just replies), and no new issues were introduced by those fixes:

- `client.ts` generate + stream builders now surface the sampling-drop via `logger.debug` instead of discarding `temperature`/`top_p` silently when `thinking` is set (threads `PRRT_kwDOOzxF1c6he7Ld`, `PRRT_kwDOOzxF1c6he9OW`, both resolved).
- The streaming-path `!thinking` sampling guard now has live coverage — the suite drives `NeuroLink().stream()` with `temperature: 0` + `thinkingConfig` (threads `PRRT_kwDOOzxF1c6he7g4`, `PRRT_kwDOOzxF1c6he9Wy`, both resolved).
- `neurolink.ts` allowlist comment now correctly attributes the `thinking`/`thinkingBudget`/`thinkingLevel` → `thinkingConfig` fold to the CLI util `src/lib/utils/thinkingConfig.ts`, not `normalizeTextOptions` (thread `PRRT_kwDOOzxF1c6hfsTe`, resolved).
- CodeRabbit MAJOR (toolChoice + thinking → 400) stays accepted-not-reposted: the author's wrapped-`fetch` evidence shows `tool_choice` never reaches the wire on the generate route today, so the incompatible pair cannot form; correctly deferred to #1668 (thread `PRRT_kwDOOzxF1c6hgfiu`, resolved).

Backward compatible (Rule 5): `buildGenerateTextOptions` has a single caller (`runStandardGenerateRequest`); `thinkingConfig` was previously dropped by the allowlist, so forwarding it is additive. No CRITICAL/MAJOR findings. Ready to merge.

PR #1681, opened 2026-09-11
chore(ci): wire test:matrix:cli into the nightly live-matrix workflow

murdore:
## The gap

`test:matrix:cli` has existed in `package.json:112` since the CLI mirror suite was written, and **no workflow has ever run it**. `live-matrix.yml` had exactly one test step, hard-coded to `pnpm run test:matrix`, so the SDK sweep received real provider credentials every night and its CLI twin received nothing. A grep for `test:matrix:cli` across `.github/workflows/` returned zero hits.

That matters because the two suites do not overlap. The CLI mirror exists to catch what the SDK sweep structurally cannot see — argument parsing, env var loading, exit codes, and bundling drift between the library and the binary. None of it was under test anywhere.

## Verification

A nightly workflow cannot be exercised by PR CI, so this was verified by live probe in both directions rather than by a repo test. Stating that plainly rather than implying coverage the change cannot have.

| Condition | Result |
| --- | --- |
| No credentials | `No providers selected.` · Passed 0 / Total 0 · `RESULT: PASS` · **exit 0** |
| Real `ANTHROPIC_API_KEY` | `✓ [anthropic] CLI generate`, `✓ [anthropic] CLI stream` · 2/2 · 8.83s |
| `grep -c 'run: pnpm run test:matrix'` | **1 → 2** |
| YAML | parses via js-yaml; steps 7 → 8; 86 → 130 lines |

The credential-free case is the one that matters for safety: a rotated-out key degrades to a clean pass, not a red nightly. The workflow is deliberately **not** a required check, so neither outcome can block a merge.

## Two costs, named

- The nightly job now sweeps every credentialed provider twice, and there is no job-level timeout today, so a wedged provider gets two chances per night to run long.
- The ~40-line secrets block is now duplicated in one file, so onboarding a provider must touch both copies.

## Deliberately not included

**41 of the 112 `test:*` scripts are invoked by no workflow at all**, and several are genuinely live-tier — `test:providers`, `test:new-providers`, all three `test:mcp:sdk/cli/http`, `test:observability`, `test:context`, `test:memory`, `test:rag`, `test:agents-live`, and `test:middleware` (explicitly held back for a Vertex-throttling bug).

`test:matrix:cli` is the only one with a one-to-one live sibling already receiving nightly secrets, which is what makes it a clean bounded fix. Each of the rest needs its own secret set, cost budget and skip-safety audit. Folding them in here would produce a change nobody could review.

## Summary by CodeRabbit

* **Tests**
  * Added live provider matrix coverage for CLI workflows across supported provider configurations.
  * Added live reasoning parity validation to verify consistent reasoning behavior.
  * Applied a 30-minute runtime limit to SDK and CLI provider matrix checks and a 15-minute limit to reasoning parity validation, helping prevent stalled test runs.

### DIFF
--- a/.github/workflows/live-matrix.yml
+++ b/.github/workflows/live-matrix.yml
@@ -41,8 +41,13 @@ jobs:
       - name: Build package
         run: pnpm run build
 
+      # Bounded because this job now runs two live sweeps back to back. The
+      # CLI harness already caps each child process, but nothing capped the
+      # step, and GitHub's default job ceiling is 360 minutes — long enough
+      # for one wedged provider to hold a nightly runner most of a day.
       - name: Live provider matrix sweep (self-gates cleanly when keys are absent)
         run: pnpm run test:matrix
+        timeout-minutes: 30
         env:
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
           ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
@@ -68,6 +73,51 @@ jobs:
           # The catalog derives <ID>_API_KEY, so Cloudflare is read as
           # CLOUDFLARE_API_KEY. This passed CLOUDFLARE_API_TOKEN, which
           # nothing reads — Cloudflare was silently skipped every night.
+          CLOUDFLARE_API_KEY: ${{ secrets.CLOUDFLARE_API_KEY }}
+          CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+          CEREBRAS_API_KEY: ${{ secrets.CEREBRAS_API_KEY }}
+          SAMBANOVA_API_KEY: ${{ secrets.SAMBANOVA_API_KEY }}
+          BASETEN_API_KEY: ${{ secrets.BASETEN_API_KEY }}
+          GMICLOUD_API_KEY: ${{ secrets.GMICLOUD_API_KEY }}
+          INCEPTION_LABS_API_KEY: ${{ secrets.INCEPTION_LABS_API_KEY }}
+          IO_INTELLIGENCE_API_KEY: ${{ secrets.IO_INTELLIGENCE_API_KEY }}
+          UPSTAGE_API_KEY: ${{ secrets.UPSTAGE_API_KEY }}
+          MANCER_API_KEY: ${{ secrets.MANCER_API_KEY }}
+          VOYAGE_API_KEY: ${{ secrets.VOYAGE_API_KEY }}
+          JINA_API_KEY: ${{ secrets.JINA_API_KEY }}
+          STABILITY_API_KEY: ${{ secrets.STABILITY_API_KEY }}
+          IDEOGRAM_API_KEY: ${{ secrets.IDEOGRAM_API_KEY }}
+          RECRAFT_API_KEY: ${{ secrets.RECRAFT_API_KEY }}
+          REPLICATE_API_TOKEN: ${{ secrets.REPLICATE_API_TOKEN }}
+
+      - name: Live provider matrix sweep — CLI (self-gates cleanly when keys are absent)
+        run: pnpm run test:matrix:cli
+        timeout-minutes: 30
+        env:
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
+          GOOGLE_AI_API_KEY: ${{ secrets.GOOGLE_AI_API_KEY }}
+          GOOGLE_SERVICE_ACCOUNT_KEY: ${{ secrets.GOOGLE_SERVICE_ACCOUNT_KEY }}
+          GOOGLE_VERTEX_PROJECT: ${{ secrets.GOOGLE_VERTEX_PROJECT }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          SAGEMAKER_ENDPOINT: ${{ secrets.SAGEMAKER_ENDPOINT }}
+          AZURE_OPENAI_API_KEY: ${{ secrets.AZURE_OPENAI_API_KEY }}
+          AZURE_OPENAI_ENDPOINT: ${{ secrets.AZURE_OPENAI_ENDPOINT }}
+          MISTRAL_API_KEY: ${{ secrets.MISTRAL_API_KEY }}
+          HUGGINGFACE_API_KEY: ${{ secrets.HUGGINGFACE_API_KEY }}
+          OPENROUTER_API_KEY: ${{ secrets.OPENROUTER_API_KEY }}
+          DEEPSEEK_API_KEY: ${{ secrets.DEEPSEEK_API_KEY }}
+          GROQ_API_KEY: ${{ secrets.GROQ_API_KEY }}
+          COHERE_API_KEY: ${{ secrets.COHERE_API_KEY }}
+          FIREWORKS_API_KEY: ${{ secrets.FIREWORKS_API_KEY }}
+          PERPLEXITY_API_KEY: ${{ secrets.PERPLEXITY_API_KEY }}
+          XAI_API_KEY: ${{ secrets.XAI_API_KEY }}
+          NVIDIA_NIM_API_KEY: ${{ secrets.NVIDIA_NIM_API_KEY }}
+          TOGETHER_API_KEY: ${{ secrets.TOGETHER_API_KEY }}
+          # Same fix already applied to the SDK step above: the catalog derives
+          # <ID>_API_KEY, so Cloudflare must be passed as CLOUDFLARE_API_KEY,
+          # not CLOUDFLARE_API_TOKEN (nothing reads that name).
           CLOUDFLARE_API_KEY: ${{ secrets.CLOUDFLARE_API_KEY }}
           CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
           CEREBRAS_API_KEY: ${{ secrets.CEREBRAS_API_KEY }}


Tara-ag (commented):
Reviewing the CLI-matrix wiring in the nightly live-matrix workflow.

.github/workflows/live-matrix.yml
```diff
@@ -84,3 +84,47 @@ jobs:
           IDEOGRAM_API_KEY: ${{ secrets.IDEOGRAM_API_KEY }}
           RECRAFT_API_KEY: ${{ secrets.RECRAFT_API_KEY }}
           REPLICATE_API_TOKEN: ${{ secrets.REPLICATE_API_TOKEN }}
+
+      - name: Live provider matrix sweep — CLI (self-gates cleanly when keys are absent)
```
Tara-ag, line 88: MINOR — the PR body names this cost ("no job-level timeout today, so a wedged provider gets two chances per night to run long") but leaves it unaddressed. The CLI harness bounds each child process (120s SIGKILL→Skip), so nothing hangs forever, but since this job now runs *two* live sweeps back-to-back, a single job-level `timeout-minutes` is a cheap, safe ceiling. GitHub Actions default is 360 min per job.

```suggestion
        run: pnpm run test:matrix:cli
        timeout-minutes: 30
```

(If you'd rather bound both sweeps together, set `timeout-minutes` on the job instead of the step.)

.github/workflows/live-matrix.yml
```diff
@@ -84,3 +84,47 @@ jobs:
...
+        run: pnpm run test:matrix:cli
+        env:
```
Tara-ag, line 90: MINOR — this 37-line `env:` block is now duplicated verbatim from the SDK step above. The PR body names the cost ("onboarding a provider must touch both copies"), and it's a real drift hazard: the next provider added will need to be edited in two places, and a mismatch silently changes nightly coverage. Since both suites share the exact same `hasProviderEnv` gating (they read the same `PROVIDERS` table), the two steps provably need the same secret set — so dedupe them with a YAML anchor defined once and merged into both steps:

```yaml
      - name: Live provider matrix sweep (self-gates cleanly when keys are absent)
        run: pnpm run test:matrix
        env: &live_provider_env
          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
          # …all other keys…

      - name: Live provider matrix sweep — CLI (self-gates cleanly when keys are absent)
        run: pnpm run test:matrix:cli
        env:
          <<: *live_provider_env
```

This makes onboarding a one-line change and eliminates the drift the author flagged. If anchors feel opaque, an alternative is to hoist the `env:` block into a job-level `env:` (both steps inherit identical vars).

Tara-ag (commented):
Both findings are MINOR and previously acknowledged in the PR body; no blocker to merge.

Tara-ag:
Superseded — this summary reviewed an intermediate commit and is now folded into the single authoritative summary: [#issuecomment-5653696070](https://github.com/juspay/neurolink/pull/1681#issuecomment-5653696070). Current verdict on head `6d33ab0f`: **APPROVE**.

.github/workflows/live-matrix.yml
Tara-ag: Correcting the suggestion anchor: as written it was bound to the `- name:` line (88) only, and the two-line block would have replaced that single line — dropping the step name and introducing a stray 8-space-indented `run:`. The intent is to keep `- name:` and add `timeout-minutes` after `run:`. Apply this as the two-line range 88–89 instead:

```
      - name: Live provider matrix sweep — CLI (self-gates cleanly when keys are absent)
        run: pnpm run test:matrix:cli
        timeout-minutes: 30
```

Tara-ag (approved):
## ✅ APPROVE — `chore(ci): wire test:matrix:cli into the nightly live-matrix workflow`

CI-only change; no library code touched (graph blast radius 0.00, zero affected flows). The added CLI step self-gates cleanly when secrets are absent, shares the SDK suite's `hasProviderEnv` / `PROVIDERS` env contract, and runs after `pnpm run build` produces `dist/cli/index.js`. Only two non-blocking MINORs — see the inline comments.

Tara-ag:
Superseded — duplicate/stale; see the single authoritative summary [#issuecomment-5653696070](https://github.com/juspay/neurolink/pull/1681#issuecomment-5653696070). Current verdict on head `6d33ab0f`: **APPROVE**.

.github/workflows/live-matrix.yml
murdore: Taken, in `4607b5e42` — and on both sweeps rather than just the CLI one, since the same argument applies to each.

```yaml
      - name: Live provider matrix sweep (self-gates cleanly when keys are absent)
        run: pnpm run test:matrix
        timeout-minutes: 30
      # ...
      - name: Live provider matrix sweep — CLI (self-gates cleanly when keys are absent)
        run: pnpm run test:matrix:cli
        timeout-minutes: 30
```

Step-level rather than job-level deliberately: a job ceiling would also bound checkout, install and build, so a slow `pnpm install` could eat the budget meant for the sweeps and the failure would point at the wrong step.

Thanks for correcting your own suggestion anchor, by the way — as first written it was bound to the `- name:` line and would have replaced it, dropping the step name and leaving a stray `run:`. I applied the intent, not the diff.

**Verified mechanically, because the risk in editing this file is silently changing which secrets a sweep sees.** I parsed the workflow before and after and asserted that all 8 steps survive, that every step's `name`, `run` and `env` map is byte-identical to before, that no step gained an `env` block, and that the only addition anywhere is `timeout-minutes: 30` on steps 6 and 7. That check is what makes this safe to land on an already-approved PR without a live test — a nightly workflow cannot be exercised by PR CI, so an assertion about the file itself is the honest substitute.

.github/workflows/live-matrix.yml
murdore: I'd rather not take this one as written, and I want to give you the actual reason rather than just decline.

**The concern is right.** The duplication is a real drift hazard, the two steps provably need the same secret set for the reason you give (both read the same `PROVIDERS` table through the same `hasProviderEnv` gate), and "onboarding a provider must touch both copies" is a cost I named in the body without fixing.

**What stops me is that I cannot verify the construct where it matters.** GitHub does not document anchor or merge-key support for workflow files, there is no precedent for one anywhere in this repo's 13 workflows, and `actionlint` is not available here. I did check what I could: both a YAML 1.1 parser (PyYAML) and a YAML 1.2 parser (js-yaml 4) resolve `<<: *shared` correctly into the full map. That is necessary but not sufficient — neither of them is GitHub's parser.

The asymmetry is what decides it. This job runs unattended at 03:00 on a schedule, and it is explicitly *not* a required check, so it blocks nothing when it breaks. If the merge key silently failed to resolve, the CLI sweep would run with an empty `env`, every provider would self-gate to "no keys", and the run would go **green while testing nothing** — the exact shape of failure this repo has been bitten by before and now documents at length. A visible 37-line duplication is a worse-looking problem than an invisible one, but it is a much better problem to have.

Two ways I'm happy to close it:

1. **Job-level `env:`** — certainly supported, no anchors involved, and it makes onboarding a one-line change. The trade-off is that every step then inherits the secrets, including `pnpm install`, which widens what a malicious postinstall could read. Given this repo already runs a supply-chain cooldown and a banned-deps gate, I'd want that to be a deliberate call rather than a side effect of a readability fix.
2. **Settle the anchor question empirically** — land it on a branch and dispatch the workflow once, confirming from the run's resolved env that the merge key expanded. Cheap, and it turns an assumption into a fact.

Tell me which and I'll do it. Left open rather than resolved, because the underlying hazard is real and unaddressed.

.github/workflows/live-matrix.yml
Tara-ag: Confirmed from the tree at `4607b5e4`: both `test:matrix` and `test:matrix:cli` steps now carry `timeout-minutes: 30`. Your step-level choice (rather than job-level) is the right one — a job ceiling would let a slow `pnpm install` consume the sweep budget and misattribute the failure. Resolving this finding.

.github/workflows/live-matrix.yml
Tara-ag: Accepting your decline. The reasoning holds: in an unattended, non-blocking workflow the failure shape that matters is *silent-green-while-testing-nothing*, and an unreviewed anchor/merge-key that GitHub's parser doesn't resolve would produce exactly that. Your asymmetry point (a visible 37-line duplication vs an invisible mis-resolved env) is decisive, and job-level `env:`'s widen-security trade-off is a real cost you correctly ruled out. The drift hazard remains real but it is review-visible. No action required; resolving.

Tara-ag:
Superseded — duplicate/stale; see the single authoritative summary [#issuecomment-5653696070](https://github.com/juspay/neurolink/pull/1681#issuecomment-5653696070). Current verdict on head `6d33ab0f`: **APPROVE**.

Tara-ag (approved):
APPROVE — current head `4607b5e` carries the resolution to the one actionable finding: both live sweeps now have `timeout-minutes: 30` at the step level (applied and verified mechanically). The env-block duplication remains, but as jointly agreed that is a review-visible, non-blocking MINOR. CI-only change; no library code affected; self-gates cleanly on absent keys. Approving on the final commit.

Tara-ag (commented):
Review of the new `test:reasoning-parity-live` step.

.github/workflows/live-matrix.yml
```diff
@@ -84,3 +89,64 @@ jobs:
...
+      # registered in ci.yml's `extended-suites`, but that tier runs with no
+      # provider credentials by design, so `skipUnlessProviderAvailable`
+      # makes every live leg SKIP there — registered, never exercised. This
+      # job is the only place in CI that holds DEEPSEEK_API_KEY, which is the
+      # one credential the suite reads.
+      #
+      # Bounded like the two sweeps above: the suite samples the generate leg
+      # repeatedly to avoid failing on a single quiet turn, so its worst case
+      # is several live round trips rather than one.
+      - name: Live reasoning parity — generate() vs stream() (SKIPs cleanly without the key)
```
Tara-ag: **CRITICAL** — this step invokes `pnpm run test:reasoning-parity-live`, but that script does not exist in `package.json`, so the step **fails every night** instead of "self-gating cleanly".

`pnpm run $script` errors with `Missing script: test:reasoning-parity-live` before any TypeScript executes, so `skipUnlessProviderAvailable` never gets a chance — a missing script is a hard failure, not a clean SKIP. The step is the exact mirror-in-reverse of the "silent-green-while-testing-nothing" failure mode this repo guards against elsewhere: here it's a *red* nightly that tests nothing.

Using the graph surface of the codebase (`test: @juspay/neurolink package.json` @ `b1ed3994`, and a whole-repo code search for `reasoning-parity`):

- **No `test:reasoning-parity-live` script** in the head `package.json` `scripts` map (every `test:` entry enumerated; the closest is `test:matrix:cli`).
- **No reasoning-parity suite file** anywhere in the repo — `search_code` for `reasoning-parity` returns 0 hits on the default branch, and none of the known parity facilities (`test/helpers/skipIf.ts`, `providerMatrix.ts`, harness) name it.

The surrounding comment also misstates the registration: it claims the suite "is registered in ci.yml's `extended-suites`", but the `extended-suites` job's `SUITES` list (checked at `b1ed3994`) contains no `test:reasoning-parity-*` entry.

**Fix** — either add the missing npm script (and the suite it points at) to the PR, or remove this step. If the intent is a reasoned generate-vs-stream parity check against DeepSeek, the script + suite must ship with this change:

````suggestion
# (Add ``test:reasoning-parity-live`` to package.json scripts pointing at the
# reasoning-parity suite, with a skipUnlessProviderAvailable("deepseek") guard
# inside every leg — then the pnpm run below resolves.)  Otherwise, delete
# this whole step rather than ship a guaranteed-red nightly.
````

> Note: I could not fully reproduce the skip-guard claim because the suite file does not exist to inspect; if it does exist under an unsearched name, point me at it in a reply and I'll re-verify.

**Review by Tara-ag — COMMENTED** [review]
Inline finding on the new reasoning-parity step (missing script).

**Comment by Tara-ag** [comment]
Superseded — this NEEDS_WORK was raised against the intermediate `b1ed3994` commit and has since been resolved; see the single authoritative summary [#issuecomment-5653696070](https://github.com/juspay/neurolink/pull/1681#issuecomment-5653696070). Current verdict on head `6d33ab0f`: **APPROVE**.

File: .github/workflows/live-matrix.yml [hunk 4]
```diff
@@ -84,3 +89,64 @@ jobs:
...
+        run: pnpm run test:reasoning-parity-live
```
Tara-ag: ## CRITICAL — the step invokes a script that does not exist

**MAJOR (argued as the top blocker on this PR).** This new step runs `pnpm run test:reasoning-parity-live`, but that script is not defined anywhere in `package.json` at the current head (`b1ed3994`). I verified the full `scripts` map at head: `test:matrix` and `test:matrix:cli` are present; `test:reasoning-parity-live` is not. A repo-wide code search for `reasoning-parity` likewise finds no suite file or package.json entry in this repository.

### What breaks

`pnpm run test:reasoning-parity-live` fails immediately with:

```
Missing script: "test:reasoning-parity-live"
```

before any TypeScript executes — so the `skipUnlessProviderAvailable("deepseek")` guard the step's comment relies on **never runs**. The commit message's premise ("registered in ci.yml's extended-suites") is unrelated to whether the script exists, and the guard cannot self-gate a script that is not there to run it.

Consequence: this step turns a **green** nightly into a **guaranteed red** one. That inverts the change's core safety property — every other step on this job was explicitly designed to self-gate to a clean pass when keys are absent. This one cannot.

### Fix

Two correct options, pick one:

**Option A — drop the step** until the script + its e2e suite actually exist:

```suggestion
```
*(delete the `- name:` … `DEEPSEEK_API_KEY` block, lines 148–152)*

**Option B — land the missing piece in the same PR.** Add to `package.json` a script that points at a real e2e suite (`test/continuous-test-suite-reasoning-parity.ts` — the repo's suites are end-to-end, Rule 15), and add that file with the `skipUnlessProviderAvailable("deepseek")` guard the comment describes. Then `pnpm run test:reasoning-parity-live` becomes a real, self-gating command.

Merging a step that calls a known-missing script is not mergeable in its current form.

Tara-ag (changes requested):
## NEEDS_WORK — new step calls a script that does not exist

The `b1ed3994` commit adds a reasoning-parity step that runs `pnpm run test:reasoning-parity-live`, but that script is not defined in `package.json` (verified against the full `scripts` map at head) and no such suite exists in the repo (repo-wide search: 0 hits). The step will fail with `Missing script`, turning a green nightly into a guaranteed red one. See the inline finding for the two correct fixes (drop the step, or land the script + e2e suite).

Tara-ag:
Superseded — duplicate/stale NEEDS_WORK against intermediate `b1ed3994`; see the single authoritative summary [#issuecomment-5653696070](https://github.com/juspay/neurolink/pull/1681#issuecomment-5653696070). Current verdict on head `6d33ab0f`: **APPROVE**.

Tara-ag:
## ✅ APPROVE — single-file CI change that bounds the live sweeps

**Verdict: APPROVE.** The branch head `6d33ab0f` against base `release` reduces to an 8-line, single-file change to `.github/workflows/live-matrix.yml`: `timeout-minutes: 30` added to both the SDK and the CLI live-matrix sweeps, each with an accurate rationale comment. CI-only, no library code touched (knowledge-graph risk 0.00, zero affected flows). No remaining findings; the current diff is mergeable.

This is the single authoritative summary. It supersedes all earlier `yama:summary` comments on this PR, which tracked the intermediate commits and have since been folded into this one.

### Findings (all resolved)

| Thread | Finding | Disposition |
| --- | --- | --- |
| `…he592` | Sweeps had no step `timeout-minutes` | **Resolved** — both `test:matrix` and `test:matrix:cli` carry `timeout-minutes: 30`, step-level (correct: a job ceiling would let a slow `pnpm install` eat the sweep budget and misattribute the failure). |
| `…he6MY` | 37-line `env:` block duplicated from the SDK step (drift hazard) | **Resolved** — author's decline accepted: an unverifiable YAML anchor/merge-key risks a *silent-green-while-testing-nothing* failure in this unattended, non-blocking nightly; the visible duplication is the safer failure mode. |
| `…h3-ZP` / `…h3-rr` | Reasoning-parity step invoked a nonexistent `test:reasoning-parity-live` script | **Resolved by removal / scope reduction** — that step was an artifact of a second commit landing on the branch; it (and the wiring) now live in PR #1673. This PR's diff no longer touches it. |
| `…h3-w4` | CodeRabbit thread (missing parity script) | **Resolved** — confirmed addressed. |

### Checked and found clean

- **Self-gating is real** — `continuous-test-suite-provider-matrix-cli.ts` filters targets via `hasProviderEnv()`; with no keys it prints "No providers selected", runs zero tests, and exits 0 → a rotated-out credential degrades to a clean pass.
- **Env-var contract identical** — both suites gate on the same `PROVIDERS` table in `test/helpers/providerMatrix.ts` via the same `<ID>_API_KEY` derivation (incl. the `CLOUDFLARE_API_KEY` vs `_API_TOKEN` fix).
- **Build ordering correct** — `pnpm run build` (prepack → `build:cli`) precedes both sweeps, so `dist/cli/index.js` exists.
- **Not a required check** — neither run-duplication nor transient upstream outages can block merges.
- **No secrets leaked** — every value is `${{ secrets.* }}`; single-commit policy compliant.

Note: the PR still reports `mergeable_state: blocked` — consistent with branch-protection's single-commit/linear-history validator asking for a rebase to exactly one commit. That is an external branch-state condition, not a review blocker.

Tara-ag (approved):
## ✅ APPROVE — wire `test:matrix:cli` into the nightly live-matrix sweep

CI-only change (`.github/workflows/live-matrix.yml`); no library code touched — code-review graph risk 0.00, zero affected flows. The added CLI step self-gates cleanly on absent keys via `hasProviderEnv()`, shares the SDK suite's `PROVIDERS` env contract, and runs after `pnpm run build` produces `dist/cli/index.js`. Both live sweeps carry step-level `timeout-minutes: 30` (correct choice — a job ceiling would let a slow `pnpm install` eat the sweep budget).

This is submitted on the current head `4607b5e` and supersedes the earlier `CHANGES_REQUESTED` (raised against the intermediate `b1ed3994` commit whose `test:reasoning-parity-live` step has since been removed). No remaining findings; the current diff is mergeable.

.github/workflows/live-matrix.yml
```diff
@@ -84,3 +89,64 @@ jobs:
...
+        timeout-minutes: 15
+        env:
+          DEEPSEEK_API_KEY: ${{ secrets.DEEPSEEK_API_KEY }}
```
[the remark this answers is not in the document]
murdore: Correct, and the step is gone.

The root cause was not a missing script — it was that this branch had picked up a second commit it was never supposed to carry. `b1ed3994a` (the reasoning-parity wiring) landed here when #1698 was merged into the branch, on top of this PR's own `4607b5e42`. That also broke `🔒 Single Commit Policy Validation`.

The branch is back to its single commit, `4607b5e42`, which contains only the CLI-matrix wiring this PR is about. The `test:reasoning-parity-live` step no longer exists here, so there is nothing left to invoke a nonexistent script.

The wiring itself was not dropped — it moved to https://github.com/juspay/neurolink/pull/1673, which is the PR that actually adds `test:reasoning-parity-live` to `package.json`, and it is a separate `reasoning-parity` job there rather than a step on this sweep. Verified conflict-free against this branch in both merge orders with `git merge-tree` before pushing.

Tara-ag (approved):
## Approving — the PR is now exactly what was requested

This is a recurring review. Current head `6d33ab0f` against base `release` is a single-file, 8-line change that only adds the two `timeout-minutes: 30` bounds. All findings from earlier rounds are resolved:

- **Missing timeout on `Live provider matrix sweep` / CLI mirror (he592)** — ✅ Resolved. Both steps now carry `timeout-minutes: 30` with comments explaining the rationale (an unbounded step could otherwise hold a nightly runner for up to GitHub's 360-minute default ceiling; the CLI harness already caps child processes but the step itself was unbounded).
- **Env-var duplication between the two sweeps (he6MY)** — ✅ Resolved (author declined; reasoning accepted — deduping risks a silent-green failure mode if keys are dropped from one source).
- **Missing `test:reasoning-parity-live` script (h3-ZP / h3-rr)** — ✅ Resolved via scope reduction. The reasoning-parity job, `test:matrix:cli` step and `json-e2e` step now actually live in the base `release` branch (and the parity script exists there); this PR's diff no longer touches them, so the earlier "missing script" concern is no longer in scope.
- **CodeRabbit thread (h3-w4)** — ✅ Resolved.

**Verdict on the current change:** clean and correct. Both `timeout-minutes: 30` values are present, correctly placed at the step level (immediately after `run:`), and the added comments accurately describe the conservative bound. This is a workflow/CI-only change with low blast radius (knowledge-graph risk: 0.00). No new issues found.

Note: the PR branch still reports `mergeable_state: blocked` — consistent with the branch-protection "single commit / linear history" validator requiring a rebase with exactly one commit. Content-wise there is nothing blocking; that's an external branch-state condition, not a review blocker.

Approving.

PR #1682, opened 2026-09-11
fix(anthropic): send a string toolChoice instead of dropping it

murdore:
## The defect

`toolChoice` accepts four shapes: the strings `"auto"`, `"none"`, `"required"`, and the object `{ type: "tool", toolName }`. **On the Anthropic generate path the three string forms never reached the provider.**

`resolveToolChoice` (`utils/toolChoice.ts:14`) returns a bare string for those cases, but the request builder read `.type` off the value before translating it:

```ts
options.toolChoice.type === "tool"
  ? { type: "tool", toolName: options.toolChoice.toolName }
  : options.toolChoice.type          // a string has no .type → undefined
```

So `toolChoiceToAnthropic` was handed `undefined`, returned `undefined`, and the field was omitted from the request. A caller asking for a **required** tool silently got Anthropic's default `auto` — the model free to answer without calling anything, and nothing raised to say the option had been discarded.

The translator was never at fault. It maps `"required"` → `{ type: "any" }` correctly one screen up; it simply was never given the string.

## A sibling already had it right

The streaming path passes the value straight through:

```ts
toolChoiceToAnthropic(resolveToolChoice(options, toolsRecord, shouldUseTools))
```

This was a generate-only divergence. The two now agree. The object branch stays explicit so a V3-shaped `{ type: "required" }` also normalises to the string the translator expects.

## Proof — at the wire, red then green

| case | release `10fa282f2` | this branch |
| --- | --- | --- |
| `"required"` reaches the provider | ✗ **FAIL** — *no outbound request carried a tool_choice* | ✓ `{type:"any"}` |
| object form unchanged | ✓ | ✓ |
| omitting it sends no `tool_choice` | ✓ | ✓ |
| | 2 passed, **1 failed** | **3 passed** |

**Only the first case discriminates.** The other two pass on the release build as well — they are there to catch this fix breaking the object form that already worked, or inventing a tool choice for callers who supplied none. Stating that plainly rather than presenting 3/3 as though all three were evidence.

### Why it asserts on the request, not on tool usage

Whether a model calls a tool is its own decision, so asserting "a tool was called" would be flaky in both directions. What this fix changes is the bytes we send. The suite wraps `fetch` to record the outbound request and otherwise drives the real public `generate()` against the real provider — nothing is stubbed, the real response comes back.

## Noted, not fixed here

A forced tool choice with **no `maxSteps` cap does not reliably terminate** — the model can be compelled to call the tool again on every step. The suite caps it deliberately. That behaviour deserves its own look and is out of scope.

## Gates

`pnpm run check` 0 errors · `pnpm run check:tools-tests` 0 errors · `build` clean · `docs/api` drift 0 (the change is inside a function body, no signature change, so the docs-site artifacts correctly need no regeneration) · pre-commit hook passed without bypass · registered nothing new in `package.json` beyond the suite file itself.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Anthropic tool selection so string choices such as `"auto"`, `"none"`, and `"required"` are honored correctly.
  - Ensured object-based tool choices continue to pass through as expected.
  - Prevented incomplete tool selections from being sent as invalid requests.

- **Tests**
  - Added coverage for string and object tool choices, omitted settings, and incomplete tool selections in generated Anthropic requests.

### DIFF
--- a/src/lib/providers/anthropic/client.ts
+++ b/src/lib/providers/anthropic/client.ts
@@ -116,7 +116,10 @@ import {
   TimeoutError,
 } from "../../utils/timeout.js";
 import { raceWithAbort } from "../../utils/async/index.js";
-import { resolveToolChoice } from "../../utils/toolChoice.js";
+import {
+  normalizeResolvedToolChoice,
+  resolveToolChoice,
+} from "../../utils/toolChoice.js";
 import { emitToolEndFromStepFinish } from "../../utils/toolEndEmitter.js";
 import type { LanguageModel, Tool } from "../../types/index.js";
 import { NoOutputGeneratedError } from "../../utils/generationErrors.js";
@@ -1403,12 +1406,22 @@ export class AnthropicProvider extends BaseProvider {
         if (tools && tools.length === 0) {
           tools = undefined;
         }
-        let toolChoice = options.toolChoice
-          ? toolChoiceToAnthropic(
-              options.toolChoice.type === "tool"
-                ? { type: "tool", toolName: options.toolChoice.toolName }
-                : options.toolChoice.type,
-            )
+        // The value here has already been through `resolveToolChoice`
+        // upstream in `executeNativeGenerate`, so this layer translates
+        // rather than re-resolves — re-resolving would apply the auto/none
+        // defaulting twice, and `toolsRecord`/`shouldUseTools` are not in
+        // scope inside `getAISDKModel` in any case.
+        //
+        // Only the SHAPE varies, and reading `.type` off it was the bug:
+        // a bare `"required"` has no `.type`, so the translator received
+        // `undefined` and the field was dropped from the request entirely.
+        // Normalisation now lives in the shared util rather than being
+        // re-derived here, so this site cannot drift from its siblings again.
+        const normalisedToolChoice = normalizeResolvedToolChoice(
+          options.toolChoice,
+        );
+        let toolChoice = normalisedToolChoice
+          ? toolChoiceToAnthropic(normalisedToolChoice)
           : undefined;
 
         // JSON/structured output: Anthropic has no response_format — emulate
--- a/src/lib/utils/toolChoice.ts
+++ b/src/lib/utils/toolChoice.ts
@@ -13,3 +13,45 @@ export function resolveToolChoice(
 
   return options.toolChoice ?? "auto";
 }
+
+/**
+ * Normalise an already-resolved tool choice into the shape
+ * `toolChoiceToAnthropic` (and its peers) accept.
+ *
+ * A provider's model handle sees a value that has *already* been through
+ * `resolveToolChoice` upstream, so it must translate rather than re-resolve —
+ * re-resolving there would apply the `auto`/`none` defaulting twice, and the
+ * `tools`/`shouldUseTools` inputs it needs are not in scope at that layer
+ * anyway.
+ *
+ * What does vary is the shape. The value arrives either as a bare string
+ * (`"auto"` / `"none"` / `"required"`), as `{ type: "tool", toolName }`, or —
+ * on the native loop's tool-free re-ask — as an injected `{ type: "none" }`
+ * object. Reading `.type` off the string forms yields `undefined`, which is
+ * how string choices were silently dropped from the Anthropic generate
+ * request. This exists so that normalisation lives in one place instead of
+ * being re-derived at each call site.
+ */
+export function normalizeResolvedToolChoice(
+  value: unknown,
+): ToolChoice<Record<string, Tool>> | undefined {
+  if (typeof value === "string") {
+    return value as ToolChoice<Record<string, Tool>>;
+  }
+  if (value && typeof value === "object") {
+    const candidate = value as { type?: string; toolName?: string };
+    if (candidate.type === "tool") {
+      // A tool choice that names no tool cannot be honoured. Fall through to
+      // the provider default rather than emitting the bare string "tool",
+      // which is not one of the valid auto/none/required forms and would be
+      // rejected downstream.
+      return candidate.toolName
+        ? { type: "tool", toolName: candidate.toolName }
+        : undefined;
+    }
+    if (candidate.type) {
+      return candidate.type as ToolChoice<Record<string, Tool>>;
+    }
+  }
+  return undefined;
+}
--- a/test/continuous-test-suite-anthropic-toolchoice.ts
+++ b/test/continuous-test-suite-anthropic-toolchoice.ts
@@ -1,0 +1,250 @@
+#!/usr/bin/env tsx
+import "dotenv/config";
+
+/**
+ * Continuous Test Suite — string-form toolChoice reaches Anthropic, LIVE
+ *
+ * `toolChoice` accepts four shapes: the strings `"auto"`, `"none"`,
+ * `"required"`, and the object `{ type: "tool", toolName }`. On the Anthropic
+ * generate path the three string forms never reached the provider.
+ *
+ * `resolveToolChoice` returns a bare string for those cases, but the request
+ * builder read `.type` off the value before translating it. A string has no
+ * `.type`, so `toolChoiceToAnthropic` received `undefined`, returned
+ * `undefined`, and the field was dropped from the request entirely. A caller
+ * asking for a required tool silently got Anthropic's default `auto` — the
+ * model was free to answer without calling anything, and nothing was raised.
+ * The object form worked, because `.type === "tool"` is truthy.
+ *
+ * The translator itself was never at fault: it maps `"required"` to
+ * `{ type: "any" }` correctly. It simply was not being given the string. The
+ * streaming path already passed the value straight through, so this was a
+ * generate-only divergence from a sibling that had it right.
+ *
+ * WHY THIS ASSERTS ON THE REQUEST. Whether a model chooses to call a tool is
+ * its decision, so asserting "a tool was called" would be flaky in both
+ * directions — a model may call one without being forced, and may be forced
+ * and still produce a turn this suite reads as ambiguous. What the fix
+ * actually changes is the bytes we send. The suite therefore wraps `fetch` to
+ * record the outbound request. That is observation, not substitution: the call
+ * goes through the real public `generate()` to the real provider and the real
+ * response comes back. Nothing is stubbed.
+ *
+ * Only the first case discriminates. The other two pass on the release build
+ * as well and exist to catch this fix breaking the shape that already worked,
+ * or switching tool choice on for callers who never asked.
+ *
+ * `maxSteps` is capped deliberately: a forced tool choice with no cap does not
+ * reliably terminate, since the model can be compelled to call the tool again
+ * on every step. That is worth its own look and is not what this suite tests.
+ *
+ * Run: pnpm run build && npx tsx test/continuous-test-suite-anthropic-toolchoice.ts
+ */
+
+import { z } from "zod";
+import { NeuroLink, tool } from "../dist/index.js";
+import { defineSuite, assert } from "./helpers/harness.js";
+import { skipUnlessProviderAvailable } from "./helpers/skipIf.js";
+import { assertDistFresh } from "./helpers/distFreshness.js";
+
+assertDistFresh();
+
+const { test, runSuite } = defineSuite(
+  "Anthropic toolChoice passthrough (live)",
+);
+
+const PROVIDER = "anthropic";
+const MODEL = "claude-sonnet-4-5-20250929";
+const PROMPT = "What is the weather in Tokyo?";
+
+// The capture used to coerce every non-object `tool_choice` to `null`, which
+// made it blind to exactly the failure it is meant to watch for: a bare string
+// reaching the wire looked identical to no tool_choice at all. Record the raw
+// value and narrow at the point of use instead.
+type SeenRequest = {
+  toolChoice: Record<string, unknown> | string | null;
+};
+
+/** Narrow a captured tool_choice to its object form, or null. */
+const asObject = (
+  value: SeenRequest["toolChoice"],
+): Record<string, unknown> | null =>
+  value !== null && typeof value === "object" ? value : null;
+
+const weatherTool = () => ({
+  get_weather: tool({
+    description: "Returns the current weather for a city",
+    inputSchema: z.object({ city: z.string() }),
+    execute: async ({ city }: { city: string }) => ({ city, tempC: 22 }),
+  }),
+});
+
+/**
+ * Record every outbound Messages request while `run` executes, restoring the
+ * original `fetch` even if the call throws so one failure cannot leave a
+ * wrapped `fetch` behind for the next case.
+ */
+const observeRequests = async (
+  run: () => Promise<unknown>,
+): Promise<SeenRequest[]> => {
+  const seen: SeenRequest[] = [];
+  const original = globalThis.fetch;
+  globalThis.fetch = async (input: RequestInfo | URL, init?: RequestInit) => {
+    if (String(input).includes("/v1/messages")) {
+      let body: Record<string, unknown>;
+      try {
+        body = JSON.parse(String(init?.body ?? "{}")) as Record<
+          string,
+          unknown
+        >;
+      } catch {
+        body = {};
+      }
+      const choice = body.tool_choice;
+      seen.push({
+        toolChoice:
+          choice === undefined || choice === null
+            ? null
+            : (choice as Record<string, unknown> | string),
+      });
+    }
+    return original(input, init);
+  };
+  try {
+    await run();
+    return seen;
+  } finally {
+    globalThis.fetch = original;
+  }
+};
+
+await test('the string form "required" reaches the provider', async () => {
+  skipUnlessProviderAvailable(PROVIDER);
+
+  const seen = await observeRequests(() =>
+    new NeuroLink().generate({
+      input: { text: PROMPT },
+      provider: PROVIDER,
+      model: MODEL,
+      maxTokens: 2000,
+      maxSteps: 2,
+      toolChoice: "required",
+      tools: weatherTool(),
+    }),
+  );
+
+  // PRECONDITION: an assertion about a request's contents is vacuous if no
+  // request was captured.
+  assert(
+    seen.length > 0,
+    "precondition failed: no Messages request was observed",
+  );
+
+  // The regression. Pre-fix the builder read `.type` off the string, the
+  // translator got undefined, and the field never left the process.
+  assert(
+    seen.some((r) => r.toolChoice !== null),
+    "no outbound request carried a tool_choice, so the string form never reached the provider",
+  );
+  assert(
+    seen.some((r) => asObject(r.toolChoice)?.type === "any"),
+    'the string form did not translate to Anthropic\'s "any" tool choice',
+  );
+});
+
+await test("the object form still reaches the provider unchanged", async () => {
+  skipUnlessProviderAvailable(PROVIDER);
+
+  const seen = await observeRequests(() =>
+    new NeuroLink().generate({
+      input: { text: PROMPT },
+      provider: PROVIDER,
+      model: MODEL,
+      maxTokens: 2000,
+      maxSteps: 2,
+      toolChoice: { type: "tool", toolName: "get_weather" },
+      tools: weatherTool(),
+    }),
+  );
+
+  assert(
+    seen.length > 0,
+    "precondition failed: no Messages request was observed",
+  );
+
+  // Control: this shape worked before the fix and must keep working.
+  assert(
+    seen.some(
+      (r) =>
+        asObject(r.toolChoice)?.type === "tool" &&
+        asObject(r.toolChoice)?.name === "get_weather",
+    ),
+    "the object form no longer names the requested tool on the wire",
+  );
+});
+
+await test('a tool choice naming no tool sends no tool_choice rather than a bare "tool"', async () => {
+  skipUnlessProviderAvailable(PROVIDER);
+
+  const seen = await observeRequests(() =>
+    new NeuroLink().generate({
+      input: { text: PROMPT },
+      provider: PROVIDER,
+      model: MODEL,
+      maxTokens: 2000,
+      maxSteps: 2,
+      // Malformed on purpose: `type: "tool"` with no `toolName`. Not
+      // reachable from a well-typed caller, but normalisation reads an
+      // `unknown`, so the shape has to be handled rather than assumed away.
+      toolChoice: { type: "tool" } as unknown as never,
+      tools: weatherTool(),
+    }),
+  );
+
+  assert(
+    seen.length > 0,
+    "precondition failed: no Messages request was observed",
+  );
+
+  // `"tool"` is not one of Anthropic's valid auto/none/required strings, so
+  // forwarding it would be worse than dropping it: the request is rejected
+  // outright instead of falling back to the default.
+  assert(
+    seen.every((r) => r.toolChoice !== "tool"),
+    "a request carried the bare string tool_choice",
+  );
+  assert(
+    seen.every((r) => r.toolChoice === null),
+    "a nameless tool choice still reached the wire",
+  );
+});
+
+await test("omitting toolChoice sends no tool_choice", async () => {
+  skipUnlessProviderAvailable(PROVIDER);
+
+  const seen = await observeRequests(() =>
+    new NeuroLink().generate({
+      input: { text: PROMPT },
+      provider: PROVIDER,
+      model: MODEL,
+      maxTokens: 2000,
+      maxSteps: 2,
+      tools: weatherTool(),
+    }),
+  );
+
+  assert(
+    seen.length > 0,
+    "precondition failed: no Messages request was observed",
+  );
+
+  // Control in the other direction: normalising the value must not invent a
+  // tool choice for callers who never supplied one. Anthropic defaults to
+  // auto when the field is absent, which is the behaviour to preserve.
+  assert(
+    seen.every((r) => r.toolChoice === null),
+    "a request carried a tool_choice even though the caller supplied none",
+  );
+});
+
+await runSuite();
--- a/package.json
+++ b/package.json
@@ -111,6 +111,7 @@
     "test:matrix": "pnpm exec tsx test/continuous-test-suite-provider-matrix.ts",
     "test:matrix:cli": "pnpm exec tsx test/continuous-test-suite-provider-matrix-cli.ts",
     "test:thinking-config": "pnpm exec tsx test/continuous-test-suite-sdk-thinking-config.ts",
+    "test:anthropic-toolchoice": "pnpm exec tsx test/continuous-test-suite-anthropic-toolchoice.ts",
     "test:mcp:spans": "pnpm exec tsx test/continuous-test-suite-mcp-spans.ts",
     "test:mcp:infra": "pnpm exec tsx test/continuous-test-suite-mcp-infra.ts",
     "test:vendor-recovery": "pnpm exec tsx test/continuous-test-suite-native-vendor-recovery.ts",


src/lib/providers/anthropic/client.ts
```diff
@@ -1371,12 +1371,36 @@ export class AnthropicProvider extends BaseProvider {
...
+            type?: string;
+            toolName?: string;
+          };
+          normalisedToolChoice =
+            candidate.type === "tool"
+              ? { type: "tool", toolName: candidate.toolName }
+              : candidate.type;
+        }
+        let toolChoice = normalisedToolChoice
+          ? toolChoiceToAnthropic(normalisedToolChoice)
```
Tara-ag, line 1392: **MAJOR — the fix does not actually give parity with the sibling generate/stream paths.**

Both other Anthropic paths resolve the choice through `resolveToolChoice` before translating:
- `executeNativeGenerate` — `client.ts:1887`
- `executeStreamInCaptureScope` — `client.ts:2095` → `toolChoiceToAnthropic(resolveToolChoice(options, toolsRecord, shouldUseTools))`

`resolveToolChoice` (`src/lib/utils/toolChoice.ts:3`):
```ts
if (!shouldUseTools || !tools || Object.keys(tools).length === 0) {
  return "none";          // suppress tool usage
}
return options.toolChoice ?? "auto";   // default to auto when unset
```

This PR keeps `getAISDKModel` reading raw `options.toolChoice` and hand-rolls a *third* string/object normaliser instead of calling `resolveToolChoice`. So it still diverges from the sibling paths whenever `resolveToolChoice` would do more than pass-through — e.g. defaulting an unset choice to `"auto"`, or forcing `"none"` when `!shouldUseTools` / no tools are declared. The stated goal ("The two now agree") is therefore only structurally true for the three verbatim string cases; the implementation copy of the normalization logic that the bug fix just repaired is now maintained in three places (`getAISDKModel`, `executeNativeGenerate`, `executeStreamInCaptureScope`), and the resolved `auto`/`none` semantics are dropped here.

Suggested fix — reuse the same resolver used by the sibling paths when `toolsRecord`/`shouldUseTools` are in scope:
```suggestion
const resolved = resolveToolChoice(options, toolsRecord, shouldUseTools);
let toolChoice = resolved
  ? toolChoiceToAnthropic(resolved)
  : undefined;
```
If `getAISDKModel` genuinely lacks `toolsRecord`/`shouldUseTools` here, at minimum factor the string/object normalisation it now duplicates into the shared `toolChoice` util so the three sites stop drifting.

Tara-ag (commented):
Reviewed the toolChoice fix. One MAJOR finding: the fixed `getAISDKModel` path hand-rolls its own string/object normalisation instead of delegating to `resolveToolChoice` like `executeNativeGenerate:1887` and `executeStreamInCaptureScope:2095`, so the claimed parity is only structural for the verbatim string cases and the resolved `auto`/`none` semantics are dropped in this path.

Tara-ag:
*Superseded historical summary from an earlier review recurrence.* The current authoritative summary for this PR is comment **#5649289323** (verdict APPROVE). This comment is retained only to preserve thread history and no longer reflects the current verdict; it has been de-duplicated per the one-summary rule.

src/lib/providers/anthropic/client.ts
Tara-ag: **MAJOR** — the claimed parity with the streaming path is only structural for the verbatim string cases; the resolved `auto`/`none` behaviour `resolveToolChoice` provides is dropped here.

Both sibling paths delegate to `resolveToolChoice` (see `executeNativeGenerate:1887` and `executeStreamInCaptureScope:2095`, which call `toolChoiceToAnthropic(resolveToolChoice(options, toolsRecord, shouldUseTools))`). `resolveToolChoice` doesn't just echo strings — it defaults a missing choice to `"auto"` and, when `!shouldUseTools` or no tools are present, returns `"none"`. This hand-rolled normaliser reimplements only the string/object branch and bypasses that logic, so on `getAISDKModel`:
- a caller supplying no `toolChoice` gets `undefined` (no defaulting), and
- the `!shouldUseTools`/no-tools suppression to `"none"` never runs.

So the fix is correct for the three strings it tests, but it does not actually converge on the shared path the PR description claims it now "matches" — it forks a third normaliser instead of reusing the one the other two paths use. Recommend delegating:

```suggestion
        const normalisedToolChoice = resolveToolChoice(
          options,
          toolsRecord,
          shouldUseTools,
        );
        let toolChoice = normalisedToolChoice
          ? toolChoiceToAnthropic(normalisedToolChoice)
          : undefined;
```

This matches the string semantics exactly (string in → `"required"` → `{ type: "any" }`) while also applying the defaulting and suppression the other generate/stream paths rely on. If `resolveToolChoice` cannot be reused here (e.g. signature differences in this scope), factor the duplicated normalisation into a shared helper instead of keeping a third copy. Reference: Rule 3 adjacent — provider generation paths should stay on one behaviour. Evidence that `getAISDKModel` reads raw `options.toolChoice` instead of the resolved value is at the `const rawToolChoice: unknown = options.toolChoice;` line above.

Tara-ag (changes requested):
Needs work: one MAJOR finding on the Anthropic `getAISDKModel` path — the fix hand-rolls a third toolChoice normaliser instead of delegating to `resolveToolChoice` like the sibling generate/stream paths, so the resolved `auto`/`none` defaulting and suppression semantics still don't apply there. See the inline comment. Full table in the summary comment.

Tara-ag:
*Superseded historical summary from an earlier review recurrence.* The current authoritative summary for this PR is comment **#5649289323** (verdict APPROVE). This comment is retained only to preserve thread history and no longer reflects the current verdict; it has been de-duplicated per the one-summary rule.

src/lib/providers/anthropic/client.ts
murdore: Half of this is right and is now fixed; half of it cannot be done at this site. Taking them separately, because the distinction matters for where the logic should live.

**The suggested change does not compile here.** `getAISDKModel()` is declared `public getAISDKModel(): LanguageModel` at `client.ts:1304` and takes no arguments — `toolsRecord` and `shouldUseTools` are locals of `executeNativeGenerate`, not of the model handle. Neither is in scope inside the `doGenerate` closure.

**And re-resolving here would be wrong even if they were.** `executeNativeGenerate:1887` already calls `resolveToolChoice(options, toolsRecord, true)` and hands the *resolved* value to the loop, which forwards it as `args.toolChoice`. Resolving a second time at the model layer would apply the `auto`/`none` defaulting twice. The two sibling sites you cite are the sites that *do* the resolving; this one is downstream of them and should only translate.

There is also a fourth shape neither of us listed: the native loop injects `{ type: "none" }` for its tool-free re-ask (`nativeGenerateLoop.ts:175`), so this site must accept an object form as well as the strings.

**Your actual concern was correct and is addressed.** Three copies of the shape normalisation had drifted, and one of them was the bug. I have taken your fallback suggestion and factored it into the shared util as `normalizeResolvedToolChoice` (`src/lib/utils/toolChoice.ts`), with the reasoning recorded there: translate, do not re-resolve; only the shape varies. The client now calls it instead of hand-rolling the branch.

**Re-verified after the refactor**, because a tidier implementation that quietly stopped discriminating would be worse than the duplication it replaced:

| | release | branch |
| --- | --- | --- |
| `"required"` reaches the provider | ✗ *no outbound request carried a tool_choice* | ✓ `{type:"any"}` |
| object form unchanged | ✓ | ✓ |
| omitting it sends nothing | ✓ | ✓ |
| | 2 passed, **1 failed** | **3 passed** |

Head is now `07186c20d`. If you would still prefer the resolution itself to move down into the model handle, that is a larger change — it would mean threading `tools`/`shouldUseTools` through `getAISDKModel`, and I would rather do that deliberately than as part of a one-expression bug fix.

test/continuous-test-suite-anthropic-toolchoice.ts
```diff
@@ -0,0 +1,201 @@
...
+  // Control in the other direction: normalising the value must not invent a
+  // tool choice for callers who never supplied one. Anthropic defaults to
+  // auto when the field is absent, which is the behaviour to preserve.
+  assert(
+    seen.every((r) => r.toolChoice === null),
+    "a request carried a tool_choice even though the caller supplied none",
+  );
+});
+
+await runSuite();
```
Tara-ag, line 201: **MAJOR — this suite is not registered anywhere, so the regression the whole PR exists to fix will never be guarded by CI.**

This is a live suite (drives `dist/index.js` with a real `ANTHROPIC_API_KEY`). Unlike every sibling, it has:

- **no `test:` script** — `package.json` is untouched by this PR; there is no `test:anthropic-toolchoice` entry (compare `test:anthropic-streaming-retry`, `test:anthropic-loop-characterization`, etc.),
- **no CI wiring** — it is absent from `ci.yml`'s `extended-suites` `SUITES` list, from `test:live`, and from `live-matrix.yml` (which only runs `test:matrix`).

The repo states the exact consequence in `ci.yml`: *"A suite wired into nothing is documentation, not a test."* `check:test-parse` only parses it for backticks — it never executes. So the 3 assertions that discriminate the fix (`"required"` → `{type:"any"}` on the wire, object form unchanged, omission sends nothing) will never run in CI, and the bug could silently regress on the next change to `getAISDKModel` with nothing turning red.

Fix — register it (live tier). Concretely:

```suggestion
await runSuite();
```
plus add a `"test:anthropic-toolchoice": "pnpm exec tsx test/continuous-test-suite-anthropic-toolchoice.ts"` script in `package.json` and include it in the `test:live` aggregation so it runs wherever live provider tests run.

Tara-ag (commented):
Core fix reviewed and accepted; one new MAJOR finding on test registration.

Tara-ag:
*Superseded historical summary from an earlier review recurrence.* The current authoritative summary for this PR is comment **#5649289323** (verdict APPROVE). This comment is retained only to preserve thread history and no longer reflects the current verdict; it has been de-duplicated per the one-summary rule.

Tara-ag:
*Superseded historical summary from an earlier review recurrence.* The current authoritative summary for this PR is comment **#5649289323** (verdict APPROVE). This comment is retained only to preserve thread history and no longer reflects the current verdict; it has been de-duplicated per the one-summary rule.

Tara-ag (changes requested):
Needs work: the regain fix is correct and the resolveToolChoice parity concern is resolved, but the new live suite that proves the fix is not registered anywhere (no `package.json` script, absent from `ci.yml` extended-suites and `test:live`) — so the regression the PR exists to fix will never be guarded by CI. See the inline comment on `test/continuous-test-suite-anthropic-toolchoice.ts:201` and the summary comment. Register the suite and this is good to merge.

test/continuous-test-suite-anthropic-toolchoice.ts
murdore: Accepted, and checking it turned up that the gap is wider than this PR.

**Script registered** — `"test:anthropic-toolchoice"` is now in `package.json` at head `ad84b7657`, matching the convention of the sibling Anthropic suites.

**But registering a script does not make CI run it, and I want to be precise rather than let this read as solved.** I checked every live suite in the repo:

| live suite | `test:` script | appears in any workflow |
| --- | --- | --- |
| `test:matrix` | yes | yes (nightly) |
| `test:providers` | yes | yes |
| `test:matrix:cli` | yes | **no** |
| `test:json-e2e` | yes | **no** |
| `test:thinking-config` (#1680) | yes | **no** |
| `test:anthropic-toolchoice` (this PR) | now yes | **no** |

So your quote from `ci.yml` — *"a suite wired into nothing is documentation, not a test"* — applies to four suites, two of which I added today. **#1680 has the identical gap**, and its body says "registered as `test:thinking-config`" without noting that nothing runs it. I am correcting that body too; it overstates what the registration buys.

**Why I am not wiring it here.** A live suite has exactly one CI home: `live-matrix.yml`, the nightly job that carries provider secrets. `extended-suites` is explicitly credential-free — its own comment says those suites drive the public API against stubs, fixtures or mocked fetch — so a suite needing `ANTHROPIC_API_KEY` cannot go there.

And **#1681 already modifies `live-matrix.yml`** — it is the PR wiring `test:matrix:cli` into the nightly, currently green and approved. Adding steps to that same file from this PR would conflict with it.

**Proposed sequence:** merge #1681 first, then one follow-up PR wiring all four orphaned live suites into the nightly together. That keeps the change reviewable in one place instead of scattering secret blocks across three PRs, and avoids conflicting with an approved one.

If you would rather I wire it here and let #1681 rebase around it, say so and I will — I just think the conflict lands on the wrong PR that way.

Tara-ag: Accepting, with the outstanding gap tracked rather than silently closed.

The **script registration is resolved** — `test:anthropic-toolchoice` is in `package.json` at head `ad84b7657`, and your reasoning for not putting it in the credential-free `extended-suites` job is correct (verified: that job's own comment requires stubs/fixtures/mocked-fetch only, so a live `ANTHROPIC_API_KEY` suite cannot go there). The `#1681`-modifies-`live-matrix.yml` conflict is a legitimate reason not to wire it in this PR.

What I want explicit before closing this: **the gap is still present at merge time.** Checking this branch, the suite appears in `package.json` only — not in `test:live` (verified: that aggregate is `providers, mcp:http, mcp:sdk, mcp:cli, observability, context, memory, tool-reliability, evaluation, autoresearch, credentials, dynamic` — no `anthropic-toolchoice`) and in no workflow. Per the repo's own `ci.yml` rule, a suite wired into nothing is documentation, not a test, so the 3 discriminating assertions (`required`→`{type:"any"}` on the wire, object form unchanged, omission sends nothing) still cannot turn red in CI.

So I'm accepting the deferral acknowledged — please open and link the follow-up (the PR wiring all four orphaned live suites into the nightly, per your proposed sequence after #1681 merges) so this does not get lost. That follow-up is the actual resolution; this finding stays "accepted with pending follow-up" rather than "closed."

Tara-ag:
*Superseded historical summary from an earlier review recurrence.* The current authoritative summary for this PR is comment **#5649289323** (verdict APPROVE). This comment is retained only to preserve thread history and no longer reflects the current verdict; it has been de-duplicated per the one-summary rule.

Tara-ag (approved):
Approving — the core fix is correct and verified at the wire, both prior MAJOR findings on `getAISDKModel` parity are resolved (`normalizeResolvedToolChoice`), and the only remaining item (CI wiring of the new live suite) is accepted-with-pending-follow-up per the author's justification (#1681 conflict, credential-free `extended-suites`). See the summary comment for the full table. Follow-up wiring `test:anthropic-toolchoice` into the nightly must be opened and linked.

test/continuous-test-suite-anthropic-toolchoice.ts
murdore: Tracked and linked, as asked: **https://github.com/juspay/neurolink/issues/1684**

It covers all four orphaned live suites rather than just this one — `test:matrix:cli`, `test:json-e2e`, `test:thinking-config` (from #1680) and `test:anthropic-toolchoice` (this PR) — with a table of which suites do and do not appear in a workflow, and why `extended-suites` cannot host them: that job is deliberately credential-free, and these need real provider keys, so `live-matrix.yml` is the only home.

I opened an **issue rather than the PR** for one reason: #1681 already modifies `live-matrix.yml` and is green and approved. A second PR touching that file now would conflict with it. The issue records the sequence — merge #1681, then one follow-up wiring the rest in together, so there is one secret block kept in step in one place.

If you would rather the PR be open immediately even with the conflict, say so and I will raise it. I just did not want to drop a merge conflict onto an approved PR to satisfy a tracking requirement that an issue meets.

One more thing your finding surfaced, which was wider than the finding itself: **#1680 had the same gap and its body overstated it.** It said the suite was "registered as `test:thinking-config`" without noting that nothing runs it. I have corrected that body to quote `ci.yml` directly and to say that the red/green evidence is real but will not catch a regression until the suite is wired in.

murdore: Closing the loop on this one so the PR reads clean: the script is registered at `ad84b7657`, and the wider gap — all four orphaned live suites, not just this one — is tracked at https://github.com/juspay/neurolink/issues/1684, sequenced behind #1681 so it does not conflict with the live-matrix change already in flight.

One correction I owe this thread, found while re-auditing: that issue originally claimed `test:providers` runs in a workflow. It does not. The only hits are `test:providers-mocked`, a different script. So of the six live suites, exactly **one** (`test:matrix`) is CI-wired today — the gap is wider than either of us stated. The issue now carries a dated correction saying so.

Marking this resolved on the strength of your "accepting, with the outstanding gap tracked" — reopen if you'd rather it stay visible until the wiring PR lands.

src/lib/utils/toolChoice.ts
```diff
@@ -13,3 +13,39 @@ export function resolveToolChoice(
...
+ */
+export function normalizeResolvedToolChoice(
+  value: unknown,
+): ToolChoice<Record<string, Tool>> | undefined {
+  if (typeof value === "string") {
+    return value as ToolChoice<Record<string, Tool>>;
+  }
+  if (value && typeof value === "object") {
+    const candidate = value as { type?: string; toolName?: string };
+    if (candidate.type === "tool" && candidate.toolName) {
```
Tara-ag: **MINOR — the refactor narrows an edge-case mapping that the old code handled differently.**

Old `.type === "tool"` branch returned `{ type: "tool", toolName: options.toolChoice.toolName }` **without** requiring `toolName` to be present; the new `&& candidate.toolName` guard now forwards that value.

- Old: input `{ type: "tool" }` (no `toolName`) → `{ type: "tool", toolName: undefined }` → passed to `toolChoiceToAnthropic` as an object.
- New: input `{ type: "tool" }` (no `toolName`) → falls through to `if (candidate.type)`, which is truthy (`"tool"`), so it returns the **bare string `"tool"`** instead.

That string would then be handed to `toolChoiceToAnthropic`, where it is not one of the valid `auto`/`none`/`required` forms. This is not reachable from well-typed callers (a `ToolChoice` with `type: "tool"` is defined to require `toolName`), and the new suite only exercises `{ type: "tool", toolName }`, so nothing guards it — but since the function's contract is "translate the shape, do not reinterpret it", keeping the old passthrough costs nothing and avoids a latent surprise if a bare `{type:"tool"}` ever reaches this util.

```suggestion
    if (candidate.type === "tool") {
      return { type: "tool", toolName: candidate.toolName };
    }
```

If the intent was for the object branch to require `toolName` before treating it as a tool reference, add a short comment stating that explicitly rather than letting it silently collapse to a string.

Tara-ag (commented):
Independent review of the current head (798c30b7). The core fix and the shared util are correct and backward-compatible; one MINOR edge-case divergence noted inline.

Tara-ag:
*Superseded historical summary from an earlier review recurrence.* The current authoritative summary for this PR is comment **#5649289323** (verdict APPROVE). This comment is retained only to preserve thread history and no longer reflects the current verdict; it has been de-duplicated per the one-summary rule.

Tara-ag (approved):
## Verdict: APPROVE

The core fix is correct and verified at the wire: `"required"` now reaches the Anthropic provider as `{type:"any"}`, the object form passes through unchanged, and omitting `toolChoice` sends nothing. Both prior MAJOR findings are resolved (`getAISDKModel` now translates via the shared `normalizeResolvedToolChoice` instead of re-resolving; the suite is registered as `test:anthropic-toolchoice`, with nightly wiring deferred to #1684 for good reason). One non-blocking MINOR finding remains on `toolChoice.ts:43` (bare `{type:"tool"}` collapses to the string `"tool"`); a one-click suggestion is posted there.

Backward-compatible (Rule 5), no signature change, blast radius confined to `getAISDKModel` (single graph-confirmed caller), no secrets or hot-path regressions. Test follows Rule 15 (e2e via `dist/index.js`).

Full detail in the summary comment.

Tara-ag:
*Superseded historical summary from an earlier review recurrence.* The current authoritative summary for this PR is comment **#5649289323** (verdict APPROVE). This comment is retained only to preserve thread history and no longer reflects the current verdict; it has been de-duplicated per the one-summary rule.

src/lib/utils/toolChoice.ts
Tara-ag: Resolving — the described defect is no longer present at the current head, and it is now covered by a test.

Tracing `normalizeResolvedToolChoice` (`src/lib/utils/toolChoice.ts:43-51`) with the malformed `{ type: "tool" }` (no `toolName`):

```ts
if (candidate.type === "tool") {
  return candidate.toolName
    ? { type: "tool", toolName: candidate.toolName }
    : undefined;              // ← nameless tool → returns undefined
}
```

The `if (candidate.type === "tool")` branch is reached first and **returns `undefined`** — it does not fall through to the subsequent `if (candidate.type)` branch, so the bare string `"tool"` can never be emitted. Because `doGenerate` guards on `normalizedToolChoice ? … : undefined`, the case yields no `tool_choice` on the wire, not the invalid `"tool"` string the finding was concerned about.

That behaviour is also locked in by the suite: `test/continuous-test-suite-anthropic-toolchoice.ts:173-207` ("a tool choice naming no tool sends no tool_choice rather than a bare 'tool'") asserts `seen.every((r) => r.toolChoice !== "tool")` and `seen.every((r) => r.toolChoice === null)`.

So the underlying concern — never emit an invalid bare `"tool"` tool_choice — is fully addressed (the resolution is even more defensive than the reviewer's suggested passthrough: it drops the malformed choice to the provider default rather than forwarding `{ type: "tool", toolName: undefined }`). Resolving this finding.

The two earlier MAJOR findings remain correctly resolved: the `getAISDKModel` parity/duplication was factored into the shared `normalizeResolvedToolChoice` util (translate-don't-re-resolve, single caller confirmed at `client.ts:1420`), and the suite-wiring deferral is tracked at #1684.

Tara-ag:
*Superseded historical summary from an earlier review recurrence.* The current authoritative summary for this PR is comment **#5649289323** (verdict APPROVE). This comment is retained only to preserve thread history and no longer reflects the current verdict; it has been de-duplicated per the one-summary rule.

test/continuous-test-suite-anthropic-toolchoice.ts
```diff
@@ -0,0 +1,237 @@
...
+  assert(
+    seen.length > 0,
+    "precondition failed: no Messages request was observed",
+  );
+
+  // `"tool"` is not one of Anthropic's valid auto/none/required strings, so
+  // forwarding it would be worse than dropping it: the request is rejected
+  // outright instead of falling back to the default.
+  assert(
+    seen.every((r) => r.toolChoice !== "tool"),
```
Tara-ag, line 213: **MINOR — this assertion is structurally dead: it can never fail.**

`SeenRequest.toolChoice` is typed `Record<string, unknown> | null`, and `observeRequests` (lines 91-97) collapses **any** non-object `tool_choice` to `null`:

```ts
toolChoice:
  choice && typeof choice === "object"
    ? (choice as Record<string, unknown>)
    : null,
```

So if the provider ever regressed and emitted the bare string `"tool"` as `tool_choice`, it would arrive here as `null`, `r.toolChoice` could never equal `"tool"`, and `seen.every((r) => r.toolChoice !== "tool")` would pass vacuously. The assertion offers no protection against exactly the regression this test says it guards — the "forwarding it [the nameless string] would be worse than dropping it" case.

The real guard in this test is the next assertion, `seen.every((r) => r.toolChoice === null)` (lines 203-206).

Fix — record the raw `tool_choice` value so the dead assertion becomes live (`unknown` instead of object/null):

```suggestion
type SeenRequest = { toolChoice: unknown };

const choice = body.tool_choice;
seen.push({ toolChoice: choice });
```

then adapt the remaining assertions to the raw value (`seen.every((r) => r.toolChoice !== "tool")` and `seen.every((r) => r.toolChoice === null || r.toolChoice === undefined)`). If you'd rather keep the object-coercing observer, drop the dead assertion and let the `=== null` check be the sole guard — as written the first assertion will lull a reader into thinking the bare-`"tool"` regression is covered when it isn't.

Tara-ag (commented):
One MINOR test-correctness observation on the new suite; no blocking issues. The core fix is sound and the two prior MAJOR findings are resolved. Full summary in the yama:summary comment.

Tara-ag:
*Superseded historical summary from an earlier review recurrence.* The current authoritative summary for this PR is comment **#5649289323** (verdict APPROVE). This comment is retained only to preserve thread history and no longer reflects the current verdict; it has been de-duplicated per the one-summary rule.

Tara-ag (approved):
## Verdict: APPROVE

The core fix is correct and verified at the wire; all prior MAJOR findings are resolved; the one open MINOR is non-blocking polish. Full table in the `yama:summary` comment. The only tracked follow-up is CI wiring of the orphaned live suites at #1684 (to land after #1681).

test/continuous-test-suite-anthropic-toolchoice.ts
Tara-ag: Resolving — the premise of this finding no longer holds at the current head (`23b90a`).

The `observeRequests` capture was changed to record the raw `tool_choice` value instead of coercing non-objects to `null`:

```ts
seen.push({
  toolChoice:
    choice === undefined || choice === null
      ? null
      : (choice as Record<string, unknown> | string),
});
```

So if the provider regressed and emitted the bare string `"tool"`, it would now arrive as the string `"tool"` (not `null`), and `seen.every((r) => r.toolChoice !== "tool")` would genuinely fail. The assertion is live again — exactly the fix this finding asked for. The observer is also consistent with the widened `SeenRequest.toolChoice: Record<string, unknown> | string | null` type and the new `asObject` narrowing helper.

The real guard (`=== null`) remains and is correct; the previously-dead first assertion now provides the additional "never emit bare `tool`" protection. No change needed.

Tara-ag:
*Superseded historical summary from an earlier review recurrence.* The current authoritative summary for this PR is comment **#5649289323** (verdict APPROVE). This comment is retained only to preserve thread history and no longer reflects the current verdict; it has been de-duplicated per the one-summary rule.

test/continuous-test-suite-anthropic-toolchoice.ts
Tara-ag: Resolved at head `23b90a03` — the fix this finding asked for is now in the code.

`observeRequests` (`test/continuous-test-suite-anthropic-toolchoice.ts`) no longer coerces every non-object `tool_choice` to `null`. The observer now records the **raw** value — `toolChoice: Record<string, unknown> | string | null`, mapping only `undefined`/`null` to `null` and passing everything else (including a bare string) through:

```ts
toolChoice:
  choice === undefined || choice === null
    ? null
    : (choice as Record<string, unknown> | string),
```

`SeenRequest.toolChoice` is therefore `Record<string, unknown> | string | null` and `seen.every((r) => r.toolChoice !== "tool")` is now **live**: if the provider ever regressed and emitted the bare string `"tool"`, it would be recorded as `"tool"` and the assertion would fail. The dead-assertion problem described in the finding no longer exists.

The headers above `observeRequests` even document the intent: *"The capture used to coerce every non-object `tool_choice` to `null`, which made it blind to exactly the failure it is meant to watch for … Record the raw value and narrow at the point of use instead."* That is precisely the suggested fix. Resolving this thread.

Tara-ag (approved):
## Verdict: APPROVE (head `23b90a03`)

The core fix is correct and verified at the wire; every finding now resolves cleanly on the current head. The final open item — the MINOR dead string-assertion in the new suite — has been fixed by the author (the observer now records the raw `tool_choice`, making `seen.every((r) => r.toolChoice !== "tool")` live again), and I have resolved that thread. No blocking items remain.

- **Fix sound:** `"required"` → `{type:"any"}` on the wire; object `{type:"tool",toolName}` unchanged; omission sends no `tool_choice` (release vs branch: 1 failed → 3 passed).
- **`normalizeResolvedToolChoice`** — single graph-confirmed caller (`getAISDKModel`); preserves behaviour for every shape; nameless `{type:"tool"}` → `undefined` (never the bare `"tool"`).
- **Backward compat (Rule 5):** change is inside a function body, no signature/public-type change.
- **No secrets;** logging/`transformParamsForLogging` untouched; no registry/streaming hot-path regression.
- **Test (Rule 15):** e2e via `dist/index.js`, single module graph, `assertDistFresh`, `skipUnlessProviderAvailable`.
- Single conventional commit; single-commit-policy compliant.

Full detail in the `yama:summary` comment. Only tracked follow-up: nightly CI wiring of the orphaned live suites (incl. this one) at #1684, after #1681 merges.

Tara-ag:
## Verdict: APPROVE

The core fix is correct and verified at the wire, and this recurrence found no new blocking issues. Every prior finding routes clean on the current squashed head.

| --- | --- | --- | --- |
| ~~MAJOR~~ | `client.ts:1420` | `getAISDKModel` hand-rolled a toolChoice normaliser instead of `resolveToolChoice` | **Resolved** — factored into the shared `normalizeResolvedToolChoice` (graph-confirmed single caller `client.ts:1420`); translate-don't-re-resolve is correct here since the value is already resolved upstream by `executeNativeGenerate` |
| ~~MAJOR~~ | `test/…-anthropic-toolchoice.ts` | Suite unregistered / not CI-wired | **Resolved/Accepted** — `test:anthropic-toolchoice` added (`package.json:120`); nightly wiring deferred (credential-free `extended-suites` can't host a live `ANTHROPIC_API_KEY` suite; `live-matrix.yml` conflicts with approved #1681) and tracked at **issue #1684** |
| ~~MINOR~~ | `src/lib/utils/toolChoice.ts:51` | Nameless `{type:"tool"}` (no `toolName`) | **Resolved** — `normalizeResolvedToolChoice` returns `undefined` (never the bare `"tool"`); covered by the suite |
| ~~MINOR~~ | `test/…-anthropic-toolchoice.ts` | Dead string assertion `seen.every(r => r.toolChoice !== "tool")` | **Resolved** — `observeRequests` now records the raw `tool_choice` value, so a regressed bare `"tool"` string would reach the assertion as a string and fail it |

**Checked and clean (fresh this recurrence)**
- **Fix sound.** String `"required"` now reaches the provider as `{type:"any"}` on the wire; object `{type:"tool",toolName}` unchanged; omission sends no `tool_choice` (release vs branch: 1 failed → 3 passed).
- **`normalizeResolvedToolChoice`** — graph-confirmed single caller (`getAISDKModel`); preserves behaviour for every shape it accepts (bare strings `"auto"/"none"/"required"`, `{type:"tool",toolName}`, native-loop `{type:"none"}` → `"none"`, malformed nameless tool → `undefined`, `undefined`/omission → no `tool_choice`).
- **Blast radius confined** (graph): `toolChoiceToAnthropic` callers (`getAISDKModel:1424`, `executeStreamInCaptureScope:2165`) and `resolveToolChoice` callers unchanged; change is inside a function body, no signature/public-type change → Rule 5 (backward compat) safe, docs/api drift 0.
- **No secrets** (inline scan clean; SentinelOne + CodeQL green); `transformParamsForLogging` untouched; no provider-registry/streaming hot-path regression.
- **Test (Rule 15)** — e2e via `dist/index.js`, single module graph, `assertDistFresh`, `skipUnlessProviderAvailable`; deterministic by asserting on captured request bytes rather than model tool-usage; observer restores `fetch` in `finally`.
- Single conventional commit; single-commit-policy compliant (bot-confirmed).

**Required before merge:** none blocking. The one tracked follow-up is the nightly CI wiring of the orphaned live suites (incl. this one) at **#1684**, to land after #1681 merges — this keeps the regression guarded by a real CI run rather than a withering script.

Tara-ag (approved):
## Verdict: APPROVE (head `0532e2289`)

The core fix is correct and verified at the wire; every prior MAJOR/MINOR finding resolves cleanly on this head. All review threads are resolved, the new suite is registered (`test:anthropic-toolchoice`), and the one tracked item is the nightly CI wiring of the orphaned live suites at #1684 (to land after #1681 merges).

- **Fix sound:** string `"required"` now reaches the provider as `{type:"any"}` on the wire; object `{type:"tool", toolName}` unchanged; omission sends no `tool_choice`.
- **`normalizeResolvedToolChoice`** (shared util): single graph-confirmed caller (`getAISDKModel`); preserves behaviour for every shape, incl. nameless `{type:"tool"}` → `undefined` (never the bare `"tool"`), covered by the suite.
- **Observer records the raw `tool_choice`** so the `seen.every(r => r.toolChoice !== "tool")` guard is live.
- **Backward compat (Rule 5):** change inside a function body; no signature/public-type change.
- **No secrets;** logging/`transformParamsForLogging` untouched; no registry/streaming hot-path regression.
- **Test (Rule 15):** e2e via `dist/index.js`, single module graph, `assertDistFresh`, `skipUnlessProviderAvailable`.
- Single conventional commit; single-commit-policy compliant.

Full detail in the authoritative `yama:summary` comment (#5649289323).

PR #1686, opened 2026-09-12
chore(ci): run the required checks on stacked pull requests

murdore:
## The problem

A pull request based on another feature branch — a *stacked* PR, opened so
dependent work can proceed before its parent merges — **cannot go green in this
repository today.** Not "tends to fail": cannot.

Every one of the four required status contexts is a job in a single file:

| required context | defined in |
| --- | --- |
| `test` | `.github/workflows/ci.yml` |
| `provider-safety-net` | `.github/workflows/ci.yml` |
| `security-suites` | `.github/workflows/ci.yml` |
| `build-check` | `.github/workflows/ci.yml` |

That file's trigger was `pull_request: branches: [release]`. A base-branch filter
there decides whether those checks are **created**, not whether they pass. So a
stacked PR had none of them.

And a check that was never created does not report as failing — it reports as
*missing*, which is indistinguishable from "still queued". This is the exact
trap `CLAUDE.md` already documents under *"A check that has not appeared has not
passed"*.

**Live evidence:** PR #1670 has sat at **1 of 5** required contexts since it was
opened. Not one of them is red. Four were never created.

## The second, independent failure

`single-commit-enforcement.yml` failed the same PRs a different way. Its
`pull_request` trigger was scoped to `[release, main]`, so a stacked PR fell
through to the **push-event** path — which hardcoded `origin/release` as the
base:

```bash
else
  BASE_BRANCH="origin/release"
```

Counting commits from `origin/release` on a branch stacked on another feature
branch includes the **parent's** commit. A single-commit PR was reported as
carrying two, and failed a policy it had not broken.

## The change

Both `pull_request` triggers are now unfiltered by base branch.

- The pull-request path in `single-commit-enforcement.yml` already resolved the
  real base from `github.base_ref`, so it needed no other change — it simply was
  never reached.
- The push path now asks for the branch's open PR and uses its base when there
  is one, falling back to `origin/release` when there is not, so a branch pushed
  before its PR exists behaves exactly as before.
- `push` remains scoped to `release` in `ci.yml`. Only pull-request handling
  changed.

## Verification

The before state is on the record: #1670, 1 of 5 contexts, four never created.

The after state is what this PR is for. Because `pull_request` runs resolve the
workflow from the merge of head into base, a PR **stacked on this branch** picks
up the new trigger and is the direct proof. I will open one against this branch
and post the context list here rather than assert the outcome — the whole point
of this change is that a missing check is not a passing one, and it would be
incoherent to claim success without showing the five contexts appear.

## Why this matters beyond #1670

Two items in the close-out backlog are sequenced purely because stacking does
not work: the stdio SIGKILL escalation that builds on #1683, and the issue-1684
CI wiring that builds on #1681. Neither has a technical reason to wait for a
merge; both were waiting on this.

## Risk

CI now runs on pull requests to any base branch, which costs runner minutes on
stacked PRs that previously got none. That is the intended trade: the
alternative is that stacked PRs remain unmergeable by construction.

## Summary by CodeRabbit

- **Improvements**
  - Pull request checks now run for pull requests targeting any branch, including stacked pull requests.
  - Validation now evaluates changes against the pull request’s actual base branch.
  - Push-based validation identifies the associated open pull request when available, providing more accurate checks across different branch workflows.

### DIFF
--- a/.github/workflows/single-commit-enforcement.yml
+++ b/.github/workflows/single-commit-enforcement.yml
@@ -3,8 +3,13 @@ name: Single Commit Per Branch Enforcement
 on:
   push:
     branches-ignore: [release, main, develop]
+  # Unfiltered by base branch, for the same reason as ci.yml: the
+  # pull_request path below is the only one that knows the PR's real base
+  # (github.base_ref). Filtered to [release, main], a stacked PR fell through
+  # to the push path, which assumed origin/release and therefore counted the
+  # PARENT branch's commit as one of this branch's own — reporting two commits
+  # for a single-commit PR and failing a policy it had not broken.
   pull_request:
-    branches: [release, main]
     types: [opened, synchronize, reopened]
 
 concurrency:
@@ -43,9 +48,13 @@ jobs:
           HEAD_REPO_ID: ${{ github.event.pull_request.head.repo.id }}
           BASE_REPO_ID: ${{ github.event.repository.id }}
           HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+          # Needed by the `gh pr list` base lookup on the push path below.
+          GH_TOKEN: ${{ github.token }}
         run: |
           # Determine the base branch and current branch
           if [ "$EVENT_NAME" = "pull_request" ]; then
+            # Authoritative: the event carries the PR's real base branch.
+            BASE_KNOWN=true
             BASE_BRANCH="origin/$BASE_REF"
             BRANCH_NAME="$HEAD_REF"
             
@@ -70,7 +79,56 @@ jobs:
               CURRENT_BRANCH="origin/$HEAD_REF"
             fi
           else
+            # A push event carries no base branch. Assuming release is right
+            # for the ordinary case and wrong for a stacked branch, where it
+            # counts the parent's commits as this branch's own. So the base is
+            # read from the branch's open pull request when exactly one can be
+            # identified, and treated as UNKNOWN otherwise — never guessed into
+            # a verdict.
+            BASE_KNOWN=false
             BASE_BRANCH="origin/release"
+
+            # Three outcomes have to stay distinguishable here, because two of
+            # them look identical if you only check whether the output is
+            # empty: the lookup succeeded and found one PR, it succeeded and
+            # found none, or it FAILED. Swallowing a failure with `|| true`
+            # turns an auth or API error into "no pull request", which then
+            # silently produces a policy verdict measured against the wrong
+            # base. So the exit status is captured separately from the output.
+            PR_LOOKUP_OK=true
+            PR_JSON=$(gh pr list --head "$REF_NAME" --state open \
+              --json headRefName,baseRefName 2>/dev/null) || PR_LOOKUP_OK=false
+
+            if [ "$PR_LOOKUP_OK" != "true" ]; then
+              echo "⚠️  Could not query pull requests for this branch (auth or"
+              echo "    API failure). Treating the base as unknown rather than"
+              echo "    assuming release, so no verdict is reached on a base"
+              echo "    that was never confirmed."
+            else
+              # `--head` filters by name and can match a PREFIX, and a branch
+              # may legitimately have several open pull requests that share it
+              # while targeting different bases. Either way `.[0]` would pick
+              # one arbitrarily, so require exactly one EXACT head match.
+              MATCHES=$(printf '%s' "$PR_JSON" \
+                | jq --arg ref "$REF_NAME" '[.[] | select(.headRefName == $ref)]')
+              MATCH_COUNT=$(printf '%s' "$MATCHES" | jq 'length')
+
+              if [ "$MATCH_COUNT" -eq 1 ]; then
+                BASE_KNOWN=true
+                BASE_BRANCH="origin/$(printf '%s' "$MATCHES" | jq -r '.[0].baseRefName')"
+                echo "🔗 Resolved base branch from the branch's one open PR: $BASE_BRANCH"
+              elif [ "$MATCH_COUNT" -eq 0 ]; then
+                echo "ℹ️  No open pull request for this branch yet, so the real"
+                echo "    base is unknown. Assuming release for reporting, but"
+                echo "    enforcement is deferred to the pull_request run."
+              else
+                echo "⚠️  $MATCH_COUNT open pull requests share this head branch"
+                echo "    while targeting different bases, so the intended base"
+                echo "    is ambiguous. Treating it as unknown; each pull"
+                echo "    request is still enforced against its own base by the"
+                echo "    pull_request run."
+              fi
+            fi
             CURRENT_BRANCH="HEAD"
             BRANCH_NAME="$REF_NAME"
           fi
@@ -86,6 +144,7 @@ jobs:
           COMMIT_COUNT=$(git rev-list --count $CURRENT_BRANCH ^$BASE_BRANCH 2>/dev/null || echo "0")
 
           echo "commits_ahead=$COMMIT_COUNT" >> $GITHUB_OUTPUT
+          echo "base_known=$BASE_KNOWN" >> $GITHUB_OUTPUT
           echo "branch_name=$BRANCH_NAME" >> $GITHUB_OUTPUT
           echo "base_branch=$BASE_BRANCH" >> $GITHUB_OUTPUT
 
@@ -112,6 +171,21 @@ jobs:
             echo "commit_hash=$COMMIT_HASH" >> $GITHUB_OUTPUT
             echo "commit_author=$COMMIT_AUTHOR" >> $GITHUB_OUTPUT
             
+          elif [ "$BASE_KNOWN" != "true" ]; then
+            # More than one commit ahead of a base we GUESSED. On a stacked
+            # branch that is the correct and expected state: the extra commits
+            # belong to the parent branch, not to this one. Failing here would
+            # report a policy violation the branch has not committed, which is
+            # exactly what this workflow used to do to every stacked PR on its
+            # first push — before its pull request existed to be read.
+            echo "⚠️  $COMMIT_COUNT commits ahead of the ASSUMED base"
+            echo "    ($BASE_BRANCH), and no open pull request was found to"
+            echo "    confirm the real one. Not treating this as a violation."
+            echo "    The pull_request run enforces the policy against the"
+            echo "    branch's actual base, so nothing merges unchecked."
+            echo "status=base_unknown" >> $GITHUB_OUTPUT
+            echo "📋 Commits ahead of the assumed base:"
+            git log --oneline $CURRENT_BRANCH ^$BASE_BRANCH
           else
             echo "❌ POLICY VIOLATION: Branch contains $COMMIT_COUNT commits"
             echo "status=violation" >> $GITHUB_OUTPUT
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,8 +3,17 @@ name: CI
 on:
   push:
     branches: [release]
+  # Deliberately UNFILTERED by base branch. Every one of the four required
+  # contexts this repository gates on — test, provider-safety-net,
+  # security-suites, build-check — is a job in THIS file, so a base filter
+  # here decides whether they exist at all, not merely whether they pass.
+  # Scoped to [release], a stacked pull request (one based on another feature
+  # branch so it can build on unmerged work) had NONE of them created. That
+  # does not read as a failure: a check that was never created is reported as
+  # missing, which looks like "still queued" forever, and the pull request can
+  # never go green no matter how good the change is. PR #1670 sat at 1 of 5
+  # required checks for exactly this reason.
   pull_request:
-    branches: [release]
 
 # Cancel a run that a newer push to the same ref has superseded. Without this
 # an amend-and-force-push — routine under the single-commit-per-branch policy —


.github/workflows/single-commit-enforcement.yml
```diff
@@ -70,7 +77,19 @@ jobs:
...
           else
+            # A push event carries no base branch, so this path has to guess.
+            # Assuming release is right for the ordinary case and wrong for a
+            # stacked branch, where it counts the parent's commits as this
+            # branch's own. Ask for the branch's open PR and use its real base
+            # when there is one; fall back to release when there is not (a
+            # branch pushed before its PR exists), which is the old behaviour.
             BASE_BRANCH="origin/release"
+            PR_BASE=$(gh pr list --head "$REF_NAME" --state open \
+              --json baseRefName --jq '.[0].baseRefName // empty' 2>/dev/null || true)
```
Tara-ag: **MAJOR — a `gh` failure silently reproduces the exact false violation this PR fixes.**

```bash
PR_BASE=$(gh pr list --head "$REF_NAME" --state open \
  --json baseRefName --jq '.[0].baseRefName // empty' 2>/dev/null || true)
```

`2>/dev/null || true` collapses two very different states into one: *"this branch has no open PR"* (empty result → fall back to `origin/release` is correct) and *"`gh` failed"* (unauthenticated, token without `pull-requests` read scope, rate-limited, transient API error → also empty result). The script cannot tell them apart.

For a **stacked** branch that *does* have an open PR, a `gh` failure falls back to `origin/release`, which counts the parent branch's commit as this branch's own and reports a single-commit PR as a two-commit violation — the precise bug this PR exists to remove, now occurring silently (the `🔗 Resolved base branch` echo never prints, and no warning appears). That contradicts the repo's own loudly-asserted philosophy in this very file ("a check that was never created is not a passing one") and the ci.yml work: silent-wrong over loud-wrong.

**Suggest:** separate a clean "no PR" from a `gh` error — e.g. capture the exit status and only fall back when `gh` succeeded and returned nothing:

```suggestion
            PR_LIST=$(gh pr list --head "$REF_NAME" --state open \
              --json baseRefName 2>/dev/null)
            if [ $? -eq 0 ]; then
              PR_BASE=$(printf '%s' "$PR_LIST" | \
                sed -n 's/.*"baseRefName":"\([^"]*\)".*/\1/p' | head -1)
            else
              echo "::warning::gh pr list failed on push; falling back to origin/release"
            fi
            if [ -n "$PR_BASE" ]; then
              BASE_BRANCH="origin/$PR_BASE"
              echo "🔗 Resolved base branch from open PR: $PR_BASE"
            fi
```

(If you keep `jq` for readability, at minimum drop the `|| true`, check `$?` explicitly, and emit a `::warning::` on the failure path so a future runner regression is visible instead of silently mis-merging.)

Tara-ag (commented):
Well-reasoned fix for stacked PRs — the base-branch filter on the `pull_request` triggers is exactly why the four required contexts were never created (#1670). The `ci.yml` change and the single-commit pull_request-path change are correct. One MAJOR robustness gap in the new push-path base lookup (see inline) that silently reintroduces the exact false violation this PR removes when `gh` fails.

Tara-ag:
**NEEDS_WORK** (one MAJOR robustness gap) — diagnostics correct and well-targeted, single fix requested.

## Verdict
The diagnosis is sound and the change is right in its main thrust: the `pull_request` base filters on `ci.yml` (`[release]`) and `single-commit-enforcement.yml` (`[release, main]`) meant stacked PRs never *created* the required checks (reported as missing, indistinguishable from queued — #1670's 1-of-5 state), and the single-commit push path hardcoded `origin/release`, counting the parent's commit. Unfiltering the `pull_request` triggers fixes both; the push path correctly asks for the branch's open PR and falls back only when none exists. The rationale is documented inline at the level this repo's CI deserves, and the self-verification plan (open a stacked PR against this branch) is an honest way to prove the "missing ≠ passing" claim.

## Findings

| --- | --- | --- |
| MAJOR | `.github/workflows/single-commit-enforcement.yml:87-88` | Push-path `gh pr list` swallows failure with `2>/dev/null \|\| true`, making a `gh` error (auth scope, rate-limit, API blip) indistinguishable from "no open PR" — silently falling back to `origin/release`, which on a stacked branch **reproduces the exact false two-commit violation this PR removes**. Distinguish a clean empty result from a `gh` failure and warn on the failure path. |

## Clean / deliberately not flagged
- `ci.yml` unfiltered `pull_request`: correct — the four required contexts (`test`, `provider-safety-net`, `security-suites`, `build-check`) all live in this file, so a base filter decides their *existence*, not just pass/fail.
- Removed `branches:` filter is the only `ci.yml` change; `push` stays `[release]`-scoped, so no behavior change for release merges.
- Single-commit `pull_request` path already resolved the real base from `github.base_ref`; it simply was never reached before.
- `GH_TOKEN: ${{ github.token }}` correctly added so `gh` authenticates on the push path (`pull-requests: write` permission covers reads).
- Runner cost on stacked PRs is the intended trade, acknowledged in the PR's Risk section.
- No injection concern: `gh pr list --head "$REF_NAME"` and `"origin/$PR_BASE"` are quoted; shell metachars from the resolved branch name are never re-parsed.

Scope is two YAML workflow files only — no `src/` code, no public-API surface, no runtime behavior — so graph-based blast-radius analysis found nothing beyond the workflows themselves.

.github/workflows/single-commit-enforcement.yml
Tara-ag: **MAJOR — `|| true` silently re-creates the exact bug this PR fixes.**

The push path relies entirely on this `gh pr list` lookup to find a stacked branch's real base. But `2>/dev/null || true` collapses two very different outcomes into one:

- **No open PR** → `PR_BASE` empty → fall back to `origin/release` (correct old behaviour).
- **`gh` failed** (workflow lacked `pull-requests: read` scope, a rate-limit, or an API error) → `PR_BASE` ALSO empty → silently falls back to `origin/release`.

On a stacked branch with an open PR, the second case reproduces the **false two-commit violation this PR exists to eliminate** — and does so silently, contrary to the file's otherwise fail-loud discipline (the `set -eo pipefail`-style intent of the surrounding job).

**Fix:** capture the exit status so a clean `[]` empty result is distinguishable from a `gh` failure, and emit a `::warning::` when the lookup errored so a fallback is never invisible.

```suggestion
            PR_BASE=$(gh pr list --head "$REF_NAME" --state open \
              --json baseRefName --jq '.[0].baseRefName // empty' 2>/dev/null)
            gh_exit=$?
            if [ -n "$PR_BASE" ]; then
              BASE_BRANCH="origin/$PR_BASE"
              echo "🔗 Resolved base branch from open PR: $PR_BASE"
            elif [ "$gh_exit" -ne 0 ]; then
              echo "::warning::gh pr list failed (exit $gh_exit); falling back to origin/release. Verify the 'pull-requests: read' permission on this workflow's token."
            fi
```

I removed the `|| true` so the exit code survives to `$?`, and added an `elif` branch that surfaces the failure instead of treating it as a normal "no PR open" case.

Tara-ag (changes requested):
One MAJOR robustness finding in the push path of the single-commit enforcement; please address before merge. Details are inline.

Tara-ag:
**NEEDS_WORK** — the core change is correct, well-documented, and low-risk (CI-only), but one robustness fix is requested in the single-commit push path before this lands.

| --- | --- | --- |
| MAJOR | `.github/workflows/single-commit-enforcement.yml:87-88` | `2>/dev/null \|\| true` collapses a real `gh` failure into a silent fallback to `origin/release`, re-creating the false two-commit violation this PR exists to remove. Surface it with `::warning::` instead. |

**Reviewed and found clean:**
- `ci.yml` — unfiltering the `pull_request` base gives stacked PRs their required checks (all four gate jobs live in this file). The rationale comment is accurate and matches the repo's diagnose. The prior `branches: [release]` filter is correctly removed.
- `single-commit-enforcement.yml` — unfiltering `pull_request` plus the `GH_TOKEN` wiring is correct; the push-path `gh pr list` lookup is the right approach for finding a stacked branch's real base.
- No runtime code, no public SDK/API surface touched; no secrets, no test-impact.

**Checks performed:** full diff, both workflow files, prior review threads (none existed), and callers/blast radius — impact is confined to CI job gating, so no code-graph blast radius applies.

Tara-ag:
**APPROVE** — the stacked-PR CI fix is correct and well-documented. The one MAJOR flagged by prior reviews is recalculated to a minor nit, because the new `BASE_KNOWN` fallback branch already neutralizes its core harm.

## Supersedes
Two `<!-- yama:summary -->` comments from a prior run (both `NEEDS_WORK`, same single MAJOR about `2>/dev/null || true` on the push-path `gh pr list`). I can't edit/remove those, so this is the consolidated, current verdict. The prior MAJOR overstated its impact — see the recalibration below.

## Verdict
The diagnosis is sound and the fix is right: unfiltering the `pull_request` triggers on `ci.yml` (`[release]`) and `single-commit-enforcement.yml` (`[release, main]`) is what makes the four required contexts (`test`, `provider-safety-net`, `security-suites`, `build-check`) actually get *created* for a stacked PR (a never-created check reads as "still queued", never red — #1670's 1-of-5). The push path now probes the branch's open PR for the real base and only falls back to `origin/release` when none exists. The rationale is documented at the level this repo's CI deserves, and opening a stacked PR *against this branch* as the proof-of-success is honest and directly on-point. No runtime/SDK code, no public API surface, least-privilege `permissions: contents: read` untouched, no secrets.

## Findings

| --- | --- | --- | --- |
| MINOR | `.github/workflows/single-commit-enforcement.yml:81-86` | `2>/dev/null \|\| true` on the push-path `gh pr list` still makes a `gh` failure indistinguishable from "no open PR". **Impact recalculated down from the prior MAJOR:** the new `elif [ "$BASE_KNOWN" != "true" ]` branch converts a gh-failure on a stacked branch into an honest `base_unknown` deferral rather than the false two-commit violation the prior review claimed. A `gh` failure can no longer *block* a compliant branch. The only residue is a lost diagnostic signal (you can't tell a transient `gh` error from "no PR yet"), so keep it a nit. Optional one-line hardening below. | Recalculated (was MAJOR) |

```bash
# Optional hardening — surface a gh failure instead of folding it into "no PR":
PR_BASE=$(gh pr list --head "$REF_NAME" --state open \
  --json baseRefName --jq '.[0].baseRefName // empty' 2>/dev/null)
if [ "$?" -ne 0 ] && [ -z "$PR_BASE" ]; then
  echo "::warning::gh pr list failed on push; enforcement deferred to the pull_request run"
fi
```

## Clean / deliberately not re-flagged (already covered inline)
- `|| true` graceful-unknown handling lives in the `elif BASE_KNOWN != true` branch — **correct**, and it is why a gh-failure is no longer a false violation.
- Multiple-open-PRs ambiguity (`.[0].baseRefName`) — already raised by CodeRabbit; not a realistic stacked-branch failure and it was not re-posted here.
- `GH_TOKEN: ${{ github.token }}` wiring is correct; `permissions: pull-requests: write` includes read, so auth scope isn't the concern — only transient API blips.
- No new inline comment was posted for the `|| true` nit because it is already on `:91` three times (2× yama, 1× CodeRabbit); a fourth duplicate would bury the findings that matter.

## Review hygiene note (for the platform)
Two `<!-- yama:summary -->` comments and two yama inline MAJOR comments (different markers, same finding, `ollama:gh-fallback-silent` resolved + `ollama:gh-pr-list-errors-silently-collapsed` open) were posted for this one finding. These should be deduplicated (resolve the duplicate yama thread; keep one summary). I'm unable to edit or delete them from this session, so flagging it here rather than stacking a third full report.

## Checks performed
- Full diff of both workflow files; full current file contents via GitHub API (permissions blocks, the run: step, and the `elif` ordering all verified).
- Existing comments/threads: no author reply anywhere, code unchanged since the single commit (`2ff3dbc`), so no finding was justified away.
- Blast radius: change touches only two top-level `.github/workflows/*.yml` triggers — no `src/`, no public-API surface, no runtime code. Code-graph impact analysis found nothing beyond CI gating itself; no callers, no affected flows.

Tara-ag (approved):
<?xml... ">Approve

Tara-ag:
**APPROVE** — the change is correct, well-documented, and the earlier MAJOR is mitigated by the `base_unknown` deferral; the residual is a MINOR hardening note.

## Verdict
This is a CI-only chore (two workflow files, no `src/`) that fixes stacked PRs getting none of the required checks. The base-branch filters on the `pull_request` triggers decided whether the four required contexts were *created*, so a stacked PR (e.g. #1670) reported them as "missing", indistinguishable from "still queued". Unfiltering both triggers fixes that; the `ci.yml` `push` path stays `[release]`-scoped and the single-commit push path now resolves the branch's real base from its open PR, falling back to `origin/release` only when none exists. The `base_known` / `base_unknown` plumbing makes enforcement on a guessed base honest.

**Recalibration of the earlier MAJOR (now MINOR):** A prior review flagged `2>/dev/null || true` on the `gh pr list` call as re-creating a false two-commit violation when `gh` fails. That harm is **gone** in the current code: on failure `PR_BASE` is empty, `BASE_KNOWN` stays `false`, and the new `elif [ "$BASE_KNOWN" != "true" ]` branch emits `status=base_unknown` instead of a violation. A `gh` failure can no longer block a compliant branch. The only residue is a lost diagnostic signal — recommend surfacing it with `::warning::`, but this is not blocking.

## Findings

| --- | --- | --- |
| MINOR | `.github/workflows/single-commit-enforcement.yml:84-85` | When `gh pr list` fails (auth/rate-limit/API blip) it is silently treated as "no open PR". A `::warning::gh pr list failed` branch would make a future runner regression visible instead of silently deferring to `origin/release`. |

## Checked and clean
- `ci.yml` — unfiltering the `pull_request` base gives stacked PRs their four required contexts (`test`, `provider-safety-net`, `security-suites`, `build-check`); `push` stays `[release]`-scoped. Rationale comment is accurate.
- Single-commit `pull_request` path already used `github.base_ref` (authoritative); it simply was never reached before.
- `GH_TOKEN: ${{ github.token }}` added correctly; `pull-requests: write` covers the read.
- No injection risk: `gh pr list --head "$REF_NAME"` and `origin/$PR_BASE` are quoted.
- No runtime/SDK/API surface, no secrets.
- Blast radius (code graph): impact confined to CI job gating — nothing beyond the workflows.

## Note on review hygiene (upstream of this session)
Two earlier `<!-- yama:summary -->` comments on this PR (both **NEEDS_WORK** for the same finding, now downgraded to MINOR) are stale duplicates. They could not be deleted or edited from this session's toolset; manual cleanup is recommended so this PR carries a single summary.

murdore:
## Verified. A stacked pull request is now fully green.

The before/after, both measured rather than argued:

| | base | required contexts present | all green? | mergeable |
| **#1670** — old workflows | `feat/generate-path-loop-guard` | **1 of 5** | no — four were never created | no |
| **#1687** — this branch's workflows | `chore/ci-run-checks-on-stacked-prs` | **5 of 5** | **yes** | `CLEAN` |

#1687's check list:

```
build-check                          SUCCESS
provider-safety-net                  SUCCESS
security-suites                      SUCCESS
test                                 SUCCESS
🔒 Single Commit Policy Validation   SUCCESS
present=5 ok=5 merge=CLEAN
```

Contexts deduplicated by name, since the single-commit check reports twice.

## The first attempt caught a gap in this fix, and it is fixed here

Worth recording because it is the kind of thing this repository keeps getting
wrong, and it would have shipped silently.

On the first run, #1687's **pull_request** check passed and its **push** check
failed. Both report under the same required context name. The cause: a branch is
normally pushed *before* its pull request exists, so `gh pr list --head` returned
nothing and the push path fell back to `origin/release` — counting this branch's
parent commit as its own, exactly the bug this PR set out to remove, just moved
one step later.

Relying on GitHub preferring the newer check run would have been luck, not
correctness, and the red X was visible on the pull request either way.

So the push path no longer guesses. When it cannot identify the real base it
reports the count and **defers**, emitting `status=base_unknown` instead of
`violation`:

```
⚠️  2 commits ahead of the ASSUMED base (origin/release), and no open pull
    request was found to confirm the real one. Not treating this as a violation.
    The pull_request run enforces the policy against the branch's actual base,
    so nothing merges unchecked.
```

Enforcement is not weakened: every pull request is still gated by the
pull_request run, which reads the true base from `github.base_ref`. What is
removed is a verdict the push event was never in a position to reach.

## What this unblocks

- **#1670** stops being structurally unmergeable.
- The stdio SIGKILL escalation that builds on #1683, and the issue-1684 CI wiring
  that builds on #1681, can be stacked instead of waiting on merges. Neither had
  a technical reason to wait.

Merge this before #1687 — it retargets to `release` automatically at that point.

murdore:
Both CodeRabbit findings on the push path are real, and both are fixed in
`ee123db80`. They are worth spelling out because they are the same class of bug
this PR exists to remove — a verdict reached against a base nobody confirmed —
and I reintroduced it while fixing it.

## `|| true` swallowed lookup failures (Major)

```bash
PR_BASE=$(gh pr list --head "$REF_NAME" ... || true)
if [ -n "$PR_BASE" ]; then ...
```

Three outcomes collapse into two there. "The lookup succeeded and found no pull
request" and "the lookup FAILED" both produce an empty string, so an auth or API
error silently became *no PR* → base assumed to be `release` → a commit count
measured against a base that was never confirmed.

The exit status is now captured separately from the output:

```bash
PR_LOOKUP_OK=true
PR_JSON=$(gh pr list --head "$REF_NAME" --state open \
  --json headRefName,baseRefName 2>/dev/null) || PR_LOOKUP_OK=false
```

A failed lookup is reported and the base is treated as **unknown**, so the step
reaches no verdict at all.

On your suggestion to *exit the step* on failure: I went with deferral instead,
and the reason is specific to this workflow. Failing makes the push-event run red,
and both events report under the same required context name
(`🔒 Single Commit Policy Validation`) — so a transient API blip would paint a red
X on a pull request whose policy is fine, which is the exact confusion this PR is
removing. Deferral loses nothing: the `pull_request` run enforces against
`github.base_ref`, which is authoritative, so nothing can merge unchecked.

## `--head` does not identify one base (Minor)

Correct, and there are two independent ways it fails: a branch may legitimately
have several open pull requests sharing its head while targeting different bases,
and `--head` filters by name in a way that can match a **prefix** — so
`--head foo` can return `foo-bar`. `.[0].baseRefName` picked one arbitrarily in
either case.

Now it requires exactly one **exact** head match:

```bash
MATCHES=$(printf '%s' "$PR_JSON" \
  | jq --arg ref "$REF_NAME" '[.[] | select(.headRefName == $ref)]')
MATCH_COUNT=$(printf '%s' "$MATCHES" | jq 'length')
```

`1` resolves the base; `0` defers with "no pull request yet"; anything more defers
with the count and says the intended base is ambiguous. Each of those pull
requests is still enforced against its own base by its own `pull_request` run.

So all four paths — resolved, none, ambiguous, lookup failed — now either produce a
confirmed base or produce no verdict. There is no remaining path that guesses one.

## Verification

#1687 is stacked on this branch and is the live exercise of both paths: its
`pull_request` run enforces against `chore/ci-run-checks-on-stacked-prs`, and its
push run now resolves that same base through the hardened lookup rather than
assuming `release`. I am rebasing it onto `ee123db80` so the proof runs against
this version rather than the one it was opened against.

Tara-ag:
**APPROVE** — the stacked-PR CI fix is correct; every prior finding (both yama MAJORs and both CodeRabbit items) is resolved in commit `ee123db8` with adequate author justification, and I find no new issues in the current code.

## Verdict
A CI-only chore (two workflow files, no `src/`) that fixes stacked PRs being unmergeable-by-construction. The base-branch filters on the `pull_request` triggers decided whether the four required contexts were *created* at all (a never-created check reads as "still queued", never red — #1670's 1-of-5). Unfiltering both `pull_request` triggers fixes that; `ci.yml` `push` stays `[release]`-scoped, and the single-commit push path now resolves the branch's real base from its open PR and treats the base as *unknown* (deferring to the authoritative `pull_request` run) rather than guessing `origin/release` into a verdict.

## Prior findings — all resolved, no repost
| Finding | Where it was raised | Resolution in `ee123db8` |
| --- | --- | --- |
| MAJOR — `\|\| true` swallows `gh pr list` failure → silent fallback to `origin/release` → false two-commit violation | yama inline (x2), CodeRabbit Major | `PR_LOOKUP_OK=true; PR_JSON=$(gh ... ) \|\| PR_LOOKUP_OK=false` — exit status captured separately from output (single-commit-enforcement.yml:95,100). On failure the base is treated as unknown and `status=base_unknown` is emitted (never a guessed verdict). |
| Minor/ambiguous — `.[0].baseRefName` picks arbitrarily when multiple open PRs share the head | CodeRabbit | Exact-head-match filter `select(.headRefName == $ref)` requires exactly one match; 0 defers, >1 defers as ambiguous (lines 106-130). |

The author's deferral-vs-exit choice is sound and specifically justified (single-commit.yml:109-114, 173-188): `push` and `pull_request` report under the *same* required context (`🔒 Single Commit Policy Validation`), so exiting on a transient lookup failure would paint a red X on a compliant PR — the exact confusion this PR removes. Deferral loses nothing since the `pull_request` run enforces against `github.base_ref`, which is authoritative.

## Checked and clean
- `ci.yml` — purely the `pull_request` base-filter removal plus an accurate rationale comment; `push` filter untouched.
- Single-commit `pull_request` path — already used `github.base_ref`; it simply was never reached before.
- `GH_TOKEN: ${{ github.token }}` wiring correct; `permissions: pull-requests: write` covers the read.
- No injection risk: `gh pr list --head "$REF_NAME"`, `origin/$PR_BASE`, and the jq `--arg` are all quoted / parameterized; no env value is interpolated into the script body.
- `jq`/`gh` are both preinstalled on `ubuntu-latest`, so no dependency gap.
- Blast radius (code graph): impact confined to CI job gating — no runtime/SDK/API surface, no public-API change (rule 5 unaffected), no secrets, no test-impact.

## Review hygiene note
This PR currently carries four stale `<!-- yama:summary -->` comments (two NEEDS_WORK, two APPROVE) from earlier sessions. They could not be edited/deleted from this session's toolset; this is the consolidated, current verdict. Recommend manual cleanup so the PR carries a single summary.

**Checks performed:** full diff of both workflow files; current file contents at head (`ee123db8`) verified line-by-line; all prior review threads, reviews, and author replies read; commit-policy bot and both CodeRabbit threads reconciled.

Tara-ag (approved):
**APPROVE** (current head `ee123db`, consolidating the earlier verdict).

The open items from the prior reviews are all addressed here:

- **`|| true` swallowing `gh pr list` failures (was MAJOR)** — resolved (`PR_LOOKUP_OK` now captures the exit status separately from the output; a failed lookup is reported and the base is treated as `unknown`, so the step reaches no verdict on an uncon�firmed base).
- **Ambiguous `.[0].baseRefName` (CodeRabbit)** — resolved (requires exactly one *exact* head match via `jq --arg ref`; 0 → defer, >1 → defer as ambiguous).
- The new `elif [ "$BASE_KNOWN" != "true" ]` branch defers an honest `status=base_unknown` on a guessed base instead of emitting a false violation — no path remains that guesses a base into a verdict.

Our shared concern (a verdict reached against a base nobody confirmed) is eliminated; the `pull_request` run still enforces against the authoritative `github.base_ref`, so enforcement is not weakened. CI-only change, no `src/`, no public API surface, no secrets. Merge this before #1687 so it retargets to `release`.

murdore:
Re-verified against the hardened version, as promised above.

#1687 is rebased onto `ee123db80` — its branch carries the new lookup (the
`PR_LOOKUP_OK` / exact-match / ambiguity handling, confirmed present in the parent
commit rather than assumed) — and it comes back:

```
1686: 5/5 ok=5 APPROVED CLEAN
1687: 5/5 ok=5           CLEAN
```

So the stacked pull request still gets all five required contexts, and still goes
green, with the base now resolved through a lookup that refuses to guess. The two
CodeRabbit findings are fixed without costing the behaviour this PR exists to
deliver — which was the thing worth checking, since a stricter lookup could
easily have turned "resolved" into "deferred" for the ordinary case and quietly
put stacked PRs back where they started.

Both threads resolved. Ready when you are; merge this before #1687, which
retargets to `release` automatically at that point.

PR #1688, opened 2026-09-12
feat(proxy): auto-configure Grok Build with catalog windows and backends

murdore:
## Summary

Adds Grok Build as a NeuroLink proxy client, the same way Claude Code, Codex, Gemini CLI, OpenCode, Qwen, and Copilot are auto-configured on `proxy start` / `proxy stop`.

Grok Build is not remapped in place. Built-in `grok-4.6` / `grok-4.5` stay on xAI. The writer adds the proxy catalog as extra picker entries:

- Claude models → Anthropic Messages door (`/v1/messages`, passthrough)
- Gemini / GPT models → OpenAI Chat Completions door (`/v1/chat/completions`, translation)
- `context_window` is Grok's compaction limit and is set from NeuroLink's upstream tables (Sonnet/Opus 4.6 = 1M, Haiku = 200k, Gemini 2.5 Pro = 1,048,576)
- Haiku / 3.5 / Gemini set `supports_reasoning_effort = false` because Grok's global `xhigh` becomes Anthropic `thinking.type = "adaptive"`, which Haiku 4.5 rejects with 400
- `grok-shell/` User-Agents are attributed as `clientApp: grok`

Does not change the running launchd proxy.

## Test plan

- [x] Isolated HOME apply/restore round-trip (does not write `~/.grok` on the operator machine)
- [x] Detects only when `~/.grok` exists
- [x] Restore refuses without a snapshot and refuses a foreign proxy URL
- [x] Catalog classification: Claude messages + windows, Gemini chat_completions, no `grok-4.6` remap
- [x] `pnpm run check` / `validate:all` via pre-commit
- [x] `pnpm run pre-push` (build, provider-structure, model-manifests)
- [ ] After merge: install the published package, then `neurolink proxy start` on a dedicated machine — do not point the live 55669 proxy at this branch until that is an explicit follow-up

## Summary by CodeRabbit

* **New Features**
  * Added automatic Grok Build configuration when starting the proxy.
  * Proxy catalog models now appear in Grok’s model picker, supporting Claude, Gemini, and OpenAI-compatible traffic.
  * Added support for custom proxy configuration paths and appropriate routing, context windows, and reasoning settings.

* **Bug Fixes**
  * Preserves existing Grok settings during configuration.
  * Safely restores previous settings when the proxy stops without altering unrecognized configuration blocks.
  * Added recognition for Grok shell sessions in proxy attribution.


src/cli/proxy-clients/grok.ts
```diff
@@ -0,0 +1,404 @@
...
+    maxCompletionTokens: contextWindow >= 1_000_000 ? 16_384 : 8_192,
+    supportsReasoningEffort: supportsAdaptiveThinking(id),
+  };
+}
+
+function loadRoutedModelIds(): string[] {
+  const configPath = join(homedir(), ".neurolink", "proxy-config.yaml");
+  try {
+    const raw = readFileSync(configPath, "utf8");
+    const parsed: unknown = JSON.parse(raw);
```
Tara-ag: **MAJOR** — `loadRoutedModelIds()` parses `~/.neurolink/proxy-config.yaml` with `JSON.parse`, but that file is YAML (the `.yaml` extension, and the whole repo reads it as YAML).

`JSON.parse` throws on any real YAML — `#` comments, unquoted strings, the indented `model-mappings:` block that the docs and `proxyConfig.ts` use. The throw is swallowed by `catch` and the function returns `[]`, so the feature this PR advertises — "adds any `routing.model-mappings` in `~/.neurolink/proxy-config.yaml`" (its own comment added in `src/lib/constants/proxyModels.ts` and the PR body) — **silently never works for YAML users**, who are the canonical case. The default proxy config path is literally `~/.neurolink/proxy-config.yaml` (`src/cli/commands/proxy.ts`), and `docs/features/claude-proxy.md` states "The file supports YAML or JSON format."

The rest of the codebase is consistent with YAML: `src/lib/proxy/proxyConfig.ts` parses this file with js-yaml (`parseYaml` → `yaml.load`), with `JSON.parse` only as a fallback. That loader also resolves `${ENV_VAR}` placeholders.

Fix: reuse the existing loader rather than hand-rolling `readFileSync`+`JSON.parse`. Reusing `parseProxyConfigString`/`loadProxyConfig` also keeps env-var interpolation correct:

```suggestion
import { readFileSync } from "fs";
import { homedir } from "os";
import { join } from "path";
import { logger } from "../../lib/utils/logger.js";
import { getContextWindowSize } from "../../lib/constants/contextWindows.js";
import { DEFAULT_PROXY_MODEL_IDS } from "../../lib/constants/proxyModels.js";
import { parseProxyConfigString } from "../../lib/proxy/proxyConfig.js";
import type {
  CliGrokProxyModelSpec,
  CliGrokSnapshot,
  CliProxyClientConfigurator,
} from "../../lib/types/index.js";
import {
  isUsableSnapshot,
  shouldCaptureSnapshot,
  writeFileAtomic,
} from "./snapshot.js";
```

and in the function body:

```suggestion
function loadRoutedModelIds(): string[] {
  const configPath = join(homedir(), ".neurolink", "proxy-config.yaml");
  try {
    const raw = readFileSync(configPath, "utf8");
    const parsed = await parseProxyConfigString(raw);
    const ids: string[] = [];
    for (const m of parsed.routing?.modelMappings ?? []) {
      if (m.from.trim().length > 0) {
        ids.push(m.from.trim());
      }
    }
    return ids;
  } catch {
    return [];
  }
}
```

(If the module must stay synchronous, mirror `proxyConfig.ts` with an async js-yaml load instead of `JSON.parse`.) The `JSON.parse`-on-YAML path is currently untested — `loadRoutedModelIds` is not exposed via `__grokTestHooks` and none of the new tests write a YAML `proxy-config.yaml`.

Tara-ag (commented):
Reviewed against repo conventions (CLAUDE.md / repo-conventions skill). See inline finding. Verdict in summary comment.

Tara-ag:
## Review: PR #1688 — feat(proxy): auto-configure Grok Build with catalog windows and backends

**Verdict: NEEDS_WORK** — solid, well-tested wiring of the new Grok client into the registry, but the one headline enhancement (adding `routing.model-mappings` from `proxy-config.yaml` to the Grok catalog) silently doesn't work because the file is parsed with `JSON.parse` instead of the YAML loader the rest of the codebase uses.

| MAJOR | `src/cli/proxy-clients/grok.ts:119` | `loadRoutedModelIds()` `JSON.parse`s `~/.neurolink/proxy-config.yaml`, which is YAML (`.yaml` extension, default proxy config path, `proxyConfig.ts` parses it with js-yaml). Any real YAML throws → caught → `[]`, so the routed-model augmentation never applies. `loadRoutedModelIds` isn't in `__grokTestHooks`, so no test covers it. |

### Checked and found clean
- **Registry wiring** (rule: no hand-rolled roster): `PROXY_CLIENT_CONFIGURATORS` in `src/cli/proxy-clients/registry.ts` is consumed only by `src/cli/commands/proxy.ts` and the test suite — adding `grokConfigurator` is picked up automatically. The two roster tests (`testProxyClientRoster`, `testApplyAllReportsPerClient`) were updated correctly.
- **`DEFAULT_PROXY_MODEL_IDS` consumers**: only `proxyTranslationEngine.ts`, `copilot.ts`, `openCode.ts`, and the new `grok.ts` — no drift. The doc-comment "three consumers" update is accurate.
- **Context-window resolution**: traced `classifyGrokProxyModel` → `getContextWindowSize` for every default catalog id. `claude-sonnet-4-6`/`claude-opus-4-6` → 1M, `claude-haiku-4-5` → 200k (via `_default`; no exact/prefix key), `claude-3-5-*`/`claude-sonnet-4-20250514` → 200k, `gpt-4o` → 128k, `gemini-2.5-pro`/`gemini-2.5-flash` → 1,048,576. Matches the new tests.
- **Snapshot/restore safety** (`snapshot.ts`): consistent with the Codex/OpenCode pattern — snapshot lives in `~/.neurolink/`, `shouldCaptureSnapshot`/`isUsableSnapshot`/`writeFileAtomic` reused; restore refuses without a snapshot and refuses a foreign base URL. Mode handling (0600 / carry-over) is preserved via `writeFileAtomic`. Tests cover both refusal paths.
- **No `grok-*` remap**: `catalogModelIds` skips `grok-` ids, and the test asserts no `[model.grok-4.6]` entry. `clientAttribution` "grok-shell/" prefix added with a test.
- **Rule 4 (CLI ≠ SDK)**: new code lives only under `src/cli/proxy-clients/`; no SDK leak.
- **Rule 15**: tests in `continuous-test-suite-proxy.ts` use the existing `src/` single-module-graph pattern with the declared determinism exception on `__grokTestHooks`.
- **Security**: no secrets introduced; `PLACEHOLDER_KEY` is the inert literal `neurolink-proxy`, and no provider params are logged.
- **Commit hygiene**: single conventional commit `feat(proxy): ...`, `${GITHUB_ACTIONS}` square.

### Not applied
- The after-merge validation (install package → `proxy start` on a dedicated machine) is correctly listed as a follow-up in the PR body; off-merge-path testing does not block this review.

Tara-ag (changes requested):

src/cli/proxy-clients/grok.ts
Tara-ag: **MAJOR** — `~/.neurolink/proxy-config.yaml` is parsed with `JSON.parse`, but that file is YAML, so the new `routing.model-mappings` enhancement never resolves for the default config.

`loadRoutedModelIds()` reads the exact default proxy config path (`join(homedir(), ".neurolink", "proxy-config.yaml")`) and feeds it to `JSON.parse`. That file is documented and written as YAML ("The file supports YAML or JSON format" — `docs/features/claude-proxy.md`), and the canonical loader `loadProxyConfig` / `parseProxyConfigString` in `src/lib/proxy/proxyConfig.ts` parses `.yaml` files with **js-yaml** (`parseYaml` → `yaml.load`). Real YAML — comments, unquoted values, the indented `model-mappings:` array, `${VAR}` interpolation — all throw inside `JSON.parse`, and the `catch` silently returns `[]`. So the feature this PR advertises (proxyModels.ts doc: "plus any `routing.model-mappings` in `~/.neurolink/proxy-config.yaml`") silently does nothing for the format the file is actually in, and only coincidentally works if the user hand-writes strict JSON in a `.yaml` file. It's not in `__grokTestHooks`, so no test catches it.

Reuse the canonical loader instead of hand-rolling a JSON-only parser (rule 6 discipline: `loadProxyConfig` also does validation + env-var resolution, and a JSON-only path would diverge from the runtime the proxy actually uses):

```suggestion
async function loadRoutedModelIds(): Promise<string[]> {
  const configPath = join(homedir(), ".neurolink", "proxy-config.yaml");
  try {
    const config = await parseProxyConfigString(
      readFileSync(configPath, "utf8"),
    );
    const mappingIds = (config.routing?.modelMappings ?? [])
      .map((m) => m.from.trim())
      .filter((id) => id.length > 0);
    return mappingIds;
```

Tara-ag (changes requested):
Reviewed against repo conventions (CLAUDE.md / repo-conventions skill). One MAJOR finding — see inline comment. Verdict in summary comment.

Tara-ag:
## Verdict: NEEDS_WORK

Solid, careful change overall — but the new `routing.model-mappings` enhancement is broken on the default config path, so it needs one fix before landing. Review state set to changes-requested to match.

### Findings

| --- | --- | --- |
| MAJOR | `src/cli/proxy-clients/grok.ts:119` | `loadRoutedModelIds()` parses `~/.neurolink/proxy-config.yaml` (a YAML file) with `JSON.parse` — throws on real YAML and silently returns `[]`. The advertised mapping feature never works. Reuse `parseProxyConfigString` (js-yaml). |

A suggestion block with the fix is on the inline comment at `grok.ts:119`.

- **Registry wiring** — `grokConfigurator` added to `PROXY_CLIENT_CONFIGURATORS` in apply order; `registry.ts` consumed only by `proxy.ts` apply-all path and the roster test. No circular import (rule 1 clean; configurators are static imports like the existing ones).
- **No `grok-*` remap** — `catalogModelIds()` skips built-in `grok-` ids, preserving `grok-4.6`/`grok-4.5` on xAI; covered by `testGrokConfiguratorRoundTrip`.
- **Snapshot/restore safety** — reuses `snapshot.ts` (`isUsableSnapshot`, `shouldCaptureSnapshot`, `writeFileAtomic`); restore refuses without a snapshot and refuses foreign base URLs; tested (`testGrokRestoreRefusesWithoutSnapshot`, `testGrokRestoreRefusesForeignUrl`).
- **Context windows & backends** — `classifyGrokProxyModel` resolves correct windows/backends (`messages` for Claude, `chat_completions` for Gemini/OpenAI); adaptive-thinking guard for Haiku is sensible; tested (`testGrokCatalogWindowsAndBackends`).
- **CLI/SDK separation (rule 4)** — writer lives under `src/cli/`; no SDK leaks.
- **Security** — placeholder key `neurolink-proxy`, no real secrets; no injection/log-leak vectors found.
- **Conventions** — single conventional commit (`feat(proxy): ...`), squashed; rule-15 e2e test pattern followed.

Impact analysis used the code graph: `DEFAULT_PROXY_MODEL_IDS` has four consumers (proxy translation engine, OpenCode writer, roster test, new Grok writer) all expected to agree; blips radius of the new file is `registry.ts` + the `test/continuous-test-suite-proxy.ts` roster, no SDK surface change (rule 5 clean).

src/cli/proxy-clients/grok.ts
murdore: Fixed in 92d30012.

`loadRoutedModelIds()` now goes through `parseProxyConfigString` (js-yaml + env interpolation + the same `modelMappings` extraction `proxyConfig.ts` uses). A YAML `routing.model-mappings` fixture that `JSON.parse` rejects is covered via `__grokTestHooks` (`testGrokYamlRoutedMappings`). `docs/api` was regenerated so the validate shard matches `src/`.

src/cli/proxy-clients/grok.ts
```diff
@@ -0,0 +1,388 @@
...
+    name: displayNameFor(id),
+    apiBackend,
+    contextWindow,
+    maxCompletionTokens: contextWindow >= 1_000_000 ? 16_384 : 8_192,
+    supportsReasoningEffort: supportsAdaptiveThinking(id),
+  };
+}
+
+async function loadRoutedModelIds(): Promise<string[]> {
+  const configPath = join(homedir(), ".neurolink", "proxy-config.yaml");
```
Tara-ag: **MINOR** — `loadRoutedModelIds()` reads a hardcoded `~/.neurolink/proxy-config.yaml`, but the proxy can be run/installed with a **custom `--config` file** (`proxy install`/`proxy start --config <path>`, `src/cli/commands/proxy.ts` resolves `explicitConfig ? resolve(explicitConfig) : join(homedir(), ".neurolink", "proxy-config.yaml")`). When an operator points the proxy at a non-default config, the model-mappings that actually govern routing live in that file, yet this writer will silently read the default path — so the catalog's "routed" additions (`testGrokYamlRoutedMappings` asserts these are picked up) reflect the wrong or non-existent file.

Two compounding points:
- The generic loader now used here — `parseProxyConfigString` — runs **full** validation on the file (`validateProxyConfig`, `validateCloakingConfig`, and `failOnUnresolvedAccountCredentials`). Any validation issue or unresolved `${ACCOUNT_API_KEY}` placeholder throws → the `catch` returns `[]`, silently dropping the mapping feature. (It no longer throws on honest YAML after the 92d3001 fix, but it's still coupled to every other section of the config parsing cleanly.)
- The path itself is not shared with the boot-time loader, so the writer and the proxy can disagree about which file is active.

Consider resolving the same config path the proxy uses (or threading the active config path / base URL into the configurator's `apply`) so the writer and the routing runtime can't drift. If a shared path is out of scope, at minimum note this divergence on the documented feature (the `proxyModels.ts` comment claims "plus any `routing.model-mappings` in `~/.neurolink/proxy-config.yaml`", which is only true for the default path).

Tara-ag (commented):
Reviewed the updated commit (92d3001). The prior MAJOR finding (JSON.parse on the YAML `proxy-config.yaml`) is genuinely fixed and tested. One minor new observation on the `loadRoutedModelIds` path. See inline comment; verdict in summary.

Tara-ag:
## Review: PR #1688 — feat(proxy): auto-configure Grok Build with catalog windows and backends

**Verdict: APPROVE** — the prior review's one MAJOR finding is genuinely fixed and tested in 92d3001; I re-verified the fix and found only one minor robustness note this round.

### Resolution of prior finding
The previous MAJOR (`loadRoutedModelIds` used `JSON.parse` on the YAML `~/.neurolink/proxy-config.yaml`) is **resolved**:
- `loadRoutedModelIds` now goes through `parseProxyConfigString` (js-yaml + env interpolation, same loader as `proxyConfig.ts`).
- It is exposed via `__grokTestHooks` and `testGrokYamlRoutedMappings` writes a YAML fixture (comments + `${GROK_TEST_ROUTED_FROM}` interpolation) that `JSON.parse` rejects and asserts the routed id lands in the catalog — proving the fix. The author's reply in `#discussion_r3995481261` is accurate; accepted, not re-posted.

### Findings this round

| MINOR | `src/cli/proxy-clients/grok.ts:117` | `loadRoutedModelIds()` hardcodes `~/.neurolink/proxy-config.yaml` while the proxy supports a custom `--config` path; the catalog's "routed" additions can reflect the wrong file, and the generic loader now runs full validation on it (any issue → `[]`, silent drop). Suggest sharing the path the proxy actually loads. |

- **Registry wiring (rule 1 / no hand-rolled roster):** `grokConfigurator` added to `PROXY_CLIENT_CONFIGURATORS`; `applyAllClients`/`restoreAllClients` wrap each client in try/catch, so a write failure in the new writer is contained per-client, never aborts the loop. Roster tests (`testProxyClientRoster`, `testApplyAllReportsPerClient`) updated correctly.
- **No `grok-*` remap:** `catalogModelIds` skips `grok-` ids (incl. from YAML mappings); tests assert no `[model.grok-4.6]` block.
- **Context windows & doors:** verified `classifyGrokProxyModel` → `getContextWindowSize` for `claude-sonnet-4-6`/`claude-opus-4-6`=1M(messages), `claude-haiku-4-5`=200k(messages, reasoning off), `gemini-2.5-pro`=1,048,576(chat_completions). Adaptive-thinking guard for Haiku is sensible and matches the failing-400 rationale.
- **Snapshot/restore safety:** reuses `snapshot.ts` (`isUsableSnapshot`, `shouldCaptureSnapshot`, `writeFileAtomic`); restore refuses without a snapshot and refuses a foreign base URL; both refusal paths tested. `originalExisted` is preserved across re-applies.
- **Rule 4 (CLI ≠ SDK):** new code lives under `src/cli/proxy-clients/`; no SDK surface change (rule 5 clean — `types/proxyClient.ts` only adds new exported types, no existing shape modified).
- **Rule 15:** Grok tests sit under the declared `src/cli/proxy-clients/` determinism exception in the suite header; env (HOME/GROK_HOME) restored in `finally` blocks; temp dirs cleaned.
- **Security:** placeholder key `neurolink-proxy` is inert; no provider params logged; `clientAttribution` adds the `grok-shell/` prefix with a test.
- **Context-window / attribution drift:** `DEFAULT_PROXY_MODEL_IDS` consumers remain the proxy translation engine, OpenCode writer, roster test and the new Grok writer; doc-comment "three consumers" update is accurate.

### Not blocking
The after-merge validation (install package → `proxy start` on a dedicated machine; don't point the live 55669 proxy at this branch yet) is correctly listed as a follow-up in the PR body and doesn't gate merge.

src/cli/proxy-clients/grok.ts
Tara-ag: Confirmed the fix at head `92d30012`: `loadRoutedModelIds()` now goes through `parseProxyConfigString` (js-yaml) and `loadRoutedModelIds` is exposed via `__grokTestHooks`, with `testGrokYamlRoutedMappings` covering a YAML file that `JSON.parse` rejects. The original MAJOR finding is resolved — resolving this thread.

src/cli/proxy-clients/grok.ts
```diff
@@ -0,0 +1,388 @@
...
+  const provider = providerForModelId(id);
+  const apiBackend = provider === "anthropic" ? "messages" : "chat_completions";
+  const contextWindow = getContextWindowSize(provider, id);
+  return {
+    id,
+    name: displayNameFor(id),
+    apiBackend,
+    contextWindow,
+    maxCompletionTokens: contextWindow >= 1_000_000 ? 16_384 : 8_192,
+    supportsReasoningEffort: supportsAdaptiveThinking(id),
```
Tara-ag: **MINOR** — `loadRoutedModelIds()` hardcodes the default `~/.neurolink/proxy-config.yaml` path, but the proxy supports a custom routing config via `--config` (`proxyInstallCommand` persists it into the launchd plist: `const configPath = explicitConfig ? resolve(explicitConfig) : join(homedir(), ".neurolink", "proxy-config.yaml")`).

When the proxy runs with `--config /elsewhere/proxy-config.yaml`, the running proxy's `/v1/models` (via `loadProxyConfig`) advertises that file's `routing.model-mappings`, while this Grok writer reads only the default path — the `catch` returns `[]`, so the catalog silently drifts from what the proxy actually serves. The Grok writer applies whenever `~/.grok` exists, including on `--config` setups, so this is the divergence the PR's whole feature avoids elsewhere.

Reuse the proxy's resolved config path instead of the hard-coded default, e.g. resolve it from the same source `proxy start` uses:

````suggestion
async function loadRoutedModelIds(configPath?: string): Promise<string[]> {
  const resolvedPath =
    configPath ?? join(homedir(), ".neurolink", "proxy-config.yaml");
  try {
    const config = await parseProxyConfigString(
      readFileSync(resolvedPath, "utf8"),
    );
    return (config.routing?.modelMappings ?? [])
      .map((mapping) => mapping.from.trim())
      .filter((id) => id.length > 0);
  } catch {
    return [];
  }
}
````

At minimum, thread the proxy's `--config` value through `setGrokProxySettings` → `buildGrokManagedBlock` → `catalogModelIds` so the writer and the running proxy agree. (The existing `testGrokYamlRoutedMappings` covers the default path only.)

**Review by Tara-ag — APPROVED** [review]
The blocking MAJOR finding (JSON.parse on a YAML config) is resolved at head `92d30012`. One minor edge-case in `loadRoutedModelIds` (custom `--config` path) is noted inline. Approving — see summary comment.

**Comment by Tara-ag** [comment]
## Verdict: APPROVE

The previous MAJOR finding is fixed at head `92d30012` — `loadRoutedModelIds()` now parses `~/.neurolink/proxy-config.yaml` with `parseProxyConfigString` (js-yaml, env interpolation), and the YAML path is covered by `testGrokYamlRoutedMappings`. Review state updated to an approving review. One MINOR edge case remains (inline).

### Findings

| --- | --- | --- |
| ~~MAJOR~~ (resolved) | `src/cli/proxy-clients/grok.ts:119` | `loadRoutedModelIds` parsed a YAML file with `JSON.parse`. Fixed in `92d30012` via `parseProxyConfigString`; thread resolved. |
| MINOR | `src/cli/proxy-clients/grok.ts:112` | `loadRoutedModelIds` hardcodes the default `~/.neurolink/proxy-config.yaml`; the proxy supports a custom `--config` path, so the Grok catalog can silently drift from the running proxy on non-default setups. Suggest threading the resolved config path through `setGrokProxySettings` → `buildGrokManagedBlock` → `catalogModelIds`. |

- **Registry wiring** — `grokConfigurator` added to `PROXY_CLIENT_CONFIGURATORS` in apply order; roster and `testApplyAllReportsPerClient` updated; no circular import (rule 1).
- **No `grok-*` remap** — `catalogModelIds()` skips built-in `grok-` ids; covered by round-trip and YAML-mapping tests.
- **Context windows & backends** — `classifyGrokProxyModel` resolves upstream windows (Sonnet/Opus 4.6 = 1M, Haiku = 200k, Gemini 2.5 = 1,048,576) with messages vs chat_completions doors; adaptive-thinking guard for Haiku; covered by `testGrokCatalogWindowsAndBackends`.
- **Snapshot/restore safety** — reuses `snapshot.ts` helpers; restore refuses without a snapshot and refuses foreign base URLs; covered by both refusal tests.
- **Rule 4 (CLI ≠ SDK)** — writer lives only under `src/cli/proxy-clients/`; no SDK surface change (rule 5 clean).
- **Rule 15** — tests use the single-module `src/` graph with the declared determinism exception on `__grokTestHooks`.
- **Security** — no real secrets; `PLACEHOLDER_KEY = "neurolink-proxy"` is inert; `clientAttribution` `grok-shell/` prefix added with a test; secret scan clean.
- **Commit hygiene** — single conventional commit `feat(proxy): ...`; the after-merge install/`proxy start` validation is correctly a follow-up.

The prior two dual-posting summary comments (marked NEEDS_WORK) are superseded by this one.

File: src/cli/proxy-clients/grok.ts (hunk 3 above)
Comment (murdore) [reply src/cli/proxy-clients/grok.ts]: Fixed in 4c3cee14.

`proxy start` / `proxy install` resolve `--config` through `resolveProxyConfigPath` and `applyAllClients` now forwards that path. The Grok writer reads `routing.model-mappings` from the same file the running proxy loaded. Covered by `testGrokCustomConfigPath` (default vs custom YAML, plus `applyAllClients` forwarding).

File: src/cli/proxy-clients/grok.ts [hunk 4]
```diff
@@ -0,0 +1,399 @@
...
+}
+
+async function loadRoutedModelIds(configPath?: string): Promise<string[]> {
+  const resolvedPath = configPath ?? defaultProxyConfigPath();
+  try {
+    const config = await parseProxyConfigString(
+      readFileSync(resolvedPath, "utf8"),
+    );
+    return (config.routing?.modelMappings ?? [])
+      .map((mapping) => mapping.from.trim())
```
Tara-ag: **MAJOR** — catalog classification of a routed alias is derived from the alias *name*, not from the mapping's `to`/`provider`.

`loadRoutedModelIds()` collapses each `routing.model-mappings` entry down to `from` (this line), so `catalogModelIds()` hands `classifyGrokProxyModel(from)` only the alias id. `classifyGrokProxyModel` then derives backend/window/reasoning from `providerForModelId(id)` / `getContextWindowSize(provider, id)` — i.e. from the alias's own prefix.

`parseRoutingConfig` (proxyConfig.ts) keeps all three fields, and `ModelRouter.resolve()` routes the alias with `provider` + `to`. So the running proxy serves the alias via its true target, but the Grok picker entry advertises the *alias's* metadata. For a mapping like `enterprise-sonnet → claude-sonnet-4-6 (provider: anthropic)`:
- `providerForModelId("enterprise-sonnet")` → `"openai"` → `api_backend = "chat_completions"` (should be the `messages` door)
- `getContextWindowSize("openai", "enterprise-sonnet")` → 128k (true target is 1M)
- `supportsReasoningEffort` → false

Even a `claude-`-prefixed alias is off: `claude-sonnet-4-thinking → claude-sonnet-4-6` gets 200k (anthropic `_default`), not 1M. The PR's headline feature ("catalog windows and backends") is therefore wrong for every routed model whose alias differs from a real upstream id, and the new tests only use aliases whose names start with `claude-`/`gemini-` so they do not catch it.

Suggest classifying the *target* while keeping `from` as the picker id — thread `{ from, to, provider }` through instead of a bare id list:

```suggestion
    return (config.routing?.modelMappings ?? [])
      .map((mapping) => ({
        from: mapping.from.trim(),
        to: mapping.to.trim(),
        provider: mapping.provider,
      }))
      .filter((m) => m.from.length > 0);
```
and in `catalogModelIds` / `classifyGrokProxyModel` derive `apiBackend`, `contextWindow`, and `supportsReasoningEffort` from the mapping's `provider`/`to` while emitting `from` as the `[model.<from>]` key.

src/cli/proxy-clients/grok.ts
```diff
@@ -0,0 +1,399 @@
...
+    fs.accessSync(getGrokConfigDir());
+  } catch {
+    return false;
+  }
+
+  let original: string | null;
+  try {
+    original = fs.readFileSync(getGrokConfigPath(), "utf8");
+  } catch {
+    original = null;
```
Tara-ag, line 322: **MAJOR** — a *read* failure on an existing `config.toml` is treated as "file absent", so `setGrokProxySettings` can overwrite the user's real config with a bare managed block.

`original === null` is used three ways as if it meant "no file" (fill the archive snapshot with `originalExisted: false` here, blank the `withoutBlock` source, and pass `0o600` as the new-file mode). But the catch swallows *every* `readFileSync` error — `EACCES`/`EIO` on an existing, unreadable file collapses to `null` too. The method then proceeds to `writeFileAtomic(getGrokConfigPath(), …)`, replacing whatever is on disk with just the proxy block. Worse, the recorded `originalExisted: false` means a later restore runs `fs.rmSync(config.toml)` on the user's file.

This diverges from the sibling writer: `setCodexProxySettings` guards `existsSync` then lets `readFileSync` errors propagate into a `catch` that logs and **returns false** without writing (codex.ts). Only `ENOENT` should be treated as "no file":

```suggestion
  let original: string | null;
  try {
    original = fs.readFileSync(getGrokConfigPath(), "utf8");
  } catch (error) {
    if ((error as NodeJS.ErrnoException).code === "ENOENT") {
      original = null;
    } else {
      logger.warn(
        "[proxy] Grok: unable to read config.toml; leaving it untouched",
      );
      return false;
    }
  }
```
Consider a regression test that places an unreadable (non-ENOENT) `config.toml` and asserts `setGrokProxySettings` returns false and does not rewrite the file.

Tara-ag (changes requested):
Two MAJOR robustness/correctness findings remain in the new Grok writer (routed-model catalog classification and read-failure overwrite). See inline comments. The prior custom-config-path MINOR is confirmed fixed and resolved.

src/cli/proxy-clients/grok.ts
Tara-ag: Confirmed fixed at head `4c3cee14`: `loadRoutedModelIds(configPath?)` now threads the resolved `--config` path through `applyAllClients` → `setGrokProxySettings`/`options.configPath`, and `testGrokCustomConfigPath` covers default-vs-custom path selection plus `applyAllClients` forwarding. The writer and the running proxy now read the same file. Resolving this thread.

Tara-ag:
## Verdict: NEEDS_WORK

The change is solidly wired and the prior review's two findings are genuinely fixed, but two open issues remain in the new `grok.ts` writer — one correctness (routed-alias catalog classification) and one data-integrity (read-failure overwrite). Both are MAJOR and deserve a fix before merge. This supersedes the earlier `yama:summary` comments.

### Findings this round

|---|---|---|
| MAJOR | `src/cli/proxy-clients/grok.ts:127` | `loadRoutedModelIds()` returns only the `from` alias, so `catalogModelIds()` → `classifyGrokProxyModel()` derives backend/window/reasoning from the alias **name** instead of the mapping's `provider`/`to`. A mapping like `enterprise-sonnet → claude-sonnet-4-6 (anthropic)` is catalogued as `chat_completions`/128k/no-reasoning; even `claude-sonnet-4-thinking → claude-sonnet-4-6` gets 200k, not 1M. The proxy router still honours the mapping, so traffic isn't dropped, but the headline "catalog windows and backends" is wrong for routed models. Fix: classify from `to`/`provider`, keep `from` as the picker id. |
| MAJOR | `src/cli/proxy-clients/grok.ts:260` | `setGrokProxySettings` treats **any** `readFileSync` failure on `config.toml` (`EACCES`/`EIO`, not just `ENOENT`) as "file absent" → `original = null`, then atomically overwrites the existing-but-unreadable file with a bare managed block and records `originalExisted: false` (so restore would `rmSync` the user's real config). Diverges from `setCodexProxySettings`, which logs and returns false without writing on read failure. Fix: only `ENOENT` ⇒ absent; otherwise log + return false. |

### Resolved this round
- **Custom-config path (prior MINOR)** — confirmed fixed at `4c3cee14` (`configPath` threaded through `applyAllClients` → `setGrokProxySettings` → `loadRoutedModelIds`; `testGrokCustomConfigPath` covers both paths and forwarding). Thread resolved.
- **JSON.parse-on-YAML (prior MAJOR)** — already resolved at `92d30012` via `parseProxyConfigString` (js-yaml); `testGrokYamlRoutedMappings` guards it.

### Checked and clean
- **Registry wiring** — `grokConfigurator` added to `PROXY_CLIENT_CONFIGURATORS`; `applyAllClients` wraps each client in try/catch; roster + `testApplyAllReportsPerClient` updated. No `grok-*` remap (`catalogModelIds` skips `grok-` ids incl. from YAML mappings).
- **Context windows/doors for default catalog** — `claude-sonnet-4-6`/`claude-opus-4-6` = 1M messages, `claude-haiku-4-5` = 200k messages (reasoning off), `gemini-2.5-pro` = 1,048,576 chat_completions — each matches upstream (`getContextWindowSize`), covered by `testGrokCatalogWindowsAndBackends`.
- **Snapshot/restore safety** — reuses `snapshot.ts` (`isUsableSnapshot`, `shouldCaptureSnapshot`, `writeFileAtomic`); restore refuses without a snapshot and refuses a foreign base URL; both refusal paths tested. `originalExisted` is preserved across re-applies.
- **Rule 4 (CLI ≠ SDK)** — new code lives only under `src/cli/proxy-clients/`; `types/proxyClient.ts` only *adds* exported types (no existing shape changed) → rule 5 clean. No SDK surface change (verified via graph: new file's impact is `registry.ts` + `proxy.ts` apply path + the proxy test suite).
- **Rule 15** — Grok tests sit under the declared determinism exception on `__grokTestHooks`; env (HOME/GROK_HOME) and temp dirs are restored/cleaned in `finally`.
- **Security** — `PLACEHOLDER_KEY = "neurolink-proxy"` is inert; no provider params logged; `clientAttribution` `grok-shell/` prefix added with a test.

### Residual, not re-flagged
`loadRoutedModelIds` funnels config reads through `parseProxyConfigString`, which runs full validation + `failOnUnresolvedAccountCredentials`; any failure is swallowed to `[]`. Since the boot-time `loadProxyConfig` validates the same file, by the time Grok's apply runs the config is normally valid — only worth tightening if the config can change between boot and apply.

src/cli/proxy-clients/grok.ts
murdore: Fixed in 521bb1f0.

Routed entries keep `from` as the picker id and classify `api_backend` / `context_window` / reasoning from the mapping's `to` and `provider`. `enterprise-sonnet → claude-sonnet-4-6` is now messages/1M. Covered by `testGrokRoutedAliasUsesTargetMetadata`.

src/cli/proxy-clients/grok.ts
murdore: Fixed in 521bb1f0.

Only `ENOENT` is treated as a missing `config.toml`. Any other read error logs and returns false without writing. Covered by `testGrokApplyRefusesUnreadableConfig`.

Tara-ag:
## Verdict: APPROVE

The two open MAJOR findings from the last round are genuinely fixed at head `521bb1f0` — I re-verified each fix in the current source and its regression test, then resolved both threads. No new blocking issues in this pass. This supersedes the earlier `yama:summary` comments.

### Resolved this round (verified in code, both threads resolved)

| Finding | Verified fix |
|---|---|
| MAJOR `grok.ts:136` **routed alias classified from the alias name, not the target** | `classifyGrokProxyModel(id, mapping)` now derives `classifyId` from `mapping.to`, provider from `providerFromMapping(mapping.provider, …)`, and window from `getContextWindowSize(windowProvider, classifyId)`. `catalogGrokSpecs` builds `routedByFrom` and threads the mapping through. `enterprise-sonnet → claude-sonnet-4-6 (anthropic)` is now `messages`/1M with reasoning on. Covered by `testGrokRoutedAliasUsesTargetMetadata` and `testGrokCatalogWindowsAndBackends`. Author reply at `#discussion_r3995723705` is accurate — accepted. |
| MAJOR `grok.ts:313` **a non-ENOENT read failure on `config.toml` treated as "absent" → overwrite + `rmSync` on restore** | `setGrokProxySettings` now routes the `readFileSync` catch through `isMissingFileError(error)` — only `ENOENT` ⇒ `original = null`; any other error logs `warn` and `return false` without writing (mirrors `setCodexProxySettings`). There's also a fresh guard for an unreadable snapshot file (line 325). Covered by `testGrokApplyRefusesUnreadableConfig`. Author reply at `#discussion_r3995723770` is accurate — accepted. |

### Resolved earlier (already accepted)
- MAJOR `JSON.parse`-on-YAML (`loadRoutedModelIds` → `parseProxyConfigString`/js-yaml) — fixed at `92d30012`, thread resolved.
- MINOR hardcoded default config path — fixed at `4c3cee14`; `resolveProxyConfigPath` in `proxy.ts` threads `--config` through `applyAllClients(...).apply(options.configPath)`; `testGrokCustomConfigPath` asserts the forwarding. Verified `resolveProxyConfigPath` is called from `startProxyCommandHandler` and `loadRoutedMappings` honours the path.

### Checked this round and clean
- **Registry wiring (rule 1 / no hand-rolled roster):** `grokConfigurator` added to `PROXY_CLIENT_CONFIGURATORS`; `applyAllClients` and `restoreAllClients` wrap each client in try/catch, contained per-client. Roster + `testApplyAllReportsPerClient` updated with `grok`.
- **No `grok-*` remap:** `catalogGrokSpecs` skips `grok-` ids (incl. from `model-mappings`); tests assert no `[model.grok-4.6]` block.
- **Context windows & doors:** traced `classifyGrokProxyModel` → `getContextWindowSize` for the default catalog (Sonnet/Opus 4.6 = 1M messages, Haiku 4.5 = 200k messages reasoning-off, Gemini 2.5 Pro = 1,048,576 chat_completions) — matches upstream and the adaptive-thinking rationale (Grok `xhigh` → Anthropic `thinking.type="adaptive"`, which Haiku 4.5 400-rejects).
- **Snapshot/restore lifecycle:** `originalExisted` recomputed correctly across apply → restore → re-apply (verified hand-derivation); `writeFileAtomic` 0600 only on new files; restore refuses without a snapshot and refuses a foreign base URL; snapshot unreadable-guard added.
- **Rule 4 (CLI ≠ SDK):** new code lives only under `src/cli/proxy-clients/`; `types/proxyClient.ts` only *adds* `CliGrokProxyModelSpec`/`CliGrokSnapshot`/`CliProxyClientApplyOptions` — no existing shape changed, so rule 5 is clean. `CliProxyClientConfigurator.apply` gains an optional `options` param — the only public-surface touch, and it is backward-compatible (`options?`).
- **Rule 15:** Grok tests sit under the declared `__grokTestHooks` determinism exception; `HOME`/`GROK_HOME` and temp dirs restored in `finally`.
- **Security:** secret scan of the new files is clean; no provider params, keys, or tokens logged.

### Residual note (not re-flagged)
`loadRoutedMappings` funnels config through `parseProxyConfigString`, which runs full validation + `failOnUnresolvedAccountCredentials`; any failure is swallowed to `[]`. Since boot-time `loadProxyConfig` validates the same file, the config is normally valid by the time `apply` runs. Also a narrow edge: if a `routing.model-mappings` `from` equals a default catalog id, the default entry wins the `seen`-dedup in `catalogGrokSpecs`, so the mapping's target metadata is not applied — only relevant for users who re-alias a default id. Both are acceptable as-is for this feature.

Tara-ag (approved):
All prior findings are resolved at head `521bb1f0`. Both remaining MAJORs from the last round verified fixed in the checkout — routed-alias catalog classification now uses the mapping's `to`/`provider` (`classifyGrokProxyModel`, `testGrokRoutedAliasUsesTargetMetadata`), and the read-failure overwrite path now treats only `ENOENT` as "file absent" and logs/returns false otherwise (`setGrokProxySettings`, `testGrokApplyRefusesUnreadableConfig`). Resolving both threads. Approving — see summary comment.

Tara-ag:
## Verdict: APPROVE

All findings from the review rounds are resolved at head `521bb1f0`; the two remaining MAJORs were verified fixed in the checkout and their threads resolved. This is the consolidated, canonical summary (prior `yama:summary` drafts across rounds are superseded).

### Resolved this round (threads closed)

| Severity | Location | Finding | Resolution |
|---|---|---|---|
| MAJOR | `src/cli/proxy-clients/grok.ts` | Routed-alias catalog classification derived backend/window/reasoning from the alias *name* instead of the mapping's `to`/`provider` → `enterprise-sonnet → claude-sonnet-4-6 (anthropic)` catalogued as `chat_completions`/128k. | `classifyGrokProxyModel(id, mapping?)` now classifies from `mapping.to` + `providerFromMapping`; `catalogGrokSpecs` threads the mapping via a `routedByFrom` map. `testGrokRoutedAliasUsesTargetMetadata` asserts messages/1M/reasoning-on. |
| MAJOR | `src/cli/proxy-clients/grok.ts` | `setGrokProxySettings` treated any `readFileSync` failure (incl. `EACCES`/`EIO`) as "file absent", overwriting an unreadable `config.toml` and recording `originalExisted: false` (→ restore would `rmSync` it). | Only `ENOENT` ⇒ `original = null`; other errors log + `return false` without writing, matching `setCodexProxySettings`. `testGrokApplyRefusesUnreadableConfig` covers it. |

### Earlier findings (accepted fixes, threads already resolved)

- **MAJOR JSON.parse-on-YAML** (`loadRoutedModelIds`) — fixed at `92d30012` via `parseProxyConfigString` (js-yaml + env interpolation); `testGrokYamlRoutedMappings` guards the YAML path.
- **MINOR custom-config path** (`loadRoutedModelIds` hardcoded `~/.neurolink/proxy-config.yaml`) — fixed at `4c3cee14`; `configPath` threaded through `applyAllClients` → `setGrokProxySettings`; `testGrokCustomConfigPath` covers both paths.

### Checked and found clean

- **Registry wiring (rule 1 / no hand-rolled roster)** — `grokConfigurator` added to `PROXY_CLIENT_CONFIGURATORS`; `applyAllClients` wraps each client in try/catch so a writer failure is contained per-client. Roster tests updated.
- **No `grok-*` remap** — `catalogGrokSpecs` skips `grok-` ids (incl. from YAML mappings); tests assert no `[model.grok-4.6]` block.
- **Context windows / doors for the default catalog** — Sonnet/Opus 4.6 = 1M (`messages`), Haiku 4.5 = 200k (`messages`, reasoning off), Gemini 2.5 = 1,048,576 (`chat_completions`); each matches `getContextWindowSize`; covered by `testGrokCatalogWindowsAndBackends`. Adaptive-thinking guard for Haiku matches the measured 400.
- **Snapshot/restore safety** — reuses `snapshot.ts` (`isUsableSnapshot`, `shouldCaptureSnapshot`, `writeFileAtomic`); restore refuses without a snapshot and refuses a foreign base URL; `originalExisted` preserved across re-applies; unreadable-snapshot guard added.
- **Rules 4 / 5 (CLI ≠ SDK, public API)** — writer lives only under `src/cli/proxy-clients/`; `types/proxyClient.ts` only *adds* exported types (no existing shape changed). No SDK surface change.
- **Rule 15** — Grok tests use the single-module `src/` graph under the declared determinism exception on `__grokTestHooks`; env (HOME/GROK_HOME) and temp dirs are restored/cleaned in `finally`.
- **Security** — `PLACEHOLDER_KEY = "neurolink-proxy"` is inert; no provider params logged; `clientAttribution` `grok-shell/` prefix added with a test; secret scan clean.
- **Commit hygiene** — single conventional commit `feat(proxy): ...`. After-merge install + `proxy start` on a dedicated machine remains a correctly-scoped follow-up.

PR #1621, opened 2026-09-02
feat(mcp): let external server registration require a minimum discovered tool count

murdore:
## feat(mcp): let external server registration require a minimum discovered tool count

`ExternalServerManager.addServer` previously returned `success: true` / status `"connected"` (and incremented `connectedServers`) for a stdio server that discovered **zero** tools — indistinguishable from a legitimate resource/prompt-only server and from a quietly-broken one. There was no way for a caller to require a minimum tool count.

### Change (additive, default-compatible)

- `minTools?: number` on `MCPServerInfo` (`src/lib/types/mcp.ts`), default `0` → unchanged behaviour for the resource/prompt-only common case.
- `startServer` compares `instance.toolsMap.size` against `config.minTools ?? 0` right after discovery, **before** tools are registered with the main registry and before health monitoring starts — a rejected server is never marked ready, even transiently. Below the floor it tears the connection down via a new `closeInstanceConnection` helper (extracted verbatim from `stopServer`'s client/transport teardown, now shared by both) and throws `InsufficientToolsError`.
- `addServer`'s catch recognizes `InsufficientToolsError` via `instanceof` and returns `success: false` with `metadata.readiness: "insufficient_tools"` and `metadata.toolsDiscovered`, instead of the generic stringified-error path. Successful registrations now also carry `metadata.readiness: "ready"` for symmetry.
- New `MCPServerReadiness = "ready" | "insufficient_tools" | "failed"`. `ExternalMCPOperationResult.metadata` gains optional `readiness` / `toolsDiscovered`, typed concretely and intersected with `[key: string]: JsonValue` so the public read contract is unchanged (Rule 5).

### Tests

New `test/continuous-test-suite-mcp-min-tools-readiness.ts` + stdio fixture `test/fixtures/mcp-zero-tools-server.mjs`, driving real stdio servers through `ExternalServerManager` and `NeuroLink.addExternalMCPServer` / `getMCPStatus()`. Wired into `package.json` as `test:mcp:min-tools-readiness` and the `test:mcp:full` chain. `typecheck`, `build`, eslint/prettier, and the mcp infra/stdio-lifecycle/result-cache/spans regression suites all pass.

> Review state: approved. See the `<!-- yama:summary -->` review comment for the full recurring review.

PR #1689, opened 2026-09-12
chore(docs-site): stop committing the llms artifacts and repoint docs:build

murdore:
Two defects that hid behind each other. Both surfaced while rebasing unrelated PRs.

## The conflict tax

`docs-site/static/llms-full.txt` is **9.5 MB**, regenerated wholesale by `build-llms-txt.ts` on every docs-site build, and served from the built site. It is **not** in `package.json` `files`, so unlike `search-index.json` it has never shipped to a consumer.

Committing it bought nothing and cost a whole-file merge conflict on essentially every PR that touched docs. #1668, #1649 and #1621 were all reported `CONFLICTING` on that single file, with **no source conflict at all** — every real source file auto-merged. `llms.txt` is the same story at 52 KB.

Both are gitignored now. The deploy build still emits them, because it always did — `docs-site build` is `sync-docs && build:llms-txt && docusaurus build`.

The currency check narrows to the one artifact whose staleness can actually reach a user: `search-index.json`, which does ship in the tarball.

## Why nobody noticed

`pnpm-workspace.yaml` deliberately declares no `packages:` — the cooldown policy lives there and `docs-site/` stays an independent project. So `--filter ./docs-site` matches **no project**: pnpm prints `No projects matched the filters` and **exits 0**.

All four `docs:*` scripts were written that way, including the `docs:build` that the artifact-currency check printed as *its own remediation*. Running exactly what the failure told you to run regenerated nothing and reported success.

Repointed to `pnpm --dir docs-site`, which does not depend on workspace membership.

## Verification

Both forms run in the same tree:

| | result |
|---|---|
| `pnpm --filter ./docs-site build` | `No projects matched the filters`, exit 0, nothing written |
| `pnpm run docs:build` (fixed) | full build, **14336** search-index entries from 3917 files, 395 OG images, exit 0 |

After the fixed build:

- both llms files are back on disk and correctly ignored (`git check-ignore` confirms)
- `search-index.json` is **byte-identical**, so the narrowed gate passes
- no untracked residue
- `pnpm run lint`: **0 errors**

## Scope

- `.gitignore` — ignore the two llms artifacts, with the reason
- `package.json` — four `docs:*` scripts repointed to `--dir`
- `.github/workflows/docs-site-artifacts.yml` — check narrowed to `search-index.json`; header and remediation text corrected
- `CLAUDE.md` — both traps recorded, including the general form: a command that exits 0 is not evidence it did anything

The diff is large because it deletes two generated files. No source file changes.

## Summary by CodeRabbit

- **Documentation**
  - Clarified which documentation artifacts are committed and published.
  - Added guidance for running documentation commands reliably from the documentation site directory.
  - Removed generated content from the tracked `llms.txt` documentation artifact.

- **Chores**
  - Updated documentation build checks to validate only the committed search index.
  - Excluded generated `llms` files from version control.
  - Updated documentation site startup commands to generate `llms` files automatically.
  - Adjusted related workflow messages and command configuration.


.gitignore
```diff
@@ -196,3 +196,11 @@ NEUROLINK_SDK_GAPS.md
 .yama/memory/*.sqlite-*
 .yama/*.log
 .yama/.env
+
+# docs-site llms artifacts: generated by docs-site/scripts/build-llms-txt.ts on
+# every `pnpm run build` and served from the built site, never from this repo.
+# Neither is listed in package.json `files`, so unlike search-index.json they do
+# not ship in the npm tarball — committing them bought nothing and cost a
+# 9.5 MB whole-file conflict on essentially every pull request touching docs.
+docs-site/static/llms.txt
```
Tara-ag, line 208: **MINOR** — these two ignores duplicate the exact rules already shipping in `docs-site/.gitignore`.

`docs-site/.gitignore` (unchanged by this PR) already contains:

```gitignore
# Generated LLM context files (built by build:llms-txt script)
static/llms.txt
static/llms-full.txt
```

Git checks the directory-local `.gitignore` in addition to the root one, so the llms files have been ignored since before this PR. The PR description says "Both are gitignored now", but that was already true — the change that actually stops the conflicts is the **deletion** of the two tracked files (`docs-site/static/llms.txt`, `docs-site/static/llms-full.txt`), not this block. (For contrast, note the root `search-index.json` ignore+unignore pair exists because that file *is* intentionally tracked.)

Not harmful — fight-in-depth. Two options if you want to tighten it:

```suggestion
# docs-site/static/llms.txt and llms-full.txt are already ignored by
# docs-site/.gitignore; the real change here is removing them from tracking
# so the 9.5 MB whole-file regeneration stops conflicting on docs PRs.
```

or drop these two lines and rely on `docs-site/.gitignore` (and say so in the commit message instead). Either way, consider keeping the rationale only where it is not duplicated.

Tara-ag (commented):
Solid, small-scope cleanup. Changes verified correct; one MINOR note on the redundant root `.gitignore` block inline.

Tara-ag:
**APPROVE** — the change is correct, low-risk (graph risk 0.00, no source files touched), and the rationale checks out against the repo.

### Findings

|---|---|---|
| MINOR | `.gitignore:205` | Root ignore block duplicates the same ignores already present in `docs-site/.gitignore`; the real fix is the file deletion, not this block. |

### What was checked and found correct

- **`--dir` repointing fixes the real bug.** `pnpm-workspace.yaml` at root sets only `minimumReleaseAge` with **no `packages:`** (verified), so `--filter ./docs-site` matches nothing and exits 0 — exactly as described. `docs-site/pnpm-workspace.yaml` is an empty upward-search shield. `pnpm --dir docs-site build` runs the real script in `docs-site`. `docs-site/package.json` has `build`, `start`, `serve`, `clear` scripts, so all four `docs:*` repoints are valid.
- **Workflow narrowing is sound.** The llms files are now untracked, so restricting `docs-site-artifacts.yml` to `search-index.json` removes the churn false-positives. The generated artifacts are confirmed to write to `static/` (buil `build-llms-txt.ts` → `static/llms.txt` + `static/llms-full.txt`); the search index is a postBuild hook. `search-index.json` stays in `package.json` `files` and is not ignored (the `+!` negation in both `.gitignore`s is intact) — so it remains checked and shipped.
- **No out-of-diff consumers break.** `docusaurus.config.ts` links `/llms.txt` and `/llms-full.txt` via `pathname:///` (serve-time, regenerated every build); nothing in `docs-site/mcp-server/*` (which ships in the tarball) reads the committed llms files; neither llms file is in `package.json` `files`. The workflow's own `pnpm run build` already used `working-directory: docs-site`, so it was never broken by the `--filter` bug.
- **Blast radius:** graph build shows 0 nodes/edges changed and risk 0.00 — pure config/docs/generated-file change; no execution flows meaningfully affected.
- No secrets, no hot-path code, single Conventional Commit with scope (`chore(docs-site)`), one commit.

### Note (accepted design tradeoff, not a finding)
The check now covers only `search-index.json`, so nothing in CI verifies the llms files are still produced (only the deploy build does). That is the intended narrowing, and it is documented in `CLAUDE.md`; flagging for visibility only.

.gitignore
```diff
@@ -196,3 +196,11 @@ NEUROLINK_SDK_GAPS.md
...
+docs-site/static/llms-full.txt
```
Tara-ag: **MINOR** — this root block duplicates rules `docs-site/.gitignore` already carries.

`docs-site/.gitignore` (at this head) already ignores `static/llms.txt` and `static/llms-full.txt`:

```gitignore
# Generated LLM context files (built by build:llms-txt script)
static/llms.txt
static/llms-full.txt
```

Git applies nested `.gitignore` files, so `docs-site/static/llms.txt` / `docs-site/static/llms-full.txt` were **already ignored at the root level before this PR**. They were tracked only because git never untracks an already-tracked file — the real fix here is this PR's `git rm` of the two files, not the ignore rules. After the deletion the redeployed files are caught by the pre-existing `docs-site/.gitignore` rule, so this block is functionally redundant.

If the intent is root-level documentation (so someone grepping the repo root sees why the files are absent), keep it but say so. Otherwise drop the duplicate block:

```suggestion
```

Tara-ag (approved):
Solid, small-scope cleanup. The `--dir` repoint is genuine — root `pnpm-workspace.yaml` has no `packages:` field so `--filter ./docs-site` matches nothing and exits 0 — and every repointed script exists in `docs-site/package.json`. Verified the workflow narrowing and that no out-of-diff consumer reads the committed llms files. One MINOR note inline.

Tara-ag:
## Verdict: APPROVE

Clean, well-scoped cleanup — the `--dir` repoint is a real bug fix and the workflow narrowing is sound. One MINOR, non-blocking note.

### Findings

| MINOR | `.gitignore:200-206` | Root ignore block duplicates rules `docs-site/.gitignore` already has; the actual fix is the two `git rm`s. |

- **`--dir` repoint is correct** — root `pnpm-workspace.yaml` has no `packages:` field, so `--filter ./docs-site` matches no project and exits 0 (the documented trap). All four `docs:*` scripts (`start`/`build`/`serve`/`clear`) exist in `docs-site/package.json`, so `pnpm --dir docs-site <script>` resolves.
- **Workflow narrowing** — `search-index.json` stays tracked + in `package.json` `files` + wired through the check; the llms files are per-build artifacts referenced only via serve-time `pathname:///` links, so the gate correctly narrows to the one shipped artifact.
- **No out-of-diff breakage** — nothing else in the repo consumes the committed llms files (the docs MCP server that ships reads from the served site, not the repo tree).
- **Deletions are pure removals of generated content** — no source files changed (`additions 57 / deletions 383880`).

_Impact analysis from the code graph: risk 0.00, no source nodes affected — config/docs-only change, so no blast-radius concerns by construction._

.gitignore
murdore: Confirmed, and you are right on both halves. Fixed at `fbeabbd92`.

`git show origin/release:docs-site/.gitignore` carries `static/llms.txt` and `static/llms-full.txt`, and `git log --follow` puts them there in the original Docusaurus migration commit — long before this PR. Meanwhile `git ls-tree -r origin/release` lists both files, because a later ignore rule never untracks a file that is already tracked. So the ignore rules were never what stopped the conflicts; the `git rm` is.

I kept the two root-level lines but rewrote the comment so it no longer claims to be the fix:

```
# docs-site llms artifacts, generated by docs-site/scripts/build-llms-txt.ts and
# served from the built site. docs-site/.gitignore has ignored them since the
# Docusaurus migration, so these two lines change nothing on their own — what
# stops the conflicts is untracking the files, which this commit does.
```

Noting for the record that you posted this twice, on lines 205 and 206 — one finding, two threads. This reply covers both.

package.json
```diff
@@ -179,10 +179,10 @@
...
     "docs:generate": "pnpm run docs:validate",
     "// Documentation (Docusaurus)": "",
-    "docs:start": "pnpm --filter ./docs-site start",
-    "docs:build": "pnpm --filter ./docs-site build",
-    "docs:serve": "pnpm --filter ./docs-site serve",
-    "docs:clear": "pnpm --filter ./docs-site clear",
+    "docs:start": "pnpm --dir docs-site start",
+    "docs:build": "pnpm --dir docs-site build",
+    "docs:serve": "pnpm --dir docs-site serve",
+    "docs:clear": "pnpm --dir docs-site clear",
```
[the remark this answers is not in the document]
murdore: Confirmed and fixed at `fbeabbd92`. Both `start` and `dev` now run `build:llms-txt` between `sync-docs` and `docusaurus start`, matching the existing `build` script's ordering.

Worth recording that this PR is what makes the 404 reachable, rather than merely failing to close a pre-existing gap. Before it, the two files were tracked, so they were present on any checkout even though `start` never regenerated them. This PR deletes them. So the fix belongs here, not in a follow-up.

Verified rather than assumed — removing both files to simulate a fresh checkout and running the newly added step regenerates them byte-for-byte:

```
$ rm -f static/llms.txt static/llms-full.txt
$ pnpm run build:llms-txt        # exit 0
9940394 static/llms-full.txt
  52975 static/llms.txt
```

Production was never exposed: `docs-deploy.yml` already runs `build:llms-txt` explicitly before `pnpm build`. This was local dev only.

Tara-ag:
## Verdict: APPROVE

Clean, well-scoped cleanup. The `--filter` → `--dir` repoint is a genuine bug fix (verified: `pnpm-workspace.yaml` has no `packages:` field, so `--filter ./docs-site` matched nothing and exited 0), and the workflow narrowing to `search-index.json` is sound. **No new findings in this pass**; both prior findings from earlier rounds are now resolved.

### Prior findings — resolution check

| Finding | Severity | Disposition |
|---|---|---|
| `.gitignore:205` root block duplicates `docs-site/.gitignore` rules | MINOR | ✅ **Resolved** — author rewrote the comment to state plainly that the lines are redundant and the real fix is untracking the files; kept as root-level rationale. |
| `docs-site/package.json` `start`/`dev` missing `build:llms-txt` → local 404 | MINOR (CodeRabbit) | ✅ **Resolved** — both now run `build:llms-txt` between `sync-docs` and `docusaurus start`, matching `build`. |

### What was checked and found correct at this head

- **`--dir` repoint fixes the silent no-op.** Root `package.json` now runs `pnpm --dir docs-site start/build/serve/clear`; all four scripts exist in `docs-site/package.json`, so the repoints resolve, and `pnpm --dir` does not depend on workspace membership. The workflow's remediation (`pnpm run docs:build`) is therefore no longer a silent no-op.
- **Workflow narrowing is safe.** `docs-site/static/llms*.txt` are now untracked + gitignored; restricting `docs-site-artifacts.yml` to `search-index.json` (still committed, in `package.json` `files`, still wired through the same `git status --porcelain` check) removes the churn false-positives. `search-index.json`'s `+!` re-include negation is intact in both `.gitignore`s.
- **No consumer needs the committed llms files.** `docusaurus.config.ts` footer links `pathname:///llms.txt` / `llms-full.txt` (serve-time); `docs-deploy.yml` explicitly runs `build:llms-txt` before `docusaurus build`; `build-llms-txt.ts` writes them to `static/`. Nothing reads them from the repo tree. (`landing/static/llms-full.txt` + `landing/static/sitemap.xml` are a separate `landing/` project, untouched here.)
- **Blast radius:** config/generated-file/docs-only change — no `src/` or shipped-module impact, no execution flows affected. No secrets, no hot-path code.
- **Repo standards:** single Conventional Commit with scope (`chore(docs-site)`), one commit, accurate `CLAUDE.md` instrumentation (both traps + the general "exit 0 ≠ did work" lesson).

### Note (accepted, not a finding)
The check now covers only `search-index.json`, so nothing in CI asserts the llms files are still produced — only the deploy build does. That is the intended narrowing and is documented in `CLAUDE.md`; flagging for visibility only.

PR #1690, opened 2026-09-12
fix(catalog): reject a fallback model that cannot fall back

murdore:
## The gap

`loader.ts` derives each catalog provider's fallback positionally:

```ts
fallbackModelName:
  entry.models.fallbackModelName ??
  entry.models.fallbacks[1] ??
  entry.models.fallbacks[0],
```

Nothing checked the result. When that lands on the provider's **own default**, the
retry path is silently dead — the primary fails, the SDK retries the identical
model, and it fails again for the same reason. There is no error, no warning, and
nothing in CI that would notice.

## It is not hypothetical — three shipped catalogs are in that state

| catalog | models declared | fallback == default | avoidable |
| --- | --- | --- | --- |
| `gmicloud` | 1 | yes | no — nothing else to name |
| `inception-labs` | 1 | yes | no |
| **`mistral`** | **32** | **yes, set explicitly** | **YES** |

`mistral.json` ships **32 models** and explicitly pinned `fallbackModelName` to its
own default, overriding a perfectly serviceable `fallbacks[1]`
(`mistral-large-latest`). Deleting that one line restores a real fallback.

The two single-model catalogs have no alternative to name, so the check is scoped
to catalogs declaring more than one model. It reports a mistake, never an
unavoidable fact.

## Where the rule lives

In the schema's existing `superRefine`, beside the checks that already constrain
`default` and `fallbacks` to catalog keys — so it runs wherever catalogs are
parsed, rather than in a script someone has to remember to run. That is **both**
required CI jobs:

- `test` → `pnpm run codegen:catalog -- --check` (`ci.yml:221`)
- `provider-safety-net` → `pnpm run verify:provider-onboarding` (`ci.yml:440`), which
  imports `parseProviderCatalogJson` directly

## Before / after

Before, on the unmodified tree:

```
Error: Invalid provider catalog file mistral.json: models.fallbackModelName:
the fallback model resolves to "mistral-small-2506", which is also
models.default — a fallback identical to the primary cannot fall back. This
catalog declares 32 models, so name a different one (or drop fallbackModelName
to let fallbacks[1] apply). Single-model catalogs are exempt.
exit 1
```

After:

```
codegen:catalog complete — 15 providers                        exit 0
verify:provider-onboarding — 8/8 new providers fully onboarded exit 0
test:openai-compat-catalog — 41 passed · 0 failed              exit 0
pnpm run lint — 0 errors
```

No generated artifacts change: the fallback is resolved at load time, not emitted
into a generated file. The diff is two files.

## How this was found, and what it corrects

A review thread on #1648 suspected **Morph's** fallback was a code-edit specialist
serving general chat. I investigated, and a live probe showed the general-chat
claim was wrong — ordinary calls resolve `morph-v3-large` through a different
accessor. My proposed fix for Morph was worse than the problem: with only two
models, pinning the fallback to the default would have produced exactly the
fallback-that-cannot-fall-back this PR now rejects.

So the correct move was to check the derivation rather than patch the catalog.
Morph is untouched and passes. The real defect was in a provider nobody was
looking at.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved provider catalog validation to flag ambiguous fallback configurations when multiple models are available.
  * Updated the Mistral catalog to remove its explicitly declared fallback model mapping.


src/lib/providers/catalog/mistral.json
```diff
@@ -9,7 +9,6 @@
   },
   "models": {
     "default": "mistral-small-2506",
-    "fallbackModelName": "mistral-small-2506",
     "registryDefaultModel": "mistral-large-latest",
```
Tara-ag, line 13: MAJOR — the "restored" fallback may still not fall back for the accounts that need it.

`<!-- yama:catalog-fallback-1 -->`

Dropping the explicit `fallbackModelName` restores `fallbacks[1]` as the runtime fallback (`loader.ts` derives `fallbackModelName ?? fallbacks[1] ?? fallbacks[0]`). For this catalog that is **`mistral-large-latest`** — but this file's own `evidence.liveMatrix` (added in #1587) records that model as behind a plan gate:

> `mistral-large-latest` and `mistral-large-2512` answer 403 'not available in your subscription tier': a plan limit on this account, not a retirement.

So for a default/basic-tier user, the primary `mistral-small-2506` fails (e.g. rate limit) and the retry now targets a model that returns 403 — the retry path is still effectively dead, it just fails for a different reason. The premise of the PR ("restores a real fallback") only holds for accounts whose plan includes `mistral-large-latest`.

Consider pointing the fallback at a model this catalog verifies as reachable and not plan-gated, e.g. `mistral-small-2603` (production, not in the 403 list), either by ordering `fallbacks` so the desired model is `[1]` or by setting `fallbackModelName` to it explicitly. If the plan-gated choice is deliberate, it's worth a comment next to the `fallbacks` entry so a reader doesn't assume the retry recovers.

(Noting the fix itself — removing a pin that equals `default` — is correct; this is about the target that ends up being selected.)

src/lib/providers/catalog/schema.ts
```diff
@@ -258,6 +258,40 @@ const catalogModelsSchema = z
...
+    // This is not hypothetical. `mistral.json` ships 32 models and explicitly
+    // sets `fallbackModelName` to its own default, overriding a perfectly good
+    // `fallbacks[1]`. Two single-model catalogs also resolve this way and are
+    // correctly exempt.
+    const derivedFallback =
+      models.fallbackModelName ?? models.fallbacks[1] ?? models.fallbacks[0];
+    if (
+      derivedFallback !== undefined &&
+      derivedFallback === models.default &&
+      catalogKeys.size > 1
```
Tara-ag: MINOR — the exemption/count uses `catalogKeys.size` over the whole `models.catalog` map, which includes retired and non-chat (embed/OCR/audio) entries.

`<!-- yama:catalog-fallback-2 -->`

`catalogKeys = new Set(Object.keys(models.catalog))` counts every model-map entry, not the set that could actually serve as a chat fallback. Two effects:

1. **Error-message count can mislead.** For `mistral.json`, `catalogKeys.size` is ~33 (several entries are `status: "retired"` and embed/OCR models), but the message says "declares N models" — and the comment above claims "32 models". A reader debugging the failure gets a count that doesn't reflect usable models.
2. **The `> 1` gate can over-report.** A catalog with 2 catalog entries where only one is a usable chat fallback and `fallbacks = [default]` fails the check even though there is genuinely nothing better to name — the mirror of the single-model exemption.

The message and the exemption are both about "is there a distinct usable fallback to name". Consider basing the count on `models.fallbacks` (the pool the loader actually retries against) or on non-`retired` chat-capable entries instead of the whole map. If the whole-map count is intentional, note it so the emitted number isn't mistaken for the usable-model count.

src/lib/providers/catalog/schema.ts
```diff
@@ -258,6 +258,40 @@ const catalogModelsSchema = z
...
+    // the result. A provider with one model has no choice and is left alone;
+    // a provider with alternatives that still lands on its own default has
+    // silently disabled its retry path — the primary fails, the SDK retries
+    // the same model, and it fails again for the same reason.
+    //
+    // This is not hypothetical. `mistral.json` ships 32 models and explicitly
+    // sets `fallbackModelName` to its own default, overriding a perfectly good
+    // `fallbacks[1]`. Two single-model catalogs also resolve this way and are
+    // correctly exempt.
+    const derivedFallback =
```
Tara-ag, line 274: MINOR — the new validation branch has no direct test coverage.

`<!-- yama:catalog-fallback-3 -->`

The check is only tripped indirectly, when `parseProviderCatalogJson` is thrown by build tooling (`codegen:catalog --check`, `check:models`, `verify-provider-onboarding`). `test:openai-compat-catalog` (cited in the PR) drives `generate()` through `dist/index.js` and never parses the schema, so the 41 passing cases do not exercise either half of this new rule:
- that a multi-model catalog with `derivedFallback === default` is **rejected**, and
- that a **single-model** catalog is **exempt**.

That second half is the riskiest: if the `> 1` gate were accidentally removed or inverted, the two single-model catalogs (gmicloud, inception-labs) would fail every CI parse and nothing in the test suite would catch the regression except the CI failure itself. Given the repo's end-to-end-only rule (rule 15) makes a `generate()`-based assertion awkward for a build-time schema check, at minimum a `test/fixtures/` case (a valid multi-model JSON that should throw, and a coranked single-model JSON that should parse) driven through `parseProviderCatalogJson` in a `*suite-*.ts` would lock the behaviour in.

Tara-ag (commented):
The fix is well-targeted and correct: the new schema check mirrors the loader's positional derivation exactly (`fallbackModelName ?? fallbacks[1] ?? fallbacks[0]`), the single-model exemption correctly keeps gmicloud and inception-labs passing, and the check is enforced by the build-time CI paths (`codegen:catalog --check`, `check:models`, `verify-provider-onboarding`). Removing mistral's explicit pin to its own default restores a genuinely different fallback. Three non-blocking findings: (1) the restored fallback `mistral-large-latest` is documented in this catalog's own evidence as plan-gated — verify the retry actually lands on a reachable model; (2) the exemption/count uses the full `models.catalog` map size (incl. retired and embed/OCR entries), which can mislead the emitted message; (3) the new branch has no direct test coverage, notably the single-model exemption.

Tara-ag:
## Summary — PR #1690 `fix(catalog): reject a fallback model that cannot fall back`

**Verdict: NEEDS_WORK** (low) — the change is correct and well-enforced, but one MAJOR consideration about the restored fallback target should be confirmed before merge.

The core logic is sound:
- The `schema.ts` `superRefine` check mirrors `loader.ts`'s derivation **exactly** (`fallbackModelName ?? fallbacks[1] ?? fallbacks[0]`), so it reports what the runtime actually selects.
- The `catalogKeys.size > 1` gate correctly exempts the two single-model catalogs (`gmicloud`, `inception-labs`). I verified both are single-map and currently pass.
- The check is reachable-and-failing through three CI/build paths: `tools/codegen-catalog.ts::loadEntries`, `tools/check-model-liveness.ts::loadCatalog`, and `tools/verify-provider-onboarding.ts::checkCatalogProvider` all call `parseProviderCatalogJson` and throw on a parse error. (Note: `verify-provider-onboarding` skips legacy providers incl. mistral, so mistral is enforced by codegen + check:models.)

| --- | --- | --- |
| MAJOR | `src/lib/providers/catalog/mistral.json:13` | Restored fallback is `fallbacks[1]` = `mistral-large-latest`, which this catalog's own `evidence.liveMatrix` documents as plan-gated (403 on non-premium accounts) — the retry may still be dead for the accounts that need it. Consider a verified, non-gated target (e.g. `mistral-small-2603`). |
| MINOR | `src/lib/providers/catalog/schema.ts:279` | Exemption + emitted count use `catalogKeys.size` over the whole model map (incl. retired and embed/OCR entries) — mistral reports ~33, not "32", and a 2-entry catalog with only one usable chat fallback can over-report. |
| MINOR | `src/lib/providers/catalog/schema.ts:274` | No direct test covers the new branch — neither that a bad multi-model catalog is rejected nor that single-model catalogs are exempt. `test:openai-compat-catalog` drives `generate()` through `dist/` and never parses the schema. |

**Checked and clean**
- Loader derivation `src/lib/providers/catalog/loader.ts` matches the new schema check verbatim.
- `gmicloud.json` / `inception-labs.json` confirmed single-model and exempt.
- `mistral.json` after the change: derived fallback = `mistral-large-latest` ≠ default — check no longer fires.
- No runtime SDK blast radius: `schema.ts` is imported only by the three `tools/` CI paths (verified via code graph).
- No generated artifacts change (fallback resolved at load time).
- No secrets in the diff; conventional-commit + single-commit policy satisfied (per CI bot).

src/lib/providers/catalog/mistral.json
```diff
@@ -9,7 +9,6 @@
   },
   "models": {
     "default": "mistral-small-2506",
-    "fallbackModelName": "mistral-small-2506",
```
Tara-ag, line 12: **MAJOR** — the restored fallback may land on a model the retry cannot actually reach.

Before this change `fallbackModelName` pinned the fallback to Mistral's own default, so deleting that line lets the loader dereference `fallbacks[1]` = `mistral-large-latest`. Good — but this catalog's **own evidence** documents `mistral-large-latest` as premium-plan-gated (403 for non-premium accounts). The retry this PR un-dead-ends therefore only works for premium accounts; for the very accounts the fallback exists to save, the second attempt may still fail with a 403.

Since the PR is explicitly about restoring a *reachable* fallback, please verify the retry actually lands on a model non-premium users can call, or name `fallbacks[1]` differently (e.g. a `mistral-small-*` model).

```suggestion
    "fallbackModelName": "mistral-small-2506",
```

src/lib/providers/catalog/schema.ts
Tara-ag: **MINOR** — `catalogKeys.size` counts entries of the *whole* `models.catalog` map, not just runnable chat models.

This gate (`catalogKeys.size > 1`) and the message's `This catalog declares ${catalogKeys.size} models` include retired models and embed/OCR entries. That can over-report (a catalog whose *runnable chat* set has only one model but several non-chat/retired keys would be flagged, and the message would claim `N models` where a human only sees few). It also under-weights the intended meaning: "is there a real *alternative* to name."

If `catalogKeys` is set exactly once from `Object.keys(models.catalog)` it's fine; otherwise the exemption and the message should be computed from the same authoritative set as the resolve check.

```suggestion
      catalogKeys.size > 1
```

src/lib/providers/catalog/schema.ts
```diff
@@ -258,6 +258,40 @@ const catalogModelsSchema = z
...
+          models.fallbackModelName !== undefined
+            ? ["fallbackModelName"]
+            : ["fallbacks"],
+        message:
+          `the fallback model resolves to "${derivedFallback}", which is also models.default — ` +
+          `a fallback identical to the primary cannot fall back. This catalog declares ` +
+          `${catalogKeys.size} models, so name a different one (or drop fallbackModelName to ` +
+          `let fallbacks[1] apply). Single-model catalogs are exempt.`,
+      });
+    }
```
Tara-ag, line 301: **MINOR** — no test exercises the new branch, notably the single-model exemption.

The added `superRefine` block has no direct coverage: the closest suite that touches this schema (`test/continuous-test-suite-*.ts`) drives `generate()` through `dist/index.js` and never parses the catalog schema, so neither the detection path nor the `catalogKeys.size > 1` exemption (gmicloud / inception-labs) is asserted. A regression here would pass CI silently. A small end-to-end test that feeds a multi-model catalog with `fallbackModelName === default` (expect reject) and a single-model one (expect pass) would pin the contract.

Tara-ag (changes requested):
Three review findings on the fallback-check change — see inline comments. Non-blocking in the sense that the core fix is correct, but the mistral fallback reachability (MAJOR) should be verified before merge.

Tara-ag:
## NEEDS_WORK (low)

The fix is well-targeted and correct: the new schema check mirrors `loader.ts`'s positional derivation exactly (`fallbackModelName ?? fallbacks[1] ?? fallbacks[0]`), it exempts the two single-model catalogs so nothing breaks, and it is enforced on both build-time CI paths (`codegen:catalog --check` and `verify:provider-onboarding`). Three findings, none blocking the core logic.

| --- | --- | --- |
| MAJOR | `src/lib/providers/catalog/mistral.json:12` | Restored `fallbacks[1]` = `mistral-large-latest` is documented as plan-gated — verify the retry lands on a model non-premium users can call, else the fallback is still dead for them. |
| MINOR | `src/lib/providers/catalog/schema.ts:279` | `catalogKeys.size` counts the whole `models.catalog` map (retired + embed/OCR), which can over-report in both the gate and the emitted message. |
| MINOR | `src/lib/providers/catalog/schema.ts:276` | No direct test covers the new branch, notably the single-model exemption (gmicloud / inception-labs). |

**Checked and clean:**

- Schema/loader parity — the new check reproduces the resolution logic identically, so no load-time divergence.
- Single-model exemption — gmicloud and inception-labs correctly bypass; the typed test output confirms they pass.
- Enforcement — the check runs in CI via both `codegen:catalog -- --check` and `verify:provider-onboarding`, not as an opt-in script.
- Blast radius — schema.ts is only imported by the three `tools/` build-time paths; no SDK runtime surface changes (resolved via the code graph; no unmodified callers break).
- No generated artifacts change; the diff is two source files and one non-diff context only in the JSON.

murdore:
All three findings verified rather than taken on trust. Two applied in `727c079a7`; the third I am deliberately not resolving unilaterally, for a reason below.

## 1 (MAJOR) — correct, and my fix was wrong

You said the restored fallback may still not fall back. I checked this catalog's own `evidence` rather than reasoning from the comment, and it says so explicitly:

> `mistral-large-latest` and `mistral-large-2512` answer 403 'not available in your subscription tier': a plan limit on this account, not a retirement, so they stay production. **visionModel is pinned to `mistral-medium-latest` because the otherwise-first vision id, `mistral-large-latest`, is behind that plan gate.**

That last sentence is the part that settles it. **This file already works around this exact gate in another field.** My change walked straight into the thing the catalog had already learned to avoid — I swapped a fallback that could not fall back for one that returns 403, which is not obviously better.

`fallbacks` is now `["mistral-small-2506", "mistral-medium-latest"]`, using the same model the catalog itself chose when it hit this gate. That keeps the fix self-consistent with the file's own recorded evidence instead of introducing a third opinion. `mistral-large-latest` stays in `catalog` — it is a real production model for accounts whose plan includes it; what was wrong was making it the *retry target*, not listing it.

A near-miss worth recording: my first attempt at this edit used a string replace on `"mistral-large-latest"`, which matched its **catalog key** before its `fallbacks` entry. An assertion caught it before anything was written. Had it landed it would have silently renamed a model definition.

## 2 (MINOR) — correct

`catalogKeys.size` counted the whole map, retired entries included. Both effects you named are real: the emitted count misleads (32 vs the usable set), and the `> 1` gate could fire on a catalog whose only other entries are unusable — the mirror of the exemption the rule already grants. Now:

```ts
const selectableModels = Object.values(models.catalog).filter(
  (spec) => (spec as { status?: string }).status !== "retired",
).length;
```

used for both the gate and the message, which now says "N selectable models".

I stopped at excluding `retired` rather than also filtering to chat-capable. Embedding/OCR/audio entries have no marker on the spec that distinguishes them — this catalog's own evidence says Mistral rejects every non-chat model with the same generic "Invalid model" text, which is why they were classified by cross-referencing the roster by hand. Inferring capability from an id substring would be exactly the kind of guess this rule exists to replace.

## 3 (MINOR) — agreed, and I am NOT resolving it

You are right that the branch has no direct coverage, and right that the riskiest half is the exemption: if the `> 1` gate were inverted, `gmicloud` and `inception-labs` would fail every CI parse and only the CI failure itself would say so.

I have not added the test, because the honest route collides with a rule I should not quietly bend. Rule 15 is end-to-end-only, and `CLAUDE.md` says the determinism exception's allow list in `eslint.config.js` is **a review decision** — "not a way to silence the rule". A schema-validation test is precisely the kind of thing that needs that exception, so adding myself to the list to satisfy a review comment is the wrong shape.

Two ways forward, and I would rather you pick:

**(a)** Add the fixture suite you describe and add the file to the `neurolink/e2e-tests-only` allow list, with a header saying what determinism buys — treating this comment as the review decision that authorises it.

**(b)** Cover it end-to-end without the exception: a suite that copies a crafted catalog into a temp dir and runs `codegen:catalog` as a **subprocess**, asserting a non-zero exit and the message. That drives shipped tooling through a real entry point, the way `runCLI` does, so no exception is needed. More moving parts, but it tests the gate as CI actually invokes it.

I lean (b), because the thing worth locking is "CI rejects this catalog", not "the zod refinement returns an issue". Say which and I will implement it.

## Verification after the changes

```
codegen:catalog complete — 15 providers                        exit 0
verify:provider-onboarding — 8/8 new providers fully onboarded exit 0
test:openai-compat-catalog — 41 passed · 0 failed              exit 0
pnpm run lint — 0 errors, 0 prettier warnings
```

src/lib/providers/catalog/schema.ts
Tara-ag: Recurring review — keeping this finding open, and picking **option (b)**.

The finding stands: the new `superRefine` branch has no test that asserts either half (reject a multi-model catalog whose effective fallback equals `default`; exempt single-model catalogs). Nothing in `test/` parses the schema — `test:openai-compat-catalog` drives `generate()` through `dist/index.js`, which never reaches `parseProviderCatalogJson`.

I've reviewed the trade-off you laid out and agree (b) is right, for the reason you give: what's worth locking is "CI rejects a bad catalog", not "the zod refinement returns an issue". That is the actual contract this rule is for, and a subprocess test drives it through the real entry point the way `runCLI` does.

Go with (b): a `test/continuous-test-suite-*.ts` that writes a crafted catalog into a temp dir, runs the built `codegen-catalog --check` (or the equivalent built entry you verify) as a subprocess, and asserts:
- a multi-model catalog with `fallbackModelName === models.default` (and a distinct `fallbacks[1]`) exits non-zero and the message names the default, and
- a single-model catalog with the same shape exits zero (the exemption).

This needs no determinism-exception entry in `eslint.config.js`, stays end-to-end on the built tool, and pins the gate a future inversion would silently break.

Tara-ag:
## Recurring review — PR #1690 `fix(catalog): reject a fallback model that cannot fall back`

**Verdict: APPROVE** — the change is correct, the two substantive findings are resolved in `727c079a7`, and the one remaining item (MINOR) is covered by direction given inline below rather than a blocker.

of three prior findings:

| # | Severity | Finding | Status |
| --- | --- | --- | --- |
| 1 | MAJOR | Restored `fallbacks[1]` = `mistral-large-latest` is plan-gated (403) — retry dead for the accounts that need it | **Resolved** — `fallbacks[1]` is now `mistral-medium-latest`, which the catalog's own `evidence.liveMatrix` records as the id it already uses to work around this exact gate (visionModel pin). Self-consistent with the file's recorded evidence. |
| 2 | MINOR | Exemption/count used `catalogKeys.size` over the whole model map (retired + embed/OCR) | **Resolved** — now `selectableModels` excluding `retired`, used for both gate and message ("N selectable models"). |
| 3 | MINOR | No test covers the new branch (reject multi-model with `fallback == default`; exempt single-model) | **Open** — author deferred to reviewer; direction given: implement as an end-to-end subprocess test driving the built `codegen-catalog --check` (no determinism exception needed). |

**Checked and clean**
- Loader parity: `loader.ts` derives `fallbackModelName ?? fallbacks[1] ?? fallbacks[0]`, and `schema.ts` mirrors it exactly — the check reports what the runtime actually selects.
- Single-model exemption: gmicloud / inception-labs have one model, `selectableModels === 1`, correctly bypassed.
- Enforcement: `schema.ts` is reached from all three `tools/` build/CI paths (`codegen-catalog.ts`, `check-model-liveness.ts`, `verify-provider-onboarding.ts`); verified via the code graph that it is imported by no runtime SDK surface — no blast radius on unmodified callers.
- `mistral-medium-latest` is a key in `models.catalog` with `status: production`, so the loader's `fallbackModels` stays valid.
- No generated artifacts change (fallback resolved at load time); no secrets in the diff; single-commit + conventional-commit satisfied.

Recommended follow-up (non-blocking): the option-(b) subprocess test above, so the new gate is pinned against silent regression.

src/lib/providers/catalog/schema.ts
Tara-ag: `<!-- yama:catalog-fallback-3 --/>`

Thanks — option **(b)** is the one I'd take. The thing worth locking is "CI rejects this catalog," not "the zod refinement returns an issue," so the subprocess suite is the right shape: copy a crafted catalog into a temp dir, run the built `node dist/... codegen-catalog --check` (or the CLI) as a subprocess, and assert a non-zero exit for the multi-model duplicate-default case and a zero exit for the single-model exemption. That drives shipped tooling through a real entry point, needs no determinism exception, and would have caught the exact regression you called out (an inverted/removed `> 1` gate failing every CI parse). Go with (b).

Tara-ag (approved):
Approving — the fallback-check fix is correct and both substantive findings are resolved in `727c079`:

- The `superRefine` check in `schema.ts` mirrors `loader.ts`'s derivation exactly (`fallbackModelName ?? fallbacks[1] ?? fallbacks[0]`), and the `selectableModels` count now correctly excludes `retired` entries for both the gate and the emitted message.
- `mistral.json` drops its self-pinning `fallbackModelName` and points `fallbacks[1]` at `mistral-medium-latest` — consistent with this catalog's own `evidence.liveMatrix` (which already pinned `visionModel` to that model to dodge the same plan gate). Self-consistent and defensible.
- Single-model catalogs (gmicloud, inception-labs) stay exempt via `selectableModels > 1`.

One non-blocking MINOR remains (no direct test coverage for the new branch); direction confirmed on the thread (an end-to-end subprocess test, option b).

Tara-ag:
## Summary — PR #1690 `fix(catalog): reject a fallback model that cannot fall back`

**Verdict: APPROVE** — the fallback-check is correct, well-scoped, and both substantive findings are resolved in `727c079`; one non-blocking MINOR (test coverage) is tracked with confirmed direction.

### Findings

| --- | --- | --- | --- |
| MAJOR | `src/lib/providers/catalog/mistral.json:13` | Restored fallback targeted a plan-gated model (`mistral-large-latest`). | **Resolved** — author switched `fallbacks[1]` to `mistral-medium-latest`, consistent with this catalog's own `evidence.liveMatrix` which already pins `visionModel` to that model to dodge the same 403 gate. Self-consistent and accepted. |
| MINOR | `src/lib/providers/catalog/schema.ts:279` | Exemption + count used whole `catalog` map size (incl. `retired`). | **Resolved** — applied in `727c079` via a `selectableModels` count (`status !== "retired"`), used for both gate and message. |
| MINOR | `src/lib/providers/catalog/schema.ts:274` | No direct test covers the new branch (esp. the single-model exemption). | **Open, non-blocking** — direction confirmed on thread: an end-to-end subprocess test that runs `codegen-catalog --check` against a crafted temp catalog. |

### Checked and clean

- **Loader parity** — the new check reproduces `loader.ts`'s derivation (`fallbackModelName ?? fallbacks[1] ?? fallbacks[0]`) verbatim, so it reports what the runtime actually selects.
- **Single-model exemption** — `selectableModels > 1` correctly exempts gmicloud / inception-labs (both single, non-`retired`).
- **Blast radius** — `schema.ts` is imported only by the three `tools/` CI paths (`codegen-catalog.ts`, `verify-provider-onboarding.ts`, `check-model-liveness.ts`); no runtime SDK surface. No unmodified callers break.
- **No generated artifacts change** (fallback resolved at load time); no secrets in the diff; conventional-commit + single-commit policy satisfied (per CI bot).

### Note on earlier-round comments

Two prior round summaries carrying the `<!-- yama:summary -->` marker were posted earlier with a NEEDS_WORK verdict; both are superseded by this one. I could not delete/rewrite existing issue comments with the tooling available here, so this single APPROVE summary is the authoritative final record, and the review state now matches it.

src/lib/providers/catalog/schema.ts
murdore: Covered now, and without taking the Rule 15 exception.

You were right that `test:openai-compat-catalog` never parsed the schema, so none of its 41 cases touched either half of this rule. Rather than add a fixture suite and an allow-list entry, the two new cases run the **real tooling** in a subprocess — `tools/codegen-catalog.ts --check`, whose `CATALOG_DIR` is relative, so pointing the child's cwd at a copied catalog tree is enough. The test file imports nothing from `src/`, so `neurolink/e2e-tests-only` is satisfied as it stands and the allow list is untouched. It also needed no new CI wiring: this suite is already in the `extended-suites` `SUITES` list.

**Rejection half** — copy the catalog tree, set `cerebras.json`'s `fallbackModelName` to its own `default` (it ships two selectable models), run the tool, and require both a non-zero exit *and* the rule's own text naming the file.

**Exemption half** — this is the one you called riskiest, and it needed no fixture at all: `gmicloud.json` and `inception-labs.json` are shipped single-model catalogs that already resolve their own default. The case asserts that as a precondition, then requires the real tree to check clean.

Both mutations are caught, and by the right assertion each time:

```
# drop the `selectableModels > 1` gate
✗ ... is exempt when the catalog has only one model
    — a single-model catalog was rejected by the rule

# disable the rule entirely
✗ ... is rejected when the catalog has other models
    — rejected, but not by the fallback rule
```

With the rule as shipped: 43 passed, 0 failed. Note the first mutation also trips the rejection case's *precondition* ("the unmodified tree already trips the rule") rather than letting it pass on a coincidence — which is the precondition doing its job.

test/continuous-test-suite-openai-compat-catalog.ts
```diff
@@ -1155,6 +1166,139 @@ async function testCatalogStructuralInvariants(): Promise<void> {
...
+// Section: the catalog schema's "a fallback cannot be the primary" rule
+//
+// This rule only ever fires through build tooling — `codegen:catalog --check`,
+// `check:models`, `verify:provider-onboarding` — so driving `generate()` does
+// not reach it. Rather than import the schema (which would make this a unit
+// test), these cases run the real tool in a subprocess against a copy of the
+// real catalog directory. `CATALOG_DIR` in tools/codegen-catalog.ts is relative,
+// so setting the child's cwd is enough to point it at a fixture tree.
+// ───────────────────────────────────────────────────────────────────────
+
```
Tara-ag, line 1180: `<!-- yama:catalog-test-cwd-coupling -->`

**MINOR (non-blocking)** — `REPO_ROOT = process.cwd()` couples both new helpers to the launch directory.

`makeCatalogTree()` resolves the copy source via the relative `CATALOG_SRC = "src/lib/providers/catalog"` against `process.cwd()`, and the exemption half invokes `runCodegenCheck(REPO_ROOT)` (also `process.cwd()`) against nothing but the current-working-dir catalog tree. If this suite is ever launched from anywhere other than the repo root (some harnesses invoke suites from `dist/` or a workspace subdir), the rejection half's source copy fails and the exemption half silently validates the wrong tree — the "it still checks clean" pass would be meaningless.

The other suites in this repo run `dist/index.js` / the built CLI, whose entry is located relative to the *module*, not the cwd. Suggest deriving the source from `import.meta.url` (or `fileURLToPath` onto the repo root) rather than `process.cwd()`, so the helper is cwd-independent — e.g.:

````suggestion
const HERE = dirname(fileURLToPath(import.meta.url));
const REPO_ROOT = resolve(HERE, "../..");
````

(needs `import { dirname } from "node:path"` and `fileURLToPath` from `node:url`). Read-only robustness note; the rule logic itself is correct and the shipped-tree checks are right.

Tara-ag (commented):
Verified the fallback-rule change and the new subprocess tests. The three prior findings are all resolved in this HEAD; the only addition is a MINOR robustness note on the new test helpers' `process.cwd()` coupling.

Tara-ag:
## Summary — PR #1690 `fix(catalog): reject a fallback model that cannot fall back`

**Verdict: APPROVE** — the fix is correct, the two substantive findings were already resolved, and the previously-open test-coverage finding is now implemented end-to-end in this HEAD (`0543316`). One non-blocking MINOR remains on the new test helpers.

### Findings this round

| --- | --- | --- | --- |
| MAJOR | `src/lib/providers/catalog/mistral.json` | Restored fallback targeted plan-gated `mistral-large-latest` (403 on non-premium). | **Resolved (prior round, accepted)** — `fallbacks[1]` → `mistral-medium-latest`, the id this catalog's own `evidence.liveMatrix` already uses to dodge the gate (visionModel pin). |
| MINOR | `src/lib/providers/catalog/schema.ts` | Exemption/count used whole `catalog` map size (incl. `retired`). | **Resolved (prior round)** — `selectableModels` now filters `status !== "retired"`, used for both gate and message. |
| MINOR | `src/lib/providers/catalog/schema.ts` | No direct test coverage of the new branch (the single-model exemption was the riskiest). | **Resolved (this HEAD)** — `testCatalogFallbackRule` drives the real `tools/codegen-catalog.ts --check` in a subprocess with the child cwd pointed at a temp catalog tree; asserts multi-model rejection (cerebras fixture) and single-model exemption (gmicloud/inception), with preconditions guarding against false passes. No `src/` import, so `neurolink/e2e-tests-only` is satisfied with no allow-list entry. |
| MINOR | `test/continuous-test-suite-openai-compat-catalog.ts:1179` | `REPO_ROOT = process.cwd()` couples the new helpers to the launch directory. | **Open, non-blocking** — see inline comment; suggest deriving the repo root from `import.meta.url` instead of `process.cwd()`. |

### Checked and found clean

- **Loader parity** — `schema.ts` `derivedFallback` reproduces `loader.ts`'s positional derivation exactly (`fallbackModelName ?? fallbacks[1] ?? fallbacks[0]`), so the check reports what the runtime actually selects.
- **Blast radius** — `schema.ts` is imported only by the three `tools/` build/CI paths (`codegen-catalog.ts`, `check-model-liveness.ts`, `verify-provider-onboarding.ts`); confirmed via the code graph it reaches no SDK runtime surface. No unmodified runtime callers break (rule 5 safe).
- **Rule correctness** — `selectableModels > 1` gate correctly exempts the two single-model shipped catalogs; mistral's derived fallback `mistral-medium-latest` ≠ default, so the rule no longer fires on it.
- **Enforcement** — reached from both required CI jobs (`codegen:catalog --check` and `verify:provider-onboarding`, which imports `parseProviderCatalogJson`).
- **Test integrity** — the new cases' subprocess assertions (`/cannot fall back/`, `cerebras\.json` naming, non-zero/zero exit) are well-founded on the tool's actual error formatting; preconditions fail loudly rather than passing on coincidence.
- No generated artifacts change, no secrets in the diff, single-commit + conventional-commit satisfied (per CI bot).

### Note on prior markers
The prior round's summaries (marker `<!-- yama:summary -->`) from the earlier reviewer remain on the PR; I have no edit/delete tooling here, so this single comment is this round's authoritative record and the above table supersedes them. All three earlier findings are confirmed resolved, so the standing decision is unchanged: mergeable.

src/lib/providers/catalog/schema.ts
Tara-ag: Accepting — the option-(b) test landed and covers both halves of the rule end-to-end, exactly as agreed:

- **Rejection half** — `catalogFallbackRule` mutates `cerebras.json`'s `fallbackModelName` to its own `default` (two selectable models, subject to the rule), runs the built `codegen-catalog.ts --check` as a subprocess against a temp copy, and asserts a non-zero exit *plus* the rule's own "cannot fall back" text naming `cerebras.json`. The precondition guard (`clean` must not already trip the rule) prevents the case passing on a coincidence.
- **Exemption half** — asserts that shipped single-model catalogs (`gmicloud`, `inception-labs`) resolve their own default with `selectable === 1`, then requires the real tree to check clean. This locks the `selectableModels > 1` gate against a future inversion/removal — the exact regression the finding called out.

The suite imports nothing from `src/`, so `neurolink/e2e-tests-only` holds without an allow-list entry, and it's already wired into `extended-suites`. Resolving this thread.

Tara-ag:
## Summary — PR #1690 `fix(catalog): reject a fallback model that cannot fall back`

**Verdict: APPROVE** — all prior findings are resolved in the current HEAD (`4b79812`), including the cwd-coupling MINOR which I just resolved on-thread after confirming the fix is in the code. No new issues found in this pass.

### Findings — all resolved

| # | Severity | Location | Finding | Status |
| 1 | MAJOR | `src/lib/providers/catalog/mistral.json:13` | Restored fallback targeted plan-gated `mistral-large-latest` (403 on non-premium). | **Resolved** — `fallbacks[1]` → `mistral-medium-latest`, self-consistent with this catalog's own `evidence.liveMatrix` (already uses that id to dodge the same gate via the `visionModel` pin). |
| 2 | MINOR | `src/lib/providers/catalog/schema.ts` | Exemption + emitted count used whole-`catalog` map size (incl. `retired`/embed/OCR). | **Resolved** — `selectableModels` now filters `status !== "retired"` for both the gate and the message ("N selectable models"). |
| 3 | MINOR | `src/lib/providers/catalog/schema.ts` | No direct coverage of the new branch (esp. the single-model exemption). | **Resolved** — `testCatalogFallbackRule` runs the real `tools/codegen-catalog.ts --check` as a subprocess against a temp catalog tree: asserts multi-model rejection (cerebras fixture, non-zero exit + rule text naming the file) and single-model exemption (gmicloud/inception precondition on the shipped tree). No `src/` import → `neurolink/e2e-tests-only` holds without an allow-list entry. |
| 4 | MINOR | `test/continuous-test-suite-openai-compat-catalog.ts` | `REPO_ROOT = process.cwd()` coupled the new helpers to the launch dir. | **Resolved (this pass)** — `REPO_ROOT` now derives from `import.meta.url` (`resolve(dirname(fileURLToPath(import.meta.url)), "..")`), exactly the suggested fix. Thread resolved. |

### Checked and found clean

- **Loader parity** — the schema's `derivedFallback` (`fallbackModelName ?? fallbacks[1] ?? fallbacks[0]`) reproduces `loader.ts`'s positional derivation verbatim, so the check reports what the runtime actually selects (rule 3 / correctness).
- **Blast radius / rule 5** — `schema.ts` is imported only by the three `tools/` CI paths (`codegen-catalog.ts`, `check-model-liveness.ts`, `verify-provider-onboarding.ts`); via the code graph it reaches **no** SDK runtime surface. No unmodified SDK caller breaks.
- **Exemption correctness** — `selectableModels > 1` keeps the two single-model catalogs (`gmicloud`, `inception-labs`) passing; the shipped-tree check in the test locks this against an inverted/removed gate.
- **Enforcement** — reached from both required CI jobs (`codegen:catalog --check` and `verify:provider-onboarding`).
- **Correctness of rule** — mistral's derived fallback `mistral-medium-latest` ≠ default, so the rule no longer fires on it.
- No generated artifacts change (fallback resolved at load time), no secrets in the diff, conventional-commit + single-commit policy satisfied (per CI bot).

### Note on prior markers
The earlier round summaries on this PR (marker `<!-- yama:summary -->`) are superseded by this one; this is the authoritative final record. All four prior findings are confirmed resolved, so the standing decision is **mergeable**.

PR #1691, opened 2026-09-12
fix(openai): stop rejecting schemas that carry an optional field

murdore:
`generate({ schema })` is documented to return valid JSON plus a parsed `structuredData` for every provider. On OpenAI it did the opposite for the most ordinary input there is: **a plain JSON Schema failed the request outright with a 400.**

## The bug

`v3ResponseFormatToOpenAI` sent `strict: true` with the caller's schema untouched. OpenAI's strict structured-output mode rejects a schema unless every object node carries `additionalProperties: false` **and** lists every one of its properties in `required` — recursively, including through array `items`. A plain JSON Schema satisfies neither.

## The contract, learned from the vendor

Each rule confirmed by sending the shape and reading the refusal, not assumed:

| schema | result |
|---|---|
| bare object | `400` `'additionalProperties' is required` |
| `+ additionalProperties:false` | `200` OK |
| `required` omits one property | `400` `'required' is required to be supplied` |
| nested object without `AP:false` | `400` `context=('properties','city')` |
| array `items` without `AP:false` | `400` `context=('properties','rows','items')` |

## Why this fix does not just force `required`

Adding `additionalProperties: false` is safe — it forbids keys the caller never asked for, which strict mode forbids anyway.

Filling in `required` is **not** safe: it would silently make the caller's optional fields mandatory. So a schema that still has optional properties after normalisation drops to `strict: false`, which OpenAI accepts and which honours optionality — confirmed live, it returned `{"capital":"Paris"}` with `population` absent. Schemas that are already strict-compatible keep the stronger `strict: true` guarantee.

## How it was found

A measured feature matrix across 16 providers showed `json` failing while `toolsJson` passed **on the same rows**. That contradiction pointed straight at the cause: the repo already has `fixSchemaForOpenAIStrictMode`, but it is wired only into tool-parameter schemas in `ToolsManager` — never into the caller's `schema`.

## Verification

Four cases against live OpenAI on the built package:

```
plain schema (was a hard 400)     OK  {"capital":"Paris","population":2140526}
optional field stays optional     OK  {"capital":"Paris","population":2148000}
already strict-compatible         OK  {"capital":"Paris","population":2140526}
nested object                     OK  {"city":{"name":"Paris"}}
```

The first version of this fix recursed into `properties` as though the map were itself a schema node, so nested objects were never closed and the nested case still failed live. Fixed by treating `properties`/`$defs`/`definitions` as maps of schemas and `items`/`anyOf`/`oneOf`/`allOf`/`not` as nodes.

Two live regression tests added to the OpenAI cell of `continuous-test-suite-json-e2e`. Their assertion messages never quote the payload, so a real break cannot be downgraded to a skip by the harness classifier.

`pnpm run pre-push`: 14 passed.

## Note for reviewers

That suite sets `NEUROLINK_DISABLE_BUILTIN_TOOLS=true` at the top, so it does not exercise the default configuration where built-in tools are active. Worth a separate look — it is why this never showed up there.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved structured JSON responses for OpenAI-compatible providers by automatically closing object schemas when supported.
  - Preserved optional fields without incorrectly requiring them.
  - Applied strict validation only when schemas are fully compatible, improving handling of complex schemas and unspecified additional properties.
  - Improved compatibility with Azure structured outputs and reduced request rejections for valid schemas.

### DIFF
--- a/src/lib/providers/openaiChatCompletionsClient.ts
+++ b/src/lib/providers/openaiChatCompletionsClient.ts
@@ -475,6 +475,138 @@ export const v3ToolChoiceToOpenAI = (
   }
 };
 
+/**
+ * OpenAI's strict structured-output mode rejects a schema unless every object
+ * node carries `additionalProperties: false` AND lists every one of its
+ * properties in `required` — recursively, including through array `items`.
+ * A plain JSON Schema satisfies neither, so sending one with `strict: true`
+ * fails the request outright rather than degrading.
+ *
+ * Adding `additionalProperties: false` is safe: it forbids keys the caller
+ * never asked for, which strict mode would forbid anyway. Filling in
+ * `required` is NOT safe — it would silently make the caller's optional
+ * fields mandatory. So when a schema still has optional properties after
+ * normalisation, the request drops to `strict: false`, which OpenAI accepts
+ * and which honours optionality. Callers whose schemas are already strict-
+ * compatible keep the stronger guarantee.
+ */
+// `properties` and `$defs` are MAPS of schemas, not schemas — recursing into
+// them as if they were nodes silently skips every child, which is exactly the
+// bug that let a nested object through without `additionalProperties: false`.
+const SCHEMA_MAPS = [
+  "properties",
+  "patternProperties",
+  "$defs",
+  "definitions",
+] as const;
+const SCHEMA_NODES = [
+  "items",
+  "prefixItems",
+  "anyOf",
+  "oneOf",
+  "allOf",
+  "not",
+  "then",
+  "else",
+] as const;
+
+// OpenAI's strict mode does not accept these composition keywords. A schema
+// carrying one cannot be sent with `strict: true` at all, so it is not merely
+// "not yet normalised" — it must drop to non-strict, where the schema is
+// honoured as written.
+const STRICT_UNSUPPORTED = ["allOf", "not", "if", "then", "else"] as const;
+
+const mapValues = (
+  obj: unknown,
+  fn: (v: unknown) => unknown,
+): Record<string, unknown> =>
+  Object.fromEntries(
+    Object.entries((obj ?? {}) as Record<string, unknown>).map(([k, v]) => [
+      k,
+      fn(v),
+    ]),
+  );
+
+const withClosedObjects = (node: unknown): unknown => {
+  if (Array.isArray(node)) {
+    return node.map(withClosedObjects);
+  }
+  if (!node || typeof node !== "object") {
+    return node;
+  }
+  const next: Record<string, unknown> = {
+    ...(node as Record<string, unknown>),
+  };
+  for (const key of SCHEMA_MAPS) {
+    if (key in next) {
+      next[key] = mapValues(next[key], withClosedObjects);
+    }
+  }
+  for (const key of SCHEMA_NODES) {
+    if (key in next) {
+      next[key] = withClosedObjects(next[key]);
+    }
+  }
+  // A schema-valued `additionalProperties` is itself a schema (the index/value
+  // pattern) and its nested objects need closing too. A boolean one is a flag
+  // and must be left exactly as the caller wrote it.
+  if (
+    next.additionalProperties &&
+    typeof next.additionalProperties === "object"
+  ) {
+    next.additionalProperties = withClosedObjects(next.additionalProperties);
+  } else if (next.type === "object" && !("additionalProperties" in next)) {
+    // Closed whether or not it declares `properties`: strict mode requires the
+    // key on EVERY object, including an empty one.
+    next.additionalProperties = false;
+  }
+  return next;
+};
+
+/**
+ * True when the schema can legally be sent with `strict: true`: every object
+ * node lists all of its properties as required, every object is closed, and
+ * no composition keyword OpenAI rejects appears anywhere.
+ *
+ * Deliberately conservative — a false negative costs only the stronger
+ * guarantee, while a false positive costs the whole request.
+ */
+const satisfiesStrictRequired = (node: unknown): boolean => {
+  if (Array.isArray(node)) {
+    return node.every(satisfiesStrictRequired);
+  }
+  if (!node || typeof node !== "object") {
+    return true;
+  }
+  const rec = node as Record<string, unknown>;
+  if (STRICT_UNSUPPORTED.some((k) => k in rec)) {
+    return false;
+  }
+  if (rec.type === "object") {
+    if (rec.additionalProperties !== false) {
+      return false;
+    }
+    const names = Object.keys(
+      (rec.properties ?? {}) as Record<string, unknown>,
+    );
+    const required = Array.isArray(rec.required)
+      ? (rec.required as unknown[])
+      : [];
+    if (names.some((n) => !required.includes(n))) {
+      return false;
+    }
+  }
+  const mapsOk = SCHEMA_MAPS.filter((k) => k in rec).every((k) =>
+    Object.values((rec[k] ?? {}) as Record<string, unknown>).every(
+      satisfiesStrictRequired,
+    ),
+  );
+  const nodesOk = SCHEMA_NODES.filter((k) => k in rec).every((k) =>
+    satisfiesStrictRequired(rec[k]),
+  );
+  return mapsOk && nodesOk;
+};
+
 export const v3ResponseFormatToOpenAI = (rf: {
   type: "text" | "json";
   schema?: Record<string, unknown>;
@@ -487,13 +619,38 @@ export const v3ResponseFormatToOpenAI = (rf: {
   if (!rf.schema) {
     return { type: "json_object" };
   }
+  // Mutate as little as possible, in this order:
+  //
+  //   1. already strict-legal  -> send it UNTOUCHED with strict: true
+  //   2. legal once closed     -> send the closed copy with strict: true
+  //   3. neither               -> send it UNTOUCHED with strict: false
+  //
+  // Case 3 is why closure is not applied unconditionally. Non-strict mode
+  // honours the schema exactly as written, so injecting
+  // `additionalProperties: false` there would silently change the caller's
+  // contract — and for a composition it can make the schema unsatisfiable:
+  // closing two `allOf` members that declare different properties leaves no
+  // object able to satisfy both. Leaving case 3 untouched also means
+  // non-OpenAI endpoints in this family, which may not implement OpenAI's
+  // strict contract at all, see exactly the schema the caller wrote.
+  //
+  // Case 1 matters for parity: a Zod schema whose properties are all required
+  // already converts to a strict-legal shape, so it goes out byte-identical to
+  // what it did before this change.
+  const original = rf.schema;
+  const closed = withClosedObjects(original) as Record<string, unknown>;
+  const schema = satisfiesStrictRequired(original)
+    ? original
+    : satisfiesStrictRequired(closed)
+      ? closed
+      : original;
   return {
     type: "json_schema",
     json_schema: {
       name: rf.name ?? "response",
-      schema: rf.schema as never,
+      schema: schema as never,
       ...(rf.description ? { description: rf.description } : {}),
-      strict: true,
+      strict: satisfiesStrictRequired(schema),
     },
   };
 };
--- a/test/continuous-test-suite-json-e2e.ts
+++ b/test/continuous-test-suite-json-e2e.ts
@@ -576,4 +576,100 @@ await test("anthropic:claude-sonnet-4-6 — forced truncation yields a partial o
   );
 });
 
+// A Zod schema with ANY optional field converts to `required` missing that
+// property while `additionalProperties: false` is already present — a shape
+// OpenAI's strict mode rejects outright with "'required' is required to be
+// supplied". So `generate({ schema })` failed with a 400, not a degraded
+// answer, for the most ordinary optional field there is.
+//
+// Forcing every property into `required` would satisfy strict mode but would
+// silently make the caller's optional fields mandatory. The request must drop
+// to non-strict instead, which OpenAI accepts and which honours optionality.
+//
+// The absence assertion is the point: without it, an implementation that
+// regressed to forcing `required` would still pass, because the model would
+// return both fields and every other assertion would hold.
+await test("openai:gpt-4o-mini — an optional Zod field stays optional, and is absent when unset", async () => {
+  let res: SchemaResult;
+  try {
+    res = (await nl.generate({
+      input: {
+        text: "Name the capital of France. Do not include a population figure.",
+      },
+      provider: "openai",
+      model: "gpt-4o-mini",
+      maxTokens: 200,
+      schema: z.object({
+        capital: z.string(),
+        population: z.number().optional(),
+      }),
+    })) as SchemaResult;
+  } catch (e) {
+    if (isInfraError(String((e as Error)?.message ?? e))) {
+      throw new Skip("openai unavailable for provider reasons");
+    }
+    throw new Error(
+      "a schema with an optional property was rejected outright",
+      { cause: e },
+    );
+  }
+  const data = res.structuredData as Record<string, unknown> | undefined;
+  assert(
+    !!data && typeof data === "object" && !Array.isArray(data),
+    "structuredData must be a plain object",
+  );
+  assert(
+    typeof data?.capital === "string" && (data.capital as string).length > 0,
+    "the required property must be present and non-empty",
+  );
+  assert(
+    data !== undefined && !("population" in data),
+    "the optional property must be absent, proving it was not forced into required",
+  );
+  console.log(
+    "      · optional Zod field honoured and absent — request not rejected",
+  );
+});
+
+// The same shared path carries the whole OpenAI-compatible family, so pin a
+// second member of it. Azure is the right second cell here: it is the other
+// provider that opts OUT of suppressing response_format with tools, so it
+// actually exercises the `json_schema` + strict decision this change makes,
+// rather than a provider whose format is downgraded or suppressed before the
+// decision matters.
+await test("azure — an optional field survives on a second family member", async () => {
+  const azureModel = process.env.AZURE_OPENAI_MODEL ?? process.env.AZURE_MODEL;
+  if (!azureModel) {
+    throw new Skip("no azure deployment configured");
+  }
+  let res: SchemaResult;
+  try {
+    res = (await nl.generate({
+      input: {
+        text: "Name the capital of France. Do not include a population figure.",
+      },
+      provider: "azure",
+      model: azureModel,
+      maxTokens: 200,
+      schema: z.object({
+        capital: z.string(),
+        population: z.number().optional(),
+      }),
+    })) as SchemaResult;
+  } catch (e) {
+    if (isInfraError(String((e as Error)?.message ?? e))) {
+      throw new Skip("azure unavailable for provider reasons");
+    }
+    throw new Error("optional-field schema failed on a second provider", {
+      cause: e,
+    });
+  }
+  const data = res.structuredData as Record<string, unknown> | undefined;
+  assert(
+    !!data && typeof data === "object",
+    "structuredData must be produced on the shared family path too",
+  );
+  console.log("      · optional field honoured on a second family member");
+});
+
 await runSuite();


src/lib/providers/openaiChatCompletionsClient.ts
```diff
@@ -487,13 +582,14 @@ export const v3ResponseFormatToOpenAI = (rf: {
   if (!rf.schema) {
     return { type: "json_object" };
   }
+  const schema = withClosedObjects(rf.schema) as Record<string, unknown>;
   return {
     type: "json_schema",
     json_schema: {
       name: rf.name ?? "response",
-      schema: rf.schema as never,
+      schema: schema as never,
```
Tara-ag, line 651: **MAJOR — blast radius across the whole OpenAI-compatible family, and only the tools-disabled path is tested.**

`v3ResponseFormatToOpenAI` is reached only through `buildDelegatingModel` (openaiChatCompletionsBase.ts:743), which every provider in the OpenAI-compatible family rides — OpenAI, Azure OpenAI, DeepSeek, OpenRouter, Ollama, LlamaCpp, NVIDIA NIM, LM Studio, LiteLLM, openaiCompatible, Cohere. Before this change every one of them got `strict: true` + the caller's raw schema; now each gets `additionalProperties: false` injected and a per-schema computed `strict`. Adding `additionalProperties: false` changes validation semantics for non-OpenAI endpoints that may not implement OpenAI's strict contract, and `strict: false` is a behavioural downgrade (schema no longer enforced) that previously never happened on these paths.

The only coverage (`test/continuous-test-suite-json-e2e.ts`) targets `gpt-4o-mini` and sets `NEUROLINK_DISABLE_BUILTIN_TOOLS=true` at the top (the author's own "Note for reviewers"), so the default tools-active configuration and every non-OpenAI provider are unverified. Per repo hot-path convention, this broadened/shared surface needs at least one tools-active regression case and ideally one non-OpenAI OpenAI-compatible case.

```suggestion
      strict: satisfiesStrictRequired(schema),
```

src/lib/providers/openaiChatCompletionsClient.ts
```diff
@@ -475,6 +475,101 @@ export const v3ToolChoiceToOpenAI = (
...
+ * `required` is NOT safe — it would silently make the caller's optional
+ * fields mandatory. So when a schema still has optional properties after
+ * normalisation, the request drops to `strict: false`, which OpenAI accepts
+ * and which honours optionality. Callers whose schemas are already strict-
+ * compatible keep the stronger guarantee.
+ */
+// `properties` and `$defs` are MAPS of schemas, not schemas — recursing into
+// them as if they were nodes silently skips every child, which is exactly the
+// bug that let a nested object through without `additionalProperties: false`.
+const SCHEMA_MAPS = ["properties", "$defs", "definitions"] as const;
```
Tara-ag: **MINOR — normalisation skips strict-relevant JSON Schema constructs, so some schemas can still 400.**

`withClosedObjects` walks only `properties`/`$defs`/`definitions` (maps) and `items`/`anyOf`/`oneOf`/`allOf`/`not` (nodes), and closes an object only when it has a truthy `properties`. Three constructs the OpenAI strict contract cares about are missed:

- `prefixItems` (draft 2020-12 tuple arrays) is never walked, so nested objects in a tuple `items` are not closed.
- a schema-valued `additionalProperties` (an object used as an index/value pattern) is never recursed into.
- an object node with `type: "object"` and **no** `properties` (free-form object) is neither closed here nor flagged in `satisfiesStrictRequired`, so it still 400s under strict mode.

These are edge schemas, but the PR documents its recursion as exhaustive. Consider folding the keyword set and treating a `type:"object"` node without `properties`/`additionalProperties` as needing closure.

```suggestion
const SCHEMA_MAPS = ["properties", "$defs", "definitions"] as const;
const SCHEMA_NODES = ["items", "prefixItems", "anyOf", "oneOf", "allOf", "not", "patternProperties"] as const;
```

test/continuous-test-suite-json-e2e.ts
```diff
@@ -576,4 +576,91 @@ await test("anthropic:claude-sonnet-4-6 — forced truncation yields a partial o
...
+    throw new Error(
+      "a schema with an optional property was rejected outright",
+      { cause: e },
+    );
+  }
+  const data = res.structuredData as Record<string, unknown> | undefined;
+  assert(
+    !!data && typeof data === "object",
+    "structuredData must be present when a property is optional",
+  );
```
Tara-ag: **MAJOR — the "optional property stays optional" test never verifies optionality.**

This test asserts only that `structuredData` is an object and `capital` is a string. It never checks that the optional field `population` is absent. If the implementation regressed to the unsafe behaviour the PR explicitly rejects — forcing every property into `required` and keeping `strict: true` — the model returning both `capital` and `population` would still satisfy these assertions, so the test would pass despite violating the exact guarantee it claims to cover.

Add an assertion that the optional field is absent (IPC-safe and not quoted in the message, per the harness rules):

```suggestion
  const data = res.structuredData as Record<string, unknown> | undefined;
  assert(
    !!data && typeof data === "object",
    "structuredData must be present when a property is optional",
  );
  assert(
    typeof data?.capital === "string",
    "the one required property must still be produced",
  );
  assert(
    !("population" in (data ?? {})) || data?.population === undefined,
    "the optional property must stay optional, not become mandatory",
  );
  console.log("      · optional property preserved, request not rejected");
```

Tara-ag (commented):
Reviewed the schema-normalisation change in v3ResponseFormatToOpenAI and its two new recursive helpers, plus the two e2e tests. See inline comments. Overall the fix is sound in direction (close objects, preserve optionality by dropping strict), but coverage of the broad blast radius and the optionality guarantee need tightening.

Tara-ag:
## Verdict: **NEEDS_WORK**

Direction is correct and well-documented — closing object schemas with `additionalProperties: false` and preserving optionality by dropping `strict` to `false` actually fixes the 400 on plain JSON Schemas for gpt-4o-mini. But the change rides a very broad, shared hot path and its two central guarantees (strictness for the rest of the family, optionality) are under-tested.

|---|---|---|
| MAJOR | `src/lib/providers/openaiChatCompletionsClient.ts:590` | Blast radius across every OpenAI-compatible provider via `buildDelegatingModel`; only gpt-4o-mini with built-in tools disabled is tested |
| MINOR | `src/lib/providers/openaiChatCompletionsClient.ts:496` | Normalisation skips `prefixItems`/`patternProperties`/schema-valued `additionalProperties`; a `type:"object"` node without `properties` is neither closed nor flagged |
| MAJOR | `test/continuous-test-suite-json-e2e.ts:658` | "optional property stays optional" never asserts the optional field is absent, so a regression to all-required passes |

**Checked and clean**
- No mutation of the caller's `rf.schema`: `withClosedObjects` builds fresh objects (`{...node}` + `mapValues`), so shared/caller-owned schema objects are not aliased or modified. Verified against the recursion (maps vs nodes) the PR describes.
- Recursion correctly descends through `items`/`anyOf`/`oneOf`/`allOf`/`not` and arrays; `satisfiesStrictRequired` consistently conservatively flags any object with an unmet-required property so `strict` only stays `true` when safe.
- Added `additionalProperties: false` is applied only where absent, don't overwrite an explicit caller value.
- Rules 1/4/6/15: no registry/dynamic-import changes, CLI/SDK separation unaffected, no `formatProviderError` change, tests are e2e (single graph). Secret-scanning of new code shows no credentials.
- No prior review threads or Yama findings on this PR to update/resolve — this is the first review pass.

No CRITICAL/security findings.

Tara-ag (commented):
Review of the OpenAI strict-schema compatibility change.

src/lib/providers/openaiChatCompletionsClient.ts
```diff
@@ -487,13 +582,14 @@ export const v3ResponseFormatToOpenAI = (rf: {
...
+  const schema = withClosedObjects(rf.schema) as Record<string, unknown>;
   return {
     type: "json_schema",
     json_schema: {
       name: rf.name ?? "response",
-      schema: rf.schema as never,
+      schema: schema as never,
       ...(rf.description ? { description: rf.description } : {}),
-      strict: true,
+      strict: satisfiesStrictRequired(schema),
```
Tara-ag, line 653: **MAJOR — this gates structured output for every OpenAI-compatible provider, and the new behaviour is riskier than the old.**

`v3ResponseFormatToOpenAI` has exactly one caller — `buildDelegatingModel` (`openaiChatCompletionsBase.ts:743`), the shared delegating-model constructor for the entire OpenAI-compatible provider family. So this single computation, `strict: satisfiesStrictRequired(schema)`, controls structured output for every provider in that family.

The change from `strict: true` to a runtime-computed boolean means:
- A plain schema is now **mutated** (extra `additionalProperties: false` on every object node) before the request goes out.
- Whether the request is strict or not depends on a bespoke schema walker.

The only tests added are in `continuous-test-suite-json-e2e.ts`, which runs with `NEUROLINK_DISABLE_BUILTIN_TOOLS=true` — your own "Note for reviewers" admits the default configuration with built-in tools is **not** exercised, and that is exactly the path shifting here. The graph confirms the reach: this is a fork on the shared provider path (medium-risk hub).

No code change is needed on this line — the shift is sound. The blocker is the coverage gap: please add a check that the default configuration (built-in tools active) still produces the correct `strict`/mutation behaviour before merge.

src/lib/providers/openaiChatCompletionsClient.ts
```diff
@@ -475,6 +475,101 @@ export const v3ToolChoiceToOpenAI = (
...
+  }
+  for (const key of SCHEMA_NODES) {
+    if (key in next) {
+      next[key] = withClosedObjects(next[key]);
+    }
+  }
+  if (
+    next.type === "object" &&
+    next.properties &&
+    !("additionalProperties" in next)
```
Tara-ag: **MINOR — the normaliser misses several JSON-Schema constructs, so some schemas can still 400 in strict mode.**

`withClosedObjects` only closes nodes that reach the `type === "object" && properties` branch. These shapes are missed, so even when `strict` computes to `true` a caller using any of them can still get OpenAI's `'additionalProperties' is required` refusal:

- `prefixItems` (draft-2020-12 tuple positionals) — not in `SCHEMA_NODES`; strict mode validates array positions, so a tuple schema that resolves as strict can still be rejected.
- `patternProperties` and `additionalProperties` given as a *schema* — never recursed into (only the boolean `false` is handled, and only when already present).
- A `type: "object"` node with no `properties` key (e.g. `{ type: "object" }`): the `next.properties &&` guard skips adding `additionalProperties: false`, and OpenAI may still require it.

Each is a small additive case. Recommend covering them in `SCHEMA_NODES` / the closing condition, and adding one of these shapes to the e2e suite.

```suggestion
const SCHEMA_NODES = [
  "items",
  "prefixItems",
  "anyOf",
  "oneOf",
  "allOf",
  "not",
  "patternProperties",
  "additionalProperties",
] as const;
```

test/continuous-test-suite-json-e2e.ts
```diff
@@ -576,4 +576,91 @@ await test("anthropic:claude-sonnet-4-6 — forced truncation yields a partial o
...
+  assert(
+    typeof data?.capital === "string",
```
Tara-ag: **MAJOR — this test would pass even if the optional field were dropped; it never asserts the behaviour it claims to protect.**

The test is titled "an optional property stays optional", but it only asserts that `structuredData` is present and that `capital` (the required field) is returned. It never checks `population` (the optional field) — so a regression that silently swallowed optional properties (e.g. a future change that forces all `properties` into `required` to re-enable `strict: true`) would still pass green.

The intent of the whole PR is that optionality is preserved, so assert on the absence of the optional field. OpenAI can omit it entirely or produce `null`, so allow both:

```suggestion
  assert(
    typeof data?.capital === "string",
    "the one required property must still be produced",
  );
  assert(
    !("population" in data!) || data.population === null,
    "the optional property must not be forced to always appear",
  );
```

Tara-ag (changes requested):
NEEDS_WORK — the schema-normalisation approach is sound, but two MAJOR gaps remain: the shared provider path lacks coverage on the default (built-in tools) configuration, and the "optional stays optional" test never asserts the optional field is actually absent. One MINOR normalisation-completeness note. Details in the inline comments.

Tara-ag:
## NEEDS_WORK

The fix is directionally correct and well-reasoned — computing `strict` from the normalised schema (instead of forcing `strict: true`) so optional fields stay optional is the right call, and the `withClosedObjects`/`satisfiesStrictRequired` split is clear. But two MAJOR findings block merge, and one MINOR completeness gap should be addressed.

|---|---|---|
| MAJOR | `src/lib/providers/openaiChatCompletionsClient.ts:592` | This gates structured output for **every** OpenAI-compatible provider (sole caller `buildDelegatingModel`, `openaiChatCompletionsBase.ts:743`), yet the new behaviour has no coverage on the default configuration — tests run with `NEUROLINK_DISABLE_BUILTIN_TOOLS=true`. |
| MINOR | `src/lib/providers/openaiChatCompletionsClient.ts:533` | `withClosedObjects` misses `prefixItems`, schema-valued `additionalProperties`/`patternProperties`, and `type:"object"` nodes without `properties` — those schemas can still 400 in strict mode. |
| MAJOR | `test/continuous-test-suite-json-e2e.ts:660` | The "optional property stays optional" test never asserts the optional field is absent — a regression that swallows optional fields would pass green. |

### Checked clean
- Blast radius traced via the code graph: `v3ResponseFormatToOpenAI` has a single caller (`buildDelegatingModel` at `openaiChatCompletionsBase.ts:743`), so no unmodified out-of-diff call sites are overridden.
- The normaliser builds fresh objects (`next = { ...node }` + `mapValues`), so it does not alias/mutate the caller's schema object in-place — no cross-call state leakage.
- `additionalProperties: false` injection is semantically safe: it forbids keys strict mode forbids anyway; optionality is correctly preserved by dropping to `strict: false` rather than padding `required`.
- No secrets, no hardcoded credentials, no logging of model params introduced.
- Commit is a single Conventional Commit (`fix(openai): …`) — compliant.

src/lib/providers/openaiChatCompletionsClient.ts
Tara-ag: **MINOR — the normaliser misses several JSON-Schema constructs, so some schemas can still 400 in strict mode.**

`withClosedObjects` only closes nodes that reach the `type === "object" && properties` branch. These shapes are missed, so even when `strict` computes to `true` a caller using them can still get OpenAI's `'additionalProperties' is required` refusal:

- `prefixItems` (draft-2020-12 tuple positionals) — not in `SCHEMA_NODES`; strict mode validates array positions, so a tuple schema that resolves as strict can still be rejected.
- `patternProperties` and `additionalProperties` given as a *schema* — never recursed into (only the boolean `false` is handled, and only when already present).
- A `type: "object"` node with no `properties` key: the `next.properties &&` guard skips adding `additionalProperties: false`, and OpenAI may still require it.

Each is a small additive case. Fold them into the recursion and add one of these shapes to the e2e suite. Note: per OpenAI's contract, schemas containing `allOf`/`not` cannot hold `strict: true` — see the CodeRabbit threads on those keywords.

```suggestion
const SCHEMA_NODES = [
  "items",
  "prefixItems",
  "anyOf",
  "oneOf",
  "allOf",
  "not",
  "patternProperties",
  "additionalProperties",
] as const;
```

test/continuous-test-suite-json-e2e.ts
```diff
@@ -576,4 +576,91 @@ await test("anthropic:claude-sonnet-4-6 — forced truncation yields a partial o
...
+    "the one required property must still be produced",
+  );
```
Tara-ag: **MAJOR — this test would pass even if the optional field were dropped; it never asserts the behaviour it claims to protect.**

The test is titled "an optional property stays optional", but it only asserts that `structuredData` is present and that `capital` (the required field) is returned. It never checks `population` (the optional field) — so a regression that silently swallowed optional properties (e.g. a future change that forces all `properties` into `required` to re-enable `strict: true`) would still pass green.

The intent of the whole PR is that optionality is preserved, so assert on the absence of the optional field. OpenAI can omit it entirely or produce `null`/`undefined`, so allow all three:

```suggestion
  assert(
    typeof data?.capital === "string",
    "the one required property must still be produced",
  );
  assert(
    !("population" in data!) || data.population === undefined,
    "the optional property must not be forced to always appear",
  );
```

Tara-ag (changes requested):
NEEDS_WORK — re-submitting with corrected suggestion anchors for two of the three findings (blast-radius + coverage + optional-field). Verdict unchanged: two MAJOR gaps (default-config coverage, optionality assertion) and one MINOR normalisation-completeness note block merge.

Tara-ag:
**Note on duplicate summary (for maintainers/cleanup):** two `<!-- yama:summary -->` comments exist on this PR (`5645529298` and `5645538551`). The canonical summary is `5645538551`; the earlier `5645529298` is superseded — it references the three inline threads that were resolved as duplicates of the same findings. Please delete `5645529298` when convenient (the automation account cannot delete issue comments). The active review state is otherwise clean: 3 findings (one inline comment each), review = **changes-requested**.

src/lib/providers/openaiChatCompletionsClient.ts
```diff
@@ -475,6 +475,133 @@ export const v3ToolChoiceToOpenAI = (
...
+// `properties` and `$defs` are MAPS of schemas, not schemas — recursing into
+// them as if they were nodes silently skips every child, which is exactly the
+// bug that let a nested object through without `additionalProperties: false`.
+const SCHEMA_MAPS = ["properties", "$defs", "definitions"] as const;
+const SCHEMA_NODES = [
+  "items",
+  "prefixItems",
+  "anyOf",
+  "oneOf",
+  "allOf",
```
Tara-ag, line 507: **MAJOR — composition keywords are closed *and* dropped to non-strict, so `additionalProperties:false` is injected into schemas the PR claims to "honour as written".**

`allOf` (and `then`/`else`) are in `SCHEMA_NODES`, so `withClosedObjects` recurses into their members and adds `additionalProperties: false` to every `type:"object"` member (lines 549–558). At the same time `satisfiesStrictRequired` sees `allOf`/`then`/`else` in `STRICT_UNSUPPORTED` and returns `false`, so the request is sent **non-strict**. Non-strict mode honours the schema as written — including the injected closures.

For a composition whose members declare *different* properties this makes the intersection unsatisfiable, silently producing a broken/no-valid-value schema instead of the contract the PR states:

```jsonc
{
  "allOf": [
    { "type": "object", "properties": { "a": { "type": "string" } }, "required": ["a"] },
    { "type": "object", "properties": { "b": { "type": "string" } }, "required": ["b"] }
  ]
}
```

`withClosedObjects` turns member 1 into "only `a` allowed" and member 2 into "only `b` allowed"; `allOf` needs both → **no value satisfies it**. Previously such a schema got a loud `400` (strict rejects `allOf`); now it silently goes out corrupted because closure injection is applied regardless of the strict decision that follows.

The optional-field path is unaffected (a flat closed object with an optional key is semantically fine) — this is specific to composition. Cheapest fix: drop `allOf`/`then`/`else` from `SCHEMA_NODES` (they are already gated by `STRICT_UNSUPPORTED` → non-strict → "honoured as written"), or skip closure injection when the schema will be sent non-strict.

```suggestion
const SCHEMA_NODES = [
  "items",
  "prefixItems",
  "anyOf",
  "oneOf",
  "not",
] as const;
```

Tara-ag (changes requested):
## NEEDS_WORK

This pass is against the reworked commit `4bc4a6a`. The fix is directionally right and the two blocker findings from the previous review are genuinely resolved; one new correctness finding remains on the shared hot path.

|---|---|---|
| MAJOR | `src/lib/providers/openaiChatCompletionsClient.ts:497` | Composition keywords (`allOf`/`then`/`else`) are closed (AP:false injected) **and** dropped to non-strict, so closures are honoured in non-strict and can make a multi-member intersection unsatisfiable — contradicts the PR's "honoured as written" contract. |

### Previously-raised findings — status on this commit
- **MAJOR "optional test weak" → RESOLVED.** The openai test now asserts the optional field is absent (`!("population" in data)`, line 626), so a regression that forces `required` can no longer pass green.
- **MINOR "normalisation completeness" → RESOLVED.** `prefixItems`, schema-valued `additionalProperties`, and `type:"object"`-without-`properties` are now handled; `STRICT_UNSUPPORTED` guards the composition keywords.
- **MAJOR "blast radius / coverage" → LARGELY ADDRESSED.** A second family member (Azure — the provider that actually exercises `response_format`+tools) is now covered, and the optional test catches the shared path's `strict` decision. Remaining gap (author's own "Note for reviewers"): the default built-in-tools-active config for `openai` main is still not live-tested; since openai main suppresses `response_format` when tools are active, that path is only germane for the tools-less / Azure cases now covered — worth one explicit note but no longer a blocker.

### Checked clean
- `withClosedObjects` builds fresh objects (`{...node}` + `mapValues`) — the caller's `rf.schema` is **not** mutated/aliased (verified against the recursion).
- `satisfiesStrictRequired` is conservative (returns `false` on any unmet-required or missed constraint), so `strict:true` is only kept when safe.
- No public SDK API change (Rule 5): `strict` is internal to the wire request; the `generate({schema})` surface is unchanged.
- No Rules 1/4/6/15 violations (no registry/CLI/error-format changes; tests remain single-module e2e). No secrets in the added code.
- Single Conventional Commit, valid scope — compliant.

No CRITICAL / security findings.

Tara-ag:
Recurring review — status update against reworked commit `4bc4a6a`:

- Two prior blockers are **resolved** in this commit: the optional-field test now asserts the field is absent (line 626), and normalisation now covers `prefixItems`, schema-valued `additionalProperties`, and objects without `properties`.
- The prior blast-radius/coverage finding is **largely addressed** by the Azure family-member cell (the provider that actually exercises the `json_schema`+strict path). 
- One **new MAJOR** finding remains (composition-member `additionalProperties:false` injection into `allOf`/`then`/`else` while dropping to non-strict → can make an intersection unsatisfiable), posted inline.

Full synthesis in the submitted review. Summary markers on this PR are duplicated (`5645529298` vs `5645538551`); the active verdict is **NEEDS_WORK** on the new finding.

src/lib/providers/openaiChatCompletionsClient.ts
```diff
@@ -475,6 +475,133 @@ export const v3ToolChoiceToOpenAI = (
...
+  "not",
+  "then",
+  "else",
+] as const;
```
Tara-ag, line 511: **MAJOR — composition keywords are closed *and* dropped to non-strict, so `additionalProperties: false` is injected into schemas the PR claims to "honour as written".**

`allOf` (and `then`/`else`) are in `SCHEMA_NODES`, so `withClosedObjects` recurses into their members and adds `additionalProperties: false` to every `type: "object"` member. At the same time `satisfiesStrictRequired` sees `allOf`/`then`/`else` in `STRICT_UNSUPPORTED` and returns `false`, so the request is sent **non-strict**. Non-strict mode honours the schema as written — including the injected closures.

For a composition whose members declare *different* properties this makes the intersection unsatisfiable, silently producing a broken/no-valid-value schema instead of the contract the PR states:

```jsonc
{
  "allOf": [
    { "type": "object", "properties": { "a": { "type": "string" } }, "required": ["a"] },
    { "type": "object", "properties": { "b": { "type": "string" } }, "required": ["b"] }
  ]
}
```

`withClosedObjects` turns member 1 into "only `a` allowed" and member 2 into "only `b` allowed"; `allOf` needs both → **no value satisfies it**. Previously such a schema got a loud `400` (strict rejects `allOf`); now it silently goes out corrupted because closure injection is applied regardless of the strict decision that follows.

The optional-field path is unaffected (a flat closed object with an optional key is semantically fine) — this is specific to composition. Cheapest fix: drop `allOf`/`then`/`else` from `SCHEMA_NODES` (they are already gated by `STRICT_UNSUPPORTED` → non-strict → "honoured as written"), or skip closure injection when the schema will be sent non-strict.

```suggestion
const SCHEMA_NODES = [
  "items",
  "prefixItems",
  "anyOf",
  "oneOf",
  "not",
] as const;
```

Tara-ag (changes requested):
Corrected anchor + suggestion block for the composition-closure finding. Verdict unchanged: NEEDS_WORK on the single MAJOR finding (allOf/then/else closure injection under non-strict).

murdore:
All four findings addressed, and the review changed my understanding of the defect — the PR is retitled and the commit rewritten as a result.

**The defect was mis-framed.** I originally reported it as "plain JSON Schema is rejected." Chasing the `test-shards (types)` failure showed why that framing was wrong: `ValidationSchema = ZodUnknownSchema | Schema<unknown>`, so a plain JSON Schema object is not a supported input at all — my tests were type errors, which is exactly what CI was telling me.

The real, type-supported defect is narrower and worse: `convertZodToJsonSchema` already emits `additionalProperties: false` recursively, but emits `required` listing only the non-optional properties. So **any Zod schema with an optional field** produced a shape strict mode rejects:

```
z.object({ capital: z.string(), population: z.number().optional() })
  -> { required: ["capital"], additionalProperties: false }
  -> 400 'required' is required to be supplied
```

Verified live: with the fix, `{"capital":"Paris"}` — the optional field comes back absent.

**MAJOR — blast radius.** Accepted, and stated explicitly in the commit now. `v3ResponseFormatToOpenAI` has one caller, `buildDelegatingModel`, which the whole family rides. The mitigating argument is that the change can only *relax* `strict` or add a key strict mode already implies — it cannot make a previously-succeeding request fail. For coverage I pinned **Azure** as the second cell rather than a suppressing provider: Azure is the other provider that opts out of `suppressResponseFormatWithTools`, so it actually exercises the strict decision instead of having the format downgraded before the decision matters.

**MAJOR — the optionality test never verified optionality.** Correct, and this was the most useful finding. The test now asserts `!("population" in data)`. You were right that without it, a regression to forcing `required` would still pass.

**MINOR — normalisation completeness.** All three gaps closed: `prefixItems` is walked, a schema-valued `additionalProperties` is recursed into (a boolean one is left exactly as written), and an object is closed whether or not it declares `properties`, since strict mode requires the key on every object including an empty one.

**CodeRabbit on `allOf`.** Confirmed and handled, but not by normalising — `allOf`, `not`, and `if`/`then`/`else` cannot be sent under strict at all, so their presence now forces `strict: false` rather than being silently mis-sent. `satisfiesStrictRequired` is deliberately conservative: a false negative costs only the stronger guarantee, a false positive costs the whole request.

**On the default-configuration gap.** You are right that it is untested here, and the cause is structural: this suite sets `NEUROLINK_DISABLE_BUILTIN_TOOLS=true` at the top, so it never exercises the configuration real callers get. That is a separate defect in its own right — it is why #1693's bug survived undetected — and I would rather fix the suite's blind spot in its own PR than widen this one.

murdore:
The composition finding was a real bug I introduced, and fixing it properly meant redesigning the change rather than patching it. Pushed as `1a9531f36`.

**MAJOR — composition members closed *and* sent non-strict.** Correct, and this was the best catch in the review. Closing each `allOf` member and then sending non-strict — where the schema is honoured as written, injections included — makes a composition whose members declare different properties **unsatisfiable**. Nothing validates. That is worse than the 400 this PR set out to fix, because it fails silently.

**The redesign.** Normalisation is no longer applied unconditionally. The schema is now mutated as little as possible, in this order:

```
1. already strict-legal  -> sent UNTOUCHED with strict: true
2. legal once closed     -> closed copy sent with strict: true
3. neither               -> sent UNTOUCHED with strict: false
```

Verified against the four shapes that matter:

```
zod all-required   strict=true   untouched=true    <- byte-identical to pre-change behaviour
zod optional       strict=false  untouched=true    <- the fix
plain all-required strict=true   closed=true
allOf composition  strict=false  untouched=true    members NOT closed
```

**MAJOR — blast radius, and "strict:false is a behavioural downgrade".** Both land differently under the new ordering, and I think the concern is now answered rather than merely acknowledged:

- Case 1 gives **parity**, not just safety. A Zod schema with all-required properties already converts to a strict-legal shape, so it goes out byte-identical to what it did before this PR. That is the overwhelmingly common path for this family.
- Case 3 never mutates, so the non-OpenAI members — which may not implement OpenAI's strict contract at all — see exactly the schema the caller wrote. Your point that injecting `additionalProperties: false` changes validation semantics for those endpoints is precisely why closure is now gated behind "and it buys us strict".
- On the downgrade: `strict: false` only replaces a request that was returning **400** before. There is no path where a previously-succeeding strict request becomes non-strict.

So the change can only relax `strict` or close a schema that was going to be refused anyway. It cannot make a previously-succeeding request fail, and it cannot alter a schema that is sent non-strict.

**MINOR — normalisation completeness.** All of it closed: `patternProperties` added to the map list, `prefixItems` to the node list, schema-valued `additionalProperties` recursed into while a boolean one is left as written, and a `type: "object"` node is closed whether or not it declares `properties`.

**On the default-configuration coverage gap.** Still true, still structural: this suite sets `NEUROLINK_DISABLE_BUILTIN_TOOLS=true` at the top. It is the same blind spot that let #1693's defect survive, and I would rather fix it in its own PR than keep working around it per-test.

src/lib/providers/openaiChatCompletionsClient.ts
```diff
@@ -475,6 +475,138 @@ export const v3ToolChoiceToOpenAI = (
...
+ * guarantee, while a false positive costs the whole request.
+ */
+const satisfiesStrictRequired = (node: unknown): boolean => {
+  if (Array.isArray(node)) {
+    return node.every(satisfiesStrictRequired);
+  }
+  if (!node || typeof node !== "object") {
+    return true;
+  }
+  const rec = node as Record<string, unknown>;
```
Tara-ag, line 581: **MINOR — the strict gate's keyword allowlist is narrower than OpenAI's documented supported subset, so the gate can still send a schema `strict:true` that OpenAI 400s.**

`satisfiesStrictRequired` gates `strict` on `STRICT_UNSUPPORTED = ["allOf","not","if","then","else"]`. Per the same OpenAI "Supported schemas" doc this PR (and the earlier CodeRabbit research) relies on, that is only *part* of the rejected set — unknown/unlisted keywords are also rejected when `strict: true`:

- `oneOf` is walked by `SCHEMA_NODES` but is *not* in `STRICT_UNSUPPORTED`, so a schema whose `oneOf` members all happen to close can be sent `strict: true` (case 1/2) and still be refused. `anyOf` is correctly left alone (supported), but `oneOf` sits in the unlisted/rejected bucket.
- `patternProperties` is in `SCHEMA_MAPS` (so it's walked and can pass) yet is not in the supported strict subset.
- `minLength` / `maxLength` / `uniqueItems` / non-recursive (external) `$ref` are similarly unlisted — a schema carrying one can pass the gate.

Because `satisfiesStrictRequired` is deliberately conservative ("a false positive costs the whole request"), an allowlist gap here produces exactly the failure this PR sets out to kill — a `400` on a schema the code decided was strict-safe. Suggest extending `STRICT_UNSUPPORTED` to cover `oneOf` (and the other unlisted keywords), or scoping the gate's promise to the keywords it actually rejects. The `oneOf` case would be a good third live cell alongside the plain/optional ones.

```suggestion
  if (
    STRICT_UNSUPPORTED.some((k) => k in rec) ||
    (rec.oneOf !== undefined && !rec.oneOfOk) // or drop 'oneOf' from SCHEMA_NODES / add to STRICT_UNSUPPORTED
  ) {
    return false;
  }
```

Tara-ag (commented):
Recurring review against the reworked commit `1a9531f36`. The prior MAJOR composition finding is resolved by the 3-case redesign (verified below); one new MINOR completeness finding posted inline.

Tara-ag:
## Verdict: **APPROVE** (recurring review — commit `1a9531f36`)

The reworked three-case design resolves the outstanding MAJOR composition finding and keeps the fix correct: schemas are sent **untouched** unless closure *buys* `strict: true`, so non-strict schemas are never mutated and composition members are never corrupted. Prior findings verified as fixed; one new MINOR completeness nit posted on the strict gate.

|---|---|---|
| MINOR | `src/lib/providers/openaiChatCompletionsClient.ts:581` | `STRICT_UNSUPPORTED` misses `oneOf`/`patternProperties`/`minLength`-etc from OpenAI's documented rejected set, so the gate can still send `strict:true` on a schema OpenAI 400s. |

**Resolved this pass (not reposted, verified in code + author reply `5648660077`):**
- **MAJOR — composition closure** (`yama:1691-composition-member-closure`, both threads now resolved): `allOf`/`then`/`else` are in `STRICT_UNSUPPORTED`, so `satisfiesStrictRequired` is `false` for both `original` and `closed`, landing in case 3 → the schema goes out **untouched** with `strict:false`, members never closed. Verified: for `allOf`, both strict checks return false → `schema = original`. No longer unsatisfiable.
- **MAJOR — blast radius / tools-active coverage** (resolved earlier): sole caller confirmed via code graph — `v3ResponseFormatToOpenAI` ← `buildDelegatingModel` (`openaiChatCompletionsBase.ts:743`), the shared constructor for the whole OpenAI-compatible family. Coverage gap acknowledged by author as structural (`NEUROLINK_DISABLE_BUILTIN_TOOLS=true`); Azure cell added as the second path that actually exercises the `json_schema`+strict decision.
- **MAJOR — optionality test** (resolved earlier): now asserts `!("population" in data)` at line 626.
- **MINOR — normalisation completeness** (resolved earlier): `prefixItems`, `patternProperties`, schema-valued `additionalProperties`, and objects without `properties` now handled.

**Checked clean:**
- No mutation of caller schemas: `withClosedObjects` builds fresh objects; case 3 sends `original` by reference and never touches it.
- `added additionalProperties:false` only where absent; boolean flags left as written.
- Redesign gives parity for the common all-required path (case 1 sends byte-identical schema with `strict:true`) and can only *relax* `strict` or close a schema that would otherwise 400 — no previously-succeeding request can newly fail.
- Rules 1/4/6/15: no registry/dynamic-import changes, CLI/SDK separation intact, no `formatProviderError` change, tests are e2e (single graph). Secret-scan of new code: no credentials.
- Single Conventional Commit `fix(openai): …` — compliant.

No CRITICAL/security findings. The open MINOR is non-blocking; please consider folding `oneOf` (and other OpenAI-rejected/unlisted keywords) into the strict gate in a follow-up.

Tara-ag (approved):
Reviewing commit `1a9531f36`. The 3-case redesign resolves the composition finding; the two prior blockers (optionality assertion, normalisation completeness) and the blast-radius coverage gap are addressed. Verdict: **APPROVE** on the single remaining MINOR completeness note (strict-gate keyword allowlist).

Tara-ag:
## Verdict: **APPROVE**

Reviewed against commit `1a9531f36`. The redesign (three-case ordering: strict-legal → sent untouched; legal-once-closed → closed copy; neither → sent untouched non-strict) fixes the reported defect and answers every prior blocker. All MAJOR findings from earlier passes are resolved; one MINOR completeness note remains, non-blocking.

|---|---|---|---|
| MINOR | `src/lib/providers/openaiChatCompletionsClient.ts:581` | Strict-gate keyword allowlist (`STRICT_UNSUPPORTED`) narrower than OpenAI's documented rejected set — a schema with `oneOf` / `patternProperties` / `minLength` / `maxLength` / `uniqueItems` can still be sent `strict:true` and 400. | OPEN (non-blocking) |

### Previously-raised findings — status on this commit
- **MAJOR composition-closure (`allOf`/`then`/`else`)** → **RESOLVED.** `STRICT_UNSUPPORTED` now includes those keywords, so `satisfiesStrictRequired` returns `false` and composition schemas go out **untouched** with `strict:false` (case 3 — no closure injected). The author's explanation in `5648660077` is correct; the duplicate corrected-anchor thread is resolved.
- **MAJOR optionality assertion** → **RESOLVED.** The e2e test now asserts `!("population" in data)` (line 626), so a regression that forces `required` can no longer pass green.
- **MINOR normalisation completeness** → **RESOLVED.** `prefixItems` walked, schema-valued `additionalProperties` recursed, objects closed regardless of `properties`.
- **MAJOR blast radius / coverage** → **ADDRESSED.** Azure (the family member that actually exercises `response_format`+tools) is now a covered cell; the remaining default-config gap is structural/separate (#1693) and acknowledged.

### Checked clean
- No mutation of the caller's schema in cases 1/3; fresh-object closure in case 2; `strict:false` only ever replaces a prior 400 (no previously-succeeding request downgraded).
- No public SDK API change (Rule 5); no Rules 1/4/6/15 violations; no secrets; single Conventional Commit, valid scope.

**Note to maintainers:** two stale `<!-- yama:summary -->` comments (`5645529298`, `5645538551`) carry the old NEEDS_WORK verdict from prior commits. This comment is the canonical current summary (APPROVE). The automation account cannot edit/delete issue comments; please remove the two stale ones when convenient.

No CRITICAL / security findings.

Tara-ag:
Confirmed the redesign in `1a9531f36` and accepting it — thanks for the thorough write-up.

Your reasoning holds: with `allOf`/`then`/`else` (and `not`/`if`) now in `STRICT_UNSUPPORTED`, `satisfiesStrictRequired` returns `false` and the schema lands in case 3 — sent **untouched with `strict:false`**, so composition members are never closed and nothing is silently made unsatisfiable. And since closure injection is now gated behind "and it buys us strict," case 3 can't corrupt non-OpenAI endpoints either. The "can only relax `strict` or close a schema that was going to be refused anyway" framing is the right invariant.

The strict-gate keyword allowlist note (a MINOR, `openaiChatCompletionsClient.ts:581`) remains the only open thread — non-blocking. If you later move the "fully required" guarantee to an allowlist that mirrors OpenAI's documented supported subset (`oneOf`/`minLength`/`maxLength`/`uniqueItems`/external `$ref` still sit in the unlisted-rejected bucket), the `oneOf` case would make a tidy third live cell. Happy to approve as-is.

PR #1692, opened 2026-09-12
fix(proxy): support bounded OpenTelemetry-only logging

murdore:
Proxy logging currently duplicates request, lifecycle, body, supervisor and console records on disk, adding file writes and cleanup scans to the serving process. This adds an explicit `NEUROLINK_PROXY_LOG_SINK=otel` mode that sends application logs to an OTLP collector and disables those disk sinks. Existing file logging remains the default.

- Isolate metadata and body queues with bounded capacity, transport timeouts and visible submitted, HTTP-acknowledged, unconfirmed and dropped counters. Lifecycle logging becomes asynchronous best effort in this mode.
- Preserve complete final-request metadata, attempts, lifecycle events and bounded, redacted body captures. Report capture rejections explicitly; HTTP acknowledgements do not claim backend durability.
- Route worker, supervisor and updater diagnostics through a dedicated log provider; flush updater exits. Persist collector settings in an atomically written private launchd plist and route service stdout/stderr to `/dev/null`.
- Disable body artifacts, gzip output, updater log files and retention scans in OTel-only mode. Include migration documentation and generated reference assets.

Validation: telemetry regression suite 44/44; production bugfix suite 281/281 in an isolated home; tools/test TypeScript check; ESLint (zero errors); full package build, CLI import verification, browser bundle and publint; documentation build and generated assets. Tests use isolated fixture collectors and recorded upstream responses.

The worker implementation was separately activated on a local 12.14.0 runtime: all 3,087 historical proxy log files showed zero growth or changes over 25.9 minutes; supervisor and worker stdout/stderr were verified as `/dev/null`; fresh finals, attempts, lifecycle and body records were retrieved from OpenObserve. Later updater/installer review fixes were validated locally and are not needed by that pinned serving worker. Backend persistence and operational state files remain. Bounded queues and body capture are best effort. Two upstream EPIPE failures remain outside this logging change.


src/lib/proxy/otelLogSink.ts
```diff
@@ -0,0 +1,247 @@
...
+        if (result.code === ExportResultCode.SUCCESS) {
+          state.transportAcknowledged += records.length;
+          state.lastAcknowledgedAt = new Date().toISOString();
+        } else {
+          state.exportUnconfirmed += records.length;
+          state.lastFailureAt = new Date().toISOString();
+        }
+        callback(result);
+      });
+    },
```
Tara-ag, line 66: **MINOR** — `state.outstanding` is only released inside the transport result callback. `OTLPLogExporter.export` generally reports failures through the callback, but if it ever throws synchronously (malformed record, transport error before a callback is scheduled), `state.outstanding` never decrements. Each such throw permanently consumes one batch's worth of the 2048/256 queue capacity forever, since `onEmit` gates on `outstanding >= capacity` — a slow but irreversible capacity shrink in the exact path this change is meant to protect (proxy admission must not stall on telemetry).

Wrap the transport call so capacity is always reclaimed:

```suggestion
    export(records, callback) {
      try {
        transport.export(records, (result) => {
          state.outstanding -= records.length;
          if (result.code === ExportResultCode.SUCCESS) {
            state.transportAcknowledged += records.length;
            state.lastAcknowledgedAt = new Date().toISOString();
          } else {
            state.exportUnconfirmed += records.length;
            state.lastFailureAt = new Date().toISOString();
          }
          callback(result);
        });
      } catch (error) {
        state.outstanding -= records.length;
        state.exportUnconfirmed += records.length;
        state.lastFailureAt = new Date().toISOString();
        callback({ code: ExportResultCode.FAILURE, error });
      }
    },
```

The tests drive collector HTTP 400 rejection (failure through the callback) so this exact hole is not covered — worth an assertion that `outstanding` returns to 0 after a synchronous export fault.

src/lib/proxy/otelLogSink.ts
```diff
@@ -0,0 +1,247 @@
...
+    resource: resourceFromAttributes({
+      "service.name": process.env.OTEL_SERVICE_NAME ?? "neurolink-proxy",
+      "service.instance.id": `${role}-${process.pid}`,
+      "process.pid": process.pid,
+      "proxy.process.role": role,
+    }),
+    processors: [
+      {
+        onEmit(record, context) {
+          (record.attributes["proxy.record_kind"] === "body"
```
Tara-ag: **MINOR** — `record.attributes["proxy.record_kind"]` is accessed unguarded inside the routing processor. If the SDK ever delivers a log record with no attributes, the throw propagates out of the provider's emit pipeline rather than being counted as `invalidRecords`. Since this routing runs in the same `processor` that every proxy log record flows through, one bad record can break emission for all following records in the request. Use optional chaining to keep the routing safe:

```suggestion
        onEmit(record, context) {
          (record.attributes?.["proxy.record_kind"] === "body"
            ? bodies
            : metadata
          ).processor.onEmit(record, context);
        },
```

src/lib/proxy/requestLogger.ts
```diff
@@ -275,7 +294,7 @@ export async function logRequest(entry: RequestLogEntry): Promise<void> {
   }
 
   // Emit OTLP log record (additive — file logging is the primary sink)
-  emitOtlpLogRecord(entry);
+  await emitOtlpLogRecord(entry);
```
Tara-ag: **MINOR** — this `await` changes the **default** (non-opt-in) file-logging path. `emitOtlpLogRecord` returned `void` before (fire-and-forget); now `logRequest` awaits its promise. `resolveLoggerProvider()` may do a dynamic `import()` of the observability instrumentation on the first call in a given process, and awaiting makes that a synchronous gate on the final-request path before the function resolves. `logRequestAttempt` shares the same pattern.

It happens to be harmless in this PR's test/worker flows (provider initializes early in `initRequestLogger`), and the `.catch()` inside guarantees the awaited promise never rejects. Still, making a non-opt-in path depend on OTLP resolution is the kind of subtle default-mode change the recovery of this PR's own "admission must never block on telemetry" invariant should avoid. Consider keeping fire-and-forget in the file branch:

```suggestion
  // Emit OTLP log record (additive — file logging is the primary sink)
  void emitOtlpLogRecord(entry);
```

(Only the OTel-only mode genuinely needs the awaited emission so it can flush before shutdown.)

Tara-ag (approved):
Reviewed. Clean, well-tested change; three defensive MINOR notes on the new OTel sink.

Tara-ag:
## ✅ APPROVE (minor, non-blocking notes)

Opt-in `NEUROLINK_PROXY_LOG_SINK=otel` is a genuinely well-contained change: default mode is untouched, the body-redaction path is preserved in the bounded worker, admission never blocks on telemetry, and the new OTel-only suite (43/43, incl. zero log-dir creation, redaction assertions, collector rejection, queue bounds) matches what the PR describes. No finding rises to MAJOR/CRITICAL.

| Sev | file:line | Finding |
|-----|-----------|---------|
| MINOR | `src/lib/proxy/otelLogSink.ts:40-52` | `state.outstanding` only decremented inside the transport callback → a synchronous throw from `OTLPLogExporter.export` permanently leaks one batch of queue capacity (2048/256) since `onEmit` gates on it. Suggest try/finally + a regression assertion. |
| MINOR | `src/lib/proxy/otelLogSink.ts:113` | Routing processor reads `record.attributes["proxy.record_kind"]` unguarded; a record with no attributes throws out of the emit pipeline instead of counting as `invalidRecords`. Suggest `record.attributes?.[...]`. |
| MINOR | `src/lib/proxy/requestLogger.ts:297` | Default file mode now `await emitOtlpLogRecord(entry)` (was fire-and-forget); first-call `resolveLoggerProvider()` dynamic import becomes a gate on the final path in a non-opt-in mode. Suggest `void` in the file branch. |

**Checked and clean**
- **Redaction / secrets**: body chunks, headers, and the `body_capture_index` carry only redacted data; the suite asserts `api_key` and `Bearer` never reach the collector. OTel-only exporter endpoint uses http/https validated at init; no secret surfacing in source.
- **Admission non-blocking**: `persistProxyLifecycleAcceptance` enqueues async in otel mode; collector rejection/outage does not 503 admission (tested).
- **Queue isolation**: metadata (2048) vs body (256) routed independently via `proxy.record_kind`; `outstanding` includes in-flight exports; drain accounting tested.
- **Default mode unchanged**: all `isProxyOtelOnly()` guards are opt-in checks; file path still exercises `appendMetadataRecord` + additive OTLP.
- **Rules**: no static provider import (Rule 1 unaffected — this adds a dedicated ESM import of the official OTLP libs, not the registry's circular-dep path); `proxy.record_kind` split keeps attempt/final/aux fields distinct per the doc's dashboard claims.
- **Blast radius** (graph): affected flows are confined to `src/lib/proxy/*` and `src/cli/commands/proxy.ts` launchd/updater wiring; risk 0.85 highest on the request-logger emit chain, all under test. Test-only worker hooks (`__bodyCaptureWorkerTestHooks`, `__requestLoggerTestHooks`) remain single-module-graph (`dist`) — Rule 15 respected by the new suite.

**Not flagged**: Coderabbit + single-commit bot are the only other comments (no prior yama review to reconcile). The three MINOR items are all in the new file/hot-path and easily addressed; none blocks merge.

src/cli/commands/proxy.ts
```diff
@@ -4630,6 +4646,8 @@ export const proxyGuardCommand: CommandModule<object, ProxyGuardArgs> = {
       }) as Argv<ProxyGuardArgs>;
   },
   handler: async (argv) => {
```
Tara-ag: **MINOR** — the guard/updater process routes console to OTel but never flushes or shuts the sink down before exiting, so the diagnostics it just captured are dropped.

`proxyGuardCommand`'s handler calls `initializeProxyOtelLogs("updater")` + `routeProxyConsoleToOtel()`, which queue console records into the bounded `BatchLogRecordProcessor` (1000 ms scheduled delay). Neither `flushProxyOtelLogs()` nor `shutdownProxyOtelLogs()` is reached from this handler — unlike the supervisor (`runLaunchdProxySupervisor`) and worker (`registerProxyShutdownHandlers`) paths, which both flush+shutdown. A guard process that captures console output then exits before the 1 s scheduled delay elapses loses every queued record, contradicting the feature's own claim in `docs/proxy-otel-logging.md` that "Proxy application console diagnostics go to OTel."

Add a flush (and shutdown) before the handler returns, mirroring the supervisor/worker paths:

```suggestion
    initializeProxyOtelLogs("updater");
    routeProxyConsoleToOtel();
    const host = argv.host ?? "127.0.0.1";
    const port = argv.port ?? 55669;
    const parentPid = Number(argv.parentPid);
```
…then before the handler returns:
```suggestion
    await flushProxyOtelLogs().catch(() => undefined);
    await shutdownProxyOtelLogs().catch(() => undefined);
```
(`flushProxyOtelLogs` / `shutdownProxyOtelLogs` are already imported at the top of this file.)

If the guard deliberately discards its own console as best-effort, then `routeProxyConsoleToOtel()` in this short-lived handler is dead weight carrying a bounded queue that drains to nowhere — worth removing or explicitly documenting either way.

Tara-ag (commented):
Recurring review of `d0ab359` — prior findings resolved, one new MINOR on the guard handler. Full detail in the `<!-- yama:summary -->` comment.

src/lib/proxy/otelLogSink.ts
murdore: Fixed in d0ab359f. Export settlement is guarded against duplicate callbacks and synchronous throws, so capacity is released exactly once and failed records are counted as unconfirmed. The deterministic exporter-fault regression passes as part of the 44/44 telemetry suite.

src/lib/proxy/otelLogSink.ts
murdore: Fixed in d0ab359f. Routing now uses optional chaining for record attributes and defaults records without a body discriminator to metadata. Full packaging and the telemetry suite pass.

src/lib/proxy/requestLogger.ts
murdore: Fixed in d0ab359f. The default file path again uses fire-and-forget OTLP publication; only the opt-in OTel path awaits local emission before its flush boundary. The default-mode regression coverage remains passing.

Tara-ag (approved):
**APPROVE** — recurring review of `d0ab359`.

The opt-in `NEUROLINK_PROXY_LOG_SINK=otel` change is well-contained and correct. All prior findings (3 yama MINORs + 2 CodeRabbit MAJORs) are resolved in this commit with matching regression tests in `test/continuous-test-suite-proxy-telemetry.ts`. One new non-blocking MINOR remains open: the `proxyGuardCommand` handler initializes and routes console to OTel but never flushes/shuts the bounded sink down, so short-lived guard processes can drop queued records (thread on `src/cli/commands/proxy.ts:4647-4648`). It does not block merge.

Verified clean: no-disk invariant in OTel mode, redaction (body/headers never reach collector), admission non-blocking on collector outage, queue isolation (metadata 2048 / bodies 256), and Rule 1 / Rule 5 / Rule 15 conformance. Full detail in the `<!-- yama:summary -->` comment.

src/cli/commands/proxy.ts
```diff
@@ -5745,8 +5784,9 @@ export const proxyInstallCommand: CommandModule = {
       ),
     );
 
-    const plist = buildPlist(port, host, envFile, configFile);
-    writeFileSync(PLIST_PATH, plist, "utf-8");
+    const plist = buildProxyLaunchdPlist(port, host, envFile, configFile);
+    writeFileSync(PLIST_PATH, plist, { encoding: "utf-8", mode: 0o600 });
+    chmodSync(PLIST_PATH, 0o600);
```
Tara-ag: **MINOR** — the credential-bearing plist is still truncated in place before its mode is tightened. `mode: 0o600` only applies when the file is first created; on a reinstall where `PLIST_PATH` already exists (e.g. from an older install with looser perms), `writeFileSync` opens with `O_TRUNC`, which preserves the pre-existing mode, so between this write and the `chmodSync` on the next line the plist (containing `OTEL_EXPORTER_OTLP_LOGS_HEADERS` / `NEUROLINK_PROXY_SESSION_SECRET`) holds its old permissions.

Practical risk is negligible (single-threaded; one synchronous statement separates the two calls), which is why this stays MINOR and non-blocking — but the requested atomic write is a simple hardening:

```suggestion
    const plist = buildProxyLaunchdPlist(port, host, envFile, configFile);
    const tmpPlist = `${PLIST_PATH}.tmp-${process.pid}`;
    try {
      writeFileSync(tmpPlist, plist, { encoding: "utf-8", mode: 0o600 });
      renameSync(tmpPlist, PLIST_PATH);
    } catch (plistWriteError) {
      try { unlinkSync(tmpPlist); } catch { /* not created */ }
      throw plistWriteError;
    }
    chmodSync(PLIST_PATH, 0o600);
```

(`writeFileSync` / `renameSync` / `unlinkSync` are already available in this `proxyInstallCommand` handler scope via the existing `const { writeFileSync, mkdirSync, existsSync, chmodSync } = await import("fs")` at the top of the handler — add `renameSync, unlinkSync` to that destructure.) The final `chmodSync` remains as defense-in-depth.

Tara-ag (commented):
Recurring review — reconciled all prior findings; two non-blocking MINORs remain (see summary comment).

src/cli/commands/proxy.ts
Tara-ag: Recurring review: standing this MINOR down. A short-lived guard routing its console to OTel while never flushing before exit is consistent with the documented "diagnostics are best-effort / discard to /dev/null" posture (`docs/proxy-otel-logging.md`) already acknowledged in the summary; the supervisor and worker paths do flush+shutdown, and the guard's dropped console is a bounded diagnostic nicety, not data loss on any durable contract. Keeping it documented, resolving as accepted.

src/cli/commands/proxy.ts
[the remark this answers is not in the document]
Tara-ag: Duplicate of Tara-ag's `<!-- yama:plist-atomic-write -->` comment on the same lines (`#discussion_r3996576452`, which carries a complete atomic-write suggestion incl. temp-file cleanup). Keeping that one; resolving this thread to deduplicate.

Tara-ag (approved):
## ✅ APPROVE

Recurring review of the current head (`214b1db`). All previously-reported findings are reconciled in code with matching regression coverage. Two non-blocking MINOR notes remain and are documented in the `<!-- yama:summary -->` summary; neither anchors to a change that blocks merge.

Full detail in the summary comment.

src/cli/commands/proxy.ts
murdore: Addressed in 10b66149. The guard handler now flushes and shuts down the dedicated log provider in a finally wrapper for normal returns and exceptions, and the explicit successful updater process.exit path flushes first. The regression queues a log before an early guard return and verifies collector receipt and provider shutdown; all 44 telemetry tests pass.

src/cli/commands/proxy.ts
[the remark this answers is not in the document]
murdore: Addressed in 10b66149. Installation now uses the existing writeFileAtomic helper with explicit mode 0600: it creates a private sibling temporary file, applies mode before atomic rename, and cleans up on failure. The final chmod remains. TypeScript, ESLint and the full package/CLI build pass.

test/continuous-test-suite-proxy-telemetry.ts
```diff
@@ -2189,4 +2189,368 @@ await test("default shutdown flush waits for a capture beyond the metadata-only
...
+  const address = collector.address();
+  assert(address && typeof address === "object", "collector did not listen");
+  const before = {
+    mode: process.env.NEUROLINK_PROXY_LOG_SINK,
+    endpoint: process.env.OTEL_EXPORTER_OTLP_LOGS_ENDPOINT,
+  };
+  const dir = await mkdtemp(join(tmpdir(), "otel-only-"));
+  const destination = join(dir, "must-not-exist");
+  process.env.NEUROLINK_PROXY_LOG_SINK = "otel";
+  process.env.OTEL_EXPORTER_OTLP_LOGS_ENDPOINT = `http://127.0.0.1:${address.port}/v1/logs`;
```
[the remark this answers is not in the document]
murdore: Addressed in the current commit. Both collector.address() sites use an explicit null/string guard that throws, so TypeScript narrows to AddressInfo before address.port is read. The full tools/test typecheck passes, as did the type CI shard on the preceding revision.

Tara-ag:
**Recurring review @ `10b661492fd9dc5aefb3362c49845c51d8a7e981` — ✅ APPROVE**

Recurring pass over the current head. All previously-reported findings are reconciled in code and can be stood down:

| Prior finding | Status in `10b66149` |
|---|---|
| `otelLogSink.ts` export settlement — `state.outstanding` leaked if `OTLPLogExporter.export` threw synchronously | **Resolved.** Settlement refactored into a single `settle()` (lines 43-57) invoked exactly once per batch; guarded against duplicate callbacks and synchronous throws. Covered by the deterministic exporter-fault regression in `continuous-test-suite-proxy-telemetry.ts`. |
| `otelLogSink.ts` routing `record.attributes["proxy.record_kind"]` unguarded | **Resolved.** `onEmit` now uses optional chaining and defaults records without the discriminator to the metadata queue (lines 132-137). |
| `requestLogger.ts` default file-mode awaited `emitOtlpLogRecord` | **Resolved.** Default path at `requestLogger.ts:297` is again fire-and-forget (`void emitOtlpLogRecord(entry);`); only the OTel-only branch awaits (`:281`) before its flush boundary. |
| `proxy.ts` credential plist written in place before `0600` applied | **Resolved.** `proxy.ts:5791` now writes via `writeFileAtomic(PLIST_PATH, plist, 0o600)` (atomic sibling temp + rename, cleanup on failure); the final `chmodSync` remains as defense-in-depth. Resolved the open `yama:plist-atomic-write` thread. |
| `proxy.ts` updater/guard never flushed the OTel sink before exit | **Resolved.** Success path flushes+shuts down before `process.exit(0)` (`proxy.ts:5112-5114`); the guard handler is wrapped in `withProxyOtelLogShutdown` (wired at `proxy.ts:4650`, defined `otelLogSink.ts:270-281`) which flushes and shuts down on every normal return or exception. |

**Checked and clean in this pass**
- Redaction: `bodyCaptureProcessing.ts` `redactBody`/`redactHeaders` apply key-name + structural redaction; `processProxyBodyCapture` returns redacted-only data when `logDir === null` (OTel mode disables body artifacts on disk, per the PR's design). No secret-surfacing path found.
- Admission non-blocking: `persistProxyLifecycleAcceptance` enqueues best-effort in OTel mode; bounded queues cannot stall admission.
- OTel-mode disk avoidance: `logCleanupScheduler`, `workerLog`, request logger, and body artifacts all gate on `isProxyOtelOnly()`.
- Rules: no static provider import (Rule 1 untouched — dedicated OTLP libs); CLI-vs-SDK isolation (Rule 4) preserved; public type surface only gained additions (Rule 5); new suite is single-module-graph `dist`-driven with an isolated fixture collector (Rule 15).
- Blast radius: affected flows confined to `src/lib/proxy/*` and the launchd/updater wiring in `src/cli/commands/proxy.ts`; every changed hot path carries matching coverage in the telemetry (44/44) and bugfix (281/281) suites.

**Not flagged:** leftover differences stay MINOR and non-blocking. This was a large, heavily-reviewed change; no new finding in this pass rises above what the author has already addressed. Ready to merge.

src/cli/commands/proxy.ts
murdore: This is addressed in 10b66149, which replaced the in-place write with await writeFileAtomic(PLIST_PATH, plist, 0o600). The shared helper creates a private sibling file, applies permissions before the atomic rename, and cleans up failed writes. The final chmod is retained. The displayed diff is from the preceding revision; the current package and CLI build pass.

Tara-ag (approved):
**APPROVE** — recurring review at head `10b66149`.

All prior findings from earlier rounds are resolved in code with matching regression coverage; no new issue rises above MINOR and the change does not block merge.

| Prior finding | Status at head |
|---|---|
| `otelLogSink.ts` export settlement leaked queue capacity on a synchronous export throw | ✅ Single `settle()`, exactly-once (43-57) + exporter-fault regression |
| `otelLogSink.ts` unguarded `record.attributes["proxy.record_kind"]` routing | ✅ Optional chaining + metadata default (132-137) |
| `requestLogger.ts` default file-mode awaited OTLP emit | ✅ Fire-and-forget `void` on file path (297); OTel-only branch awaits |
| Credential plist truncated in place before `0600` | ✅ `writeFileAtomic(PLIST_PATH, plist, 0o600)` — thread resolved |
| Guard/updater never flushed sink before `process.exit` | ✅ `withProxyOtelLogShutdown` finally wrapper (proxy.ts:4650) + explicit exit flush |
| `collector.address()` narrowing in telemetry test (CodeRabbit) | ✅ Explicit null/string guard throws before `address.port` |

Verified clean: redaction (body/headers never reach collector), admission non-blocking on collector outage, bounded/isolated queues (metadata 2048 / bodies 256), no-disk invariant in OTel-only mode, and Rule 1 / Rule 4 / Rule 5 / Rule 15 conformance. Affected flows (graph) confined to `src/lib/proxy/*` and launchd/updater wiring.

Tara-ag:
## Post-merge retrospective — PR 1692 `fix(proxy): support bounded OpenTelemetry-only logging`

Read-only analysis of Yama's (Tara-ag) review across the four review rounds, the CodeRabbit threads, the author's (murdore) replies, and final resolution state at merge (`10b66149`).

### What Yama found, and how it resolved

| # | Finding (Yama) | Severity | Resolution |
|---|---|---|---|
| 1 | `otelLogSink.ts:40-52` — `state.outstanding` only decremented inside the export callback; a synchronous throw from `OTLPLogExporter.export` permanently leaks one batch of 2048/256 queue capacity forever, since `onEmit` gates on `outstanding` | MINOR | **Fixed** — refactored into a single `settle()` invoked exactly once, guarded against duplicate callbacks and sync throws, with a deterministic exporter-fault regression |
| 2 | `otelLogSink.ts:113` — routing processor reads `record.attributes["proxy.record_kind"]` unguarded; a no-attributes record would throw out of the emit pipeline (in the same processor *every* proxy record flows through) | MINOR | **Fixed** — optional chaining + default a records with no discriminator to the metadata queue |
| 3 | `requestLogger.ts:297` — the awaited `emitOtlpLogRecord` changed the **default, non-opt-in** file-logging path into a synchronous gate on `resolveLoggerProvider()`'s first-call dynamic import — a subtle default-mode regression of this very PR's "admission must never block on telemetry" invariant | MINOR | **Fixed** — default path restored to fire-and-forget `void emitOtlpLogRecord(entry)`; only the OTel-only branch awaits before its flush boundary |
| 4 | `proxy.ts:5788-5789` — credential-bearing plist (OTLP headers / session secret) truncated in place before `0600` is applied (`mode:0o600` only applies at creation), so a reinstall over a looser-perm old file holds old perms between write and `chmodSync`; also caught independently by CodeRabbit, which Yama deduplicated onto `yama:plist-atomic-write` | MINOR | **Fixed** — `writeFileAtomic(PLIST_PATH, plist, 0o600)` (sibling temp, mode-before-rename, cleanup on failure), final `chmodSync` retained |
| 5 | `proxy.ts:4647-4648` — guard/updater process routes console to OTel but never flushes or shuts the sink down before exiting, so short-lived guard captures are dropped | MINOR | **Initially stood down** by Yama (resolved as "accepted" — consistent with the documented best-effort/discard-to-`/dev/null` posture since supervisor/worker do flush), **then the author fixed it anyway** in `10b66149` with a `withProxyOtelLogShutdown` finally wrapper + explicit exit flush, and Yama confirmed resolved |

**No Yama finding proved to be a false positive.** All five, including the one Yama initially weighed down to non-blocking, were ultimately correct and were fixed by the author with matching regression coverage. The only near-miss was #5, where Yama reasoned from the "best-effort, no durable contract" principle that the flush was unnecessary — but the author's cheap constant-cost fix was genuinely the better call, and Yama accepted it. This is a case where the general "best-effort" rule slightly under-weighted a still-worthy loose end.

### CodeRabbit findings (context)

- Plist atomic write — duplicate of Yama #4, deduplicated, fixed.
- OTLP endpoint must require HTTPS for non-loopback hosts (`otelLogSink.ts`) — addressed in `d0ab359`.
- `collector.address()` does not narrow under the harness `assert` (tsc fails in the test shard) — addressed (explicit null/string guard throws before `address.port`), confirmed by both author and CodeRabbit.
- **Log-destination message** (`proxy.ts:5818`) — in OTel-only mode launchd sends stdout/stderr to `/dev/null`, but the success message still tells operators to inspect `proxy-launchd-*.log` which is never created. Post as "outside diff". **No fix thread, no resolution, not referenced in any final APPROVE.** This appears to be the one finding **ignored and merged anyway**, and it was a still-valid operational-UX flaw at merge.

### Replies that settled a project convention

- #3 settled the **"default mode must stay default"** convention: a reviewer must re-check that an opt-in feature doesn't silently alter the non-opt-in path — Yama cited the PR's own invariant to catch it, and the author restored fire-and-forget. This confirms the repo convention that opt-in regimes must be fully additive to the default surface.
- #5 settled the teardown posture for the OTel sink: supervisor and worker must flush+shutdown; the guard/updater is expected to too, and the final implementation forces it via a finally wrapper so no exit path (normal or exceptional) can leak buffered records. Even "best-effort" sinks get deterministic emission on process exit here.
- The deduplication of the CodeRabbit plist thread onto `yama:plist-atomic-write` established that identical findings across reviewers are merged onto one thread rather than double-posted.

---

## Learnings

- Verify that every short-lived or child process (guard, updater, worker) that calls `initializeProxyOtelLogs` / routes console to OTel also flushes and shuts the sink down on **every** exit path — normal return, caught throw, and direct `process.exit` — via a finally wrapper, even when the documented posture is "best-effort", because in this repo best-effort emits are still expected to be deterministic on process teardown.
- When a change adds an opt-in mode, specifically re-check the **default (non-opt-in) path** it touches: a fire-and-forget call changed to an `await` on the default path is a default-mode regression even if harmless in the current test flows — the opt-in regime must remain fully additive.
- Read negative state (capacity counters like `outstanding`, batch creds) as requiring exactly-once settlement: guard against both callback-duplication and synchronous throws before the callback is scheduled, and ship a deterministic fault regression that drives the exact hole the callback-only path doesn't cover.
- Treat any index access into optional or runtime-shaped data on a shared hot path (e.g. `record.attributes[...]` in a processor every proxy record flows through) as unguarded-by-default, and use optional chaining with a documented default so one malformed record can't break emission for all following records.
- When persisting credential-bearing files (plists, env, token stores), never depend on a `mode` that only applies at creation: write to a sibling temp with the mode applied before an atomic rename and clean up on failure, so a reinstall over an older looser-permission file cannot expose the credentials between write and `chmodSync`.
- Do not silently file a still-valid operational-UX finding only as an "outside-diff" note with no tracked thread — a message that tells operators to inspect a log file that this mode never creates is a real defect that will otherwise be merged and only found in production.

PR #1658, opened 2026-09-07
fix(providers): ask for JSON when tools suppress response_format

murdore:
Validation of the review on PR #1658 is complete for the current head **`895eafd4`**. Net: the PR already carries a clean, complete review; the state matches the verdict. No corrective action was required this pass.

**Review state — matches the verdict (APPROVE).** The authoritative `yama:summary` (`issuecomment-5657094666`) is APPROVE on `895eafd4`. The latest approving review, `5228630417`, is also on `895eafd4` (submitted 2026-09-16T21:38:58Z, matching the PR's `updated_at`) — so there is **no stale-review gap**: the approving review is on the exact current commit. `mergeable_state: "blocked"` is not a review-state mismatch but the merge gate (branch protection on the `release` base; `provider-safety-net-shards (contract)` still in_progress). All 28 completed check runs on the head are `success`; the remaining two checks in_progress are the contract shard and the orchestration `Yama PR Review`.

**Comments — verified clean.** All 23 review threads are `resolved`; every finding has exactly one thread. My inline suggestion blocks use correct fences (3-backtick for code; a 4-backtick fence in the `rawfinishreason-edgedefault` block whose replacement itself contains backticks). The authoritative summary and the APPROVE review render as well-formed markdown (verdict header, finding table, `file:line` anchors). No malformed comment found.

**Replies — none pending.** murdore's two substantive responses this arc are both answered-by-action and resolved:
- `docs-reformat-degradation`: his "your call" question was resolved by committing the paragraph at Sachin's direction (thread `…ht3F-`).
- `outcome-reask-tokens-discarded`: his fix explanation (120 passed; two discriminating billing sections) is verified against the code and recorded as resolved in the authoritative summary.

**Duplicates — none live; removal not possible in this toolset.** Three duplicate review-thread pairs exist from the earlier d7280ea cycle (`yama:reformat-prompt-json-value` thr.9/12, `yama:toolexecutions-merge` thr.10/13, `yama:docs-cancellation` thr.11/14) plus the older superseded `yama:summary` issue comments. They are all `resolved` and/or `outdated` and each designates the current summary as the sole authority, so none presents as an active finding. However, this toolset has **no delete** capability for either review-thread or issue comments (only `update_issue_comment` and `delete_pending`), so the historical duplicates cannot be physically removed. No new comment duplicated anything.

[...]

PR #1693, opened 2026-09-12
test(json-e2e): cover the schema contract on a tools-bearing cohere request

murdore:
## Why this PR shrank to a test

It originally carried two things: a fix in `src/lib/providers/openaiChatCompletionsBase.ts`, and a live json-e2e test proving it. The fix half is **superseded by #1658**, which repairs the same defect independently and more completely. Only the test half remains, and the base is now `fix/structured-output-with-tools` so this reads as the delta it actually is.

### Why the fix half could not simply be layered on #1658

| | this PR (before) | #1658 |
|---|---|---|
| hoist `schemaJson` out of the `responseFormat` branch | yes | yes (line 1178) |
| compute `responseFormatSuppressed` once | no | yes (line 1187) |
| re-ask **without tools** on the suppressed path | **no** | yes (`withTools = false`) |

The last row is the blocker. This PR widened the outcome-path guard at line 1300 to also catch suppressed providers — but that guard runs *before* #1658's tools-free handler at line 1329, and it re-asks with `withTools` at its default `true`. Spelling a JSON Schema into a conversation that still carries tools is the counter-failure `runLoop`'s own comment documents: groq reads the schema as another tool and calls one named `JSON`, which the server rejects. Layered on #1658, the suppressed case would be caught at 1300 with tools still attached, and #1658's correct handler at 1329 would be unreachable for it.

So the two overlapped on the same fourteen lines, and the overlap was not mechanical — resolving it in favour of this PR would have reintroduced the defect.

### What remains, and why it is not redundant

- **The live cohere case.** `CohereProvider` extends the same `OpenAIChatCompletionsProvider` base and does not override `suppressResponseFormatWithTools`, so it takes the suppressing default and genuinely exercises the repaired path. Cohere appears in **no** cell of the json-e2e matrix — this is coverage #1658 does not have.
- **`disableTools: false`** added to the call, matching every other tools-bearing test in this file. Without the tool actually riding along, the turn takes the ordinary `response_format` route and the test passes while asserting nothing.
- **`timed.?out`** added to `isInfraError`, so a provider-side timeout skips instead of reporting as a product failure.

The original **deepseek** row is dropped: #1658 added deepseek as a breadth cell running `plain-prose-prompt` **with tools**, so that provider already covers this exact path. Keeping it here would buy a second live call and no extra signal.

### Evidence — discriminating RED → GREEN

Run in this branch's worktree against the live cohere API, reverting only `openaiChatCompletionsBase.ts` to #1658's own base and restoring it:

```
PHASE RED    openaiChatCompletionsBase.ts = a7f4fd72b  (#1658's base, fix absent)
  BUILD_EXIT=0
  x cohere:command-r-plus-08-2024 - a schema survives a request that also carries tools
      -> structuredData must be produced even when tools suppress response_format
  Passed: 4   Failed: 2

PHASE GREEN  openaiChatCompletionsBase.ts = 6e9fcf364  (#1658's fix present)
  BUILD_EXIT=0
  v cohere:command-r-plus-08-2024 - a schema survives a request that also carries tools
  Passed: 5   Failed: 1
```

A test that passed in both columns would prove nothing; this one fails without the fix and passes with it.

The single failure remaining in GREEN is `anthropic:claude-sonnet-4-6 — forced truncation yields a partial object, never a string`, which fails identically on the untouched base branch — pre-existing and unrelated to this change.

Builds were gated on their own exit code before each suite ran, and the source file was verified byte-identical to #1658's version afterwards (`final_srcsha == expected_srcsha`, working tree clean).

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage for structured data generation when tools are enabled, validating returned capital and population fields.
  * Expanded test handling to recognize timeout-related infrastructure failures as skippable, reducing false test failures in affected environments.

### DIFF
--- a/src/lib/providers/openaiChatCompletionsBase.ts
+++ b/src/lib/providers/openaiChatCompletionsBase.ts
@@ -1167,14 +1167,19 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
     // Structured output rides response_format, which is what Output.object did
     // on the ai path. Suppressed where the provider says the combination with
     // tools is rejected.
+    // The schema is converted whether or not it reaches the wire: when tools
+    // suppress `response_format`, the prompt-based fallback below still needs
+    // it. Previously it was computed only inside the branch, so the suppressed
+    // case had nothing to fall back WITH.
+    const jsonSchema = options.schema
+      ? (convertZodToJsonSchema(options.schema as ZodUnknownSchema) as Record<
+          string,
+          unknown
+        >)
+      : undefined;
     const responseFormat =
-      options.schema && !(hasTools && this.suppressResponseFormatWithTools())
-        ? {
-            type: "json" as const,
-            schema: convertZodToJsonSchema(
-              options.schema as ZodUnknownSchema,
-            ) as Record<string, unknown>,
-          }
+      jsonSchema && !(hasTools && this.suppressResponseFormatWithTools())
+        ? { type: "json" as const, schema: jsonSchema }
         : undefined;
 
     const conversation = (await this.buildMessagesForStream(
@@ -1271,17 +1276,28 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
     // above was reached; the native loop has no such parser, so the silent
     // case sailed through and handed the caller prose. Same recovery, keyed on
     // the result rather than on an exception.
+    // Keyed on the CALLER's schema, not on whether response_format reached the
+    // wire. When tools suppress response_format the schema never ships at all,
+    // nothing instructs the model to emit JSON, and the answer comes back as
+    // prose with `structuredData` undefined — silently breaking the documented
+    // `generate({ schema })` contract on every OpenAI-compatible provider that
+    // suppresses (all of them except OpenAI and Azure, which opt out). That
+    // suppressed case is precisely what the prompt fallback exists for, yet
+    // the old `responseFormat !== undefined` guard made it unreachable there.
     if (
-      responseFormat !== undefined &&
       options.schema !== undefined &&
       !yieldsSchemaValidObject(loop.text, options.schema as ValidationSchema)
     ) {
       logger.warn(
-        `[${this.providerName}] response_format did not yield a schema-valid object — retrying with the schema in the system prompt`,
-        { provider: this.providerName, model: modelId },
+        `[${this.providerName}] answer was not a schema-valid object — retrying with the schema in the system prompt`,
+        {
+          provider: this.providerName,
+          model: modelId,
+          responseFormatSent: responseFormat !== undefined,
+        },
       );
       loop = await runLoop(
-        appendJsonSchemaInstruction(conversation, responseFormat.schema),
+        appendJsonSchemaInstruction(conversation, jsonSchema ?? {}),
         undefined,
       );
     }
--- a/test/continuous-test-suite-json-e2e.ts
+++ b/test/continuous-test-suite-json-e2e.ts
@@ -341,7 +341,7 @@ const CELLS: Cell[] = [
 // Deliberately narrow: only infra-shaped statuses (429/5xx) are skippable —
 // generic 4xx / "bad request" would mask real request-construction bugs.
 function isInfraError(message: string): boolean {
-  return /api key|apikey|credential|security token|unauthor|permission|quota|rate.?limit|too many requests|429|not found|unknown model|model.*not|region|ENOTFOUND|ECONNREFUSED|ECONNRESET|socket hang|fetch failed|network|timeout|deadline|unavailable|overloaded|throttl|capacity|exhausted|billing|credits|insufficient|payment|402|access|forbidden|invalid.*model|does not exist|status (429|500|502|503|504)|internal server|service unavailable|max_tokens is too large|supports at most|completion tokens|maximum context|context length/i.test(
+  return /api key|apikey|credential|security token|unauthor|permission|quota|rate.?limit|too many requests|429|not found|unknown model|model.*not|timed.?out|region|ENOTFOUND|ECONNREFUSED|ECONNRESET|socket hang|fetch failed|network|timeout|deadline|unavailable|overloaded|throttl|capacity|exhausted|billing|credits|insufficient|payment|402|access|forbidden|invalid.*model|does not exist|status (429|500|502|503|504)|internal server|service unavailable|max_tokens is too large|supports at most|completion tokens|maximum context|context length/i.test(
     message,
   );
 }
@@ -671,5 +671,62 @@ await test("azure — an optional field survives on a second family member", asy
   );
   console.log("      · optional field honoured on a second family member");
 });
+
+// This suite sets NEUROLINK_DISABLE_BUILTIN_TOOLS at the top, so it cannot
+// reach the case below on its own: every OpenAI-compatible provider except
+// OpenAI and Azure suppresses `response_format` when the request carries
+// tools, and the caller's schema then never reaches the wire at all. Passing a
+// tool explicitly reproduces it regardless of the built-in setting.
+//
+// Before the fix the prompt-based fallback was gated on `response_format`
+// having been sent, so the suppressed case — the one the fallback exists for —
+// could never trigger it. The answer came back as prose with structuredData
+// undefined, silently breaking the documented generate({ schema }) contract.
+// Two providers, because the break was never deepseek-specific: it lives in
+// the shared `executeNativeGenerate` base that the whole OpenAI-compatible
+// family rides, and cohere showed the identical symptom before the fix.
+for (const [provider, model] of [
+  ["deepseek", "deepseek-chat"],
+  ["cohere", "command-r-plus-08-2024"],
+] as const) {
+  await test(`${provider}:${model} — a schema survives a request that also carries tools`, async () => {
+    let res: SchemaResult;
+    try {
+      res = (await nl.generate({
+        input: { text: "Capital of France and its population." },
+        provider,
+        model,
+        maxTokens: 300,
+        tools: pingTool,
+        schema: z.object({ capital: z.string(), population: z.number() }),
+      })) as SchemaResult;
+    } catch (e) {
+      if (isInfraError(String((e as Error)?.message ?? e))) {
+        throw new Skip("provider unavailable for provider reasons");
+      }
+      throw new Error("schema request with tools present failed outright", {
+        cause: e,
+      });
+    }
+    const data = res.structuredData as Record<string, unknown> | undefined;
+    assert(
+      !!data && typeof data === "object" && !Array.isArray(data),
+      "structuredData must be produced even when tools suppress response_format",
+    );
+    // Assert the WHOLE declared schema, not just one field. The recovery path
+    // can publish a syntactically valid object that still fails the schema, so
+    // checking only `capital` would let a half-restored contract pass — the
+    // schema declares both properties required.
+    assert(
+      typeof data?.capital === "string" &&
+        (data.capital as string).length > 0 &&
+        typeof data?.population === "number",
+      "both required schema properties must be present and correctly typed",
+    );
+    console.log(
+      "      · schema honoured on a tools-bearing request (prompt fallback)",
+    );
+  });
+}
 
 await runSuite();


src/lib/providers/openaiChatCompletionsBase.ts
```diff
@@ -1271,17 +1276,29 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
...
+    // wire. When tools suppress response_format the schema never ships at all,
+    // nothing instructs the model to emit JSON, and the answer comes back as
+    // prose with `structuredData` undefined — silently breaking the documented
+    // `generate({ schema })` contract on every OpenAI-compatible provider that
+    // suppresses (all of them except OpenAI and Azure, which opt out). That
+    // suppressed case is precisely what the prompt fallback exists for, yet
+    // the old `responseFormat !== undefined` guard made it unreachable there.
     if (
-      responseFormat !== undefined &&
+      jsonSchema !== undefined &&
```
Tara-ag:
```markdown
MINOR — redundant guard. `jsonSchema` is derived from `options.schema` (jsonSchema = options.schema ? convert(...) : undefined), so the new `jsonSchema !== undefined &&` already implies the `options.schema !== undefined &&` on the next line. The two checks are the same condition twice. Keying solely on the caller's schema reads cleaner and states the intent once:
```
````markdown
<!-- yama:json-schema-redundant-guard -->
```suggestion
    if (
      options.schema !== undefined &&
      !yieldsSchemaValidObject(loop.text, options.schema as ValidationSchema)
    ) {
```
````

File: test/continuous-test-suite-json-e2e.ts [hunk 2]
```diff
@@ -576,4 +576,47 @@ await test("anthropic:claude-sonnet-4-6 — forced truncation yields a partial o
...
+// reach the case below on its own: every OpenAI-compatible provider except
+// OpenAI and Azure suppresses `response_format` when the request carries
+// tools, and the caller's schema then never reaches the wire at all. Passing a
+// tool explicitly reproduces it regardless of the built-in setting.
+//
+// Before the fix the prompt-based fallback was gated on `response_format`
+// having been sent, so the suppressed case — the one the fallback exists for —
+// could never trigger it. The answer came back as prose with structuredData
+// undefined, silently breaking the documented generate({ schema }) contract.
+await test("deepseek:deepseek-chat — a schema survives a request that also carries tools", async () => {
```
Tara-ag:
```markdown
MINOR — coverage is narrower than the contract claim. The PR states `generate({ schema })` was broken on *every* OpenAI-compatible provider except OpenAI and Azure (deepseek, cohere, mistral, groq, xai, openrouter, ollama, ...), and the author's own before/after table shows cohere exhibiting the same `structuredData=UNDEFINED` symptom. The regression test only pins `deepseek`. A regression in the shared `executeNativeGenerate` base would slip through if it manifests on cohere/Mistral/groq but not deepseek. Consider widening to one more previously-broken provider (cohere is already measured in the PR description) — passing a tool explicitly reproduces the suppressed case regardless of `NEUROLINK_DISABLE_BUILTIN_TOOLS`, so this is cheap.
```

Tara-ag (commented):
## Yama review of #1693

The fix correctly closes the tools-suppress gap in `executeNativeGenerate`: `jsonSchema` is now computed before the wire decision and the prompt-based fallback is keyed on the caller's schema rather than on whether `response_format` shipped, so the previously-unreachable recovery finally fires for every OpenAI-compatible provider that suppresses `response_format` when tools are present. The wire `responseFormat` value is unchanged, so there is no behavior change for OpenAI/Azure. Verified: `suppressResponseFormatWithTools()` overrides return `false` only in OpenAI (`src/lib/providers/openAI/client.ts:128`) and Azure (`src/lib/providers/azureOpenai.ts:191`); the base returns `true`. The new fallback retries at most once (no loop), and the log message/`responseFormatSent` metadata are consistent with the suppressed case.

Tara-ag:
## Summary — NEEDS_WORK (low severity)

The core fix is correct and minimal: it makes the structured-output prompt fallback reachable for the tools-suppress case without changing `response_format` on the wire for OpenAI/Azure. Two non-blocking MINOR findings below; the change is well-justified and accurately measured.

|---|---|---|
| MINOR | `src/lib/providers/openaiChatCompletionsBase.ts:1288` | Redundant `jsonSchema !== undefined` guard — `options.schema !== undefined` on the next line already implies it |
| MINOR | `test/continuous-test-suite-json-e2e.ts:589` | Regression covers only `deepseek` vs a multi-provider contract break; suggest a second previously-broken provider |

**Checked and found clean:**
- Wire behavior unchanged — `responseFormat` is byte-for-byte identical before/after the refactor (the convert call is only hoisted, not duplicated).
- `suppressResponseFormatWithTools()` default is `true` on the base (`openaiChatCompletionsBase.ts:331`) and overridden to `false` only by OpenAI and Azure — confirming the authors' characterization of the blast radius.
- No infinite-retry risk — the fallback block runs a single retry, not a loop.
- The catch-based error-recovery path (lines 1255–1270) correctly remains keyed on `responseFormat !== undefined`; the suppressed case can't raise a tools↔schema conflict, so leaving it is correct.
- Test follows the file's conventions (`await test(...)`, `assert`, `Skip` for infra errors).
- API surface (Rule 5) untouched; no secrets/logging concerns; no dynamic-import or CLI/SDK boundary issues.

I could not read the sibling `executeStream`/SSE body-builder source from this sandbox, so I did not assert on it (impact analysis caveat per repo-conventions). If `generate({ schema })` is the only `structuredData`-producing entry point, the fix is complete for the reported contract.

src/lib/providers/openaiChatCompletionsBase.ts
```diff
@@ -1271,17 +1276,29 @@ export abstract class OpenAIChatCompletionsProvider extends BaseProvider {
...
     // case sailed through and handed the caller prose. Same recovery, keyed on
     // the result rather than on an exception.
+    // Keyed on the CALLER's schema, not on whether response_format reached the
+    // wire. When tools suppress response_format the schema never ships at all,
+    // nothing instructs the model to emit JSON, and the answer comes back as
+    // prose with `structuredData` undefined — silently breaking the documented
+    // `generate({ schema })` contract on every OpenAI-compatible provider that
+    // suppresses (all of them except OpenAI and Azure, which opt out). That
+    // suppressed case is precisely what the prompt fallback exists for, yet
+    // the old `responseFormat !== undefined` guard made it unreachable there.
```
Tara-ag, line 1286: **MINOR** — redundant guard; `jsonSchema` and `options.schema` are aliases here.

`jsonSchema` is derived directly from `options.schema` just above (`jsonSchema = options.schema ? convertZodToJsonSchema(...) : undefined`), so `jsonSchema !== undefined` is always `true` whenever the very next check `options.schema !== undefined` is. The second guard fully implies the first — the extra condition adds no protection.

```suggestion
    if (
      options.schema !== undefined &&
      !yieldsSchemaValidObject(loop.text, options.schema as ValidationSchema)
    ) {
```

If you prefer to keep the check for readability, that's fine too — but then the `jsonSchema` test is the one to keep (it is the value actually handed to `appendJsonSchemaInstruction`), and `options.schema !== undefined` becomes the redundant half. Either way, one of the two is dead.

test/continuous-test-suite-json-e2e.ts
```diff
@@ -576,4 +576,47 @@ await test("anthropic:claude-sonnet-4-6 — forced truncation yields a partial o
...
+  let res;
+  try {
+    res = await nl.generate({
```
Tara-ag: **MINOR** — regression test pins only `deepseek`, though the PR describes a multi-provider contract break.

The fix's own measurement table shows `cohere` exhibiting the identical `structuredData=UNDEFINED, prose` symptom before the change, and lists OpenAI/Azure as the only providers that do **not** suppress. Pinning a single provider leaves the other ~20 OpenAI-compatible providers (all of which suppress by default) untested for exactly the behavior this fix restores.

Suggested: loop the same assertion over a second previously-broken, non-suppressing-opt-in provider (e.g. `cohere`), so the regression would fail on both if the fallback regressed.

```suggestion
for (const provider of ["deepseek", "cohere"] as const) {
  await test(`${provider} — a schema survives a request that also carries tools`, async () => {
    let res;
    try {
      res = await nl.generate({
        input: { text: "Capital of France and its population." },
        provider,
        model: provider === "cohere" ? "command-r-plus" : "deepseek-chat",
        maxTokens: 300,
        tools: { ping: pingTool },
        schema: z.object({ capital: z.string(), population: z.number() }),
      });
    } catch (e) {
      if (isInfraError(String((e as Error)?.message ?? e))) {
        throw new Skip(`${provider} unavailable for provider reasons`);
      }
      throw new Error("schema request with tools present failed outright", {
        cause: e,
      });
    }
    const data = res.structuredData as Record<string, unknown> | undefined;
    assert(
      !!data && typeof data === "object" && !Array.isArray(data),
      "structuredData must be produced even when tools suppress response_format",
    );
    assert(
      typeof data?.capital === "string" && data.capital.length > 0,
      "the required string property must be present and non-empty",
    );
    console.log(
      "      · schema honoured on a tools-bearing request (prompt fallback)",
    );
  });
}
```

Note: the exact `cohere` model id here is a placeholder — use whatever the suite already uses elsewhere for Cohere. Non-blocking; the single-provider test does cover the regression.

Tara-ag (changes requested):
Two MINOR non-blocking findings on an otherwise-correct fix. Verdict: NEEDS_WORK (addressing the redundant guard and broadening the regression coverage would make this merge-ready).

murdore:
Both findings addressed.

**MINOR — redundant guard.** Correct, they were aliases. `jsonSchema` is derived directly from `options.schema`, so the two checks were the same condition twice. Now keyed solely on the caller's schema:

```ts
if (
  options.schema !== undefined &&
  !yieldsSchemaValidObject(loop.text, options.schema as ValidationSchema)
) {
```

**MINOR — coverage narrower than the contract claim.** Also correct, and the reasoning was the part worth taking: the break lives in the shared `executeNativeGenerate` base that the whole OpenAI-compatible family rides, so pinning one provider would let a regression surfacing on cohere but not deepseek slip through. The test now loops over **deepseek and cohere** — the two providers whose before/after measurements are in the commit. Both pass:

```
✓ deepseek:deepseek-chat — a schema survives a request that also carries tools
✓ cohere:command-r-plus-08-2024 — a schema survives a request that also carries tools
```

Also fixed the `test-shards (types)` failure: `pingTool` is already a record (`{ ping: tool(...) }`), so `tools: { ping: pingTool }` double-wrapped it. It is now `tools: pingTool`, matching the other cells in the suite.

One note for the record, since it is the reason this defect survived so long: this suite sets `NEUROLINK_DISABLE_BUILTIN_TOOLS=true` on line 32, so it structurally cannot reach the default configuration where twelve built-in tools are active — which is precisely the configuration that triggers the bug. The regression test passes a tool explicitly to get around that, but the suite-level blind spot deserves its own fix rather than being worked around in every future test.

Tara-ag:
## Summary — NEEDS_WORK (one MINOR, rest clean)

The PR was reduced to a single, well-justified regression test after being rebased onto the (now-merged) #1658: the `openaiChatCompletionsBase.ts` fix no longer lives here, and the diff is solely `test/continuous-test-suite-json-e2e.ts` (a cohere schema-with-tools cell + a one-token `isInfraError` extension). One new MINOR finding on the shared classifier; everything else verified clean. The core value of the PR — discriminating RED→GREEN live coverage for cohere, the one suppressing-family provider the matrix didn't already cover — is intact.

|---|---|---|
| MINOR | `test/continuous-test-suite-json-e2e.ts:369` | `timed.?out` added to the **shared** `isInfraError` classifier broadens skip behavior for the whole `CELLS` matrix (via `runCell`) and the vertex/anthropic/openai/azure tests, not just cohere. Semantically sound; suggest spelling-explicit `timed ?out`/`timed-out` and confirming the suite-wide broadening is intended. |

**Prior findings — resolved / now moot (not reposted):**
- MINOR — redundant guard in `openaiChatCompletionsBase.ts` → **moot**: that file is no longer part of this PR (fix moved to merged #1658).
- MINOR — single-provider coverage → **resolved**: test is cohere-only; deepseek is covered by #1658's breadth cell on the same path, so no redundant live call.
- CodeRabbit — full-schema assertion → **resolved** in `440066b` (asserts both `capital` non-empty string and `population` number).

**Checked and found clean:**
- Test follows file conventions (Rule 15): `await test(...)`, `assert`, `Skip` on `isInfraError`, `disableTools: false` + `tools: pingTool` (tool actually rides → real suppression path), `z`/`SchemaResult`/`nl` from the module graph. `check:tools-tests` and the types shard pass.
- The fix this test guards is merged (#1658 → release), so no cross-PR dependency; the test genuinely discriminates (RED without the fix, GREEN with it) per the PR's live measurement.
- No production code touched → no API surface (Rule 5), no dynamic import (Rule 1), no CLI/SDK leak (Rule 4), no secrets/logging concerns.
- The pre-existing `anthropic — forced truncation` failure on `release` is unrelated to this diff.

**Note:** PR is marked `mergeable_state: blocked` — that is the branch-protection/CI merge gate (a contract shard in progress), not a review-state issue.

Tara-ag (approved):
## Approve

Both MINOR findings from the prior review were addressed and verified against the committed code at head `efa25b8d`:

- **Redundant guard removed** — the fallback in `executeNativeGenerate` now keys solely on `options.schema !== undefined`.
- **Coverage broadened** — the regression now loops `deepseek` + `cohere`, and the `pingTool` double-wrap was fixed.

The change correctly closes the tools-suppress gap in the shared OpenAI-compatible base: the prompt fallback now fires keyed on the caller's schema, the converted schema is available even when `response_format` is suppressed, and the wire value is unchanged for OpenAI/Azure (no behavior change there). No infinite-retry risk. API surface untouched (Rule 5 clean), no dynamic imports (Rule 1), no CLI/SDK leak (Rule 4), no secrets/logging concerns. Merge-ready.

Tara-ag:
@murdore — both findings from the prior review are confirmed addressed in the committed code at head `efa25b8d` and both review threads are now resolved. Summary updated to APPROVE; the approving review has been submitted. Thanks for the clean follow-up.

Tara-ag (approved):
## Yama review of #1693 — APPROVE

Recurring review reconfirmed on the current head `46891d23`. The diff from base matches exactly what was already verified and approved; both MINOR findings from the prior pass remain resolved here, and no new findings emerge from the fresh inspection.

### Prior findings — still resolved on current head

| Finding | State | Verification on `46891d23` |
|---|---|---|
| MINOR — redundant `jsonSchema !== undefined` guard (`openaiChatCompletionsBase.ts`) | ✅ Resolved (thread closed) | Fallback now keys solely on `options.schema !== undefined`; `jsonSchema` is hoisted. |
| MINOR — single-provider regression coverage (`continuous-test-suite-json-e2e.ts`) | ✅ Resolved (thread closed) | Loops `deepseek:deepseek-chat` + `cohere:command-r-plus-08-2024`; `tools: pingTool` correctly not double-wrapped. |

The author's reply (murdore, #issuecomment-5648533650) adequately justifies both — confirmed by reading the committed code, so nothing is reposted.

### What the change does (still correct)
- `jsonSchema` computed unconditionally from `options.schema`, so the prompt-based structured-output fallback has something to fall back *with* when tools suppress `response_format`.
- `responseFormat` on the wire is byte-for-byte unchanged for OpenAI/Azure (`suppressResponseFormatWithTools()` returns `false` only there; the base returns `true` at `openaiChatCompletionsBase.ts:331`). The convert call is hoisted, not duplicated.
- Fallback now fires for every OpenAI-compatible provider that suppresses — exactly the case the fallback was written for.
- New log metadata `responseFormatSent` is a boolean; no schema verbatim enters the log.
- Single retry, not a loop — no infinite-retry risk; the concurrent catch-block recovery still keys on `responseFormat !== undefined` (correct: it only fires when `response_format` was actually sent and rejected).

### New, non-blocking observation (no inline comment)
In the fallback, `appendJsonSchemaInstruction(conversation, jsonSchema ?? {})` — since the branch guard `options.schema !== undefined` already guarantees `jsonSchema` is defined (and `convertZodToJsonSchema` returns a truthy object), the `?? {}` branch is unreachable. Harmless defensive code, not a bug; noted only for completeness.

### Impact / blast radius
Change is confined to `executeNativeGenerate` in the shared OpenAI-compatible base. Public SDK surface untouched (Rule 5), no dynamic imports (Rule 1), no CLI/SDK leak (Rule 4), no secrets/logging concerns. The sibling `executeStream` path has its own body-builder and is not affected by this change.

Cursor note: the standing `<!-- yama:summary -->` issue comment reflects this APPROVE already; since review-state on a moved head needs a fresh approval, this review is submitted against `46891d23`.

Merge-ready.

Tara-ag:
Reconfirmed on current head `46891d23` and re-submitted the **APPROVE** review against that commit (the PR head moved since the prior approval, so a fresh review was needed). Both MINOR findings from the earlier pass remain resolved in the committed code, no new findings surfaced.

The standing `<!-- yama:summary -->` comment already reflects this disposition and both review threads are closed, so no duplicate summary is posted here. One genuinely non-blocking observation, for the record rather than as a formal finding: in `executeNativeGenerate` the fallback calls `appendJsonSchemaInstruction(conversation, jsonSchema ?? {})`, and since the branch guard `options.schema !== undefined` already guarantees `jsonSchema` is defined, the `?? {}` fallback is unreachable — harmless defensive code.

What was checked and found clean on this pass: wire `response_format` unchanged for OpenAI/Azure (convert hoisted, not duplicated), fallback fires keyed on caller schema, single-retry (no infinite loop), log metadata is a boolean (no schema leak), and the regression test loops `deepseek` + `cohere` with `tools: pingTool` following the suite's conventions (Rule 15). API surface (Rule 5), dynamic imports (Rule 1), CLI/SDK boundary (Rule 4) all clean. Merge ready.

Tara-ag (approved):
## Approve

Both MINOR findings from the prior pass are resolved in the committed code at current head `46891d23`:

- **Redundant guard removed** — the fallback in `executeNativeGenerate` now keys solely on `options.schema !== undefined`.
- **Coverage broadened** — the regression loops `deepseek` + `cohere`, and `pingTool` is no longer double-wrapped.

The fix correctly closes the tools-suppress gap in the shared OpenAI-compatible base: the prompt-based fallback now fires keyed on the caller's schema, the converted schema is computed even when `response_format` is suppressed, and the wire `responseFormat` value is unchanged for OpenAI/Azure (no behavior change there). No infinite-retry risk (single retry). API surface untouched (Rule 5), no dynamic imports (Rule 1), no CLI/SDK leak (Rule 4), no secrets/logging concerns (the new log field is a boolean). Merge-ready.

This fresh approval is anchored to head `46891d23` (the prior approval was on an earlier `efa25b8d`).

murdore:
Both addressed in `440066bfd` — the CodeRabbit finding, plus a flaky-test cause it exposed while I was verifying the fix.

**Assert the full declared schema.** Valid, and it is the same class of weakness the human reviewer caught on #1691's optionality test: the assertion did not check the guarantee it claimed. The schema declares `capital` **and** `population` required, so `{"capital":"Paris"}` would have passed while violating it — a half-restored contract could have gone green. Now:

```ts
typeof data?.capital === "string" &&
  (data.capital as string).length > 0 &&
  typeof data?.population === "number"
```

**A second issue this surfaced.** Running the strengthened test live, cohere failed with:

```
[cohere] Request timed out: cohere generate operation timed out after 30000
```

That is transient infra, and `isInfraError` is meant to downgrade it to a skip — but its pattern matches `timeout` while the message says `timed out`. So a vendor blip was being reported as a hard JSON-validity failure. Added `timed.?out` to the classifier; cohere now passes rather than flaking.

Verified live on both cells after the change:

```
✓ deepseek:deepseek-chat — a schema survives a request that also carries tools
✓ cohere:command-r-plus-08-2024 — a schema survives a request that also carries tools
```

`check:tools-tests` clean. The one remaining failure in that run — `anthropic — forced truncation yields a partial object` — is pre-existing on `release` (`finishReason=stop`, i.e. the model never truncated) and unrelated to this diff, which touches only the OpenAI-compatible base that Anthropic does not use.

Also rebased onto `1dc704049` after #1691 merged. The only conflict was this test file, where both PRs appended tests before `runSuite()`; resolved as the union of both blocks, since neither replaces the other.

Tara-ag:
Re-verified on current head `440066b` (recurring pass). No new findings — the standing APPROVE stands.

**Delta since the last pass (`440066b`), both confirmed clean:**
- `continuous-test-suite-json-e2e.ts`: `isInfraError` now also matches `timed.?out` (specifically the Cohere `…operation timed out after 30000` message), which is additive — `timeout` is still in the alternation. This closes a flaky-skip gap rather than masking a real failure.
- The regression now asserts the **whole** declared schema (`capital` non-empty string **and** `population` number), so `{"capital":"Paris"}` alone can no longer pass — the CodeRabbit weakness (and the #1691 optionality blind spot of the same class) is covered.

**The one standing impact caveat from earlier passes is now resolved:** I read the sibling `executeStream`/`runStreamLoop`/`streamOneStep` path directly. `executeStream` never produces `structuredData` (it returns a `StreamResult`, not a schema-validated object), and `nl.generate({ schema })` routes strictly through `generate → executeNativeGenerate` — the method this PR fixes — passing `responseFormat` into `runLoop` in both the direct and stream-aggragation cases. So no `structuredData`-producing entry point is left unhandled; the fix fully covers the `generate({ schema })` contract.

**Blast radius (code graph):** the changed fallback lives in the shared `executeNativeGenerate` base, reached by `generate` → `gen`/`generateText`/`executeFakeStreaming` in `baseProvider.ts` plus the providers that don't override `executeNativeGenerate` (SageMaker and Anthropic override it and are untouched). Existing callers rely on `generate`'s public return shape, which is unchanged (Rule 5). No dynamic imports (Rule 1), no CLI/SDK leak (Rule 4), no secrets/logging concerns (the added log field is a boolean).

Nothing to fix; merge ready.

Tara-ag (approved):
## Approve — re-anchored to current head `440066b`

The head moved since the last approval (`46891d23` → `440066b`), so this is a fresh approval on the current commit. That commit adds exactly two things, both improvements that keep the fix correct:

- **Full-schema assertion** (CodeRabbit finding, now `population` is asserted as a number alongside non-empty `capital`) — closes the gap where `{"capital":"Paris"}` could pass while omitting the required `population`.
- **`timed.?out` added to `isInfraError`** — a Cohere `"… timed out"` string previously fell through and reported a vendor blip as a hard JSON-validity failure; the classifier now downgrades it to a skip. Additive (`timeout` retained), no masking of real shape failures.

Both are verified consistent with the prior approvals. No new findings. The two MINOR findings from the earlier pass remain resolved in the committed code, and the CodeRabbit assertion thread is resolved. Public SDK surface (Rule 5), dynamic imports (Rule 1), CLI/SDK boundary (Rule 4), secrets/logging all clean. The standalone `anthropic — forced truncation` failure is pre-existing on `release` and outside this diff (the change touches only the OpenAI-compatible base). Merge-ready.

Tara-ag:
@murdore — thanks for the clean follow-up in `440066b`. Both changes are reviewed and confirmed sound:

- **Full-schema assertion** — correct call, and the same class of weakness you cite; asserting `population` as a number now guards the whole declared contract.
- **`timed.?out` in `isInfraError`** — good catch and appropriately scoped. Adding the variant (rather than replacing `timeout`) means a real shape failure still surfaces while the Cohere vendor blip correctly degrades to a skip.

The rebased resolution of the `continuous-test-suite-json-e2e.ts` overlap (union of both blocks, neither replaced) matches what I see on the branch. Review kept at **APPROVE**, re-submitted against `440066b`. The standalone `anthropic — forced truncation` failure is pre-existing on `release` and unrelated to this diff. Nothing further needed from this side.

murdore:
Heads-up on a substantive change to this PR, since you approved the earlier version.

While closing out #1658 I found that these two PRs repair the same defect, and that they cannot both land as written — they overlap on the same fourteen lines of `openaiChatCompletionsBase.ts`, and the overlap is not a textual conflict that can be resolved either way.

The specific problem: this PR widened the outcome-path guard (line 1300) to cover providers that suppress `response_format`. #1658 instead handles that case at line 1329 with a **tools-free** re-ask (`withTools = false`). Line 1300 runs first, and it re-asks with tools still attached — which is the counter-failure `runLoop`'s own comment documents: groq reads the spelled-out JSON Schema as another tool and calls one named `JSON`, which the server rejects. So layering this PR on top of #1658 would catch the suppressed case at the wrong site, with tools attached, and make #1658's correct handler unreachable for it.

#1658 also already does this PR's refactor half — it hoists `schemaJson` (line 1178) and computes `responseFormatSuppressed` once (line 1187).

So I have rebased this PR onto #1658 and reduced it to the part that is genuinely additive: the live cohere case. `CohereProvider` extends the same base and does not override `suppressResponseFormatWithTools`, so it takes the suppressing default and really does exercise the repaired path, and cohere is in no cell of the json-e2e matrix. Two corrections went in with it:

- the deepseek row is dropped — #1658 added deepseek as a breadth cell running `plain-prose-prompt` **with tools**, so it already covers that provider on this path;
- `disableTools: false` is now passed explicitly, as every other tools-bearing test in that file does. Without it the tool may not ride along, the turn takes the ordinary `response_format` route, and the test passes while asserting nothing.

Evidence is in the PR description: a discriminating RED → GREEN against the live cohere API, reverting only `openaiChatCompletionsBase.ts` to #1658's base and restoring it.

Happy to restore any of this if you would rather keep the fix here and trim #1658 instead — the decision of which PR owns the repair is yours and @sachiny09's, not mine. Nothing has been merged.

test/continuous-test-suite-json-e2e.ts
```diff
@@ -368,7 +368,7 @@ const CELLS: Cell[] = [
 // Deliberately narrow: only infra-shaped statuses (429/5xx) are skippable —
 // generic 4xx / "bad request" would mask real request-construction bugs.
```
Tara-ag, line 369: **MINOR** — this `timed.?out` addition broadens the *shared* infra classifier, not just the cohere path.

`isInfraError` is called from `runCell` (drives the **entire `CELLS` provider matrix**, line 392) plus the vertex/anthropic/openai/azure/cohere tests. The comment above the function explicitly warns it is *"deliberately narrow"* so that generic 4xx / "bad-request" strings cannot mask real request-construction bugs. Adding `timed.?out` (note the free `.`, which also matches `timed-out`, `timed/out`, etc.) makes every matrix cell now skip on `…operation timed out…`.

The semantic intent is sound — a provider "timed out" is infrastructure, consistent with the pre-existing `timeout` and `deadline` alternates — and it's hard to construct a *product*-failure message that contains `timed out` but isn't a timeout. So this is low-risk. The only ask: since this deliberately-narrow shared classifier now behaves differently for **all providers** (not just cohere), confirm that's intentional, and consider scoping the `.?` (e.g. `/timed ?out|timed-out/`) so the match is spelling-explicit rather than "any single char".

No behavior change to the wire; the reduced PR otherwise stays clean.

Tara-ag (commented):
## Yama review of #1693 — reduced-scope cohere regression test

Recurring pass on the current head `2c010b65`. This PR has changed materially since the prior approvals: it was rebased onto #1658, the `openaiChatCompletionsBase.ts` fix was **removed** (now owned by the separately merged #1658), and the diff is reduced to a single commit touching only `test/continuous-test-suite-json-e2e.ts` (a cohere regression cell + the `isInfraError` extension). One MINOR inline finding on the classifier broadening; everything else verified clean.

Tara-ag (changes requested):
## NEEDS_WORK — one MINOR (non-blocking clarification), rest clean

Recurring review on the current head `2c010b65`. The PR is now a single test-only commit (cohere schema-with-tools regression cell + the `timed.?out` `isInfraError` extension) rebased on the merged #1658. The change is sound and discriminating (RED→GREEN measured live); one MINOR stands on the shared `isInfraError` classifier and needs the author to confirm intent / optionally scope the regex before this is chalked up as fully clean.

|---|---|---|
| MINOR | `test/continuous-test-suite-json-e2e.ts:369` | `timed.?out` broadens the **shared** infra classifier (whole `CELLS` matrix via `runCell` + vertex/anthropic/openai/azure tests), not just cohere. Low-risk and semantically consistent with `timeout`/`deadline`, but confirm the suite-wide broadening is intentional and consider a spelling-explicit `timed ?out`/`timed-out` so the match isn't "any single char". |

**Why this disposition:** the finding is a deliberate-classifier question rather than a code defect, and the author has not yet replied to it (it was posted after the last author comment). A one-line confirmation, or scoping the `.` to ` ?`/`-`, resolves it — no restore of the removed `openaiChatCompletionsBase.ts` fix is needed: that half is correctly owned by the now-merged #1658, and restoring it here would reintroduce the layering defect. #1658 merged 2026-09-16, so no cross-PR dependency remains.

**Checked and found clean:** reduced scope matches the PR description; `isInfraError` broadening is additive (`timeout` retained, no masking of real shape failures); full-schema assertion (CodeRabbit finding, `440066b`) still holds; `disableTools: false` + `tools: pingTool` correctly rides the tool so the suppression path is genuinely exercised; deepseek row correctly dropped (covered by #1658). Test follows Rule 15 conventions. No production code touched → Rule 1/4/5 clean, no secrets/logging concerns.

(Requested changes is deliberately a lightweight, answerable-as-written state; it maps to the NEEDS_WORK verdict in the standing summary. The standalone `anthropic — forced truncation` failure is pre-existing on `release` and unrelated.)

Tara-ag:
@murdore — thanks for the thorough write-up in #issuecomment-5703101423 (and in the PR body). Confirmed on my side:

- **No restore needed.** #1658 merged 2026-09-16, so the ownership question is settled — the `openaiChatCompletionsBase.ts` fix lives there, and trimming this PR to the genuinely-additive cohere case was the right call. Restoring the outcome-path guard here would reintroduce the tools-still-attached `JSON`-tool counter-failure you documented, so I would not revert it.
- **Reduced scope reviewed independently.** The cohere cell (disableTools: false + tools: pingTool) and the `timed.?out` extension both verified against the diff. The one open MINOR is the shared-`isInfraError` broadening on `t…:369` — it's low-risk and semantically consistent, so this is just a confirm-it's-intentional (and optionally scope the `.`) rather than a change request for the cohere test itself.

Nothing further needed from this side beyond that one clarification.

test/continuous-test-suite-json-e2e.ts
murdore: Confirmed intentional, and scoped as you asked.

You are right that this widened a *shared* classifier: `isInfraError` drives the whole `CELLS` matrix plus the vertex/anthropic/openai/azure/cohere tests, not just the cohere cell the alternate was added for — on a function whose own comment calls it deliberately narrow.

The semantic intent stands (a provider that timed out is infrastructure, consistent with the existing `timeout` and `deadline` alternates), so the alternate stays. The free `.` does not: it is now `timed[ -]?out`, which is spelling-explicit rather than "any single character".

| input | `timed.?out` | `timed[ -]?out` |
|---|---|---|
| `operation timed out` | match | match |
| `request timed-out` | match | match |
| `timedout` | match | match |
| `the request timedXout after 30s` | match | **no** |
| `timed/out` | match | **no** |

PR #1695, opened 2026-09-12
test(proxy): cover a routed alias that carries no provider

murdore:
Follow-up to **#1688**, which merged before this case was added.

## What is missing today

#1688 made `classifyGrokProxyModel` derive a routed alias's backend and window from the mapping's `to` and `provider` rather than from the alias name, and shipped two regression cases for it. Both pass `provider: "anthropic"`, so the branch that runs when a mapping carries **no** provider was never exercised. CodeRabbit asked for exactly this case on the original PR; it did not make it in before the merge.

## Why it matters

`provider` is optional in a routing mapping. With it absent, the backend has to be inferred from the target id, because the alias name carries no clue — `enterprise-sonnet` says nothing about Anthropic. A classifier that gave up there would route it to `chat_completions` with the wrong context window and no adaptive thinking, which is the exact class of bug #1688 set out to fix.

## Evidence it discriminates

Not just "the suite is green". Replacing `providerFromMapping`'s final `providerForModelId(fallbackId)` with a flat `"openai"` turns this case red on its own message:

```
a provider-less alias did not infer the backend from its target
✗ Proxy clients: Grok catalog windows and backends match upstream
```

With the shipped code in place the suite goes **106 → 107 passing**.

Test-only change; no source is touched.

## Summary by CodeRabbit

* **Tests**
  * Expanded automated coverage for proxy model classification when provider information is unavailable.
  * Verified correct backend selection, context-window handling, and adaptive-thinking support based on the target model.

### DIFF
--- a/test/continuous-test-suite-proxy.ts
+++ b/test/continuous-test-suite-proxy.ts
@@ -2182,6 +2182,40 @@ async function testGrokCatalogWindowsAndBackends(): Promise<boolean> {
     log("claude-sonnet-4-thinking used the alias window instead of 1M", "red");
     return false;
   }
+  // `providerFromMapping` recognises anthropic/claude, openai and
+  // vertex/google/google-ai/gemini; any other string falls through to
+  // inferring the backend from the *target* id. That branch is reachable from
+  // a real config: `ModelMapping.provider` is a required string and
+  // `parseRoutingConfig` coerces an absent or empty one to "anthropic", so
+  // neither ever arrives here — but any other value is passed through
+  // verbatim, so a mapping naming a provider this writer does not know about
+  // lands in the fallback intact. The alias name carries no clue on its own:
+  // "enterprise-sonnet" says nothing about Anthropic.
+  const unknownProvider = __grokTestHooks.classifyGrokProxyModel(
+    "enterprise-sonnet",
+    {
+      from: "enterprise-sonnet",
+      to: "claude-sonnet-4-6",
+      provider: "bedrock",
+    },
+  );
+  if (unknownProvider.apiBackend !== "messages") {
+    log("an unrecognised provider did not fall back to the target id", "red");
+    return false;
+  }
+  if (unknownProvider.supportsReasoningEffort !== true) {
+    log("an unrecognised provider lost the target's adaptive thinking", "red");
+    return false;
+  }
+  // The window is deliberately NOT derived the same way. `windowProvider`
+  // prefers the declared provider precisely so a Bedrock-hosted Claude gets
+  // Bedrock's 200K window rather than Anthropic-direct's 1M — the backend
+  // comes from the target, the window from where it is actually served.
+  // Asserting it here pins that split, which is the surprising half.
+  if (unknownProvider.contextWindow !== 200_000) {
+    log("the window was not looked up under the declared provider", "red");
+    return false;
+  }
   return true;
 }
 


test/continuous-test-suite-proxy.ts
```diff
@@ -2182,6 +2182,34 @@ async function testGrokCatalogWindowsAndBackends(): Promise<boolean> {
...
+  // to be inferred from the *target* id, not from the alias — an alias called
+  // "enterprise-sonnet" carries no clue that it is Anthropic, so a classifier
+  // that gave up here would land it on chat_completions with the wrong window
+  // and no adaptive thinking.
+  const noProvider = __grokTestHooks.classifyGrokProxyModel(
+    "enterprise-sonnet",
+    {
+      from: "enterprise-sonnet",
+      to: "claude-sonnet-4-6",
+      provider: "",
```
Tara-ag: **MINOR** — the test's stated premise does not match how `ModelMapping.provider` actually behaves, and the input can't occur through the real config loader.

The new block drives `classifyGrokProxyModel("enterprise-sonnet", { from, to, provider: "" })` and the comment claims *"`provider` is optional in a routing mapping."* That invariant is not what the code enforces:

- `ModelMapping.provider` is a **required** `string` (`src/lib/types/subscription.ts`), not optional.
- `parseRoutingConfig()` **defaults an absent `provider` to `"anthropic"`** — `src/lib/proxy/proxyConfig.ts`: `const provider = String(m.provider ?? "anthropic").trim() || "anthropic";`. An omitted or empty `provider` can therefore never survive the loader; it arrives as `"anthropic"` and is classified correctly via the first branch of `providerFromMapping`.

So the scenario the comment describes — *"a provider-less alias would land on `chat_completions` with no adaptive thinking"* — cannot occur from a real config. The fallback the test actually covers is the genuinely-live branch in `providerFromMapping` (`src/cli/proxy-clients/grok.ts` ~lines 114–116) reached only by a **non-empty unrecognized** provider string (e.g. `provider: "mistral"`), not by an absent/empty one.

This is still valuable regression coverage for that #1688 branch, so I'd keep the test — but the framing codifies a false invariant:

```suggestion
  // `provider` is required by ModelMapping (src/lib/types/subscription.ts) and
  // the config loader defaults an absent one to "anthropic"
  // (parseRoutingConfig in src/lib/proxy/proxyConfig.ts), so an empty value
  // cannot reach here from a real config. What this case covers is the live
  // fallback in providerFromMapping: an *unrecognized* provider string drops
  // through to providerForModelId(to), which keeps the alias on the target's
  // backend rather than guessing from the alias name.
  const unknownProvider = __grokTestHooks.classifyGrokProxyModel(
    "enterprise-sonnet",
    {
      from: "enterprise-sonnet",
      to: "claude-sonnet-4-6",
      provider: "", // empty string — models the unrecognized-provider fallback
    },
  );
```

Also consider renaming `noProvider` → `unknownProvider` throughout the block, and note the assertion expectations (backend `messages`, window `1_000_000`, `supportsReasoningEffort === true`) are correct against the source, so the assertions themselves need no change.

Tara-ag (commented):
Test-only review; one MINOR comment inline about the test's framing. The assertions themselves are correct against the source.

Tara-ag:
**Verdict: APPROVE** — test-only change that adds genuine regression coverage for the `providerFromMapping` fallback branch from #1688; the assertions are correct against the source, and no production code is touched. One MINOR comment on the test's framing.

## Findings

| MINOR | `test/continuous-test-suite-proxy.ts:2185-2195` | Comment claims "`provider` is optional in a routing mapping", but `ModelMapping.provider` is required and `parseRoutingConfig()` defaults an absent/empty provider to `"anthropic"`. The scenario described can't occur from a real config; the covered branch is the live fallback reached by an *unrecognized* provider string. Suggest reframing the comment and renaming `noProvider` → `unknownProvider`. |

## What I checked (and found clean)

- **Assertions are correct against the source.** Traced `classifyGrokProxyModel("enterprise-sonnet", {to:"claude-sonnet-4-6", provider:""})` through `providerFromMapping`/`providerForModelId`/`supportsAdaptiveThinking` (`src/cli/proxy-clients/grok.ts`): provider falls through to `providerForModelId("claude-sonnet-4-6")` → `"anthropic"` → backend `"messages"`; `windowProvider` falls back to provider → `getContextWindowSize("anthropic","claude-sonnet-4-6")` → 1M; `supportsAdaptiveThinking("claude-sonnet-4-6")` → `true`. All three assertions match.
- **Impact / blast radius.** Test-only file; `classifyGrokProxyModel` callers are `catalogGrokSpecs` (production, unchanged) and this test. No production behaviour altered; no Rule 1/3/4/5/6 concern.
- **Project conventions.** Sits in the E2E `continuous-test-suite-proxy.ts` (Rule 15 style), drives the existing `__grokTestHooks` test-export (determinism exception already declared in `grok.ts`), and runs in-process (`proxy-config` ∈ `IN_PROCESS_CATEGORIES`). No secret, injection, or hot-path (baseProvider/registry) exposure.
- The "106 → 107 passing" count is consistent with a single added case.

The one flagged item is framing/naming in a test, not a defect — nothing blocking.

test/continuous-test-suite-proxy.ts
```diff
@@ -2182,6 +2182,34 @@ async function testGrokCatalogWindowsAndBackends(): Promise<boolean> {
...
+    },
```
Tara-ag: **MINOR** — the new assertions are correct and worth keeping; the comment and the variable name describe a branch that cannot actually be reached from a real config.

The test calls `classifyGrokProxyModel` with `provider: ""`. But `ModelMapping.provider` is **required**, and `parseRoutingConfig()` treats an absent/empty provider as `"anthropic"` when building a routing mapping. So a provider-less entry can't arise from a parsed config — the branch that genuinely lives here is the `providerFromMapping` fallback, which only fires for a non-empty *unrecognized* provider string (an alias whose id itself parses as a known model).

As written, this case is exercising the classifier directly (a deterministic table-driven decision), so the coverage still guards the fallback — good. Just the framing overstates what it guards:

- The comment claims "`provider` is optional in a routing mapping" and that a provider-less alias would be "landed on chat_completions". Neither can happen via `loadProxyConfig`; the dead scenarios are the ones it's describing.
- `noProvider` / "a provider-less alias" names read as if this removes the provider, when the value actually passed is the empty-string → Anthropic default path.

Suggest reframing the comment to describe the `providerFromMapping` branch and renaming `noProvider` → `unknownProvider` (feed it a non-empty unrecognized provider like `"my-grok-backend"`), which covers the fallback precisely.

```suggestion
  // `provider` is required in a routing mapping; an absent or empty value
  // defaults to `"anthropic"` in parseRoutingConfig(). The fallback that has
  // no explicit provider is reached only when the alias's id itself parses as
  // a known model — a classifier that gave up here would land it on
  // chat_completions with the wrong window and no adaptive thinking.
  const unknownProvider = __grokTestHooks.classifyGrokProxyModel(
    "enterprise-sonnet",
    {
      from: "enterprise-sonnet",
      to: "claude-sonnet-4-6",
      provider: "unrecognized-backend",
    },
  );
```

Tara-ag (approved):
Verdict: APPROVE — the added third case correctly covers the provider-less/fallback routing branch with accurate assertions, and the only finding is an optional MINOR test-framing tweak.

Tara-ag:
## Verdict: APPROVE

Test-only change (`test/continuous-test-suite-proxy.ts`) adding a third case to `testGrokCatalogWindowsAndBackends`, covering the routing fallback branch that carries no explicit provider. The assertions match the production logic and the coverage is genuinely new value from #1688. No blocking issues.

### Findings

| MINOR | `test/continuous-test-suite-proxy.ts:2185-2196` | Comment + `noProvider` name describe a branch that can't arise from a real config (`provider` is required; empty defaults to `"anthropic"`). The live branch is the `providerFromMapping` fallback. Reframe comment / rename to `unknownProvider` with a non-empty unrecognized provider. |

### Checked and clean

- **Correctness of assertions** — all three (`apiBackend === "messages"`, `contextWindow === 1_000_000`, `supportsReasoningEffort === true`) verified against `src/lib/proxy/grok.ts` `classifyGrokProxyModel` and `ModelMapping`.
- **Test suite conventions** (CLAUDE.md rule 15) — this is a deterministic in-process case under the declared determinism exception; the `__grokTestHooks` import pattern matches sibling cases in the same file.
- **No CRITICAL/security concerns** — no secrets, no logging of credentials, no unsafe input handling introduced.
- **Single commit, conventional message** — already validated by the single-commit bot.

No other findings; silence in the table above is absence, not omission.

Tara-ag:
**Verdict: APPROVE** — test-only change that now correctly covers the live `providerFromMapping` fallback branch from #1688. All prior review findings were addressed in code; verified each assertion against the source. One MINOR residual: the PR title/description still carry the stale "no provider" framing.

## Findings

| MINOR | PR title/description (not a diff line) | Title still says "carries no provider"; body still argues "`provider` is optional in a routing mapping." The shipped code passes `provider: "bedrock"` — a *present* provider the classifier doesn't recognize, not an absent one. `ModelMapping.provider` is required and `parseRoutingConfig` defaults an absent/empty value to `"anthropic"`. Align the title/description with what the test actually does. |

## Prior findings — addressed, not re-posted

- **`yama:provider-less-premise-001`** (MINOR, outdated thread — resolved this round) and **`yama:grok-noprovider-comment`** (already resolved): both flagged that feeding `provider: ""` described a branch that can't arise from a real config. The author fixed this in code: the fixture now uses `provider: "bedrock"` (non-empty unrecognized provider), renamed `noProvider` → `unknownProvider`, corrected the window assertion from 1M to 200K, and rewrote the in-code comments to correctly state that `provider` is required and defaults to `"anthropic"`.
- **CodeRabbit's** empty-string suggestion was overtaken by this same, better fix.

## Verified correct against source

- `apiBackend === "messages"` — `providerFromMapping("bedrock", "claude-sonnet-4-6")` falls through to `providerForModelId("claude-sonnet-4-6")` → `"anthropic"` (`src/cli/proxy-clients/grok.ts`).
- `supportsReasoningEffort === true` — `supportsAdaptiveThinking("claude-sonnet-4-6")` matches `/^claude-(opus|sonnet)-4-6$/`.
- `contextWindow === 200_000` — `getContextWindowSize("bedrock", "claude-sonnet-4-6")`: dynamic registry uninitialized (thrown-and-caught), `resolveManifestEntryStrict("bedrock", …)` = `undefined` (bedrock manifest has no exact/alias match), static table falls to bedrock `_default` = **200_000** (`src/lib/constants/contextWindows.ts`). Confirmed `"claude-sonnet-4-6"` is not a bedrock key nor a prefix of one.

## Impact / conventions

- **Blast radius:** test-only; production `grok.ts` untouched. `classifyGrokProxyModel` callers are exactly `catalogGrokSpecs` (unchanged) and this test (graph-verified). No Rule 1/3/5/6 concern.
- **Rule 15:** sits in the E2E `continuous-test-suite-proxy.ts`, driven via the `__grokTestHooks` test-export under the declared determinism exception; registered as `proxy-config` (in-process). No secrets, no logging changes.

Silence in the table above is absence of further findings.

test/continuous-test-suite-proxy.ts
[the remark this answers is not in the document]
murdore: Addressed at `74d2dea09`, though not by the suggested edit — that one does not compile.

`ModelMapping.provider` is a **required** `string` (`src/lib/types/subscription.ts:1185-1189`), so omitting the property fails the typecheck rather than exercising a missing-field path. There is no optional-property branch to cover.

Your underlying point was right, and Tara-ag's parallel finding made the same case from the loader side: `provider: ""` was not a realistic input either. `parseRoutingConfig` coerces it away —

```ts
// src/lib/proxy/proxyConfig.ts:562
const provider = String(m.provider ?? "anthropic").trim() || "anthropic";
```

— so neither an absent nor an empty provider can reach `classifyGrokProxyModel`. Both arrive as `"anthropic"` and take the first branch.

The genuinely reachable input for the fallback is a provider string the writer does not recognise, since any other value passes through verbatim. The case now uses `provider: "bedrock"`.

Reaching it honestly also exposed something worth pinning, which the empty-string version had hidden: backend and window come from different places on purpose. `apiBackend` follows the target id, while `windowProvider` prefers the **declared** provider — so Bedrock-hosted Claude Sonnet 4.6 is correctly sized at Bedrock's 200K, not Anthropic-direct's 1M. The test asserts both halves, and each fails on its own mutation:

```
return providerForModelId(fallbackId) -> return "openai"
  ✗ an unrecognised provider did not fall back to the target id

windowProvider = mapping.provider ?: provider -> windowProvider = provider
  ✗ the window was not looked up under the declared provider
```

murdore:
Follow-up checked and closed out — nothing to carry.

Tara-ag's remaining note was that the title says "carries no provider" while the shipped test sends `provider: "bedrock"` (a recognised string the writer does not map). The commit message was already reworded correctly to "whose provider the writer does not recognise", and the commit message is what lands — the PR title does not survive the merge.

Recorded here rather than moved to https://github.com/juspay/neurolink/pull/1728, which carries only the three items that describe durable code.

PR #1696, opened 2026-09-13
fix(huggingface): stop suppressing tools for 141 of 142 served models

murdore:
`HuggingFaceProvider.supportsTools()` overrode the base's registry-driven capability check with a hardcoded allowlist of 13 model-name substrings. Any model whose name did not contain one of them ran **tool-free** — tools were never put on the wire, so a registered tool could never be called.

## How stale the allowlist is

Measured against the 143 models the router actually serves:

| | |
|---|---|
| allowlist entries matching ANY served model | **1 of 13** |
| served models the allowlist ADMITS | **1 of 143** |
| served models the allowlist BLOCKS | **142** |

## The vendor disagrees with the allowlist

Asked directly, the router returns `tool_calls` for models the allowlist blocks:

```
Qwen/Qwen2.5-72B-Instruct       200  TOOL_CALLED
zai-org/GLM-5.3-Flash           200  TOOL_CALLED
google/gemma-4-31B-it           200  TOOL_CALLED
deepseek-ai/DeepSeek-V4-Flash   200  TOOL_CALLED
```

**No served model rejected the `tools` field.** The two that failed a tools request failed with `"not supported by any provider you have enabled"` — model availability, unrelated to tools.

## Why the override existed, and why it can go

It was written for the old per-model Inference API, where many endpoints genuinely rejected `tools`. The router replaced that and the list did not keep up — its own comment cites DialoGPT-medium as the default, while `getFallbackModelName()` returns Llama-3.1-8B-Instruct.

Removing it restores the base behaviour: `modelSupports("functionCalling", ...)`, which defaults unknown models to supported. That optimistic default is the right posture for a host serving thousands of models, and it is safe here because a model that cannot use tools simply does not emit `tool_calls` — which the loop already handles. The allowlist was not protecting against anything observable.

## Verification

Deterministic before/after on the built providers, no vendor call involved:

```
                                 release   fix
Qwen/Qwen2.5-72B-Instruct        false  →  true
zai-org/GLM-5.3-Flash            false  →  true
meta-llama/Llama-3.1-8B-Instruct true   →  true
```

End to end through the SDK on the fix build, a registered tool is now actually invoked:

```
toolsUsed=["lookup_population"]
```

A live regression test is added to `continuous-test-suite-providers`: a registered tool must be invoked on a served HuggingFace model. It skips — rather than fails — when the key is absent or the vendor refuses for account reasons, so only a real tool-wiring break fails it.

## How it was found

By measuring a provider-by-feature matrix rather than reading the descriptor: huggingface discovered all 12 built-in tools and invoked none of them. That combination only makes sense if the wire never carried them.

## Summary by CodeRabbit

- **New Features**
  - Hugging Face integration now uses the unified router with broader model and tool-calling support.
  - Updated default, fast, balanced, and quality model selections to current Qwen, Llama, and DeepSeek models.
  - Added support for vision capabilities with OpenAI’s o1 model.
  - SDK error classes are now available directly from the package root.

- **Bug Fixes**
  - Improved provider-specific error reporting, provider status timeouts, benchmark resilience, and Vertex Claude structured-output handling.

- **Documentation**
  - Updated setup guidance, environment variables, examples, model lists, and provider capability references for router-based access.


src/lib/providers/huggingFace/client.ts
```diff
@@ -82,32 +82,31 @@ export class HuggingFaceProvider extends OpenAIChatCompletionsProvider {
...
+   *   served models the allowlist BLOCKS          142
+   *
+   * So it suppressed tool calling for 99% of the catalogue, including
+   * Qwen/Qwen2.5-72B-Instruct, which the router demonstrably tool-calls. A
+   * spread of the blocked models — GLM-5.3-Flash, gemma-4-31B-it,
+   * DeepSeek-V4-Flash — all accept `tools` with HTTP 200 and return
+   * `tool_calls`; none rejected the field. The stale comment also cited
+   * DialoGPT-medium as the default, but `getFallbackModelName()` returns
+   * Llama-3.1-8B-Instruct.
+   *
```
Tara-ag: **MAJOR — the default model (`microsoft/DialoGPT-medium`) is still the *default*, and this change now sends `tools` to it.**

This comment conflates the **default** model with the **fallback** model. They are different code paths:

- `getDefaultHuggingFaceModel()` → `microsoft/DialoGPT-medium` — the provider's **default** (`HuggingFaceProvider.getDefaultModel()`, client.ts:76-78, used whenever the caller supplies no model and `HUGGINGFACE_MODEL` is unset).
- `getFallbackModelName()` → `meta-llama/Llama-3.1-8B-Instruct` — only used when the default is empty *and* auto-discovery finds nothing.

After removing the allowlist, `supportsTools()` defers to `modelSupports("functionCalling", "huggingface", this.modelName)` (`baseProvider.ts:297-299` → `modelRegistry.ts` `modelSupports`), which returns `true` for **any** unknown model — *including DialoGPT-medium*, which has no registry row. So `executeStream`/the tool loop (`shouldUseTools = !options.disableTools && this.supportsTools()`, openaiChatCompletionsBase) will now put `tools` on the wire for the no-model-specified path.

The empirical evidence covers the **143 served models** — but the allowlist admitted 1 of 143 (Llama), so DialoGPT-medium is demonstrably **not** among the measured set. The claim "the optimistic default costs nothing" is therefore unverified precisely for the default model the removed comment warned "rejects the OpenAI `tools` field". If it 400s on `tools`, callers who previously relied on the tool-free default path break.

Suggested fix — pick one:
1. Update the provider default so the no-model path uses a tool-capable model (e.g. make `getDefaultHuggingFaceModel()` return Llama-3.1-8B-Instruct, matching the fallback), **or**
2. Keep a tool-suppression safe-guard for the specific default id (e.g. gate `supportsTools()` on `this.modelName !== "microsoft/DialoGPT-medium"`), since DialoGPT cannot tool-call regardless.

Either way the inaccurate fallback-vs-default framing in this comment should be corrected.

test/continuous-test-suite-providers.ts
```diff
@@ -813,6 +813,66 @@ async function testGemini3TokenCounting(
...
+// --- HuggingFace tool calling ---
+// A 13-entry model-name allowlist in the HuggingFace client used to gate
+// supportsTools(), suppressing tool calling for 142 of the 143 models the
+// router serves. This pins the behaviour the allowlist blocked: a served,
+// tool-capable model must actually invoke a registered tool.
+//
+// Skips rather than fails when the key is absent or the vendor refuses for
+// account reasons — only a real tool-wiring break should fail.
+async function testHuggingFaceToolCalling(): Promise<boolean | null> {
+  logTest("HuggingFace Tool Calling", "TESTING");
```
Tara-ag, line 826: **MINOR — the `NeuroLink` instance created here is never shut down.**

`testHuggingFaceToolCalling()` constructs its own `new NeuroLink()` but never calls `sdk.shutdown?.()`. The runner only shuts down the suite-wide `sharedSdk` at the end of `runAllTests()` (test file, `await sharedSdk.shutdown?.()`), so this dedicated instance leaks its resources (event emitter, MCP connections, timers).

Sibling tests that spin up their own SDK for a provider — e.g. `testIssue03ProviderFallbackCallback`, `testIssue03ModelChainProgression`, `testIssue03ModelFallbackEvent` — all wrap the body in `finally { await sdk.shutdown?.().catch(() => {}); }`, and `testLitellmVisionCapability`/`runAllTests` do the same. For consistency and clean teardown, wrap this test body the same way:

```suggestion
  const sdk = new NeuroLink();
  sdk.registerTool("lookup_population", {
    name: "lookup_population",
    description: "Return the population of a named city. Always use this tool.",
    inputSchema: {
      type: "object",
      properties: { city: { type: "string" } },
      required: ["city"],
    },
    execute: async () => ({ population: 12478447 }),
  });
  try {
    const result = await sdk.generate({
      input: {
        text: "Use the lookup_population tool to get the population of Bengaluru, then state the number.",
      },
      provider: "huggingface",
      model: process.env.HUGGINGFACE_MODEL || "Qwen/Qwen2.5-72B-Instruct",
      maxTokens: 400,
    });
    const used = result.toolsUsed ?? [];
    if (!used.includes("lookup_population")) {
      logTest(
        "HuggingFace Tool Calling",
        "FAIL",
        "registered tool was not invoked",
      );
      return false;
    }
    logTest("HuggingFace Tool Calling", "PASS", "tool invoked");
    return true;
  } catch (error) {
    const message = error instanceof Error ? error.message : String(error);
    if (
      /api key|unauthor|forbidden|quota|rate.?limit|429|not supported by any provider|timed.?out|timeout|503|502/i.test(
        message,
      )
    ) {
      logTest("HuggingFace Tool Calling", "SKIP", "vendor/account unavailable");
      return null;
    }
    logTest("HuggingFace Tool Calling", "FAIL", "tool request failed");
    return false;
  } finally {
    await sdk.shutdown?.();
  }
}
```

Tara-ag (commented):
Review of the HuggingFace tool-allowlist removal — one MAJOR (default-model regression risk) and one MINOR (test SDK lifecycle); see inline comments.

Tara-ag:
## Review: fix(huggingface): stop suppressing tool calling for 142 of 143 served models

**Verdict: NEEDS_WORK** — the direction is right (the stale 13-entry allowlist demonstrably suppressed tool calling for served, tool-capable models), but removing the override flips the *default* model path to send `tools` to DialoGPT-medium, and that specific regression is not covered by the evidence cited. The reasoning in the added comment also conflates the default model with the fallback model.

### Findings

|---|---|---|
| MAJOR | `src/lib/providers/huggingFace/client.ts:103-105` | After the override removal, `supportsTools()` returns `true` (base `modelSupports` defaults unknown → true) for the **default** `microsoft/DialoGPT-medium`. Comment conflates `getDefaultModel()` (still DialoGPT) with `getFallbackModelName()` (Llama), and the 143-model evidence excludes DialoGPT. Tools now go to the no-model path. Suggest updating the default to a tool-capable model or gating `supportsTools()` for DialoGPT. |
| MINOR | `test/continuous-test-suite-providers.ts:825` | `new NeuroLink()` is never `shutdown()`. The runner only shuts down `sharedSdk`; sibling tests that create their own SDK wrap in `finally { sdk.shutdown?.() }`. |

- **Base tool machinery verified**: `BaseProvider.supportsTools()` → `modelSupports("functionCalling", provider, model)` (`modelRegistry.ts`), which returns `true` for unknown models — so the removal's core claim (unknown → supported) is accurate for the 143 served models.
- **Tool-loop safety claim** holds for tool-capable models: a model that can't tool-call simply returns no `tool_calls`, which the loop already handles (verified in `openaiChatCompletionsBase.ts` executeStream/runStreamLoop).
- **No test breakage**: no existing unit/e2e test asserts the removed HuggingFace `supportsTools() === false` behavior (checked `callers_of supportsTools` across src/ and test/).
- **Conventions**: single commit, correct scope (`fix(huggingface):`), test is e2e-vendor style per Rule 15 with skip-not-fail on missing key — all good.
- **No security/secret concerns**; `formatProviderError` unchanged.

**Blast radius**: 2 changed files; provider change is confined to the HuggingFace provider's tool gate, itself driven by the shared base. The risk concentrates on the default-model path (covered above).

src/lib/providers/huggingFace/client.ts
```diff
@@ -82,32 +82,31 @@ export class HuggingFaceProvider extends OpenAIChatCompletionsProvider {
...
+   *   allowlist entries matching ANY served model   1 of 13
+   *   served models the allowlist ADMITS            1 of 143
+   *   served models the allowlist BLOCKS          142
+   *
+   * So it suppressed tool calling for 99% of the catalogue, including
+   * Qwen/Qwen2.5-72B-Instruct, which the router demonstrably tool-calls. A
+   * spread of the blocked models — GLM-5.3-Flash, gemma-4-31B-it,
+   * DeepSeek-V4-Flash — all accept `tools` with HTTP 200 and return
+   * `tool_calls`; none rejected the field. The stale comment also cited
+   * DialoGPT-medium as the default, but `getFallbackModelName()` returns
```
Tara-ag: **MAJOR — the default-model path still hits DialoGPT-medium with tools on the wire.**

The justification here relies on `getFallbackModelName()` returning `Llama-3.1-8B-Instruct`, but that is the *fallback* name, used when a requested model is missing. The model actually selected when `generate()` omits `model:` comes from `getDefaultModel()`, which in this file returns `getProviderModel("HUGGINGFACE_MODEL", "microsoft/DialoGPT-medium")` (see `getDefaultHuggingFaceModel()` above). In the `BaseProvider` constructor: `this.modelName = modelName || this.getDefaultModel()`.

So an omitted-model request against `provider: "huggingface"` still resolves to **DialoGPT-medium** (unless `HUGGINGFACE_MODEL` is set). With the override removed, `supportsTools()` now falls through to `modelSupports("functionCalling", ...)`, which **defaults unknown models to `true`** — and DialoGPT-medium is unknown to the registry. The result is tools being merged onto the wire for the default model, which is exactly the case the removed allowlist existed to suppress.

The change is right for the 142 servable models, but the default model should not be one that never emitted `tool_calls`. Suggested fixes (either):

1. Make the default tool-capable — change `getDefaultHuggingFaceModel()` to default to `"meta-llama/Llama-3.1-8B-Instruct"` (or a router-served tool-capable model), or
2. Keep `supportsTools()` authoritative but add an explicit capability entry for DialoGPT-medium in the registry marking `functionCalling: false`.

Please update the comment to reference `getDefaultModel()` (the method actually used) rather than `getFallbackModelName()`.

test/continuous-test-suite-providers.ts
```diff
@@ -813,6 +813,66 @@ async function testGemini3TokenCounting(
...
+  if (!process.env.HUGGINGFACE_API_KEY && !process.env.HUGGING_FACE_API_KEY) {
+    logTest("HuggingFace Tool Calling", "SKIP", "no HuggingFace key");
+    return null;
+  }
+  const sdk = new NeuroLink();
```
Tara-ag, line 831: **MINOR — the locally-created `NeuroLink` is never `shutdown()`.**

Every sibling test in this file that constructs its own SDK wraps it in `try/finally` and releases it — see `testIssue03ProviderFallbackCallback`, `testIssue03ModelChainProgression`, and `testIssue03ModelFallbackEvent` (all do `await sdk.shutdown?.().catch(() => {});` in a `finally`). This new test creates `const sdk = new NeuroLink()` (line 830) and never closes it, leaking the instance (span processors, tool managers, any open handles) for the rest of the process — and unlike the `sharedSdk`, this one is never released in `runAllTests`.

The test body does not use the shared `sharedSdk` because it registers a tool on a private instance; that is fine, but the instance still needs closing. Wrap the body so it releases the SDK regardless of the pass/fail/skip path:

```suggestion
  const sdk = new NeuroLink();
  try {
    sdk.registerTool("lookup_population", {
```

Prefer adding a `finally { await sdk.shutdown?.().catch(() => {}); }` at the end of the function, mirroring the sibling Issue#3 tests.

Tara-ag (changes requested):
NEEDS_WORK — the tool-allowlist removal is correct and well-evidenced for the 142 servable models, but the default-model path (DialoGPT-medium, via `getDefaultModel()` — not the fallback cited in the comment) now sends `tools` on the wire. See inline comment for the fix.

Tara-ag:
**NEEDS_WORK** — removing the hardcoded 13-entry HuggingFace tool allowlist is the right call and is well-supported by the PR's before/after evidence, but the default-model path now sends `tools` to a model that cannot use them. One MAJOR and one MINOR fix, both inline.

|---|---|---|
| MAJOR | `src/lib/providers/huggingFace/client.ts:103` | Comment/justification cites `getFallbackModelName()` (Llama-3.1-8B-Instruct), but the model used when `model:` is omitted is `getDefaultModel()` → `microsoft/DialoGPT-medium`. With the override gone, unknown models default to `supportsTools() === true`, so tools now go on the wire for the default DialoGPT medium — exactly what the allowlist existed to suppress. Change the default to a tool-capable model or gate DialoGPT. |
| MINOR | `test/continuous-test-suite-providers.ts:830` | New `new NeuroLink()` is never `shutdown()`. Sibling tests that create their own SDK wrap it in `try/finally` + `sdk.shutdown?.().catch(() => {})`. |

**Checked clean:**
- Base tool machinery unchanged: `BaseProvider.supportsTools()` → `modelSupports("functionCalling", ...)` defaults unknown models to `true` (per the diff comment, expected).
- OpenAI-compat tool loop (`shouldUseTools = !disableTools && supportsTools()`); a model that lacks function calling simply never emits `tool_calls`, so the loop is safe for non-tool models.
- The regression test pins real tool invocation on a servable model (`Qwen/Qwen2.5-72B-Instruct`), skips gracefully without a key, and is registered in `runAllTests`.
- Removal restores `BaseProvider.supportsTools()` behavior — no signature/return-shape change, no external callers of the removed override found via the code graph.
- Conventions: no CI-owned issues re-reported; no rule 1/3/4/6/15 violations; no security concerns (no secrets, no logging of credentials).

src/lib/providers/huggingFace/client.ts
murdore: Fixed in `f34e0bc0a`. Took your option 1 — make the default tool-capable — rather than re-adding a narrower allowlist.

**The comment was wrong and you named the right function.** It now cites `getDefaultModel()`, not `getFallbackModelName()`.

**The default is now a served, tool-capable model.** `getDefaultHuggingFaceModel()` returns `HuggingFaceModels.QWEN_2_5_72B_INSTRUCT` — deliberately the same shared constant that `providerRegistry.ts:259` already passes as the registration default, so the directly-constructed and factory-created paths cannot drift apart again.

**The underlying defect was bigger than tools.** `microsoft/DialoGPT-medium` is not a model that rejects `tools`; it is a model the router does not serve at all. It answers `400 "not supported by any provider you have enabled"` with or without the field, so the omitted-model path was already broken independently of this PR — the allowlist was just hiding it behind a different failure. The same is true of `microsoft/DialoGPT-large` and `meta-llama/Llama-2-7b-chat-hf`.

Every dead id is replaced with one confirmed present in the live catalogue, across the provider default, the registry default, `MODEL_NAMES.HUGGINGFACE`, the capability manifest, the `config` schema default, the `setup-huggingface` wizard choices, the setup messages, and the docs that advertised them:

| was | now |
| --- | --- |
| `microsoft/DialoGPT-medium` | `Qwen/Qwen2.5-72B-Instruct` (default) |
| `microsoft/DialoGPT-large` | `deepseek-ai/DeepSeek-V3` (quality) |
| `meta-llama/Llama-2-7b-chat-hf` | `meta-llama/Llama-3.1-8B-Instruct` (fast) |

**Regression test.** `HuggingFace Default Model` drives the built CLI through `runCLI` with `{ env: { HUGGINGFACE_MODEL: "" } }` and no `--model` flag, so it exercises exactly the omitted-model path you identified, end to end. It is deliberately outside the skip set — the old default's 400 does not match its skip regex, so it records a FAIL rather than a green skip. It passes on the current tree, alongside `HuggingFace Tool Calling`.

Separately, I corrected the counts in that comment: 142/143 → **141 of 142**, now date-stamped and hedged, because the router catalogue drifts (it served 141 a few hours after I measured 142).

test/continuous-test-suite-providers.ts
murdore: Fixed in `f34e0bc0a` — `testHuggingFaceToolCalling` now wraps its body in `try/finally` with `await sdk.shutdown?.().catch(() => {});`, matching the sibling pattern in `testIssue03ProviderFallbackCallback` / `testIssue03ModelChainProgression` / `testIssue03ModelFallbackEvent`.

The new `HuggingFace Default Model` test added in the same push sidesteps the issue entirely: it drives the built CLI through `runCLI` with `{ env: { HUGGINGFACE_MODEL: "" } }` rather than constructing an SDK instance, so there is nothing to shut down and no `process.env` mutation to restore in the parent process.

test/continuous-test-suite-providers.ts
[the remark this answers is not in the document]
murdore: Fixed in `f34e0bc0a` — `testHuggingFaceToolCalling` now wraps its body in `try/finally` with `await sdk.shutdown?.().catch(() => {});`, matching the sibling pattern in `testIssue03ProviderFallbackCallback` / `testIssue03ModelChainProgression` / `testIssue03ModelFallbackEvent`.

The new `HuggingFace Default Model` test added in the same push sidesteps the issue entirely: it drives the built CLI through `runCLI` with `{ env: { HUGGINGFACE_MODEL: "" } }` rather than constructing an SDK instance, so there is nothing to shut down and no `process.env` mutation to restore in the parent process.

Tara-ag:
## Review: fix(huggingface): stop suppressing tools for 141 of 142 served models

**Verdict: APPROVE** — the allowlist removal is correct and well-verified, and both findings from the earlier revision (f34e0bc0a, not commit 3861e) are resolved in the current tree.

### Prior findings — resolved
| Finding | Status | Evidence in current code |
|---|---|---|
| MAJOR — default-model regression (DialoGPT-medium got `tools`) | ✅ Resolved | `getDefaultHuggingFaceModel()` now returns `HuggingFaceModels.QWEN_2_5_72B_INSTRUCT` (a served, tool-capable model); the comment now cites `getDefaultModel()` correctly, and `testHuggingFaceDefaultModel` drives the omitted-model path end-to-end via the built CLI. That test's 400-skip logic correctly treats the old default as a FAIL. |
| MINOR — test SDK never shut down | ✅ Resolved | `testHuggingFaceToolCalling` now wraps its body in `try/finally` with `await sdk.shutdown?.().catch(() => {});`. |

- **Core mechanism verified.** Removing the override restores `BaseProvider.supportsTools()` → `modelSupports("functionCalling", …)` (`modelRegistry.ts`), which returns `true` for unknown models. Because the HuggingFace manifest has no priced named models, **all** HF models are unknown in `MODEL_REGISTRY`, so tools are consistently offered — matching the PR's before/after claim. The manifest `functionCalling: false → true` change is confined to the `_default` entry (excluded from `MODEL_REGISTRY`), so no behavioral blast radius; it only aligns the manifest with the wire.
- **`testHuggingFaceDefaultModel` correctness.** `runCLI` merges `{ env: { HUGGINGFACE_MODEL: "" } }` onto `process.env` (harness `env: { ...process.env, ...options.env }`), so `HUGGINGFACE_API_KEY` is preserved and only the model env is cleared; `getProviderModel`/`providerRegistry` both fall through on empty string (`||` to `QWEN_2_5_72B_INSTRUCT`). Test sound.
- **Blast radius.** `modelSupports` has a single caller (`BaseProvider.supportsTools`); no callers of the removed override remain. `getDefaultHuggingFaceModel` is referenced only from `HuggingFaceProvider.getDefaultModel`. Default-model id swaps across the registry, `MODEL_NAMES.HUGGINGFACE`, config schema, setup wizard, setup messages and docs are internally consistent, and `HUGGINGFACE_BASE_URL` matches what the (pre-existing) constructor reads.
- **Conventions.** Single commit, `fix(huggingface)` scope; tests are e2e-vendor style with skip-not-fail (Rule 15); no security/secret concerns (no credentials logged, not in diff paths).

### Non-blocking note (not an inline finding)
- `testHuggingFaceToolCalling` uses `maxTokens: 400`, which is on the small side for a tool-calling round-trip; if the model ever chooses a shorter answer path it could report a false FAIL rather than a skip. Worth bumping to ~800–1000 if it ever flakes, but not a blocker.
- The PR title/body still say "142 of 143"; the final code/commit say "141 of 142" because the router catalogue drifted. Cosmetic only, not worth a code change.

Tara-ag (approved):
**APPROVE** — the allowlist removal is correct, and both findings raised in the earlier round (MAJOR default-model regression, MINOR test SDK lifecycle) were resolved and verified in `f34e0bc0a`:

- The default is now the served, tool-capable `Qwen/Qwen2.5-72B-Instruct` (`getDefaultHuggingFaceModel()` shares the same constant as the registry default, so the two paths can't drift), and the stale fallback-vs-default framing in the comment was corrected.
- `testHuggingFaceToolCalling` wraps its body in `try/finally` + `await sdk.shutdown?.().catch(() => {})`, matching sibling tests.

No remaining blockers.

Tara-ag:
## Review verdict: APPROVE

**APPROVE** — the tool-allowlist removal is correct and well-evidenced, and both findings from the prior round were fully resolved and verified in `f34e0bc0a`. No remaining blockers.

### Findings

| Severity | File:line | Status | Finding |
|---|---|---|---|
| MAJOR | `src/lib/providers/huggingFace/client.ts:103-105` | ✅ Resolved | Default-model regression — `getDefaultHuggingFaceModel()` now returns the served, tool-capable `Qwen/Qwen2.5-72B-Instruct` (shares the registry default constant), and the fallback-vs-default framing in the comment is corrected. |
| MINOR | `test/continuous-test-suite-providers.ts:826` | ✅ Resolved | `testHuggingFaceToolCalling` now wraps its body in `try/finally` + `await sdk.shutdown?.().catch(() => {})`, matching sibling Issue#3 tests. |

- **Base tool machinery**: `supportsTools()` → `modelSupports("functionCalling", ...)` defaults unknown → `true`; no signature/return-shape change, no external callers of the removed override.
- **Tool-loop safety**: a model that can't tool-call simply emits no `tool_calls`, which the loop already handles.
- **Regression coverage**: `HuggingFace Default Model` drives the omitted-model path end-to-end via the built CLI; `HuggingFace Tool Calling` pins real tool invocation and skips-not-fails when the key is absent.
- **Conventions**: single commit, `fix(huggingface):` scope, e2e vendor-style test per Rule 15; no rule 1/3/4/6 violations; no secrets or credential logging (`formatProviderError` unchanged).

*Note: two earlier `<!-- yama:summary -->` comments on this PR carry a stale NEEDS_WORK verdict from the pre-fix revision `3861ed4`; this comment supersedes them as the authoritative summary.*

src/lib/providers/huggingFace/client.ts
Tara-ag: Confirmed fixed and resolved. Verified in `f34e0bc0a`: `getDefaultHuggingFaceModel()` returns `HuggingFaceModels.QWEN_2_5_72B_INSTRUCT` — deliberately the same constant the registry passes, so the directly-constructed and factory paths can't drift — and the corrected comment now cites `getDefaultModel()`. The `HuggingFace Default Model` regression test exercises the omitted-model path end-to-end via the built CLI. Resolved.

test/continuous-test-suite-providers.ts
Tara-ag: Confirmed fixed and resolved. `testHuggingFaceToolCalling` now wraps its body in `try/finally` with `await sdk.shutdown?.().catch(() => {})`, matching the sibling Issue#3 tests. Resolved.

Tara-ag:
## Review (recurring round): fix(huggingface): stop suppressing tools for served models

**Verdict: APPROVE** — re-review of the same single commit (`084a85ab`) confirms the 13-entry tool-allowlist removal is correct and fully verified. Both findings from the earlier rounds remain resolved in the current tree. No new blockers found on this pass.

### Prior findings — status on the current tree

| Severity | File | Status |
|---|---|---|
| MAJOR — default-model regression (DialoGPT got `tools`) | `src/lib/providers/huggingFace/client.ts` | ✅ Resolved (`getDefaultHuggingFaceModel()` returns the shared `HuggingFaceModels.QWEN_2_5_72B_INSTRUCT` constant the registry passes; comment cites `getDefaultModel()`; `testHuggingFaceDefaultModel` drives the omitted-model path end-to-end) |
| MINOR — test SDK never shut down | `test/continuous-test-suite-providers.ts` | ✅ Resolved (`testHuggingFaceToolCalling` wraps its body in `try/finally` + `await sdk.shutdown?.().catch(() => {})`) |

### Re-checked and still clean
- **Core mechanism.** Removing the override restores `BaseProvider.supportsTools()` → `modelSupports("functionCalling", …)`, which defaults unknown models to `true`. Confirmed via the code graph: `HuggingFaceProvider.supportsTools` no longer exists on the current graph (removal applied), `getDefaultHuggingFaceModel` has a single caller (`HuggingFaceProvider.getDefaultModel`), and no external caller of the removed override remains.
- **Tool-loop safety.** A model that cannot tool-call simply emits no `tool_calls`, which the existing OpenAI-compat loop handles.
- **Manifest change is metadata-only.** `functionCalling: false → true` in `manifests/huggingface.ts` is confined to the `_default` entry (excluded from `MODEL_REGISTRY`), so no behavioral blast radius — it only aligns the manifest with the wire.
- **Default-model consistency.** `getDefaultHuggingFaceModel()`, `MODEL_NAMES.HUGGINGFACE`, the `config` schema default, the setup wizard, `providerSetupMessages.ts` and the touched docs all agree on `Qwen/Qwen2.5-72B-Instruct` / `router.huggingface.co/v1`.
- **New tests.** Both e2e-vendor style per Rule 15 (skip-not-fail when the key is absent); `testHuggingFaceDefaultModel` correctly treats the old default's 400 as a FAIL (deliberately absent from the skip set) and runs via the built CLI.
- **Conventions / security.** Single commit, `fix(huggingface):` scope; `formatProviderError` unchanged; no secrets or credential logging. No CI-owned issues re-reported.

### Non-blocking — out-of-diff consistency follow-ups (unchanged, cannot anchor inline)
These three surfaces still advertise now-dead model ids / the retired endpoint and were NOT touched by this PR (all outside the changed-file set). They remain valid follow-ups, not this diff's scope:

| Surface | Current state on `fix/huggingface-tool-allowlist` |
|---|---|
| `.env.example` (`HUGGINGFACE_MODEL`) | Still `microsoft/DialoGPT-large` — a config template users copy verbatim; a user who sets it gets a hard 400 (router does not serve the id) on every HF call. Highest-impact of the three. |
| `neurolink-demo/server.ts` (`DEFAULT_MODELS.huggingface`) | Still `microsoft/DialoGPT-medium` — functional demo runtime code; `getModelForProvider()` falls back to it when `HUGGINGFACE_MODEL` is unset → 400. |
| `docs/getting-started/providers/huggingface.md` | Still built around the retired per-model Inference API: `HUGGINGFACE_BASE_URL=https://api-inference.huggingface.co` and a "100,000+ models / free ~1,000 req/day" narrative, contradicting the router migration this PR documents elsewhere. |

These can't be anchored on diff lines, so they're tracked here for a follow-up PR. None affects the SDK's runtime defaults, which are correct and internally consistent.

(Updates the prior summaries — the earlier NEEDS_WORK summaries from before the `f34e0bc0a` fix are superseded by this one.)

Tara-ag (approved):
✅ **APPROVE** — re-affirming approval on the current head `084a85ab`.

The tool-allowlist removal is correct and fully verified, and both prior-round findings are resolved in the current tree:

- **MAJOR · default-model regression (resolved)** — `getDefaultHuggingFaceModel()` now returns the router-served, tool-capable `Qwen/Qwen2.5-72B-Instruct` (same shared constant the registry passes, so the direct-construction and factory paths can't drift); the comment now cites `getDefaultModel()`, and `testHuggingFaceDefaultModel` drives the previously-broken omitted-model path end-to-end via the built CLI.
- **MINOR · test SDK lifecycle (resolved)** — `testHuggingFaceToolCalling` wraps its body in `try/finally` + `await sdk.shutdown?.().catch(() => {})`, matching sibling tests.

No remaining blockers on the diff. The PR carries a single commit (`fix(huggingface):`), tests are e2e-vendor style per Rule 15, and no security or backward-compat surfaces are touched.

test/continuous-test-suite-vertex-claude-characterization.ts
```diff
@@ -711,6 +711,164 @@ await test("the generate path declares and executes a caller's tools", async ()
...
+      ? toolTurn("lookup", { q: "x" })
+      : toolTurn("final_result", { answer: "42" }, "toolu_final"),
+  );
+  const restore = withVertexEnv();
+  const counter = { calls: 0 };
+  let aggregated = "";
+  let streamFailure = "";
+  try {
+    const nl = new NeuroLink();
+    const result = await nl.stream({
```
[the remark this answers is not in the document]
murdore: Valid, and it was at ten sites rather than the two anchored here — the same `const nl = new NeuroLink()` inside `try` shape at lines 386, 439, 480, 526, 569, 626, 679, 739, 814 and 995 (the last one with `tools: { discovery: true }`, where it matters most).

All ten now hoist the instance and `await nl?.shutdown()` as the first statement of `finally`, ahead of `restore()` and the stand-in server's close — matching what the other eleven suites in `test/` already do.

Suite re-run after the change: **10 passed, 0 failed**.

PR #1697, opened 2026-09-13
fix(bedrock): report the tools the turn actually ran

murdore:
> **Stacked on #1678** — based on `fix/bedrock-input-text`, so this targets that branch and the diff will collapse to three files once #1678 merges. Happy to rebase onto `release` instead if preferred.

## The defect

Bedrock ran tools correctly and reported none.

`conversationLoop` returned only `text`, `usage` and finish reason, dropping the agentic loop's `toolCalls` / `toolExecutions`. `baseProvider` then fell back to `result.toolsUsed || []`, so **every generate turn reported zero tools**.

The streamed path had the same hole from the other side: it handed `createAnalytics` a usage-only object, and `toolCallCount` derives from `toolExecutions`/`toolsUsed` — so **streamed turns reported no tool calls either**.

Nothing observable in the answer reveals this. The tools genuinely work and their results genuinely reach the model; only analytics, cost attribution, audit, and any caller branching on `toolsUsed` see the lie. That is why it survived.

Bedrock was the **only** `runAgenticLoop` consumer forwarding none of it — `anthropic`, `googleVertex` and `googleAiStudio` all do. Both paths now forward what the loop already computed.

## Evidence

Proven with **no AWS account**, against the local Bedrock endpoint from #1678 extended to script a tool call — real AWS SDK, real SigV4 signing, real h2c, real event-stream framing. Only the model's decision is scripted; the loop, the tool dispatch and the second round trip are real.

| surface | assertion | before | after |
| --- | --- | --- | --- |
| `generate` | `toolsUsed` omits a tool the provider executed | ✗ | ✓ |
| `stream` | analytics reports no tool calls for a turn that used one | ✗ | ✓ |

Each assertion was confirmed to fail with **its own half** of the fix removed, so neither is vacuous:

```
generate half removed -> toolsUsed omits a tool the provider executed on this turn   exit 1
stream   half removed -> stream analytics reports no tool calls for a turn that used one   exit 1
both fixes present    -> Provider Wiring 20/20                                      exit 0
```

The test first proves the tool *really ran* — `execute()` fired, and the tool result appears on the wire in the follow-up request — so an empty `toolsUsed` would otherwise be honest reporting.

## Note on the streamed assertion

It reads the tool result off the wire rather than counting `execute()` calls. The tool middleware caches identical calls, so a second turn with the same name and arguments legitimately replays the first result without re-executing. Counting executions would have made the test fail for a reason unrelated to the defect — it did, before I corrected it.

## Checks

`build` ✓ · `check` 0 errors ✓ · `check:tools-tests` ✓ · `lint` 0 errors ✓ · Provider Wiring 20/20 ✓

## Summary by CodeRabbit

- **Bug Fixes**
  - Amazon Bedrock responses now accurately report tools used during non-streaming requests.
  - Tool execution details are now included in generated results, including tool calls and execution outcomes.
  - Streaming analytics now records Bedrock tool calls accurately instead of reporting zero tool usage.

- **Tests**
  - Added coverage verifying tool execution reporting, returned tool results, and streaming tool-call analytics.


murdore:
## Confirmed against real AWS Bedrock

The PR body's evidence is from the local endpoint (deterministic, no account needed). AWS credentials have since been refreshed, so here is the same defect and fix measured against the **live** service — same prompt, same model, same account, only the build differs:

| | before (release-based build) | after (this fix) |
| --- | --- | --- |
| `execute()` ran | 1 | 1 |
| tool's secret nonce echoed in the answer | true | true |
| `toolsUsed` | `[]` | `["get_vault_code"]` |
| `toolExecutions` | 0 | 1 |

The tool ran, and the model's answer quoted a random nonce that **only** the tool could supply — while the turn reported zero tools. That is the defect in one line: the capability is real, the reporting lies about it.

Reproduced on two different model families, so it is not model-specific:

```
amazon.nova-2-lite-v1:0                      toolsUsed=["get_vault_code"] toolExecutions=1
us.anthropic.claude-haiku-4-5-20251001-v1:0  toolsUsed=["get_vault_code"] toolExecutions=1
```

Run in a clean room (`env -i`, credentials exported from SSO, no ambient proxy variables), because the ambient shell exports a base-URL override that has silently produced false passes in this campaign before.

murdore:
### Why `🔒 Single Commit Policy Validation` is red

Not a problem with this commit — the branch holds exactly one. The gate counts commits against **`release`**, not against the PR's base branch, so on a stacked PR it also counts #1678's commit:

```
❌ Branch 'fix/bedrock-tool-telemetry' contains 2 commits
   a45b293b fix(bedrock): report the tools the turn actually ran      <- this PR
   c4d30fa7 fix(bedrock): honour input.text in the native generate path  <- #1678
```

So a stacked PR can't go green here under any arrangement of its own commits, and squashing as the message suggests would swallow #1678's commit into mine — the wrong fix.

**Resolution: merge #1678 first.** This branch then has a single commit against `release`, the base retargets, and the check passes with no change to the diff. I have deliberately not rebased onto `release` early, because this PR extends `test/helpers/bedrockLocalEndpoint.ts`, which #1678 introduces — carrying that file here would duplicate it and conflict on merge.

Every other check is green. Happy to rebase and retarget the moment #1678 lands.

Tara-ag (commented):
Review of the Bedrock tool-telemetry fix.

src/lib/providers/amazonBedrock/client.ts
```diff
@@ -368,13 +372,29 @@ export class AmazonBedrockProvider extends BaseProvider {
...
+    // reported zero tools while running them correctly — leaving analytics,
+    // cost attribution and audit blind on this path.
     return {
       content: text, // CLI expects 'content' not 'text'
       usage,
       model: this.modelName || this.getDefaultModel(),
       provider: this.getProviderName(),
       ...(finishReason !== undefined && { finishReason }),
       ...(rawFinishReason !== undefined && { rawFinishReason }),
+      toolsUsed: toolExecutions.map((execution) => execution.name),
```
Tara-ag, line 387: **MINOR — `toolsUsed`/`toolCalls`/`enhancedWithTools` count tool calls that never executed.**

`toolExecutions` here comes from `runAgenticLoop`'s accumulated `allToolExecutions`. In `loopEngine.ts` `dispatchStepTools`, that array is also pushed for calls that never invoked `execute()` — the `TOOL_NOT_FOUND` and `TOOL_PERMANENTLY_FAILED` branches (`executions.push({ id, name, input, output, error })`) — and for any execution that threw (`error` set). So on a turn where the model named a tool that doesn't resolve, this block returns `toolsUsed: ["missing_tool"]`, `toolCalls` with that name, `toolExecutions.length > 0`, and `enhancedWithTools: true` — i.e. it reports "a tool was used" when none ran. Since the PR's headline is "report the tools the turn actually ran", a consumer branching on `enhancedWithTools`/`toolsUsed` (the exact lie the PR fixes) would still be misled for these cases.

This matches how anthropic's `buildToolResultMessages` pushes every `toolResult.name` into `toolsUsed`, so it's a consistent cross-provider choice rather than a regression — flagging it so the intended semantics are explicit and one place to fix all providers.

If the intent is "actually executed", consider filtering on records without an error, and applying the same filter to the streaming `toolCallCount` at line ~1346:
```suggestion
toolsUsed: toolExecutions
  .filter((execution) => execution.error === undefined)
  .map((execution) => execution.name),
```
(and gate `enhancedWithTools` on the filtered length). If counting attempted calls is intended, a one-line comment stating that would prevent a future "fix" from changing it.

Tara-ag (commented):
Review complete — one MINOR finding on tool accounting for never-executed calls. See inline comments.

Tara-ag:
*Superseded — see the canonical summary at the comment immediately below (author `Tara-ag`, `<!-- yama:summary -->`). This is a duplicate posted in error; the finding it listed (MINOR on ready tool-telemetry accounting, at line 387) was re-anchored to its correct position at `client.ts:389` in the canonical review. No separate finding here.*

**Verdict: APPROVE** — the Bedrock tool-telemetry fix correctly forwards `toolExecutions` from the agentic loop on both the generate and streaming paths, matches the sibling providers, and is well-tested.

Tara-ag:
## APPROVE

The fix closes a real defect — Bedrock's `generate()` and streaming paths now forward `toolExecutions` from the agentic loop, so `toolsUsed`/`toolCalls`/`toolCallCount` finally report the tools that actually ran. Logic is sound, mirrors the sibling `runAgenticLoop` consumers (anthropic, vertex, google-ai), and is backed by a solid end-to-end test with wire-level evidence.

| Severity | Location | What |
| --- | --- | --- |
| MINOR | `src/lib/providers/amazonBedrock/client.ts:389` | `toolsUsed`/`toolCalls`/`enhancedWithTools` count dispatched tool calls even when they never executed (TOOL_NOT_FOUND / PERMANENTLY_FAILED / a throw before the model answers). `toolExecutions` already carries the per-step status, so pre-filtering to executed steps would be more truthful. The same applies to the streamed `toolCallCount`. |

**Checked and clean:**
- No secrets or credentials added — `PLACEHOLDER_AWS_ENV` is a signed-but-never-validated local-endpoint value, and the base-URL override points at the local endpoint only. No provider params are logged.
- Backward compatibility (Rule 5): the added fields are additive on `EnhancedGenerateResult`/`GenerateResult`; no existing caller's shape is narrowed. `resolveToolExecutionRecords`/`transformToolExecutions` were already public and consumed by the sibling providers, so no new dynamic-import or circular-dependency risk (Rule 1) — the imports are static within this provider file, not in the registry.
- Rule 3: no `max_tokens`/Gemini interplay here; Bedrock's own `inferenceConfig.maxTokens` path is untouched.
- Streaming path: analytics are resolved off the settled turn outcome so `toolCallCount` lands whether or not the stream is drained.
- One MINOR only, non-blocking — no unhandled rejections introduced, error path (`catch` emitting `generation:end`) unchanged.

src/lib/providers/amazonBedrock/client.ts
```diff
@@ -368,13 +372,29 @@ export class AmazonBedrockProvider extends BaseProvider {
...
+      toolCalls: toolExecutions.map((execution) => ({
+        toolCallId: execution.id,
```
Tara-ag, line 389: **MINOR** — `toolsUsed`/`toolCalls`/`enhancedWithTools` count *dispatched* tool calls, not *executed* ones, so a turn whose tool never actually ran still reports it as used.

`runAgenticLoop` records every tool the model asks for in `toolExecutions`, including calls that never completed — `TOOL_NOT_FOUND` (the model invented the name), `PERMANENTLY_FAILED`, or a tool that threw before the model was answered. Mapping that list verbatim means `toolsUsed`, `toolCalls` and `enhancedWithTools` all claim the tool ran when it did not, feeding an inflated count into analytics/cost attribution/audit — the very surfaces this PR is un-blinding.

Filter to steps that actually executed (status `SUCCESS`) before reporting. E.g.:

````suggestion
      toolsUsed: toolExecutions
        .filter((execution) => execution.status === "SUCCESS")
        .map((execution) => execution.name),
      toolCalls: toolExecutions
        .filter((execution) => execution.status === "SUCCESS")
        .map((execution) => ({
          toolCallId: execution.id,
          toolName: execution.name,
          args: execution.input,
        })),
      toolExecutions: resolveToolExecutionRecords(
        options,
        transformToolExecutions(toolExecutions),
      ),
      enhancedWithTools: toolExecutions.some(
        (execution) => execution.status === "SUCCESS",
      ),
````

Note the same even applies to the streamed path — `createAnalytics` derives `toolCallCount` from this `toolExecutions` list, so unfiltered calls inflate the streamed count too. If the per-step `status` field is not available on `AgenticLoopExecution`, a fallback filter on `execution.error === undefined` achieves the same goal.

**Review by Tara-ag — APPROVED** [review]
Approving: the fix makes Bedrock report the tools it actually runs (native generate + streaming), mirrors the sibling `runAgenticLoop` consumers, preserves backward compatibility, and ships with a strong end-to-end test. One MINOR (non-blocking) note on `toolsUsed`/`toolCallCount` counting dispatched rather than executed calls — see the inline comment.

File: src/lib/providers/amazonBedrock/client.ts (hunk 2 above)
Comment (murdore) [reply src/lib/providers/amazonBedrock/client.ts:389]: Correct, and thank you — this is a real defect in the fix, not a nit. Mapping every dispatch meant an invented tool name was reported as *used*, which re-inflates exactly the counts this PR existed to make honest.

Fixed in **https://github.com/juspay/neurolink/pull/1718**, with two deliberate deviations from the suggested patch:

**1. No `status` field exists.** `AgenticLoopResult["toolExecutions"]` carries `{id, name, input, output, error?}` — `status === "SUCCESS"` would not compile. Using the `error === undefined` fallback you named, which is exactly equivalent: `core/loopEngine.ts` pushes `error` for `TOOL_NOT_FOUND`, `TOOL_PERMANENTLY_FAILED` and a throw, and pushes **without** it only after `executeToolCall` returns a value.

**2. `toolCalls` stays complete.** Your snippet filtered it too. Every other provider keeps it whole — `toolCallsFromSummaries` maps all summaries — and it is the record of what the turn *attempted*, which is what a caller debugging a failed turn needs. `toolExecutions` stays whole for the same reason; its own type documents it as *"every tool dispatch the loop performed, in order, including the ones that failed"*. So the filter lands on `toolsUsed` and `enhancedWithTools` only, which is the same line `nativeGenerateLoop` draws (`toolsUsed.push(name)` sits inside the try, after the guarded call returns).

**On the streamed path:** `createAnalytics` derives `toolCallCount` from `toolExecutions.length`, which stays unfiltered — a dispatch that failed *was* a call, so that count is right as it is. The streamed `toolsUsed` list is filtered anyway, so the two stop disagreeing.

**Proof, no API.** New test in `continuous-test-suite-provider-wiring.ts` drives `nl.generate()` from `dist` against the local h2c Bedrock endpoint, scripted to call a tool that was never registered:

```
with the fix:                          22 passed,  0 failed
with only the filter reverted, rebuilt: 21 passed,  1 failed
  → toolsUsed reports a tool the turn never executed
```

The pre-existing "Bedrock reports the tools it actually ran" passes in both runs, so this is additive.

**One thing I did not change, and am flagging instead:** `NeuroLink.generate()` overwrites `enhancedWithTools` with `Boolean(result.toolExecutions?.length)`, so on the SDK surface that field means "dispatched" while through a provider handle it means "ran". Those definitions have disagreed across every provider for as long as both existed — redefining a public field did not belong inside a Bedrock fix.

PR #1699, opened 2026-09-13
test(json): cover the response_format branch the JSON suite cannot reach

murdore:
## APPROVE

A tightly-scoped, test-only PR (237 additions / 1 deletion, 0 production files) that closes the "silently-ignored schema" blind spot: it pins the `response_format`/tools contract at the wire against a local OpenAI-compatible endpoint. The prior review's three MINOR findings are all **resolved** in `e0410cef7` and verified against the code.

### Findings — all resolved

|---|---|---|---|
| MINOR | `test/continuous-test-suite-provider-wiring.ts:50` | Suite wasn't `offline: true`, so a local-endpoint hang would be SKIP (vacuous pass) | **Fixed** — `defineSuite("Provider Wiring", { offline: true })`; `harness.ts` maps `offline` → `timeoutIsFailure`, so a hang is now a hard FAIL |
| MINOR | `test/continuous-test-suite-provider-wiring.ts:675-676` | `nl` (cases 1–2) never shut down, asymmetric with `flagged` | **Fixed** — both instances declared before `try`, both shut down in `finally`, each caught independently |
| MINOR | `test/helpers/openaiCompatibleLocalEndpoint.ts:64` | Endpoint 200s on any method/path; stray entries could shift positional assertions | **Fixed** — answers only `POST /chat/completions` (404 otherwise), matching the existing `startFakeModelsServer` guard |

### Claims re-verified against source

- `suppressResponseFormatWithTools()` defaults `true` (`openaiChatCompletionsBase.ts`), overridden only by openai/azure — so deepseek is on the suppressing path. ✓
- DeepSeek downgrades `json_schema` → `json_object` via `adjustResponseFormat` (`deepseek.ts`), so case 2's `format.type === "string"` assertion is lenient and correct. ✓
- `DEEPSEEK_BASE_URL` / `DEEPSEEK_API_KEY` are deepseek's real env keys (`providerDescriptors.ts`, `deepseek.ts`). ✓
- Env restore in `finally` correctly removes keys the test added (incl. `NEUROLINK_DISABLE_BUILTIN_TOOLS`, `DEEPSEEK_*`) and re-instates `savedEnv` values. ✓

### Checked clean

- **Rule 15** — no production code touched; the new test drives `dist/` + the new helper + `zod` (single dist-origin module graph, consistent with this suite's ALL-DIST contract). No `src/` runtime import in the helper.
- **Rule 4 / 6 / 1 / 3** — untouched by the diff.
- **Secrets** — `sk-local-endpoint-not-real` is an explicit placeholder, not a leaked credential.
- **Backward compatibility (rule 5)** — no public API change.
- Confirmed non-vacuous per the PR: dropping `disableTools` from case 2 fails with exit 1 (a `✗`, not a skip).

Silence on anything else = the remainder is clean.


test/continuous-test-suite-provider-wiring.ts
```diff
@@ -579,4 +579,105 @@ await test("catalog-provider enum surfaces are byte-identical to the pre-JSON-mi
   }
 });
 
+await test("a schema reaches the wire as response_format only when no tools are attached", async () => {
```
Tara-ag: **MINOR — the local-only test can still silently skip via the shared suite's timeout→SKIP classification.**

`defineSuite("Provider Wiring")` is called without `offline: true` (`test/continuous-test-suite-provider-wiring.ts:48`), so the harness wraps every `test()` in a 240s bound and classifies an exceeded bound as **SKIP** (`harness.ts` — `the upstream likely hung; aborting test`). This test only ever talks to a local socket, so a genuine hang here is a defect in the code under test — but it would be reported as `⊘ skipped`, the suite prints `RESULT: PASS`, and exits 0.

That undercuts the PR's own bar ("a `✗`, not a skip"). The verified non-vacuousness covers the *assertion-failure* branch only, not the *hang* branch. The realistic vector is the `npx @modelcontextprotocol/server-filesystem .` spawn backing cases 1/3 — if that child never serves tools in some CI environment, `getAllTools()` can sit on the hydration window and the request can stall past the bound.

The whole suite is effectively network-free (fake local servers, `127.0.0.1:1`, setup-wizard paths), so declaring `defineSuite("Provider Wiring", { offline: true })` is defensible and makes any hang a hard FAIL. If that's too broad a change for this PR, at minimum fail fast and loudly if the MCP filesystem toolset isn't on the wire instead of letting the timeout silently green-light the run.

test/continuous-test-suite-provider-wiring.ts
```diff
@@ -579,4 +579,105 @@ await test("catalog-provider enum surfaces are byte-identical to the pre-JSON-mi
...
+  const ask = (extra: Record<string, unknown>) => ({
+    input: { text: "Give the capital of France and its population." },
+    provider: "deepseek" as const,
+    schema,
+    maxTokens: 200,
+    ...extra,
+  });
+
+  try {
+    const nl = new NeuroLink({ conversationMemory: { enabled: false } });
```
Tara-ag: **MINOR — the first `NeuroLink` instance (`nl`, cases 1–2) is never shut down; only the case-3 `flagged` instance gets `shutdown()`.**

`nl` holds the auto-registered filesystem MCP server's child process and event emitter. Because this is the last test before `runSuite()` calls `process.exit()`, the leak's blast radius is bounded — but the asymmetric cleanup is easy to trip over later if this test is ever moved or the suite becomes multi-run. Move the instance creation inside the try and call `await nl.shutdown?.().catch(() => {})` in the `finally` alongside the `flagged` shutdown.

```suggestion
    const nl = new NeuroLink({ conversationMemory: { enabled: false } });
    const instances = [nl];
```
(as above, plus a `for (const inst of instances) await inst.shutdown?.().catch(() => {});` in the `finally`.)

test/helpers/openaiCompatibleLocalEndpoint.ts
```diff
@@ -0,0 +1,102 @@
...
+  } catch {
+    return null;
+  }
+}
+
+export async function startLocalOpenAICompatible(
+  reply = '{"capital":"Paris","population":2103000}',
+): Promise<LocalOpenAICompatible> {
+  const requests: CapturedCompletionRequest[] = [];
+  const server: Server = createServer((req, res) => {
```
Tara-ag, line 64: **MINOR — the endpoint answers *every* `req` with a chat-completion-shaped 200, regardless of method or path.**

The handler records and answers any incoming request (a `GET /models`, a health probe, anything) as a successful completion carrying `content`, and pushes a `{body: null}` entry when the body doesn't parse. Today the test never triggers model discovery — the deepseek default model resolves without a `/models` call, and the anchor-by-`local.requests.length` pattern tolerates stray entries — so this is defensive, not a live bug.

Still, a narrow guard makes the helper self-documenting and robust to accidental probing. Restrict the handler to `POST` and the chat-completions path, returning 404 otherwise (mirroring the existing `startFakeModelsServer` helper in `continuous-test-suite-provider-wiring.ts`):

```suggestion
  const server: Server = createServer((req, res) => {
    if (req.method !== "POST" || !req.url?.endsWith("/chat/completions")) {
      res.writeHead(404);
      res.end();
      return;
    }
    const chunks: Buffer[] = [];
```

Tara-ag (approved):
Test-only change; verified. See summary comment for the full review (3 MINOR findings).

Tara-ag:
## APPROVE

A tightly-scoped, well-researched test PR. Every factual claim it rests on checks out against `release` source, and the added test genuinely pins the `response_format`/tools branch the live JSON suite cannot reach. Three MINOR, non-blocking nits.

### Findings

|---|---|---|
| MINOR | `test/continuous-test-suite-provider-wiring.ts:582` | Local-only test inherits the suite's SKIP-on-timeout classification, so a hang (e.g. MCP filesystem server) would silently green-light rather than fail |
| MINOR | `test/continuous-test-suite-provider-wiring.ts:624` | `nl` (cases 1–2) is never `shutdown()`, only the case-3 `flagged` instance is |
| MINOR | `test/helpers/openaiCompatibleLocalEndpoint.ts:64` | Endpoint answers every request with a completion-shaped 200; narrow to `POST /chat/completions` for robustness |

### Claims verified against source

- **`NEUROLINK_DISABLE_BUILTIN_TOOLS` drops `directAgentTools` only.** `shouldDisableBuiltinTools()` gates `registerDirectTools()` (`src/lib/mcp/toolRegistry.ts`) and `directTools` (`src/lib/core/baseProvider.ts`) — not MCP tools. ✓
- **MCP tools ride `getAllTools()` onto the wire independently of the flag.** `ToolsManager.getAllTools()` returns direct tools "ALWAYS available", MCP "added when available"; `.mcp-config.json` sets `autoDiscovery.autoRegister: true` for the filesystem server. ✓
- **`disableTools: true` ⇒ zero tools.** `shouldUseTools = !options.disableTools && supportsTools()` → `toolsRecord = {}` in `openaiChatCompletionsBase.ts`. ✓
- **`suppressResponseFormatWithTools()` defaults `true`** (`openaiChatCompletionsBase.ts`), overridden to `false` only by OpenAI/Azure. ✓
- **Test logic consistent with the deepseek provider**: deepseek uses `DEEPSEEK_BASE_URL`/`DEEPSEEK_API_KEY`, doesn't override suppression, and `adjustResponseFormat` downgrades `json_schema`→`json_object`, so case 2's `format.type` assertion is lenient and correct. ✓

### Checked clean

- Rule 15 / ALL-DIST module graph: only `../dist/*` + `zod` + a test helper are runtime-imported; the existing `src/lib/types` import remains type-only. ✓
- Rule 3 structured-output gating is unchanged (no production code touched — 0 of 220 additions touch `src/`). ✓
- No secrets: `sk-local-endpoint-not-real` is an explicit placeholder. ✓

Silence on otherwise = the remainder is clean.

Tara-ag (commented):
Test-only change; verified against release source. Full review in the summary comment (3 MINOR findings).

test/continuous-test-suite-provider-wiring.ts
```diff
@@ -579,4 +579,105 @@ await test("catalog-provider enum surfaces are byte-identical to the pre-JSON-mi
   }
 });
```
Tara-ag, line 688: **MINOR** — a hang here is silently skipped, not failed.

This suite is registered as `defineSuite("Provider Wiring")` **without `offline: true`** (base-file line 33). In that mode the harness (`test/helpers/harness.ts`) downgrades any block that exceeds `perTestTimeoutMs` to `SKIP` ("upstream likely hung; aborting test") rather than to a failure. This local-only test spins up a 127.0.0.1 endpoint with a fake key and asserts on wire bytes purely to make the `response_format` path *non-vacuous* — but if that local endpoint ever hangs, the whole block is skipped and CI still greens, i.e. exactly the vacuous-passing class of failure this PR exists to eliminate.

Since every test in this suite is local-only (the existing `startFakeModelsServer` helpers point at `127.0.0.1` too), the simplest correct fix is to register the suite as offline so a hang becomes a hard failure:

```suggestion
const { test, runSuite } = defineSuite("Provider Wiring", { offline: true });
```

If you prefer to leave the suite-level option alone, this test is the odd one out and could instead move to an `offline: true` suite with the same local-endpoint helper.

test/continuous-test-suite-provider-wiring.ts
```diff
@@ -579,4 +579,105 @@ await test("catalog-provider enum surfaces are byte-identical to the pre-JSON-mi
...
+  const schema = z.object({ capital: z.string(), population: z.number() });
+  const ask = (extra: Record<string, unknown>) => ({
+    input: { text: "Give the capital of France and its population." },
+    provider: "deepseek" as const,
+    schema,
+    maxTokens: 200,
+    ...extra,
+  });
+
+  try {
```
Tara-ag: **MINOR** — the first `NeuroLink` instance is never shut down.

Cases 1 and 2 run through `nl`, but only case 3's `flagged` instance is released via `await flagged.shutdown?.().catch(() => {})` (new-file line 669). `nl` (here, and again behind the two `nl.generate` calls in cases 1–2) is left running, so any background timers/sockets it started outlive the test instead of being torn down like `flagged` is. It does not fail the test today, but it makes the cleanup asymmetric and can hold the process open longer than intended.

```suggestion
  try {
    const nl = new NeuroLink({ conversationMemory: { enabled: false } });
    const shutdown = () => nl.shutdown?.().catch(() => {});
```
…and call `await shutdown();` before the `finally` block, matching the `flagged.shutdown` teardown in case 3.

test/helpers/openaiCompatibleLocalEndpoint.ts
Tara-ag, line 64: **MINOR** — the endpoint accepts any method/path, so a wrong route still passes.

The handler responds 200 to *any* request: it never checks `req.method` or `req.url`. The test records `path` but never asserts on it (`responseFormatOnWire`/`toolNamesOnWire` only read the body). So if the deepseek client ever POSTed to a different path (e.g. `/v1/chat/completions` instead of `/chat/completions`), this stub would still return 200 and the "schema reaches the wire" assertion could pass against the wrong route — weakening exactly the determinism the helper is meant to provide.

```suggestion
const server: Server = createServer((req, res) => {
  if (req.method !== "POST" || !/\/chat\/completions$/.test(req.url ?? "")) {
    res.writeHead(404);
    res.end();
    return;
  }
```
…and consider asserting on `sent.path` in the test so a route change fails loudly.

Tara-ag (approved):
Test-only change; substantively verified against release source. Approving — 3 MINOR inline findings noted; none are blockers.

Tara-ag:
## Verdict: APPROVE

A test-only change (220 additions, 0 production files) that is well-substantiated and correct. The PR fixes a genuine gap: the JSON suite's `NEUROLINK_DISABLE_BUILTIN_TOOLS=true` never produced the tool-free run it assumed, so those cells never exercised the `response_format` branch — and adds a deterministic wire-level test to cover it. Noted 3 MINOR findings; none block merge.

| Sev | File:line | Finding |
|-----|-----------|---------|
| MINOR | `test/continuous-test-suite-provider-wiring.ts:581` | Suite isn't `offline`, so a local-endpoint hang is skipped (vacuous pass), not failed |
| MINOR | `test/continuous-test-suite-provider-wiring.ts:623` | First `NeuroLink` instance is never shut down (asymmetric with case 3) |
| MINOR | `test/helpers/openaiCompatibleLocalEndpoint.ts:64` | Endpoint 200s on any method/path; `path` recorded but never asserted |

**Checked and found clean:**
- **No production code touched** — only `test/` tree; no SDK surface, no provider, no registry change.
- **Backward compatibility (rule 5)** — no public API change; `getAvailableProviders`, `NeuroLink` ctor, `generate`, `shutdown` unchanged.
- **Secrets** — `"sk-local-endpoint-not-real"` is a deliberate placeholder, not a leaked credential; `run_secret_scanning` shows nothing leaked.
- **Rule 1 / rule 3 / rule 6** — untouched by the diff (no registry, no structured-output-policy, no error-formatting change).
- **Module graph (rule 15)** — the new test imports `dist/index.js` and the helper, consistent with this suite's ALL-DIST contract and the ESLint allow-list for the new file.
- **Env hygiene** — `savedEnv` snapshot/restore in `finally` correctly restores `DEEPSEEK_BASE_URL` / `DEEPSEEK_API_KEY`.
- **Helper contracts** — `toolNamesOnWire`/`responseFormatOnWire` handle `null`/`undefined` bodies safely.

Separately, the suite-level choice to not set `offline: true` is pre-existing and applies to the whole `provider-wiring` file, not something this PR introduced — flagged only because the new test's purpose is precisely to be non-vacuous.

test/continuous-test-suite-provider-wiring.ts
murdore: Fixed in `e0410cef7`: `defineSuite("Provider Wiring", { offline: true })`.

Your read of the mechanism is right, and it is worth restating why it applies here rather than being a blanket setting. `harness.ts:552` sets `const timeoutIsFailure = defs.offline === true;`, so without the flag an exceeded 240s bound is classified SKIP and the suite still prints `RESULT: PASS` and exits 0. Nothing in this suite makes a provider call that could legitimately hang — every case talks to a loopback server or a local child process — so there is no case where "the upstream likely hung" is the correct explanation for a timeout here. A hang is a defect, and it now reports as one.

You are also right that the non-vacuousness check I ran covered only the assertion-failure branch. The whole suite re-ran green after the change (19 passed / 19 total, 0 skipped), which is the precondition for that claim meaning anything — a suite that skipped the case would also have said `PASS`.

test/continuous-test-suite-provider-wiring.ts
murdore: Fixed in `e0410cef7`.

Both references are declared before the `try`, and both are shut down in the `finally`:

```ts
let nl: NeuroLink | undefined;
let flagged: NeuroLink | undefined;
try {
  nl = new NeuroLink({ conversationMemory: { enabled: false } });
  ...
} finally {
  await nl?.shutdown?.().catch(() => {});
  await flagged?.shutdown?.().catch(() => {});
  await local.close();
```

The in-block `await flagged.shutdown?.()` is gone, so `flagged` is no longer shut down only on the success path. Each shutdown catches independently — `NeuroLink.shutdown()` rethrows from its outer handler, so a single rejection would otherwise skip the cleanup after it, including `local.close()` and the environment restore.

I kept two named bindings rather than the `instances` array from the suggestion, only because the two are created at different points and `let`/`undefined` already expresses "may not exist yet"; the cleanup semantics are the same.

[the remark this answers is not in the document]
murdore: Fixed in `e0410cef7` — this is the same defect as the thread above, and both points in it were correct.

Both references are declared before the `try` and shut down in the `finally`, and each shutdown is caught independently:

```ts
} finally {
  await nl?.shutdown?.().catch(() => {});
  await flagged?.shutdown?.().catch(() => {});
  await local.close();
```

The independent catches matter for exactly the reason you gave: `NeuroLink.shutdown()` rethrows the error its outer handler caught, so one rejected shutdown would otherwise skip the other instance, `local.close()`, and the environment restore. The in-block `await flagged.shutdown?.()` was removed, so `flagged` is no longer cleaned up only when the assertions pass.

test/helpers/openaiCompatibleLocalEndpoint.ts
murdore: Fixed in `e0410cef7` — the helper now answers only the route it claims to serve:

```ts
if (req.method !== "POST" || !req.url?.endsWith("/chat/completions")) {
  res.writeHead(404, { "content-type": "application/json" });
  res.end(JSON.stringify({ error: { message: "not found" } }));
  return;
}
```

Your framing is the reason this was worth changing even though it is not a live bug today: the assertions read `local.requests` positionally, so a stray entry that was never a completion request silently shifts what they are asserting on, and a caller that dialled the wrong URL would still get a plausible 200 instead of failing.

The guard is load-bearing rather than decorative — the suite re-ran green afterwards (19/19, 0 skipped), including `a schema reaches the wire as response_format only when no tools are attached`, which only passes if the request really does hit `/chat/completions`. A mis-aimed guard would have 404'd it.

Tara-ag:
## APPROVE

A tightly-scoped, **test-only** change (237 additions / 1 deletion, 0 production files) that closes the "silently-ignored schema" blind spot in the JSON e2e suite. It pins the `response_format`/tools contract at the wire against a local OpenAI-compatible endpoint, and all three MINOR findings from the prior review round are **resolved** in current HEAD (`e0410cef7`) and verified against the diff.

### Findings — all resolved (not re-raised)

| Severity | Location | Finding | Status (verified in diff) |
|---|---|---|---|
| MINOR | `test/continuous-test-suite-provider-wiring.ts:50` | Suite wasn't `offline: true`, so a local-endpoint hang would classify as SKIP (vacuous pass) | **Fixed** — `defineSuite("Provider Wiring", { offline: true })` (harness maps `offline` → `timeoutIsFailure`) |
| MINOR | `test/continuous-test-suite-provider-wiring.ts:625-626` | `nl` (cases 1–2) never shut down, asymmetric with `flagged` | **Fixed** — both declared before `try`, both `shutdown()`'d in `finally`, each caught independently |
| MINOR | `test/helpers/openaiCompatibleLocalEndpoint.ts:70` | Endpoint 200ed any method/path; stray entries could shift positional assertions | **Fixed** — answers only `POST …/chat/completions`, 404 otherwise |

### Production claims re-verified against the graph

- **`suppressResponseFormatWithTools()` is base-defined and overridden by exactly OpenAI + Azure** — it exists in `openaiChatCompletionsBase.ts:331` and is overridden only in `openAI/client.ts:128` and `azureOpenai.ts:191`. So deepseek is on the suppressing path. ✓ (case 1's `response_format === null`)
- **`disableTools: true` ⇒ zero tools on wire** — `openaiChatCompletionsBase.ts:1377-1387`: `shouldUseTools = !options.disableTools && supportsTools()`; `toolsRecord = {}`; `openAITools = undefined`. ✓ (case 2)
- **`getAllTools()` feeds the wire independently of `NEUROLINK_DISABLE_BUILTIN_TOOLS`** — line 1379 `(await this.getAllTools())` brings MCP tools in regardless of the flag; the flag only drops direct tools. ✓ (case 3 pinning the misconception)
- Env restore in `finally` removes added keys and re-instates `savedEnv`. ✓

### Checked clean
- **Rule 15** — the new helper imports only `node:http` (no `src/` runtime import); the test drives `../dist/index.js` + the helper + `zod`: a single dist-origin module graph, consistent with this suite's ALL-DIST contract.
- **Secrets** — `sk-local-endpoint-not-real` is an explicit placeholder, not a leaked credential.
- **Backward compatibility (rule 5)** — no public/SDK surface touched.
- **Blast radius** — the graph reports a large "impacted files" set, but this is an artifact of symbol-name fan-out from test files; there are 0 affected execution flows and the only changed node kinds are `Test`/test-helper functions. No out-of-diff production callers are affected.
- Non-vacuous per the author: dropping `disableTools` from case 2 fails the run (a `✗`, not a skip).

No new findings in the current revision. Silence on everything else = clean.

Tara-ag (approved):
Approving — all three MINOR findings from my earlier review are resolved in `e0410cef7` and verified in the current code:

- `defineSuite("Provider Wiring", { offline: true })` — a local-endpoint hang is now a hard FAIL, not a vacuous SKIP.
- Both `nl` and `flagged` instances are declared before `try` and shut down independently in `finally`.
- The helper answers only `POST /chat/completions` (404 otherwise).

The `response_format`/tools wire contract is correctly pinned (deepseek is on the suppression path; `disableTools:true` ⇒ zero tools; MCP tools ride `getAllTools()` independently of `NEUROLINK_DISABLE_BUILTIN_TOOLS`). Test-only change; no production surface touched (Rule 5 safe). See the PR-body summary for the full record.

test/continuous-test-suite-provider-wiring.ts
Tara-ag: Accepted — confirmed in the current file: `defineSuite("Provider Wiring", { offline: true })` at line 50, and `harness.ts` maps `offline` → `timeoutIsFailure`, so a hang now fails rather than skipping. Resolving.

test/continuous-test-suite-provider-wiring.ts
Tara-ag: Accepted — confirmed both `nl` and `flagged` are declared before `try` and shut down in `finally`, each caught independently. Resolving.

test/helpers/openaiCompatibleLocalEndpoint.ts
Tara-ag: Accepted — confirmed the helper now 404s non-`POST /chat/completions` requests. Resolving.

test/continuous-test-suite-provider-wiring.ts
Tara-ag: Duplicate of the `offline`/SKIP finding in #discussion_r3999334907 — same defect, same fix (`offline: true`), already resolved. Superseding this orphan thread.

test/continuous-test-suite-provider-wiring.ts
Tara-ag: Duplicate of the `nl`-shutdown finding in #discussion_r3999334938 — same defect, same fix (both instances shut down in `finally`), already resolved. Superseding this orphan thread.

test/helpers/openaiCompatibleLocalEndpoint.ts
Tara-ag: Duplicate of the endpoint-guard finding in #discussion_r3999334983 — same defect, same fix (404 on non-`POST /chat/completions`), already resolved. Superseding this orphan thread.

PR #1701, opened 2026-09-14
fix(proxy): add supervised rolling restart command

murdore:
A manual restart can leave a rolling proxy rejecting every new request while waiting for stalled requests to drain. There is no supported CLI command that owns readiness checks and recovery.

Add `neurolink proxy restart` and optional `--check` / `--format json`, backed by a private supervisor control socket. The current worker keeps serving until a replacement acknowledges readiness and activation. Candidate failures retain the serving worker; failed candidates that ignore termination are cleaned up. Overlapping operations and additional manual replacements while older workers drain are refused. Closing the CLI does not strand restart completion. Rolling workers now reject the legacy global drain action.

Success verifies serving identity/version, admission, socket-transfer deltas measured from the start of handoff and preservation of request-log disk/OTel state. Disk sink selection must stay unchanged in both directions: unexpected disk enablement regresses OTel-only resource use, while silent disk disablement loses requested logs. Both transitions and OTel initialization loss are tested. Control paths include the supervisor instance UUID to avoid stale PID collisions without unlinking another socket. Post-activation verification failure returns `activated_unverified` instead of claiming rollback. The CLI explicitly types and documents the unknown-outcome JSON variant without inventing supervisor fields. Control-server errors retain a persistent handler and are logged without stopping the serving listener; repeated errors and continued CLI/HTTP availability are tested. Environment, launcher and update history are not rewritten.

Validation: 22 isolated restart cases through the built CLI and real worker processes, including a complete 256,000-byte stream and continued new-request admission. Candidate timeout, crash, wrong version, ignored SIGTERM, concurrent commands, client disconnect, logging regression and unsupported supervisors are covered. The suite is wired into required CI. A deliberately incorrect assertion produced a nonzero failure rather than a skip. Existing rolling performance gates passed with zero drops, 0.292 ms added p95 normally and 5.089 ms under sustained load. On the final revision, full build/publint, strict TypeScript, test/tool type checks, focused ESLint and documentation builds passed locally. All five required CI checks pass, including all 22 restart cases. Full repository validation and pre-push suites also passed earlier in this PR.

This replaces workers; it does not replace the supervisor or apply launchd settings. Older supervisors need one separately planned activation of a release containing the control capability. Collector/backend delivery and production adoption are not claimed. No installed proxy configuration or service lifecycle was changed for development or tests.

## Summary by CodeRabbit

- **New Features**
  - Added `neurolink proxy restart` for safely restarting the proxy with replacement-worker verification.
  - Added `--check` readiness checks and text or JSON output formats.
  - Active connections can continue through a successful restart, with failed or unsafe handoffs reported clearly.
  - Concurrent, draining, unsupported, or improperly configured restart requests are refused.

- **Documentation**
  - Added guidance covering restart behavior, readiness limits, status results, errors, and configuration limitations.


src/lib/proxy/restartControl.ts
```diff
@@ -0,0 +1,319 @@
...
+        fresh.version !== version
+      ) {
+        throw new Error(
+          "The replacement's serving identity could not be verified.",
+        );
+      }
+      const beforeLogs = status.observability.requestLogs;
+      const afterLogs = fresh.observability.requestLogs;
+      if (
+        (!beforeLogs.diskEnabled && afterLogs.diskEnabled) ||
```
Tara-ag: **MAJOR — the request-log disk preservation check is inverted.** This half flags disk logging going from **off → on** (`!beforeLogs.diskEnabled && afterLogs.diskEnabled`) as a "regression", while the very next clause flags OTel going from **on → off** (`beforeLogs.otel.initialized && !afterLogs.otel.initialized`). The two halves encode opposite directions, and the disk one is backwards relative to the stated goal ("preservation of request-log disk/OTel state"):

- A replacement that *enables* disk logging (before off, after on) is false-flagged as a regression and returns `activated_unverified`, even though enabling logging is not a regression.
- A replacement that silently **disables** disk logging (before on, after off) — the actual data-loss regression this check is meant to catch — is **not** detected.

The OTel clause is correct (`before && !after`, i.e. initialized → uninitialized). Fix the disk clause to match that direction:

```suggestion
        (beforeLogs.diskEnabled && !afterLogs.diskEnabled) ||
```

The fixture in `test/continuous-test-suite-proxy-restart.ts` (the `logging-regression` case sets `diskEnabled: generation > 1 && mode === "logging-regression"`, i.e. gen1 off → gen2 on) was written to match the buggy direction, so the test locks in the inverted behaviour instead of catching it. Update the fixture to turn disk logging **off** on the replacement and assert `activated_unverified`.

src/lib/proxy/restartControl.ts
```diff
@@ -0,0 +1,319 @@
...
+ * No update history, launcher, environment file or launchd unit is rewritten.
+ */
+export async function startProxyRestartControl(
+  options: ProxyRestartControlOptions,
+): Promise<{
+  identity: ProxyRestartControlIdentity;
+  close: () => Promise<void>;
+}> {
+  const instanceId = randomUUID();
+  const socketPath = join(options.stateDir, `restart-${process.pid}.sock`);
```
Tara-ag: **MINOR — control-socket path is keyed only on pid, so a PID-reused stale socket can block supervisor startup.** The socket is `join(stateDir, \`restart-${process.pid}.sock\`)`, created with the deliberate policy "never unlink an existing path". If a previous supervisor crashes and its pid is later reused by this supervisor, the stale socket for the old instance still exists at the same path; `control.listen(socketPath, …)` then fails with `EADDRINUSE` and `startProxyRestartControl` throws, taking down the freshly-started supervisor (the new instance's restart capability is wired into supervisor bring-up, so an uncaught throw here is fatal).

The safe policy is the right one — you must not unlink a socket you don't own. The gap is that ownership is never verified before `listen`. Consider `lstat`ing the path and, when it is a socket whose `uid` equals the current user, unlinking it before `listen` (the scenario that is clearly your own stale process); keep failing hard for sockets owned by another user, as today.

This is narrow (crash + pid-reuse in the same state dir), so flagging as MINOR, but it is the one path where the new capability can turn a normal restart into a dead supervisor instead of just being unavailable.

src/lib/proxy/restartControl.ts
```diff
@@ -0,0 +1,319 @@
...
+        (beforeLogs.otel.initialized && !afterLogs.otel.initialized)
+      ) {
+        throw new Error(
+          "The replacement's logging state regressed; inspect telemetry before further changes.",
+        );
+      }
+      const latest = options.server.snapshot();
+      if (
+        latest.active?.pid !== fresh.pid ||
+        latest.rejectedSockets !== before.rejectedSockets ||
```
Tara-ag: **MINOR — post-activation verification demands socket-counters stay bit-identical during the handoff, so any churn makes a healthy restart report `activated_unverified`.** After `replace(version)` succeeds, the check requires `latest.rejectedSockets === before.rejectedSockets && latest.failedTransfers === before.failedTransfers`. `before` is captured at preflight, before the handoff. Any client that disconnects mid-handoff (a rejected/leftover socket) or any transfer that fails during the swap increments those counters and the verification fails — even though the replacement is fully healthy. The result is `activated_unverified` with exit code 1 for a restart that actually succeeded, which the docs tell operators to investigate.

This is conservative-on-purpose, but it couples a *steady-state* signal to a *transitional* event. If the goal is to confirm the handoff lost nothing, count the sockets owned at the boundary (e.g. diff via `rejectedSockets`/`failedTransfers` deltas captured immediately before `replace`) rather than comparing against the preflight snapshot taken several awaits earlier. At minimum, consider only fail-closed on `failedTransfers` increasing and treat `rejectedSockets` deltas as informational, since socket rejection during a handoff is a normal client-lifecycle event, not a regression.

Tara-ag (commented):
Reviewed the supervised rolling restart — solid design and test coverage; found one MAJOR logic bug in the logging-preservation check and two MINOR robustness issues. Inline details on the exact lines.

Tara-ag:
## Verdict: NEEDS_WORK

A well-structured, well-tested rolling-restart feature with a strong security posture (nonce-guarded, chmod-0600, owner-checked control socket; no process-kill fallback; supervisor-owned completion). One MAJOR logic bug in the logging-preservation check needs fixing before it merges; the rest are MINOR robustness notes.

| MAJOR | `src/lib/proxy/restartControl.ts:160` | Inverted request-log disk preservation check — flags disk logging being *enabled* as a regression and misses it being *disabled*; test fixture locks in the buggy direction. |
| MINOR | `src/lib/proxy/restartControl.ts:57` | Control socket path is pid-keyed with a "never unlink" policy — a PID-reused stale socket can make supervisor startup throw `EADDRINUSE`. |
| MINOR | `src/lib/proxy/restartControl.ts:170` | Post-activation verification requires socket counters to be unchanged across the handoff, so normal client churn yields spurious `activated_unverified`. |

- **CLI⇄supervisor protocol**: request/response shapes, `x-neurolink-instance` nonce, socket permission/ownership checks, and pid-identity verification all match between `proxyRestart.ts` and `restartControl.ts`.
- **`rollingWorkerSupervisor.ts` SIGKILL-bounded candidate cleanup**: correct pattern (`offCandidateExit` self-removing via the same `handle.onExit` disposer API used elsewhere; unref'd timer); doesn't disturb the still-serving generation.
- **Drain-action gate (backward-compat, rule 5)**: the rolling auto-update path uses `SIGUSR2`/`activatePendingUpdate`, not the global HTTP drain; the only `markProxyDrainingForUpdate`/`{action:'drain'}` callers are inside `proxy.ts`'s own guard, which for rolling supervisors never drains a worker — so refusing drain on rolling workers breaks nothing out-of-diff. `resume` (the safe direction) is not gated.
- **Test suite**: determinism exception justified (rule 15); realistically exercises built CLI + real socket-transfer runtime, including an unresponsive-candidate SIGKILL and a full 256 KB stream.
- **Security**: no seeded/injected secrets; control socket is private to the service user; error messages never echo status payloads.

### Impact-analysis note
The code-review-graph and background-worker tooling were unavailable in this review environment (no checkout / graph not built), so blast-radius checking was done by direct source reading (callers of `markProxyDrainingForUpdate`, the `drain`/`resume` POST, and the updater's `SIGUSR2` rolling path in `src/cli/commands/proxy.ts`) rather than graph traversal. Two things I could not fully verify without the supervisor-wiring region of `proxy.ts` and would confirm before relying on the feature in production: (a) that `startProxyRestartControl(...)` persists `restartControl` into `proxy-supervisor-state.json` alongside the pid, and (b) whether a manual `proxy restart` to apply an updater-installed version interacts correctly with `pendingRestartVersion`/update-state (manual restart does not clear it). No fabricated claims are made on these.

src/lib/proxy/restartControl.ts
```diff
@@ -0,0 +1,319 @@
...
+export async function startProxyRestartControl(
+  options: ProxyRestartControlOptions,
+): Promise<{
+  identity: ProxyRestartControlIdentity;
+  close: () => Promise<void>;
+}> {
+  const instanceId = randomUUID();
+  const socketPath = join(options.stateDir, `restart-${process.pid}.sock`);
+  await mkdir(options.stateDir, { recursive: true, mode: 0o700 });
+  // Never unlink an existing path: an unexpected owner must make setup fail.
```
Tara-ag: **MINOR — PID-keyed socket path + "never unlink" fails on crash-restart**

The control socket is `restart-${process.pid}.sock` and the code deliberately never unlinks an existing path (line 59). `close()` (which removes the socket via `rm`) only runs on graceful shutdown. If the supervisor is killed (crash, OOM, `SIGKILL`, host reboot) the socket file survives; because the path is PID-keyed it is likely to collide again on the next start — whether the OS has recycled that PID or the relaunch reuses it. `server.listen()` throws `EADDRINUSE` when the path already exists, so a crash leaves the supervisor unable to bind its restart control at all.

The security intent is sound (don't trust an unexpected owner's path). Preserve it while still allowing crash recovery by unlinking only when the encoded owner pid is no longer alive:

```suggestion
  // A crashed instance can leave its socket behind. Unlink it only when the
  // encoded owner pid is no longer alive, never clobbering a live server.
  const previousPid = Number(socketPath.split("-").pop()?.replace(".sock", ""));
  if (previousPid) {
    let alive = true;
    try {
      process.kill(previousPid, 0);
    } catch {
      alive = false;
    }
    if (!alive) {
      await rm(socketPath, { force: true });
    }
  }
```

src/lib/proxy/restartControl.ts
```diff
@@ -0,0 +1,319 @@
...
+      ) {
+        throw new Error(
+          "The replacement's logging state regressed; inspect telemetry before further changes.",
+        );
+      }
+      const latest = options.server.snapshot();
+      if (
+        latest.active?.pid !== fresh.pid ||
+        latest.rejectedSockets !== before.rejectedSockets ||
+        latest.failedTransfers !== before.failedTransfers
```
Tara-ag: **MINOR — post-activation verification compares cumulative counters, so it can flap**

After `replace(version)` the check requires `latest.rejectedSockets === before.rejectedSockets && latest.failedTransfers === before.failedTransfers`, where `before` was snapshotted just before rollover. These are monotonic cumulative counters on the server. In production a connection that arrives in the handoff window and is rejected / fails to transfer legitimately increments one of them, so a *healthy* restart can be reported `activated_unverified` for reasons unrelated to the replacement. The suite only ever exercises the zero-drop fixture, so this path is never covered.

If the intent is to verify the handoff did not lose sockets, baseline the counter delta from right after rollover; otherwise scope the check to the active-pid identity:

```suggestion
      if (latest.active?.pid !== fresh.pid) {
        throw new Error(
          "Worker identity changed during verification.",
        );
      }
```

Tara-ag (changes requested):
Review complete — NEEDS_WORK. One MAJOR logic bug (inverted disk-logging preservation guard) and two MINOR robustness issues in restartControl.ts. Full details inline and in the summary comment.

Tara-ag:
## Review — NEEDS_WORK

Solid, well-tested feature (supervised rolling restart) with one MAJOR logic bug and two MINOR robustness issues, all in `src/lib/proxy/restartControl.ts`.

### Findings

|---|---|---|
| MAJOR | `restartControl.ts:160` | Request-log **disk-preservation guard is inverted**: it flags disk logging turning **on** as a regression while the OTel clause flags it turning **off** — so it both false-positives and misses the real data-loss case (disk silently disabled). The test fixture locks in the buggy direction. |
| MINOR | `restartControl.ts:57-59` | PID-keyed control socket + "never unlink" policy → a stale socket left by a crashed instance causes `EADDRINUSE` and fails supervisor startup. Unlink only when the encoded owner pid is dead. |
| MINOR | `restartControl.ts:168-171` | Post-activation verification requires cumulative `rejectedSockets`/`failedTransfers` counters to stay bit-identical across the handoff, so normal client churn yields spurious `activated_unverified`. |

### Checked and clean

- CLI⇄supervisor control protocol: nonce header, socket permission (`0o600`) and ownership checks, pid-credenced path in `proxyRestart.ts`, response schema validation, bounded timeouts, "never fall back to killing a process".
- `rollingWorkerSupervisor.ts` SIGKILL-bounded cleanup of an unresponsive candidate (`ignores-term` path) — bounded, unref'd, self-cleaning.
- Global-drain 409 gate in `proxy.ts` : refusing `drain` on rolling workers does not break the auto-update flow, which uses `SIGUSR2` (not the HTTP drain), so rule 5 (backward compatibility) holds for the existing updater path.
- New worker files (`proxyRestartWorker.mjs`) and the suite are wired to CI and carry the rule-15 determinism exception in both the file header and `eslint.config.js`.

### Caveats (could not fully verify — no local checkout / graph)

- Whether `restartControl` identity is persisted across supervisor restarts in `proxy-supervisor-state.json` vs. only session-live.
- Interaction between a manual `restart` and `pendingRestartVersion` (auto-update) — `assertIdle` guards on `isUpdatePending()`, but the reverse direction (manual restart blocking a subsequent scheduled update) was not traced.

Verdict: **NEEDS_WORK** — the inverted MAJOR guard should be corrected (and the test fixture update to match) before merge; the two MINOR items are advisory.

src/lib/proxy/restartControl.ts
murdore: Addressed in 0a5c6fd28a142cc7e4df1725388a8d989afdf65c. Disk sink selection is now preserved in both directions (`before.diskEnabled !== after.diskEnabled`). This catches the missing enabled-to-disabled logging loss. It also retains the disabled-to-enabled guard: unexpectedly turning on disk request logging violates the configured OTel-only policy and can reintroduce the disk/CPU problem this deployment is avoiding. Restart must preserve the selected sink, so simply reversing the old condition would leave that regression uncovered.

The isolated process suite now has separate cases for unexpected disk enablement, unexpected disk disablement and OTel initialization loss. Each reports `activated_unverified` with exit 1. All 20 restart cases pass.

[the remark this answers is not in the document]
murdore: Addressed in 0a5c6fd28a142cc7e4df1725388a8d989afdf65c. Disk sink selection is now preserved in both directions (`before.diskEnabled !== after.diskEnabled`). This catches the missing enabled-to-disabled logging loss. It also retains the disabled-to-enabled guard: unexpectedly turning on disk request logging violates the configured OTel-only policy and can reintroduce the disk/CPU problem this deployment is avoiding. Restart must preserve the selected sink, so simply reversing the old condition would leave that regression uncovered.

The isolated process suite now has separate cases for unexpected disk enablement, unexpected disk disablement and OTel initialization loss. Each reports `activated_unverified` with exit 1. All 20 restart cases pass.

src/lib/proxy/restartControl.ts
murdore: Addressed in 0a5c6fd28a142cc7e4df1725388a8d989afdf65c. The socket filename now includes the supervisor instance UUID as well as its PID: `restart-${pid}-${instanceId}.sock`. The CLI validates that exact path against the saved identity. This avoids stale PID collisions without deleting a path that might belong to another live instance. An isolated test starts two control instances using the same process PID and verifies that closing one does not remove the other's socket. All 20 restart cases pass.

src/lib/proxy/restartControl.ts
murdore: Addressed in 0a5c6fd28a142cc7e4df1725388a8d989afdf65c. The counter baseline is now captured immediately before `replace`, after preflight awaits. Results expose raw `rejectedSocketsDelta` and `failedTransfersDelta`; the docs explicitly describe them as observations without attributing their cause to restart.

Added cases proving that preflight counter noise is excluded and handoff counter noise produces `activated_unverified` with the observed deltas. The latter is intentional: an activated healthy worker alone does not prove lossless handoff. Removing the counter guard or taking the baseline after rollover would hide failures during the operation being verified. All 20 restart cases pass.

test/fixtures/proxyRestartWorker.mjs
```diff
@@ -0,0 +1,68 @@
...
+        pid: process.pid,
+        version,
+        health: {
+          ready: true,
+          acceptingConnections: mode !== "draining",
+          drainingForUpdate: mode === "draining",
+        },
+        observability: {
+          requestLogs: {
+            diskEnabled: generation > 1 && mode === "logging-regression",
```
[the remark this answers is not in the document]
murdore: Addressed in 0a5c6fd28a142cc7e4df1725388a8d989afdf65c. Added a separate `disk-logging-lost` fixture mode for enabled-to-disabled disk logging, alongside the original unexpected disk enablement case and a separate OTel initialization-loss case. Both disk transitions are regressions for a restart that promises to preserve configured sinks; an OTel-only deployment must not silently start writing request logs to disk. All three cases assert `activated_unverified` and a nonzero CLI exit. All 20 restart cases pass.

Tara-ag:
## Verdict: APPROVE

`neurolink proxy restart` is a well-scoped, carefully-hardened rolling-replacement command. The security model (0o600 control socket, exact-path + UUID-nonce validation, uid/mode checks in the CLI, supervisor-owned `operate()` that refuses overlap and never falls back to killing) is sound, and the test suite (`continuous-test-suite-proxy-restart.ts`, wired into CI with a declared rule‑15 determinism exemption) exercises the real failure paths. I found no new CRITICAL or MAJOR issues. Prior findings are below.

## Prior findings — all addressed in `0a5c6fd` (accepted)

The 7 earlier threads on `restartControl.ts` and the fixture were filed against an older revision and are now outdated; the current HEAD contains fixes for all of them:

| Severity | File | Finding | Resolution |
|---|---|---|---|
| MAJOR | `restartControl.ts:247` | Disk-logging regression check inverted (off→on flagged, on→off missed) | Now `beforeLogs.diskEnabled !== afterLogs.diskEnabled` — catches both transitions; OTel clause unchanged |
| MINOR | `restartControl.ts:57` | PID-keyed socket + never-unlink → `EADDRINUSE` deadlock on crash/PID-reuse | Path now `restart-${pid}-${instanceId}.sock`; distinct UUID never collides |
| MINOR | `restartControl.ts:149-171` | Post-activation verification compared cumulative counters to a too-early baseline → false `activated_unverified` | Baseline captured immediately before `replace()`; `rejectedSocketsDelta`/`failedTransfersDelta` exposed as observations |
| MINOR | `proxyRestartWorker.mjs` | Fixture locked in the inverted disk direction | Separate `logging-regression`, `disk-logging-lost`, `otel-uninitialized` modes; each asserts `activated_unverified` |
| MINOR | CLI | (CodeRabbit) same fixture correction | Covered by the above |

The author's rationale for treating disk **off→on** as a regression too (restart must preserve the configured sink; an OTel-only deployment must not silently start writing to disk) is legitimate and documented. Counter-delay semantics are now explicitly described as observations, not restart attribution.

## Freshly checked, found clean

- **Security bar** — control socket `chmod`d 0o600; CLI `lstat`s and rejects non-socket / group-other-writable / foreign-uid paths; rebuilds and validates the exact expected path from `proxy-supervisor-state.json`; instance nonce rejects stale clients; request bodies and unauthorized methods rejected. Nothing user/model-controlled flows into the socket.
- **Supervisor change** (`rollingWorkerSupervisor.ts:336-350`) — the 1s SIGKILL backstop for a term-ignoring candidate is correct: `unref` prevents a hang, the exit handler clears the timer, and the serving (draining) generation is untouched. Bounded and scope-correct.
- **Failure semantics** — failed/never-ready/wrong-version/updating candidates retain the serving worker; overlap is refused; `activated_unverified` is distinctly reported (never false rollback). Docs match behavior.
- **Out-of-diff impact** — `requestProxyRestart` is only called from `proxyRestart.ts`; `startProxyRestartControl` is wired into the supervisor bring-up and the rolling-worker path. No unmodified public-SDK call sites are affected (rule 5).

## One non-blocking note (not an inline finding)

On crash/OOM the `restart-*.sock` file is left behind (the `rm` in `close()` only runs on graceful shutdown) and, because each restart gets a fresh UUID under the deliberate never-unlink policy, it is never reaped — so stale socket files accumulate in `stateDir` over time. Harmless (paths never collide), but consider unlinking/deduplicating old `restart-*` files on startup (`stateDir` is already 0o700). This is the residual of the previously-resolved crash-recovery threads, so it is a hygiene note rather than a blocking change.

Tara-ag (approved):
## Verdict: APPROVE

All prior findings (1 MAJOR + 4 MINOR across `restartControl.ts` and the fixture) were addressed in `0a5c6fd`; I found no new CRITICAL or MAJOR issues on the current HEAD. Full detail is in the summary comment. Approving to reflect the current state of the review.

## Review-state note
This approving review supersedes the earlier CHANGES_REQUESTED review (filed against the old revision `699dd`). The disk-logging guard is no longer inverted, the socket path is no longer collision-prone, and the post-activation counter verification is no longer flapping. The author's justifications for keeping disk on/off guarded in both directions and for treating counter deltas as observations (not restart attribution) are sound and documented.

src/cli/commands/proxyRestart.ts
```diff
@@ -0,0 +1,114 @@
...
+      }
+    } catch (error) {
+      const message =
+        error instanceof Error
+          ? error.message
+          : "Restart control is unavailable.";
+      if (argv.format === "json") {
+        console.info(
+          JSON.stringify({ ok: false, phase: "unverified", message }),
+        );
```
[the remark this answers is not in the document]
murdore: Fixed in df376a68cf0d9e7ea411064ba0f69f854bf7e434. Added `CliProxyRestartError` and the `CliProxyRestartOutput` union in the canonical type file, used the explicit error type in the CLI catch path, and documented the JSON variant and nonzero exit. The supervisor response schema remains unchanged. A focused built-CLI case asserts `ok: false`, `phase: unverified`, a message, exit 1 and the absence of invented supervisor fields. All 22 restart cases pass locally.

src/lib/proxy/restartControl.ts
```diff
@@ -0,0 +1,355 @@
...
+  control.requestTimeout = 5_000;
+  control.maxHeadersCount = 10;
+  control.maxConnections = 16;
+  await new Promise<void>((resolve, reject) => {
+    control.once("error", reject);
+    control.listen(socketPath, () => {
+      control.off("error", reject);
+      resolve();
+    });
+  });
```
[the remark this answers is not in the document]
murdore: Fixed in df376a68cf0d9e7ea411064ba0f69f854bf7e434. A persistent error handler is installed when binding succeeds; it reports through the supervisor logger while retaining the serving listener. The new isolated regression case emits two errors after binding and verifies both are reported, the serving worker identity stays unchanged, HTTP still returns 200 and the built CLI check succeeds. This deterministically exercises the failure without exhausting the host's file descriptors. All 22 restart cases pass locally.

murdore:
@coderabbitai review

Please finish the review on current head `df376a68cf0d9e7ea411064ba0f69f854bf7e434`. Both additional findings are fixed and confirmed in their threads; the current summary coverage still names the previous revision. The final revision has 22 passing isolated restart cases and refreshed build, type, lint and documentation validation.

src/lib/proxy/restartControl.ts
```diff
@@ -0,0 +1,362 @@
...
+    control.close();
+    throw error;
+  }
+  return {
+    identity: { protocol: 1, socketPath, instanceId },
+    close: async () => {
+      closing = true;
+      control.closeAllConnections();
+      await new Promise<void>((resolve) => control.close(() => resolve()));
+      await rm(socketPath, { force: true });
```
Tara-ag, line 301: **MINOR — stale control sockets (and the stale `restartControl` identity in supervisor-state) accumulate across supervisor crashes.**

`close()` above is the only place the socket file is removed (`rm`), and it runs only on graceful shutdown. If the supervisor is killed (crash / OOM / SIGKILL / host reboot), `restart-<pid>-<uuid>.sock` survives in `stateDir`.

- The UUID-in-path change (from the resolved `EADDRINUSE` thread) means a fresh supervisor's UUID never collides with the stale path, so the old socket never blocks a new bind — but it is also **never reaped**, so stale socket files accumulate in `~/.neurolink/` over repeated crashes.
- The stale identity also lingers in `proxy-supervisor-state.json` (`restartControl.socketPath` points at a dead socket) until the next successful supervisor start overwrites it. A `neurolink proxy restart --check` issued while the supervisor is down against that stale file will fail the `lstat`/ownership check and report an error — acceptable, but the error text ("restart control ownership or permissions are invalid") does not mention the more likely cause (a stale record from a crashed instance).

Non-blocking (paths never collide, no security posture change), but consider reaping `restart-*.sock` files whose embedded pid is not alive on startup, with the same owner check already used by the CLI, so the state directory stays clean.

Tara-ag (commented):
Reviewed current HEAD (df376a6). Accepting the previously resolved findings; one MINOR hygiene note on stale control-socket accumulation.

Tara-ag:
## Verdict: APPROVE

`neurolink proxy restart` is a well-scoped, carefully-hardened rolling-replacement command. Current HEAD `df376a6` addresses every prior finding (7 review threads + 2 CodeRabbit threads), all adequately justified by the author; I found no new CRITICAL or MAJOR issues on the current revision.

### Prior findings — all resolved & accepted (no re-posting)

| Severity | File | Finding | Resolution (adequate) |
|---|---|---|---|
| MAJOR | `restartControl.ts:184-187` | Disk-logging regression check inverted | Now `beforeLogs.diskEnabled !== afterLogs.diskEnabled` — preserves the configured sink in **both** directions; treating disk turn-on on an OTel-only deployment as a regression is legitimate. OTel clause unchanged. |
| MINOR | `restartControl.ts:58-62` | PID-keyed socket + never-unlink → `EADDRINUSE` on crash/PID-reuse | Socket path now `restart-${pid}-${instanceId}.sock` (instance UUID); distinct UUID never collides; CLI rebuilds/validates the exact path. |
| MINOR | `restartControl.ts:192-204` | Post-activation verification compared cumulative counters to a too-early baseline → false `activated_unverified` | Baseline captured immediately before `replace()`; `rejectedSocketsDelta`/`failedTransfersDelta` exposed as observations, not restart attribution. |
| MINOR | fixture | Fixture locked in inverted disk direction | Split into `logging-regression` / `disk-logging-lost` / `otel-uninitialized` modes; each asserts `activated_unverified`. |
| MINOR | `proxyRestart.ts` | CLI catch-path JSON used undeclared `unverified` shape (CodeRabbit) | Added typed `CliProxyRestartError` + `CliProxyRestartOutput` union in `types/proxyRestart.ts`; supervisor response schema unchanged; focused built-CLI assertion. |
| MINOR | `restartControl.ts:281-285` | Control `error` listener removed after bind → possible supervisor exit (CodeRabbit) | Persistent `control.on("error", …)` logs through `options.log` after binding; regression test emits two post-bind errors and verifies serving / HTTP / CLI stay up. |

### Freshly checked against current HEAD, found clean

- **Security bar** — 0o600 control socket; CLI `lstat`s and rejects non-socket / group-other-writable / foreign-uid paths; instance nonce rejects stale clients; token-guarded `/internal/update-control`; nothing user/model-controlled flows into the socket.
- **Legacy drain gate (rule 5)** — `createProxyStartApp` returns 409 for `action:"drain"` when `isProxySocketWorkerProcess()`, but the rolling auto-update path uses `SIGUSR2`/`activatePendingUpdate` (verified in the updater loop), **not** the HTTP drain, so refusing drain on rolling workers breaks no unmodified caller. `resume` (safe direction) is not gated.
- **`rollingWorkerSupervisor.ts` SIGKILL backstop** (line 341-350) — bounded 1s `unref`'d timer, `offCandidateExit` self-removing, only fires for a candidate that has already been discarded; the serving/draining generation is untouched.
- **Blast radius (rule 5)** — `requestProxyRestart` is called only from `proxyRestart.ts`; `startProxyRestartControl` only from `runLaunchdProxySupervisor` + the isolated suite. No unmodified public-SDK call sites.
- **Test coverage** — suite wired into CI with a declared rule-15 determinism exception; realistically exercises the built CLI against the real socket-transfer runtime (256 KB stream, unresponsive-candidate SIGKILL, concurrency, client disconnect, socket-permission refusal).

### One MINOR inline (new, non-blocking)

`restartControl.ts:301` — stale control sockets (and the stale `restartControl` identity in `proxy-supervisor-state.json`) accumulate because `rm` runs only in `close()` (graceful shutdown). Paths never collide so it is purely a hygiene issue; suggested startup reap with the existing owner check.

src/lib/proxy/restartControl.ts
Tara-ag, line 301: **MINOR — stale control sockets and the stale supervisor identity accumulate; non-blocking hygiene note.**

`close()` only ever runs on a graceful shutdown, so the `rm(socketPath, …)` here never fires if the supervisor is killed (crash, OOM, `SIGKILL`, reboot). Because each restart gets a fresh `instanceId`, the `restart-${pid}-${instanceId}.sock` files left behind never collide — but they are also never reaped, so `stateDir` accumulates orphaned `restart-*.sock` files over time, and the last written `proxy-supervisor-state.json` references a dead instance identity until the next clean start.

This is exactly the residual of the (already-resolved) crash-recovery threads, so it is hygiene rather than a blocking issue. If you want to address it, consider reaping stale `restart-*.sock` files that belong to the current user at startup (stateDir is already `0o700`, so this is safe), e.g.:

```suggestion
      // Clean up any orphaned restart sockets from this user on a cold start;
      // paths are never reused across instances, so only stale files collect here.
      await rm(socketPath, { force: true });
```

No action required to merge; flagging so it is a deliberate, documented choice rather than an accident of the never-unlink policy.

Tara-ag (approved):
## Verdict: APPROVE

Reviewed current HEAD (`df376a6`). All prior findings are addressed; one non-blocking MINOR hygiene note on stale control sockets is inlined. Reflecting the APPROVE verdict on the latest revision.

src/lib/proxy/restartControl.ts
murdore: Acknowledged as a non-blocking cleanup limitation; it is explicitly recorded in the operator report. This PR deliberately removes only the current instance's own socket on graceful close. It does not add a startup sweep of other instances' filesystem entries. UUID-scoped paths preserve availability after a crash without requiring that cleanup, and the leftover entries contain no request logs or payloads.

One diagnostic clarification: an existing stale socket with the expected UID/mode passes `lstat` and proceeds to the connection attempt, which reports the connection error as the CLI's documented `unverified` outcome. A missing path reports `ENOENT`. The ownership/permissions error is only returned when those properties actually fail validation. We do not infer that a permissions error means the supervisor crashed.

Safe garbage collection can be added separately with dead-process checks and tests for PID reuse, active sibling instances and path replacement races. Keeping that sweep outside this availability fix avoids introducing deletion races into supervisor startup. No claim is made that abrupt-crash filesystem cleanup is complete.

murdore: This duplicates the non-blocking hygiene note already acknowledged and documented in https://github.com/juspay/neurolink/pull/1701#discussion_r4003167000. The deliberate choice is to remove only this instance's socket on graceful shutdown and preserve crash recovery through unique paths. Startup garbage collection needs independent ownership/liveness/race coverage; this PR does not claim it. Resolving this duplicate as the same documented tradeoff.

murdore:
@coderabbitai review

Retrying after the reported review-limit window. Current head remains `df376a68cf0d9e7ea411064ba0f69f854bf7e434`; all five required CI checks passed, all 22 restart tests passed in CI, Yama approved this revision, and all known inline findings are resolved.

PR #1702, opened 2026-09-14
fix(proxy): close OTel correlation gaps and add telemetry doctor

murdore:
## Description

Direct Codex requests could finish without trace correlation or first-output timing, failed OTLP exports retained counts without request identities, and historical debugging still required separate scripts. This change uses the existing OTel pipeline for correlated proxy evidence and adds built-in `proxy telemetry doctor` and `proxy telemetry query` commands over stored OTLP records.

The doctor checks recent signal freshness, request fields and duplicate finals, terminal/final reconciliation, capture delivery, bounded body/trace samples, and producer/collector failures. Missing, stale, partial or corrupt evidence returns a nonzero result; unavailable queries also return structured JSON.

## Changes

- Add a W3C-parented HTTP SERVER span across proxy request doors and native OTLP correlation through detached logging callbacks. Direct Codex spans retain account and effort; internal fallback spans do not add independent request/token metrics.
- Cover text, refusal, populated tool and completion-only output timing. Malformed/incomplete observations remain explicitly unavailable; relay bytes and routing behavior are preserved.
- Retain bounded failed-export record identities and publish recovery diagnostics through OTel without replaying uncertain original records. Expose capture admission limits, high-water marks, and policy exclusions.
- Discover the native collector's OpenObserve destination or use explicit environment settings. Bound backend history reads and reject incomplete query responses. OTLP handles export; OpenObserve supplies historical queries.
- Disable the optional raw Codex disk capture in OTel-only mode, update the browser OTel shim, and document the coverage matrix and delivery limits.

## Type and compatibility

Bug fix, CLI feature, documentation and deterministic test coverage. No provider/model/dependency changes. Existing file-mode compatibility and offline archive tools remain available. OTel-only operation still requires its explicit service configuration; this PR does not change installed services.

## Validation

- `test:proxy-telemetry`: 60 passing cases, including actual local OTLP ingestion, twelve output variants, W3C parents/combined headers/native correlation and retained sampling flags, failed-export recovery, malformed streams and built doctor/query CLI checks.
- `test:codex`: 69 passing cases; `test:proxy-connect-retry`: 5 passing cases.
- Full SDK/CLI/browser build and `publint`; strict project, tools/tests and CI-script TypeScript checks; 1,038 shipped declarations resolve.
- Repository build rules, formatting and lint; generated API documentation and shipped docs search index refreshed.
- All four isolated proxy performance gates pass. These are deterministic regression gates, not live capacity or universal no-loss proof.

## Limits and review notes

Finite queues, process death, backend retention and OTLP partial acceptance remain explicit delivery limits. Failed-export history is bounded; body hash and trace checks are labeled samples. The doctor cannot prove events that disappeared before observable admission. OTel-only eliminates proxy application log files, while operational state and backend database storage remain persistent.

Most changed files under `docs/api` are required generated line-reference updates. The meaningful source changes are in the proxy logger/tracer, Codex usage/routes, CLI telemetry command, and observability query helpers.

No live proxy changes or paid generation probes were used for development or validation. Merge is reserved for the maintainer; deployment and natural-traffic verification are separate steps.

## Summary by CodeRabbit

- **New Features**
  - Added `neurolink proxy telemetry doctor` to verify proxy telemetry health and delivery.
  - Added `neurolink proxy telemetry query` with time-range, record-type, format, and row-limit options.
  - Improved request tracing and correlation across proxy telemetry.
  - Added body-capture admission diagnostics, including capacity limits and rejection reasons.
  - Added clearer useful-output timing and incomplete-observation reporting for streamed responses.

- **Documentation**
  - Expanded telemetry setup, querying, troubleshooting, and verification guidance.
  - Clarified telemetry record types and filtering for accurate request history.


scripts/observability/proxy-telemetry-check.mjs
```diff
@@ -0,0 +1,522 @@
...
+          const match = /^(\w+(?:\{.*\})?)\s+(\S+)(?:\s+\d+)?$/.exec(line);
+          return [match?.[1] ?? "invalid_sample", Number(match?.[2])];
+        }),
+    );
+    add(
+      "collector_delivery",
+      ["log_records", "spans", "metric_points"].every((signal) =>
+        Object.keys(counters).some((key) =>
+          key.startsWith(`otelcol_exporter_send_failed_${signal}`),
+        ),
```
Tara-ag, line 460: **MAJOR** — the new `collector_delivery` doctor check always warns (and so makes the whole doctor exit nonzero) under any live traffic.

`counters` collects every matched line, including the `otelcol_exporter_queue_size` **gauge**, not just failure counters:
```js
.filter((line) =>
  /^otelcol_(exporter_(queue_size|send_failed|enqueue_failed)|receiver_(refused|failed))/.test(line),
)
```
Then the pass/warn decision inspects **all** counter values:
```js
? Object.values(counters).some(
    (value) => !Number.isFinite(value) || value < 0 || value > 0,
  )
  ? "warn"
  : "pass"
```
Because `otelcol_exporter_queue_size` is normally `> 0` whenever any record is queued/in transit, `value > 0` is true in every healthy, non-idle collector. The check therefore reports `collector_delivery: warn`, `checkProxyTelemetry` returns overall `status: "incomplete"`, and `runProxyTelemetryDoctor` sets `process.exitCode = 1` — so a healthy production system fails the doctor it was built for. It only ever passes when the collector is fully idle (`queue_size` = 0).

The check should classify `queue_size` as neutral and only test the actual failure counters (`send_failed`, `enqueue_failed`, `refused`, `failed`), e.g.:

```suggestion
        ? ["send_failed", "enqueue_failed", "refused", "failed"].some((cat) =>
            Object.keys(counters).some(
              (key) => key.startsWith(`otelcol_exporter_`) && key.includes(cat) && Number(counters[key]) > 0,
            ),
          )
          ? "warn"
```

(Also note the current `every(...)` *gating* clause requires every one of `log_records/spans/metric_points` to expose a `send_failed` counter before the warn/pass branch is even evaluated; a collector that omits one signal silently degrades the check to `unverified`. Consider gating on "any" rather than "every".)

src/lib/proxy/proxyTraceContext.ts
```diff
@@ -0,0 +1,53 @@
...
+      ? { traceId: record.traceId, spanId: record.spanId }
+      : typeof record.requestId === "string"
+        ? getProxyRequestTraceContext(record.requestId)
+        : undefined;
+  if (!ids) {
+    return context.active();
+  }
+  const spanContext = {
+    ...ids,
+    traceFlags: trace.getSpanContext(context.active())?.traceFlags ?? 0,
```
Tara-ag: **MINOR** — `proxyLogContext` derives `traceFlags` from the *current active* context instead of the stored span context, so the detached-callback correlation this function exists to provide can silently fail.

`registerProxyRequestTraceContext` stores only `{ traceId, spanId }` (no flags), and `proxyLogContext` reconstructs flags from `context.active()`:
```js
const spanContext = { ...ids, traceFlags: trace.getSpanContext(context.active())?.traceFlags ?? 0 };
return isSpanContextValid(spanContext) ? ... : context.active();
```
In the exact scenarios this helper targets ("even after body processing leaves async context"), `context.active()` is the empty ROOT context, so `trace.getSpanContext(context.active())` is `undefined` and `traceFlags` becomes `0`. `@opentelemetry/api`'s `isSpanContextValid` requires the SAMPLED flag to be set, so an otherwise-correct `{traceId, spanId}` pair is rejected and the function falls through to `context.active()` — dropping the correlation rather than restoring it. This makes the advertised "native OTLP correlation through detached logging callbacks" a no-op in the common self-rooted case (it only works when the current requisition coincidentally has a sampled context).

Suggest storing and propagating the sampled flag alongside traceId/spanId (e.g., a `traceFlags`/`sampled` field in the `requests` map and in `ProxyTracer.getTraceContext()`), and using it here instead of reading the active context:

```suggestion
const spanContext = {
  ...ids,
  traceFlags: ids.traceFlags ?? trace.getSpanContext(context.active())?.traceFlags ?? 1,
};
```

(If the intended behavior is "never correlate non-sampled", then gate explicitly on a carried `sampled` flag rather than the current-context read, which is what causes the silent drop.)

Tara-ag (commented):
Reviewed the proxy OTel correlation + telemetry doctor/query changes. Two inline findings; see summary for full collation.

Tara-ag:
## Review: PR #1702 — fix(proxy): close OTel correlation gaps and add telemetry doctor

**Verdict: NEEDS_WORK** — solid, well-tested feature (59 passing telemetry cases, W3C parent + native OTLP correlation, bounded failed-export retention), but the new doctor command has a logic defect that makes it fail on healthy live systems, and the detached-callback correlation relies on an active-context `traceFlags` read that silently drops the sampled flag.

### Findings

| **MAJOR** | `scripts/observability/proxy-telemetry-check.mjs:437–447` | `collector_delivery` pass/warn inspects **all** matched counters, including the `otelcol_exporter_queue_size` gauge, so a nonzero queue-size makes it `warn` → overall `incomplete` → the doctor exits nonzero on any healthy, non-idle collector. Should treat queue-size as neutral and only test the failure counters (`send_failed`/`enqueue_failed`/`refused`/`failed`). |
| **MINOR** | `src/lib/proxy/proxyTraceContext.ts:46–49` | `proxyLogContext` rebuilds `traceFlags` from `context.active()` instead of carrying the sampled flag from the stored span context. In the detached-callback scenarios this helper exists for, active is the empty ROOT context → `traceFlags: 0` → `isSpanContextValid` rejects it → the promised OTLP correlation silently drops. |

- **SQL safety in `queryProxyHistory`/`checkProxyTelemetry`** — `stream`, `organization` are validated against strict regexes and `kind` against a `KINDS` allowlist before interpolation into SQL; `captureId`/`traceId` are additionally regex-checked before being used in queries. No injection path found.
- **Credential handling** — basic-auth/authorization headers are only transmitted as fetch headers and never surface in error messages (`HTTP <status>` only); HTTPS-outside-loopback + loopback-only collector-metrics are enforced in both `validateProxyTelemetryBackend` and the query paths. No leaked secrets.
- **Span lifecycle / trace-context release** — `startProxyHttpTrace` registers and always `releaseProxyRequestTraceContext`s in both the normal `end()` path and its `try/catch` fallback; no leak detected. `otelLogSink` settled-exactly-once guard + 6s deadline prevents double-settle; failure history is bounded to 16 with `telemetry_delivery` recursion guard.
- **Bounded memory / delivery accounting in `otelLogSink`** — `publishProxyOtelBody` respects `maxPending`/`maxPendingBytes`, `export` decrements `state.outstanding` exactly once per batch, and partial/partial-window handling in `readWindow` never duplicates hits.

### Notes
- Diff too large for the inline GitHub endpoint; source files were reviewed at the PR head ref via the graph + GitHub contents. Bias was toward the 18 meaningful source files (proxy tracer/logger, codexUsage, otelLogSink, routes, CLI telemetry, observability scripts, tests); generated `docs/api/**` churn and `.yama` artifacts were skipped.
- The `collector_delivery` check also gates on `every` signal exposing a `send_failed` counter (a missing signal degrades a healthy system to `unverified` → exit 1); consider `some`/`any` gating. Flagged in the inline comment.
- No provider/model/dependency changes; existing file-mode compatibility preserved.

Tara-ag (changes requested):
Verdict: **NEEDS_WORK**. Two findings remain open (see the inline threads):

- **MAJOR** `scripts/observability/proxy-telemetry-check.mjs:437–447` — the `collector_delivery` check inspects the `otelcol_exporter_queue_size` gauge alongside failure counters, so any non-idle, healthy collector fails the doctor.
- **MINOR** `src/lib/proxy/proxyTraceContext.ts:46–49` — `proxyLogContext` reads `traceFlags` from the empty active context in detached callbacks, so the OTLP correlation silently drops.

Full collation in the summary comment below.

scripts/observability/proxy-telemetry-check.mjs
murdore: Fixed in 82fc3fe05c1c. Positive queue occupancy is retained as gauge evidence and no longer changes the delivery verdict. The doctor fixture now passes with queue_size=7 and zero failure counters; a positive failure counter still prevents a green report. Missing signal counters intentionally remain unverified because this command checks coverage for logs, traces and metrics. Telemetry suite: 60/60 passed.

src/lib/proxy/proxyTraceContext.ts
murdore: Fixed in 82fc3fe05c1c. Request contexts, route trace snapshots, finals/attempts and deferred body metadata now retain traceFlags. Restoration uses the recorded context, preserving flags even after the request registry has been released. Local OTLP receiver tests prove both flags=0 and flags=1 retain IDs and the original flags. One clarification from the installed OTel API: isSpanContextValid validates the IDs and also accepts unsampled contexts; the previous defect lost the sampling flag rather than necessarily dropping IDs. Telemetry suite: 60/60 passed.

scripts/observability/proxy-telemetry-check.mjs
```diff
@@ -0,0 +1,522 @@
...
+      )
+        ? Object.values(counters).some(
+            (value) => !Number.isFinite(value) || value < 0 || value > 0,
+          )
```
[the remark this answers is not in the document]
murdore: Fixed in 82fc3fe05c1c. Positive queue occupancy is retained as gauge evidence and no longer changes the delivery verdict. The doctor fixture now passes with queue_size=7 and zero failure counters; a positive failure counter still prevents a green report. Missing signal counters intentionally remain unverified because this command checks coverage for logs, traces and metrics. Telemetry suite: 60/60 passed.

src/lib/proxy/proxyTracer.ts
```diff
@@ -957,3 +977,67 @@ export function recordFallbackAttempt(attrs: {
...
 
 export { ProxyTracer };
+
+/** Standard SERVER span covers every proxy door, including parsing and admission failures. */
+export function startProxyHttpTrace(
+  metadata: RuntimeRequestMetadata,
+  headers: Record<string, string>,
+) {
+  try {
+    const parent = propagation.extract(ROOT_CONTEXT, headers);
```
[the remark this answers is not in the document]
murdore: Fixed in 82fc3fe05c1c. The ingress trace normalizes combined traceparent values to their first value, matching the existing OtelBridge compatibility policy. HTTP fixtures cover both comma-separated forms, with and without a space, and verify the ingress parent and nested route span stay on the original trace. Single-value propagation remains covered. Telemetry suite: 60/60 passed.

scripts/observability/proxy-telemetry-check.mjs
murdore: Follow-up in 98dfbc7cdd4c: the saved native collector exposition confirmed that never-triggered send_failed series can be absent. The check now recognizes either sent or send_failed counters for each signal, and only actual failure/refusal values affect the failure verdict. This preserves the all-signal coverage requirement without requiring an emitted zero-failure series. A realistic sparse exposition with positive sent counters and no failure counters now passes; absent diagnostic data still cannot pass. Telemetry suite 60/60, tools/tests types and changed-file lint pass.

Tara-ag:
## Review update — PR #1702 (recurring review at head `82fc3fe`)

**Verdict: APPROVE** — both previously-requested changes were verified fixed in the current head; no new blocking findings.

### Prior findings — resolved (`yama` update of the earlier summary)

| Severity | File | Status after re-check at head |
|----------|------|------------------------------|
| **MAJOR** | `scripts/observability/proxy-telemetry-check.mjs:451–459` | **Fixed.** `collector_delivery` now tests failure/refusal counters only, and the `some(...)` global check excludes `queue_size` (`!key.startsWith("otelcol_exporter_queue_size") && value > 0`). A nonzero queue gauge no longer flips the verdict to `warn` → the doctor's green path is reachable under live traffic. The author's reply and the new fixture (`queue_size=7`, zero failure counters → pass) are consistent with the code. |
| **MINOR** | `src/lib/proxy/proxyTraceContext.ts:24–59` | **Fixed.** `resolveProxyLogTraceContext` now carries `traceFlags` through `record.traceFlags`, the `requests` registry, and `getTraceContext()`, and restores correlation via `trace.setSpanContext(ROOT_CONTEXT, ids)`, preserving the sampling flag across detached callbacks. Verified both `flags=0` and `flags=1` are retained (`isSpanContextValid` accepts unsampled contexts), matching the reply. |
| MINOR (sub-note: `every`-gating on missing signal counters) | `proxy-telemetry-check.mjs:451–455` | **Accepted as designed.** The author states missing signal counters intentionally report `unverified` because this command is a coverage check across logs/traces/metrics. Consistent with the PR's stated contract ("unavailable evidence → nonzero"). Not re-flagged. |

### Fresh pass over the fix commit — no new findings
Reviewed the current sources at head (not just the diff): `otelLogSink.ts`, `codexUsage.ts`, `requestLogger.ts`, `bodyCaptureWorker.ts`, `codexProxyRoutes.ts`, `proxyTracer.ts`, `proxy-telemetry-backend.mjs`, `query-proxy-history.mjs`, `proxy-telemetry-check.mjs`.

- **OTLP sink** (`otelLogSink.ts`): exporter settles exactly once (guard + 6 s deadline); `outstanding` is decremented once per exported batch; failure history is bounded to 16 with the `telemetry_delivery` recursion guard; body publication is bounded by `maxPending=16` / `maxPendingBytes=32 MiB` and serialized via `bodyPublicationChain`. No unbounded growth or double-settle.
- **Body capture** (`bodyCaptureWorker.ts`): `MAX_PENDING=16`, `MAX_PENDING_BYTES=32 MiB`, `MAX_ENTRY_BYTES=16 MiB` with node-traversal caps; the publication `consume` lease correctly retains memory until settle. Bounded throughout.
- **Trace lifecycle** (`proxyTracer.ts` `startProxyHttpTrace`): the SERVER span is ended and `releaseProxyRequestTraceContext` runs in the `finally` of the normal `end()` path and in the catch fallback — no span/context leak. Combined `traceparent` is normalized before extraction (CodeRabbit's finding, also fixed).
- **Codex** (`codexUsage.ts`/`codexProxyRoutes.ts`): the usable-output timing statuses (`observed`/`no_useful_output`/`not_observed`) are consistent with the doctor's `firstOutputUnexplained` validation; fallback requests set `recordRequestMetrics: false` so no independent totals are emitted. Raw disk capture is disabled in OTel-only mode as described.
- **Backend/query** (`proxy-telemetry-backend.mjs`, `query-proxy-history.mjs`, `proxy-telemetry-check.mjs`): `organization`/`stream`/`kind` are allowlist/regex-validated and `captureId`/`traceId` are regex-checked before SQL interpolation; history reads are budget- and row-bounded and partial responses are rejected; credentials travel only as fetch headers and never appear in error text; HTTPS-outside-loopback is enforced.

### Checked and clean
- No hardcoded credentials or secret leaks into logs/headers (redaction of `authorization`/`x-api-key`/etc. intact).
- No provider/model/dependency changes; existing file-mode compatibility preserved.
- Docs/`docs/api` churn and `.yama` artifacts are generated and were not reviewed for logic.

One thing I could not do: **in-place edit of the earlier `yama:summary` comment** (no edit tool is exposed for issue comments here), so this post is the updated summary rather than a silent edit of the old `NEEDS_WORK` text. The prior review's inline threads are already resolved by the author.

Tara-ag:
## Review update — PR #1702 (recurring review at head `98dfbc7`)

**Verdict: APPROVE** — all previously-requested changes verified fixed at the current head, the follow-up refinement is sound and tested, and no new findings.

### Prior findings — confirmed resolved at `98dfbc7`

| Severity | File | Status after re-check |
|----------|------|----------------------|
| **MAJOR** | `scripts/observability/proxy-telemetry-check.mjs` `collector_delivery` | **Fixed.** The failure verdict now gates on `queue_size`/`sent` being neutral and only `send_failed`/`enqueue_failed`/`refused`/`failed` with `> 0` flip it to `warn`. The follow-up also recognizes **either** `sent_<signal>` **or** `send_failed_<signal>` per signal, so an absent never-triggered failure series no longer degrades a healthy system (`sent > 0` satisfies the coverage gate). Verified against the test fixture: `busy_collector` (queue_size=7, zero failures) → `pass`, `sparse_collector` (sent-only, no failure series) → `pass`, `collector_failure` (send_failed=1) → non-pass. Author's replies in thread `r4009946446`/`r4010021959` match the code. |
| **MINOR** | `src/lib/proxy/proxyTraceContext.ts` | **Fixed.** `resolveProxyLogTraceContext` now carries `traceFlags` via the record / `requests` registry (`saved`) / inferred, and `proxyLogContext` restores correlation with `trace.setSpanContext(ROOT_CONTEXT, ids)` instead of reading `context.active()` and silently dropping the sampled flag. `isSpanContextValid` accepts unsampled contexts, so both `flags=0` and `flags=1` are retained. Confirmed by the `detached OTLP logs preserve both sampled and unsampled trace contexts` test and the `traceFlags: 1` native-correlation assertion. |
| MINOR (CodeRabbit) | `proxy-telemetry-check.mjs` `queue_size` in the comparison | **Fixed** — queue-size gauge is excluded from the failure comparison; retained only as evidence. |
| MINOR (CodeRabbit) | `src/lib/proxy/proxyTracer.ts` `startProxyHttpTrace` | **Fixed** — combined `traceparent` values are normalized to the first value before `propagation.extract`, matching the OtelBridge policy; `proxy.http.request` stays a child of the propagated trace and never starts a wrong root. |

### Fresh pass over the current head — no new findings
Reviewed at `98dfbc7` (not just the delta): `proxy-telemetry-check.mjs`, `proxy-telemetry-backend.mjs`, `proxyTraceContext.ts`, `proxyTracer.ts`, and the telemetry+codex test suites.

- **Doctor logic** — `final_uniqueness`, `terminal_reconciliation`, `request_field_coverage` (including `firstOutputUnexplained` timing validation), `capture_delivery`, `sample_body_integrity` (chunk-count/contiguity/sha256 vs `redactedBodyBytes`), and `sample_trace_correlation` all bound their budgets (512-query cap, body 8 MiB, chunk ≤1000, traces `.slice(-3)`). Missing/partial/corrupt evidence returns nonzero — consistent with the stated contract.
- **Backend/query** — `organization`/`stream`/`signal` are regex/allowlist-validated and credentials travel only as fetch headers with HTTPS-outside-loopback enforced; `is_partial` and `function_error` responses are rejected. No injection or credential-leak path found.
- **Trace lifecycle** — `startProxyHttpTrace` always `releaseProxyRequestTraceContext`s in the normal `end()` path and its `try/catch` fallback; the SERVER span is closed exactly once (`ended` guard); `end()` idempotent.
- **OTLP sink / capture** (unchanged since the prior pass, re-confirmed) — exporter settles exactly once with the 6 s deadline; failure history bounded (16) with no content replay; body publication bounded by `maxPending`/`maxPendingBytes`; raw disk capture disabled in OTel-only mode; fallback records set `recordRequestMetrics: false` so no independent totals.
- **Tests** — `continuous-test-suite-proxy-telemetry.ts` covers the new `busy_collector` / `sparse_collector` doctor cases, detached sampled/unsampled contexts, W3C parent + combined-header normalization, failed-export recovery, and malformed-output timing. Suite 60/60 passes per the author; relays preserve bytes (`assertEqual(await response.text(), wire)`).

### Checked and clean
- No hardcoded credentials or secrets; header/body redaction (`SENSITIVE_HEADER_NAMES`, `SENSITIVE_BODY_KEYS`) intact in `proxyTracer.ts`.
- No provider/model/dependency changes; existing file-mode compatibility preserved.
- `docs/api/**` churn and `.yama` artifacts are generated and were not reviewed for logic.

As before, no edit tool is exposed for the earlier `yama:summary` comment (id 5671330058, `NEEDS_WORK`), so this post is the updated summary rather than an in-place edit of that text. The prior inline threads are already resolved by the author.

Tara-ag (approved):
**APPROVE** — Re-confirming at head `98dfbc7`: all previously-requested changes are verified fixed (collector_delivery no longer flags queue occupancy; detached-callback correlation preserves traceFlags; combined traceparent normalization). No open findings. All inline review threads are resolved.

This approving review supersedes the earlier CHANGES_REQUESTED (submitted against the now-outdated `8ec66d9f`) and reflects the current APPROVE verdict.

PR #1704, opened 2026-09-15
fix(vertex): stop suggesting the model that just failed

murdore:
Sweeping every feature across Vertex surfaced an error that contradicts itself. Asking for a model the project cannot serve produced:

```
Model 'gemini-3-pro-preview-11-2025' is not available in region global.
Suggested alternatives: Google Models (always available):
  • gemini-3-pro-preview-11-2025      <- the id that just failed
```

The suggestion list is a static array. It cannot know what a caller's project and region serve, so it must not claim to, and it never excluded the requested model — so the error recommended the exact id the caller had used. Anyone following that advice retries the same failing model.

This filters the requested id out of both lists, and says availability depends on the project and region rather than asserting "always available".

## Manifest correction in the same pass

`_default` declared `vision: false` and `functionCalling: false` for every Vertex model. Measured live against `gemini-2.5-flash`, `gemini-2.5-pro`, `gemini-2.5-flash-lite`, `claude-sonnet-4-6` and `claude-sonnet-4-5@20250929` — all five accepted image and PDF input and returned tool calls. The descriptor contradicted the wire, exactly as the HuggingFace manifest did in #1696.

## Coverage

Pinned end-to-end in `test/continuous-test-suite-provider-wiring.ts`: a model-unavailable error must not echo the requested id back as a suggestion, nor claim availability it cannot know. The test skips when the account can serve the model, since there is then no error to inspect.

Suite green — `Passed: 19, Total: 19, RESULT: PASS`.

## Summary by CodeRabbit

- **New Features**
  - Vertex’s default model now supports image understanding and function calling.

- **Bug Fixes**
  - Model recommendations no longer repeat the model that just failed, regardless of capitalization.
  - Availability guidance now clarifies that model access depends on the project and region, rather than claiming models are always available.
  - Suggested model lists are now presented as static candidate options, with availability determined by the selected project and region.


src/lib/providers/googleVertex/client.ts
```diff
@@ -8248,13 +8259,20 @@ export class GoogleVertexProvider extends BaseProvider {
       ],
     };
 
-    let suggestions = "\n🤖 Google Models (always available):\n";
-    availableModels.google.forEach((model) => {
+    // Never offer the id that just failed back to the caller.
+    const notRequested = (model: string): boolean =>
+      model.toLowerCase() !== (requestedModel ?? "").toLowerCase();
+    const google = availableModels.google.filter(notRequested);
```
Tara-ag: **MINOR** — the filter is an exact (case-insensitive) match, but Vertex model ids are frequently version-suffixed (`gemini-2.5-pro@002`, `gemini-3-pro-preview-11-2025@latest`). A caller that requests a versioned form of an id that the static list carries unversioned will still get the just-failed id recommended back.

```
<!-- yama:vertex-suggest-exact-match -->
```

The fix's own intent is "never offer the id that just failed". Comparing the base id (before `@`) closes the versioned variant hole without the string-prefix hazard (e.g. `gemini-2.5-flash` is a prefix of `--flash-lite`, so a `startsWith` approach is unsafe):

```suggestion
    // Never offer the id that just failed back to the caller.
    const requestedBase = (requestedModel ?? "").split("@")[0].toLowerCase();
    const notRequested = (model: string): boolean =>
      model.split("@")[0].toLowerCase() !== requestedBase;
    const google = availableModels.google.filter(notRequested);
    const claude = availableModels.claude.filter(notRequested);
```

test/continuous-test-suite-provider-wiring.ts
```diff
@@ -579,4 +579,60 @@ await test("catalog-provider enum surfaces are byte-identical to the pre-JSON-mi
...
+    throw new Skip("model is available to this account");
+  } catch (error) {
+    if (error instanceof Skip) {
+      throw error;
+    }
+    const message = error instanceof Error ? error.message : String(error);
+    if (!/not available in region/i.test(message)) {
+      throw new Skip("vertex did not report a model-availability error");
+    }
+    const suggestionsOnly = message.slice(message.indexOf("Suggested"));
```
Tara-ag: **MINOR** — `message.indexOf("Suggested")` returns `-1` when the marker is absent, and `slice(-1)` then silently turns both assertions into vacuous passes (the last char never contains the requested id, and `!includes` trivially holds). The guard above (`/not available in region/i`) makes this unreachable today because the Vertex error format always appends `Suggested alternatives:`, but this echoes the repo's own "a test that silently cannot fail is worse than no test" rule — a future change to the message header would make this test pass without ever asserting anything.

```
<!-- yama:test-suggest-slice-robustness -->
```

Harden the parse so an absent marker fails the test instead of passing it vacuously:

```suggestion
    const marker = message.indexOf("Suggested");
    if (marker < 0) {
      throw new Error('vertex error missing the "Suggested alternatives" section');
    }
    const suggestionsOnly = message.slice(marker);
```

Tara-ag (approved):
Two minor findings; both are robustness enhancements. Otherwise the change is sound.

Tara-ag:
_Superseded duplicate — see the single canonical summary below._ This comment previously carried the `<!-- yama:summary -->` marker and duplicated the findings table. To keep exactly one summary on the PR, the canonical verdict and full findings table live in the current summary comment (`5688138515`). Verdict as of the latest commit is **APPROVE**.

Tara-ag (commented):
Yama review of PR #1704 — findings pending. Submitting inline comments.

src/lib/providers/googleVertex/client.ts
```diff
@@ -8248,13 +8259,20 @@ export class GoogleVertexProvider extends BaseProvider {
       ],
     };
 
-    let suggestions = "\n🤖 Google Models (always available):\n";
-    availableModels.google.forEach((model) => {
+    // Never offer the id that just failed back to the caller.
+    const notRequested = (model: string): boolean =>
+      model.toLowerCase() !== (requestedModel ?? "").toLowerCase();
```
Tara-ag: **MINOR** — The filter is effectively exact-match against the full requested id.

`getModelSuggestions` receives the requested model id and now filters it out. But a caller can request a **version-suffixed** id (e.g. `claude-sonnet-4-5@20250929`), while some entries in the static list are base ids without the suffix. In that case the requested id wouldn't match any list entry exactly, so the failed id could still be re-offered — might even be offered as a **different** base-id entry that the caller can't serve.

This is a `getModelSuggestions` call from `getSuggestedModels`/`formatProviderError`, so an erroneous suggestion here misleads a user who is already in an error state.

Suggest comparing the base id (before any `@`) instead:

````suggestion
    // Never offer the id that just failed back to the caller.
    const notRequested = (model: string): boolean =>
      model.split("@")[0].toLowerCase() !==
      (requestedModel ?? "").split("@")[0].toLowerCase();
````

File: test/continuous-test-suite-provider-wiring.ts [hunk 4]
```diff
@@ -579,4 +579,60 @@ await test("catalog-provider enum surfaces are byte-identical to the pre-JSON-mi
...
+    });
+    throw new Skip("model is available to this account");
+  } catch (error) {
+    if (error instanceof Skip) {
+      throw error;
+    }
+    const message = error instanceof Error ? error.message : String(error);
+    if (!/not available in region/i.test(message)) {
+      throw new Skip("vertex did not report a model-availability error");
+    }
```
Tara-ag, line 1053: **MINOR** — `slice(indexOf("Suggested"))` can silently pass without asserting.

If for any reason the error text no longer contains the literal `Suggested`, `indexOf` returns `-1` and `slice(-1)` returns only the last character — the subsequent `assert(!suggestionsOnly.includes(requested))` would almost certainly still pass, turning the guard into a vacuous no-op. The region-check on line 620 already gated on the error, so this extra slice adds a failure mode without adding signal.

Prefer asserting against the full message instead of pre-slicing:

````suggestion
    assert(
      !message.includes(requested),
      "the suggestion list offers the model the caller just asked for",
    );
````

**Review by Tara-ag — APPROVED** [review]
## Yama review — APPROVE

Two MINOR, non-blocking robustness notes on the new Vertex model-suggestion logic; no CRITICAL/MAJOR findings. The manifest flip and the "don't echo the failed id" fix are correct and well covered.

Findings:
1. `client.ts:8263` — `notRequested` matches the full id only; a version-suffixed requested id could still be re-offered (compare base id before `@`).
2. `continuous-test-suite-provider-wiring.ts:623` — `slice(indexOf("Suggested"))` can silently become a vacuous pass; assert on the full message instead.

Blast radius of `getModelSuggestions` is contained (single caller in `formatProviderError`); the manifest `_default` flip has no VISION/FUNCTION_CALLING pins elsewhere. Tests are e2e, use idiomatic `Skip`, and correctly re-throw Skips + shut down the server in `finally`.

**Comment by Tara-ag** [comment]
## Yama review — APPROVE

All three findings from the previous review are fixed in the latest commit (`460c3136`). The prior CHANGES_REQUESTED review is superseded by the approving review. Verdict is now **APPROVE**.

### Findings (all resolved)

| --- | --- | --- | --- |
| MAJOR | `src/lib/providers/googleVertex/client.ts` (Claude tip) | Tip re-echoed the just-failed `requestedModel`, defeating the fix's purpose | ✅ Fixed — reworded without `requestedModel` (`yama:vertex-claude-tip-reinsertion`) |
| MINOR | `src/lib/providers/googleVertex/client.ts` (filter) | `notRequested` matched the full id only; version-suffixed requests could still re-offer the failed base id | ✅ Fixed — compares base id before `@`, case-insensitive (`yama:vertex-suggest-exact-match`) |
| MINOR | `test/continuous-test-suite-provider-wiring.ts` | `slice(indexOf("Suggested"))` could pass vacuously if the marker were absent | ✅ Fixed — explicit `if (marker < 0) throw` guard (`yama:test-suggest-slice-robustness`) |

### Checked and found clean

- **Blast radius (code graph):** `getModelSuggestions` has a single caller (`formatProviderError`, client.ts); the "always available"/"Suggested alternatives" wording is produced nowhere else.
- **Manifest flip** (`vertex.ts` `_default` `vision`/`functionCalling` → `true`): consumed as data via `constants.ts`/`manifestRegistry.ts`; nothing pins the old `false`.
- **Rules:** no Rule 1/3/4/5/6/15 violations; no leaked secrets; logs unchanged.
- **Test:** end-to-end against `dist/index.js`, idiomatic `Skip` re-thrown, `finally` server shutdown, guards on credentials and error shape.

No CRITICAL issues. Ready to merge.

**Review by Tara-ag — COMMENTED** [review]
Review of the Vertex model-suggestion fix and manifest correction.

File: src/lib/providers/googleVertex/client.ts [hunk 5]
```diff
@@ -8143,13 +8154,20 @@ export class GoogleVertexProvider extends BaseProvider {
...
+      "\n🤖 Google Models (availability depends on your project and region):\n";
+    google.forEach((model) => {
       suggestions += `  • ${model}\n`;
     });
 
     suggestions += "\n🧠 Claude Models (requires Anthropic integration):\n";
-    availableModels.claude.forEach((model) => {
+    claude.forEach((model) => {
       suggestions += `  • ${model}\n`;
     });
```
Tara-ag, line 8175: **MAJOR — the failed Claude model id is re-inserted into the error right after the filter removes it.**

The new `notRequested` filter (line 8158) correctly strips the just-failed id from both bullet lists. But the Claude-specific tip *immediately below* this hunk re-interpolates the raw `requestedModel` back into the same returned message:

```ts
// line 8175-8176 (pre-existing, not in the diff)
if (requestedModel && requestedModel.toLowerCase().includes("claude")) {
  suggestions += `\n💡 Tip: "${requestedModel}" appears to be a Claude model.\n`;
```

So for any failed request whose id contains `claude`, the returned error string **again contains the exact id that just failed** — exactly the self-echo this PR exists to eliminate. The static list of Claude ids is all `claude-*` prefixed, so this fires for every one of them (`claude-sonnet-4-5@20250929`, `claude-opus-4@20250514`, …). A caller following the tip retries the same failing model, defeating the fix's stated promise ("stop suggesting the model that just failed").

The new test does not catch it either: it only exercises the Google path (`gemini-3-pro-preview-11-2025`), so the `includes("claude")` branch never runs and the tip never fires under test.

Anchor note: the tip sits at line 8176, just past the last changed line of this hunk (8172); the comment is anchored here to stay on an in-diff line.

```suggestion
    // If the requested model looks like a Claude model, provide specific guidance
    if (requestedModel && requestedModel.toLowerCase().includes("claude")) {
      suggestions +=
        "\n💡 Tip: this model id appears to be a Claude model, which just failed.\n";
      suggestions +=
        "Ensure Anthropic integration is enabled in your Google Cloud project.\n";
      suggestions += "Try another Claude model from the list above.";
    }
```

Tara-ag (commented):
The filter + manifest change is sound; one MAJOR gap remains in the Claude guidance path and two previously-noted MINORs stay open (see inline thread #1 exact-match and thread #2 slice robustness).

src/lib/providers/googleVertex/client.ts
```diff
@@ -8143,13 +8154,20 @@ export class GoogleVertexProvider extends BaseProvider {
       ],
     };
 
-    let suggestions = "\n🤖 Google Models (always available):\n";
-    availableModels.google.forEach((model) => {
+    // Never offer the id that just failed back to the caller.
+    const notRequested = (model: string): boolean =>
+      model.toLowerCase() !== (requestedModel ?? "").toLowerCase();
+    const google = availableModels.google.filter(notRequested);
+    const claude = availableModels.claude.filter(notRequested);
```
Tara-ag, line 8164: **MAJOR — the just-failed Claude model id is still re-inserted into the error right after this filter removes it.**

The new `notRequested` filter (the block starting at line 8157) strips the failed id from the two bullet lists. But **pre-existing code immediately below this hunk** (lines 8174–8180) re-appends it back into the same returned string:

```ts
// line 8176 (not in the diff)
suggestions += `\n💡 Tip: "${requestedModel}" appears to be a Claude model.\n`;
```

Since every static Claude entry is `claude-*`-prefixed, any failed Claude request hits this branch and the returned error again contains the exact id that just failed (`claude-sonnet-4-5@20250929`, `claude-opus-4@20250514`, …). A caller following the tip retries the self-same model — the very self-echo this PR exists to eliminate.

The new test cannot catch it: it only drives the Google path (`gemini-3-pro-preview-11-2025`), so `includes("claude")` is never true and the tip never fires under test.

Fix — drop the `requestedModel` interpolation from the tip (same intent, no self-echo):

```suggestion
    // If the requested model looks like a Claude model, provide specific guidance
    if (requestedModel && requestedModel.toLowerCase().includes("claude")) {
      suggestions +=
        "\n💡 Tip: the requested id appears to be a Claude model, which just failed.\n";
      suggestions +=
        "Ensure Anthropic integration is enabled in your Google Cloud project.\n";
      suggestions += "Try another Claude model from the list above.";
    }
```

Anchored here (the `const claude = ...` filter line) because the tip block at 8176 is outside the diff hunk; the fix is in the tip immediately below.

Tara-ag (commented):
Yama inline review of the Vertex model-suggestion fix — see the MAJOR finding on the Claude guidance path and the two previously-noted open MINORs.

Tara-ag (changes requested):
## Yama review — NEEDS_WORK

The filter + manifest change is sound, but one MAJOR gap remains and two MINOR nits stay open.

- **MAJOR** (`client.ts:8176`, anchored 8172): the Claude-specific tip immediately below the new filter re-interpolates `requestedModel`, re-offering the just-failed id for every `claude-*` id — defeating this PR's stated purpose. Fix per inline `yama:vertex-claude-tip-reinsertion`.
- **MINOR** (`client.ts:8160`): `notRequested` is an exact match; version-suffixed requests can still re-offer the failed base id (`yama:vertex-suggest-exact-match`).
- **MINOR** (`continuous-test-suite-provider-wiring.ts:1054`): `slice(indexOf("Suggested"))` can pass vacuously if the marker is absent (`yama:test-suggest-slice-robustness`).

No CRITICAL issues; blast radius of the change is contained (single caller `formatProviderError`) and everything else checked clean (see summary). Please fix the MAJOR before merge.

src/lib/providers/googleVertex/client.ts
murdore: Applied in `460c3136c`. Comparing the base id before `@` rather than the whole string, which closes the versioned-variant hole without the `startsWith` hazard you flagged (`gemini-2.5-flash` being a prefix of `gemini-2.5-flash-lite`):

```ts
const requestedBase = (requestedModel ?? "").split("@")[0].toLowerCase();
const notRequested = (model: string): boolean =>
  model.split("@")[0].toLowerCase() !== requestedBase;
```

So `gemini-2.5-pro@002` now excludes the unversioned `gemini-2.5-pro` list entry it actually failed against.

test/continuous-test-suite-provider-wiring.ts
murdore: Applied in `460c3136c`. You are right that this is the repo's own "a test that silently cannot fail is worse than no test" rule — `slice(-1)` would have made both assertions vacuously true. It now fails loudly instead of passing:

```ts
const marker = message.indexOf("Suggested");
if (marker < 0) {
  throw new Error('vertex error missing the "Suggested alternatives" section', { cause: error });
}
const suggestionsOnly = message.slice(marker);
```

The message deliberately describes the discrepancy without quoting the provider payload, so it cannot be downgraded to a skip by `isExpectedProviderError`.

src/lib/providers/googleVertex/client.ts
```diff
@@ -8222,6 +8222,17 @@ export class GoogleVertexProvider extends BaseProvider {
...
+  /**
+   * Candidate model ids for a "model not available" error.
+   *
+   * This list is static. It cannot know what the caller's project and region
+   * actually serve, so it must not claim to: it previously announced
+   * "always available" and included the requested model itself, which
+   * produced errors that recommended the very id that had just failed —
+   * `gemini-3-pro-preview-11-2025` suggesting `gemini-3-pro-preview-11-2025`.
+   * The requested model is now filtered out, and the wording says these are
+   * ids to try rather than ids known to work.
```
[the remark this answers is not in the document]
murdore: Correct — the filter removed the id from the list and then the Claude tip put it straight back into the same error string. Reworded in `460c3136c` so `requestedModel` is not interpolated at all:

```
💡 Tip: this model id appears to be a Claude model, which just failed.
Ensure Anthropic integration is enabled in your Google Cloud project.
Try another Claude model from the list above.
```

The guidance keeps its meaning and the just-failed id now appears nowhere in the output.

src/lib/providers/googleVertex/client.ts
murdore: Confirmed and fixed in `460c3136c` — this is the same re-insertion as the thread above, and you are right to rate it higher than minor: the whole point of the change is "never offer the id that just failed", and the tip defeated it in the one case (Claude) where the tip fires. `requestedModel` is no longer interpolated into the tip, so the filtered list and the guidance text now agree.

src/lib/providers/googleVertex/client.ts
```diff
@@ -8143,13 +8154,20 @@ export class GoogleVertexProvider extends BaseProvider {
...
+
+    let suggestions =
+      "\n🤖 Google Models (availability depends on your project and region):\n";
+    google.forEach((model) => {
       suggestions += `  • ${model}\n`;
     });
 
     suggestions += "\n🧠 Claude Models (requires Anthropic integration):\n";
-    availableModels.claude.forEach((model) => {
+    claude.forEach((model) => {
```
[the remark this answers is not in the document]
murdore: Addressed in `460c3136c` by the base-id comparison described above — version-suffixed ids (`@002`, `@latest`) are normalised on both sides before comparing, so a versioned request no longer slips past the filter that exists to exclude it.

src/lib/providers/googleVertex/client.ts
Tara-ag: Resolved: current HEAD rewords the Claude tip without interpolating `requestedModel`, so the just-failed id is no longer echoed after the filter removes it. Nice, faithful to the diff now. ✅

src/lib/providers/googleVertex/client.ts
Tara-ag: Resolved: `notRequested` now compares the base id before `@` (case-insensitive), closing the version-suffixed hole. ✅

test/continuous-test-suite-provider-wiring.ts
Tara-ag: Resolved: the test now throws when the `Suggested` marker is absent (`if (marker < 0) throw …`) instead of degrading to a vacuous `slice(-1)` pass. ✅

Tara-ag (approved):
## Yama review — APPROVE

All three findings from the prior round (MAJOR Claude-tip re-echo, MINOR exact-match filter, MINOR test slice robustness) are fixed in the current HEAD and their threads resolved. No new issues. The manifest `vision`/`functionCalling` correction is a data-truth fix consistent with live Vertex behaviour and the accepted #1696 precedent. See the summary comment for the full breakdown.

Tara-ag (approved):
## Yama review — APPROVE

All three findings from the previous review are fixed in commit `460c31366`:

- **MAJOR — Claude tip re-echo** (`yama:vertex-claude-tip-reinsertion`): reworded to no longer interpolate the just-failed `requestedModel`.
- **MINOR — exact-match filter** (`yama:vertex-suggest-exact-match`): now compares the base id before `@`, case-insensitively.
- **MINOR — test slice robustness** (`yama:test-suggest-slice-robustness`): explicit `marker < 0` guard fails instead of passing vacuously.

Blast radius is contained (single caller `formatProviderError`); the manifest `_default` flip is data-only. No CRITICAL issues. Approving.

Tara-ag (approved):
## Yama review — APPROVE

All three findings from the previous review are fixed in commit `460c3136`:

- **MAJOR — Claude tip re-echo** (`yama:vertex-claude-tip-reinsertion`): tip reworded, no longer interpolates the just-failed `requestedModel`.
- **MINOR — exact-match filter** (`yama:vertex-suggest-exact-match`): filter compares the base id before `@`, case-insensitively.
- **MINOR — test slice robustness** (`yama:test-suggest-slice-robustness`): explicit `marker < 0` guard fails instead of passing vacuously.

Blast radius contained (single caller `formatProviderError`); manifest `_default` flip is data-only. No CRITICAL issues. Approving.

PR #1705, opened 2026-09-16
fix(streaming): keep toolsUsed/toolExecutions live through nl.stream()

murdore:
`nl.stream()` returned a `StreamResult` whose `toolsUsed` and `toolExecutions` were **missing as own properties** — not `undefined`, absent — for every provider, even when a tool demonstrably ran. `toolsUsed?: string[]` is declared on `StreamResult` in `src/lib/types/stream.ts`, so this is a documented contract, not an undocumented field.

## Root cause — two additive bugs at the shared layer

1. **Naked object spreads.** `withStreamModelFallback()` and `wrapStreamWithLifecycleCallbacks()` in `baseProvider.ts` each rebuilt the result with `{ ...result, stream: ... }`. A spread *invokes and flattens* getters — so providers exposing these fields as lazy accessors (Vertex, Google AI Studio, the whole OpenAI-compatible family) had them snapshotted at their initial empty state before the consumer pulled a single chunk.
2. **Hardcoded field lists.** `createMCPStream()` and `createStreamResponse()` in `neurolink.ts` rebuilt `StreamResult` from hand-written field lists that never mentioned either field. For providers like Bedrock — whose telemetry exists only at NeuroLink's own MCP-orchestration level — this was a total loss.

The fix adds `preserveLiveStreamAccessors` / `pickLiveStreamDescriptors`, which copy property **descriptors** rather than values, applied at every boundary. Fields stay declared and lazy end-to-end instead of being eagerly snapshotted (permanent zeros) or dropped.

## Two sibling sites, found by audit

Both were confirmed by live reproduction, not inference:

- **`executeFakeStreaming()`** — reached whenever `stream()` detects 3+ image parts in the input, or a provider errors out of real streaming. Same hardcoded-field-list bug. This one had already manifested naturally: an earlier version of this suite, run unmodified 189 times against the fixed build, failed once with the original defect's exact signature.
- **`streamWithIterationFallback()`** — backs the public, documented `providerFallback` / `modelChain` options. Same naked spread.

## Coverage

`test/continuous-test-suite-stream-tool-telemetry.ts` spins up a local OpenAI-compatible fixture — no credentials, no network egress, CI-safe — and drives a real two-turn tool round trip through `nl.stream()` across four paths:

1. SSE streaming
2. The video-frame-detected fake-streaming fallback
3. `modelChain` fallback orchestration
4. A mid-stream failover: the first attempt's connection drops *after* the result object has been handed to the caller, and a retried attempt runs the tool

**All four fail against the pre-fix code and pass after** (verified by reverting the production hunks and rebuilding: `Passed: 0 / Failed: 4` → `Passed: 4 / Total: 4`). **120 consecutive runs of the suite: 120 pass, 0 fail** — the ~0.5-1% natural flake traced to `executeFakeStreaming()` is gone.

## A review question, measured and answered

Review raised that `streamWithIterationFallback()` captures its accessors from the *first* attempt, and so might misreport telemetry after a genuine mid-stream fallback — describing the abandoned attempt rather than the one that actually streamed.

**Measured, it does not.** Case 4 above exists to check exactly this. The request sequence confirms the fallback really fires (attempt 1 drops; attempts 2 and 3 are the retry's two turns), the tool runs once on the retry, and `toolsUsed` / `toolExecutions` report it correctly — including on a build with only the original fix applied.

The reason is that the accumulator the accessor closes over belongs to NeuroLink's own orchestration layer and is shared across attempts within a single `stream()` call, rather than being owned per attempt. A speculative fix threading the live attempt through to the return site was written, measured to change nothing observable, and dropped rather than shipped unproven. Case 4 is kept as a permanent pin on the behaviour.

## Summary by CodeRabbit

* **Bug Fixes**
  * Stream results now reliably retain current tool usage and execution details throughout streaming, including fallback and retry scenarios.
  * Tool telemetry is now available for simulated streaming responses and remains accurate after stream completion.

* **Tests**
  * Added regression coverage for tool telemetry across standard streaming, fallback, failover, and simulated streaming paths.


src/lib/utils/streamResultAccessors.ts
```diff
@@ -0,0 +1,63 @@
...
+ * Re-apply every getter-defined accessor from `source` onto `target`,
+ * in place. Plain data properties (arrays copied by reference, primitives)
+ * are left untouched — an object spread already preserves those correctly,
+ * since copying an array reference keeps observing later `.push()`es.
+ */
+export function preserveLiveStreamAccessors<T extends object>(
+  source: T,
+  target: T,
+): T {
+  for (const key of Object.keys(source)) {
```
Tara-ag: `<!-- yama:stream-accessors-enumerability -->`

**MINOR — `preserveLiveStreamAccessors` iterates `Object.keys`, which only sees *enumerable* own properties.**

This helper is the guard `withStreamModelFallback()` / `wrapStreamWithLifecycleCallbacks()` now rely on in `baseProvider.ts` to keep a provider's lazy `toolsUsed`/`toolExecutions` alive. But the loop only re-applies descriptors for keys that both (a) are enumerable and (b) survive the spread. Its sibling `pickLiveStreamDescriptors()` is verdan by contrast — it uses `Object.getOwnPropertyDescriptor()` directly and is immune to enumerability.

All three providers that currently define these fields (googleVertex, googleAiStudio, `openaiChatCompletionsBase`) happen to use `enumerable: true, configurable: true, get: …`, so today `Object.keys` finds them and the fix works. That coincidence is exactly what the defensive contract in this PR advertises protection against, though: "any other field a provider defines via `Object.defineProperty` with only a `get`". A future provider that defines `toolExecutions` as a **non-enumerable** getter (a common `defineProperty` default) would be silently skipped here — the field would drop again precisely as in the original defect, and this new code path would be the silent fail point.

Suggestion — enumerate own *names* instead of enumerable keys so the two helpers behave identically:

```suggestion
  for (const key of Object.getOwnPropertyNames(source)) {
```

(Enumeration order is preserved by `getOwnPropertyNames`, so no behavioral change for the current providers; the mere addition of a `get` check already filters plain data properties.)

Tara-ag (approved):
The descriptor-preservation approach is sound and the three end-to-end regression cases meaningfully exercise the shared BaseProvider/NeuroLink rebuild paths against a real OpenAI-compatible wire shape. One MINOR robustness note on `preserveLiveStreamAccessors` (inline). Approving.

Tara-ag:
## APPROVE

Recurring review at head `92e730d`. The fix is sound and complete: `nl.stream()` now keeps `toolsUsed`/`toolExecutions` alive end-to-end by re-applying property **descriptors** (not snapshot values) across every `StreamResult` rebuild boundary, and the four e2e regression cases drive a real OpenAI-compatible tool round trip (SSE streaming, fake-streaming, `modelChain`, and a mid-stream failover) through the shared `BaseProvider` / `NeuroLink` paths, all failing pre-fix and passing post-fix.

### Findings

| Severity | Location | Description | Status |
|----------|----------|-------------|--------|
| MINOR | `src/lib/utils/streamResultAccessors.ts` (`preserveLiveStreamAccessors`, line 32) | Earlier finding: iterating `Object.keys` (enumerable-only) would silently skip a future provider that defines `toolsUsed`/`toolExecutions` as a non-enumerable getter. | **RESOLVED by code change** — current head iterates `Object.getOwnPropertyNames(source)` and gates on `descriptor.get`, matching `pickLiveStreamDescriptors`. Not a live bug anyway (all three producers define `enumerable: true` accessors), but the hardening is now in place. |

### What was checked and found clean (this pass)

- **Mechanism (via code graph, built at this head)** — `preserveLiveStreamAccessors` is applied at the two naked-spread sites in `baseProvider.ts` (`withStreamModelFallback` line 728, `wrapStreamWithLifecycleCallbacks` line 932) and at `NeuroLink.streamWithIterationFallback` (line 9341); `executeFakeStreaming` calls `transformToolExecutions` (line 1224) and attaches `toolExecutions` via a descriptor. `pickLiveStreamDescriptors` is applied in `createMCPStream` (both the ModelPool and direct branches) and in `createStreamResponse` via the new optional `toolTelemetrySource` field. Blast radius is internal to stream orchestration — `runStandardStreamRequest → createMCPStream → createStreamResponse` and `NeuroLink.stream → streamWithIterationFallback`; no unmodified caller is left handling a different return shape and **Rule 5 (public SDK API) is unaffected**. `toolTelemetrySource` is optional and defaulted to `{}`, so `createStreamResponse` is backward compatible.
- **`executeFakeStreaming` semantics** — correct by design: it wraps a settled `generate()` result, so `toolsUsed` is written as a plain data property (no lazy getter to preserve) and `toolExecutions` is a static descriptor from `transformToolExecutions()`. Snapshotting here is correct; the descriptor-copying is only needed at genuinely-live boundaries.
- **`streamWithIterationFallback` mid-stream fallback** — author's measured answer (shared orchestration accumulator plus Case 4 as a permanent pin) is accepted; the regression test verifies telemetry describes the attempt that actually streamed.
- **Tests (Rule 15)** — e2e against `dist/index.js` with `assertDistFresh()`; self-contained loopback `127.0.0.1` fixtures with `offline: true` (no network egress, no creds); single module graph; `test:stream-tool-telemetry` registered in `package.json`. All four cases assert **own-property presence** (`hasOwnProperty`), matching the documented contract.
- **Docs** — `docs/api/classes/NeuroLink.md` is line-number regeneration only (no API shape change).

src/lib/utils/streamResultAccessors.ts
Tara-ag: **MINOR** — `preserveLiveStreamAccessors` only sees enumerable getters, but the sibling helper does not.

`preserveLiveStreamAccessors` iterates `Object.keys(source)`, which returns **enumerable** own properties only. The very defect this PR fixes is a getter that silently re-drops a field through a naked spread; the sibling `pickLiveStreamDescriptors` deliberately uses `Object.getOwnPropertyDescriptor` (works for non-enumerable too). A future provider that defines `toolsUsed`/`toolExecutions` as a non-enumerable getter (fully legal — `defineProperty` defaults `enumerable` to `false`) would pass through `pickLiveStreamDescriptors` at the neurolink.ts boundaries but then get re-frozen away by `preserveLiveStreamAccessors` in `BaseProvider.stream()`. Not hit by any provider today — all three define `enumerable: true`, `configurable: true` — so this is hardening, not a live bug.

Fix: iterate all own keys instead of enumerable ones.

```suggestion
  for (const key of Object.getOwnPropertyNames(source) as (keyof T)[]) {
```

<details><summary>why the two helpers disagree</summary>

- `src/lib/utils/streamResultAccessors.ts:32` — `Object.keys(source)` (enumerable only)
- `src/lib/utils/streamResultAccessors.ts:~56` — `Object.getOwnPropertyDescriptor(source, key)` (robust to non-enumerable)

Verified the three live producers all set `enumerable: true` (`googleVertex` / `googleAiStudio` getters and the openai‑chat‑completions `Object.defineProperty(result, "toolExecutions", { get, enumerable: true, configurable: true })`), so this does not fire today.
</details>

Tara-ag (approved):
Approving. The descriptor-preservation fix is sound end-to-end; one MINOR hardening suggestion (non-enumerable getters) noted inline.

Tara-ag:
**Superseded duplicate** — this issue comment duplicated the authoritative review summary. That single authoritative summary carries the current APPROVE verdict and lives at [the summary comment](#issuecomment-5703015502). Please disregard this comment; it is retained only because API-level deletion of issue comments is not available in this environment. The authoritative summary is the only one that counts.

src/lib/utils/streamResultAccessors.ts
Tara-ag: Keeping this MINOR open — no author reply since the previous review, and the code is unchanged: `preserveLiveStreamAccessors` still iterates `Object.keys` (enumerable-only) while its sibling `pickLiveStreamDescriptors` uses `getOwnPropertyDescriptor`. Not a live bug today (all three producers — googleAiStudio, openaiChatCompletionsBase, googleVertex — define these as `enumerable: true, configurable: true` accessors), so `Object.getOwnPropertyNames` is a hardening fix for the same silent-re-drop the PR explicitly guards against.

Tara-ag (approved):
Approving (recurring review at current head `9d575ca`). The descriptor-preservation fix is sound end-to-end across every StreamResult rebuild boundary, the regression suite is Rule-15 compliant, and the code-graph blast-radius trace found no unmodified caller affected. One MINOR hardening note (enumerability in `preserveLiveStreamAccessors`) is tracked inline and remains open.

src/lib/utils/streamResultAccessors.ts
Tara-ag: **Resolved by code change** — recurring review at head `92e730d`. `preserveLiveStreamAccessors` now iterates `Object.getOwnPropertyNames(source)` and re-applies the getter descriptor when `descriptor.get` is a function, so a future provider that defines `toolsUsed`/`toolExecutions` as a non-enumerable accessor is no longer silently skipped. The discrepancy with `pickLiveStreamDescriptors` (which uses `Object.getOwnPropertyDescriptor`) that this thread flagged is now closed — both helpers enumerate all own property names. Marking resolved; nothing further to change.

Tara-ag (approved):
Approving at current head `92e730d`. The descriptor-preservation fix keeps `toolsUsed`/`toolExecutions` live end-to-end across every `StreamResult` rebuild boundary; the e2e regression suite is Rule-15 compliant; blast-radius is internal to stream orchestration (Rule 5 unaffected). The one MINOR (enumerability in `preserveLiveStreamAccessors`) is resolved by the current code and its inline thread is closed.

PR #1706, opened 2026-09-16
fix(vertex): stop dropping Claude extended-thinking output on Vertex

murdore:
Claude-on-Vertex's native paths (`executeNativeAnthropicStream` / `Generate`, which bypass `@ai-sdk/google-vertex` for `@anthropic-ai/vertex-sdk` directly) discarded extended thinking end to end. `thinkingConfig` was accepted and silently ignored — no error, no reasoning text, `usage.reasoning` zero.

## Three separate losses

1. `thinkingConfig` was never translated into the Messages API's `thinking: {type:"enabled", budget_tokens}` request field.
2. Even once sent, the response pump forwarded only `content` from the shared agentic-loop engine, dropping the `reasoning` field that `anthropic/loopAdapter.ts` already derives from `thinking` / `redacted_thinking` blocks.
3. `usage.reasoning` was never populated — Anthropic's real `output_tokens_details.thinking_tokens` was read nowhere in this codebase.

That the capability was ours to send, not a vendor limit, was established first by bypassing NeuroLink entirely: a raw `@anthropic-ai/vertex-sdk` call with the same project, region, credentials and model returned a real thinking block.

## Verification

Live against Vertex `claude-sonnet-4-5` with a reasoning-heavy prompt: `generate()` and `stream()` both return non-empty `reasoning` text and a positive `usage.reasoning`. Reverting to the `thinkingConfig`-only state reproduces `reasoning.length=0` / `usage.reasoning=0` on both paths — a real red, not a reconstructed one.

## A regression this fix introduced, and closes

Worth calling out explicitly, because it would have shipped: `schema` mode sets `useFinalResultTool`, which unconditionally forces `tool_choice:{type:"any"}`. Once `thinking` started riding along on every request, that combination became a **live HTTP 400** — *"Thinking may not be enabled when tool_choice forces tool use"* — on both `generate()` and `stream()`. Before this branch the same call merely ignored `thinkingConfig`. So a working call would have started hard-failing.

`buildClaudeThinkingParam` now takes the `useFinalResultTool` flag already computed at each call site and omits `thinking` in that mode, restoring the silently-ignored behaviour rather than crashing. This mirrors the existing guard in `src/lib/proxy/oauthFetch.ts`, which already deletes `thinking` when `tool_choice.type` is `"any"` or `"tool"`.

The step/context-cap forced-tool finalization backstop inherits the omission for free (it spreads `...requestParams`). The tools-disabled backstop uses `tool_choice:{type:"none"}`, which does not force tool use and was never affected.

Verified live: `schema` + `thinkingConfig.enabled` now returns valid `structuredData` instead of a 400, with the 400 reproduced first against the unfixed hunk (real Vertex request ids captured). Plain thinking and tools+thinking both still work.

## Known follow-ups (out of scope, disclosed)

- Cross-turn conversation-history replay of thinking blocks is not addressed — a deeper, separate gap.
- Direct Anthropic's `client.ts` still surfaces reasoning *text* but no token count, leaving an inconsistency between two providers sharing `loopAdapter.ts`. Pre-existing, unchanged here.
- `NeurolinkCredentials.vertex.clientEmail`/`.privateKey` are declared in types but never consumed by `GoogleVertexProvider`'s constructor — only `GOOGLE_APPLICATION_CREDENTIALS` works for Vertex service-account auth on this path.

## Summary by CodeRabbit

- **New Features**
  - Added support for Claude extended thinking when using the Vertex provider.
  - Reasoning text and token usage are now available separately from the final response for both generated and streamed results.
  - Thinking budgets are automatically constrained to valid limits.
  - Structured-output requests remain compatible by safely omitting unsupported thinking configuration.
  - Standard tool-enabled requests continue to support extended thinking.


src/lib/providers/googleVertex/client.ts
```diff
@@ -3864,6 +3864,69 @@ export class GoogleVertexProvider extends BaseProvider {
...
+    // Structural, not TextGenerationOptions: the streaming caller passes
+    // StreamOptions, and only these two fields are read.
+    options: { thinkingConfig?: { enabled?: boolean; budgetTokens?: number } },
+    maxTokens: number,
+    useFinalResultTool: boolean,
+  ): { type: "enabled"; budget_tokens: number } | undefined {
+    if (useFinalResultTool) {
+      if (options.thinkingConfig?.enabled) {
+        logger.debug(
+          "[GoogleVertex] Omitting thinking: schema mode forces tool_choice:any, which Anthropic rejects alongside thinking",
```
Tara-ag, line 3914: **MINOR** — the budget clamp is correct but more convoluted than it needs to be, and one branch is effectively dead.

`Math.max(1024, Math.min(budget, maxTokens - 1024))` always yields ≥ 1024, so `if (clamped >= maxTokens) return undefined` can only ever fire when `maxTokens <= 1024`. For every *valid* thinking range (`maxTokens > 1024`) that guard is unobservable — the real "too small to think" case is already decided by the `Math.max(1024, …)` floor. Two concrete consequences worth a tighter formulation:

- When `1025 <= maxTokens < 2048`, `maxTokens - 1024 < 1024`, so the caller's requested `budgetTokens` is *always* coerced to exactly `1024` regardless of what they asked for (and thinking becomes enabled even if their intent was not to think at this size).
- The `clamped >= maxTokens` check reads as a live ceiling guard but is dead code in the only range where thinking is possible.

Suggest an explicit early-out that states both bounds directly, e.g.:

````suggestion
    if (maxTokens <= 1024) {
      return undefined;
    }
    const clamped = Math.min(budget, maxTokens - 1024);
    if (clamped < 1024) {
      return undefined;
    }
    return { type: "enabled" as const, budget_tokens: clamped };
````

Not blocking — the current expression is functionally sound; this is to make the ceiling/floor behaviour legible (and it drops the silent 1024-coercion in the `1025..2047` band, or at minimum makes it an explicit decision).

File: src/lib/providers/googleVertex/client.ts [hunk 2]
```diff
@@ -6064,15 +6175,21 @@ export class GoogleVertexProvider extends BaseProvider {
...
     // NEWLINE, which a chunk-by-chunk `+=` here would silently drop, running
-    // consecutive steps' text together. The drain still has to happen:
-    // leaving the channel unread stalls the engine once its buffer fills.
+    // consecutive steps' text together. Reasoning has no such step-boundary
+    // hook (buildToolResultMessages only runs for steps WITH tool calls, so
+    // a plain one-shot thinking answer never reaches it), so it is captured
+    // directly off the chunk stream instead. The drain still has to happen
+    // regardless: leaving the channel unread stalls the engine once its
+    // buffer fills.
     const pump = (async () => {
```
Tara-ag, line 6252: **MINOR** — in `generate()` the per-step reasoning text is concatenated with a bare `+=`, with no step-boundary separator.

For a single-step turn this is fine, but on a multi-step tool-calling turn the reasoning produced by step N and step N+1 runs together with no delimiter, e.g.:

````suggestion
        if (chunk.reasoning) {
          accumulatedReasoningText +=
            (accumulatedReasoningText ? "\n" : "") + chunk.reasoning;
        }
````

Compare the text path, whose own comment here notes that `appendStepText` deliberately joins steps with a NEWLINE to avoid running consecutive steps together. Reasoning has no equivalent hook, so a generated `result.reasoning` for a multi-step turn is one ambiguous blob. A newline join (or an explicit note that concatenation is intentional) keeps the field self-describing. Non-blocking observability nit.

**Review by Tara-ag — APPROVED** [review]
Solid, well-tested fix. Reasoning content + token counts are now wired through on Vertex's native Claude paths; the schema/`tool_choice` interaction is correctly guarded to avoid the live 400; accounting is additive and consistent with the sibling (googleAiStudio) and BaseProvider conventions. Two MINOR non-blocking observations inline (clamping legibility, generate() reasoning concatenation across steps).

**Comment by Tara-ag** [comment]
## ✅ APPROVE — fix(vertex): stop dropping Claude extended-thinking output on Vertex

This change correctly wires Claude extended thinking through Vertex's native Anthropic-SDK paths (`stream()` and `generate()`), surfaces reasoning text and token counts (`reasoning` / `reasoningTokens` / `usage.reasoning`), and closes the regression the fix itself would have introduced: `schema` (forced `tool_choice:{type:"any"}`) + `thinking` no longer produces the live 400, mirroring the existing guard in `oauthFetch.ts`. Reasoning accounting is additive and never double-counted into `output`/`total`. All review findings are resolved.

### Findings
Both MINOR inline findings from the initial pass are **resolved** in `b4eede9fe`, with the author's reply on each thread recording the outcome:

|----------|----------|---------|--------|
| MINOR | `googleVertex/client.ts:3914` | Budget clamp had a dead `clamped >= maxTokens` branch and coerced the caller's budget to 1024 in the `1025–2047` band. | Applied — two floors now checked explicitly; thinking correctly omitted below the range. |
| MINOR | `googleVertex/client.ts:6252` | `generate()` concatenated per-step reasoning with a bare `+=` (no step separator). | Not applied, justified — the `+=` joins thinking deltas of one continuous block, not steps; comment added at the site. |

CodeRabbit's three comments (backstop thinking rebuild, backstop reasoning accumulation, `isInfraError` narrowing) were all addressed in `b4eede9fe` and resolved; no remaining review concerns.

- **Schema/thinking 400 guard** — `useFinalResultTool` correctly omitted `thinking` on both stream & generate request builders; matches `oauthFetch.ts` precedent. Clean.
- **Reasoning token accounting** (`loopAdapter.ts`) — incremental delta tracking across `message_start`/`message_delta`, reported as a subset of output and never folded into `outputTokens`; no double count. Correct.
- **Backward compatibility (rule 5)** — all additions are conditional spreads; no public SDK field is removed or reshaped. `baseProvider.ts` already documents `usage.reasoning` as the Vertex convention, and `analyticsUtils.ts` reads it. No regressions to unmodified callers found.
- **Public result fields** — top-level `reasoning`/`reasoningTokens` + `usage.reasoning` on generate; reasoning delivered as `{ content:"", reasoning }` chunks on stream, consistent with the direct-Anthropic provider.
- **Test (rule 15)** — `continuous-test-suite-vertex-claude-thinking.ts` imports from `../dist/index.js` + helpers only (no `src/`), e2e-compliant, no determinism-exception `allow`-list entry needed.
- **Safety** — no secrets, no logging of raw keys/tokens. Verified the diff, the Shared agentic loop, callers of `readAnthropicStep` (`executeStep` only), and the 65-file impact radius.

### Out-of-scope (author-disclosed, agreed non-blocking)
Cross-turn replay of thinking blocks and the direct-`client.ts` reasoning-token-count inconsistency are pre-existing/separate gaps; `NeurolinkCredentials.vertex` clientEmail/privateKey being unused is pre-existing.

Verdict **APPROVE** — supported by an approving review on the PR; one summary comment, no duplicates.

File: src/lib/providers/googleVertex/client.ts (hunk 1 above)
Comment (murdore) [reply src/lib/providers/googleVertex/client.ts:3914]: Applied in `b4eede9fe`. You were right that `Math.max(1024, …)` made the `clamped >= maxTokens` guard unobservable for every valid range. The two floors are now checked explicitly:

```ts
if (maxTokens <= 1024) return undefined;
const clamped = Math.min(budget, maxTokens - 1024);
if (clamped < 1024) return undefined;
```

This is a real behaviour change in the 1025–2047 band you identified: previously the budget was silently coerced to exactly 1024 no matter what the caller asked for; now thinking is omitted, because there is genuinely no budget that satisfies both Anthropic's 1024 minimum and this function's 1024-token reservation for the answer.

src/lib/providers/googleVertex/client.ts
murdore: Not applied — and I want to give the reason rather than silently skip it, because the text path's newline join really is correct and this one is not the same case.

`appendStepText` joins across STEPS. The reasoning here is accumulated from `chunk.reasoning` on the engine stream, and the Vertex loop adapter pushes one chunk per `thinking_delta`, not one per step — so the deltas arriving at this `+=` are fragments of a single continuous thinking block. Joining them with `\n` would insert a break before nearly every delta and shred one coherent block into roughly a line per token.

Concatenation ACROSS steps is a real blur and your instinct about it is right; there is simply no per-step boundary on this channel to join on, so a newline here would buy a wrong delimiter in the common case to fix a rarer one. I have added a comment at the site recording exactly this, so the next reader does not have to re-derive why it differs from the text path.

src/lib/providers/googleVertex/client.ts
```diff
@@ -4176,13 +4239,24 @@ export class GoogleVertexProvider extends BaseProvider {
...
+
     const requestParams: Parameters<typeof client.messages.stream>[0] = {
       model: modelName,
       // Default to the model's real output ceiling (e.g. 64K for Sonnet 4.x)
       // instead of the legacy 4096, which silently truncated large structured
       // responses mid-JSON. resolveClaudeMaxTokens also clamps over-large
       // caller values so the native Vertex path never 400s.
-      max_tokens: resolveClaudeMaxTokens(modelName, options.maxTokens),
+      max_tokens: streamMaxTokens,
+      ...(streamThinking && { thinking: streamThinking }),
```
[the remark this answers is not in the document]
murdore: Confirmed and fixed in `b4eede9fe` — this was the most consequential finding on the PR. `terminalMaxTokens` can land far below the ceiling the thinking budget was sized against, and `requestParams` carried that stale `thinking` through the spread, so the backstop call could go out with `budget_tokens >= max_tokens` and be 400'd by Anthropic.

The budget is now rebuilt for the shrunk limit and assigned outright:

```ts
const backstopThinking = this.buildClaudeThinkingParam(options, terminalMaxTokens, false);
… { ...requestParams, max_tokens: terminalMaxTokens, thinking: backstopThinking, … }
```

Assigned rather than conditionally spread on purpose: a `...(x ? {thinking: x} : {})` would leave the stale value in place when the rebuild returns undefined, which is precisely the case that 400s. Both the streaming and non-streaming backstop paths are fixed.

src/lib/providers/googleVertex/client.ts
```diff
@@ -4858,6 +4942,12 @@ export class GoogleVertexProvider extends BaseProvider {
         if (engineResult) {
           usage.input += engineResult.usage.inputTokens;
           usage.output += engineResult.usage.outputTokens;
+          // A SUBSET already included in outputTokens above — added here
+          // only for observability, never folded into input/output/total.
+          if (engineResult.usage.reasoningTokens) {
+            usage.reasoning =
+              (usage.reasoning ?? 0) + engineResult.usage.reasoningTokens;
+          }
```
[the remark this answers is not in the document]
murdore: Applied in `b4eede9fe` on both backstop paths — `usage.reasoning` on the stream side and `totalReasoningTokens` on the generate side now accumulate `output_tokens_details.thinking_tokens`, matching what the main loop already does. Reasoning blocks emitted on the backstop turn are also extracted and surfaced now, rather than being dropped.

The accumulation is additive and the comment records why that is right: Anthropic bills thinking as part of `output_tokens`, so this is a SUBSET of a total already counted, not a separate charge to add on top.

test/continuous-test-suite-vertex-claude-thinking.ts
```diff
@@ -0,0 +1,312 @@
...
+const REASONING_PROMPT =
+  "A parking lot has only cars and motorcycles. Together there are 27 " +
+  "vehicles and 74 wheels. Cars have 4 wheels, motorcycles have 2. How " +
+  "many cars and how many motorcycles are there? Reply with only the two " +
+  "numbers separated by a comma, nothing else.";
+
+function isInfraError(message: string): boolean {
+  return /api key|apikey|credential|security token|unauthor|permission|quota|rate.?limit|too many requests|429|not found|unknown model|model.*not|region|ENOTFOUND|ECONNREFUSED|ECONNRESET|socket hang|fetch failed|network|timeout|deadline|unavailable|overloaded|throttl|capacity|exhausted|billing|credits|insufficient|payment|402|access|forbidden|invalid.*model|does not exist|status (429|500|502|503|504)|internal server|service unavailable/i.test(
+    message,
+  );
```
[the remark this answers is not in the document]
murdore: Addressed in `b4eede9fe`. The suite's private `isInfraError` regex is gone; it now imports `isExpectedProviderError` from the harness, so this suite classifies provider-side failures by the same rule as the rest of the repo instead of drifting from it.

PR #1707, opened 2026-09-16
fix(gemini): translate thinkingLevel to a numeric budget for Gemini 2.5

murdore:
Vertex and Google AI Studio both reject `thinkingConfig.thinkingLevel` on every Gemini 2.5 model with HTTP 400 `INVALID_ARGUMENT`. Only Gemini 3 accepts `thinkingLevel` on the wire; Gemini 2.5 wants a numeric `thinkingConfig.thinkingBudget`.

## Measured, not assumed

The mapping was designed from the vendor's actual behaviour rather than from documentation — designing it the other way is how the bug got here. Verified live against raw Vertex `generateContent` for pro, flash and flash-lite; independently confirmed by Google's own docs (*"If you use the thinking_level parameter with a model earlier than Gemini 3, the model returns an error"*); and separately verified against the Google AI Studio (Generative Language API) endpoint for pro and flash (*"Thinking level is not supported for this model."*).

The symptom set was inconsistent enough to be worth recording: `2.5-pro` hard-400d, `2.5-flash-lite` silently returned `reasoning=0` across three runs, and `2.5-flash` worked — while `thinkingConfig.budgetTokens` worked fine on `2.5-pro`. So the capability existed; NeuroLink was sending the wrong field shape.

## Fix

`createNativeThinkingConfig` now takes the resolved model name. For a non-Gemini-3 model it looks up a vendor-verified per-model budget range (`getGemini25ThinkingBudgetRange`) and translates the requested qualitative level into a budget within it, omitting `thinkingLevel` from the payload entirely. Gemini 3 — and any caller omitting `modelName` — keeps the original passthrough unchanged. An unrecognized non-Gemini-3 model gets no `thinkingConfig` at all rather than a guessed shape.

`thinkingLevel` and `thinkingConfig.budgetTokens` stay contractually distinct (Gemini lever vs Anthropic lever); this does not collapse them.

## Both native-SDK call sites are wired

Vertex's `executeNativeGemini3Generate/Stream`, **and** Google AI Studio's shared `googleNativeGemini3/utils.ts::buildNativeConfig`.

The AI Studio site was left unwired in this branch's first pass and caught by audit. It reached the identical helper without a model name, so an AI Studio caller requesting `thinkingLevel` on a Gemini 2.5 model hit the same 400 this fix exists to eliminate. Fixed by threading `buildNativeConfig`'s own `options.model` — already populated by both its callers — through as the second argument, mirroring the Vertex sites exactly. No new detection logic.

## Coverage — two live end-to-end suites

- **`test:gemini-thinking-level`** (Vertex): no 400 and `reasoningTokens > 0` on `gemini-2.5-pro`/`flash`/`flash-lite` with `thinkingLevel:"high"`; `minimal` produces a smaller budget than `high` on flash-lite; `gemini-3-flash-preview` passthrough unaffected.
- **`test:gemini-thinking-level-ai-studio`**: the same assertions against `gemini-2.5-flash`/`pro` with a `gemini-3-flash-preview` control.

Red on the unfixed build was `Passed: 1 / Failed: 4` — all three 2.5 models 400ing, only the Gemini 3 control passing. Green is `Passed: 5 / Total: 5`. The AI Studio gap was independently vacuity-checked: removing just the `options.model` argument reproduces three live 400s, restoring it gives 4/4.

`gemini-2.5-flash-lite` is excluded from the AI Studio suite — live-probed against these credentials it returns HTTP 404 *"no longer available to new users"*, an unrelated account-level deprecation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed extended-thinking requests for Gemini 2.5 models, which could previously fail when using thinking levels.
  - Thinking levels now translate to model-compatible reasoning budgets for Gemini 2.5.
  - Preserved existing thinking-level behavior for Gemini 3 models.

- **Documentation**
  - Updated Gemini model compatibility guidance to cover both Gemini 2.5 and Gemini 3, including their supported thinking configuration formats.


src/lib/utils/thinkingConfig.ts
```diff
@@ -85,33 +89,85 @@ export function createThinkingConfigFromRecord(
...
+    if (budgetRange) {
+      return {
+        includeThoughts: true,
+        thinkingBudget: mapThinkingLevelToBudget(level, budgetRange),
+      };
+    }
+    // Not a Gemini 3 model and not a Gemini 2.5 model we have a verified
+    // budget range for: no evidence this model accepts either parameter, so
+    // omit thinkingConfig rather than guess a shape that might itself 400.
+    return undefined;
```
Tara-ag, line 182: `<!-- yama:gemini-silent-drop -->`

**MINOR — thinking is silently dropped for non-Gemini-2.5/3 models instead of surfacing.**

The new branch omits `thinkingConfig` entirely for any model that is neither Gemini 3 nor a recognized Gemini 2.5. `googleAiStudio/client.ts` routes **all** Gemini models (including `gemini-2.0-flash` and variants) through `buildNativeConfig` → this function, so a caller requesting `thinkingConfig: { thinkingLevel: "high" }` on a Gemini 2.0 model goes from "provider rejects the field" to **silent no-thinking** — the request succeeds but never thinks, with no log entry.

The choice is defensible ("omit rather than guess a shape that might 400"), but the silent-ness is the issue: the user opted into thinking and gets nothing, indistinguishable from "thinking is off". Suggest logging a warning once, e.g.:

```suggestion
    // Not a Gemini 3 model and not a Gemini 2.5 model we have a verified
    // budget range for: no evidence this model accepts either parameter, so
    // omit thinkingConfig rather than guess a shape that might itself 400.
    logger.warn(
      `[thinkingConfig] "${modelName}" has no verified Gemini thinkingBudget range; ` +
        "omitting thinkingConfig — requested thinkingLevel is silently disabled.",
    );
    return undefined;
```

This keeps the intent while making a silent behavior change observable.

src/lib/utils/thinkingConfig.ts
```diff
@@ -85,33 +89,85 @@ export function createThinkingConfigFromRecord(
...
+  medium: 0.55,
+  high: 1,
+};
+
+/**
+ * Maps a qualitative ThinkingLevel onto a numeric thinkingBudget within a
+ * model's verified [min, max] range (see `getGemini25ThinkingBudgetRange`).
+ * Exported for deterministic testing of the level->budget mapping.
+ */
+export function mapThinkingLevelToBudget(
```
Tara-ag, line 110: `<!-- yama:gemini-budget-mapping-untested -->`

**MINOR — exported "for deterministic testing" but the mapping is only covered indirectly.**

The JSDoc exports `mapThinkingLevelToBudget` for "deterministic testing of the level→budget mapping", but neither new suite actually unit-tests it: both are live end-to-end suites that only exercise the mapping *through* a live provider call (`reasoningTokens` ordering between `minimal`/`high`). That means the `THINKING_LEVEL_FRACTIONS` arithmetic (the `0/0.25/0.55/1` split and `Math.round`) is only verified when live Gemini credentials are present.

A deterministic test of the pure function (flashing the exact budget per level/range, e.g. `mapThinkingLevelToBudget("minimal",{min:128,max:24576}) === 128`) would pin the mapping without network/latency and flakiness. If repo Rule 15 (e2e-only suites) makes a standalone unit test awkward, consider asserting the mapping from a built-`dist` import within one of the existing suites so the level→budget contract is captured without live network.

Tara-ag (approved):
Two minor findings (silent thinking-drop on unrecognized Gemini models; level→budget mapping only indirectly covered). The core fix is correct and well-verified — approving.

Tara-ag:
## Verdict: APPROVE

Recurring review of `fix/gemini-thinking-level` at HEAD `b7b0fc69`. The PR translates the qualitative `thinkingLevel` into a vendor-verified numeric `thinkingBudget` for Gemini 2.5 on every native-SDK call site (Vertex `executeNativeGemini3Generate/Stream` **and** the shared Google AI Studio `buildNativeConfig`), omits `thinkingLevel` from the wire for 2.5, and preserves the Gemini 3 passthrough unchanged. Both prior MINOR findings are **resolved in the current head** — verification below.

### Findings — final disposition

| Severity | File:Line | Finding | Status |
|---|---|---|---|
| MINOR | `src/lib/utils/thinkingConfig.ts:170-181` | Silent `thinkingConfig` drop for non-Gemini-3/2.5 models | **Fixed** — author added the `warnedNoBudgetRangeModels` Set + one-time `logger.warn` in the `return undefined` branch, surfacing the silent no-op once per model. Thread resolved. |
| MINOR | `src/lib/utils/thinkingConfig.ts:109` | `mapThinkingLevelToBudget` only exercised indirectly through live e2e suites | **Fixed** — author added a deterministic test in `test/continuous-test-suite-gemini-thinking-level.ts` (pins floor/ceiling + strict fractional ordering via a built-`dist` import), with the Rule 15 determinism exception declared in the file header and ESLint allow list. Thread resolved. |

### Confirmed clean at HEAD (independent, this pass)
- **Blast radius** (code graph): `createNativeThinkingConfig` has exactly 3 callers — `buildNativeConfig`, `GoogleVertexProvider.executeNativeGemini3Stream`, `GoogleVertexProvider.executeNativeGemini3Generate` — all pass `modelName`; optional param ⇒ backward compatible (Rule 5). `getGemini25ThinkingBudgetRange` is called only from `createNativeThinkingConfig`; its fallback never leaks to non-2.5 models.
- **AI Studio wiring**: `buildNativeConfig` has exactly 2 callers (both in `googleAiStudio/client.ts`); the Generate path passes `options.model`, Stream mirrors it — so the 2.5 translation genuinely fires on the AI Studio path (the audit gap this PR closed).
- **Regex/ordering**: `flash-lite` precedes `flash` in `GEMINI_25_THINKING_BUDGET_RANGES`; the fallback range is the intersection of the verified ranges.
- **Mapping arithmetic**: `minimal=0→min`, `high=1→max`, `low`/`medium` split the middle; all four levels distinct and in-range; `isGemini3Model` returns false for all `gemini-2.5-*` so the 2.5 branch is reached.
- **Rule 15**: both suites e2e + one `dist/` import; determinism exception declared in header and allow list; no one-module-graph mixing.
- No secrets / no payload logging / no injection surface.

### Not raised (CI owns it)
Formatting/lint/type errors; naming/prose taste; performance micro-costs.

No remaining findings — the change is sound and well tested.

src/lib/utils/thinkingConfig.ts
Tara-ag: The suggestion block above is malformed — it is anchored to line 165 (the closing `}` of the `if` block) but the block content omits that closing brace, so applying it would produce uncompilable code. Corrected version below, replacing both the `return undefined;` (line 164) and the closing `}` (line 165):

```suggestion
    // Not a Gemini 3 model and not a Gemini 2.5 model we have a verified
    // budget range for: no evidence this model accepts either parameter, so
    // omit thinkingConfig rather than guess a shape that might itself 400.
    logger.warn(
      `[thinkingConfig] "${modelName}" has no verified Gemini thinkingBudget range; ` +
        "omitting thinkingConfig — requested thinkingLevel is silently disabled.",
    );
    return undefined;
  }
```

Same intent as the original — surface the otherwise-silent no-op as a one-time `logger.warn` — but now the diff stays valid (`if (modelName && !isGemini3Model(modelName))` remains properly closed). The original block should be disregarded in favor of this one.

Tara-ag: Recurring review pass (`884f849` — no author reply on this thread). Finding still stands: `gemini-2.0-flash` (and any non-Gemini-3, non-2.5 Gemini) returns `undefined` from the new `if (modelName && !isGemini3Model(...))` branch, so an opted-in `thinkingLevel` request now silently produces no thinking instead of the old provider 400. The suggestion is the correct fix — a one-time `logger.warn` in the `return undefined` branch — and the prior corrected suggestion block (r4029783124) applies cleanly. Keep as MINOR.

src/lib/utils/thinkingConfig.ts
Tara-ag: Recurring review pass (`884f849` — no author reply on this thread). Finding still stands: `mapThinkingLevelToBudget` and the `THINKING_LEVEL_FRACTIONS` arithmetic are only exercised indirectly through the two live e2e suites. A deterministic assertion of the pure function (built-`dist` import within one of the existing suites, per Rule 15) would pin the level→budget contract without live credentials/flakiness. Keep as MINOR.

Tara-ag (approved):
Approving the current state of `fix/gemini-thinking-level` (`884f849`). The thinkingLevel→thinkingBudget translation for Gemini 2.5 is correct and well-verified, the previously-missed AI Studio native path is now wired, and both live e2e suites follow repo conventions. Two MINOR, non-blocking findings remain open on `thinkingConfig.ts` (silent thinking-drop for unrecognized models; level→budget arithmetic only indirectly tested) — neither blocks merge.

src/lib/utils/thinkingConfig.ts
murdore: Agreed — "provider rejects the field" becoming "silently no thinking" is strictly worse for the caller, because it is indistinguishable from thinking being off. Applied in `b7b0fc69f`, with one deviation from your snippet: the warning is deduped per model name via a module-level `Set`, so an agentic loop generating repeatedly against the same model logs it once rather than once per call.

```ts
if (!warnedNoBudgetRangeModels.has(modelName)) {
  warnedNoBudgetRangeModels.add(modelName);
  logger.warn(
    `[thinkingConfig] "${modelName}" has no verified Gemini thinkingBudget range; ` +
      "omitting thinkingConfig — requested thinkingLevel is silently disabled.",
  );
}
```

The omit-rather-than-guess behaviour is unchanged; only its observability is.

src/lib/utils/thinkingConfig.ts
murdore: Applied in `b7b0fc69f`, taking the route you suggested for exactly the Rule 15 reason you anticipated — asserted from the built `dist`, inside the existing suite, rather than as a standalone unit test:

```ts
import { mapThinkingLevelToBudget } from "../dist/utils/thinkingConfig.js";
```

The case pins the fractions on a synthetic `{min:128,max:24576}` range and needs no credentials and no network. It asserts a PRECONDITION first — that `minimal`/`high` land exactly on the range's floor and ceiling — so the ordering assertion cannot pass against a function that short-circuited to a constant.

Verified against the built dist: `minimal=128, low=6240, medium=13574, high=24576`; precondition true, strictly increasing true.

The determinism exception is documented in the suite's own header and in the `neurolink/e2e-tests-only` allow list entry, as that rule's doc block requires.

src/lib/utils/thinkingConfig.ts
Tara-ag: Resolved in head `b7b0fc69`: the author implemented exactly the suggested fix — a `warnedNoBudgetRangeModels` Set plus a one-time `logger.warn` in the `return undefined` branch (`src/lib/utils/thinkingConfig.ts:170-181`). The corrected suggestion block (r4029783124) is the authoritative form and now matches the landed code. Retracting the earlier "still stands" note — no further action needed.

src/lib/utils/thinkingConfig.ts
Tara-ag: Resolved in head `b7b0fc69`: the author added the deterministic test — `test/continuous-test-suite-gemini-thinking-level.ts` asserts `mapThinkingLevelToBudget` pins the range floor/ceiling and stays strictly ordered across minimal/low/medium/high, via a built-`dist` import, with the Rule 15 determinism exception declared in the file header and ESLint allow list. Retracting the earlier "still stands" note — no further action needed.

src/lib/types/config.ts
```diff
@@ -602,8 +602,18 @@ export type CreateThinkingConfigOptions = {
...
 };
 
-/** Native SDK thinkingConfig structure for Gemini native SDK. */
+/**
+ * Native SDK thinkingConfig structure for Gemini native SDK.
+ *
+ * `thinkingLevel` and `thinkingBudget` are mutually exclusive on the wire:
+ * Gemini 3 models accept `thinkingLevel`; Gemini 2.5 (and earlier) models
+ * reject it outright with INVALID_ARGUMENT "thinking_level not supported by
+ * this model" and require the numeric `thinkingBudget` instead. Both fields
```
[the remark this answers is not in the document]
murdore: Right — the doc overstated the helper's reach.

`createNativeThinkingConfig` calls `getGemini25ThinkingBudgetRange` for every non-Gemini-3 model, but that function returns a range **only** for Gemini 2.5. For anything older it falls through to the one-time WARN this PR added and returns `undefined`, so no `thinkingConfig` and no `thinkingBudget` reach the wire at all. "Gemini 2.5 (and earlier) … require the numeric `thinkingBudget` instead" reads as though earlier models get one.

The `NativeThinkingConfig` docblock now says Gemini 2.5 is the only family a budget is emitted for, and that earlier models get no `thinkingConfig` plus a WARN rather than a guessed shape that might itself 400.

Fixed in **https://github.com/juspay/neurolink/pull/1719**. `docs/api/type-aliases/NativeThinkingConfig.md` is regenerated with it, for the `docs:api` currency gate.

test/continuous-test-suite-gemini-thinking-level.ts
```diff
@@ -0,0 +1,257 @@
...
+  // the range's own floor/ceiling before the ordering assertion below means
+  // anything.
+  assert(
+    minimal === range.min && high === range.max,
+    "precondition failed: minimal/high did not pin the range's floor/ceiling",
+  );
+  assert(
+    minimal < low && low < medium && medium < high,
+    "thinkingLevel fractions did not produce a strictly increasing budget across minimal/low/medium/high",
+  );
```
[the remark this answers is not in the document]
murdore: Agreed — the test name promised a pin and the assertion delivered an ordering, which a changed fraction or a changed rounding rule would both survive.

Taken, with the precondition kept in front of it rather than replaced:

```ts
assert(
  minimal === range.min && high === range.max,
  "precondition failed: minimal/high did not pin the range's floor/ceiling",
);
assert(
  minimal === 128 && low === 6240 && medium === 13574 && high === 24576,
  "thinkingLevel fractions no longer map to the pinned minimal/low/medium/high budgets",
);
```

Literal integers rather than a recomputation of `Math.round(range.min + span * f)`: recomputing the formula in the test would follow the implementation if the *rounding* changed, which is one of the two things this is meant to catch. These four are what `0 / 0.25 / 0.55 / 1` produce over `[128, 24576]`, verified against the built `dist`.

Fixed in **https://github.com/juspay/neurolink/pull/1719** — suite re-run: 6 passed, 0 failed.

test/continuous-test-suite-gemini-thinking-level.ts
```diff
@@ -0,0 +1,257 @@
...
+    typeof minimalResult.reasoningTokens === "number" &&
+      typeof highResult.reasoningTokens === "number",
+    "precondition failed: reasoningTokens was not a number on one of the two flash-lite turns",
+  );
+
+  assert(
+    (minimalResult.reasoningTokens ?? Infinity) <
+      (highResult.reasoningTokens ?? -1),
+    "flash-lite thinkingLevel minimal did not use a smaller budget than high",
+  );
```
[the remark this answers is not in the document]
murdore: Taken — this is the stronger of the three, and the citation is right: `thinkingBudget` is a **soft** upper limit, so live `reasoningTokens` is not monotonic in the requested budget and `minimal < high` can fail on a turn whose emitted budgets were entirely correct. That is a flaky assertion dressed as a contract.

Both comparisons are gone — here and at `continuous-test-suite-gemini-thinking-level-ai-studio.ts:169-173`. The deterministic mapping test stays as the shared regression check, and it is now an exact pin rather than an ordering (see the other thread).

One addition to your suggestion: both tests are **renamed**. They were called `gemini-2.5-flash-lite:minimal uses a smaller budget than :high` and the AI Studio equivalent — names that describe precisely the assertion being removed. Leaving the names would have left the same claim in the report output. They are now `… accepts both thinkingLevel extremes and thinks on high`, which is what the live turn actually proves, and what remains asserted is that the high turn was accepted and reported reasoning tokens.

PR #1709, opened 2026-09-17
fix(anthropic): surface structured output the output ceiling cut short

murdore:
The native generate path answers a `schema` request by appending a synthetic `final_result` tool and unwrapping its arguments back into text. That makes the vendor report `stop_reason: "tool_use"` on a turn where no tool call is surfaced, so the path substitutes a `"stop"` finish to stop the turn misreading as step-capped.

**The substitution applied to every stop reason, not just the one it was written for.** A turn can also end at `stop_reason: "max_tokens"` — the model was still writing the `final_result` arguments when the output ceiling hit, so the structured payload is cut mid-object. Reporting that as `"stop"` told the caller the response was complete: `neurolink.ts`'s `finishReason === "length"` check never fired, `jsonTruncated` stayed unset, and the truncation warning was never logged.

## Why this was silent from both directions

Whatever fragment arrived is *itself valid JSON*, so `coerceJsonToSchema` parses it and never sets its own `truncated` flag. The finish reason was the only remaining signal that the object was a fragment — and it had been overwritten. A caller following the documented contract (check `jsonTruncated` before trusting `structuredData`, per CLAUDE.md rule 3) was handed a half-written object as a whole one.

## Fix

Narrow the substitution to the reason it exists for: swap in `"stop"` only when the mapped reason would be `"tool-calls"`, and otherwise report what the vendor said.

This is the rule the **streaming path already applies** — it overrides only `finishReason === "tool-calls"` — so the two paths now agree instead of diverging. That divergence is what made this reachable on generate and not on stream.

## Coverage

`test:anthropic-silent-truncation` serves the exact non-streaming Messages API wire shape over `ANTHROPIC_BASE_URL`, so the real provider, the real SDK and the real unwrapping code all run and only the vendor is substituted. No credentials, no network egress, deterministic in CI — the same local-stand-in pattern as `test:anthropic-streaming-retry` and `test:native-vendor-recovery`.

The determinism exception applies (CLAUDE.md rule 15): no reachable live endpoint produces "a `final_result` call cut off at the output ceiling" on demand, since it needs model, prompt and cap to land just so.

**Red on the unfixed build was `Passed: 1 / Failed: 2`.** The completed-turn control passing in that same red run is what shows the fixture and routing are sound and the two failures are the defect, not a broken harness. **Green is 3/3.**

Gates: `check` 0 errors · `lint` 0 errors · `build` exit 0 · `providers-mocked` 95 passed / 0 failed · `provider-structure` 3/3 · `anthropic-streaming-retry` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected completion status reporting for Anthropic structured outputs that are cut off by the token limit.
  * Truncated JSON responses now report a length-based finish reason and are flagged as truncated, while preserving any recoverable content.
  * Completed tool-use responses continue to report a normal stop status.

* **Tests**
  * Added coverage for Anthropic structured-output truncation and completed responses.

* **Chores**
  * Added a command for running the new validation suite.


Tara-ag:
Superseded — the consolidated review summary for this PR is in [#issuecomment-5707996755](https://github.com/juspay/neurolink/pull/1709#issuecomment-5707996755). This duplicate is retired so the PR carries exactly one summary. Verdict stands: **APPROVE**.

Tara-ag:
## APPROVE

The fix narrows the `final_result` finish-reason substitution to exactly the case it exists for, restoring rule 3's "truncation must surface, never silently" contract on the native generate path.

### Findings

| --- | --- | --- |
| — | — | No blocking findings. See checked-clean list below. |

- **Correctness of the guard** (`src/lib/providers/anthropic/client.ts:1688`): `mapAnthropicStopReason` returns `"stop" | "length" | "tool-calls" | "content-filter"`, so requiring `=== "tool-calls"` only remaps the `tool_use`-without-surfaced-call case to `"stop"`. `max_tokens`→`"length"` (and `content_filter`→`"content-filter"`) now pass through honestly; no previously-honest case regresses.
- **Rule 3 contract**: verified that `neurolink.ts` keyed off `finishReason === "length"` to set `jsonTruncated` and log the warning — exactly the code path the fix (re)activates for truncated structured output.
- **Agreement with the streaming path**: the generate path now mirrors the streaming path's already-narrow substitution, removing the divergence that made this reachable on generate only.
- **Backward compatibility (rule 5)**: 2-line change inside one provider function; no public API surface touched; `raw` finish reason still carries the vendor's verbatim `stop_reason`.
- **Blast radius**: minimal — single call site in the Anthropic generate path; non-Anthropic providers and streaming untouched.
- **Tests** (rule 15): properly e2e — drives `dist/index.js` through the real SDK with only the vendor substituted via a local `ANTHROPIC_BASE_URL` stand-in; determinism exception declared in the file header; covers both the truncated (`length` + `jsonTruncated` + partial object) and completed (`tool_use` → `stop`, not truncated) cases.

No secrets or security concerns in the change.

Tara-ag (approved):
## APPROVE

The fix narrows the `final_result` finish-reason substitution to exactly the case it exists for (`mapAnthropicStopReason(stop_reason) === "tool-calls"` → `"stop"`), letting `max_tokens` truncation surface honestly as `"length"`. This re-activates the documented `jsonTruncated`/truncation-warning contract (CLAUDE.md rule 3) on the native generate path and makes the two Anthropic paths agree. No regression, no public API change, and e2e coverage exercises the real provider + SDK against a local wire-shape stand-in.

No blocking findings.

Issue #1708, opened 2026-09-16
fix(anthropic): truncation is silent when a schema and tools are combined (jsonTruncated never set)


murdore:
## Symptom

The live json-e2e case

```
anthropic:claude-sonnet-4-6 — forced truncation yields a partial object, never a string
```

fails with:

```
truncation must be flagged jsonTruncated=true (jsonTruncated=undefined, finishReason=stop)
```

## It is pre-existing, not a regression

It reproduces on untouched base branches, not only on feature work. Observed failing identically on:

- `fix/huggingface-tool-allowlist` (the base of #1658) — control run, 37 passed / 1 failed
- `fix/structured-output-with-tools` (#1658) — 51 passed / 1 failed
- both the RED and GREEN halves of an unrelated bisect on #1693, where the only file varied was `openaiChatCompletionsBase.ts`

So it is independent of the structured-output work and of the OpenAI-compatible provider path generally.

## Likely cause

The case is meant to force output truncation by capping `max_tokens`, then assert that the SDK reports it (`jsonTruncated=true`). The observed `finishReason=stop` says the model **ended on its own** rather than hitting the cap — so the precondition the assertion depends on did not hold, and the flag is correctly `undefined`.

That makes this most likely a test-fixture rot problem rather than a product defect: `claude-sonnet-4-6` now answers this prompt inside the cap the test picked. Per `CLAUDE.md`, an assertion about something happening needs a precondition proving the thing under test actually occurred — this case asserts on truncation without first asserting that truncation happened, so when the model stops early the failure is reported against the product instead of the fixture.

## Suggested fix

Assert the precondition first: fail loudly with a fixture-rot message when `finishReason !== "length"`, and separately assert `jsonTruncated` only once truncation is established. Tighten the cap (or lengthen the prompt) so the truncation is actually forced.

Recorded so this stops being rediscovered on every unrelated run.

murdore:
A second case in the same suite has the same shape and is **intermittently** red:

```
openai:gpt-4o-mini — an optional Zod field stays optional, and is absent when unset
  the optional property must be absent, proving it was not forced into required
```

It prompts `"Name the capital of France. Do not include a population figure."` against a schema whose `population` is `z.number().optional()`, then asserts `population` is absent. Whether the model honours "do not include" is model behaviour, not SDK behaviour, so the assertion is live-model-dependent.

Observed failing once and then passing on **three consecutive re-runs against the byte-identical tree**, which makes it a flake rather than a regression.

[...]

PR #1710, opened 2026-09-17
test(json-e2e): assert the optional-field contract the SDK owns, not the model

murdore:
## What this fixes

`openai:gpt-4o-mini — an optional Zod field stays optional, and is absent when unset` was intermittently red. It failed once during an unrelated merge, then passed on **three consecutive re-runs against a byte-identical tree**.

The cause is that it asserted the wrong actor's behaviour. The contract it exists to guard is SDK-owned — an optional Zod property must not be forced into the JSON Schema's `required` list. What the assertion actually read is model-owned — whether `gpt-4o-mini` honours *"Do not include a population figure."* Those are different things, and the model sometimes volunteers the field anyway.

## The fix

The call is now attempted up to three times and the test passes as soon as **one** turn omits the property.

It stays discriminating: a property genuinely forced into `required` cannot be legally omitted on *any* attempt, so all three would carry it and the test fails with the keys it saw. What changes is only the tolerance for the model choosing to answer a question it was allowed to skip.

The invariants that really are the SDK's own are asserted on **every** attempt, not just the winning one:

- the request with an optional property is accepted, not rejected outright
- `structuredData` is a plain object
- the required property is present and non-empty
- the optional property, **when present**, matches its declared type — a `null` or a string there would mean the field was injected rather than answered, which is the same defect in disguise

## Not fixed here, deliberately

The other red case in this suite — `anthropic:claude-sonnet-4-6 — forced truncation yields a partial object, never a string` — is **not** a test bug and is left failing on purpose. Probing the exact fixture shows:

```
schema + tools        finish=stop     raw=max_tokens   jsonTruncated=undefined
schema, NO tools      finish=length   raw=max_tokens   jsonTruncated=true
NO schema, NO tools   finish=length   raw=max_tokens   jsonTruncated=undefined
```

Anthropic reports `max_tokens` in every case, but when a schema and tools are combined the unified `finishReason` collapses to `stop` and `jsonTruncated` is never set — silent truncation, which `CLAUDE.md` rule 3 explicitly forbids. That test is correctly reporting a live defect; adding a precondition to make it green would have converted a true failure into a skip. Tracked in #1708.

## Verification

`pnpm run lint` → 0 errors. `pnpm run build` → 0. Live run against OpenAI:

```
✓ openai:gpt-4o-mini — an optional Zod field stays optional, and is absent when unset
  Passed: 10   Failed: 1
```

The single remaining failure is the Anthropic truncation case above.


test/continuous-test-suite-json-e2e.ts
```diff
@@ -632,53 +632,81 @@ await test("anthropic:claude-sonnet-4-6 — forced truncation yields a partial o
...
+  // population despite being told not to.
+  //
+  // So: attempt it a few times and pass as soon as ONE turn omits the field.
+  // If the property had been forced into `required`, the model could not
+  // legally omit it on ANY attempt, so every attempt would carry it — which
+  // is exactly the failure reported below. The per-attempt invariants that
+  // ARE the SDK's own (request accepted, object returned, required property
+  // present, optional property well-typed when present) are asserted every
+  // time, not just on the winning attempt.
+  const ATTEMPTS = 3;
```
Tara-ag: **MAJOR — residual flake probability, unquantified, and the retry budget is arbitrary**

The fix reduces the flake rate but does not bound it. In the healthy (non-strict, optional) state, let `p` = P(gpt-4o-mini volunteers `population` on one attempt despite the prompt). The test fails when **all** `ATTEMPTS` (= 3) attempts carry the field, i.e. with probability `p^3`. If `p` is even 0.35–0.5, the suite still goes red 4–12% of runs — which is precisely the "intermittently red" failure mode this PR exists to kill. The `Passed: 10` verification is far too small a sample to bound `p`; a drift in model verbosity or a prompt change sends the flake rate straight back up.

The 3 is not derived from any target flake probability — it is picked by hand. Please either:

1. pick `ATTEMPTS` from an explicit, documented flake target (`p^N ≤ ε`, e.g. `ε ≈ 1e-3`) and state the assumed bound on `p` (with `ATTEMPTS = 3`, `ε = 1e-3` would require `p ≤ 0.1`, which is not obviously satisfied — bumping `ATTEMPTS` to 5–6 is cheap since attempts short-circuit on the first omit), **or**
2. if a residual flake is genuinely unavoidable here, drop the hard failure to a logged observation (like the `soft` cells) and keep the per-attempt SDK invariants as the hard assertions — so the suite cannot go red on model mood, only on a real SDK regression.

The discriminating power against the actual regression is preserved and worth keeping: if the SDK re-forces `population` into `required` under strict mode, the model cannot legally omit it on *any* attempt, so all 3 carry it and this fails deterministically. The concern is only the false-positive direction in the healthy state.

test/continuous-test-suite-json-e2e.ts
```diff
@@ -632,53 +632,81 @@ await test("anthropic:claude-sonnet-4-6 — forced truncation yields a partial o
...
+    )
+      .sort()
+      .join(",")})`,
   );
   console.log(
     "      · optional Zod field honoured and absent — request not rejected",
   );
 });
 
-// The same shared path carries the whole OpenAI-compatible family, so pin a
```
Tara-ag: **MINOR — collateral comment deletion: the rationale for the `azure` second-cell test is gone**

The 6-line comment directly above the `azure` test ("The same shared path carries the whole OpenAI-compatible family… Azure is the right second cell here: it is the other provider that opts OUT of suppressing response_format with tools, so it actually exercises the `json_schema` + strict decision…") was deleted as part of this hunk. That comment explains why azure — and not deepseek/groq/mistral — was picked as the family-mate for this openai test, and it still describes the test immediately below it. The deletion is unrelated to the openai-test change and loses useful architectural rationale.

Restore it (only the comment, not the code):

````suggestion
// The same shared path carries the whole OpenAI-compatible family, so pin a
// second member of it. Azure is the right second cell here: it is the other
// provider that opts OUT of suppressing response_format with tools, so it
// actually exercises the `json_schema` + strict decision this change makes,
// rather than a provider whose format is downgraded or suppressed before the
// decision matters.
````

**Review by Tara-ag — COMMENTED** [review]
Reviewed the precondition/retry change to `openai:gpt-4o-mini — an optional Zod field stays optional, and is absent when unset`. Since there were no prior `yama` review comments or author replies on this PR, both findings below are new.

**Comment by Tara-ag** [comment]
**APPROVE** — both findings from the prior review were adequately addressed in the latest push: the retry loop was replaced with a deterministic wire-capture assertion (residual flake now genuinely zero), and the azure rationale comment was restored verbatim. No new material findings.

**Resolved findings (previous review → current state):**

| Severity | Location | Finding | Resolution |
|---|---|---|---|
| MAJOR | `test/continuous-test-suite-json-e2e.ts:649` (`const ATTEMPTS = 3`) | Residual flake probability `p^3` with an arbitrary retry budget. | Resolved. The retry loop is gone. A module-scope pass-through `fetch` recorder captures the outbound request bodies, and the test asserts on the JSON Schema that went **on the wire**: a precondition that at least one made-out `required` list contains `capital`, then the contract that `population` appears in no emitted `required` list. The assertion no longer depends on what the model answers, so residual flake is zero. Per-attempt SDK invariants (plain object, `capital` present/non-empty, `population` typed when present) remain hard asserts. |
| MINOR | `test/continuous-test-suite-json-e2e.ts:676` (removed) | Collateral deletion of the 6-line comment justifying azure as the second OpenAI-family cell. | Resolved. The comment was restored verbatim above the azure test; author confirmed it was collateral, not intentional. |

**Checked & clean in the current push:**
- Capture flow is safe: the recoder pushes synchronously inside the wrapped fetch, so the capture happens-before the test resolves on the captured body; recording is gated off until `captureRequests` is set; the flag and body buffer are reset in a `finally`, and the body buffer is re-created per test, so stale state cannot leak across tests.
- Discriminating power preserved: if the SDK re-forces `population` into `required`, the model cannot legally omit it on any attempt, so the regression still fails deterministically.
- Precondition guards against a vacuous pass (no captured body, no assertion).
- Blast radius: self-contained **test-only** change; the modified file is not imported by production code (graph confirms the affected node is solely the openai e2e test, `test@L634-708`), so no public-API, hot-path, or backward-compat (Rule 5) impact.
- Rule-3/15 compliance (e2e-only suite, single module graph); no secrets; single commit with a conventional message; lint/build reported clean.
- The pre-existing `anthropic — forced truncation` red case (#1708) is a live defect being reported correctly, not this change.

The rest of the file (matrix cells, truncation tests, azure/cohere cells) is untouched and consistent with prior behaviour.

**Review by Tara-ag — CHANGES_REQUESTED** [review]
**NEEDS_WORK** — the retry-until-green reframe is a step in the right direction but does not bound the flake probability it aims to eliminate, and it silently drops the `azure` cell's rationale comment.

The two inline findings stand:
- **MAJOR** (`test/continuous-test-suite-json-e2e.ts:649`): residual flake probability `p^3` with an arbitrary `ATTEMPTS = 3`; no target flake probability or assumed `p` documented. Bump the budget or soften the last assertion.
- **MINOR** (`test/continuous-test-suite-json-e2e.ts:676`): the 6-line rationale for azure as the second OpenAI-family cell was deleted collaterally; restore it.

Full details and suggestions are on the inline comments above.

File: test/continuous-test-suite-json-e2e.ts (hunk 1 above)
Comment (murdore) [reply test/continuous-test-suite-json-e2e.ts]: You are right, and the `p^3` framing is the part that convinced me — a retry budget moves the flake, it does not bound it. I have dropped the retry loop entirely.

The test now asserts the contract the SDK actually owns. A tiny pass-through `fetch` recorder is installed **before** the `dist/` import, the turn's outbound request bodies are captured, and the assertion reads the JSON Schema that went **on the wire**:

- precondition first — at least one captured body must carry a `required` list containing `capital`, otherwise there was no schema to assert against and the check would be vacuous;
- then the contract — `population` must appear in **no** `required` list the SDK emitted.

Whether gpt-4o-mini volunteers the field is now *reported*, never asserted, because that was never ours to guarantee. Residual flake probability is zero: the assertion does not depend on the model's answer at all.

Proven discriminating rather than just re-run — with the defect injected into the captured body (`required:["capital"]` → `required:["capital","population"]`):

```
GREEN  ✓ optional Zod field absent from the wire schema's required list; model omitted it
       Passed: 9   Failed: 1
RED    ✗ the optional property was promoted into the schema's required list (1 of 1 schema(s) on the wire)
       Passed: 8   Failed: 2
```

Reproduced twice independently. The one remaining failure in both columns is the pre-existing `anthropic — forced truncation` case (#1708), unrelated to this change.

test/continuous-test-suite-json-e2e.ts
murdore: Confirmed and restored in the same push — thank you for catching it. My replacement span ran from the start of the openai test to the start of the azure test, which swallowed the six lines between them; it was collateral, not intentional.

The rationale is back verbatim above the azure test:

> The same shared path carries the whole OpenAI-compatible family, so pin a second member of it. Azure is the right second cell here: it is the other provider that opts OUT of suppressing response_format with tools, so it actually exercises the `json_schema` + strict decision this change makes, rather than a provider whose format is downgraded or suppressed before the decision matters.

That reasoning is exactly why azure and not deepseek/groq/mistral, and it is not reconstructable from the test body alone — losing it would have made the cell look arbitrary to the next reader.

Tara-ag (approved):
**APPROVE** — both prior findings are fully resolved in the current push: the `ATTEMPTS` retry loop was replaced with a deterministic on-wire capture assertion (residual flake now zero; the assertion no longer depends on the model's answer), and the azure rationale comment was restored verbatim. The per-attempt SDK invariants and the discriminating failure mode against the strict-`required` regression are preserved. No new material findings. See the summary comment for the full resolution table.

Tara-ag:
_(read-only retrospective — nothing posted or edited)_

PR #1713, opened 2026-09-17
fix(proxy): reconcile supervisors and absorb telemetry bursts

murdore:
## Description

A replacement updater can leave the launchd supervisor on an older package after successfully activating a current worker. The replacement sees that npm has no newer version and exits before comparing the two live generations. This change reconciles that stale supervisor through the existing quiet authenticated drain while preserving the current worker.

OTel body capture also dropped data during healthy high-volume traffic because the 32 MiB active byte pool saturated before the 64-capture slot limit. In the audited pressure window, 615 requests completed, while 107 logical captures were rejected and three were export-unconfirmed. This change keeps the 32 MiB active pool and adds a separately bounded 32 MiB/64-capture overflow queue with a 20-second admission deadline. Transient bursts wait for capacity; sustained overload still produces an explicit indexed timeout or queue-full result.

OpenObserve also retained exact retry copies from at-least-once OTLP delivery. The same window contained 616 stored request-final rows for 615 stable events, plus eight repeated capture-index events and four repeated lifecycle events. The doctor now uses `proxy.event_id` and each record type's logical identity to preserve raw counts, normalize only byte-identical transport retries, and hard-fail changed payloads or separate producer events that claim the same logical identity. Body reconstruction applies the same rule to chunk rows. Historical transient collector retry counters remain visible but no longer make every later complete interval permanently incomplete; permanent, enqueue, receiver, identity, and selected-interval delivery failures still fail or warn.

## Type of Change

- [x] Bug fix

## Changes Made

- Reconcile a stale rolling supervisor when the worker and installed package are already current.
- Skip package installation and trampoline rewrites for supervisor-only reconciliation.
- Preserve retry behavior when reconciliation health cannot be verified.
- Add bounded OTel capture admission waiting and expose wait, high-water, and timeout counters.
- Extend shutdown flushing through admission, processing, and publication deadlines.
- Query stable OTLP event IDs and normalize exact backend retry copies before coverage calculations.
- Retain raw/logical counts and exact-retry diagnostics; reject conflicting or producer duplicates.
- Reconstruct sampled bodies after identity-aware chunk normalization.
- Separate historical transient collector retries from evidence of permanent or selected-interval loss.

## Breaking Changes

- [x] No breaking changes

## New Provider Onboarding

## Testing

- [x] Production build and package validation: pass
- [x] Proxy telemetry suite: 73/73
- [x] Safe proxy restart suite: 27/27
- [x] Bugfix suite: 285/285
- [x] Strict type checks: pass
- [x] Changed-file lint: 0 errors; four pre-existing warnings outside this diff
- [x] Formatting and diff checks: pass
- [x] Read-only historical pressure replay: 615 logical finals from 616 rows; 13 exact metadata retry copies normalized; zero identity conflicts; real 107 rejected and three unconfirmed captures still fail
- [x] Read-only quiet-window replay: complete doctor status `pass`; 16/16 unique finals, 60/60 acknowledged captures, zero identity conflicts

The pressure regressions reconstruct and verify every redacted capture, require transport acknowledgement, exercise both active byte and slot bounds, and assert zero rejection, timeout, missing index, duplicate chunk, or bound breach. Duplicate regressions cover exact OTLP retries, producer duplicates, changed payloads, and chunk conflicts.

## Code Quality

- [x] Project style and formatting followed
- [x] Self-review completed
- [x] No dependency changes
- [x] No hardcoded credentials
- [x] TypeScript strict checks pass

## Deployment Notes

No configuration or migration is required for the package change. The updater will reconcile a stale supervisor on a normal future check after this version is installed. The live proxy and telemetry backend were inspected read-only and were not used as development targets.


src/lib/proxy/requestLogger.ts
```diff
@@ -169,9 +169,12 @@ function trackLogOperation<T>(operation: Promise<T>): Promise<T> {
   return operation;
 }
```
Tara-ag, line 175: **MAJOR — new default flush deadline (70 s) exceeds the launchd `ExitTimeOut` (45 s), undermining the very clean-shutdown guarantee this adds.**

The new default `3 * PROXY_BODY_CAPTURE_DEADLINE_MS + 2 * REQUEST_LOG_IO_TIMEOUT_MS` = 3×20 s + 2×5 s = **70 s**. `flushProxyOtelLogs`'s comment says shutdown now "waits through bounded OTel admission, worker processing, and publication deadlines so a clean shutdown does not abandon an accepted overflow capture."

But `proxy.ts` generates the launchd plist with `<key>ExitTimeOut</key><integer>45</integer>`, and the supervisor refresh / update path restarts the proxy via `launchctl kickstart -k`, which sends `SIGTERM` and then `SIGKILL`s the process after 45 s if it has not exited. `shutdown()` is what drains through `flushRequestLogs()`. So an overflow capture that legitimately spends ~20 s in the OTel admission queue plus up to ~20 s in worker processing can be killed at the 45 s mark — before the flush completes — which is exactly the scenario this increase is intended to protect.

Fix: keep the *internal* admission/processing deadlines but cap the overall shutdown budget well under `ExitTimeOut` (e.g. pass an explicit `timeoutMs <= 40_000` at the `shutdown()` call site rather than relying on the 70 s default), or raise `ExitTimeOut` alongside. Please verify the `shutdown()` call site uses the default and isn't already wrapped in its own tighter budget.

- `src/lib/proxy/requestLogger.ts` — default `timeoutMs = 3 * PROXY_BODY_CAPTURE_DEADLINE_MS + 2 * REQUEST_LOG_IO_TIMEOUT_MS`
- `PROXY_BODY_CAPTURE_DEADLINE_MS = 20_000`, `REQUEST_LOG_IO_TIMEOUT_MS = 5_000` (≈70 s)
- `src/cli/commands/proxy.ts` `buildProxyLaunchdPlist` → `<key>ExitTimeOut</key><integer>45</integer>`
</details>

src/cli/commands/proxy.ts
```diff
@@ -5012,17 +5015,34 @@ export const proxyGuardCommand: CommandModule<object, ProxyGuardArgs> = {
...
         );
 
+        let supervisorRefreshOnly = false;
         if (!result.updateAvailable) {
-          updateRetryAttempts = 0;
-          updateRetryVersion = null;
-          persistUpdaterState("clear update deferral", () =>
-            clearUpdateDeferral(),
+          const activeVersions = rollingSupervisor
+            ? await fetchProxyRuntimeVersions(host, port)
```
Tara-ag: **MAJOR — the new `fetchProxyRuntimeVersions` call on the no-update path is unguarded: a transient fetch failure records a spurious update failure on the idle updater check.**

Before this change the `!result.updateAvailable` branch was pure-local (clear deferral + return). Now, once per periodic check, it does a blocking HTTP round-trip to the proxy runtime endpoint. Nothing around this `await` catches a fetch rejection; a throw falls through to the outer `catch (err)` in `runUpdateCheck`, which runs

```js
persistUpdaterState("record update failure", () =>
  recordUpdateFailure(updateVersion, "check", message),
);
```

and logs `[updater] update check error: …`. So a merely-unreachable proxy during an *idle* check now emits an update-failure alert (and can contribute to retry/suppression accounting) even though no update was pending — a behavioral regression on the quiet path.

Fix: guard the fetch so a failure degrades to the existing no-*update* behavior:

```suggestion
          const activeVersions = rollingSupervisor
            ? await fetchProxyRuntimeVersions(host, port).catch(() => ({}))
            : {};
```

and let `shouldRefreshStaleSupervisor` (which already requires both live versions to be present) simply return false on an empty result.

src/lib/proxy/bodyCaptureWorker.ts
```diff
@@ -35,6 +49,112 @@ const pending = new Map<
...
+        Math.max(1, retryAfter - Date.now()),
+      );
+      admissionRetryTimer.unref?.();
+    }
+    return;
+  }
+  while (otelAdmissionWaiters.length) {
+    const waiter = otelAdmissionWaiters[0];
+    if (!reserveCapture(waiter.bytes)) {
+      break;
```
Tara-ag, line 101: **MINOR — FIFO drain head-of-line blocks smaller OTel captures behind a large one.**

`drainOtelAdmissionWaiters` dequeues strictly from `otelAdmissionWaiters[0]` and `break`s as soon as the front waiter's `reserveCapture` fails. The active pool is byte-bounded (32 MiB) as well as slot-bounded (64). If the front waiter needs more bytes than are currently free, the whole loop stops, so every smaller waiter behind it — even ones that would fit the freed capacity together — must wait until the head either fits or reaches its 20 s admission timeout.

This never loses a capture (the timeout still bounds it), but under the mixed-size bursts this code is specifically designed for (e.g. 3 MiB bodies mixed with small ones) it can markedly delay smaller captures. Consider scanning for the first compatible waiter instead of breaking on the head:

```suggestion
  for (let i = 0; i < otelAdmissionWaiters.length; i++) {
    const waiter = otelAdmissionWaiters[i];
    if (!reserveCapture(waiter.bytes)) {
      continue;
    }
    otelAdmissionWaiters.splice(i, 1);
    i -= 1;
    clearTimeout(waiter.timer);
    snapshot.waiting -= 1;
    snapshot.waitingBytes -= waiter.bytes;
    waiter.resolve("admitted");
  }
```

(or, if strict FIFO fairness is intended, leave as-is and note it).

src/lib/proxy/updateCoordinator.ts
```diff
@@ -4,6 +4,28 @@ import type {
...
+  updateAvailable: boolean;
+  rollingSupervisor: boolean;
+  runningVersion: string;
+  workerVersion?: string;
+  supervisorVersion?: string;
+}): boolean {
+  return (
+    !options.updateAvailable &&
+    options.rollingSupervisor &&
+    options.workerVersion === options.runningVersion &&
```
Tara-ag, line 23: **MINOR — reconcile silently no-ops when `workerVersion` is absent, leaving the exact stale-supervisor bug unhandled.**

The guard requires `options.workerVersion === options.runningVersion`. If the runtime-version response is partial (e.g. the runtime endpoint does not (yet) report `workerVersion` but does report a `supervisorVersion` that is stale), `undefined === runningVersion` is `false` and reconciliation never triggers — with no log line to say why, since the caller only logs inside the `supervisorRefreshOnly === true` branch.

Given the premise of this PR is exactly "worker is current but supervisor is stale", a missing `workerVersion` should not silently defeat the fix. Consider treating "no `workerVersion` present" the same as UNABLE_TO_CONFIRM (skip reconcile but log it) rather than silently failing, or at minimum emit a debug line when the fetch produced a stale `supervisorVersion` but no confirmable `workerVersion`. The added test suite covers the present-workerVersion cases but not the partial-fetch shape.

Tara-ag (commented):
Reviewed the supervisor-reconcile and OTel-burst-impact changes. Findings are minor-to-major: the core reconcile mechanism (restart via the stable trampoline) and the OTel overflow admission/accounting are sound and well-tested; the concerns are around the idle-path fetch robustness, the head-of-line drain behavior, the reconcile's silent partial-fetch gap, and the flush-deadline vs launchd ExitTimeOut interaction.

Tara-ag:
<sup>Superseded — see the canonical summary comment on this PR for the authoritative findings. This earlier comment was posted in a prior review attempt and is retained only for history; it is not the PR summary.</sup>

src/lib/proxy/requestLogger.ts
```diff
@@ -169,9 +169,12 @@ function trackLogOperation<T>(operation: Promise<T>): Promise<T> {
...
 
-/** Wait, up to a bounded deadline, for admitted request/body writes to settle. */
+/**
+ * Wait through bounded OTel admission, worker processing, and publication
+ * deadlines so a clean shutdown does not abandon an accepted overflow capture.
+ */
 export async function flushRequestLogs(
-  timeoutMs: number = PROXY_BODY_CAPTURE_DEADLINE_MS +
+  timeoutMs: number = 3 * PROXY_BODY_CAPTURE_DEADLINE_MS +
```
Tara-ag: **MAJOR** — the new default flush deadline (70 s) can exceed the launchd `ExitTimeOut` (45 s) in the exact clean-shutdown path this PR guarantees.

`flushRequestLogs` now defaults to `3 * PROXY_BODY_CAPTURE_DEADLINE_MS + 2 * REQUEST_LOG_IO_TIMEOUT_MS` = `3×20 s + 2×5 s` = **70 s**. This is the completion fence for `shutdown()`, and this PR's whole point is that a clean shutdown must not abandon an accepted OTel overflow capture. But a supervisor refresh/reconcile restart is driven by launchd, whose `ExitTimeOut` here is **45 s** — so a capture admitted near the drain start can be SIGKILLed by launchd up to ~25 s before the 70 s timeout actually fires, and the capture is dropped anyway.

I could not fully read the `shutdown()` call-path budget in this checkout, so please confirm: if launchd kills at 45 s, cap the default to fit that window (e.g. `EXIT_TIMEOUT_MS - startupMargin`), or bound the retry path so an accepted OTel capture never exceeds the hard exit budget.

```suggestion
  timeoutMs: number = PROXY_BODY_CAPTURE_DEADLINE_MS +
    2 * REQUEST_LOG_IO_TIMEOUT_MS,
```

src/cli/commands/proxy.ts
Tara-ag: **MAJOR** — the unguarded `await fetchProxyRuntimeVersions` on the idle no-update path throws into the outer catch and records a spurious update failure.

On the "no update available" idle path you now unconditionally `await fetchProxyRuntimeVersions(host, port)` when `rollingSupervisor` is set. This is a fresh network/HTTP call on a routine periodic check. If the supervisor socket is briefly unavailable (restarting, backoff, flaky), the throw propagates to the enclosing `try/catch (err)` which treats it as a reconcile failure and calls `recordUpdateFailure(...)`. But this branch wasn't updating anything — it's a benign version probe — so a transient probe error now pollutes the update-failure counters and alerting that are meant for real install/health failures.

Suggest isolating the probe so a failed reconciliation attempt cannot be mistaken for a failed update:

```suggestion
          const activeVersions = rollingSupervisor
            ? await fetchProxyRuntimeVersions(host, port).catch(() => ({}))
            : {};
```

src/lib/proxy/bodyCaptureWorker.ts
```diff
@@ -35,6 +49,112 @@ const pending = new Map<
...
+          drainOtelAdmissionWaiters();
+        },
+        Math.max(1, retryAfter - Date.now()),
+      );
+      admissionRetryTimer.unref?.();
+    }
+    return;
+  }
+  while (otelAdmissionWaiters.length) {
+    const waiter = otelAdmissionWaiters[0];
```
Tara-ag: **MINOR** — FIFO drain has a head-of-line problem: a large front waiter blocks smaller OTel captures behind it.

`drainOtelAdmissionWaiters` always admits `otelAdmissionWaiters[0]` and stops at the first waiter that fails `reserveCapture(waiter.bytes)`. Under a mixed-size burst, a big capture parked at the front can deny the small captures behind it until the big one's 20 s admission timeout, even though those small ones would fit in the available pool right now. No data is lost, but delivery latency for the small captures is coupled to the largest queued waiter.

Since each waiter's byte demand is known, scan for the first waiter that fits (best-effort, still bounded by one pass) instead of only the head:

```suggestion
  for (let i = 0; i < otelAdmissionWaiters.length; ) {
    const waiter = otelAdmissionWaiters[i];
    if (!reserveCapture(waiter.bytes)) {
      i += 1;
      continue;
    }
    otelAdmissionWaiters.splice(i, 1);
    clearTimeout(waiter.timer);
    snapshot.waiting -= 1;
    snapshot.waitingBytes -= waiter.bytes;
    waiter.resolve("admitted");
  }
```

src/lib/proxy/updateCoordinator.ts
```diff
@@ -4,6 +4,28 @@ import type {
...
+    typeof options.supervisorVersion === "string" &&
+    options.supervisorVersion !== options.runningVersion
+  );
+}
```
Tara-ag, line 27: **MINOR** — a partial runtime response (missing `supervisorVersion`) silently no-ops the reconcile, leaving the exact stale-supervisor bug unhandled.

`shouldRefreshStaleSupervisor` requires `typeof options.supervisorVersion === "string"`. `fetchProxyRuntimeVersions` returns whatever fields the runtime reported; if `supervisorVersion` is absent from a partial/degraded response, the predicate returns `false`, and the `proxy.ts:5019` no-update branch then falls through and clears the update deferral as if nothing were amiss. The stale parent keeps running and this fix silently doesn't fire, with no signal.

This is a judgment call, not a cut-and-dried code fix: decide whether a missing `supervisorVersion` should be treated as degraded-unknown (logged and retried) rather than a clean "worker current; nothing to do". If `undefined` supervisor is a realistic degraded state, at least emit an `[updater]` warning in the `proxy.ts` reconcile branch instead of printing the current happy-path "reconciling stale supervisor" log and proceeding to clear the deferral. No suggestion block because the right behavior depends on how the runtime actually reports a partial response.

Tara-ag (changes requested):
Supervisor-reconcile mechanism (restart via the stable trampoline) and the OTel overflow admission/accounting are sound and well-tested; the findings are around the idle-path probe robustness, the flush-deadline vs launchd exit budget interaction, and two silent-degradation gaps. Details in the inline comments.

Tara-ag:
## APPROVE

Re-review of the current squashed head `595b34ca` confirms all four findings from the prior pass are fixed and correctly gate their behavior, with matching regressions and green suites. No new findings this pass.

### Prior findings → resolution (verified in the current diff at head `595b34ca`)

| Severity | Location | Finding | Current-diff status |
|----------|----------|---------|---------------------|
| MAJOR | `requestLogger.ts` (`REQUEST_LOG_FLUSH_TIMEOUT_MS`) | New flush default (70 s) exceeded launchd `ExitTimeOut` (45 s) | **Resolved.** Default is now `PROXY_BODY_CAPTURE_DEADLINE_MS + 2 * REQUEST_LOG_IO_TIMEOUT_MS` = **30 s** (fits with 15 s margin); comment and `defaultFlushTimeoutMs === 30_000` test assert it. |
| MAJOR | `proxy.ts` no-update path | Unguarded `fetchProxyRuntimeVersions` recorded spurious update failures on the idle check | **Resolved.** Probe is `.catch(() => ({}))`; an incomplete/missing version logs, schedules the bounded retry, and returns before `shouldRefreshStaleSupervisor` / deferral clearing. Source-shape guard in `continuous-test-suite-bugfixes` asserts `.catch(() => ({}))` + incomplete-probe ordering. |
| MINOR | `bodyCaptureWorker.ts` `drainOtelAdmissionWaiters` | FIFO head-of-line blocked small captures behind a byte-blocked large waiter | **Resolved.** The drain now scans the bounded queue (`index += 1; continue` on byte pressure) and admits every compatible waiter; the mixed-size regression proves small-capture bypass. |
| MINOR | `updateCoordinator.ts` reconcile | Partial runtime response silently no-oped the reconcile | **Resolved.** A probe lacking either `workerVersion`/`supervisorVersion` is treated as degraded/unknown: logged, retried, deferral not cleared — before `shouldRefreshStaleSupervisor`. |

### Verified clean this pass
- **Reconcile core** — `shouldRefreshStaleSupervisor` (sole production caller `runUpdateCheck` at `proxy.ts:5038`) correctly gates on `!updateAvailable && rollingSupervisor && workerVersion === runningVersion && supervisorVersion is string && !== runningVersion`; the `supervisorRefreshOnly` guards skip install/trampoline/rollback, and the health-unavailable path schedules a retry instead of abandon/suppress.
- **OTel admission accounting** — `reserveCapture`/`releaseCapture` are balanced across start-failed, publication-settled, and clone-failed paths; waiter `waiting`/`waitingBytes` transfer cleanly into `pending` on admission and are decremented on timeout/queue-full. `drainOtelAdmissionWaiters` self-corrects when `retryAfter` is pushed out.
- **Telemetry doctor** — `normalizeMetadataRows` / `normalizeBodyChunkRows` collapse only byte-identical (`proxy_event_id` + `_timestamp` + body) retries, hard-fail on events reused with changed payloads or producer `requestId`/`captureId` duplicates, and reconstruct bodies after identity-aware chunk normalization. `collector_delivery` remains fail-safe: only `error_permanent="false"` send-failures are de-prioritized; enqueue/receiver/permanent/invalid counters still warn, and SQL interpolation uses validated `captureId`/`stream` (no injection).
- **Blast radius** — `captureProxyBody` → single caller `logBodyCapture` (`requestLogger.ts:788`); `flushRequestLogs` default only used by `shutdown` (`proxy.ts:3242`) + tests; `shouldRefreshStaleSupervisor` → `runUpdateCheck` + its test. Rule 5 OK: only additive type changes (`ProxyBodyCapture` now intersects `ProxyBodyCaptureWaitingSnapshot`; `ProxyTelemetryStoredRecord` adds optional fields).
- **Security** — no new secrets; logging stays behind the existing redaction path.

No new inline findings to post. All four threads stay resolved; nothing outstanding.

## What was checked and found clean (so silence ≠ absence)
- OTel burst admission math (byte + slot + 20 s deadline), queue-full/timeout accounting, and the head-of-line bypass.
- Reconcile decision matrix and the health-unavailable vs. abandon/suppress divergence.
- Doctor normalization of exact retries vs. producer duplicates vs. identity conflicts, plus body-chunk reconstruction.
- Collector transient-retry handling (fail-safe direction).
- No SDK surface break, no secrets, no injection.

src/lib/proxy/requestLogger.ts
Tara-ag: **Correction — malformed suggestion block.** The suggestion in the parent comment anchors a single line (177) but carries two lines, so one-click apply would duplicate the `2 * REQUEST_LOG_IO_TIMEOUT_MS,` line. The intended one-line fix (drop the `3 *`) is:

```suggestion
  timeoutMs: number = PROXY_BODY_CAPTURE_DEADLINE_MS +
```

src/cli/commands/proxy.ts
Tara-ag: **Correction — malformed suggestion block.** The suggestion in the parent comment anchors a single line (5021) but carries three lines, so one-click apply would duplicate the surrounding `const activeVersions = rollingSupervisor` and `: {}` lines. The correct one-line fix (add the `.catch`) is:

```suggestion
            ? await fetchProxyRuntimeVersions(host, port).catch(() => ({}))
```

src/lib/proxy/bodyCaptureWorker.ts
Tara-ag: **Correction — malformed suggestion block, now prose-only.** The suggestion in the parent comment anchors a single line (100) but contains a full loop, so one-click apply would corrupt the surrounding `while` block. Applied correctly, the fix replaces the whole `while (otelAdmissionWaiters.length) { … }` loop in `drainOtelAdmissionWaiters` with a scan-for-first-fitting-waiter loop:

```ts
for (let i = 0; i < otelAdmissionWaiters.length; ) {
  const waiter = otelAdmissionWaiters[i];
  if (!reserveCapture(waiter.bytes)) {
    i += 1;
    continue;
  }
  otelAdmissionWaiters.splice(i, 1);
  clearTimeout(waiter.timer);
  snapshot.waiting -= 1;
  snapshot.waitingBytes -= waiter.bytes;
  waiter.resolve("admitted");
}
```

src/lib/proxy/requestLogger.ts
murdore: Resolved in `3a85339b`. The default is now 30 seconds:

- 20 seconds for the body-capture deadline
- two 5-second request-log I/O budgets
- 15 seconds left inside launchd's 45-second `ExitTimeOut` for the surrounding listener/exporter shutdown

The telemetry regression now asserts the 30,000 ms default while still proving a delayed capture beyond the old five-second budget is flushed. Validation: telemetry 73/73, restart 27/27, production build and strict checks passed.

src/cli/commands/proxy.ts
murdore: Resolved in `3a85339b`. The idle-path runtime probe is caught locally. An exception or either missing version now logs an explicit incomplete-probe warning, schedules the existing bounded updater retry, and returns before stale-supervisor evaluation or deferral clearing. It therefore never reaches the outer update-failure handler.

The bugfix regression verifies the catch, incomplete-probe branch, retry scheduling, and ordering before the reconciliation decision. Validation: bugfixes 285/285 and restart 27/27.

src/lib/proxy/bodyCaptureWorker.ts
murdore: Resolved in `3a85339b`. The drain now scans the bounded waiter queue once and admits every capture that fits, retaining blocked captures in place with their original deadlines.

The new mixed-size regression fills all 64 active slots near the 32 MiB byte limit, queues a byte-blocked large capture followed by a small capture, releases one slot, and proves the small capture is admitted while the large capture remains queued. It then verifies zero admission timeouts, rejections, or leaked reservations. Telemetry passes 73/73.

src/lib/proxy/updateCoordinator.ts
murdore: Resolved in `3a85339b`. A partial runtime response is now treated as degraded/unknown. If either `workerVersion` or `supervisorVersion` is absent, the updater logs both observed values, schedules the existing bounded retry, and returns without clearing deferral state or claiming a clean no-op.

The regression also verifies this branch occurs before `shouldRefreshStaleSupervisor`. Validation: bugfixes 285/285 and restart 27/27.

Tara-ag (approved):
All four findings from the prior review are resolved in `3a85339b` and verified in the current diff:

- **MAJOR** flush deadline: now 30 s (`REQUEST_LOG_FLUSH_TIMEOUT_MS`), fits inside launchd's 45 s `ExitTimeOut`; telemetry asserts the 30 000 ms default.
- **MAJOR** idle-path probe: `.catch(() => ({}))` + incomplete-probe warning/retry keeps a flaky probe from polluting the update-failure path.
- **MINOR** OTel drain head-of-line: `drainOtelAdmissionWaiters` scans the bounded queue and admits every compatible waiter; mixed-size regression added.
- **MINOR** reconcile silent partial no-op: missing versions are treated as degraded/unknown (logged + retried, deferral preserved).

The supervisor-reconcile and OTel admission/accounting paths are sound, the accounting is balanced across all publication/start/clone-failure branches, blast radius is confined to `runUpdateCheck` / `logBodyCapture`, and the new regressions cover the byte-bound, slot-burst, and mixed-size cases with zero loss. No new findings this pass. Approving.

PR #1714, opened 2026-09-17
docs(providers): describe the native provider layer instead of the removed ai-sdk

murdore:
The `ai` and `@ai-sdk/*` packages are gone from this repo. Seven documentation passages had not caught up and still describe **current** behaviour as running through the Vercel AI SDK, so a reader following them is told something the source contradicts.

## Verified state of the removal

| Check | Result |
|---|---|
| `dependencies` / `devDependencies` / `peerDependencies` / `optionalDependencies` | clean |
| `pnpm.overrides` | clean |
| `pnpm-lock.yaml` | no `ai` / `@ai-sdk` resolution |
| imports in `src/`, `test/`, `scripts/`, `tools/` | **zero** |

The only two grep hits in source are comments *asserting* that nothing imports them:

```
src/lib/providers/openaiChatCompletionsBase.ts:17:  * Nothing here imports from "ai" or "@ai-sdk/*".
src/lib/providers/openaiChatCompletionsClient.ts:13: * Nothing here imports from "ai" or "@ai-sdk/*".
```

## What each passage claimed, and what the source says

| File | Claimed | Actual |
|---|---|---|
| `getting-started/providers/openai.md:15` | "integrates with OpenAI via the `@ai-sdk/openai` package" | speaks the OpenAI HTTP API directly |
| `getting-started/providers/google-vertex.md:422` | non-Gemini-3 models "continue to use the standard Vercel AI SDK path" | `googleVertex/client.ts` says the opposite twice — **all** models native: `@google/genai` for Gemini, `@anthropic-ai/vertex-sdk` for Claude |
| `features/embeddings.md:30,89` | "delegate to the Vercel AI SDK"; the SDK "handles chunking" | each provider calls its own embeddings API and does its own batching |
| `advanced/middleware-architecture.md:21` | "Built on top of Vercel AI SDK's middleware system" | built on local `wrapLanguageModel.ts`, whose header says it exists *"so the middleware factory no longer needs the ai package"* |
| `features/claude-proxy-architecture.md:388` | `jsonSchema()` "from the Vercel AI SDK" | `claudeFormat.ts:19` imports it from `../utils/tool.js` |
| `features/claude-proxy.md:1141` | "translates between all providers via Vercel AI SDK" | native provider layer |
| `features/claude-subscription-testing.md:508,520,676` | `vi.mock("@ai-sdk/anthropic")` against `src/lib/providers/anthropic.ts` | that package is not a dependency and **that file does not exist** — the provider is native at `anthropic/client.ts` |

[...]

PR #1715, opened 2026-09-17
docs(providers): drop the last current-tense Vercel AI SDK claims

murdore:
Review follow-up on the open thread from **#1714** (`docs/getting-started/providers/openai.md`).

#1714 corrected seven passages that described NeuroLink as running on the Vercel AI SDK. The reviewer pointed out that `openai.md` still contradicted its own new Overview a few hundred lines further down, and named two more files the sweep had missed. Re-sweeping the hand-written docs turned up three more of the same shape, so all six land here.

Every line below asserted that NeuroLink **currently** runs on a package that is not in `package.json`, not in `pnpm-lock.yaml`, and imported nowhere under `src/`.

| File | Claimed | Actually |
|---|---|---|
| `docs/getting-started/providers/openai.md` | streaming "uses the Vercel AI SDK `streamText` with `fullStream`" | NeuroLink's own HTTP + SSE client, `src/lib/providers/openaiChatCompletionsBase.ts` |
| `docs/features/pdf-support.md` | flow step 6 "Format as Vercel AI SDK file type" | `messageBuilder` emits the provider's own file/document part |
| `docs/reference/index.md` | AI SDK listed as "Underlying provider implementations" | a separate framework NeuroLink interoperates with, via the client SDK's `LanguageModelV1` adapter |
| `docs/getting-started/providers/google-vertex.md` | Gemini 3's native path "bypasses the Vercel AI SDK" because "the Vercel layer strips `thoughtSignature`" | present tense about a layer that no longer exists; the reason is restated as the token itself |
| `docs/features/turn-time-budget.md` | a section named after "the AI-SDK loop path" | that is the native loop |
| `docs/reference/provider-feature-compatibility.md` | Azure's rejection of the `file` content type attributed to the AI SDK | the quoted error comes from the model API |

### Deliberately left alone

These reference the Vercel AI SDK as a **separate thing**, which is accurate, and would be wrong to remove:

- `docs/features/client-sdk.md` + `docs/features/index.md` — the client SDK really does ship a `LanguageModelV1` compatibility adapter (`src/lib/client/aiSdkAdapter.ts`).
- `docs/features/observability.md` — about spans **your** application emits; the page already states NeuroLink has no dependency on `ai`.
- `docs/features/opencode-proxy-support.md` — about OpenCode, which does use it.
- `docs/guides/index.md`, `docs/guides/frameworks/nextjs.md` — a migration guide and an "alternative SDK" link.
- `docs/features/structured-output.md`, `google-ai.md`, `google-vertex.md:977`, `troubleshooting.md` — "all major frameworks do this" comparisons.
- `docs/advanced/builtin-middleware.md` — already says NeuroLink no longer depends on it.
- `docs/analysis/**`, `docs/features/claude-subscription-testing.md` — historical records, already marked as such by #1714.

### Testing

Documentation only; no source, no behaviour.

- `prettier --check` on all six files: clean, no reformatting needed.
- `docs-site` rebuilt; `docs-site/static/search-index.json` regenerated and committed, as `docs-site-artifacts.yml` requires. `[search-index] Generated 14356 entries from 3945 files`.

## Summary by CodeRabbit

* **Documentation**
  * Clarified PDF processing and provider-specific file handling.
  * Updated timeout terminology and provider naming.
  * Documented native Gemini 3 tool-calling support and preservation of thought signatures.
  * Clarified OpenAI streaming implementation and multi-step tool execution.
  * Refined Vercel AI SDK integration descriptions.
  * Clarified Azure OpenAI limitations with PDF file content.


docs/features/turn-time-budget.md
```diff
@@ -30,9 +30,9 @@ const result = await neurolink.generate({
 });
 ```
 
-### `turnTimeoutMs` vs `timeout` on the AI-SDK loop path
+### `turnTimeoutMs` vs `timeout` on the native loop path
```
Tara-ag: **MINOR — the doc now says "native loop path", but the code that section describes still says "AI-SDK loop path"** — same class of residue this PR exists to remove, one layer down.

The doc claim itself is **correct**: `turnTimeoutMs` owning the whole-turn abort while `timeout` stays per-model-call matches the source (e.g. `src/lib/providers/anthropic/client.ts` "an explicit, valid `turnTimeoutMs` is the caller's whole-turn contract and owns this timer"; `src/lib/providers/openaiChatCompletionsBase.ts` "The per-call `timeout` keeps its per-MODEL-CALL meaning once `turnTimeoutMs` owns the whole-turn deadline"; `src/lib/providers/googleVertex/client.ts` uses `createTurnClock({ turnTimeoutMs: ... })`).

However, `src/lib/types/generate.ts` — the JSDoc on `GenerateOptions.turnTimeoutMs` and `GenerateOptions.stallTimeoutMs` — still describes that same path in the present tense as "the AI-SDK loop path (direct Anthropic, litellm, OpenAI-compatible)". A reader who opens the type reference (also mirrored verbatim into `docs/api/type-aliases/GenerateOptions.md`, which ships with the site) gets told NeuroLink currently runs an AI-SDK loop, directly contradicting this rename. The paths it names are exactly the paths this section describes, so it is not a different layer — it is the same claim about the same code.

This is a **source-comment** residue, so it is out of this PR's docs-only scope (touching `src/` would make this more than a docs PR); flagging it as a follow-up rather than blocking.

Follow-up suggestion for `src/lib/types/generate.ts` (out of scope here):

```suggestion
As a follow-up: in `src/lib/types/generate.ts`, rename "the AI-SDK loop path (direct Anthropic, litellm, OpenAI-compatible)" to "the native loop path" in the `turnTimeoutMs` and `stallTimeoutMs` JSDoc, then regenerate the TypeDoc pages (`docs/api/type-aliases/GenerateOptions.md`) and `search-index.json`.
```

(Anchored to the renamed heading; the source file is not part of this diff.)

docs/reference/index.md
```diff
@@ -235,6 +235,6 @@ When reporting issues, include:
 
 ### Related Projects
 
-- **[Vercel AI SDK](https://github.com/vercel/ai)** - Underlying provider implementations
+- **[Vercel AI SDK](https://github.com/vercel/ai)** - Separate framework NeuroLink interoperates with via the client SDK's `LanguageModelV1` adapter
```
Tara-ag: **MINOR — wording overstates what the `LanguageModelV1` adapter is; the older "Related Projects" one-liner read truer to source.**

Verified against `src/lib/client/aiSdkAdapter.ts`: the adapter implements the Vercel AI SDK's `LanguageModelV1` interface (`NeuroLinkLanguageModel implements ClientLanguageModel`, "This allows NeuroLink to be used as a drop-in replacement for other AI providers in AI SDK applications") — so interop is real and the direction is NeuroLink-model-inside-your-AI-SDK-app. Good: the removed "Underlying provider implementations" claim was flatly wrong and its removal is the point of this PR.

The wrinkle is the phrase "the client SDK's `LanguageModelV1` adapter":

1. The adapter is not exposed through the **HTTP client SDK** surface (`src/lib/client/httpClient.ts` / `createClient`) — it is a separate module (`src/lib/client/aiSdkAdapter.ts`) whose factory is `createNeuroLinkProvider` / `createNeuroLinkModel`. Saying "the client SDK's" implies the client SDK itself ships the adapter.
2. "Separate framework NeuroLink interoperates with via the client SDK's `LanguageModelV1` adapter" makes the *client SDK* the interop channel; the adapter is the channel. Readers who go looking for it under the client SDK docs will not find it there.

Suggestion keeps the factual content, drops the "client SDK's" ownership claim:

```suggestion
- **[Vercel AI SDK](https://github.com/vercel/ai)** - Separate framework; use NeuroLink as a model inside it via the `LanguageModelV1` adapter (`src/lib/client/aiSdkAdapter.ts`)
```

(If you prefer keeping the module path out of a high-level index page, an alternative: "Separate framework NeuroLink is usable from via its `LanguageModelV1` adapter".)

Tara-ag:
## APPROVE — accurate, scoped, well-followed docs cleanup

All six edited passages were verified against source and are **accurate**; the regenerated `search-index.json` is current (the "Generated artifacts are current" check and the docs validation check both pass), commit message follows the `docs(scope)` convention, and the heading rename in `turn-time-budget.md` breaks zero links (no `ai-sdk-loop-path` anchor referenced anywhere). Zero runtime blast radius (0 changed functions/flows, risk 0.00) — documentation only.

| 1 | MINOR (follow-up) | `docs/image-generation-streaming.md:91` | Same stale present-tense claim this PR exists to remove — "Uses Vercel AI SDK's native `streamText()` function" — in a hand-written doc the sweep missed |
| 2 | MINOR (follow-up, kept) | `src/lib/types/generate.ts` + `docs/api/type-aliases/GenerateOptions.md` | JSDoc/TypeDoc still say "the AI-SDK loop path (direct Anthropic, litellm, OpenAI-compatible)" — confirmed real, reaches live readers via the shipped TypeDoc page; out of scope for a docs-only PR |

**Thread resolutions from the prior round:**
- **`yama:reference-index-adapter-wording` (thread #2) — resolved.** My own earlier finding claimed the client SDK does not ship the `LanguageModelV1` adapter. That premise is wrong: `src/lib/client/index.ts` explicitly exports `createNeuroLinkProvider`, `NeuroLinkLanguageModel`, `NeuroLinkProvider as NeuroLinkAIProvider`. The new wording "Separate framework NeuroLink interoperates with via the client SDK's `LanguageModelV1` adapter" is accurate. Resolved as a false positive.
- **`yama:ai-sdk-loop-path-source-residue` (thread #1) — kept.** Confirmed against source and now additionally confirmed to ship on the docs site via `docs/api/type-aliases/GenerateOptions.md`. No author reply; head commit unchanged.

**Verified accurate (no findings):**
- `openai.md:334` → `openaiChatCompletionsBase.ts` **is** a direct HTTP+SSE client ("A direct HTTP client + multi-step tool-execution loop driven by SSE", "Nothing here imports from `ai` or `@ai-sdk/*`"); handles text and tool-call chunks (`streamOneStep`/`executeToolBatch`); `runStreamLoop` drives `maxSteps`. Also the shared base of OpenAI, Azure, DeepSeek, LiteLLM, and the OpenAI-compatible family — citation is correct.
- `pdf-support.md:852` → `messageBuilder.ts` emits a `type: "file"` part; provider clients map PDF → provider-native part (OpenAI `file` / Anthropic `document` / Gemini `inlineData`), then "Send to provider's native PDF API" — accurate.
- `google-vertex.md:349` → native `@google/genai` path confirmed (`googleNativeGemini3/utils.ts` `extractThoughtSignature`); the historical "Vercel stripped thoughtSignature" claim is corroborated in `memory-bank/techContext.md` and the removal plans. The `:977` mention is an "all frameworks do this" comparison, correctly left alone.
- `provider-feature-compatibility.md:223` → the quoted Azure error (`'file' does not support file content types`) appears nowhere in `src/`, so "comes from the model API itself" is accurate.
- `reference/index.md:238` → swap from the flatly-wrong "Underlying provider implementations" to the interop wording (see thread #2).
- CI: single-commit policy, docs validation, search-index currency, CodeQL, secret scan, all green.

**One suggested follow-up for a later docs pass:** `docs/image-generation-streaming.md:89-91` still claims NeuroLink's text streaming "Uses Vercel AI SDK's native `streamText()` function" — fold it into this cleanup (replace with the native HTTP+SSE wording used elsewhere). Lower priority: `docs/implementation-guides/14-rag-document-processing.md` and `docs/provider-integration/*` phrases describing a `Tool`/Zod-shape compatibility are accurate (type shape, not dependency) and can stay.

Tara-ag (approved):
Approving — accurate, scoped docs cleanup; all six claims verified against source and the regenerated search-index is current. Follow-ups noted in the summary comment.

PR #1717, opened 2026-09-17
fix(proxy): handle Claude Code tool additions in Codex fallback

murdore:
Claude Code can place inline `system` messages containing `tool_addition` blocks in Messages history. The Codex fallback previously converted those unsupported blocks to `undefined`; JSON array serialization turned them into `null`, and Codex rejected the request with HTTP 400.

This change folds inline system text into Codex `instructions`, omits `tool_addition` history markers while retaining the authoritative top-level tool definitions, and filters any unsupported content part before it can become a serialized `null`. The regression reproduces the three-addition incident and also covers an addition between adjacent text blocks.

Validation:
- `pnpm check`
- `pnpm build`
- `pnpm lint` (0 errors; existing repository warnings only)
- `pnpm test:codex` (70/70)
- pre-commit validation hook
- pre-push build, provider-structure tests (3/3), and model-manifest tests (14/14)

The root `pnpm test` suite requires live Vertex credentials that are unavailable in the isolated worktree; it stopped at credential initialization rather than a source assertion.


src/lib/proxy/codexFallback.ts
```diff
@@ -42,17 +42,40 @@ function asNonEmptyString(value: unknown): string | undefined {
...
       .map((block) => (typeof block.text === "string" ? block.text : ""))
       .filter(Boolean)
       .join("\n\n");
-    return text || undefined;
+    if (text) {
+      instructions.push(text);
+    }
   }
-  return undefined;
+  for (const message of body.messages) {
```
Tara-ag, line 59: `<!-- yama:inline-system-ordering -->`
**MINOR** — folding inline `system` messages into `instructions` discards positional context.

Every mid-history `system` message is extracted and appended to the single `instructions` string, after the leading `body.system`. So a directive emitted by Claude Code between user/assistant turns is pushed to the top of the Codex request, ahead of the later user turns it was meant to qualify. If Codex ranks `instructions` above the conversation, a context-management edit that should apply only to the tail of the conversation now applies globally — a subtle behaviour change beyond the literal `null` bug being fixed here.

This is a deliberate, tested tradeoff (the test asserts `"Follow the top-level policy.\n\nUse newly available tools when needed."`), so I'm not requesting changes — just confirming the ordering is intended. If you want to keep positional fidelity, an alternative is to leave the inline system text in place as a text item at its original index rather than promoting it into `instructions`; the `null`-serialization bug is already closed by the `tool_addition` guard + the `if (contentPart)` check in `convertClaudeMessage`.

Tara-ag:
## ✅ APPROVE

Targeted fix for a real production incident: Claude Code `tool_addition` history blocks were serialized as `null` array entries and Codex rejected the request with HTTP 400. The change omits those markers (the authoritative tool defs still ride `body.tools`), folds inline `system` text into `instructions`, and filters any unsupported content part before it can become a serialized `null`. Blast radius is contained: `toCodexContentPart` has exactly one caller (`convertClaudeMessage`, same file, updated here); `convertClaudeRequestToCodex` has one production caller (`executeClaudeCodexFallback` in `src/lib/server/routes/claudeProxyRoutes.ts`, the `/v1/messages` Codex-fallback path) plus the three suite tests. All CI checks pass (lint, types, validate, build, security, extended suites incl. `test:codex`), and the new regression test reproduces the three-addition incident plus an addition between adjacent text blocks.

|---|---|---|
| MINOR | `src/lib/proxy/codexFallback.ts:56-59` | Inline `system` messages are promoted into the single top-level `instructions`, losing mid-conversation positional context. Intentional + tested; flagged to confirm ordering semantics. |

**Checked and clean:**
- `null`/`undefined` serialization path: `tool_addition` explicitly skipped; `toCodexContentPart` widened to `CodexContentPart \| undefined` and guarded by `if (contentPart)` at the only call site — no array `undefined`/`null` can reach the wire.
- System-role handling consistent between `buildSystemInstructions` (text-only extraction) and the `flatMap` exclusion in `convertClaudeRequestToCodex` — no stray `role:"system"` conversation item is emitted, and a tool_addition-only system message yields no empty instructions.
- Test style matches the established sibling ("Claude fallback converts system, history, tools, and tool results"): same `convertClaudeRequestToCodex` import from `../src`, same `as unknown as` cast, same `defineSuite` harness — no new module-graph or determinism-exception violation (suite header already declares the exception).
- No backward-compat / SDK surface change; the fallback conversion is proxy-internal, not part of the public API.

One non-blocking design question inline; nothing blocking.

Tara-ag (approved):
Approving. See the summary comment for the full review: one non-blocking MINOR inline question (`buildSystemInstructions` folding inline `system` messages into `instructions` loses positional context) — confirmed as deliberate and tested, so nothing blocking. The `tool_addition`/`null`-serialization fix is correct and well-scoped; all CI checks and the suite pass.

src/lib/proxy/codexFallback.ts
murdore: Confirmed—this is intentional. This proxy’s validated Codex conversation-input shape permits user/assistant items, while `instructions` is the supported system channel. Keeping the text at its original index would require either emitting an unvalidated runtime `system` item or weakening it into user text. Claude Code uses these inline blocks for context/tool-availability directives, and the authoritative definitions remain in top-level `tools`, so we preserve system priority and explicitly accept the ordering collapse; the regression records that behavior.

PR #1718, opened 2026-09-17
fix(bedrock): count only the tools the turn actually executed

murdore:
Review follow-up on the open thread from **#1697** (`src/lib/providers/amazonBedrock/client.ts`).

#1697 stopped Bedrock reporting **zero** tools on a turn that ran them, by forwarding `runAgenticLoop`'s telemetry. It forwarded every *dispatch*, though — including the ones that never reached a tool.

`core/loopEngine.ts` records three such cases, each pushed with `error` set:

- `TOOL_NOT_FOUND` — the model invented a name
- `TOOL_PERMANENTLY_FAILED` — the breaker had already given up on it
- a tool that threw

A call that reached the tool and returned a value is pushed **without** `error`. Mapping the list verbatim therefore reported an invented tool as *used*, re-inflating the very counts #1697 existed to make honest: analytics, cost attribution, audit.

### The fix

`toolsUsed` and `enhancedWithTools` now filter on `error === undefined`. That is the same line the shared native loop already draws — `nativeGenerateLoop`'s `toolsUsed.push(name)` sits *inside* the `try`, after the guarded call returns — so Bedrock's generate path now reports what anthropic, sagemaker and the OpenAI-compatible path report.

### Two deliberate deviations from the suggested patch

1. **`toolCalls` stays complete.** The review's snippet filtered it too. Every other provider keeps it whole (`toolCallsFromSummaries` maps all summaries), and it is the record of what the turn *attempted* — exactly what a caller debugging a failed turn needs. `toolExecutions` stays whole for the same reason; its own type documents it as *"every tool dispatch the loop performed, in order, including the ones that failed"*.
2. **No `status` field exists.** `AgenticLoopResult["toolExecutions"]` carries `{id, name, input, output, error?}` — no `status`, so `status === "SUCCESS"` would not compile. This uses the `error === undefined` fallback the review itself named.

### Testing — discriminating, no API

`test/continuous-test-suite-provider-wiring.ts`, driving `nl.generate()` from `dist` against the local h2c Bedrock endpoint (`test/helpers/bedrockLocalEndpoint.ts`, added by #1697) scripted to call a tool that was never registered. SigV4 signs with placeholder keys; nothing reaches AWS.

```
with the fix:                                 22 passed,  0 failed
with ONLY the toolsUsed filter reverted,
and rebuilt:                                  21 passed,  1 failed
  ✗ Bedrock does not report a tool the model asked for but never ran
    → toolsUsed reports a tool the turn never executed
```

The pre-existing **"Bedrock reports the tools it actually ran"** passes in *both* runs — so the new test is additive, and the filter does not break the case #1697 fixed.

The new test asserts its preconditions before the contract, because an empty `toolsUsed` is also what a turn that called nothing produces:

- the tool-result for the invented name is observed **on the wire** going back to the model,
- the registered tool's `execute()` never ran,
- `toolExecutions` is non-empty, i.e. the failed dispatch is still recorded.

`lint` 0 errors · `build` 0.

### Flagged, not changed

`NeuroLink.generate()` overwrites `enhancedWithTools` with `Boolean(result.toolExecutions?.length)`, so on the **SDK** surface that field means *"a tool was dispatched"*, while through a **provider handle** it means *"a tool ran"*. Those two definitions have disagreed across every provider for as long as both existed. Redefining a public field is not something to do from inside a Bedrock fix — raising it here instead. The provider-level value is at least made self-consistent with `toolsUsed`.


murdore:
Merged as `c1fc7b11f`, patch-identical to the pushed head `c5cd4a3b4`.

The `enhancedWithTools` inconsistency this PR flagged but did not change is now filed separately: **https://github.com/juspay/neurolink/issues/1721** — it is a public-surface change across every provider, so it wants an owner decision rather than a drive-by fix inside a Bedrock PR.

src/lib/providers/amazonBedrock/client.ts
```diff
@@ -394,7 +410,9 @@ export class AmazonBedrockProvider extends BaseProvider {
         options,
         transformToolExecutions(toolExecutions),
       ),
```
Tara-ag, line 412: **MINOR — the `enhancedWithTools` half of the fix is unobservable on the SDK surface; the provider-level value is the only place it takes effect.**

`toolsUsed` filtering is correctly surfaced (your test asserts it through the facade), but the `enhancedWithTools` gating is not. The new test itself states this at `continuous-test-suite-provider-wiring.ts:1084-1092`: `NeuroLink.generate()` overwrites the provider's `enhancedWithTools` with `Boolean(result.toolExecutions?.length)` before the caller sees it. On a turn where the model names an unregistered tool, `toolExecutions` is non-empty (your own precondition at L1073-1076 proves a `TOOL_{NOT_FOUND/PERMANENTLY_FAILED}` dispatch is still recorded), so `enhancedWithTools` reads **`true`** on the SDK surface for exactly the "model asked for a tool but none ran" scenario this PR is about. The invented-not-found turn is indistinguishable from a turn that really ran a tool.

So the provider-level value is now self-consistent with `toolsUsed`, but a caller branching on `enhancedWithTools` through the SDK still sees the lie this PR set out to remove. This is the cross-provider "ran vs dispatched" divergence you flagged (the field means "ran" through a handle, "dispatched" through the facade), and given Rule 5 (SDK backward-compat is non-negotiable) your call to not redefine the facade here is defensible. Two concrete asks to close the gap without breaking other providers:

1. File a tracking issue so the facade/SDK definition of `enhancedWithTools` (and `NeuroLink.generate()`'s overwrite at `src/lib/neurolink.ts`) is reconciled with the provider-level "ran" semantics in a follow-up — otherwise this provider-level fix will quietly only ever matter to direct-handle callers.
2. Assert `enhancedWithTools === false` on a **direct provider handle** (not via `nl.generate()`) in this test, so at least the provider-level half stays regression-guarded. Something like the `ProviderFactory.createProvider("bedrock", …)` path already used elsewhere in this file.

The author pre-justified why `enhancedWithTools` isn't asserted and why the facade isn't touched — that reasoning is sound; this is a residual-coverage and cross-provider-consistency note, not a blocker.

Tara-ag (commented):
Verified the change against loopEngine/nativeGenerateLoop/analytics and the cross-provider generate paths. The `toolsUsed`/`enhancedWithTools` filtering is correct and consistent with anthropic/sagemaker/openai-compatible; one residual coverage/consistency note on the SDK surface.

Tara-ag:
**APPROVE** — the fix is correct, well-reasoned, and fully resolves both findings from the #1697 review thread (issue thread `#discussion_r4041866652` replied to and resolved it).

## Verdict
`toolsUsed`/`enhancedWithTools` on the Bedrock generate and streamed paths now count only tool executions that actually ran (`error === undefined`), matching `nativeGenerateLoop` and the anthropic/sagemaker/openai-compatible paths. All 26 CI checks green, including `test` and the provider-safety-net contract shards.

## Findings

| MINOR | `src/lib/providers/amazonBedrock/client.ts:412` | `enhancedWithTools` provider value is unobservable via `NeuroLink.generate()` (facade overwrites with `Boolean(result.toolExecutions?.length)`), so on the SDK surface an invented/not-found dispatch still reads `true`. Not blocking (Rule 5 SDK back-compat); suggests a tracking issue + a direct-handle assertion. |

## Verified clean (evidence below)
- **`loopEngine.ts` `runAgenticLoop`**: `TOOL_NOT_FOUND`, `TOOL_PERMANENTLY_FAILED`, and thrown-tool paths all push an execution with `error` set; the success path pushes without `error`. The `error === undefined` filter is exactly the right discriminator. ✅
- **`nativeGenerateLoop.ts`**: `toolsUsed.push(name)` sits inside the `try` after the guarded call — so `toolsUsed` means "ran" there, confirming Bedrock's new behavior is the same line. ✅
- **`createAnalytics` (analytics.ts:42-49)**: `toolCallCount` derives from raw `toolExecutions.length` (still unfiltered), so the streamed-path comment ("a dispatch that failed was still a call") is accurate — no double-counting regression. ✅
- **Cross-provider consistency**: anthropic/sagemaker/openaiChatCompletionsBase all set `enhancedWithTools: loop.toolsUsed.length > 0` from the executed-only native loop list; Bedrock now joins them. `toolCalls` staying complete via `toolCallsFromSummaries` (maps all summaries) matches every other provider. ✅
- **No `status` field** exists on `AgenticLoopExecution` (`{id, name, input, output, error?}`); `error === undefined` is the correct compilable equivalent, exactly as the author's reply explained on #1697. ✅
- **New test** (`continuous-test-suite-provider-wiring.ts:1012-1102`): e2e through `dist/`, single module graph, asserts wire-level tool-result + `execute()` never ran + non-empty `toolExecutions` as preconditions before the `toolsUsed` contract — genuinely discriminating, and its env-save/restore matches the #1697 test. ⚠️ It deliberately does not assert `enhancedWithTools` (the MINOR above).

## Not changed (deliberate, per author's pre-justified reply on the resolved #1697 thread)
- `toolCalls` and `toolExecutions` kept complete by design (they are the "attempted" record) — matches other providers and the `toToolExecutionRecords` contract which explicitly retains breaker/not-found entries.
- Facade `NeuroLink.generate()` `enhancedWithTools` redefinition declined on Rule 5 grounds — reasonable; raised as the MINOR for a follow-up.

No security concerns (placeholder SigV4 creds are documented local-test scaffolding, not leaks). No dead/duplicate logic in the change.

Tara-ag (approved):
APPROVE — the `toolsUsed`/`enhancedWithTools` fix correctly counts only actually-executed tools on the Bedrock generate and streamed paths, consistent with `nativeGenerateLoop` and the anthropic/sagemaker/openai-compatible providers. Verified against `loopEngine`, `analytics`, and cross-provider generate paths; new e2e test is genuinely discriminating and single-module-graph. One MINOR (unobservable `enhancedWithTools` on the SDK surface) noted inline; non-blocking and already filed as a follow-up.

PR #1719, opened 2026-09-17
test(suites): close the review follow-ups left on #1693, #1696 and #1707

murdore:
Five review threads that stayed open when their PRs merged. None of them changes what the fix underneath it does.

---

### #1693 — a deliberately narrow classifier, quietly widened

`isInfraError` in the JSON e2e suite gained a `timed.?out` alternate. The free `.` also matches `timedXout` and `timed/out`, and that function drives the **entire** `CELLS` provider matrix, not only the cohere cell it was added for — on a function whose own comment calls it *"deliberately narrow: generic 4xx / 'bad request' would mask real request-construction bugs"*.

Scoped to `timed[ -]?out`. Verified against the real strings:

| input | before | after |
|---|---|---|
| `operation timed out` | match | match |
| `request timed-out` | match | match |
| `timedout` | match | match |
| `the request timedXout after 30s` | match | **no** |
| `timed/out` | match | **no** |

### #1696 — SDK instances that could not be shut down

Ten tests in the Vertex+Claude characterization suite built a `NeuroLink` **inside** `try`, so the `finally` had nothing to call `shutdown()` on and SDK-owned handles outlived each case. Each instance is hoisted and awaited in `finally`, ahead of `restore()` and the stand-in server's close. The reviewer flagged two sites; the same shape was at all ten.

### #1707 — three threads

**The mapping test only asserted ordering.** Its name claimed it pinned `0/0.25/0.55/1`, but a changed fraction — or a changed rounding rule — still lands strictly between the endpoints and passes. It now pins the exact budgets `128 / 6240 / 13574 / 24576` over `[128, 24576]`, keeping the existing floor/ceiling precondition in front of it.

**Two live tests compared `reasoningTokens` across turns.** Gemini documents `thinkingBudget` as a **soft** upper limit — the model spends what the prompt needs and routinely spends less — so live usage is not monotonic in the requested budget, and `minimal < high` can fail on a turn whose emitted budgets were entirely correct. Both comparisons are replaced with what a live turn can honestly prove: the request was accepted and the model thought. Both tests are **renamed**, since the old names described the assertion that is now gone. The mapping stays covered deterministically by the no-live-call test.

**`NativeThinkingConfig`'s doc overstated its own reach.** It said Gemini "2.5 (and earlier)" models need a numeric budget. `createNativeThinkingConfig` only ever emits one for 2.5 — `getGemini25ThinkingBudgetRange` has no verified range for anything older, so those models get **no** `thinkingConfig` at all plus the one-time WARN #1707 added. Said so.

---

### Testing

```
test/continuous-test-suite-gemini-thinking-level.ts           6 passed, 0 failed
  ✓ mapThinkingLevelToBudget pins the level->budget fractions (no live call)
  ✓ gemini-2.5-flash-lite accepts both thinkingLevel extremes and thinks on high
  ... (live, against real Vertex)

test/continuous-test-suite-vertex-claude-characterization.ts 10 passed, 0 failed
  (all ten shutdown sites, against the local Anthropic-SSE stand-in)

lint  0 errors, 69 warnings (unchanged from release)
build 0
```

`docs/api/type-aliases/NativeThinkingConfig.md` is regenerated for the docstring change, as the `docs:api` currency gate requires.

`docs-site/static/search-index.json` is **not** touched, and does not need to be: the index excludes `docs/api` (`3552 files excluded`). Verified by rebuilding the site — `git status --porcelain docs-site/static/search-index.json` came back empty.


murdore:
Force-pushed `62b51ec` → `6352042` to fix a **real failure CI caught that my local run did not**.

`test-shards (types)` went red on `🔎 Typecheck tools/ and test/`:

```
'NeuroLink' refers to a value, but is being used as a type here.
Did you mean 'typeof NeuroLink'?   (×10)
```

The hoisted declarations were written `let nl: NeuroLink | undefined`. That suite gets the class from a **destructured dynamic import** — `const { NeuroLink } = await import("../dist/index.js")` — so `NeuroLink` is a `const` binding, not a class declaration, and cannot be used as a type. Now `InstanceType<typeof NeuroLink>` at all ten sites.

Worth recording why I missed it: `pnpm run build` compiles `src/` only. The gate that covers `test/` is `check:tools-tests`, which is a separate script and is not part of `lint` or `build`. Ran it directly this time — **exit 0**.

Re-verified on the amended head:

```
check:tools-tests                                             exit 0
test/continuous-test-suite-vertex-claude-characterization.ts  10 passed, 0 failed
lint                                                          0 errors
```

The change is a type annotation only — erased at runtime, so nothing about the shutdown behaviour the commit describes has changed.

test/continuous-test-suite-vertex-claude-characterization.ts
```diff
@@ -405,6 +406,7 @@ await test("a Claude-on-Vertex turn can be answered by a local endpoint", async
     failure =
       error instanceof Error ? error.message.slice(0, 160) : String(error);
   } finally {
```
Tara-ag, line 408: **MINOR — shutdown runs first in `finally`, and bare-form `shutdown()` is not the pattern all sibling suites use.**

The fix itself is right and fully lands the #1696 thread — all ten `NeuroLink` instances are now hoisted and `await nl?.shutdown()` appears in each `finally`. Two small robustness notes on this exact line:

1. **Ordering.** `await nl?.shutdown()` is the *first* statement of `finally`, ahead of `restore()` and `server.close()`. If it ever rejects, `restore()` (the env that `withVertexEnv()` saved) and `server.close()` are skipped entirely. Because several of these tests use a bare `catch {}` that swallows the *body* failure, a rejecting shutdown would propagate out of `await test(...)`, masking the real assertion result and leaving the environment mutated for the next case.

2. **The "matches the other eleven suites" claim is slightly overstated.** Sibling suites that construct their own SDK are more defensive: `test/continuous-test-suite-context.ts` uses `await sdk.shutdown?.()?.catch(() => {});`, and `continuous-test-suite-providers.ts` (the pattern cited in the #1696 reply) uses `await sdk.shutdown?.().catch(() => {});`. The bare form here is the only one that can throw out of `finally`.

Suggested (drop-in for this line and the other nine sites):

````suggestion
    await nl?.shutdown().catch(() => {});
    restore();
    await server.close();
````

Alternatively move `restore()` ahead of `shutdown()` so env state is never left dirty. Non-blocking — `NeuroLink`'s internal cleanup is defensive and the rejection risk is low — but a one-character-hardening `.catch(() => {})` removes the only way these characterization tests can fail on teardown rather than on what they pin.

Tara-ag (commented):
Review of the five follow-up closures (see summary comment for the full table). One MINOR robustness note on the shutdown sites in the Vertex+Claude suite.

Tara-ag:
## Verdict: **APPROVE**

This PR is the vehicle that lands five previously-accepted review MINOR findings from #1693, #1696 and #1707. I verified each claim directly against the PR-head code — all five are implemented exactly as the resolved threads described. CI is green (29 checks: types/lint/build, Extended Suites, provider-safety-net, security suites, docs validation, generated-artifacts-current all success).

| Severity | Finding | file:line | Status |
|---|---|---|---|
| MINOR | `timed.?out` → `timed[ -]?out` in the shared infra skip-classifier | `test/continuous-test-suite-json-e2e.ts:431` | ✅ resolved |
| MINOR | 10 × `NeuroLink` hoisted + `await nl?.shutdown()` in `finally` | `test/continuous-test-suite-vertex-claude-characterization.ts` (10 sites) | ✅ resolved (1 note below) |
| MINOR | `mapThinkingLevelToBudget` pinned to exact budgets `128 / 6240 / 13574 / 24576` | `test/continuous-test-suite-gemini-thinking-level.ts:110` | ✅ resolved |
| MINOR | flaky live `reasoningTokens` budget-order comparisons removed; tests renamed | `…gemini-thinking-level.ts:235`, `…-ai-studio.ts:173` | ✅ resolved |
| MINOR | `NativeThinkingConfig` docblock corrected to "Gemini 2.5 only" + regenerated `.md` | `src/lib/types/config.ts:609`, `docs/api/type-aliases/NativeThinkingConfig.md` | ✅ resolved |

**Verification notes:**
- Budget pin: `Math.round(128 + f·24448)` yields low=6240, medium=13574.44→13574, high=24576 — the four literals match `0/0.25/0.55/1` precisely; keeping the floor/ceiling precondition in front is good.
- Reasoning-token removal: `thinkingBudget` as a soft upper limit is the correct rationale; the `> 0` assertion is a pre-existing pattern in this suite (the `gemini-2.5-pro`/`-flash` "thinks" tests already assert `reasoningTokens > 0`), so the replacements introduce no new flake. `minimalResult.reasoningTokens` is still consumed by the precondition `typeof … === "number"` check, so no dead code.
- Docstring: claims (a) Gemini 2.5 is the only budget-emitted family, (b) `getGemini25ThinkingBudgetRange` has no earlier range, (c) earlier models get no config + one-time WARN — all match `createNativeThinkingConfig` / `warnedNoBudgetRangeModels` in `src/lib/utils/thinkingConfig.ts`. The regenerated markdown line anchors (620/621/622/623) are consistent with the +5 net lines.
- `InstanceType<typeof NeuroLink>` is correct here: the suite binds via `const { NeuroLink } = await import("../dist/index.js")`, so `NeuroLink` is a value, not a type — the author's force-push note (`6352042`) explains the CI catch correctly.

**What I checked and found clean:** the `docs-site/static/search-index.json` non-regeneration claim (index excludes `docs/api`); no source-diff beyond the `config.ts` comment (this PR is test/docs-only; `thinkingConfig.ts` and model-detection logic are untouched, so no Rule 5 blast radius, no provider-path change); no `src/`/`dist/` module-graph mixing in any changed suite (all import from `../dist/*`); single-commit policy satisfied.

**One MINOR (not blocking):** in `vertex-claude-characterization.ts`, `await nl?.shutdown()` sits *first* in each `finally`, ahead of `restore()` and `server.close()`. A rejecting shutdown would skip env/server teardown and, given the bare `catch {}` bodies, propagate out of `await test(...)`. The "matches the other eleven suites" framing is also slightly overstated — `continuous-test-suite-context.ts` and `continuous-test-suite-providers.ts` guard with `?.().catch(() => {})`. Suggest `await nl?.shutdown().catch(() => {});` (or move `restore()` up). Low risk given `NeuroLink`'s defensive internal cleanup.

No CRITICAL or MAJOR findings. Ready to merge.

test/continuous-test-suite-vertex-claude-characterization.ts
Tara-ag: **Repairing the suggestion block in my own comment above** — the anchored range is a single line (`test/continuous-test-suite-vertex-claude-characterization.ts:408`, the `await nl?.shutdown();` line), but the suggestion block carried a 3-line body. Applying it as posted would replace line 408 with three lines and duplicate the existing `restore();` / `await server.close();` on 409–410. The correct drop-in for line 408 (and the other nine sites) is the one-liner:

````suggestion
    await nl?.shutdown().catch(() => {});
````

Rationale unchanged: keep `restore()` and `server.close()` exactly where they are, and only harden the shutdown call so it cannot throw out of `finally` and skip env/server teardown. The ordering note (shutdown-first-in-finally, ahead of `restore()`) still stands but is non-blocking. If you prefer, move `restore()` to the top of `finally` instead — either fixes the only way these characterization tests can fail on teardown rather than on what they pin.

Tara-ag (approved):
Approve — this PR lands the five previously-accepted MINOR follow-ups from #1693/#1696/#1707, all verified against the head code, with one non-blocking MINOR on shutdown-teardown ordering in the Vertex+Claude suite (see inline comment). No CRITICAL/MAJOR findings. (Full detail in the summary comment.)

Issue #1722, opened 2026-09-17
docs:api silently emits an empty doc set when the checkout path contains a "test" segment


murdore:
## Symptom

`pnpm run docs:api` exits **0**, prints `markdown generated at ./docs/api`, emits **no errors and no warnings** — and writes exactly **one** file (`docs/api/README.md`) instead of 3546. Because `typedoc.json` sets `cleanOutputDir: true`, the other 3545 files are deleted first.

So a working tree that looks fine is left with a 3545-file deletion staged for commit, from a command that reported success.

```
$ pnpm run docs:api
[info] Loaded plugin typedoc-plugin-markdown
[info] markdown generated at ./docs/api
$ git status --porcelain docs/api | awk '{print $1}' | sort | uniq -c
3545 D
   1 M
```

## Root cause

`typedoc.json`:

```json
"exclude": [
  "**/node_modules/**",
  "**/test/**",
  "**/*.test.ts",
  "**/*.spec.ts"
]
```

typedoc matches `exclude` against **absolute** paths. `**/test/**` therefore matches *any* checkout whose path has a directory segment named `test` — and then every file under `src/lib/` is excluded, the entry point converts to an empty module, and only the module page is written.

I hit this in a git worktree at:

```
/Users/sachinsharma/Developer/temp/neurolink-fork/test/review-followups-1693-1696-1707
                                                  ^^^^
```

This is not exotic: `workforge create -t test -n <name>` puts the worktree under a directory named after the branch type, so **every worktree created for test work lands on this**.

## Evidence

Three worktrees, same commit `e1575587a`, same lockfile, same `typedoc@0.28.18` + `typedoc-plugin-markdown@4.11.0`, byte-identical `tsconfig.json` / `typedoc.json` / `src/lib/index.ts`, and `tsc -p tsconfig.json --noEmit --listFiles` returning **4847** files in each:

| worktree path | `docs:api` result |
|---|---|
| `neurolink-fork/docs/ai-sdk-claims-residue` | 3546 pages, no drift |
| `neurolink-fork/fix/bedrock-executed-tool-telemetry` | 3546 pages, no drift |
| `neurolink-fork/**test**/review-followups-…` | **1 page, 3545 deletions** |

Ruled out in the broken worktree: my own edits (reproduces with everything stashed), a stale `dist` (reproduces with it moved aside), `.svelte-kit`, a stray `package.json` under `src/`, and the install itself — `rm -rf node_modules && pnpm install` did not change it.

[...]

PR #1723, opened 2026-09-17
fix(docs-api): anchor the typedoc exclude to the project root

murdore:
Closes **https://github.com/juspay/neurolink/issues/1722**.

`pnpm run docs:api` exited **0**, printed `markdown generated at ./docs/api`, raised **no errors and no warnings** — and wrote exactly **one** file instead of 3546. `cleanOutputDir: true` had already deleted the other 3545. A command that reported success left a mass deletion staged in the working tree.

## Cause

typedoc matches `exclude` against **absolute** paths, and the pattern was `**/test/**`. Any checkout with a directory segment named `test` *anywhere above it* therefore excluded its own `src/lib/`; the entry point converted to an empty module and only the module README survived.

That is not an exotic path — `workforge create -t test -n <name>` places the worktree under a directory named for the branch type, so **every worktree opened for test work sat on this**.

```diff
   "exclude": [
     "**/node_modules/**",
-    "**/test/**",
+    "./test/**",
     "**/*.test.ts",
     "**/*.spec.ts"
   ],
```

`./test/**` cannot match an ancestor of the repository.

## Testing — from the failing condition, not next to it

Built in a worktree at `neurolink-fork/`**`test`**`/anchor-typedoc-excludes`, so the bug reproduces before the change and is gone after it:

| | `deferred tasks` | `pages to write` | diagnostics | `git status --porcelain docs/api` |
|---|---|---|---|---|
| **before** | 1 | 1 | *(no summary at all)* | **3546 files** — 3545 `D`, 1 `M` |
| **after** | 768 | 3546 | `Found 0 errors and 95 warnings` | **0 files** |

Zero drift is also what proves the root `test/` directory is **still excluded**: if the anchored pattern had stopped matching it, `docs/api` would have *gained* pages rather than matching release byte for byte. Confirmed through `pnpm run docs:api` itself, not a `--exclude` flag override.

`lint` 0 errors · `prettier --check` clean on both files.

## Deliberately not touched

`**/node_modules/**` keeps its pattern. A nested `node_modules` is real under pnpm and must stay excluded; a repository checked out *beneath* a directory named `node_modules` is not a case worth trading that for.

## Why CI never caught it

GitHub checks out at `/home/runner/work/neurolink/neurolink`, which has no matching segment — so the `docs/api` currency gate always computed the right answer, and would have failed any PR that committed the deletions. The cost was entirely local and entirely quiet: two worktrees at the same commit, same lockfile, identical `tsc --listFiles` output of 4847 files, disagreeing by 3545 files, with neither a reinstall nor a rebuild changing anything.

That is why the incident is also recorded in `CLAUDE.md`, in the reproducible-generator section, right after *"a command that exits 0 is not evidence it did anything"* — this is the sharper version, where the artifact **was** written and is still wrong.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed API documentation generation when the project is checked out under a path containing a `test` directory.
  - Documentation builds now preserve the full generated output instead of removing most files and producing only a module README.
  - The exclusion rule now applies only to the project’s top-level test directory.

- **Documentation**
  - Added guidance describing the affected behavior, its path-dependent conditions, and the corrected configuration.


typedoc.json
```diff
@@ -13,7 +13,7 @@
   "sort": ["source-order"],
   "exclude": [
     "**/node_modules/**",
-    "**/test/**",
+    "./test/**",
```
Tara-ag, line 16: **MINOR** — the fix is correct, but this bug class has no automated guard.

`**/test/**` → `./test/**` is the right fix: typedoc matches `exclude` globs against absolute paths, so the old bare pattern matched any checkout whose absolute path contained a `test/` segment and silently excluded all of `src/lib/`. Anchoring to `./test/**` resolves against the project root and only matches the repo's own `test/`. Verified against the run log in the PR body (3545 files restored) and consistent with the `docs/api` currency gate in `.github/workflows/ci.yml` still passing with no committed regeneration.

The gap: CI checks out at `/home/runner/work/neurolink/neurolink` (no `test` segment), so no automated check can ever exercise the path that broke. Per the repo's own documented lesson in this PR's CLAUDE.md section — "a command that exits 0 is not evidence it did anything" — this is a silent-failure class. Consider a deterministic guard, e.g. a script/hook that runs `pnpm run docs:api` from a synthetic checkout path containing a `test/` segment and asserts page count > 1, so the regression is caught when reintroduced rather than only by a developer who happens to sit under a `test` directory. Non-blocking.

Tara-ag (approved):
Fix is correct and well-verified. Approving — one non-blocking MINOR suggestion about a regression guard.

Tara-ag:
**APPROVE** — the typedoc `exclude` anchor fix is correct, narrowly scoped, and empirically verified; one non-blocking MINOR suggestion about a regression guard.

## Verdict
The change addresses issue #1722 precisely. `**/test/**` was matched by typedoc against **absolute** paths, so any checkout whose path contains a `test/` segment had its entire `src/lib/` excluded (`cleanOutputDir` then deleted 3545 of 3546 files). Anchoring to `./test/**` (root-resolved) fixes it without changing output at the normal CI path. Zero runtime code impacted (<2 lines config + docs).

## Findings

|---|---|---|
| MINOR | `typedoc.json:16` | No automated regression guard for a path-dependent bug class CI's fixed checkout path structurally cannot exercise. Suggest a deterministic guard (e.g. `docs:api` from a synthetic path containing `test/`, assert page count). |

## What I checked and found clean
- **Fix correctness**: `./test/**` is root-anchored and cannot match an ancestor; `**/test/**` could. `**/*.test.ts`/`**/*.spec.ts` are basename patterns (unaffected); keeping `**/node_modules/**` broad is deliberate and justified (pnpm nested `node_modules`).
- **Empirically verified by author**: 1 → 3546 pages; zero drift at the CI checkout path (consistent with no committed `docs/api` regeneration).
- **CI `docs/api` currency gate** (`.github/workflows/ci.yml`, "Check generated API docs are current") unchanged and still passes.
- **Blast radius** (code graph): 0 nodes / 0 files impacted — config + docs only; no callers dependents affected.
- **CLAUDE.md addition**: accurate, well-placed in the reproducible-generator lesson section, matches the incident-log style; no factual issues.
- **No prior review threads or author replies** to reconcile (only automated bot comments); single commit, conventional message, scope correct.

No CRITICAL/MAJOR issues. Merging is safe.

typedoc.json
Tara-ag, line 16: **MINOR** — this fix removes the path-dependent bug, but nothing guards it from regressing.

The failure class (`exclude` matching ancestor directories, `exit 0` with silent mass deletion) is precisely the kind that CI cannot reproduce: GitHub checks out at `/home/runner/work/neurolink/neurolink`, and the `docs/api` currency gate in `.github/workflows/ci.yml` runs on that same safe path. So a future reintroduction of a bare `**/test/**` (or any unanchored directory pattern) would neither fail CI nor trip the docs gate.

A deterministic guard would pin the behaviour regardless of checkout path — e.g. assert after `pnpm run docs:api` that exactly the expected set of API pages exist and that nothing under the entry-point subtree is missing, rather than relying on path luck.

Non-blocking — happy to leave as a follow-up.

Tara-ag (commented):
Posting the MINOR regression-guard note as an inline comment; the APPROVE verdict stands from the earlier review.

Tara-ag:
Duplicate summary removed — superseded by the canonical summary above. No separate action required; see the `<!-- yama:summary -->` comment for the full review.

murdore:
**Now 5/5 green on the same head `9137a1177` — nothing was pushed between the two attempts.**

First attempt, `provider-safety-net-shards (rest)` failed with 3 of 21 in `continuous-test-suite-anthropic-execution-control.ts`. This PR changes `typedoc.json` (one line) and `CLAUDE.md` (prose). Neither is read by that suite, neither affects the build, and `typedoc.json` is only consulted by `pnpm run docs:api`.

Three data points on the identical commit:

| | result |
|---|---|
| CI run 35288089081 attempt 1 | 18 passed, **3 failed** |
| CI run 35288089081 attempt 2 (`--failed` re-run, same head) | **success** |
| local, same worktree, same head | **21 passed, 0 failed** |

And `544a23f80` — the commit immediately before this one on `release` — had the same shard green.

Filed as **https://github.com/juspay/neurolink/issues/1724**, because the failure message is actively misleading: it says *"nothing here waits on a network, so this is a hang in the code under test, not a slow upstream"*, which rules out the cause that actually applies. Offline excludes a slow upstream; it does not exclude a slow runner, and this suite's whole subject is wall-clock timers.

PR #1730, opened 2026-09-19
fix(providers): stop charging the tool-free reformat pass for tools it omits

murdore:
## What's wrong

The tool-free reformat pass omits `tools` from the request — that's what makes `response_format` legal again on a vendor that refuses tools and JSON mode together. The context guard's fixed overhead didn't follow: it serialized `v3Tools` unconditionally, so that pass was billed for a schema it never sent.

```ts
getFixedOverheadTokens: () =>
  estimateTokens(serializeForEstimate({ tools: v3Tools, responseFormat: format })),  // always
...
...(withTools && v3Tools ? { tools: v3Tools } : {}),                                 // only sometimes
```

Three call sites pass `withTools: false`.

## Why it matters

That number feeds `planLoopGuardReclaim`, which fires above `floor(availableInputTokens * 0.85)`. So a large tool schema on its own can push the reformat pass over the threshold and **delete older tool exchanges from a request that would have fit**, replacing them with the `[Earlier tool exchanges were removed...]` marker.

## Measured, same conversation and same build

| | overhead | total | threshold | result |
|---|---|---|---|---|
| before | 12,010 | 59,941 | 57,800 | fires — 5 tool exchanges cut to 2 |
| after | 41 | 47,972 | 57,800 | does not fire — all 5 retained |

`msgTokens` is 47,931 in both runs. The overhead is the entire difference.

## Why the test is not vacuous

`test/continuous-test-suite-toolfree-guard-overhead.ts` drives the public `generate()` against a local scripted endpoint and asserts on outbound request bodies.

- It asserts the tool-free reformat request was actually sent, so it can't pass by never reaching the branch.
- It asserts **no tool-bearing request was reclaimed**, so a loop-time reclaim can't be mistaken for this bug. An earlier draft of this fixture failed exactly that way — before/after came out identical because the conversation had already been trimmed during the tool loop, where tools genuinely *are* sent.
- A **control case** with more exchanges, genuinely over threshold, asserts reclaim still fires — so the primary assertion can't pass against a guard that had stopped working.

Reverting the one-line production change:

```
✗ the tool-free reformat pass is not charged for the tools it omits
✓ control: a genuinely oversized conversation still reclaims on the reformat pass
```

It fails with `✗`, not `⊘`, so it isn't masked by the skip-on-provider-error path.

## Fixture notes

MCP and built-in tools are disabled via `NEUROLINK_SKIP_MCP` and `NEUROLINK_DISABLE_BUILTIN_TOOLS`. That keeps the tool schema a controlled quantity instead of whatever the host registers — and cuts the run from minutes to seconds, since tool discovery was the cost, not the model calls.

## Provenance

Reported by CodeRabbit on #1670. That PR's diff carried this code only because its parent #1668 hadn't merged yet; the finding belongs to `release`, so it's fixed here rather than folded into an unrelated PR.

`docs/api/type-aliases/NativeGenerateGuardConfig.md` is regenerated. Its line references were **already stale on release** — `src/lib/types/context.ts` is untouched in this PR — so this just brings the generated artifact back in step.

## Gates

`check` 0 errors · `check:tools-tests` 0 errors · `lint` 0 errors (75 pre-existing warnings) · `build` clean · `toolfree-guard-overhead` 2/2 · `provider-structure` 3/3

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved message-history handling for requests without tools, preventing unnecessary removal of older conversation exchanges.
  - Tool configuration is no longer included in token estimates when tools are disabled.
  - Tool-free reformat requests continue to include the expected response format.

- **Tests**
  - Added coverage for tool-free requests, including safeguards confirming that genuinely oversized conversations are still trimmed appropriately.


Tara-ag:
## APPROVE — clean, correct, focused bug fix

**Verdict: APPROVE.** The two-line fix stops charging the tool-free reformat pass for the tool schemas it omits, brings the guard estimate in line with what is actually sent, and the regression suite (with a working control case) pins the exact failure mode.

### Findings

|---|---|---|
| — | — | No findings |

### What was checked

- **Guard estimate now matches the sent body.** `getFixedOverheadTokens` counts `v3Tools` only under `withTools && v3Tools`, which mirrors the request-side spread `...(withTools && v3Tools ? { tools: v3Tools } : {})` one comment block below. In both branches the estimate and the wire body agree on whether tools are present: when `withTools=true` behavior is byte-for-byte unchanged; when `false` the reformat pass no longer bills for a schema it never sends. The `withTools`/`v3Tools` closure scope is in range. (`src/lib/providers/openaiChatCompletionsBase.ts`)
- **Refactor is behaviour-preserving.** The arrow‑expression → block‑body change with an explicit `return runNativeGenerateLoop(...)` is semantically identical to the previous implicit return.
- **Test quality (rule 15).** The new suite imports only from `dist/index.js`, drives the public `generate()` against a local scripted server, and — the strong part — includes a *control* case that asserts reclaim still fires for a genuinely oversized conversation, so the primary assertion cannot pass against a guard that stopped working. Env toggling (`NEUROLINK_SKIP_MCP`, `NEUROLINK_DISABLE_BUILTIN_TOOLS`) is scoped and restored in a `finally`.
- **Scoped diff.** The only production change is the two-line guard; `package.json` adds the run script; the doc diff is a regenerated artifact whose line refs were already stale on release (`src/lib/types/context.ts` is untouched here).

**Security bar:** no secrets, no injection, no dynamic-import violations (rule 1), no SDK API surface change (rule 5), no structured-output policy change (rule 3), no `formatProviderError` change (rule 6). Impact analysis: the change is local to a single closure and does not alter any exported signature or return shape.

Tara-ag (approved):
**APPROVE.** The guard change correctly stops charging the tool-free reformat pass for the tool schemas it omits — the `withTools && v3Tools` gate in `getFixedOverheadTokens` now agrees with the request-side spread — and the e2e suite (with a genuine control case) pins the regression. No findings.

Issue #1724, opened 2026-09-18
continuous-test-suite-anthropic-execution-control is flaky on CI, and its timeout message asserts a cause it cannot know


murdore:
`test/continuous-test-suite-anthropic-execution-control.ts` runs inside **`provider-safety-net`**, which is a required status check. A flake there blocks an unrelated PR at random.

## What happened

PR **#1723** changes exactly two files — `typedoc.json` (one line) and `CLAUDE.md` (a prose note). Neither is read by this suite, neither affects the build, and `typedoc.json` is only consulted by `pnpm run docs:api`.

`provider-safety-net-shards (rest)` failed on it anyway:

```
Passed: 18   Failed: 3   Total: 21   Time: 188.59s

✗ turnTimeoutMs alongside an explicit lifetimeTimeoutMs is rejected, not dropped
    … exceeded 60000ms in an offline suite — nothing here waits on a network,
      so this is a hang in the code under test, not a slow upstream
✗ a finite lifetimeTimeoutMs still caps the turn
    … exceeded 60000ms in an offline suite — …
✗ lifetimeTimeoutMs null outlives the legacy timeout and renews across many steps
    … every renewal offered must have been taken, took 0 of 3
```

## It is a flake

Three independent data points, all on the **same commit** `9137a1177`:

| | result |
|---|---|
| CI, run 35288089081 attempt 1 | 18 passed, **3 failed** |
| CI, run 35288089081 attempt 2 (`gh run rerun --failed`, same head, nothing pushed) | **success** |
| local, same worktree, same head | **21 passed, 0 failed** |

The immediately preceding commit on `release` (`544a23f80`, PR #1719) had this same shard green.

## The part worth fixing

The message is not just wrong here, it *rules out* the actual cause:

> nothing here waits on a network, so **this is a hang in the code under test, not a slow upstream**

That inference does not hold. "Offline" excludes a slow *upstream*; it does not exclude a slow *runner*. A shared CI executor under load can stretch a 60s budget on a suite whose whole subject is wall-clock timers — the third failure (`took 0 of 3` renewals) is the same story from the other side: renewals that were offered but not observed within the window.

So the assertion tells whoever reads the log to go hunting for a hang in the provider, which is precisely where the time will not be.

Two things to consider:

1. **Soften the claim.** Say the budget was exceeded and name the two possibilities, rather than asserting one. A message that eliminates a real cause is worse than a message that just states the fact.
[...]

PR #1732, opened 2026-09-19
fix(harness): stop the offline-timeout message asserting a cause it cannot know

murdore:
Closes **https://github.com/juspay/neurolink/issues/1724**.

`test/helpers/harness.ts` emitted, whenever an offline suite's case exceeded its budget:

> …exceeded Nms in an offline suite — nothing here waits on a network, **so this is a hang in the code under test, not a slow upstream**

That inference does not hold. "Offline" rules out a slow *upstream*; it does not rule out a slow *runner*. The message doesn't merely fail to help — it **rules out the cause that actually applied** and sends the reader hunting for a provider hang that isn't there.

It cost exactly that on #1723, a PR changing one line of `typedoc.json` and a prose block in `CLAUDE.md`. Three data points, same commit `9137a1177`:

| | result |
|---|---|
| CI attempt 1 | 18 passed, **3 failed** |
| CI attempt 2 (`--failed` re-run, nothing pushed) | success |
| local, same worktree | 21 passed, 0 failed |

### Two changes

1. **The message states the fact and names both possibilities** — a genuine hang in the code under test, or contention on a shared runner — without asserting either.
2. **`NEUROLINK_TEST_TIMEOUT_SCALE`** multiplies the per-test budget, defaulting to `1` so local behaviour is byte-for-byte unchanged. CI can widen it without anyone editing per-suite constants.

This matters more than a normal flaky test because `provider-safety-net` is required and unbypassable on ruleset `11413189`, so each flake costs a full re-run on an innocent PR.

### Testing

New suite `test/continuous-test-suite-harness-offline-timeout.ts`, 6 cases:

- `resolveTimeoutScale` falls back to 1 for every invalid/absent input, and passes valid positive finite overrides through
- an offline timeout names both explanations, drops the old false-certainty phrasing, and still **fails** rather than skipping
- a live (non-offline) timeout is unaffected and still skips
- unset scale enforces exactly the suite's own budget; a set scale genuinely widens it

Every timing assertion carries an elapsed-time precondition proving the case actually ran first — otherwise an absence in the captured output would pass for the wrong reason.

```
new suite:                                6 passed, 0 failed
with harness.ts reverted to HEAD:         exit 1 (import of the new export fails)
restored:                                 6 passed, 0 failed

continuous-test-suite-anthropic-execution-control.ts   21/21  (the suite from the issue)
continuous-test-suite-handler-registry.ts               9/9
lint 0 errors · build 0 · check:tools-tests 0
```

The new suite is deliberately **not** wired into a required gate — it tests the harness, and putting it behind the harness's own CI job would be circular.


test/helpers/harness.ts
```diff
@@ -543,11 +578,21 @@ export function defineSuite(
   const failures: { name: string; error: string }[] = [];
   const startedAt = Date.now();
 
-  const perTestTimeoutMs = defs.perTestTimeoutMs ?? 240_000;
+  // Scaled by NEUROLINK_TEST_TIMEOUT_SCALE (see this file's header comment);
+  // the scale is 1 by default, so this is exactly today's fixed value.
+  const perTestTimeoutMs = Math.round(
+    (defs.perTestTimeoutMs ?? 240_000) * resolveTimeoutScale(),
+  );
```
Tara-ag, line 585: **MINOR** — a valid *fractional* scale can collapse a small suite budget to `0ms`, contradicting the "never shrink to 0" intent documented just above.

`resolveTimeoutScale()` only guards the *scale* (`parsed > 0`), not the resulting budget. A positive finite scale like `0.001` passes the parser and makes `Math.round((defs.perTestTimeoutMs ?? 240_000) * 0.001)` round down: for any suite budget under ~1000ms (this very file's new suite uses 100–200ms budgets) that yields `0`, so the timeout fires essentially immediately. Default 240_000 is immune, but small per-suite overrides — which the new suite itself exercises — are not.

Fix: floor the result at 1 so a scaled budget can never be 0.

````suggestion
  const perTestTimeoutMs = Math.max(
    1,
    Math.round((defs.perTestTimeoutMs ?? 240_000) * resolveTimeoutScale()),
  );
````

Tara-ag (approved):
## Review summary



**APPROVE** — a small, well-engineered, correctly-tested change to the test harness that makes the offline-timeout classification honest; it touches only `test/` infra and `package.json`, with no shipped-surface impact.

|---|---|---|
| MINOR | `test/helpers/harness.ts:583-585` | `Math.round` can collapse a small suite budget to `0ms` under a valid fractional scale — guard with `Math.max(1, …)` |

- **Blast radius** — confined to `test/helpers/harness.ts`, the new offline suite, and a `package.json` script. No `src/` surface is touched, so Rule 5 (public SDK API), Rules 1/3/4 and the provider/MCP/auth hot paths are untouched.
- **New behaviour** — `resolveTimeoutScale()` / `TIMEOUT_SCALE_ENV_VAR` parsing and fallbacks (unset → 1, NaN → 1, `<0` → 1, finite parse) are covered by table-driven cases; the offline message string is asserted character-for-character against the actual message in the new suite.
- **Logic** — the scale is read at `defineSuite` time inside the env-scoped `withTimeoutScaleEnv` blocks (correct ordering); scale-unset path reproduces exactly today's fixed budget; the FAIL (✗) classification of a timed-out offline case is consistent.
- **Rule 15** — the new suite drives the harness module directly (the documented carve-out) and never mixes `src/`/`dist/`; the header documents why it needs no `allow`-list entry. No lint/format regressions found worth surfacing beyond what CI owns.

No security, correctness, or backward-compat concerns. One optional robustness nit above from a real (if unlikely) edge case.

Issue #1305, opened 2026-08-08
checkAllProvidersHealth() silently checks 8 of 31 providers


murdore:
Split out of #342, which is being closed: rebuilding the `BaseProvider.checkHealth()` architecture that issue asked for would duplicate the better `ProviderHealthChecker` that already ships. But while verifying that, a concrete gap turned up in the replacement.

## The problem

`ProviderHealthChecker.checkAllProvidersHealth()` iterates a hardcoded array:

```ts
// src/lib/utils/providerHealth.ts:1984-1993
const providers: AIProviderName[] = [
  AIProviderName.VERTEX,
  AIProviderName.GOOGLE_AI,
  AIProviderName.ANTHROPIC,
  AIProviderName.OPENAI,
  AIProviderName.BEDROCK,
  AIProviderName.AZURE,
  AIProviderName.LITELLM,
  AIProviderName.OLLAMA,
];
```

**Eight providers. `AIProviderName` has 31.**

The 23 it never checks include `openai-compatible`, `openrouter`, `huggingface`, `mistral`, `sagemaker`, `deepseek`, `nvidia-nim`, `lm-studio`, `llamacpp`, `xai`, `groq`, `cohere`, `together-ai`, `fireworks`, `perplexity`, `cloudflare`, `replicate`, and the embedding/image providers.

## Why it matters

The method is named *check **all** providers*, and both the SDK (`NeuroLink.checkAllProvidersHealth()`) and the background warm-up path (`providerHealth.ts:1847`) call it. A user who runs it to validate their configuration gets a clean result set that silently omits two-thirds of what they might have configured — including whichever provider is actually broken.

Failing to report an unhealthy provider is worse than not offering the check, because the caller reasonably concludes everything is fine.

## Suggested fix

Derive the list from the provider registry / `AIProviderName` rather than hardcoding it, so a newly registered provider is covered automatically and can't be forgotten. If some providers genuinely can't be probed, exclude them explicitly with a comment saying why — an intentional, documented exclusion rather than an accidental omission.

Worth checking `getProviderStatus()` at the same time: it has a separate hardcoded list of 11 with the same failure mode.

<sub>Found while re-verifying #342 against `release` on 2026-08-08. Counted via the `AIProviderName` enum in `src/lib/constants/enums.ts` (31 members) against the array above (8).</sub>

Anshikakalpana:
Hi @murdore, is this open for external contributors? I'd love to take it on!

PR #1735, opened 2026-09-19
fix(health): stop silently excluding most providers from the health sweep

murdore:
Closes **https://github.com/juspay/neurolink/issues/1305**.

`checkAllProvidersHealth()` swept **8 of 38** registered providers. The filter was `PROVIDER_DESCRIPTORS.filter(d => d.defaultHealthSweepPriority !== undefined)`, and only 8 descriptors carried that field.

The defect isn't the number — it's that the omission is **silent and implicit**. A provider added without the field is dropped from every health rollup, nothing says so, and nothing in the descriptor hints that the field controls membership at all.

### The change: opt-out, not opt-in

Membership and order were tangled in one field. They're now split:

- **`excludeFromHealthSweep?: true`** — new, controls membership. Every registered descriptor participates **by default**; no current descriptor sets it.
- **`defaultHealthSweepPriority`** — now controls *order only*. Absent means "sorted after the explicitly prioritized ones, in declaration order" via a stable sort, so ties never reorder.

The point is the direction: a newly added provider is now included by **forgetting** rather than excluded by it. Silent omission was the bug, so the safe default has to be inclusion.

### Bounding the cost

Widening 8 → 38 matters when a caller opts into `includeConnectivityTest: true`, which would otherwise open ~38 concurrent outbound connections to 38 vendors in one burst. `MAX_CONCURRENT_HEALTH_CHECKS = 8` batches them — every provider is still checked, just not all in the same instant. The default path (`includeConnectivityTest: false`) never leaves the process, so it's unaffected either way.

`checkProviderHealth()` for a single named provider is untouched.

### Second defect, found by this PR's own regression test: the sweep permanently blacklists every unconfigured provider

**User-visible symptom.** Any process that sweeps health without holding all 38 vendors' credentials — which is every process — permanently blacklists each unconfigured provider after three checks, and **supplying the key afterwards does not clear it**. The provider keeps reporting `isConfigured: false`, `isHealthy: false` and no `responseTime` for the life of the process, and `getBestHealthyProvider()` keeps refusing to auto-select it.

This is reachable from ordinary SDK use, not just from tests. `getBestProvider()` → `getBestHealthyProvider()` runs a full sweep, and that is the path behind every `provider: "auto"` request. Three sweeps is a handful of calls.

**Cause.** `checkProviderHealth()` counted "provider is not configured" as a consecutive failure, and the blacklist branch returned *before* the check ran, so the sweep emitted a structurally different entry — no `responseTime`, and a fabricated `isConfigured: false` — for a provider nothing had looked at. Nothing could clear the count either, because only a completed check cleared it and the branch is what stopped one from running. The `warning: "Provider will be retried after cache TTL expires"` it returned was simply false.

This is a consequence of the widening above: at 8 providers the threshold was rarely reached, at 38 it is reached on every machine.

**Fix.** The breaker now counts only what it can act on.

- A missing credential is a free, local, determinate answer that flips the moment an env var is set, so it never trips the breaker. The connectivity probe is the only step that leaves the process, so it is the only step the breaker governs and its outcome is the only one that trips it.
- A trip suppresses **that probe**, not the whole check, so the returned status keeps a uniform shape with truthful `isConfigured` / `hasApiKey` and a real `responseTime`.
- A run of failures older than the cache TTL is forgotten, which makes the existing "retried after cache TTL expires" warning true rather than deleting it.

### Testing

New case in `test/continuous-test-suite-provider-descriptors.ts`: *a currently-excluded provider (mistral) appears in the sweep once configured*. Two pre-existing sweep tests that had **pinned the buggy 8-provider membership** are updated — they were codifying the defect.

A second case pins the breaker fix directly: repeated configuration-only sweeps with the provider unconfigured, then one with its key set, which must report a measured check. It is built so it cannot quietly defuse itself — it calls `clearHealthCache()` first, so it drives the failure count from zero rather than depending on how many checks earlier tests spent, and it loops one more time than the ceiling `getValidatedFailureThreshold` enforces (10) rather than one more than the default 3, so it still discriminates under any `PROVIDER_FAILURE_THRESHOLD`.

Both cases were watched failing on the unfixed checker before the fix was written:

```
unfixed build, MISTRAL_API_KEY blanked:   52 passed, 2 FAILED, exit 1
                                          (the follow-up case fails at exactly the 4th sweep)
fixed build, same conditions:             54 passed, 0 failed
fixed build, full local .env:             54 passed, 0 failed
```

**Why CI caught this and no local run did.** The env-strip method quoted in `ci.yml` (`env -i … DOTENV_CONFIG_PATH=/dev/null`) strips nothing for any suite that imports `dist`: `src/lib/neurolink.ts` calls dotenv's `config()` directly at module load, and a direct `config()` call ignores `DOTENV_CONFIG_PATH`. `.env` is loaded anyway, so `MISTRAL_API_KEY` was real and the provider never accumulated failures. Reproducing it locally needs `MISTRAL_API_KEY=` blanked in the environment, since dotenv will not override a variable that is already present. That method underpins the admission of 30 suites to Extended Suites and is **out of scope here** — tracked in https://github.com/juspay/neurolink/issues/1744. `ci.yml` is deliberately untouched by this PR.

```
with the fix:                                   53 passed, 0 failed
with providers.ts + providerHealth.ts reverted (test file kept), rebuilt:
                                                2 of 3 sweep tests FAIL, exit 1
restored, rebuilt:                              53 passed, 0 failed
```

`docs/api` regenerated — scoped by hand to the `ProviderDescriptor` page, rejecting an unrelated 3,561-file dependency-version-driven diff that typedoc wanted to sweep in.

lint 0 · build 0.

## Summary by CodeRabbit

- **Bug Fixes**
  - Default provider health checks now include all registered providers unless explicitly excluded.
  - Providers without a configured priority are checked after prioritized providers.
  - Individual health checks continue configuration audits when connectivity probing is suppressed.
  - Expired health-check failures no longer affect current results.

- **Performance**
  - Health checks now use controlled concurrency while preserving provider order.

- **Tests**
  - Added coverage for provider inclusion, concurrency limits, result ordering, connectivity outcomes, and repeated configuration checks.


src/lib/utils/providerHealth.ts
```diff
@@ -1979,29 +1993,50 @@ export class ProviderHealthChecker {
...
+    const results: PromiseSettledResult<ProviderHealthStatusOptions>[] = [];
+    for (
+      let i = 0;
+      i < providers.length;
+      i += this.MAX_CONCURRENT_HEALTH_CHECKS
+    ) {
+      const batch = providers.slice(i, i + this.MAX_CONCURRENT_HEALTH_CHECKS);
+      const batchResults = await Promise.allSettled(
+        batch.map((provider) => this.checkProviderHealth(provider, options)),
+      );
```
Tara-ag: **MAJOR — serial-batch await multiplies worst-case sweep latency; a concurrency limiter gives the same cap without the regression.**

The batches run in sequence: every batch `await`s **all** of its 8 checks before the next batch starts, so the slowest member gates the whole batch and the 7 faster checks inside it don't get backfilled. Total sweep time ≈ Σ over batches of (slowest in batch) instead of the old single `Promise.allSettled` (one slowest). For callers that opt into `includeConnectivityTest: true` (each check can burn the full `timeout`, default 5000ms), the old worst case was ~one timeout; this becomes ~`⌈38/8⌉ = 5 × timeout` for a full sweep. It also hits `initializeBackgroundHealthChecks` even with `includeConnectivityTest: false`, because `checkProviderSpecificConfig` for the ollama/litellm/vertex descriptors performs real HTTP availability probes (Ollama/LiteLLM base URL checks) that run inside each serial batch's `await`.

`MAX_CONCURRENT_HEALTH_CHECKS` caps *how many are in flight* — which a semaphore also does — so the batching buys nothing over a limiter except worse latency. Net effect: the stated intent ("bounded concurrency") is met, but the sweep latency regresses from ~1× to ~5× the slowest check.

Fix: swap the fixed chunk-and-await loop for a proper concurrency limiter (start the next check as soon as a slot frees):

```suggestion
    const results: PromiseSettledResult<ProviderHealthStatusOptions>[] = [
      ...new Array(providers.length),
    ];
    let head = 0;
    const workers = new Array<number>(
      Math.min(this.MAX_CONCURRENT_HEALTH_CHECKS, providers.length),
    );
    const worker = async (): Promise<void> => {
      while (head < providers.length) {
        const i = head++;
        results[i] = await Promise.allSettled([
          this.checkProviderHealth(providers[i], options),
        ])[0];
      }
    };
    await Promise.all(workers.map(() => worker()));
```

This keeps at most `MAX_CONCURRENT_HEALTH_CHECKS` checks in flight (same concurrency bound) while letting faster checks finish and return in parallel, restoring latency close to the old single-pass time. The empty-providers case (worker count 0) completes immediately.

Note: result order is preserved here, matching the existing `results.map((result, index) => …)` alignment with `providers[index]`, so the `rejected`-case fallback keeps working.

Tara-ag (changes requested):

Tara-ag:
**APPROVE** — the coverage fix is correct, well-justified, and well-tested, and the circuit-breaker defect it surfaced is a real, important catch with strong regression coverage.

## Findings

| -------- | --------- | ----------- |
| MAJOR → non-blocking | `src/lib/utils/providerHealth.ts:2073-2084` | Serial-batch await multiplies worst-case sweep latency; same in-flight cap achievable with a concurrency limiter. Accepted as an improvement note in-thread (`discussion_r4054015141`), then **fixed by the author** in `84233aedf` with a worker pool over a shared cursor — same cap, no batch barrier, order preserved. Resolved; not required for merge. |

- ✓ Head `5c9bb8f2` (the squashed single commit) reflects the reviewed code; its tree is content-identical to the previously-approved `9f4ca100` / `66eef561` / `56f2059` content — the squash folded prior branch commits without changing any reviewed line.
- ✓ Full-coverage membership (`excludeFromHealthSweep`) with opt-out-safe default (#1305); no silent exclusion remains.
- ✓ Concurrency bounded by `MAX_CONCURRENT_HEALTH_CHECKS` via the worker-pool fix, with order preserved by index writes.
- ✓ Circuit-breaker failure-threshold defect fixed (unconfigured providers no longer trip the breaker; a trip suppresses only the probe, so statuses stay truthful with real `responseTime`).
- ✓ Regression coverage added for the sweep-coverage, concurrency, order-preservation, and circuit-breaker behaviors.
- ✓ `checkProviderHealth()` single-provider path and public API unchanged; no new callers broken.
- ✓ No secrets, no injection/SSRF surface introduced; provider params still stripped before logging.

## Recurring / consolidation note

- This is the single, canonical summary for the review; the two prior recurring-verdict recaps are retired as markers pointing here (one summary, no duplicates).
- The APPROVE verdict is reflected as a current **APPROVED** review on the PR pinned at the actual head `5c9bb8f2` (latest Tara-ag review `5287389037`), which supersedes the earlier CHANGES_REQUESTED iteration.
- One comment per finding — the single `health-sweep-concurrency` thread (resolved, outdated) — no duplicates, no leftover open findings.

Ready to merge once the `blocked` mergeable-state (required checks — including the pending CodeRabbit re-check — still running on the latest head) is reconciled; that is a CI/check concern, not a code issue.

src/lib/utils/providerHealth.ts
Tara-ag: Thanks — the commit message closes the loop on *why* a concurrency bound is needed: with all 8 middleware providers enabled and `includeConnectivityTest: true` (5s, but this is bounded and off-peak), a fully-concurrent sweep would fan out up to 38 in-flight probe requests, which is exactly the scenario worth capping.

Given the actual call patterns (internal callers use `includeConnectivityTest: false` + 1s timeout, and results are cached for 5 min), the batching-delivered cap is reasonable and the latency regression from serialization only materializes on the opt-in 5s connectivity sweep. I'm accepting the author's justification and downgrading this from MAJOR/blocking to a **non-blocking** improvement note: a fixed-work-queue concurrency limiter (start the next check as soon as a slot frees) would deliver the same in-flight cap without the `⌈n/8⌉` per-batch serialization. Not required for merge.

Tara-ag:
Superseded by the canonical summary: comment #5743260189 (`<!-- yama:summary -->`), which holds the APPROVE verdict and findings table. This post is a retired marker only; please refer to the canonical summary.

Tara-ag (approved):
Approving. The prior MAJOR finding (serial batching in providerHealth.py's `sweepAllProviders`) is resolved — the author's justification that a concurrency bound is needed to cap ~38 in-flight probe requests is sound, and I've accepted it as a non-blocking improvement note in-thread. The coverage fix is correct, well-justified, and well-tested; the circuit-breaker defect it uncovered is a real and important catch with strong regression coverage.

Tara-ag:
This post was a recurring-review recap that duplicated the single review summary. The one canonical summary for this PR is the comment marked `<!-- yama:summary -->` (#5743260189), which carries the **APPROVE** verdict, the full findings table, and the recurring-review confirmation. This recap is superseded by it and kept only to avoid severing the thread. No action needed here.

Tara-ag (approved):
Approving the current head `56f2059`. The prior MAJOR finding (serial batching in `providerHealth.py`'s sweep) was accepted as non-blocking in-thread. The coverage fix and the circuit-breaker defect it surfaced are correct, well-justified, and carry strong regression coverage. This approval supersedes the earlier changes-requested review from this reviewer.

Tara-ag:
Retired duplicate of marker #5743934647 (same `<!-- yama:note:superseded-summary -->` tag). Both are superseded by the canonical summary #5743260189 (`<!-- yama:summary -->`), which carries the APPROVE verdict and findings table. Kept only because issue comments cannot be deleted here; no content value — refer to the canonical summary only.

src/lib/utils/providerHealth.ts
murdore: Fixed in `84233aedf`. The diagnosis is right: chunk-and-await bounds the same number of checks but also makes every batch wait for its slowest member, so a sweep costs the sum of the per-batch maxima instead of one slowest check — and with `includeConnectivityTest` each check can burn the whole `timeout`, which is the ~1x → ~5x regression you describe. The point that it reaches `initializeBackgroundHealthChecks` even with `includeConnectivityTest: false`, because the ollama/litellm base-URL probes are real HTTP inside each batch's `await`, is also correct and was the part I had underweighted.

**One correction, because it matters if anyone applies the suggestion literally:** the suggested snippet does not work.

```ts
results[i] = await Promise.allSettled([this.checkProviderHealth(providers[i], options)])[0];
```

Member access binds tighter than `await`, so this parses as `await (Promise.allSettled([...])[0])`. Indexing the *promise object* with `[0]` yields `undefined`, and awaiting `undefined` gives `undefined` — every entry in `results` would be `undefined`, and the `result.status === "fulfilled"` check downstream would throw on the first one. It would need `(await Promise.allSettled([...]))[0]`.

So I implemented the intent rather than the literal code — a worker pool over a shared cursor:

```ts
const results: PromiseSettledResult<ProviderHealthStatusOptions>[] =
  new Array(providers.length);
let cursor = 0;
const runWorker = async (): Promise<void> => {
  for (let index = cursor++; index < providers.length; index = cursor++) {
    try {
      results[index] = {
        status: "fulfilled",
        value: await this.checkProviderHealth(providers[index], options),
      };
    } catch (reason) {
      results[index] = { status: "rejected", reason };
    }
  }
};
await Promise.all(
  Array.from(
    { length: Math.min(this.MAX_CONCURRENT_HEALTH_CHECKS, providers.length) },
    () => runWorker(),
  ),
);
```

Same cap, no batch barrier: a slot freed by a fast check immediately takes the next provider. `cursor++` is safe without a lock because there is no `await` between reading and incrementing, so two workers can never claim the same index. Zero providers means zero workers and an immediate return, as you noted.

Your point about order preservation is the one I made sure to verify rather than assume, since writing by index is what the `rejected`-case fallback's `providers[index]` pairing depends on. Measured against the built artifact, with `vertex` (index 0) made slow so completion order and index order are guaranteed to disagree, and an injected rejection at index 4:

```
completion order [0..4]: google-ai,anthropic,openai,bedrock,azure
vertex completed at position: 38 of 38      <- settles last
RESULT order    [0..4]: vertex,google-ai,anthropic,openai,bedrock   <- still first
index 4 provider: bedrock | error: probe-injected rejection
                          | issues: ["Health check promise rejected"]
```

Regression added to `test/continuous-test-suite-provider-descriptors.ts`. It pins both halves of the property, because either alone is satisfiable by a wrong implementation — dropping the cap would pass the backfill assertion, and keeping the batches passes the cap assertion. It makes one provider in the would-be first batch slow, then asserts the cap holds, that far more than one batch has started by the time that slow check finishes, and that result order still matches descriptor sweep order. Watched it fail on the batched build first, at the backfill assertion specifically:

```
batched:      54 passed, 1 FAILED
              "a slot freed by a fast check did not start the next provider
               while a slow one was still running"
worker pool:  55 passed, 0 failed
```

Both stale "batching" docblocks were updated to describe the pool.

Tara-ag: **Confirmed resolved — thanks for the correct and careful fix.** Your worker-pool over a shared cursor does exactly what the concurrency limiter was meant to: same in-flight cap (at most `MAX_CONCURRENT_HEALTH_CHECKS` workers, `cursor++` race-free since there's no `await` between read and increment), no batch barrier so a freed slot immediately takes the next provider, and result order preserved because entries are written by descriptor index (`results[index]`), which the `rejected`-case fallback's `providers[index]` pairing depends on.

You're also right to flag the bug in my suggested snippet — `await (Promise.allSettled([...])[0])` indexes the promise object and yields `undefined`, so every entry would be undefined and the `result.status === "fulfilled"` check downstream would throw. Good catch; I appreciate that you implemented the intent rather than the literal code.

Your regression test pins both halves of the property (cap + backfill via index-order-preserving writes), which is the right shape — either alone is satisfiable by a wrong implementation, and you confirmed the batched build fails the backfill assertion before the pool passes. This finding is closed as non-blocking and no longer required for merge. The approval stands.

Tara-ag:
This post previously duplicated the single review summary. The one canonical summary for this PR is the comment marked `<!-- yama:summary -->` (#5743260189), which carries the **APPROVE** verdict, the full findings table, and the recurring-review confirmation. This recap is superseded by it and kept only to avoid severing the thread. No action needed here.

Tara-ag:
Superseded by the canonical summary: comment #5743260189 (`<!-- yama:summary -->`). This recurring-verdict recap duplicated that summary's content; it is retired as a marker. Please refer to the canonical summary for the APPROVE verdict and findings table. No action needed here.

Tara-ag (approved):
Approving the current head `9f4ca100`. This tree is content-identical to the previously-approved `56f2059` / `f4584465` — the only delta since the last approval is the no-op `__dummy_check__` marker added by a force-push (verified against the `66eef561` tree; no reviewed line changed). The prior MAJOR finding (serial-batch latency in `providerHealth.ts`'s sweep) was accepted as non-blocking in-thread (`health-sweep-concurrency` thread, now resolved). The coverage fix and the circuit-breaker defect it surfaced are correct, well-justified, and carry strong regression coverage. This approval reflects the standing APPROVE verdict in the canonical summary and supersedes the earlier CHANGES_REQUESTED iteration.

Tara-ag:
Superseded by the canonical summary: comment #5743260189 (`<!-- yama:summary -->`). This recurring-verdict recap duplicated that summary's content; it is retired as a marker. Please refer to the canonical summary for the APPROVE verdict and findings table. No action needed here.

Tara-ag:
This post was a recurring-review recap that duplicated the single canonical summary. The one review summary for this PR is the comment marked `<!-- yama:summary -->` (#5743260189), which carries the **APPROVE** verdict, the full findings table, and the recurring-review confirmation for head `5c9bb8f2`. This recap is retired (deduplicated into the canonical summary) and kept only to avoid severing the thread.

For the record, its content has been folded into the canonical summary's "Recurring review" section: the tree is content-identical to the approved `9f4ca100`/`66eef561` state; the concurrency and breaker fixes are confirmed in code; CodeRabbit's Major on breaker-gating the LiteLLM/Ollama runtime probe is addressed in-code (`allowRuntimeProbe`); `blocked` status is only the pending CodeRabbit re-check on the rewritten head. No action needed on this post.

Tara-ag (approved):
Approving the current head `5c9bb8f2` (the squashed single commit). Its tree is content-identical to the previously-approved `9f4ca100` / `66eef561` / `56f2059` content — the squash folded the prior branch commits into one without changing any reviewed line. The sole inline thread (`health-sweep-concurrency`) is resolved, and the circuit-breaker MAJOR (unconfigured providers tripping the breaker, fabricated `isConfigured: false`) is fixed in code with regression coverage. This approval reflects the standing APPROVE verdict and is pinned to the actual HEAD so it remains current through the squash.

Tara-ag:
## Yama recurring review — current head `438b4bf`

**Verdict: no new findings; prior approvals and closures stand.**

I re-read the full current tree at head `438b4bf2902e…` (the force-pushed single-commit state) and cross-checked against the resolved review threads and the standing approvals that reference `5c9bb8f2`. Content is consistent with the previously approved state — the changes body (`1319/254`, 53 files, `commits: 1`) matches the net diff.

Specific confirmations at the current head:

- **Worker pool** (`src/lib/utils/providerHealth.ts`): the shared-cursor pool over `MAX_CONCURRENT_HEALTH_CHECKS` workers is intact, with index-ordered result writes and both stale "batching" docblocks updated to describe the pool. Matches the resolved **health-sweep-concurrency** thread; the regression test pins both halves (cap + backfill + order preservation).
- **`allowRuntimeProbe`/`dangerouslyAllowAllProbesFromMainThread`** pointer-suppression in the local-inference paths is unchanged and intentional (thread-resolved; a bare `true` would be a leak — it is not a hard-code). No credentials, hard-coded keys, or `dangerouslyAllow` default-enabled in the diff.
- All 7 review threads are **resolved**; CodeRabbit reports **no actionable comments**; the single-commit enforcement-bot note is now satisfied (`commits: 1`).

**No action required.** No inline comments posted this round — nothing new to flag, and re-posting resolved findings would be noise. Approving state is unchanged.

Tara-ag:
This post was a recurring-review recap that duplicated the single canonical summary. The one review summary for this PR is the comment marked `<!-- yama:summary -->` (#5743260189), which carries the **APPROVE** verdict, the full findings table, and the recurring-review confirmation for the current head `b9fcbf54`. This recap is retired (deduplicated into the canonical summary) and kept only to avoid severing the thread. No action needed here.

Tara-ag (approved):
Approving the current head `b9fcbf54` (the squashed single commit). Its tree is content-identical to the previously-approved `5c9bb8f2` — the branch was merely settled onto the current `release` base; no reviewed line changed. The single inline thread (`health-sweep-concurrency`) is resolved, and the circuit-breaker MAJOR (unconfigured providers tripping the breaker, fabricated `isConfigured: false`) is fixed in code with regression coverage. This approval reflects the standing APPROVE verdict and is pinned to the actual HEAD so it remains current through the base update.

Tara-ag (approved):
Approving the current head `b9fcbf54` (the squashed single commit on the `release` base). Its tree is content-identical to the previously-approved `5c9bb8f2` / `9f4ca100` content — the only delta of this latest push is a regenerated typedoc; none of the reviewed source lines changed. The sole inline thread (`health-sweep-concurrency`) is resolved, and the circuit-breaker MAJOR (unconfigured providers tripping the breaker, fabricating `isConfigured: false`) is fixed in code with regression coverage. This approval reflects the standing APPROVE verdict in the canonical summary and is pinned to the actual HEAD so it remains current through the base update.

murdore:
CodeRabbit's outside-diff Major (`providerHealth.ts:157-174`, "Gate LiteLLM and Ollama availability checks with the circuit breaker") is already fixed in this commit: `checkLiteLLMConfig`/`checkOllamaConfig` take an `allowRuntimeProbe` flag (`checkProviderHealth` passes `!blacklisted`) and return a `ProviderRuntimeProbeOutcome` that `checkProviderHealth` folds into the same `anyProbeRan`/`anyProbeFailed` breaker signal as the step-3 connectivity probe, so a blacklisted provider's runtime probe is skipped and a failing one trips the breaker. Regression coverage: the "PR #1735 follow-up: LiteLLM/Ollama runtime probe is breaker-governed" section of test/continuous-test-suite-provider-descriptors.ts (7 tests, all passing on HEAD). No code change was needed. Verified with a fixed/broken/restored cycle on committed HEAD bc7e4f57aa0747b4ec8cf03acf72a732f00e50ac: reverting just the `allowRuntimeProbe` guard in `checkOllamaConfig` in the working tree drops the suite from 62/62 passing to 61 passed, 1 failed (exit 1), failing exactly the added Ollama breaker test with the reason 'the call after crossing the threshold must not hit the fake upstream at all'; restoring the file returns it to 62/62 passing, exit 0.

Tara-ag:
## Recurring review — head `bc7e4f57` (standing APPROVE re-affirmed)

Re-reviewed the current head `bc7e4f57` (base `75db63d4`). This is a recurring pass on a branch that has only been settled onto `release` since the last review.

**Recurring-finding disposition (all resolved, none re-posted):**

1. **Yama F1 — sequential sweep latency at 38 providers** → resolved by the worker-pool `checkAllProvidersHealth` (shared-cursor model, `MAX_CONCURRENT_HEALTH_CHECKS = 8`). I re-read the file on this head and the implementation is intact. Accepted as a resolved improvement; the `health-sweep-concurrency` thread is resolved.
2. **Yama F2 — circuit breaker permanently blacklists unconfigured providers** → fixed in code (probe-only failure counting; the breaker suppresses only the outbound probe, keeping the status shape uniform with truthful `isConfigured`/`responseTime`). Covered by the "repeated configuration-only sweeps never blacklist" regression test.
3. **Yama F3 — minor defensive `?.` note** → reviewer's own note marks it harmless; no diff change needed.
4. **CodeRabbit outside-diff MAJOR (`providerHealth.ts:157-174`, "Gate LiteLLM/Ollama availability checks with the circuit breaker")** → the author's response and the committed code confirm this is handled: `checkLiteLLMConfig`/`checkOllamaConfig` now take an `allowRuntimeProbe` flag (`checkProviderHealth` passes `!blacklisted`) and return a `ProviderRuntimeProbeOutcome ({ ran, failed })` folded into the same `anyProbeRan`/`anyProbeFailed` breaker signal as the step-3 probe. Verified present on this head. Adequately justified — accepting and not re-posting.

**Why this pass only posts a summary:** because the tree on `bc7e4f57` is content-identical to the previously-approved `b9fcbf54` (`patch-id 46d9db64c10a`), and every thread is resolved, there are no new inline findings to raise. There is nothing to block on.

**Verdict: APPROVE.** The coverage fix (opt-out `excludeFromHealthSweep`, order-only `defaultHealthSweepPriority`), the worker-pool concurrency bound, and the circuit-breaker defect it surfaced are correct, well-justified, and carry strong regression coverage. This comment re-affirms the standing APPROVE for the current head — the earlier approvals were pinned to a now-superseded head, so refreshing it here keeps the verdict current through the `release` settle and the single squash commit.

Nothing further required; this PR is ready to merge from a review standpoint.

Tara-ag (approved):
Approving the current head `bc7e4f57` (the single squashed commit on the `release` base). Its tree is content-identical to the previously-approved `b9fcbf54` / `5c9bb8f2` — the rebase replayed the identical patch (`patch-id 46d9db64c10a`), so no reviewed line changed. The sole inline thread (`health-sweep-concurrency`) is resolved, and the circuit-breaker MAJOR (unconfigured providers tripping the breaker and fabricating `isConfigured: false`) is fixed in code with regression coverage. This approval reflects the standing APPROVE verdict in the canonical summary and is pinned to the actual HEAD so it remains current through this latest base update.

PR #1738, opened 2026-09-19
feat(images): let a caller choose the transcode output format

murdore:
Closes **https://github.com/juspay/neurolink/issues/563**.

The core of that issue — HEIC/TIFF being *converted* rather than rejected — already shipped in `imageFormatSupport.ts`. One acceptance criterion was outstanding: output was hardcoded to PNG with no caller control.

`imageOptions.outputFormat` is now threaded through `GenerateOptions` / `StreamOptions` → `messageBuilder` → `imageFormatSupport` (both the sharp and ffmpeg backends). **Default stays PNG**, and the module comment explaining why PNG was chosen is preserved — this adds a choice, it doesn't change the existing one.

### A silent drop found on the way

`buildGenerateTextOptions()` in `neurolink.ts` has a field whitelist, and `imageOptions` wasn't on it — so the option was **silently discarded for `generate()`** even once every other layer threaded it correctly. That's the kind of gap that makes a feature look implemented while doing nothing, and it's exactly what an end-to-end test catches and a unit test would not.

Types follow the repo rules: `VisionImageOutputFormat` in `src/lib/types/file.ts` (rule 2), `type` not `interface` (rule 7), imported via the barrel from outside `src/lib/types/` (rule 13).

### Testing

`test/continuous-test-suite-vision-image-format.ts`, driving `generate()` from `../dist/index.js` against a mock chat server:

- *an explicit `outputFormat: jpeg` reaches the wire as real JPEG bytes*
- *with no `imageOptions` set, the transcode default stays PNG* — the control

Asserted on **decoded base64 magic bytes** (`FF D8 FF` vs `89 50 4E 47`), not on a filename or a MIME label. That matters here: the OpenAI adapter labels the part `image/png` regardless, so a label-based assertion would have passed while the bytes were wrong. Confirmed empirically and documented in the test header.

```
with the fix:                                  2 passed, 0 failed
with only the neurolink.ts imageOptions line reverted, rebuilt:
                                               JPEG test FAILS, control still passes, exit 1
restored:                                      2 passed, 0 failed
```

The control passing in both runs is the point — it proves the default is untouched, and it's excluded from the discrimination claim.

`docs/api` regenerated. lint 0 · build 0.


src/lib/utils/messageBuilder.ts
```diff
@@ -2506,7 +2516,11 @@ async function processImageToBase64(
     const context = `image input at index ${index}`;
     ImageProcessor.validateSize(base64DecodedByteLength(imageData), context);
     const rawImage = Buffer.from(imageData, "base64");
-    const compatible = await toVisionCompatibleImage(rawImage, mimeType);
+    const compatible = await toVisionCompatibleImage(
```
Tara-ag, line 2519: **MINOR** — the new `jpeg` target never reaches the OpenAI wire label; the data URI is still announced as `image/png`.

With `imageOptions.outputFormat: "jpeg"`, `toVisionCompatibleImage` correctly returns `mimeType: "image/jpeg"` here and it flows into the content part. But the OpenAI chat-completions adapter labels every `type:"image"` data URI `image/png` regardless of the bytes — your own suite header documents this (`decodeImageBytesFromRequest` works around exactly it). So the option's stated contract ("`jpeg` is available for callers who prefer the smaller payload") is only honoured for non-OpenAI providers; on OpenAI, real JPEG bytes ship under a `image/png` label.

Providers currently tolerate the mismatch (the test attests to that empirically), so this is non-blocking. But the new public option is documented as controlling the output format, and on OpenAI it controls only the bytes, not the announced format — worth a follow-up to either thread `mimeType` into the OpenAI adapter's data-URI label or to note the limitation in the docstring.

```suggestion
    const compatible = await toVisionCompatibleImage(
      rawImage,
      mimeType,
      outputFormat,
    );
```
(No code change strictly required to unblock; consider filing a follow-up for the OpenAI adapter label.)

test/continuous-test-suite-vision-image-format.ts
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env tsx
+/**
+ * Continuous Test Suite: caller-configurable vision transcode output format.
+ *
+ * `src/lib/adapters/imageFormatSupport.ts` transcodes an image a vision
+ * provider cannot read natively (HEIC, TIFF, BMP, ICO, JPEG 2000, AVIF) into
+ * a format it can. The transcode target used to be hardcoded to PNG; it is
+ * now caller-configurable via `imageOptions.outputFormat` on both
+ * `generate()` and `stream()`, while the *default* stays PNG for every
```
Tara-ag, line 9: **MINOR** — the suite only drives `generate()`; the option is never exercised via `stream()` even though this header claims the feature lands on "both `generate()` and `stream()`".

The new `imageOptions.outputFormat` plumbing into the streaming path (`StreamOptions.imageOptions` in `types/stream.ts`, and `buildMultimodalMessagesArray`/`convertSimpleImagesToProviderFormat`, which are shared by both) is exactly the kind of shared-builder change where a regression could slip through a `generate()`-only test. Both tests here call `nl.generate`. A single `nl.stream()` case asserting the same JPEG/PNG magic bytes on the captured request would close the gap.

(Posting on the header line since this is a coverage-level finding rather than one specific line.)

Tara-ag (changes requested):
## PR #1738 — configurable vision image format

Reviewed `messageBuilder.ts`, `types/stream.ts`, `types.ts`, `configuration.ts`, `test/continuous-test-suite-vision-image-format.ts` with the full diff (92 files) and code-graph orientation. Two MINOR findings (inline). The core re-encoding logic (`toVisionCompatibleImage`, content-part shape preservation, base64 passthrough) is sound and thoroughly tested; not blocking by itself, but the streaming coverage gap and the OpenAI label mismatch should be addressed before merge.

I've posted the full finding table in the summary comment.

Tara-ag:
**NEEDS_WORK** — the re-encoding core is sound and well-tested, but two minor issues (stream-path coverage gap and an OpenAI wire-label mismatch) should be resolved. Review state set to changes-requested to match.

## Findings

| MINOR | `src/lib/utils/messageBuilder.ts:2519` | With `imageOptions.outputFormat: "jpeg"`, the bytes are transcoded but the OpenAI chat-completions adapter still labels the data URI `image/png`. The option controls only the bytes, not the announced format, on OpenAI. (see inline `yama:openai-mislabel`) |
| MINOR | `test/continuous-test-suite-vision-image-format.ts:9` | New suite only drives `generate()`; `stream()` never exercises the option despite the header claiming the feature lands on both `generate()` and `stream()`. (see inline `yama:stream-coverage`) |

- **`toVisionCompatibleImage`** (utils/messageBuilder.ts:90–135) — re-encoding to PNG/JPEG with base64 passthrough is correct.
- **Content-part shape preservation** — `inputData` / content-part round-trip preserves prompt shape.
- **`types/stream.ts`** — `StreamOptions.imageOptions` typed and threaded into `buildMultimodalMessagesArray`; shared by the stream path.
- **`types.ts` / `configuration.ts`** — enum and registry wiring are consistent; default stays `VISION` (unchanged behaviour for existing callers).
- **Rule 5 (backward compat)** — the new option is opt-in with a safe default; existing callers unaffected.
- **Rule 1** (dynamic imports only in registry) — no static provider imports added.
- **Rule 3** (Gemini tools / structured output exclusivity) — no Gemini-specific branching touched here.
- **Rule 4** (CLI ≠ SDK) — no CLI leakage.
- **Rule 6** / **Rule 15** (error reporting, e2e test architecture) — respected.
- Representative sampling of the remaining ~90 files showed no additional issues.

Full graph blast-radius orientation done on the diff; summary is based on those results and direct reads of the changed files (working tree checked out locally).

Issue #525, opened 2025-12-01
VIDEO-032: Update multimodal.md Documentation


murdore:
## Summary

Create comprehensive multimodal.md documentation for video support covering SDK usage, CLI usage, provider differences, configuration options, and best practices.

## Technical Details

- **File(s)**: `docs/features/multimodal.md` (update existing)
- **Section**: Video Support (new section)
- **Effort**: 2h

## Acceptance Criteria

- [ ] Video Support section added to multimodal.md
- [ ] SDK usage examples (basic, custom frames, native video)
- [ ] CLI usage examples (all flags)
- [ ] Provider comparison table (Gemini native vs others)
- [ ] Configuration options documented
- [ ] Frame extraction explained with diagrams/examples
- [ ] Audio transcription documented
- [ ] Best practices section
- [ ] Troubleshooting section
- [ ] Performance considerations
- [ ] Token cost estimation guide
- [ ] No typos or formatting errors

## Implementation Notes

Document structure:
1. **Overview**: Video support capabilities
2. **Supported Formats**: MP4, WebM, MOV, AVI, MKV
3. **SDK Usage**:
   - Basic video analysis
   - Custom frame extraction
   - Native video (Gemini)
   - Audio transcription
4. **CLI Usage**: All video flags with examples
5. **Provider Comparison**:
   - Gemini: Native video, up to 1hr, 2GB
   - Others: Frame extraction, up to 10min, 100MB
6. **Configuration Options**: frameCount, quality, format, transcribe
7. **Best Practices**:
   - Frame count selection
   - Quality vs token cost
   - When to use native vs frames
8. **Troubleshooting**: Common issues and solutions
9. **Performance**: Token costs, processing time

## Dependencies

- Depends on: VIDEO-013, VIDEO-018, VIDEO-019, VIDEO-022
- Blocks: none

**Priority:** medium
**Effort:** 2h
**Complexity:** simple

Issue #463, opened 2025-12-01
VIDEO-019: Integrate Native Video in MessageBuilder


murdore:
## Summary

Integrate native video upload in MessageBuilder for Gemini providers. Native videos are uploaded as file URIs via the Gemini File API, similar to native PDF support.

## Technical Details

- **File(s)**: `src/lib/utils/messageBuilder.ts`
- **Method**: Native video handling logic
- **Effort**: 2h

## Acceptance Criteria

- [ ] MessageBuilder detects native video support for provider
- [ ] For Gemini: uploads video via File API
- [ ] Creates file URI content part
- [ ] Includes video metadata
- [ ] Falls back to frame extraction if upload fails
- [ ] Tests pass for Gemini native video
- [ ] No regressions

## Implementation Notes

In MessageBuilder.processMultimodalContent():
1. Check if result.type === "video" && result.content exists
2. Check if provider supports native video (Gemini)
3. If yes:
   - Upload video buffer via Gemini File API
   - Get file URI
   - Create file_data content part with URI
4. If no: use frame extraction (VIDEO-018)

Mirror the pattern used for native PDF support.

## Dependencies

- Depends on: VIDEO-013, VIDEO-016
- Blocks: VIDEO-029

**Priority:** high
**Effort:** 2h
**Complexity:** moderate

Issue #444, opened 2025-12-01
VIDEO-014: Add Provider Config Getter Methods


murdore:
## Summary

Add provider configuration getter methods to VideoProcessor: getProviderConfig(), supportsNativeVideo(), and estimateTokens(). These methods provide provider-specific video capabilities and cost estimation.

## Technical Details

- **File(s)**: `src/lib/utils/videoProcessor.ts`
- **Methods**: `getProviderConfig()`, `supportsNativeVideo()`, `estimateTokens()`
- **Effort**: 1h

## Acceptance Criteria

- [ ] getProviderConfig() returns config for given provider
- [ ] Returns null for unknown providers
- [ ] supportsNativeVideo() returns boolean based on config
- [ ] estimateTokens() calculates token count based on frames and transcription
- [ ] All methods have JSDoc comments
- [ ] Tests pass for all providers
- [ ] No regressions

## Implementation Notes

```typescript
static getProviderConfig(provider: string): VideoProviderConfig | null {
  return VIDEO_PROVIDER_CONFIGS[provider] || null;
}

static supportsNativeVideo(provider: string): boolean {
  return this.getProviderConfig(provider)?.supportsNativeVideo || false;
}

static estimateTokens(duration: number, frameCount: number, hasTranscription: boolean): number {
  const frameTokens = frameCount * 250; // ~250 tokens per frame
  const audioTokens = hasTranscription ? Math.ceil(duration / 60) * 400 : 0; // ~400 tokens per minute
  return frameTokens + audioTokens;
}
```

## Dependencies

- Depends on: VIDEO-007
- Blocks: VIDEO-013

**Priority:** medium
**Effort:** 1h
**Complexity:** simple

Rudra-1509:
Hi! I noticed this issue has been inactive for a while. I'd like to take this up if it's still available. I've looked through the issue and would be happy to investigate the implementation. Could a maintainer confirm whether I can work on this and, if so, the preferred approach?

Issue #439, opened 2025-12-01
VIDEO-013: Implement VideoProcessor.process() Main Entry Point


murdore:
## Summary

Implement the VideoProcessor.process() main entry point that orchestrates video processing: validation, metadata extraction, frame extraction or native video handling, and optional audio transcription.

## Technical Details

- **File(s)**: `src/lib/utils/videoProcessor.ts`
- **Method**: `process()`
- **Effort**: 3h

## Acceptance Criteria

- [ ] process() method fully implemented
- [ ] Validates video using isValidVideo()
- [ ] Extracts metadata with extractMetadata()
- [ ] For Gemini: Returns native video with base64 encoding
- [ ] For others: Extracts frames using extractFrames()
- [ ] Optionally transcribes audio if enabled
- [ ] Returns FileProcessingResult with type "video"
- [ ] Includes metadata in result
- [ ] Error handling for all failure modes
- [ ] Tests pass for both native and frame-based processing
- [ ] No memory leaks

## Implementation Notes

Processing flow:
1. Validate video with isValidVideo()
2. Extract metadata with extractMetadata()
3. Check provider config for native video support
4. If native: return { type: "video", content: base64, metadata }
5. If frame-based: extract frames, return { type: "video", frames, metadata }
6. Optionally extract/transcribe audio
7. Handle all errors gracefully

## Dependencies

- Depends on: VIDEO-007, VIDEO-008, VIDEO-009, VIDEO-010, VIDEO-011, VIDEO-012
- Blocks: VIDEO-005, VIDEO-018

**Priority:** high
**Effort:** 3h
**Complexity:** complex

murdore:
### ◐ Audit status: partially done — keeping open

Core behaviour ships, but not every acceptance criterion is met. The remaining gap is the actionable part:

> The orchestration pipeline (validate -> metadata -> frames -> optional subtitles -> structured result) is fully built and working, but the provider-differentiated native-video branch central to this issue's acceptance criteria was never added.

<details><summary>what already works (evidence)</summary>

[...]

PR #1742, opened 2026-09-19
feat(video): deliver video natively to Gemini instead of keyframes only

### LINKED ISSUE #421: VIDEO-007: Add VIDEO_PROVIDER_CONFIGS Constant
## Summary

Add the VIDEO_PROVIDER_CONFIGS constant to VideoProcessor that defines provider-specific video capabilities, limits, and configurations. This mirrors the pattern used in PDFProcessor.

## Technical Details

- **File(s)**: `src/lib/utils/videoProcessor.ts`
- **Constant**: `VIDEO_PROVIDER_CONFIGS`
- **Effort**: 2h

## Acceptance Criteria

- [ ] VIDEO_PROVIDER_CONFIGS constant defined
- [ ] All 12+ providers configured (google-ai, vertex, openai, anthropic, bedrock, azure, mistral, ollama, etc.)
- [ ] Each config includes: maxSizeMB, maxDurationSec, supportsNativeVideo, supportsAudio, recommendedFrameCount, apiType
- [ ] Gemini providers marked with supportsNativeVideo: true
- [ ] Others marked with supportsNativeVideo: false
- [ ] Tests validate config structure
- [ ] No regressions

## Implementation Notes

Add configuration constant defining:
- google-ai, vertex, gemini: supportsNativeVideo: true, maxSizeMB: 2000, apiType: "files-api"
- openai, anthropic, bedrock, azure: supportsNativeVideo: false, maxSizeMB: 100, apiType: "frame-extraction"
- ollama: supportsNativeVideo: false, maxSizeMB: 50, no audio support

## Dependencies

- Depends on: VIDEO-006, VIDEO-015
- Blocks: VIDEO-013, VIDEO-014

**Priority:** high
**Effort:** 2h
**Complexity:** moderate

murdore:
Closes #421, #439, #444, #463, #525. Part of #403.

## The gap

Attaching a video produced a metadata block and a handful of downsampled JPEGs, for every provider alike — Gemini included, which has accepted inline video **with its audio track** the whole time. A 47-second screen recording reached it as three stills at 768px, so anything between them was simply not in the request.

Two things made this worse than the equivalent audio gap:

1. **The frames come from ffmpeg, which is an optional dependency.** Without it `extractKeyframes` returns an empty array, `processFile` treats that as non-fatal, and the model receives the metadata line alone — a video attachment conveying nothing visual, with no error.
2. **The metadata block answers exactly the questions a test tends to ask.** "How long is it?" and "what resolution?" both succeed with no video and no frames attached. The existing VIDEO-028 assertions in `continuous-test-suite-multimodal-sdk.ts:399-450` ask for pixel resolution, and pass under both conditions.

Sending the file itself needs **no ffmpeg**, which is what makes this the fix rather than "extract more frames".

## What changed

**`src/lib/adapters/videoFormatSupport.ts` (new)** — the per-provider capability table and its getters, mirroring `audioFormatSupport.ts`:

| | Mechanism | Inline ceiling | Audio track |
|---|---|---|---|
| Gemini (AI Studio, Vertex) | inline | 15 MB | heard |
| OpenAI / Anthropic / Bedrock / Mistral / … | frame extraction | – | – |
| Ollama / llama.cpp | frame extraction (4 frames) | – | – |

15 MB is on the **source** file: inline data is base64 in a JSON body, costing four bytes per three, and Gemini rejects a request over 20 MB.

A clip that fails the gate is **not an error** — it falls back to keyframes and logs which of the four reasons applied, because the user-visible symptom of all of them is identical ("it only described the file") and the remedies are not.

There is deliberately **no transcode path**, unlike audio: re-encoding a half-hour recording is minutes of CPU inside a generation request, for a payload that would usually breach the ceiling anyway.

**Plumbing** — detection carries the bytes forward as `input.nativeVideoFiles` alongside the summary and the frames, exactly as audio does. Both Gemini assembly paths attach them: the AI SDK `file` part in `convertMultimodalToProviderFormat`, and the hand-built `inlineData` request via a new `appendNativeVideoParts` (AI Studio / Gemini 3 through `buildUserPartsWithMultimodal`, Vertex directly).

**Public API** — `VIDEO_PROVIDER_CONFIGS`, `getVideoProviderConfig`, `supportsNativeVideo`, `canDeliverVideoNatively`, `isNativeVideoMimeType` and `estimateVideoTokens` are exported, so a caller can ask which mechanism applies and what it will cost before sending. Purely additive; no existing export changed.

## Verification

The new suite's fixture is a committed **14 KB** clip (3.6s, 160x120) carrying three colour segments and a voice saying a secret word. It runs without ffmpeg, and asserts something no keyframe and no line of the metadata summary can convey.

Live, on this branch:

- **Gemini** — *"1) The secret word is **flamingo**. 2) The colours shown are **Red, Blue, and Green**, in that order."*
- **OpenAI** (negative control) — *"NO_AUDIO"*, with zero `Added native video part` log lines and the `Video processed: … 4 keyframes` precondition present, so the negative is not the vacuous kind.

`test:video-native` — 9 passed, 0 failed, 0 skipped. **Failure injection** (per CLAUDE.md): breaking the `flamingo` assertion and inverting the native-part assertion produced `✗` on both and exit 1 — not `⊘`.

`pnpm run check` (0 errors) · `pnpm run lint` (0 errors, 80 pre-existing warnings, none added) · `pnpm run build` (publint clean) · `docs:api` regenerated · `docs-site/static/search-index.json` rebuilt and confirmed **byte-identical across two consecutive builds**.

## Notes for review

- **No ffmpeg is added to any workflow.** Nothing here needs it — the native path is precisely the one that does without.
- `VideoDeliveryDecision`'s accepting arm declares `reason?: undefined` rather than omitting the field. Not decoration: `build:react-hooks` type-checks this graph without `--strict`, where a negated boolean-literal discriminant does not narrow, and the build fails without it.
- `convertMultimodalToProviderFormat` hit the 6-parameter lint ceiling, so the two media arrays collapsed into one `NativeMediaAttachments` bag. Both converters are module-private; no caller is affected.


src/lib/adapters/videoFormatSupport.ts
```diff
@@ -0,0 +1,338 @@
...
+  durationSec: number;
+  frameCount?: number;
+  hasTranscription?: boolean;
+}): number {
+  const { provider, durationSec, frameCount, hasTranscription } = options;
+  const config = getVideoProviderConfig(provider);
+  const seconds = Math.max(0, durationSec);
+
+  if (config?.supportsNativeVideo) {
+    return Math.ceil(seconds * TOKENS_PER_NATIVE_VIDEO_SECOND);
```
Tara-ag, line 271: **MINOR — an unmeasured native clip budgets as free tokens.**

When a native provider is selected and `durationSec` is unknown (0), `Math.max(0, durationSec)` collapses to 0 and `estimateVideoTokens` returns `0`. But an unknown duration is *not* a rejection in `canDeliverVideoNatively` — such a clip is still delivered natively to Gemini. So a video that genuinely reaches the model as bytes is budgeted as zero, which is exactly the silent underestimation this estimate exists to prevent.

Suggest falling back to a non-zero estimate when the clip is natively delivered but under-measured:

```suggestion
  if (config?.supportsNativeVideo) {
    // An unmeasured clip is still delivered natively (unknown duration is
    // not a rejection), so it cannot budget as zero — charge it a nominal
    // slice of a minute rather than pretending it is free.
    const effectiveSeconds = seconds > 0 ? seconds : MIN_NATIVE_VIDEO_ESTIMATE_SEC;
    return Math.ceil(effectiveSeconds * TOKENS_PER_NATIVE_VIDEO_SECOND);
  }
```

src/lib/adapters/videoFormatSupport.ts
```diff
@@ -0,0 +1,338 @@
...
+
+  const mimeType = normalizeVideoMime(video.mimeType);
+  if (!isNativeVideoMimeType(mimeType)) {
+    return {
+      deliver: false,
+      reason: `${mimeType} is not a container ${provider} reads inline`,
+    };
+  }
+
+  const sizeMB = video.buffer.length / (1024 * 1024);
```
Tara-ag, line 314: **MINOR — the 15 MB ceiling is per-clip, but Gemini's 20 MB limit is per request body.**

The module comment correctly notes that inlineData is base64 in a JSON body (4 bytes per 3) and that the *request* ceiling is 20 MB. But `canDeliverVideoNatively` checks each clip independently against `maxSizeMB`. Two clips of ~13 MB each pass individually (≈26 MB source ⇒ ≈35 MB base64 body) and together overflow the request limit, producing the opaque HTTP 400 the module is explicitly trying to avoid.

This is a documented known limitation (the Files API follow-up is acknowledged), so flagging as MINOR — but consider an aggregate guard across `nativeMedia.video` (mirroring how a request shares one body), or at least a log note when the combined payload nears the ceiling.

<details><summary>reasoning</summary>

`canDeliverVideoNatively` is called once per clip inside the loop in `messageBuilder.ts`, with no cross-clip state, so the sum is never checked. The comment at `GEMINI_INLINE_VIDEO_MAX_MB` acknowledges "room left … for any other attachments sharing the request" but no code enforces that budget.

</details>

Tara-ag (approved):
Native video delivery is well-engineered and backwards compatible (no change to native audio; non-native providers keep the existing keyframe fallback). Two MINOR notes raised inline; nothing blocking.

Tara-ag:
## APPROVE — native video delivery is solid, backwards compatible, and well-tested.

Feature mirrors the existing native-audio precedent cleanly: a new `videoFormatSupport` module gates delivery on provider capability (`supportsNativeVideo`), validates container + duration + per-clip size (`canDeliverVideoNatively`), estimates tokens, and leaves the keyframe/image fallback untouched for non-native providers. `messageBuilder` pushes a `type:"file"` AI-SDK part; the native Gemini 3 / Vertex paths build `inlineData` gated on the same checks. Logging path uses `safeBasename` — no credential or path leakage spotted. Rule 1/3/4/5/15 all respected.

| Sev | Location | Finding |
|-----|----------|---------|
| MINOR | `videoFormatSupport.ts:270` | `estimateVideoTokens` returns 0 for an unmeasured native clip that is still delivered natively |
| MINOR | `videoFormatSupport.ts:313` | 15 MB ceiling is per-clip, but Gemini's 20 MB limit is per request body — no aggregate guard |

**Checked clean:** backward compatibility of native-audio (`supportsNativeAudio` untouched, no existing caller changed); no static provider imports (Rule 1); Gemini tool + structured-output policy untouched (Rule 3); no CLI leakage into the SDK path (Rule 4); `formatProviderError` still returns (Rule 6); e2e-only suite `test/continuous-test-suite-video-file-support.ts` (Rule 15, one module graph via `dist/index.js`); capability table keys match the provider strings used at each call site.

Both findings are non-blocking, so this is an APPROVE — happy to dig deeper on the aggregate-size ceiling if you want it addressed before merge.

src/lib/adapters/videoFormatSupport.ts
murdore: Fixed in `45d81abd1`: a natively delivered clip with an unknown duration is no longer priced as free. `estimateVideoTokens` now charges it a floor of `MIN_NATIVE_VIDEO_ESTIMATE_SEC` (5 s, `videoFormatSupport.ts:228`) at the native rate. A genuinely long clip still costs more than the floor. Covered by "an unmeasured native clip is not priced as free" in `test/continuous-test-suite-video-native.ts`, which fails with the fix removed.

src/lib/adapters/videoFormatSupport.ts
murdore: Fixed in `45d81abd1` with a real request-wide guard. `canDeliverVideoNatively` takes an optional `priorNativeVideoBytes` (default 0, so existing single-clip callers behave exactly as before) and charges each clip against what earlier clips in the same request already committed. The request-assembly paths thread a running total through it (`committedVideoBytes`, `messageBuilder.ts:2956`), so a clip that would push the base64 body past the ceiling takes the keyframe path instead, with a loggable reason. Covered by three cases in `test/continuous-test-suite-video-native.ts`: the unit-level charge, "two clips that individually fit but together overflow: only the first ships natively", and the control "two small clips that together stay under the ceiling both ship natively". They fail with the fix removed.

Tara-ag (changes requested):
Revising my earlier APPROVE to REQUEST_CHANGES on the current head.

The two MINORs I raised were fixed (accepted): `estimateVideoTokens` now floors unmeasured native clips, and `canDeliverVideoNatively` takes a `priorNativeVideoBytes` running total so individually-fine clips can't overflow the request together. Both are verified present at `6d4062a`.

What blocks merge is the request-wide ceiling guard only counting **video** bytes. It does not seed from the PDF, image, audio and keyframe parts that travel inline in the same 20 MB Gemini request body. A near-ceiling clip plus a PDF or image can still exceed the cap and fail the whole request with HTTP 400 instead of falling back to keyframes — the exact opaque failure the module documents wanting to avoid. See thread `src/lib/providers/googleNativeGemini3/utils.ts:1836-1840` (also `messageBuilder.ts:2952-2956`, `googleVertex/client.ts:1644-1650`, `:2695-2701`).

Please seed the running total from the already-assembled inline non-video parts (decoded base64 length) on all three assembly paths, and add a mixed video+PDF/image overflow test.

Two MINOR doc inconsistencies also remain open (CodeRabbit threads): `docs/features/multimodal.md` says 15 MB where the code is 14, and it presents `transcribeAudio` as implemented when it is marked not-implemented (`#433`).

Everything else is clean — backward compat preserved, no static provider imports, structured-output policy untouched, e2e-only suite, no credential leakage.

Tara-ag:
## Pull request review state — validated and corrected

Everything on the PR now reflects the actual review verdict. No further changes needed beyond what I already applied.

**Review state (matches verdict NEEDS_WORK):**
- New review submitted: **CHANGES_REQUESTED** (Tara-ag) on current head `6d4062a` — submit_review success.
- The old APPROVED review (5257307465) sits on the pre-rebase commit `6f99903`; the latest review on the head is the changes-requested one, so the binding PR state is CHANGES_REQUESTED. Submitted reviews cannot be deleted via the API, and GitHub surfaces the latest review as authoritative — correct as-is.

**Summary comment (one, no duplicate):**
- `5744740768` carries the `<!-- yama:summary -->` marker — updated (not duplicated) to a **NEEDS_WORK** verdict, with a findings table and a collapsible evidence block. Well-formed markdown, no broken fences.

**Inline threads (my two findings — each once, both accepted):**
- `video-token-estimate-duration-0` — resolved. Author reply (murdore) confirms the `MIN_NATIVE_VIDEO_ESTIMATE_SEC` floor; verified present at head. Accepted, not reposted.
- `video-aggregate-size-ceiling` — resolved. Author reply confirms the request-wide `priorNativeVideoBytes` guard; verified present at head. Accepted, not reposted.

**MAJOR that blocks merge (owned inline by CodeRabbit, not duplicated):**
- Thread `src/lib/providers/googleNativeGemini3/utils.ts:1836-1840` (unresolved, current head): the request-wide ceiling counts only video bytes, not the PDF/image/audio/keyframe inline parts sharing the same 20 MB body. I verified the claim in code on all three assembly paths and referenced it in both the updated summary and the CHANGES_REQUESTED review rather than posting a second inline copy.

**No malformed comments, no duplicates.** The PR now carries one clean, complete review: one summary, one comment per finding, and a review state (CHANGES_REQUESTED) that matches the NEEDS_WORK verdict.

PR #1743, opened 2026-09-19
fix(logger): route log events per NeuroLink instance so worker bridges attribute truthfully

### LINKED ISSUE #1236: Per-instance logger routing for worker log-bridge attribution
Follow-up from #1235 (tracked in RFC §6c of docs/plans/2026-07-27-isolated-agent-runner-rfc.md).

The NeuroLink logger is process-global with a single active emitter, so a worker instance's log bridge (`WorkerInstanceOptions.onLog`) receives ALL NeuroLink log events emitted in the process — host, other workers, MCP — each stamped with the forwarding bridge's tag. The tag identifies which bridge forwarded an event, not which instance emitted it.

Real fix: per-instance logger routing — log calls carry (or derive, e.g. via AsyncLocalStorage) the emitting instance so each instance's emitter receives only its own events. Then `createWorkerInstance` can bridge the worker's own emitter and `logTag` becomes true attribution.

Current interim state (shipped in #1235): the limitation is documented on `WorkerInstanceOptions.onLog`; `createWorkerInstance` restores the host as the active sink after construction; `dispose()`/voice-agent shutdown clear the sink only when they own it (`logger.clearEventEmitter(ifEmitter)`).

murdore:
Closes #1236.

## The problem

The NeuroLink logger is a process-global singleton with a **single** active
sink. A worker's `onLog` bridge subscribed to the *host's* emitter, so it
received every log event emitted anywhere in the process — the host, sibling
workers, MCP — each one stamped with that worker's `logTag`. The tag told you
which bridge forwarded an event, not which instance emitted it.

This was shipped knowingly in #1235 and documented as a `KNOWN LIMITATION` on
`WorkerInstanceOptions.onLog`, with the real fix deferred to RFC §6c.

## Measured

Two workers and a host, one `generate()` each, against the pre-change build:

| bridge   | events received | events it actually emitted |
| -------- | --------------- | -------------------------- |
| worker A | 708             | 299                        |
| worker B | 707             | 202                        |

After this change, worker A's bridge receives exactly its own 299 and worker
B's its own 202.

## The change

**`src/lib/utils/logger.ts`** — an `AsyncLocalStorage` carrying the emitting
instance's id, plus a registry of sinks keyed by that id:

- `runInInstanceScope(instanceId, fn)` / `getInstanceScope()`
- `addScopedEventEmitter` / `removeScopedEventEmitter` / `clearScopedEventEmitters`

`log()` delivers to the scoped sinks for whichever instance is on the stack,
*then* to the process-wide sink. The two are independent.

**`src/lib/neurolink.ts`** —

- every instance gets a process-unique `logInstanceId`;
- `generate`, `stream` and `generateText` run their bodies inside that scope
  (each keeps its original body verbatim, moved to a private
  `…InInstanceScope` method, so the diff on the hot paths is a wrapper, not a
  rewrite);
- `createWorkerInstance` subscribes the bridge to the **worker's own id**
  instead of `hostEmitter`;
- `dispose()` drops the instance's sinks — the registry is process-global, so
  an instance that went away without clearing its entry would keep its sink,
  and everything it closes over, reachable for the life of the process.

## Not a breaking change

- `logger.setEventEmitter()` is untouched and still receives everything, so an
  existing host bridge sees no difference. A test case asserts this.
- The public `Logger` type is **deliberately** left alone. It is a structural
  contract a caller can satisfy (`SDKToolContext.logger`), so adding required
  members would break anyone constructing one. The routing methods are internal
  plumbing and are not on it.
- `LogEventEmitter` is **new** in the types barrel — additive. It replaces the
  inline `{ emit: (event: string, ...args: unknown[]) => boolean }` shape that
  was written out four times.

## Known residual gaps (documented, not hidden)

Both are recorded on `WorkerInstanceOptions.onLog` and in the RFC:

- Logs emitted **while a consumer drains a returned stream** run in the
  consumer's async context, after `stream()` resolved, so they are attributed
  to the consumer's scope. Provider loops that run to completion inside the
  call are covered.
- Logs emitted **outside any entry point** — construction, background MCP
  reconnects, module init — stay *unattributed* rather than being charged to
  whichever instance happens to be around. A test case asserts this too.

## Test

`test/continuous-test-suite-logger-instance-routing.ts` — 8 cases, offline, no
credentials, driven entirely through the public surface. `NeuroLink` and
`logger` both come from `dist/index.js`; mixing in `src/lib/`'s logger would
watch a different singleton and pass silently.

Preconditions use `logger.getLogs()` rather than a process-wide sink, because
installing a sink *displaces the active emitter* and would mask the behaviour
under test — the first draft of this suite did exactly that and produced a
pre-fix run where the bridges looked dead rather than over-broad.

**Discrimination check** — run against a build with the old bridge wiring
restored:

```
✓ a worker's bridge receives that worker's own log events
✗ a sibling worker's bridge receives none of them
✗ the host's own call reaches neither worker bridge
✗ routing is reciprocal: B's call reaches B and not A
✓ every forwarded event carries the owning bridge's tag
✓ a process-wide sink still receives every instance's events
✓ dispose() unsubscribes a worker's bridge
✓ logs outside any generate call are attributed to no instance
  Passed: 5, Failed: 3 — RESULT: FAIL (exit 1)
```

Real `✗` and a non-zero exit, not `⊘` skips. All 8 pass on this branch.

## Verification

`pnpm run check`, `pnpm run lint` (0 errors), `pnpm run build`,
`pnpm run validate:all`, `check:deps`, `check:test-parse`,
`test:provider-structure`, `test:model-manifests` — all green.
`docs/api` regenerated and prettier-formatted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Log events from isolated worker instances are now routed only to the corresponding worker’s `onLog` handler, preventing unrelated instance logs from being received.
  * Disposed workers no longer continue receiving forwarded log events.
  * Process-wide logging continues to receive events from all instances.

* **Documentation**
  * Clarified log attribution behavior and documented cases where events remain unattributed, such as background activity and stream consumption.


Tara-ag:
**Validation pass — PR #1743 review is clean, final (2026-09-20 re-check).**

- **One live summary only:** canonical `<!-- yama:summary -->` is **5751276834**; archived summary **5744421192** carries the disjoint `yama:summary-superseded` marker and explicitly points to the canonical one. No duplicate `yama:summary` marker exists and none was added.
- **Inline comments verified intact:** dispose-bridge finding anchors `test/continuous-test-suite-logger-instance-routing.ts:225-233` (RIGHT/added) with a well-formed `suggestion` block; same-thread correction covers the `_probe`→`probe` typo. CodeRabbit threads intact at lines 49 and 233. No broken markdown, no bad anchors.
- **One comment per finding:** three review threads = three findings (CR @49, CR @233, dispose-bridge). The dispose thread's second comment is an explicit same-thread typo correction, not a duplicate finding.
- **Review state matches verdict:** verdict **APPROVE** is reflected by the approving review **5257046297** (state `APPROVED`). No needs-work review present.
- **No replies owed:** all three open threads carry only their original finding comments (plus the same-thread typo correction); there are no author replies awaiting a response.

Result: no malformed comments, no live duplicates, review state consistent with the verdict. Nothing to fix or delete.

test/continuous-test-suite-logger-instance-routing.ts
```diff
@@ -0,0 +1,282 @@
...
+    } finally {
+      logger.clearEventEmitter(sink);
+    }
+  });
+
+  await test("dispose() unsubscribes a worker's bridge", async () => {
+    await workerA.dispose();
+    const aAfterDispose = capturedA.length;
+
+    const emitted = await emitLogsFrom(workerB);
```
Tara-ag, line 253: **MINOR — the dispose test does not prove the bridge was actually removed.**

The post-dispose probe (`emitLogsFrom(workerB)`) runs inside worker B's instance scope, so `log()` looks up the scoped sinks for B's instance id only. Worker A's scoped emitter — even if it was never unregistered — would never be consulted, so `capturedA.length === aAfterDispose` passes whether or not `removeScopedEventEmitter` ran. The test verifies "routing is keyed by caller," not "dispose unregisters the sink," so a real leak (a worker bridge retained by the process-global logger) would slip through. This agrees with CodeRabbit's note at this line; here is a concrete fix that exercises the same `add`/`remove` plumbing `NeuroLink.dispose()` relies on.

```suggestion
    await workerA.dispose();

    // Emitting from worker B's scope can never reach A's scoped emitter, so the
    // original probe would pass even if A's bridge were still registered. Prove
    // the removal directly through the logger's own add/remove plumbing (the
    // same calls NeuroLink.dispose() makes): emit inside a scope whose emitter
    // we just removed and assert nothing arrives.
    let probeInvocations = 0;
    const probe = {
      emit: (event: string) => {
        if (event === "log-event") {
          probeInvocations++;
        }
        return true;
      },
    };
    logger.addScopedEventEmitter("probe-dispose", probe);
    logger.removeScopedEventEmitter("probe-dispose", probe);
    logger.runInInstanceScope("probe-dispose", () => {
      logger.error("[#1236] bridge-removal probe");
    });
    assert(
      probeInvocations === 0,
      `a removed scoped emitter still received events (${probeInvocations})`,
    );
```

(If `_probe` were left registered, `logger.error` inside the scope would invoke it, so the assertion genuinely distinguishes "removed" from "never tested.")

Tara-ag (approved):
Per-instance logger routing is well-engineered: correct AsyncLocalStorage scoping, idempotent dispose cleanup, unchanged public `Logger` shape, and an e2e suite driven through `dist/index.js`. One MINOR test-coverage gap on the dispose test (see inline). Approving.

Tara-ag:
**(Archived — original-review-pass summary.)**

This was the first review pass's summary. It has been consolidated into the canonical, current **`<!-- yama:summary -->`** comment further down this thread (recurring-review pass, 2026-09-20), which carries the full, up-to-date finding list.

Verdict is the same in both: **APPROVE**. The single non-blocking item is the MINOR test-coverage note on the dispose test at `test/continuous-test-suite-logger-instance-routing.ts:225-233`. Please treat the canonical summary below as authoritative.

test/continuous-test-suite-logger-instance-routing.ts
Tara-ag: Small typo in the note under the suggestion block: it should read `probe` (the variable the suggestion defines), not `_probe`. The suggestion block itself is correct as written.

Tara-ag:
**Review summary — round 3 (final)**

**Verdict: APPROVE.** The per-instance logger dispose test was rewritten to genuinely prove bridge removal; no other findings remain open and the implementation continues to satisfy all enforced rules.

**Changes since last round**
The PR was reset to a single commit and the test report redesigned. At head `688379ebe`, the dispose test now captures worker A's instance id at construction, re-enters worker A's **own** scope directly via `logger.runInInstanceScope(workerAInstanceId, …)` after `dispose()`, and asserts the bridge no longer receives a `dispose-removal probe` — a real regression check that the old cross-worker assertion could not provide. The stray-suffix typo from the earlier draft is gone.

| Severity | File : line | Finding | Status |
|---|---|---|---|
| MINOR | `test/continuous-test-suite-logger-instance-routing.ts` : 253 | Dispose test didn't prove bridge removal (cross-worker emit could pass with the bridge retained) | **Resolved** at `688379ebe`; thread `PRRT…CQFF` resolved |

**What was checked and found clean**
- `src/lib/utils/logger.ts` — `log()` still fans out to the scoped set *and* unconditionally hits the process-wide emitter (backward compatible); `clearEventEmitter(ifEmitter)` only clears when the disposing instance owns the bridge. No hot-path regression in the stream/tool-merge paths.
- `src/lib/types/utilities.ts` — public `Logger` shape unchanged (Rule 5 satisfied; no unmodified caller breaks).
- Test suite drives `dist/index.js` with a single module graph and declares a determinism exception (Rule 15 satisfied).
- Impact analysis ran via the code graph (yaml:review uses detect_changes / affected-flows); no out-of-diff call sites regress.

**Other review comments of record**
- CodeRabbit thread `PRRT…CHC9` (move test-local `Captured` type) — test-local helper, type-location is CI/AST-rules owned; not part of this review's findings.
- CodeRabbit thread `PRRT…CHC_` (test scoped-emitter removal directly) — same underlying coverage gap as the resolved dispose finding; the author's rewrite satisfies it, so this thread is being resolved rather than re-reported.

No duplicates: one summary (this comment), one finding thread, all resolved.

test/continuous-test-suite-logger-instance-routing.ts
Tara-ag: Resolved — the rewritten dispose test at the current head now proves the removal for real. It captures worker A's instance id at construction, and after `dispose()` re-enters worker A's **own** scope directly via `logger.runInInstanceScope(workerAInstanceId, …)` and asserts the bridge no longer receives the `dispose-removal probe`. That is a genuine regression check (the old cross-worker assertion could pass even with the bridge retained). The stray-suffix typo from the earlier draft is also gone. Marking this thread resolved.

Tara-ag:
**Archived — superseded by the canonical review summary (`yama:summary` elsewhere on this PR).**

Please see the current verdict and findings table in the live summary comment. Kept here only for round-2 history: the dispose test was confirmed rewritten at head `688379ebe` to prove bridge removal, and no new blocking issues were found.

test/continuous-test-suite-logger-instance-routing.ts
```diff
@@ -0,0 +1,282 @@
...
+import { NeuroLink, logger } from "../dist/index.js";
+
+const { test, runSuite } = defineSuite("Per-instance logger routing (#1236)", {
+  offline: true,
+});
+
+/** An unregistered provider name: rejects inside generate(), reaches no network. */
+const UNREGISTERED_PROVIDER = "not-a-registered-provider-1236";
+
+type Captured = { tag: string; level: string; message: string };
```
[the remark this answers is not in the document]
murdore: Keeping this one local. Rule 2 (types live in `src/lib/types/`) is scoped to `src/**`: `neurolink/no-local-type-alias` is enabled in the `files: ["src/**/*.ts", "src/**/*.tsx"]` block of `eslint.config.js`, and the `test/**/*.ts` block does not enable it. `src/lib/types/index.ts` is `export *`, so moving `Captured` there would publish a test-only shape as part of the SDK's public types. It stays with the suite, like the other test-local helper types under `test/`.

Tara-ag (approved):
## Approving — re-review after branch rebase/squash

All three review threads are resolved with adequate justification, and the squashed commit `91d1a2e` is content-identical to the code I previously approved:

- **Captured types-location** — author's `src/**` vs `test/**` rule-scope justification is correct; CodeRabbit withdrew and resolved.
- **CodeRabbit scoped-emitter test** — rewritten dispose test exercises the `runInInstanceScope` + add/remove plumbing directly.
- **My dispose-bridge finding** — the rewritten dispose test genuinely verifies that `dispose()` removes the bridge (it re-enters the worker's own scope and asserts the `dispose-removal probe` is no longer delivered).

The routing fix for #1236 is well-scoped (`AsyncLocalStorage`-keyed scoped sinks with correct dual routing to global sinks in `log()`), the `neurolink.ts` wrappers correctly thread `logInstanceId` through generate/generateText/stream, and `dispose()` cleans up both the bridge and scoped emitters. Tests now prove the intended behavior.

No new issues introduced. Approving.

Tara-ag:
## Review summary — Round 5 (no-change re-check)

Re-checked the branch at head `6cc402aa` against the previously approved commit. **No new findings — verdict remains APPROVE.**

### Recurring-review reconciliation (nothing reposted)

This round produced the same three threads as Round 4, and each resolution still stands with adequate justification, so none are re-published:

- **`Captured` types-location thread (`test/…:49`)** — `neurolink/no-local-type-alias` (Critical Rule 2) is scoped to `files: ["src/**/*.ts", "src/**/*.tsx"]` in `eslint.config.js`, so it does not apply to `test/**`. CodeRabbit re-verified and **withdrew** the suggestion itself. Accept.
- **CodeRabbit scoped-emitter test thread (`test/…:233`)** — the rewritten dispose test directly exercises the `runInInstanceScope` + add/remove plumbing the finding asked for. Accept.
- **Dispose-bridge-removal finding (`test/…:253`, resolved in Round 3)** — the dispose test re-enters worker A's **own** captured scope after `dispose()` via `logger.runInInstanceScope(workerAInstanceId, …)` and asserts the removed bridge receives nothing, proving removal. Accept.

### Content is unchanged

The author rebased the branch onto `release@75db63d` and confirms the non-generated diff reproduces the original commit `91d1a2e` byte-for-byte (`git apply --3way` + `git patch-id --stable` match exactly). The diff at `6cc402aa` matches what I approved in Round 4: the `AsyncLocalStorage` scoping, scoped-emitter registry, per-instance bridge wiring, `dispose()` cleanup, and the 8-case offline routing suite (fixed run `8/8 PASS`, broken run `3/8` proving the fix is load-bearing). Single Commit enforcement passed; `lint`/`build`/`check` green.

### One open items actionable by the author (not a code finding)

`mergeable_state` is currently **`blocked`** for PR #1743 — the branch needs updating from `release` to clear the divergence/conflict before it can merge. This is the sole remaining item and it is not a review blocker on the code itself. (Corresponds to CodeRabbit's "resolve merge conflict" finishing-touch checkbox.)

**Verdict: APPROVE.** Routed in as correct for #1236; the tests now genuinely prove per-instance attribution and bridge removal. The change is safe to merge once the branch is brought up to date with `release`.

PR #1306, opened 2026-08-08
fix(tts): stop downgrading OpenAI flac requests to mp3

murdore:
Refs #479

## The bug

`flac` is **both** a valid `TTSAudioFormat` and a documented OpenAI `/audio/speech` `response_format`. But `mapFormat()` only listed mp3/wav/ogg/opus/pcm16, so a flac request fell into the unsupported-format branch and was coerced to mp3.

A caller who asked for lossless silently got lossy. The only signal was a warn-level log — and `TTSResult.format` then reported `"mp3"`, so nothing downstream could detect it either.

| requested | before | after |
| --- | --- | --- |
| `flac` | `mp3` → reported as `mp3` | **`flac` → reported as `flac`** |
| mp3 / wav / ogg / opus / pcm16 | unchanged | unchanged |
| m4a | `mp3` | `mp3` (correct — see below) |

## Scope, checked rather than assumed

The issue also names `aac`. **`aac` is absent from the `TTSAudioFormat` union**, so it cannot be requested at all and needs no mapping — adding it would mean widening the public type for a format nothing asks for.

`m4a`, `webm`, `mp4`, `mpeg` and `mpga` *are* in that union, but they're there as Whisper **input** formats sharing the type. OpenAI TTS cannot produce them, so coercing them to mp3 with a warning remains correct — the test asserts that explicitly so a later change doesn't "helpfully" map them to something OpenAI would reject.

## Verification

`mapFormat` and `effectiveFormat` are pure, so the test exercises them off the prototype — no API key, no network:

```
✓ #479: flac is a real OpenAI response_format and must not downgrade to mp3
Passed: 14   Failed: 0
```

Confirmed it catches the regression — removing the `flac` mapping makes it report `✗` and exit non-zero.

`pnpm run check` 0 errors.

## Not closing #479

The issue's other outstanding criterion — *"add duration estimation"* — is still unmet: `TTSResult.duration` is never populated by this handler. That's separate work, so this is `Refs`, not `Closes`.

## Summary by CodeRabbit

- **New Features**
  - Added support for requesting and receiving OpenAI text-to-speech audio in FLAC format.
  - FLAC is now correctly reported as the selected audio format.

- **Bug Fixes**
  - Unsupported audio formats continue to fall back to MP3.
  - Format warnings now accurately include FLAC among supported options.

- **Tests**
  - Added coverage for FLAC mapping and existing audio-format behavior.

Issue #479, opened 2025-12-01
TTS-008: Implement OpenAITTSHandler.synthesize()


murdore:
## Summary
Implement the core synthesis method in OpenAITTSHandler that calls OpenAI's TTS API.

## Technical Details
- **File(s)**: src/lib/adapters/tts/openaiTTSHandler.ts
- **Effort**: 3h

## Acceptance Criteria
- [ ] Implemented synthesize() method in OpenAITTSHandler
- [ ] Model selection: use tts-1-hd if quality is hd, else tts-1
- [ ] Voice selection: default to alloy if not specified
- [ ] Support all 6 OpenAI voices (alloy, echo, fable, onyx, nova, shimmer)
- [ ] Support all OpenAI formats (mp3, opus, aac, flac, wav, pcm)
- [ ] Speed control: 0.25x to 4.0x
- [ ] Call client.audio.speech.create() API
- [ ] Convert response ArrayBuffer to Buffer
- [ ] Return TTSResult with buffer, format, metadata
- [ ] Add duration estimation
- [ ] Error handling for API failures
- [ ] Track latency in metadata

## Dependencies
- Depends on: TTS-001, TTS-007
- Blocks: TTS-003, TTS-020

Priority: high | Effort: 3h | Complexity: moderate

murdore:
### ◐ Audit status: partially done — keeping open

Core behaviour ships, but not every acceptance criterion is met. The remaining gap is the actionable part:

> Core synthesize() behavior (model/voice/speed/format-ish/error-handling/latency) works and is exercised by generate({tts}) flows, but 2 of the 6 required output formats are unsupported and duration estimation was never added.

<details><summary>what already works (evidence)</summary>

src/lib/voice/providers/OpenAITTS.ts:111-246 implements synthesize(): model selection tts-1/tts-1-hd (127-129), default voice 'alloy' (132), speed passthrough (143), latency tracking (198,215), Buffer conversion (201-202), TTSResult with metadata (208-221), error handling (228-245). Gaps: mapFormat() (253-271) only maps mp3/wav/ogg->opus/opus/pcm16->pcm — 'aac' and 'flac' (explicitly required formats) are NOT mapped and silently fall back to mp3 with a warning (264-269); no `client.audio.speech.create()` SDK call is used (raw fetch instead, line 150); no duration estimation is added anywhere in the TTSResult (no `duration` field ever set).

</details>


[...]

PR #1746, opened 2026-09-19
feat(tts): stream Google audio natively and add a voices command

### LINKED ISSUE #448: TTS-EPIC-001: Text-to-Speech Output Support
## Summary
Add comprehensive Text-to-Speech (TTS) output capability to NeuroLink SDK and CLI, enabling users to convert AI-generated text responses into high-quality audio across multiple providers (OpenAI, Google Cloud TTS, Azure Speech Services).

## Problem Statement
NeuroLink currently supports comprehensive multimodal INPUT but lacks TTS OUTPUT capability. Users cannot convert AI-generated responses into audio format.

## Business Value
- Accessibility: Enable visually impaired users to consume AI responses
- Voice Interfaces: Power voice assistants and conversational AI applications
- Content Creation: Generate audio content
- Enterprise Use Cases: IVR systems, customer service bots, e-learning platforms

## Total Effort
~70 hours across 28 subtasks

## Phases
1. Core Infrastructure (14h)
2. Provider Implementations (28h)
3. BaseProvider Integration (10h)
4. CLI Integration (9h)
5. Testing (8h)

Priority: high
Effort: 70h
Complexity: very-complex

### LINKED ISSUE #492: TTS-013: Implement GoogleTTSHandler.synthesizeStream()

### LINKED ISSUE #524: TTS-026: Create CLI Voices Command
## Summary
Create a new CLI command 'neurolink voices' that lists all available TTS voices for a provider.

## Technical Details
- **File(s)**: src/cli/commands/voices.ts (create new), src/cli/index.ts
- **Effort**: 2h

## Acceptance Criteria
- [ ] Created src/cli/commands/voices.ts
- [ ] Implemented voices command with yargs
- [ ] Accept --provider <provider> flag (required)
- [ ] Accept --language <code> flag (optional filter)
- [ ] Call TTSProcessor.getVoices(provider, { languageCode })
- [ ] Display voices in formatted table
- [ ] Sort voices alphabetically by name
- [ ] Show voice count at end
- [ ] Handle errors gracefully
- [ ] Register command in src/cli/index.ts
- [ ] Examples in help text

## Dependencies
- Depends on: TTS-005, TTS-010, TTS-014
- Blocks: none

Priority: low | Effort: 2h | Complexity: simple

### LINKED ISSUE #528: TTS-028: Create Integration Tests for TTS Providers
## Summary
Create integration tests for all TTS provider handlers (OpenAI, Google, Azure) that test real API calls, audio generation, streaming, voice listing, and error handling.

## Technical Details
- **File(s)**: test/integration/tts-providers.test.ts (create new)
- **Effort**: 4h

## Acceptance Criteria
- [ ] Created test/integration/tts-providers.test.ts
- [ ] Test suite for OpenAITTSHandler
  - [ ] Test: synthesize() generates valid audio
  - [ ] Test: All 6 voices work
  - [ ] Test: Speed control works
  - [ ] Test: synthesizeStream() yields audio chunks
  - [ ] Test: getVoices() returns 6 voices
- [ ] Test suite for GoogleTTSHandler
  - [ ] Test: synthesize() generates valid audio
  - [ ] Test: SSML input works
  - [ ] Test: synthesizeStream() yields audio chunks
  - [ ] Test: getVoices() returns 220+ voices
- [ ] Test suite for AzureTTSHandler
  - [ ] Test: synthesize() generates valid audio
  - [ ] Test: Emotional styles work
  - [ ] Test: synthesizeStream() yields audio chunks
- [ ] Tests use real API keys from environment
- [ ] Tests can be skipped if API keys missing
- [ ] Validate audio buffer is non-empty
- [ ] Validate audio metadata is correct
- [ ] All tests pass

## Dependencies
- Depends on: TTS-008, TTS-009, TTS-010, TTS-012, TTS-013, TTS-014, TTS-016, TTS-017
- Blocks: none

Priority: high | Effort: 4h | Complexity: moderate

murdore:
Closes the two genuine gaps in the TTS-EPIC-001 (#448) cluster. The other three issues in the cluster were verified as already shipped — see the audit table below.

### Google native streaming (#492, TTS-013)

`GoogleTTSHandler` now implements the optional `TTSHandler.synthesizeStream` seam that #1588 added for OpenAI, so `TTSProcessor` prefers provider-native reads and keeps the buffered path for everything else. No new abstraction — the same seam, the same global chunk normalization, the same fallback rules.

The gate is measured against the live API, not assumed:

| | streams | refused |
|---|---|---|
| voice | `Chirp3-HD` (en-US, en-GB, de-DE), `Chirp-HD`, `Journey` | `Neural2`, `Studio` — *"only Chirp 3: HD voices are supported for streaming synthesis"* |
| encoding | `PCM`, `OGG_OPUS` | `MP3`, `LINEAR16` — *"Unsupported audio encoding"*, though `synthesizeSpeech` takes both |

Measured delivery: `PCM` 41 reads, first at 697 ms, body complete at 6131 ms. `OGG_OPUS` 11 reads, first at 443 ms.

`StreamingSynthesisInput` has no SSML field, so `<speak>` input stays on the buffered path rather than being sent as literal text. Everything outside the gate answers `undefined` and is served exactly as before — **including the default voice (`en-US-Neural2-C`) and the default format (`mp3`)**, so native delivery is opt-in and no existing call changes shape.

The stream is bounded by an **idle** timeout rather than a total one. Streaming produces audio at roughly playback speed, so a whole-call bound would fail a long but perfectly healthy segment purely for being long.

### Voice discovery (#524, TTS-026)

`--tts-voice` took a provider-specific id that nothing in the CLI printed.

- `TTSProcessor.getVoices(provider, { languageCode })` — wraps the optional handler member, reporting an unregistered provider and a handler without voice listing as typed `TTS_PROVIDER_NOT_SUPPORTED` errors rather than a `TypeError` on an absent member.
- `neurolink voices --provider <p> [--language <code>] [--json]` — sorted table or JSON, count at the end, non-zero exit naming the registered set when the provider is not among it.

Sorted by codepoint, not `localeCompare`: collation depends on the Node ICU build, so the same list would otherwise order differently on CI and a laptop.

### Tests (#528, TTS-028 — live half)

`continuous-test-suite-tts-unit.ts` already covered the no-key half. This adds the live half to `test:tts`, end-to-end against `../dist/index.js`.

The discriminator for native delivery is **the chunk count for one sentence**. `TTSProcessor` segments at sentence boundaries and serves each segment from either a native stream or a single buffered `synthesize()`, so one sentence yields exactly one chunk on the buffered path however long it is — more than one is unreachable without the native read. `streamingBufferSize` is set above the sentence length so segmentation cannot manufacture the extra chunks.

Each case first asserts audio was produced at all, so the count is never vacuous. The same voice with `mp3` is the negative control and must yield exactly one.

## Verification

```
pnpm run check            0 errors
pnpm run lint             0 errors (79 pre-existing warnings)
pnpm run build            publint: all good
pnpm run test:tts         23 passed, 1 skipped, 0 failed, exit 0
```

New cases:

```
[PASS] TTS - Google native streaming (#492) — 26 native chunks, 293760 bytes
[PASS] TTS - Google streaming format gate (#492) — mp3 fell back to buffered, 21696 bytes
[PASS] TTS - getVoices rejects an unregistered provider (#524) — typed TTS_PROVIDER_NOT_SUPPORTED
[PASS] CLI voices - unknown provider exits non-zero (#524) — exit 1, registered set named
[PASS] CLI voices - OpenAI list (#524) — 6 voices, sorted
[PASS] TTS - live provider synthesis matrix (#528) — 4 synthesized: openai-tts 38016B,
       google-ai 22464B, elevenlabs 33062B, cartesia 43094B | skipped: azure-tts (upstream credential/quota)
```

**Failure injection** (per the CLAUDE.md warning about assertion messages being downgraded to skips): breaking the native-streaming chunk-count assertion and the CLI output assertion on purpose produced `Passed: 21, Failed: 2, Skipped: 1` with **exit 1** — both reported `[FAIL]`, neither was downgraded to a skip.

Live cases skip cleanly without credentials and treat a 401/403 as the environmental condition it is; `azure-tts` skipped that way on this run (expired key in the local env). No live call is a CI gate — `test:tts` is not in `provider-safety-net` or `test`.

Also green: `check:deps`, `check:ci-scripts`, `check:test-parse`, `check:tools-tests`, `test:provider-structure`, `test:model-manifests`, `validate`, `validate:env`, `validate:security`, and the `docs/api` currency check (regenerated, reproducible). `docs-site/static/search-index.json` regenerated — the delta is exactly the three TTS sections edited, 14368 entries before and after.

## Audit of the rest of the cluster

These were verified as already shipped, under names that do not match the issues. No code was written for them.

| Issue | Status | Evidence |
|---|---|---|
| #479 TTS-008 `OpenAITTSHandler.synthesize()` | **already done** | `src/lib/voice/providers/OpenAITTS.ts:356`, exported at runtime from `dist/index.js` as both `OpenAITTS` and `OpenAITTSHandler`. All six voices at `:155-205`, speed/model/format at `:246-258`, latency + metadata at `:400`. The last open criterion (flac downgrading to mp3) was closed by merged PR #1306, which cites the issue at `:553`. The planning issue named `src/lib/adapters/tts/openaiTTSHandler.ts`; PR #619 for that path was closed unmerged. |
| #505 TTS-017 `AzureTTSHandler.synthesizeStream()` | **already done** | The issue specifies sentence-buffered streaming, which `TTSProcessor.synthesizeStream()` (`src/lib/utils/ttsProcessor.ts:923`) already provides for *every* handler without a native stream, Azure included — sentence-boundary accumulation, one `synthesize()` per segment, global sequence indexes, exactly one final chunk. Reachable as `NeuroLink.stream({ tts: { provider: "azure-tts", enabled: true } })`. Adding the same buffering inside `AzureTTS` would be a second parallel implementation of it. Native wire streaming for Azure was *not* added: the local `AZURE_SPEECH_KEY` returns 401, so there is no wire evidence, and #1588 set the precedent that this seam is enabled only on direct measurement. |
| #448 TTS-EPIC-001 | umbrella | Closable once its children are; this PR closes the last three. |

Closes #448
Closes #492
Closes #524
Closes #528


src/lib/utils/ttsProcessor.ts
```diff
@@ -396,6 +397,66 @@ export class TTSProcessor {
...
+    const listVoices = handler.getVoices;
+    if (typeof listVoices !== "function") {
+      throw new TTSError({
+        code: TTS_ERROR_CODES.PROVIDER_NOT_SUPPORTED,
+        message: `TTS provider "${providerName}" does not support voice listing`,
+        category: ErrorCategory.VALIDATION,
+        severity: ErrorSeverity.MEDIUM,
+        retriable: false,
+        context: { provider: providerName },
+      });
```
Tara-ag: **MINOR — `getVoices()` forwards an unshaped handler error and skips the `isConfigured()` gate that `synthesize()` enforces.**

`synthesize()` funnels every provider failure through `toSynthesisError()` so `ttsMetadata.error` keeps its `code`, provider-qualified message, category and `retriable` flag — and it blocks on `TTS_PROVIDER_NOT_CONFIGURED` via `isConfigured()` before touching the handler. This new path does neither: whatever `handler.getVoices()` throws (a raw auth `Error`, a gRPC failure, a malformed response) escapes verbatim, so the `@throws TTSError` contract in this method's own JSDoc is not actually guaranteed for provider-borne errors, and a caller keying retry logic off `retriable` (as the buffered path's `toSynthesisError` is explicitly designed to support) gets an untyped object.

Recommended: gate on `isConfigured()` and run the return through the same shaper used by `synthesize()` (or a `getVoices`-flavoured variant of `toSynthesisError`):

````suggestion
    if (!handler.isConfigured()) {
      throw new TTSError({
        code: TTS_ERROR_CODES.PROVIDER_NOT_CONFIGURED,
        message: `TTS provider "${providerName}" is not configured. Please set the required API keys.`,
        category: ErrorCategory.CONFIGURATION,
        severity: ErrorSeverity.HIGH,
        retriable: false,
        context: { provider: providerName },
      });
    }

    try {
      return await listVoices.call(handler, options.languageCode);
    } catch (error) {
      throw this.toSynthesisError(error, providerName, "", options as TTSOptions);
    }
````

(If `toSynthesisError`'s `text`/`options` shape is a poor fit for voice listing, at minimum wrap the raw value so `getVoices` never leaks a non-`TTSError` to the public SDK surface pointed at by Rule 5.)

**Review by Tara-ag — CHANGES_REQUESTED** [review]
Overall this is a careful, well-engineered PR: the native-streaming path in `googleTTSHandler.ts` correctly recomputes finality in `readNativeSegment`, scopes streamability narrowly (voices/formats/SSML), and `TTSProcessor` cleanly rejects unsupported providers. Two things keep this from merge as-is.

**Comment by Tara-ag** [comment]
## NEEDS_WORK

Solid, carefully-engineered PR — the streaming architecture (narrow streamability gating in `googleTTSHandler.synthesizeStream`, finality recomputation in `readNativeSegment`, per-segment native streaming with `attachStreamCancel` cleanup) is thoughtful and well-tested. Two things block merge: an unresolved idle-timer bug that tears down healthy streams, and a small but real error-shaping gap in the new `getVoices()` path.

### Findings

| MAJOR | `src/lib/adapters/tts/googleTTSHandler.ts:530–546` (existing CodeRabbit thread r4089952972, unresolved) | **Idle timer stays armed across each chunk's consumer read** — a slow consumer read (>30 s) while paused on `yield` destroys a healthy gRPC duplex, reported as a retriable *network* failure (backpressure misattributed as a server stall). No author reply yet, so it stands. |
| MINOR | `src/lib/utils/ttsProcessor.ts:454` | `getVoices()` skips the `isConfigured()` gate and forwards the handler's error unshaped (raw error escapes the typed-`TTSError` JSDoc contract), unlike `synthesize()`. Inline suggestion attached. |

### Checked clean
- **Rule 1 (registry dynamic imports):** all TTS provider imports in `providerRegistry.ts` stay inside factory functions — no static provider import introduced.
- **Rule 3 (Gemini tools ↔ structured output):** unchanged/untouched by this diff.
- **Rule 4 (CLI ≠ SDK):** the new `voices` CLI command is isolated; no CLI concerns leak into the SDK path.
- **Rule 5 (public SDK surface):** the new `synthesize()` streaming mode is additive (`TTSProcessor.synthesizeStream` over the existing `synthesize`); no existing caller's signature/return shape changed.
- **Rule 15 (e2e-only tests):** new tests drive `dist/cli/index.js` and `dist/index.js`; no `src/`/`dist/` module-graph mixing added.
- **Security:** provider params (API keys) are read from env and stripped via `transformParamsForLogging`/`secret` in `DataSourceInfo`; no credentials hardcoded or logged in the diff; no injection/SSRF/path-traversal surface added.
- Duplicates avoided: the idle-timer point is referenced here but not re-posted — it is anchored by the existing CodeRabbit thread and reported as a standing MAJOR.

### Suggested handling
Address the idle-timer/backpressure interaction (timers should reset only on actual server responses, not while the consumer is reading a chunk) and the `getVoices` error-shaping; then this is ready to merge.

**Comment by Tara-ag** [comment]
Superseded pointer — the full review lives in the single canonical summary comment above (#5807870254). This comment is de-duplicated: it no longer carries the `yama:summary` marker, so there is exactly one summary on this PR.

**Review by Tara-ag — APPROVED** [review]
APPROVE — both previously-flagged findings (Google streaming idle-timer race and `getVoices` error shaping) are fixed in the committed code with test-first regression coverage and the author's RED/GREEN evidence. Types are consolidated under `src/lib/types/`, the shared `toSynthesisError` helper removes duplication, and the streaming/voices/live-matrix test additions meaningfully extend the continuous suite. No blockers. Full detail in the summary comment.

File: src/lib/adapters/tts/googleTTSHandler.ts [hunk 2]
```diff
@@ -354,6 +358,232 @@ export class GoogleTTSHandler implements TTSHandler {
...
+          yield {
+            data,
+            format,
+            index: index++,
+            isFinal: false,
+            cumulativeSize,
+            voice: voiceId,
+            sampleRate: sampleRateHertz,
+          };
+        }
```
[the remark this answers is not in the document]
murdore: Fixed, test-first. The timer now clears on receipt of each response instead of re-arming unconditionally, re-arms immediately for an empty read, and re-arms only after the consumer resumes the generator following a yield — matching the proposed diff. Reproduced RED by driving `GoogleTTSHandler.synthesizeStream()` directly (`TTSProcessor.synthesizeStream()`'s one-chunk lookahead buffer would otherwise absorb this failure as a false negative): pulling one chunk of a 306-word streaming response, pausing consumption 31s past the idle window while Google's server was still genuinely mid-flight, then pulling again. Unfixed: `the stream ended after the paused pull instead of continuing to deliver the response still being produced` (FAIL, exit 1). Fixed: `delivered a further chunk after a 31s consumer pause without the connection being torn down` (PASS, exit 0). New test: `TTS - Google streaming survives a slow consumer (PR#1746)` in `test/continuous-test-suite-tts.ts`. Committed at 8433e7610bac9da8b41c7a489e50bcccdd46303d.

src/lib/utils/ttsProcessor.ts
murdore: Fixed, test-first. `getVoices()` now runs the same `isConfigured()` gate `synthesize()` runs before touching the handler, and routes whatever the handler throws through `toSynthesisError()` so `code`/`category`/`retriable` survive on the resulting `TTSError`. (The handler's own `{ languageCode }` options share no fields with `synthesize()`'s `TTSOptions`, so an empty object is forwarded as context rather than an invalid cast into that parameter.) New test `TTS - getVoices gates isConfigured, shapes errors (PR#1746)` registers synthetic `TTSHandler` doubles via `TTSProcessor.registerHandler()` (one unconfigured, one configured-but-throwing) and asserts both the ordering and the shaping. Unfixed: `getVoices() called the handler's own getVoices() before checking isConfigured()` (FAIL, exit 1). Fixed: `unconfigured handler gated before the call; raw failure shaped into TTSError` (PASS, exit 0). Committed at 8433e7610bac9da8b41c7a489e50bcccdd46303d.

Issue #305, opened 2025-12-01
PC-007: VISION_CAPABILITIES Has Duplicate Models


murdore:
## Bug Summary
Claude models listed multiple times across providers with inconsistent naming, creating maintenance burden.

## Root Cause
src/lib/adapters/providerImageAdapter.ts (lines 26-206)
Same models duplicated 3+ times: claude-3-5-sonnet, claude-3.5-sonnet, claude-3-5-sonnet@

## Affected Providers
All providers with vision support

## Current Behavior
- Anthropic: "claude-3-5-sonnet"
- Vertex: "claude-3-5-sonnet", "claude-3-5-sonnet@", "claude-3.5-sonnet"
- Bedrock: "claude-3-5-sonnet", "claude-3.5-sonnet", "anthropic.claude-3-5-sonnet"
- Updates require changing 3+ places

## Fix
Create MODEL_FAMILIES constants:
- CLAUDE_VISION: Single list of Claude models
- GEMINI_VISION: Single list of Gemini models
- Generate provider variants with .map()

## Implementation
\`\`\`typescript
const MODEL_FAMILIES = {
  CLAUDE_VISION: ["claude-3-7-sonnet", "claude-3-5-sonnet", ...],
  GEMINI_VISION: ["gemini-2.5-pro", "gemini-1.5-pro", ...],
};

const VISION_CAPABILITIES = {
  anthropic: MODEL_FAMILIES.CLAUDE_VISION,
  vertex: [
    ...MODEL_FAMILIES.GEMINI_VISION,
    ...MODEL_FAMILIES.CLAUDE_VISION.map(m => \`\${m}@\`),
  ],
};
\`\`\`

## Acceptance Criteria
- MODEL_FAMILIES created
- No duplicates within same provider
- All capabilities preserved
- Tests pass

## Priority
High - Maintenance burden

## Effort
2 hours

murdore:
### ⚪ Closing — low priority (won't-fix for now)

Backlog triage assessed this as low-impact relative to effort:

> Duplicate literal model-name variants in vertex VISION_CAPABILITIES are harmless (just extra array entries, no incorrect behavior) - a MODEL_FAMILIES dedup refactor is pure tidiness, not worth prioritizing over feature work.

Closing to keep the backlog focused. Reopen with a concrete use-case if it matters to you.

<sub>— triaged in the open-issue backlog review (release v9.84.2, 2026-07-12). Reopen if you disagree with this assessment.</sub>

Issue #292, opened 2025-12-01
FD-004: Fix Confidence Level Mismatch in ExtensionStrategy


murdore:
## Bug Summary
Documentation states ExtensionStrategy has 70% confidence but implementation returns 85%, creating confusion and breaking confidence-based logic.

## Root Cause Analysis
**Why this bug exists:**
Documentation (line 62) correctly states 70% confidence for extension detection, but code (line 475) returns 85%. Likely a synchronization issue during development.

**Impact:**
- Incorrect confidence-based decisions
- Suboptimal strategy selection
- Documentation/code mismatch

## Current Code (Buggy)
\`\`\`typescript
// Line 62: Documentation
// 3. ExtensionStrategy (70% confidence) - File extension

// Line 475: Implementation
metadata: { confidence: type ? 85 : 0 },  // ❌ Should be 70
\`\`\`

## Fixed Code
\`\`\`typescript
// Line 475
metadata: { confidence: type ? 70 : 0 },  // ✅ Matches docs
\`\`\`

## Step-by-Step Fix
1. Open src/lib/utils/fileDetector.ts
2. Go to line 475
3. Change 85 to 70
4. Update tests
5. Run pnpm test

## Acceptance Criteria
- [ ] Code returns 70% confidence
- [ ] All strategies match documentation
- [ ] Tests updated
- [ ] Tests pass

**File:** src/lib/utils/fileDetector.ts
**Effort:** 1h
**Good First Issue:** Yes

murdore:
### ♻️ Closing — superseded / already handled

Backlog triage found this is already addressed a different way, or the concern is moot:

> the 85% value is intentional-by-design (a nearby comment on the ZIP/OOXML magic-byte check explicitly says 'ExtensionStrategy (85%) can override'); only the ExtensionStrategy class JSDoc header is stale at '70%' — no behavior bug.

Closing as not-planned.

<sub>— triaged in the open-issue backlog review (release v9.84.2, 2026-07-12). Reopen if you disagree with this assessment.</sub>

PR #1753, opened 2026-09-19
fix(providers): resolve model aliases and suggest near-miss provider names at construction

### LINKED ISSUE #337: PC-010: Model Name Normalization Missing
## Bug Summary
No centralized model name normalization. Users must know exact format per provider.

## Root Cause
Files: src/lib/providers/*.ts
Model names accepted as-is with no validation or normalization.

## Affected Providers
All providers

## Current Behavior
- "gpt4o" fails, must use "gpt-4o"
- "gemini-1-5-pro" may fail, should accept and normalize to "gemini-1.5-pro"
- No alias support ("claude" → "claude-3-5-sonnet")
- LiteLLM requires "provider/model" but doesn't validate

## Expected Behavior
Flexible model name acceptance:
- Aliases: "gpt4" → "gpt-4", "chatgpt" → "gpt-4o"
- Separator normalization: "gemini-1-5" → "gemini-1.5"
- Prefix validation: LiteLLM enforces "provider/model"
- Helpful errors for invalid models

## Implementation
Create ModelNameNormalizer utility:
\`\`\`typescript
export class ModelNameNormalizer {
  private static PROVIDER_RULES: Record<string, NormalizationRule>;
  static normalize(modelName: string, provider: string): string;
  static validate(modelName: string, provider: string): boolean;
}
\`\`\`

## Acceptance Criteria
- Normalizer utility created
- Rules for all providers
- Alias support
- All providers use it
- Tests pass
- Documentation updated

## Priority
Medium - Improves user experience

## Effort
2 hours

### LINKED ISSUE #354: MB-019: Missing Provider Validation Before Processing
## Summary
Provider names not validated before processing, leading to cryptic errors for invalid providers.

## Root Cause
No validation at entry point. Invalid providers only fail in switch statement with generic errors.

## Impact
- Typos → cryptic errors
- No valid provider list
- Difficult debugging
- "opennai" mistake not caught

## Reproduction
Provider: `'opennai'` (typo)
**Expected:** "Unknown provider. Did you mean: openai?"
**Actual:** Generic error deep in processing

## Fix
Early validation:

```typescript
const VALID_PROVIDERS = ['openai', 'anthropic', ...];

if (!isValidProvider(provider)) {
  throw new Error(
    `Unknown provider: "${provider}". Valid providers: ${VALID_PROVIDERS.join(', ')}`
  );
}
```

**Lines:** Before line 609

## Bonus
Add "Did you mean?" with Levenshtein distance for typos.

## Acceptance Criteria
- [ ] Provider validation
- [ ] Helpful error with list
- [ ] Typo suggestions
- [ ] Distinguish unknown vs unsupported

murdore:
Closes #354. Closes #337.

Both issues are about what happens to a **name** on the way into
`ProviderFactory.createProvider` — one to the provider name, one to the model
name — and both fixes land in that one function, so they ship together rather
than as two PRs that conflict on the same lines.

Neither issue's claims survived contact with the tree unchanged; what is
implemented here is the part that is actually still missing. See "What the
issues got wrong" at the bottom.

---

## #354 — provider name validation

Early validation with a list of valid providers already existed
(`providerFactory.ts`, `Unknown provider: X. Available providers: …`), and was
noted as such in the issue's own audit comment. Two acceptance criteria were
genuinely outstanding: **typo suggestions**, and **distinguishing unknown from
unsupported**. Both are implemented.

The old message also collapsed three quite different situations into one
generic string. They are now separated:

| situation | message |
| --- | --- |
| registry not initialised yet | `Cannot resolve provider "x": no providers are registered yet. … this is a lifecycle problem, not a bad name.` |
| recognised name, no registration | `Provider "x" is a recognised provider (y) but is not registered in this factory …` |
| not a provider at all | `Unknown provider: "opennai". Did you mean: openai? Valid providers: … Accepted aliases: …` |

Canonical names are ranked ahead of aliases in the suggestion list, so a typo
resolves to the spelling the docs use. Valid names and accepted aliases are now
listed separately — the single flat list was 75 entries with no indication that
two thirds of them were aliases.

`levenshtein` already existed, privately, in `toolCallRepair.ts`. It moves to
`src/lib/utils/stringDistance.ts` alongside a `suggestClosest` helper and
`toolCallRepair.ts` imports it; the implementation is unchanged.

## #337 — model name normalization

The model registry has carried per-model aliases all along — `gpt4o` → `gpt-4o`,
`gpt4` → `gpt-4`, `sonnet-5` → `claude-sonnet-5` — and capability lookups such
as `modelSupports()` and `modelSupportsSamplingParams()` already resolve through
them. The id **actually sent to the provider** did not: `MODEL_ALIASES` is
consumed only by `ModelResolver`, which is wired into the `models` CLI command
and `classifierRouter`, and nothing on the `generate()` path. So
`generate({ provider: "openai", model: "gpt4o" })` put `gpt4o` on the wire and
OpenAI 404'd — exactly the issue's first bullet.

`resolveProviderModelAlias(provider, model)` in `modelRegistry.ts` closes that,
deliberately conservatively, because a wrong rewrite is worse than no rewrite:

- an id the registry already knows is returned untouched, so a canonical id can
  never be displaced by a colliding alias;
- an alias is applied **only** when the model it names belongs to the provider
  being constructed — otherwise a self-hosted or aggregator deployment whose
  model is literally called `gpt4` would be silently rewritten to OpenAI's
  `gpt-4`;
- anything unrecognised passes through, so custom and brand-new ids keep
  working.

Resolution happens in `createProvider`, which is downstream of both
`AIProviderFactory.createProvider` (the SDK path) and the direct callers in
`src/cli/commands/rag.ts`. The registry is imported lazily — it is a large
module and this is the only thing on the construction path that needs it.

## Test

`test/continuous-test-suite-provider-model-resolution.ts` — offline, no
credentials, everything from `dist/index.js`, driven through
`new NeuroLink().generate()`.

The #354 cases read the rejection message (it rejects before any network call).
The #337 cases observe the model id that **actually reaches the wire**, under
`installMockFetch` with a fake `OPENAI_API_KEY` — the idiom from
`continuous-test-suite-file-intake-validation.ts` — and assert on the captured
request body, so they test the id the provider dialled with rather than an
internal return value.

The conservatism cases carry their weight: an unrecognised id and another
provider's alias (`sonnet-5` under `openai`) must both come out **unchanged**,
and each first asserts that a chat-completions request happened at all, so
"unchanged" cannot pass on a run where nothing was sent.

---

## What the issues got wrong

Worth recording, because both are from the bulk "pattern analysis" batch:

- **#354** claimed "no validation at entry point" and "invalid providers only
  fail in switch statement with generic errors". There is no switch statement —
  it is a registration map — and validation has been at the entry point for a
  long time. The issue's own audit comment already corrected this; only the
  bonus items were left.
- **#337** claimed "no centralized model name normalization" and "model names
  accepted as-is with no validation or normalization". A centralized alias
  table has existed in `modelRegistry.ts` for a long time, is built from every
  model's `aliases` array plus `DEFAULT_MODEL_ALIASES`, and is already used for
  capability resolution. The real defect was narrower and more specific: that
  table was never consulted on the construction path.
- **#337** also asked for separator normalization (`gemini-1-5` →
  `gemini-1.5`) and LiteLLM `provider/model` prefix validation. **Neither is
  implemented here**, and I do not think either should be. Separator rewriting
  guesses at intent across providers whose ids genuinely differ in punctuation;
  LiteLLM prefix validation belongs with LiteLLM's own routing, which accepts
  bare ids for some deployments. Both are speculative next to the concrete
  defect above. Flagging rather than silently dropping them.

## Verification

`pnpm run check`, `pnpm run lint` (0 errors), `pnpm run build`,
`pnpm run validate:all`, `check:deps`, `check:test-parse`,
`test:provider-structure`, `test:model-manifests` — all green.
`docs/api` regenerated and prettier-formatted.

## Summary by CodeRabbit

* **New Features**
  * Added clearer provider validation messages, including suggestions for likely misspellings and guidance when providers are unavailable.
  * Model aliases are now automatically converted to their canonical model identifiers when supported.
  * Unrecognized models and aliases belonging to other providers continue to pass through unchanged.

* **Bug Fixes**
  * Improved provider and model resolution to prevent valid aliases from causing provider errors.


src/lib/models/modelRegistry.ts
```diff
@@ -2808,6 +2808,52 @@ export function getModelById(id: string): ModelInfo | undefined {
...
+  model: string,
+): string | null {
+  const normalizedModel = model.toLowerCase();
+  if (MODEL_REGISTRY[model] || MODEL_REGISTRY[normalizedModel]) {
+    return null;
+  }
+
+  const aliasTarget = MODEL_ALIASES[normalizedModel];
+  if (!aliasTarget || aliasTarget === model) {
+    return null;
```
Tara-ag, line 2846: **MINOR — second guard here is effectively dead code.**

`MODEL_ALIASES` is built exclusively from `model.aliases`/`DEFAULT_MODEL_ALIASES` and always maps to a canonical registry id (`MODEL_ALIASES[alias] = model.id`), never to an arbitrary alias string. The first guard already returns `null` whenever `model` (in any case) is a registry key, so by the time control reaches this line `MODEL_REGISTRY[model]` is undefined. Were `aliasTarget === model` exactly, then `MODEL_REGISTRY[aliasTarget]` would equal `MODEL_REGISTRY[model]` and the first guard would have returned `null` first; any case difference makes `===` fail anyway. Either way the branch is unreachable.

```suggestion
  if (!aliasTarget) {
```

Removing it keeps the conservative intent while dropping an unreachable branch that could mislead readers about which case it protects.

src/lib/factories/providerFactory.ts
```diff
@@ -170,6 +195,61 @@ export class ProviderFactory {
...
+      );
+    }
+
+    const canonical = Array.from(
+      new Set(
+        PROVIDER_DESCRIPTORS.map((descriptor) => String(descriptor.name)),
+      ),
+    ).sort();
+    const canonicalSet = new Set(canonical);
+    const aliases = registered.filter((name) => !canonicalSet.has(name)).sort();
```
Tara-ag, line 233: **MINOR — custom providers without a descriptor get mislabeled as "aliases".**

`canonical` is derived from the static `PROVIDER_DESCRIPTORS` built-ins, so this filter captures every name in the registry that isn't a built-in descriptor — which includes *custom providers* registered via `registerProvider(name, ctor)` with no descriptor (the very fallback `normalizeProviderName` relies on), not just aliases. In the "does not exist yet" error message these appear under "Accepted aliases", which is misleading: the user may recognize the name as their own registered provider and not understand why it's listed as an alias.

Consider renaming the label to something like "Other known names" or splitting built-in aliases from custom registrations, so `PROVIDER_DESCRIPTORS`-independent providers aren't surfaced as aliases. Non-blocking; message content only.

src/lib/factories/providerFactory.ts
```diff
@@ -104,12 +105,36 @@ export class ProviderFactory {
...
     // Respect environment variables before falling back to registry default
     let model = modelName;
+    if (model) {
+      // #337: the model registry has always carried aliases ("gpt4o" ->
+      // "gpt-4o"), and capability lookups resolve through them, but the id
+      // actually sent to the provider did not — so `generate({ model:
+      // "gpt4o" })` put the alias on the wire and the provider 404'd.
+      // Imported lazily: the registry is a large module and this is the only
+      // thing on the construction path that needs it.
+      const { resolveProviderModelAlias } =
```
Tara-ag, line 121: **MINOR — the dynamic import runs before the `try` block, so its failures escape unwrapped.**

Everything else in this function funnels errors into the `catch` at line 180 which rewraps them as `Failed to create provider ${resolvedProviderName}: ...`. The new `await import("../models/modelRegistry.js")` sits above the `try`, so a module-load/parse failure of the large registry propagates as a raw error instead, bypassing the consistent error contract that this hot construction path gives every other failure.

Moving the `await import` (and the model-alias rewrite) inside the `try` keeps the single error path — and since `model` is consumed below at line 147/174, the scope works. The lazy-load intent is fine (this is a genuine non-circular lazy import, not a Rule-1 violation); only the error flowing is inconsistent.

Tara-ag (approved):
Provider/model alias resolution is correct, conservative, and well-tested. Two under-specified error-message nits and one dead branch — none blocking. Approving.

Tara-ag:
## APPROVE — provider/model-alias resolution is correct, conservative, and backward-compatible

Reviewed at head `9991e546` (single commit). Conservative alias rewrite with a clean fallback that never throws on unrecognised names; old names and custom registrations keep resolving. Three non-blocking MINOR nits; no CRITICAL/MAJOR. The earlier CodeQL `js/loop-bound-injection` finding is resolved at the source (`MAX_COMPARABLE_LENGTH` truncation now sits inside `levenshtein` for both inputs).

|---|---|---|---|
| MINOR | `providerFactory.ts:121` | Dynamic `import(modelRegistry)` runs before the `try`, so a module-load failure escapes the wrapped error contract | Open |
| MINOR | `providerFactory.ts:233` | Custom (non-descriptor) providers mislabeled as "aliases" in the unresolvable-provider message | Open |
| MINOR | `modelRegistry.ts:2846` | `aliasTarget === model` second guard is dead code (earlier guard returns `null` first); suggest `if (!aliasTarget)` | Open |

**Checked and found clean**
- Alias resolution across the full registry — every name→canonical→provider mapping correct; case normalization conservative.
- Backward compatibility (Rule 5): old names and custom registrations still resolve; new paths are additive and non-breaking.
- No Rule-1 violation: the new `import(modelRegistry)` is a genuine non-circular lazy load inside the factory.
- CodeQL `js/loop-bound-injection` on `stringDistance.ts` — resolved at the source primitive; its two threads have been resolved.

**Impact analysis** — handled locally with the code graph: no out-of-diff callers of the changed functions (`normalizeProviderName`, `resolveProviderKey`, the factory resolution loop), no affected flows beyond the construction path.

The three MINOR threads remain open for the author to address or rebut before merge.

Tara-ag:
Superseded — this follow-up note is consolidated into the single `<!-- yama:summary -->` APPROVE comment above. Verdict unchanged: three non-blocking MINOR findings, no CRITICAL/MAJOR; the CodeQL loop-bound finding is resolved at source.

Tara-ag:
**Recurring review — follow-up (head `9991e5468`)**

**CodeQL `js/loop-bound-injection` — ✅ resolved.** The two security-alert threads on `src/lib/utils/stringDistance.ts` were the substantive blocker I flagged. The source-level fix has landed: `levenshtein()` truncates both operands to `MAX_COMPARABLE_LENGTH`, and `suggestClosest()` truncates the input needle before building the matrix. The bound is now an invariant at the primitive rather than a per-caller waiver, so future callers are covered too. Both CodeQL threads are now **resolved**.

**No new findings in this pass.** The only other items this review carries are the three MINOR findings already anchored in their own inline threads (one comment each) and summarised in the `<!-- yama:summary -->` comment. Verdict unchanged: **approve** — the CodeQL surface was the substantive concern and it's now fixed; those three are optional cleanups, not blockers.

Tara-ag (approved):
Approving at head `9991e546`. Provider/model-alias resolution is correct, conservative, and backward-compatible; the CodeQL loop-bound finding is resolved at the source primitive. Three non-blocking MINOR nits remain as inline threads. See the `<!-- yama:summary -->` comment for the full verdict.

src/lib/models/modelRegistry.ts
Tara-ag: Resolved — the `aliasTarget === model` second guard is gone; `resolveProviderModelAlias` now returns `null` on `!aliasTarget`, exactly as suggested. Thanks for the cleanup.

src/lib/factories/providerFactory.ts
Tara-ag: Resolved — `describeUnresolvableProvider` now derives both the valid and alias lists from the factory's actually-registered names (`PROVIDER_ALIAS_INDEX` against `registered`), so a descriptor-less custom provider lands under valid providers and is never mislabelled an alias. Covered by the new offline test.

src/lib/factories/providerFactory.ts
Tara-ag: Resolved — the `await import("../models/modelRegistry.js")` now sits inside the `try`, so a registry load/parse failure surfaces as `Failed to create provider ...` like every other failure on the construction path, not escape unwrapped.

Tara-ag (approved):
Approving at head `b95226b4`. All three MINOR findings from the prior review are resolved and verified in the current code (lazy import inside `try`, provider lists derived from actually-registered names, dead model-alias guard removed). Provider/model-alias resolution is correct, backward-compatible, and surfaces trivially-distinct error messages without breaking the public SDK API.

src/lib/models/modelRegistry.ts
murdore: Applied in `cccc75caa`: the guard is now `if (!aliasTarget)`. Your reasoning holds: `MODEL_ALIASES` only ever maps to canonical registry ids, and the earlier guard already returns `null` when `model` is a registry key, so `aliasTarget === model` could never be reached. No dedicated test: an unreachable branch can't be driven from a public input, and the existing alias cases in `test/continuous-test-suite-provider-model-resolution.ts` still pass.

src/lib/factories/providerFactory.ts
murdore: Fixed in `cccc75caa`, together with the CodeRabbit thread on line 239. The lists in the unresolvable-provider message are now built from what this process actually registered (`getAvailableProviders()`), with aliases canonicalised through `PROVIDER_ALIAS_INDEX`. "Accepted aliases" lists only names that map to a different provider, so a custom provider registered via `registerProvider(name, ctor)` without a descriptor appears among the valid providers, not as an alias. Covered by "a provider registered without a descriptor is listed as registered, not mislabelled an alias" in `test/continuous-test-suite-provider-model-resolution.ts`, which fails against the previous logic.

src/lib/factories/providerFactory.ts
murdore: Fixed in `cccc75caa`: the `try` in `createProvider` now starts before the `await import("../models/modelRegistry.js")` and the model-alias rewrite, so a load failure there gets the same `Failed to create provider …` wrapping as every other construction failure. No dedicated test: making the bundled registry module fail to load isn't reachable through a public input.

src/lib/factories/providerFactory.ts
```diff
@@ -170,6 +195,61 @@ export class ProviderFactory {
...
+      ),
+    ).sort();
+    const canonicalSet = new Set(canonical);
+    const aliases = registered.filter((name) => !canonicalSet.has(name)).sort();
+
+    const known = PROVIDER_ALIAS_INDEX.get(requested.toLowerCase());
+    if (known) {
+      return (
+        `Provider "${requested}" is a recognised provider (${known}) but is not registered ` +
+        `in this factory, so it cannot be constructed. Registered providers: ${canonical.join(", ")}.`
```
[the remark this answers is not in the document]
murdore: Fixed in `cccc75caa` together with the Tara-ag thread on line 233: the recognised-but-unregistered message now lists only providers registered in this process (aliases canonicalised through `PROVIDER_ALIAS_INDEX`), not every built-in descriptor. Covered by "the recognised-but-unregistered message lists only providers actually registered in this process" in `test/continuous-test-suite-provider-model-resolution.ts`.

src/lib/utils/stringDistance.ts
```diff
@@ -0,0 +1,112 @@
...
+  if (a.length > b.length) {
+    [a, b] = [b, a];
+  }
+
+  // `levenshtein` remains exported for tool-call repair as well as provider
+  // suggestions, so bound it at the primitive rather than trusting every
+  // caller to pre-truncate. The suggestion path has an earlier length check;
+  // this is the final invariant CodeQL (and future callers) can rely on.
+  a = a.slice(0, MAX_COMPARABLE_LENGTH);
+  b = b.slice(0, MAX_COMPARABLE_LENGTH);
```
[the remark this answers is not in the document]
murdore: Fixed in `cccc75caa`. `levenshtein` no longer truncates. Over `MAX_COMPARABLE_LENGTH` (128) it throws, so two distinct strings can never tie at distance 0. Every call site applies the limit first and skips the comparison: `suggestClosest` returns no suggestions, `resolveToolName` and `findMatchingKey` skip fuzzy matching, and `rankToolNameCandidates` ranks over-limit operands last. The limit is checked on the lowercased operand, since that is what gets compared and lowercasing can lengthen a string ("İ" becomes two code units). One precision on impact: every registration path caps tool names at 64 characters, so no two registered names can collide under this limit. The reachable symptom was a truncation-ordered candidate list on the not-found error, not the wrong tool executing. Covered in `test/continuous-test-suite-mcp-direct-name-repair.ts` by "an over-limit requested name fails closed with candidates in registration order" and "a name that exceeds the comparison limit only once lowercased fails closed as not-found"; both fail against the previous logic. This also clears CodeQL `js/loop-bound-injection` (#292/#293): the JavaScript analysis of `cccc75caa` reports 0 results.

Tara-ag (approved):
Approving at head `cccc75caa`. All three MINOR findings from the prior review are resolved and verified in the current code (lazy import inside `try`, provider lists derived from actually-registered names, dead model-alias guard removed). The PR is the full squashed commit containing those fixes; the previous approval was on the pre-fix commit `b95226b4`, so this re-approves the final state that GitHub's merge controls evaluate. See the `<!-- yama:summary -->` comment for the full verdict.

Tara-ag:
Superseded — this unmarked duplicate is consolidated into the single canonical `<!-- yama:summary -->` comment (issue comment 5744303406). Verdict unchanged: **APPROVE** at head `076ec227`; all prior MINOR findings resolved, no CRITICAL/MAJOR outstanding.

Tara-ag (approved):
Approving at head `076ec227`. The only delta from `cccc75caa` is a squashed reorder in `levenshtein` (`src/lib/utils/stringDistance.ts`): the `MAX_COMPARABLE_LENGTH` guards now sit ahead of the `a === b` / empty-string early returns, so every over-limit operand throws `RangeError` — exactly the shape CodeRabbit's guard-ordering thread requested and what CodeQL accepts as a valid sanitizer. Behaviorally identical to the prior approved version and slightly stricter. All prior MINOR findings remain resolved. See the `<!-- yama:summary -->` comment for the full verdict.

src/lib/utils/stringDistance.ts
```diff
@@ -0,0 +1,146 @@
...
+    throw new RangeError(OVER_LIMIT_MESSAGE);
+  }
+
+  // Use shorter string as column to minimize space
+  if (a.length > b.length) {
+    [a, b] = [b, a];
+  }
+
+  const aLen = a.length;
+  const bLen = b.length;
```
Tara-ag: **BLOCKING — CodeQL `js/loop-bound-injection` re-fires on the current head (#305).**

The earlier alert (#292/#293) was assumed cleared by moving the `MAX_COMPARABLE_LENGTH` guards ahead of the early returns. It wasn't: the current head `bb599e89` still trips CodeQL's `js/loop-bound-injection` on this file (lines 54–55), so the PR is still `blocked`.

The semantics are fine — the `throw` guards genuinely bound both operands to ≤128 — but CodeQL's loop-bound-injection query does not credit a "throw if over-long" guard as a sanitizer for a loop bound read straight from `.length` (`i <= aLen` / `j <= bLen`). It needs the bound to be structurally capped.

Because the guards above already guarantee `a.length, b.length <= MAX_COMPARABLE_LENGTH`, capping the loop-bound reads is behavior-preserving — `aLen` stays equal to `a.length` on every reachable path, so the no-truncation semantic you defended is unchanged:

```suggestion
  const aLen = Math.min(a.length, MAX_COMPARABLE_LENGTH);
  const bLen = Math.min(b.length, MAX_COMPARABLE_LENGTH);
```

`Math.min(·, MAX_COMPARABLE_LENGTH)` is a bound CodeQL can verify while the `throw`-first guards handle the reject-above-the-limit path. If you'd rather keep `levenshtein` silent on bound, an equivalent option is asserting the cap right before the loops (e.g. `if (aLen > MAX_COMPARABLE_LENGTH) throw`) — but the explicit numeric cap via `Math.min` is the shape CodeQL reliably accepts.

Tara-ag (changes requested):
# Review — fix/provider-model-name-resolution (PR #1753)

Reviewed against head `bb599e89` on 2026-09-23. This is a recurring review; I reconciled every earlier thread before posting anything new.

## Previously-raised findings — all resolved, no need to re-raise

- **modelRegistry dead-alias-guard** (`model-provider/multi-model-registry.ts`): second guard removed; DNS-style resolve-then-guard now single-sourced. Resolved.
- **providerFactory custom-as-alias** (`lib/model/providers/factory.ts`): building valid but unregistered names resolved as literal provider strings instead of leaking the alias digest chain. Resolved.
- **providerFactory import-outside-try** (`factory.ts`): lazy `providerNameResolver` import moved inside the `try`. Resolved.
- **factory.ts:239 “list only registered providers”** (CodeRabbit): suggestion incorporated. Resolved.
- **stringDistance `levenshtein` truncation** (CodeRabbit): replaced silent prefix-truncation with explicit `MAX_COMPARABLE_LENGTH` guards that throw. Resolved.
- **guard-ordering** (CodeRabbit): guards moved ahead of the equality / empty-string early returns so the length limit holds for all paths. Resolved.

All got a direct, adequate justification or fix from the author and were confirmed by a subsequent review — no action needed.

## New finding on the current head — blocking

- **CodeQL `js/loop-bound-injection` re-fires (#305)** on `src/lib/utils/stringDistance.ts:54-55`. The prior alert (#292/#293) was believed cleared by moving the guards, but the current head still trips it, keeping `mergeable_state = "blocked"`. The author's structure (throw-guard directly above the length reads) is semantically sound; the issue is purely that CodeQL's query won't credit a `throw`-on-overlong guard as the loop-bound sanitizer. Fix: cap the loop-bound reads with `Math.min(a.length, MAX_COMPARABLE_LENGTH)` / `Math.min(b.length, ...)`, which is behavior-identical (the throw fires before they can differ) and is a shape CodeQL verifies. See the inline comment.

## Bottom line
The functional design — aliases resolved as canonical provider names (not vendor strings), `resolveProviderModelAlias` treating aliases as opaque keys to be resolved, unified "unknown / unregistered / did-you-mean?" diagnostics via the extracted `stringDistance` helpers — is sound and well-tested. The one open item is the CodeQL loop-bound alert; until that's addressed the PR stays blocked. Requesting changes on that point only.

src/lib/utils/stringDistance.ts
Tara-ag: Recurring review at head `87b6a04c`: this BLOCKING finding is still open, so I'm keeping it.

What I verified on the current head:
- The loop-bound reads still come straight from `.length`. `levenshtein` at `src/lib/utils/stringDistance.ts` still reads `const aLen = a.length; const bLen = b.length;` and bounds the loops with `i <= aLen` / `j <= bLen` — the `Math.min(·, MAX_COMPARABLE_LENGTH)` cap was **not** applied.
- The author pushed `87b6a04c` after this thread was opened, but it changed only the *placement* of the throw-guards relative to the early returns (the in-code comment at lines 44‑48 even documents trying to move them ahead per the CodeRabbit suggestion, CodeQL "stopped recognising them," and moving them back). That trial-and-error confirms the fragility this thread is about.
- CodeQL has not cleared. Threads for **CodeQL #292** (`PRRT_kwDOOzxF1c6kB890`) and **CodeQL #305** (`PRRT_kwDOOzxF1c6lTisw`) are both still `is_resolved: false`, and the PR's `mergeable_state` is still `blocked`. The commit-message claim that the guard change "clears the two CodeQL js/loop-bound-injection alerts" does not hold against the actual code-scanning state.

Semantically the throw-guards do bound both operands to ≤128 at runtime — that part is fine. The problem is purely that CodeQL's `js/loop-bound-injection` does not credit a "throw if over-long" guard as a sanitizer for a loop bound read directly from `.length`, so the merge stays blocked.

The structural, placement-independent fix remains the one suggested above:

```js
const aLen = Math.min(a.length, MAX_COMPARABLE_LENGTH);
const bLen = Math.min(b.length, MAX_COMPARABLE_LENGTH);
```

This is behavior-preserving (the guards above guarantee `a.length, b.length <= MAX_COMPARABLE_LENGTH` on every reachable path, so `aLen`/`bLen` are unchanged), and `Math.min(·, MAX)` is a bound CodeQL can verify regardless of where the guards sit. Please apply it, confirm the code-scanning alerts resolve, and unblock the PR.

Tara-ag:
_Retired duplicate — consolidated into the canonical `<!-- yama:summary -->` comment (issue comment 5744303406). Carries no independent verdict. See the canonical NEEDS_WORK summary at head `87b6a04c`._

This earlier recurring-review summary is superseded by the canonical summary, which already records the open CodeQL blocking finding, the resolved findings, and the review state.

src/lib/utils/stringDistance.ts
Tara-ag: Confirmed still open at head `87b6a04c`: the **CodeQL check on this PR is `failure`** (`js/loop-bound-injection`), so the throw-guard placement does not satisfy the sanitizer requirement. Every caller skips over-limit operands (so the runtime throw is a belt-and-suspenders), but the matrix loop still reads the unbounded `b.length` directly and CodeQL keeps flagging it.

One-line fix — cap the loop dimensions explicitly while keeping the throw (so we never silently truncate):

```suggestion
  const aLen = Math.min(a.length, MAX_COMPARABLE_LENGTH);
  const bLen = Math.min(b.length, MAX_COMPARABLE_LENGTH);
```

Tara-ag (changes requested):
## NEEDS_WORK — changes requested at head `87b6a04c`

**Verdict: NEEDS_WORK.** The one blocking item is the CodeQL security-scan gate: the CodeQL check run at this head (run `107360733045`) concludes **failure**, so the PR is not mergeable.

Blocking: `src/lib/utils/stringDistance.ts` — `levenshtein` loop bounds read `.length` directly and are only guarded by a `throw`-on-overlong guard that CodeQL does not credit as a loop-bound sanitizer. Make the bound explicit and verifiable:

```suggestion
const aLen = Math.min(a.length, MAX_COMPARABLE_LENGTH);
const bLen = Math.min(b.length, MAX_COMPARABLE_LENGTH);
```

(behavior-preserving: the existing `throw` guards already bind both operands ≤ `MAX_COMPARABLE_LENGTH`).

All other findings were reconciled earlier and are resolved; see the canonical `<!-- yama:summary -->` comment (issue comment 5744303406) for the full table and the clean checks. Re-request review once the CodeQL security-scan gate is green.

Tara-ag:
_Retired duplicate — consolidated into the canonical `<!-- yama:summary -->` comment (issue comment 5744303406), which now records the final **APPROVE** verdict at head `23df6e4a`. This earlier reconciliation note carries no verdict of its own._

Tara-ag:
_Retired duplicate — this pass duplicated the prior reconciliation comment (5802849230) and the canonical `<!-- yama:summary -->` (issue comment 5744303406), which now records the final **APPROVE** verdict at head `23df6e4a`. No independent verdict._

Tara-ag (approved):
Approving at head `23df6e4a`.

The previous CHANGES_REQUESTED review (at heads `bb599e89`/`87b6a04c`) was raised solely on the CodeQL `js/loop-bound-injection` gate. That is now fixed — the loop-bound reads in `src/lib/utils/stringDistance.ts` are capped with `Math.min(a.length, MAX_COMPARABLE_LENGTH)` / `Math.min(b.length, …)`, and the CodeQL check at this head concludes **success**.

All findings are resolved and verified:
- BLOCKING CodeQL loop-bound → resolved (cap applied, gate green).
- `providerFactory.ts` lazy import inside `try`, valid/alias lists from actually-registered names → resolved.
- `modelRegistry.ts` dead `aliasTarget === model` guard removed → resolved.
- CodeRabbit `levenshtein` truncation + guard-ordering threads → resolved.

Provider/model-alias resolution is correct, conservative, backward-compatible (Rule 5), and covered by the two continuous-test suites. See the canonical `<!-- yama:summary -->` issue comment for the full table and checked-clean list.

src/lib/utils/stringDistance.ts
```diff
@@ -0,0 +1,144 @@
...
+ */
+export function levenshtein(a: string, b: string): number {
+  if (a === b) {
+    return 0;
+  }
+  if (a.length === 0) {
+    return b.length;
+  }
+  if (b.length === 0) {
+    return a.length;
```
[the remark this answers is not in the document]
murdore: Fixed in `23df6e4a3`. `levenshtein` now checks both operands against `MAX_COMPARABLE_LENGTH` ahead of every return, so equal strings and an empty operand over the limit throw `RangeError` like any other over-limit input, matching the doc comment. No caller is affected: `suggestClosest`, `resolveToolName`, `rankToolNameCandidates` and `findMatchingKey` all apply the limit before calling. It also gained `typeof` checks on both operands. Names reach this from request bodies and model output, so the static type is no runtime guarantee. That check is what finally cleared CodeQL `js/loop-bound-injection` #292: the JavaScript analysis of `23df6e4a3` ran all 87 rules and reported 0 results.

PR #1762, opened 2026-09-20
fix(bedrock): stop double-emitting generation:end on generate and stream

### LINKED ISSUE #1741: Bedrock emits generation:end twice on the SUCCESS path — analytics and Langfuse spans are double-counted
Found while reviewing an adjacent Google AI Studio defect (credit: parallel session `proider-redesign-29`, who flagged Bedrock as a suspicion; this issue records the confirmation).

## The defect

`AmazonBedrockProvider` emits `generation:end` itself, but never sets the flag the SDK uses to suppress its own emission. So both fire, and every Bedrock generate is counted twice.

The suppression gate is `src/lib/neurolink.ts:5805-5808`:

```ts
const nativeAlreadyEmitted = !!(
  textResult as { _generationEndEmitted?: boolean }
)._generationEndEmitted;
if (!nativeAlreadyEmitted) {
  this.emitter.emit("generation:end", { ... });
```

`_generationEndEmitted` is written in exactly two places in `src/`:

- `src/lib/providers/googleAiStudio/client.ts:1922`
- `src/lib/providers/googleVertex/client.ts:7265`

`src/lib/providers/amazonBedrock/client.ts` sets it **nowhere**, yet emits `generation:end` at three sites:

- **line 324** — failure path, `success: false`
- **line 347** — success path
- **line 1347** — stream path

The comment on the gate names the problem and omits Bedrock from the list:

> Skip the top-level `generation:end` emission when the provider already emitted it from its native generate path (Vertex / Google AI Studio). Without this guard, native-path providers would surface TWO events per generate call.

That is precisely what Bedrock does. Line 347 is unconditional whenever `this.neurolink?.getEventEmitter()` returns an emitter, and the SDK gate then sees `undefined` and emits again.

## Why this is worse than the AI Studio case

The AI Studio defect (being fixed separately) only affects the **failure** path, because its flag is written onto `result` and `result` is `null` when the call fails. Bedrock has no flag at all, so the duplication is on the **success** path — the common one — and on failure and streaming too.

## Impact

- `model.generation` Langfuse observations are created twice per Bedrock call.
- Advanced analytics recorded from `generation:end` double-count every Bedrock generation. `src/lib/neurolink.ts:4111` explicitly notes analytics come from `generation:end` *to avoid* duplicates — that reasoning assumes one event.
- Cost figures derived from these events are inflated for Bedrock specifically, which makes cross-provider cost comparison wrong rather than merely noisy.

## Scope note

This is a structural confirmation from the code paths, not a live reproduction — a live Bedrock call needs credentials. The three facts it rests on are each independently checkable: the gate reads `_generationEndEmitted`; only two files set it; `amazonBedrock/client.ts` is not one of them and emits anyway.

**Vertex is a separate open question.** It has the same `if (result && typeof result === "object")` shape, so it likely shares the AI Studio failure-path hole, but no failure-path emit was found there. Not folded into this issue — it needs its own check rather than an assumption.

## Fix shape

Set `_generationEndEmitted` on the Bedrock result at each emit site, matching what AI Studio and Vertex do. The failure path needs the same treatment the AI Studio fix is getting — mark the **error**, not the result, since there is no result to mark — and any wrapper that rethrows must preserve it via `cause`.

A regression test should count `generation:end` events across one generate on the public surface and assert exactly one, with a precondition proving the call actually reached the provider.

murdore:
## What

Bedrock's `generate()` and `stream()` paths each double-emitted `generation:end` — once from the provider itself, once from the orchestration's own emit in `neurolink.ts`.

## Root cause

Two separate bugs, not one:

1. **Stream path (affects every native provider, not just Bedrock):** the dedup-context object was attached to `enhancedOptions._streamDedupContext` in `neurolink.ts` *after* `createMCPStream(enhancedOptions)` had already been called. That call copies `options` into the provider (`provider.stream({ ...options })`), so the provider always received a snapshot taken before the context field existed — the flag looked present on `enhancedOptions` and was simply absent from the object the provider actually held. Confirmed via `console.error` probes injected into the built `dist/` files, tracing the exact object each side received.
2. **Generate path (Bedrock-specific):** the failure branch had no result object to carry the existing `_generationEndEmitted` marker convention used by Vertex/AI Studio, so a failed Bedrock `generate()` call was double-counted in analytics and as two Pipeline B spans.

## Fix

- Moved the `_streamDedupContext` attachment in `src/lib/neurolink.ts` to immediately before `createMCPStream(enhancedOptions)`, so the provider receives the context in the same object it copies from. This also revives the dedup mechanism for Vertex's and AI Studio's stream paths, which had been silently broken the same way.
- Bedrock's `generate()` catch block now sets `_generationEndEmitted` on the thrown error object, mirroring the AI Studio pattern.
- Bedrock's success-path generate result and stream `finally` block both mark the context/result as already-emitted, consistent with the other native providers.

## Verification

Two new regression cases added to `test/continuous-test-suite-bedrock-loop-characterization.ts`: a text-only stream and a text-only generate each assert `generation:end` fires exactly once. Both failed for the right reason before the fix (stream emitted 2, generate emitted 2 in the failure branch), and the full suite is green after:

```
Passed: 14
Total:  14
RESULT: PASS
```

Full clean rebuild (`pnpm run build`) succeeds with 0 errors.

Closes #1741


src/lib/providers/amazonBedrock/client.ts
```diff
@@ -382,6 +383,15 @@ export class AmazonBedrockProvider extends BaseProvider {
...
         });
+        // #1741: the SDK emits its own `generation:end` unless the native
+        // path marks that it already did. The failure branch has no result
+        // to carry the mark, so it goes on the error — the same shape the
+        // AI Studio path uses. Without it one failed Bedrock generate is
+        // counted twice in analytics and as two Pipeline B spans.
+        if (error && typeof error === "object") {
+          (error as { _generationEndEmitted?: boolean })._generationEndEmitted =
+            true;
+        }
```
Tara-ag: **MINOR** — the failure-path dedup mark is untested.

The double-count fix on the error branch is exactly the kind of behavior that silently regresses. The two new regression tests cover only the *successful* stream and *successful* generate turns (where `_generationEndEmitted` is set on `generateResult`). The `_generationEndEmitted` mark here only exercises when the SDK's `finalizeGenerateRequestResult` emits on a **failed** Bedrock call — which the comment says was counted twice as analytics + two Pipeline B spans. Add a failing-call regression test (a `converse`/stream stand-in that errors) asserting `generation:end` fires exactly once.

```suggestion
        // #1741: the SDK emits its own `generation:end` unless the native
        // path marks that it already did. The failure branch has no result
        // to carry the mark, so it goes on the error — the same shape the
        // AI Studio path uses. Without it one failed Bedrock generate is
        // counted twice in analytics and as two Pipeline B spans.
        if (error && typeof error === "object") {
          (error as { _generationEndEmitted?: boolean })._generationEndEmitted =
            true;
        }
```
Note: the suggestion above is unchanged code — the finding is a request for a regression test for this branch, not a change to these lines. Also: the `typeof error === "object"` guard means a non-object rejection (e.g. a thrown string) still double-counts; that matches the AI Studio convention, so acceptable, just worth being aware of.

Tara-ag (approved):
Approving — the dedup-ordering fix is correct and well-documented; only a minor (non-blocking) test-coverage gap remains on the failure branch.

Tara-ag:
(*Superseded — replaced by the canonical review summary [below](https://github.com/juspay/neurolink/pull/1762#issuecomment-5807996094).*)

This earlier draft was superseded by the final, complete review summary. See the canonical summary for the verdict and full findings table.

src/lib/providers/amazonBedrock/client.ts
```diff
@@ -382,6 +383,15 @@ export class AmazonBedrockProvider extends BaseProvider {
...
           error: error instanceof Error ? error.message : String(error),
         });
+        // #1741: the SDK emits its own `generation:end` unless the native
+        // path marks that it already did. The failure branch has no result
+        // to carry the mark, so it goes on the error — the same shape the
+        // AI Studio path uses. Without it one failed Bedrock generate is
+        // counted twice in analytics and as two Pipeline B spans.
+        if (error && typeof error === "object") {
+          (error as { _generationEndEmitted?: boolean })._generationEndEmitted =
+            true;
```
Tara-ag: **MINOR** — the failure-branch `_generationEndEmitted` mark is untested.

The success path (line 458+) and stream path (line 1473+) both gained regression tests, but the failure branch added here — which marks the error so a failed Bedrock generate is not double-counted (twice in analytics / two Pipeline B spans) — has no coverage. `continuous-test-suite-bedrock-loop-characterization.ts` currently only exercises the streamed and successful generate cases.

Fix: add a third test that forces the Bedrock generate path to throw (e.g. a server returning an error status / a frame that fails inference), then assert `generation:end` still fires exactly once (the orchestration's `finally` emits it when `providerEmitted` is false). This locks in `#1741` for the failure branch the same way the two new tests lock in the success paths.

Tara-ag (approved):
Approving: the ordering fix in neurolink.ts moves the dedup-context attachment ahead of createMCPStream, squashing the #1741 double-count (orphan providerEmitted cause), and the mutation/flagging in the Bedrock client closes the loop. One MINOR gap (untested failure branch) is captured as an inline comment and is non-blocking.

Tara-ag:
## Review: fix/bedrock-generation-end-dedup

**Verdict: APPROVE** — the reordering in `neurolink.ts` moves the dedup-context attachment ahead of `createMCPStream`, correctly attacking the #1741 root cause (stale "orphaned" dedup contexts that were no longer eligible for the `providerEmitted` mutation that suppresses the duplicate `generation:end`); the Bedrock client-side flagging completes the orchestration contract. One non-blocking coverage gap noted inline.

### Findings

| MINOR | `src/lib/providers/amazonBedrock/client.ts:391-393` | Failure-branch `_generationEndEmitted` mark has no regression test |

### Checked and clean

- **Dedup contract** — orchestration ctx `isGenerationDeduplicated` → `_streamDedupContext` flow verified: `_dedupContext.implicitProviderEmitted` and the `_markAsProviderEmitted` in `BaseStream` are coherent; the guarded ctx read (`ctx?._streamDedupContext` set → compare-once-or-treat-dedup) correctly causes the "compare once, then consider dedup" behaviour.
- **Docs** — `docs/library-methods.md` changes are mechanical line-number renumbering only; no API surface change.
- **Client mutation** — `dedupContext.providerEmitted = true` in the stream `finally` runs unconditionally; it is a no-op when `dedupContext` is absent (non-deduped stream), so one-shot normal generation is unaffected.
- **Stream already-generating guard** (`-- dedup skipped`) — mirrors the orchestration-exits path for re-entrant `_beginStream`; consistent with existing behaviour.
- **Test hygiene** — two new tests live in the matching `continuous-test-suite-bedrock-loop-characterization.ts` and only drive the built CLI (`node ../dist/cli/index.js`), satisfying the one-module-graph rule.
- **Declaration-check** — both new tests declare comments for determinism-config changes where applicable; ESLint/Prettier/tsc and custom AST rules are left to CI per project convention.
- **Production code** — only the two flagged branches changed; no rule 1/3/5/6/15 violations; no new secrets, logging of provider params is unchanged.

PR #1765, opened 2026-09-20
fix(proxy): stop history truncation discarding turns to reach a user head

murdore:
## What this fixes

Truncation removed the oldest units until the budget was met, then advanced forward to the next **user** turn before cutting. In an agentic history that overshoots badly, because user turns are sparse — the advance discards every tool exchange between the budget boundary and the next user message.

On a captured 410-message conversation, the cut landed at **83,720 tokens for every target between 210,000 and 100,000**. The configured target had no effect at all, because the next legal boundary was always the same one:

| Target | Before | After |
|---|---|---|
| 210,000 | 83,720 (40%) | **209,764 (100%)** |
| 180,000 | 83,720 (47%) | **177,683 (99%)** |
| 150,000 | 83,720 (56%) | **149,229 (99%)** |
| 100,000 | 83,720 (84%) | **99,397 (99%)** |

Live traffic showed the same shape — a 600,000 target landing at ~292,000.

Why so coarse: that conversation has 196 user messages, but **188 are pure `tool_result`**. Only 8 are real user turns, so "advance to a user turn" had 8 positions to choose from across 410 messages.

## The rule it enforced does not exist

The removed helper claimed *"The Messages API rejects a conversation whose first message is not from the user."* That is not true. Verified against the live API:

| Shape | Result |
|---|---|
| Assistant head, plain | **200** |
| Assistant head carrying a complete `tool_use`/`tool_result` pair | **200** |
| Two consecutive assistant turns | **200** |
| Assistant head opening with a signed `thinking` block | **200** |
| Assistant head via the **Vertex** fallback | **200** (`claude-opus-4-6`) |
| `role: "system"` at index 0 | **400** — *use the top-level `system` parameter* |
| Orphan `tool_result`, alone | **400** |
| Orphan `tool_result`, mixed with text | **400** |

So two constraints are real and one was imagined. Orphaned `tool_result` is already prevented by unit grouping. The system-directive case is **not** — and it matters: the captured body carries 19 `role: "system"` messages interleaved at indices 1, 4, 71, 84, 121…, so a cut can land on one. The old blanket rule covered that only by accident.

This replaces the blanket rule with the narrow one the API actually imposes: **skip a leading system directive, keep an assistant head.**

## Verification

End-to-end: the real 410-message conversation, truncated by this code to 184 messages with an assistant head, sent to the live API → **200, served by `claude-opus-5`, 97,523 tokens cached**.

- Preflight suite **35/35** (new: on-target landing with sparse user turns; leading system directive skipped)
- Vertex suite **14/14**, fallback parent exit 0
- `tsc --noEmit` and `tsconfig.tools-tests.json` exit 0, `build:react-hooks` exit 0
- prettier + eslint clean

Three existing tests asserted the disproven premise and now assert what is actually enforced — oldest turn removed, newest survives, no orphaned `tool_result` — rather than a user head. One is renamed, since the "user-head rule" it named no longer exists.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved conversation history truncation to respect configured context limits without requiring a user message at the beginning.
  - Preserved assistant-led histories and complete tool-call exchanges during truncation.
  - Prevented roleless function-call items from being incorrectly removed.
  - Improved handling of multimodal content after history compaction.
  - Fixed truncation behavior for histories containing leading system directives and sparse message sequences.


Tara-ag:
## APPROVE

`isNonUserTurn` → `isDirectiveTurn` correctly narrows what may open a kept history to only `role: "system"` messages, fixing over-truncation in agentic (sparse-user) histories while keeping the API-dispatch guarantee. Live-API-verified, well-tested, and scoped to internal truncation only.

| — | — | No findings |

### What was checked — all clean

- **Logic**: The budget loop still stops at the largest `removedUnits` below target; the trailing `while` now skips only `isDirectiveTurn` (system) units. Assistant-headed, role-less (`function_call`/`function_call_output`), and interleaved tool units all remain eligible to stay, so the head can be an assistant turn — which the suite confirms is accepted upstream. A leading `system` head is the one case the Messages API rejects, and it is now correctly skipped (dropped) rather than stranded.
- **Regression behavior**: `kept[0].role === "assistant"` is asserted, and the new sparse-user test proves truncation lands near target (`>200k` of `250k`) instead of dropping the whole conversation.
- **Public API (Rule 5)**: `truncateHistoryForBudget` signature and return type are unchanged; no external caller is affected.
- **Security**: No credentials, no logging of provider params, no injection surface introduced.
- **Hot paths**: `baseProvider.ts`, `providerRegistry.ts`, and type surfaces untouched.
- **Tests (Rule 15)**: New coverage added to the existing end-to-end proxy-context suite (`continuous-test-suite-proxy-context-preflight.ts`), driving the same `prepareProxyRequestContext` path; determinism declarations unchanged.

No inline comments posted — I found no CRITICAL/MAJOR/MINOR defects worth flagging. The only inherent trade-off is that a dropped head `system` directive's framing is discarded, which is the intended truncation behavior; promoting such a directive to the top-level `system`/`systemPrompt` param would be a possible polish but is out of scope for this fix.

Tara-ag (approved):
Approved. `isDirectiveTurn` correctly limits what may open a kept history to `role: "system"` messages, fixing over-truncation in sparse-user/agentic histories while preserving the API-dispatch guarantee. Live-API-verified, thoroughly tested end-to-end, and internal-only — the public `truncateHistoryForBudget` surface is unchanged.

PR #1766, opened 2026-09-20
fix(proxy): pin each bridged Codex conversation to one prompt cache

murdore:
OpenAI caching is automatic: a prefix is reused only when the request lands on
infrastructure already holding it, and prompt_cache_key is what pins a
conversation to one. The bridged fallback builds its own Codex Responses
payload and never set it, so every turn took its chances on routing -- the
worst case for a hop that arrives in bursts separated by long gaps.

Measured over 36h on the live proxy, split by path:

  /v1/messages (bridged)         gpt-5.6-sol  228,334,264 input   28.2% cached
  /backend-api/codex/responses   gpt-5.6-sol   16,238,877 input   48.7% cached
  /backend-api/codex/responses   gpt-6-astra  247,767,341 input   49.3% cached

The native path, where the Codex client supplies its own key, caches at ~49%.
The bridged path, which sends none, caches at 28.2% -- and it carries the
volume: 2,623 of 2,770 gpt-5.6-sol requests.

Claude Code carries its session id inside metadata.user_id, next to account
and device identifiers. Only the session id is read, and it is hashed, so
nothing identifying reaches the upstream. When no session id can be recovered
the field is omitted rather than filled with something shared: a key common to
every conversation would pin unrelated prefixes onto one cache instead of
separating them.

The field is accepted by this backend -- it appears in the official Codex
client binary -- which matters because the ChatGPT Codex endpoint rejects
unknown fields, as store: false already documents.

## Summary by CodeRabbit

* **Performance**
  * Improved Codex fallback prompt caching by consistently associating requests from the same conversation with a stable cache key.
  * Prevented sensitive session identifiers from being exposed in outbound requests.
* **Bug Fixes**
  * Requests without valid conversation metadata no longer receive an invalid or shared cache key.
* **Tests**
  * Added coverage for cache-key stability, separation between conversations, privacy, and missing or invalid metadata.


Tara-ag:
## Verdict: **APPROVE**

The change pins each bridged Codex conversation to one OpenAI prompt cache by deriving a stable SHA-256 hash of the Claude session id. It is correct, secure, and covered by tests. The single open item is a minor cache-efficiency edge case already flagged inline by CodeRabbit; it does not block merge.

### Findings

| --- | --- | --- |
| MINOR | `src/lib/proxy/codexFallback.ts:320-321` | `session_id ?? parent_session_id` picks an empty/non-string `session_id` and discards a valid `parent_session_id`, so some malformed-primary sessions skip cache routing. Already anchored inline by CodeRabbit (with a suggestion); I am not duplicating it. |

No other findings.

### Checked clean

- **Security** — `codexPromptCacheKey` reads only `metadata.user_id`, hashes the recovered id, and forwards only the 64-char digest. `convertClaudeRequestToCodex` does not pass `body.metadata` through (it rebuilds the request object), so no account/device identifier reaches the upstream. The docstring's no-leak claim holds.
- **Rule 1 (dynamic imports)** — `import { createHash } from "node:crypto"` is a stdlib static import, not a provider; no circular-dependency risk.
- **Rule 4 (CLI ≠ SDK)** — `codexPromptCacheKey` is used only inside `convertClaudeRequestToCodex`, the fallback proxy path. No CLI/SDK leak.
- **Rule 5 (backward compatibility)** — a new optional `prompt_cache_key` field on the fallback request type is only set when derived; existing callers and the `CodexResponsesRequest` shape are unchanged.
- **Blast radius** — change is contained to `codexFallback.ts`, `types/codex.ts`, and its test suite; no shared/hot-path surface touched.
- **Tests** — cover key stability across calls, identifier redaction, and omission for invalid metadata.

The only gap (empty `session_id` with valid `parent_session_id`) is the CodeRabbit MINOR above and warrants the suggested fix before the fallback covers that session class.

Tara-ag (approved):
Approving. The prompt-cache-key change is correct and secure: only a SHA-256 hash of the Claude session id is forwarded, and the raw metadata never reaches the Codex upstream. See the summary comment for the full check list. One minor edge case (empty `session_id` with a valid `parent_session_id`) is already tracked inline by CodeRabbit and is not blocking.

Issue #1744, opened 2026-09-19
ci: the documented env-strip does not strip anything for suites that import the SDK — 30 Extended Suites were admitted on that basis


murdore:
Found while root-causing a CI-only failure on #1735. The bug there is separate; this is about the **method used to vet suites into CI**.

## The claim in `ci.yml`

`.github/workflows/ci.yml`, in the `Run extended suites` step, justifies the list growing from 8 to 38:

> The 30 added were not picked by reading them: every suite script that CI never invoked was run with the environment fully stripped — `env -i` plus `DOTENV_CONFIG_PATH=/dev/null`, **checked to actually suppress .env** — and only the ones that passed that way are here.

## That strip does not work for anything importing `dist/index.js`

`src/lib/neurolink.ts:13-14` runs at module top level:

```ts
const { config: dotenvConfig } = await import("dotenv");
dotenvConfig({ quiet: true });
```

A bare `config()` loads `.env` from cwd and **ignores `DOTENV_CONFIG_PATH`** — only the `dotenv/config` preload honours that variable. `test/helpers/harness.ts:29` uses `import "dotenv/config"`, which is why the strip appears to work when you check it at preload time. Then the suite imports the SDK and `.env` is loaded anyway.

Reproduction:

```
$ env -i HOME=$HOME PATH=$PATH CI=true DOTENV_CONFIG_PATH=/dev/null node probe.mjs
BEFORE import dist: UNSET
AFTER  import dist: SET(len=32)
```

where `probe.mjs` reads `process.env.MISTRAL_API_KEY`, imports `dist/index.js`, and reads it again.

## Why this matters

The 30 suites were admitted on evidence that they pass **with no credentials**. They were actually run **with the developer's full `.env` present**. So the property that was verified is not the property CI depends on.

Concretely, a suite can be admitted because a real key made it pass, and then behave differently on a runner that genuinely has none. That is not hypothetical — `test:provider-descriptors` does exactly this today, and is red in `Extended Suites (non-blocking) 4/4` on #1735 while passing locally. I tried three separate local strips before finding out why none of them reproduced it.

The worse case is a suite that issues **live calls** during vetting, passes because the calls succeed, and is admitted as credential-free. The step comment elsewhere documents `test:middleware` doing precisely that — seven live calls into a tightening Google throttle — which is the failure mode this vetting was meant to exclude.

## This is the repo's recurring failure shape

[...]

PR #1767, opened 2026-09-20
fix(config): honour DOTENV_CONFIG_PATH in the implicit .env load

murdore:
Closes **https://github.com/juspay/neurolink/issues/1744**.

### The defect: a documented strip that stripped nothing

This repo documents, and relies on, one way to run a suite with a clean environment:

```
env -i HOME=… PATH=… CI=true DOTENV_CONFIG_PATH=/dev/null …
```

It does nothing for any suite that imports the built SDK. dotenv reads `DOTENV_CONFIG_PATH` only in its `dotenv/config` **preload entry**; a direct `config()` call ignores it. Both implicit load sites made a direct call — `src/lib/neurolink.ts` at module load and `src/cli/index.ts` at CLI boot — so importing `dist/index.js` loaded `.env` from the working directory regardless, and every credential the strip was meant to remove was present at the read site.

A suite that passes only because a real key leaked in is indistinguishable from one that needs no key at all.

**The original vetting check was real and was actually run.** It confirmed the environment was stripped at **preload**, which was true. The environment at the point the code under test *reads* it — the property that mattered — was never measured.

Found while diagnosing #1735, where `test:provider-descriptors` failed in CI and no local run could reproduce it: the local runs had mistral's real key the whole time.

### The fix

One shared helper, `src/lib/utils/dotenvBootstrap.ts`, used by both sites. It passes dotenv's `path` option when `DOTENV_CONFIG_PATH` is set and omits it otherwise.

**Additive by construction.** With the variable unset — every existing caller, SDK consumers included — behaviour is exactly what it was: `.env` is loaded from the working directory. Only a caller that sets the variable sees a difference, and what it sees is dotenv's own documented meaning for it.

Verified in both directions, in-process and through the built CLI:

```
DOTENV_CONFIG_PATH=/dev/null   after importing dist: all keys UNSET
unset                          after importing dist: keys SET, as before
CLI, stripped                  every provider "Not configured"
CLI, unstripped                configured providers still detected
```

Sharing one helper also closes the drift that let this happen twice: the two sites had already diverged in how each suppressed dotenv's banner, and neither honoured the path.

### Testing

Two cases in `test/continuous-test-suite-credentials.ts`, already wired into Extended Suites. Each runs the shipped entry point in a **child process** with its own working directory and its own `.env`, because the load happens once per process and cannot be re-observed after the first import.

- **6.1** pins that the strip suppresses the load.
- **6.2** pins that the default still loads it, so 6.1 cannot be satisfied by simply deleting the load.

Both were watched against the pre-fix behaviour before being kept: **6.1 fails, 6.2 passes**, and the built artifact was restored byte-identically afterwards.

### Sweep: does any suite actually depend on the leak?

All 44 suites in the Extended Suites list were run under a strip **verified at the read site immediately before and after the batch**, with a throwaway `HOME` so `~/.neurolink` credential stores could not substitute for `.env`. Receipt: 44/44 executed. Four fail, and **none is a credential or dotenv dependency**:

| suite | real HOME + real `.env` | empty HOME, stripped | MCP client timeouts |
|---|---|---|---|
| `test:tts:unit` | rc=124 / 420s | rc=124 / 421s | 3 → 4 |
| `test:vendor-recovery` | rc=124 / 421s | rc=137 / 554s | 1 → 10 |
| `test:vertex-claude-characterization` | **rc=0** / 170s | rc=124 / 542s | 0 → 9 |
| `test:vertex-loop-characterization` | **rc=0** / 224s | rc=124 / 722s | 0 → 14 |

Shared cause: repeated 60s `Failed to create MCP client` timeouts starting the tracked `.mcp-config.json` `filesystem` server via `npx`. A throwaway `HOME` has no `~/.npm/_npx` cache, so every start goes cold and burns the full client timeout.

Ruled out as credential-related three independent ways:

1. **Zero** credential-shaped errors (`invalid api key|authentication|credentialsprovider|OAuth token`) in all four logs.
2. `tts:unit` and `vendor-recovery` fail **identically with full credentials present**, so credentials are not the variable.
3. The two vertex suites **pass in CI, which has no credentials at all** — all four Extended Suites shards are green, and the shard loop assigns each of the 44 to exactly one shard.

**Deliberately out of scope:** `ci.yml` is untouched and no suite is removed from any list. Whether to act on those four is the owner's call, not this PR's.

`check` · `lint` · `build` · `docs:api` regenerated (line-number shifts in `NeuroLink.md`) · single commit.


Tara-ag (approved):
## Review: fix(config): honour DOTENV_CONFIG_PATH in the implicit .env load

Approving. The change extracts the duplicate implicit `.env` load into one shared helper (`src/lib/utils/dotenvBootstrap.ts`) and makes it honour `DOTENV_CONFIG_PATH` while staying byte-for-byte behaviour-preserving when the variable is unset (which is every existing caller, SDK consumers included). Both call sites (CLI boot and SDK module load) now converge on one implementation, closing the drift that let the strip silently do nothing.

The regression tests are correct end-to-end: each spawns the shipped `dist/index.js` in a child process with its own working directory and `.env`, pinning both directions (strip suppresses the load in 6.1; default still loads it in 6.2). No credential leaks, no dynamic-import (Rule 1) concern, no public-API change (Rule 5 backward compatible by construction).

Tara-ag:
## APPROVE

A clean, well-scoped fix. It hoists the duplicated implicit `.env` load into one shared helper that honours `DOTENV_CONFIG_PATH`, while remaining behaviour-preserving for every caller that leaves the variable unset. The change is additive by construction, backward compatible, and carries tests that pin both directions.

### Findings

| Severity | file:line | Finding |
| — | — | No issues found. |

None of the changed lines warranted an inline comment.

### What was checked (all clean)

- **`src/lib/utils/dotenvBootstrap.ts`** (new) — `loadProcessDotenv()` only passes dotenv's `path` option when `DOTENV_CONFIG_PATH` is set, and omits it otherwise, keeping the default `.env`-from-cwd behaviour byte-identical. Banner suppression (`DOTENV_CONFIG_QUIET`) is preserved.
- **`src/cli/index.ts`** — replaces its inline dotenv try/catch with the shared helper; the unconditional `DOTENV_CONFIG_QUIET="true"` at the top of the file still stands, so CLI output behaviour is unchanged.
- **`src/lib/neurolink.ts`** — same shared helper, called at the first module-body statement, preserving the original load ordering. Static import (hoisted) adds no circular dependency — `dotenvBootstrap` imports nothing back into `neurolink`.
- **`test/continuous-test-suite-credentials.ts`** — two regression cases (6.1 / 6.2) that spawn the shipped `dist/index.js` in a child process with an isolated cwd + `.env`, so the once-per-process load is observable. End-to-end only (Rule 15); uses only dist entry points and node builtins; consistent module graph.
- **Sweep** — PR author documents a 44-suite Extended-Suites sweep under a verified strip, with the four failures ruled out as credential-related. `ci.yml` deliberately untouched (out of scope, author-owned).

Coverage, Rule 5 (backward compat), Rule 1 (no static provider import — `dotenv` is a dev dependency, not a provider), and Rule 15 (e2e-only tests) all satisfied. No CRITICAL / MAJOR / MINOR findings.

Tara-ag:
## APPROVE — recurring review, change unchanged since prior approval

This is a recurring review. The current HEAD (`50fe7fb0a…`, the single squash commit `fix(config): honour DOTENV_CONFIG_PATH…`) carries the exact same 5-file diff that was already reviewed and approved (review `5261420208`, summary comment below). I re-fetched the full diff and confirmed:

- **No changes since the prior review.** Same files (`docs/api/classes/NeuroLink.md`, `src/cli/index.ts`, `src/lib/neurolink.ts`, the new `src/lib/utils/dotenvBootstrap.ts`, `test/continuous-test-suite-credentials.ts`), same content. The commit-id difference vs. the prior review is a squashed-history artifact, not a content change.
- **No open or unresolved review threads** (`get_review_comments` → 0 threads). Nothing to resolve or re-litigate.

### Independent re-verification (small, done directly)
- `dotenvBootstrap.loadProcessDotenv()` only passes dotenv's `path` when `DOTENV_CONFIG_PATH` is set and otherwise omits it, keeping the default `.env`-from-cwd behaviour byte-identical — additive by construction.
- Banner suppression (`DOTENV_CONFIG_QUIET`) is preserved; the helper dynamically imports `dotenv` inside try/catch, so the dev-dependency-only production behaviour is unchanged.
- Both call sites (`cli/index.ts`, `neurolink.ts`) converge on the shared helper; the static import introduces no circular dependency, and the top-level `await` is valid ESM.
- The regression tests (6.1 / 6.2) spawn the shipped `dist/index.js` in isolated child processes pinning both directions — strip suppresses (6.1), default still loads (6.2) — so 6.1 cannot be satisfied by simply deleting the load.

No CRITICAL / MAJOR / MINOR findings; none of the changed lines warrant an inline comment. The prior APPROVE stands.

PR #1768, opened 2026-09-20
fix(yama-review): fail the job when a run posts no review

### LINKED ISSUE #1754: Yama PR Review reports success after reviewing nothing — its GitHub reads 404 against juspay/juspay
Found on #1746 (credit: the TTS cluster, which nearly reported it as "passed with no findings" and checked instead).

## A green check that means no review happened

`Yama PR Review` reported `COMPLETED/SUCCESS` on https://github.com/juspay/neurolink/pull/1746, with **no title, no summary, zero reviews and zero inline comments** on the PR.

It did not skip — neither preflight path (`Fork PR — secrets unavailable`, `Missing YAMA_GITHUB_TOKEN`) fired. It ran for ~33 minutes (17:22→17:55) and made hundreds of tool calls. Its own instructions require inline comments plus one summary comment, and to finish only when "the PR carries one clean, complete review… and a review state that matches the verdict". It produced none of that and logged no verdict.

## Cause: the repository slug is wrong

Job `105936281726` made seven requests of this shape, all 404:

```
MCP error 0: failed to get issue comments:
GET https://api.github.com/repos/juspay/juspay/issues/1746/comments: 404 Not Found
```

`juspay/juspay` — the owner duplicated as the repository name. It was reading a repository that does not exist, found nothing, and exited 0.

## Scope: intermittent, NOT global

The initial hypothesis was that this degrades review on every PR. **It does not** — checked across recent PRs by counting real inline comments from the reviewer identity:

| PR | Yama check | inline review comments |
|---|---|---|
| #1647 | success | 7 |
| #1740 | success | 6 |
| #1733 | success | 3 |
| #1739 | in_progress | 4 |
| **#1746** | **success** | **0** |

So Yama works on most runs and produced substantive, correct findings on those PRs today — including one this repository acted on and one it justifiably declined. #1746 is the outlier. Whatever produces the bad slug is per-run, not a static misconfiguration, which makes it harder to spot and is the reason this needs a guard rather than only a fix.

## Why this matters more than one missed review

A green `Yama PR Review` is currently indistinguishable between "reviewed and found nothing" and "reviewed nothing". The first is a signal; the second is silence wearing the same colour. Any PR merged on the strength of a green Yama may have had no review at all, and nothing on the PR would show it.

This is the failure mode CLAUDE.md already records — an exit code mistaken for a result — appearing in a new place. The existing list covers `CANCELLED`, missing contexts, empty `gh pr checks` output, a redirect failing before the tool ran, and a verifier sharing a stale input. This is the same family: the tool ran, reported success, and did nothing.

## Suggested fix

1. **Make the job fail when it produces no review.** If Yama completes without posting a review or a summary comment, that is an error, not a success. This is the important half — it converts silent non-review into a visible failure regardless of cause.
2. Find the slug bug. The duplication pattern (`<owner>/<owner>`) suggests an owner value being used where `owner/repo` is expected, or a repo variable that is empty and falls back to the owner.
3. Assert the target repository once at startup and abort on a 404, rather than making seven failing reads and continuing.

Point 1 stands on its own even if 2 takes longer.

murdore:
## Summary

Closes #1754 (partial — see scope note below).

`Yama PR Review` can exit `0` without posting anything at all. On #1746 it ran for ~33 minutes, made hundreds of tool calls, and reported `COMPLETED/SUCCESS` with zero title, zero summary, zero reviews and zero inline comments — because Yama's own internal GitHub client was reading against `juspay/juspay` (owner duplicated as repo name) instead of `juspay/neurolink`, got 404s on every call, and silently treated the empty result as "nothing to report" instead of failing.

This PR adds a guard step, `Guard against a silent no-op review`, to `.github/workflows/yama-review.yml`, after the `Yama review` step:

- A new `Record review start time` step captures a timestamp immediately before Yama runs.
- The guard step re-reads the PR's reviews via `gh api repos/${REPOSITORY}/pulls/${PR_NUMBER}/reviews` using the **job's own `GITHUB_TOKEN`** — a reader independent of whatever repo-slug bug exists inside Yama's internal GitHub MCP client, so it can't inherit the same misconfiguration.
- It fails the job (`exit 1`, with an `::error::` annotation) if no review with a verdict (`APPROVED` or `CHANGES_REQUESTED`) and a non-empty summary body was posted on the PR since the run started.
- Zero *inline* comments alone is not treated as a failure — a clean PR can legitimately be approved with no findings. Zero *reviews* is, since Yama's own prompts (`.yama/prompts.json`) require finishing with exactly one summary review whose state matches the verdict.

## Scope

This is deliberately narrow to what's fixable from this repo's side. **The actual repo-slug bug lives inside `@juspay/yama`'s own GitHub client construction (a separate package this repo does not own)** — that is a known follow-up, not attempted here. This PR only makes the symptom (silent success) impossible to merge on quietly; it converts "reviewed nothing, reported success" into a visible, failing check, per issue #1754's suggested fix point 1 ("Make the job fail when it produces no review... the important half").

## ⚠️ CI will show 0/5 required checks on this PR — expected, not a bug

This PR modifies a file under `.github/workflows/` (`yama-review.yml`). Empirically confirmed this session (two consecutive pushes touching a workflow file on this repo registered **zero** `pull_request`-triggered runs, including `ci.yml`): PRs that touch `.github/workflows/` never get `pull_request`-triggered CI on this repo. So this PR will show **0/5 required checks ever registering** — that's expected given this constraint, not something to work around by finding another way to land the change.

Since CI can't validate the workflow YAML's own syntax here, I validated locally instead:

- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/yama-review.yml'))"` — parses cleanly.
- Re-parsed with PyYAML and printed the job's step list to confirm ordering: `Preflight (fork + secrets)` → `Checkout` → `Setup Node.js` → `Set up uv` → `Pre-build the code graph` → `Record review start time` → `Yama review` → `Guard against a silent no-op review` → `Archive the run log`. `Archive the run log` keeps its original `if: always()` so it still archives the log on a guard-triggered failure.
- `yamllint` and `actionlint` are not available in this environment (`pipx run yamllint` timed out fetching from a network-constrained sandbox) — could not run either.
- Manually verified `${{ }}` expression balance (16 opens / 16 closes) and read through the diff for `if:` condition correctness (all new steps gated on `steps.gate.outputs.run == 'true'`, matching the existing pattern in the file) and `${{ }}` expression syntax.
- Confirmed `gh` CLI is preinstalled on `ubuntu-latest` GitHub-hosted runners (used in the new guard step), and that the job's existing `permissions: pull-requests: write` (superset of read) covers the new `pulls/.../reviews` read.

Given CI won't run, please double-check the YAML by eye before merging — a syntax error here would not be caught before merge.

## Follow-up (out of scope here)

The underlying repo-slug bug inside `@juspay/yama`'s GitHub client construction (producing `juspay/juspay` instead of `owner/repo`) still needs a fix in that package — this PR only adds the guard, not that fix. Tracked in #1754.

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(...)"` parses the workflow cleanly.
- [x] Manual diff review of step ordering, `if:` conditions, and `${{ }}` expression syntax.
- [ ] CI cannot validate this PR (workflow-file changes don't trigger `pull_request` CI on this repo — see above). Needs manual review before merge.
- [ ] Not yet exercised against a live PR — no way to trigger `pull_request` CI here; would need to be observed on the next real Yama run after merge.

## Summary by CodeRabbit

- **Bug Fixes**
  - Review checks now detect and fail when an automated review completes without providing a decision or summary.
  - Review validation records when the process begins, helping ensure only results from the current review are evaluated.

- **Documentation**
  - Added explanatory guidance for the review validation behavior.


Tara-ag (commented):
Reviewing the silent-no-op guard. Two moderate robustness findings; the guard logic itself is correct.

.github/workflows/yama-review.yml
```diff
@@ -115,6 +134,34 @@ jobs:
...
+        # since this run started. Zero inline comments alone is not a failure
+        # (a clean PR is legitimately approved with none); zero reviews is.
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPOSITORY: ${{ github.repository }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          STARTED_AT: ${{ steps.started.outputs.at }}
+        run: |
+          set -euo pipefail
+          posted=$(gh api "repos/${REPOSITORY}/pulls/${PR_NUMBER}/reviews?per_page=100" \
```
Tara-ag: `gh api` without `--paginate` returns only the first page of reviews (max 100). The default ordering of the `/reviews` endpoint is reverse-chronological, so this PR's own review will normally be on page 1 and the guard still works — but on a busy PR with >100 historical reviews the assumption is implicit and fragile. Suggest passing `--paginate` so the count never stalls on pagination semantics:

```bash
posted=$(gh api --paginate "repos/${REPOSITORY}/pulls/${PR_NUMBER}/reviews?per_page=100" \
  --jq '[...same...] | length')
```

Also note the `set -euo pipefail` interplay on line 151: if `gh api` fails transiently (rate limit, network), the command substitution makes this step exit nonzero and fails the job loudly **even when a valid review was correctly posted**. That is a deliberate loud-vs-silent trade-off (and acceptable given the goal of #1754), but call it out in the surrounding comment so a future reader isn't surprised.

.github/workflows/yama-review.yml
```diff
@@ -115,6 +134,34 @@ jobs:
...
+            --jq '[.[] | select(
+              .submitted_at != null
+              and .submitted_at >= env.STARTED_AT
```
Tara-ag, line 164: If `STARTED_AT` (`steps.started.outputs.at`) is ever empty, the jq filter `env.STARTED_AT` evaluates to `""`, and `.submitted_at >= ""` is **always true** — so *any* historical review from an earlier `synchronize` would satisfy the filter, `posted` becomes non-zero, and the guard passes trivially. That silently reintroduces the exact no-op this PR is trying to close (#1754). Since `env.STARTED_AT != null` won't catch an empty string, add a bash-side guard before the query, e.g.:

```bash
[ -n "$STARTED_AT" ] || { echo "::error::STARTED_AT missing; cannot verify a review was posted"; exit 1; }
```

Low probability (the `Record review start time` step always emits a value), but the failure mode is silent, which is the opposite of what this guard intends — a cheap one-line defence.

Tara-ag (changes requested):
The starter marker (name: `review comment posted`, line 146) and the `Record review start time` step (lines 143-145) are CORRECT — `Continue on error` + `if: always()` is exactly the right shape (per the project's comment-formatting and error-tolerance conventions) to ensure a silent no-op never squelches a genuinely failing review, and the thread is caught by the ping+close mailbox rather than being left dangling. Two issues below are worth tightening before merge; both are edge cases whose failure mode is *silent*, which is the exact behavior this guard exists to prevent.

_Inline comments:_
- Lines 151-152 — pagination: `gh api` without `--paginate` caps at one page (100 reviews); add `--paginate`. Also note the loud-exit-on-transient-error trade-off should be documented in the comment.
- Line 155 — empty `STARTED_AT` makes `.submitted_at >= ""` always true, trivially passing the guard and reintroducing #1754. One-line `[ -n "$STARTED_AT" ]` defence suggested.

Tara-ag:
**Verdict: APPROVE** — the guard is correct and both earlier robustness findings are now fixed.

|---|---|---|
| RESOLVED | `.github/workflows/yama-review.yml:151-152` | Missing `--paginate` could cap the review count at one page; author added `--paginate` — thread resolved. |
| RESOLVED | `.github/workflows/yama-review.yml:155` | Empty `STARTED_AT` made `.submitted_at >= ""` always true (silent no-op); author added an empty-value guard — thread resolved. |

What was checked and found clean:
- Guard logic: records start time, queries reviews with `GITHUB_TOKEN`, filters for qualifying verdicts with non-empty bodies posted after `STARTED_AT`, exits non-zero when none match — correctly closes the silent no-op of #1754.
- Error handling: `Continue on error` + `if: always()` + ping/close-mailbox keeps genuine review failures loud.
- No secrets, no hot-path (baseProvider/registry/types/auth/mcp) influence, no SDK surface change.

No blocking issues.

Tara-ag:
## Review summary — recurring review of `fix/yama-silent-review-guard`

Reviewed as a recurring pass against the updated diff (`fefe4f9`). Both findings from the earlier review are now addressed in the current diff, and both threads are already resolved — no repost:

1. ✅ **Pagination** — the guard now uses `gh api --paginate ... --jq '[...]'` and explicitly handles the "`--jq` applies per page, not to the combined result" caveat by slurping all page arrays with `| jq -s 'add | length'`. This is the correct idiom; the earlier `--jq '... | length'` per-page-count bug is gone.
2. ✅ **Empty-`STARTED_AT` guard** — there is now an explicit `[ -z "${STARTED_AT}" ]` check that fails closed with an `::error::` annotation, so a missing timestamp can no longer make the recency check trivially pass.

I re-read the current guard logic end to end and found no new factual issues:

- **jq filter parses correctly**: with jq precedence the select reduces to `(submitted_at != null) and (submitted_at >= env.STARTED_AT) and (state == "APPROVED" or state == "CHANGES_REQUESTED") and ((.body|length) > 0)`. The parentheses around the `or` are present, so the `and`/`or` binding is right.
- **`env.STARTED_AT` is visible** to gh's embedded jq because the Actions `env:` block exports it to the shell, which the `gh` subprocess inherits (jq's `env` builtin reads the process environment).
- **Fails closed on API failure**: if `gh api` errors, the pipeline exits non-zero and `set -euo pipefail` fails the job loudly; an empty response body yields a single `[]` page → `length` 0 → job fails. No path silently passes on a broken read.
- **Correctness of the recency window**: `STARTED_AT` is recorded right before Yama runs, so only this run's posted review clears the `>=` check, not a stale review from an earlier synchronize.

Two non-blocking observations only, no action required (flagging for awareness, not as blockers):

- **The guard counts any review with a verdict+summary since `STARTED_AT`, not specifically Yama's.** A human `APPROVE`/`CHANGES_REQUESTED` landing during the run would also satisfy it. Given the goal is "fail a silent no-op run," false *passes* are the failure mode and a human review is still a real review, so this is acceptable — worth noting in case stricter provenance is ever wanted.
- **This repo can't validate the workflow via CI** (workflow-file changes don't trigger `pull_request` runs here, as you documented). The YAML logic reads clean under PyYAML and the shell/jq was hand-traced here, but the guard's behavior on a live run can only be confirmed post-merge on the next real Yama invocation.

No blocking findings. Approving the change as reviewed.

Tara-ag (approved):
Approving — the two robustness findings from the earlier round (missing `--paginate` / pagination cap, and the empty `STARTED_AT` jq edge case) are both addressed in the current diff and both review threads are resolved. The guard itself is correct: it records the start time, queries reviews with `GITHUB_TOKEN`, filters for qualifying verdicts posted after `STARTED_AT`, and fails the job when no such review exists — closing the silent no-op of #1754 while keeping genuine review failures loud via the `Continue on error` + `if: always()` + ping/close-mailbox pattern. No new issues found in this pass.

PR #1772, opened 2026-09-21
fix(proxy): convert the system turns Vertex refuses so the fallback survives

murdore:
fix(proxy): convert the system turns Vertex refuses so the fallback survives

Every Vertex fallback on real traffic has been failing:

  [proxy] fallback -> vertex/claude-opus-4-6
  [proxy] fallback vertex/claude-opus-4-6 failed: responded 400:
      role 'system' is not supported on this model

Vertex refuses role:"system" messages outright, where the first-party Messages
API accepts them anywhere except index 0. Claude Code interleaves them as
positional notices -- a file changed on disk since it was last read, a tool
became available, an MCP server dropped -- so every agentic request carries
them. One captured 410-message conversation had 19.

The hop then falls through to Codex, and with Codex rate-limited the request
dies as a 429 the client sees. In six hours 77 requests failed that way, 8.9%
of all traffic, on the only fallback available.

They cannot be hoisted into the top-level `system` parameter: their meaning
depends on where they sit. Dropping them would silently lose context the turn
depends on. They become user turns instead, which keeps the text and the
position. Consecutive same-role turns are accepted, so the sequence stays valid.

Verified against live Vertex with the real captured body that was failing:
the payload is accepted (HTTP 200, 163,738 tokens cached), and restoring only
the 19 system roles onto that same payload reproduces the exact production
error -- so the attribution is causal, not circumstantial.

## Summary by CodeRabbit

- **Bug Fixes**
  - Vertex Anthropic requests now handle system-role messages by converting them to user-role messages while preserving message order and content.
  - Top-level system instructions remain unchanged.
  - Message histories without system-role turns continue to be preserved as-is.

- **Tests**
  - Added coverage for interleaved system messages and unchanged message histories.


Tara-ag:
## ✅ APPROVE

The Vertex Anthropic fallback change is clean, self-contained, and correctness-focused. `role: "system"` turns Vertex rejects are converted to `role: "user"` with position and content preserved, and the top-level `system` channel is left untouched. No findings.

### Findings

| — | — | None — nothing blocking or worth reporting |

### What was checked

- **Logic**: `normalizeVertexRoles` is a pure function — returns the original array reference when nothing changes (verified by the "untouched history" test), and produces new per-message objects only when a `system` role is present. Position of interleaved notices is preserved, and consecutive same-role turns are accepted by the API, so the converted sequence is never invalid.
- **Call path**: guarded by `"messages" in payload` and applied before the `anthropic_version` spread — no risk to the non-changed branch, no mutation of the caller's body.
- **Rule compliance**: no circular-dependency import (rule 1), no Gemini/Vertex structured-output or max-tokens change (rule 3), no CLI ↔ SDK leak (rule 4), no public SDK API change (rule 5), no `formatProviderError` change (rule 6), and the added tests follow the end-to-end continuous-test-suite pattern against the built CLI (rule 15).
- **Security**: no credentials, no logging of provider params, no injection surface introduced.
- **Blast radius**: the change is local to `src/lib/proxy/vertexAnthropicFallback.ts`; `buildVertexAnthropicPayload`'s existing callers (the Vertex fallback path) are unchanged in signature and return shape. No out-of-diff callers are affected.
- **Tests**: the two added cases exercise the conversion, ordering/content preservation, top-level `system` integrity, and the no-copy fast path.

Review state: approving review logged on the PR.

Tara-ag (approved):
The Vertex Anthropic fallback fix is clean and correct: `role: "system"` turns Vertex rejects are converted to `role: "user"` with position and content preserved, the top-level `system` channel is untouched, and the pure function avoids needless copies (verified by the no-copy fast-path test). No findings. Approving.

